PR #3, opened 2021-12-07
feat(auth): start building a foundation for an auth lib

codyoss:


auth/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+            return Ok(source);
+        }
+
+        Err(Error::Other("unable to detect default credentials".into()))
+    }
+
+    /// Creates a source from a file type credential such as a Service Account
+    /// Key file or a gcloud user credential.
+    fn file_source(
+        file_path: String,
```
SurferJeffAtGoogle, line 140: File systems can have file names that are not valid UTF-8, and therefore cannot be represented as a Rust String.
This code could never open such a file.  To fix, the type of `file_path` should match the type of the argument to `std::fs::read()`.  See https://doc.rust-lang.org/std/fs/fn.read.html.

auth/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+        config: CredentialConfig,
+    ) -> Result<Box<dyn Source + Send + Sync + 'static>> {
+        let contents = std::fs::read(file_path)?;
+        let file: Key = serde_json::from_slice(&contents)?;
+        let source: Box<dyn Source + Send + Sync + 'static> = match file.cred_type {
+            "authorized_user" => {
+                let source = UserSource::from_file_contents(&contents)
```
SurferJeffAtGoogle, line 147: Where is `UserSource` defined?

SurferJeffAtGoogle (changes requested):

auth/src/lib.rs
```diff
@@ -0,0 +1,214 @@
+//! The `google-cloud-auth` crate provides a convenient way of fetch credentials
```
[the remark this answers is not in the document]
codyoss: Done

auth/src/lib.rs
```diff
@@ -0,0 +1,214 @@
...
+    /// The actual token.
+    pub value: String,
+    // TODO(codyoss): Token leaks chrono lib, should we have our own type or is this okay...
+    //                We could just use a unix int64 timestamp from zulu?
+    /// The time when a token expires, if known.
+    pub expires: Option<DateTime<Utc>>,
+}
+
+impl AccessToken {
+    pub(crate) fn is_valid(&self) -> bool {
```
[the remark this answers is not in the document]
codyoss: Went with `needs_refresh` for now.

auth/src/lib.rs
codyoss: Switched to `PathBuf` for now, or would you prefer a `AsRef<Path>`?

auth/src/lib.rs
codyoss: `source.rs`

auth/src/metadata.rs
```diff
@@ -0,0 +1,139 @@
...
+
+use super::{Error, Result};
+
+// TODO(codyoss): cache a client
+// TODO(codyoss): move client impl elsewhere
+// TODO(codyoss): Instead of vecs things could be slices... don't need ownership
+// TODO(codyoss): funcs could take &str or impl Into<String>?
+
+/// A wrapper around a HTTP client used to talk to a Google Cloud metadata service.
+struct MetadataClient {
```
SurferJeffAtGoogle, line 29: I see no reason for this struct to exist.  How about a factory function instead:
```
fn newMetadataClient() -> reqwest::Client {
```

auth/src/metadata.rs
```diff
@@ -0,0 +1,139 @@
...
+    }
+}
+
+/// Makes a request to the supplied metadata endpoint.
+///
+/// ```rust
+/// let project_id = get("project/project-id")?;
+/// println!("{}", project_id);
+/// ```
+// TODO(codyoss): I wonder if this should be a pathbuf
```
SurferJeffAtGoogle, line 52: I think String is fine here.

auth/src/metadata.rs
```diff
@@ -0,0 +1,139 @@
...
+#[allow(dead_code)]
+pub async fn get(suffix: String) -> Result<String> {
+    get_with_query::<()>(suffix, None).await
+}
+
+async fn get_with_query<T: Serialize + ?Sized>(
+    suffix: String,
+    query: Option<&T>,
+) -> Result<String> {
+    let host = env::var("GCE_METADATA_HOST")
```
SurferJeffAtGoogle, line 62: Please insert a comment with a link to where this environment variable and IP address are documented.

auth/src/metadata.rs
```diff
@@ -0,0 +1,139 @@
...
+        .unwrap_or_else(|_| -> String { String::from("169.254.169.254") });
+    let suffix = suffix.trim_start_matches('/');
+    let url = format!("http://{}/computeMetadata/v1/{}", host, suffix);
+
+    // TODO(codyoss): retry
+    let req = MetadataClient::new().c.get(url);
+    let req = if let Some(query) = query {
```
SurferJeffAtGoogle, line 69:
```
let req = query.map(|q| req.query(q));
```

auth/src/metadata.rs
```diff
@@ -0,0 +1,139 @@
...
+    Ok(content)
+}
+
+/// Checks the environment to determine if code is executing in a Google Cloud
+/// environment.
+pub async fn on_gce() -> bool {
+    if std::env::var("GCE_METADATA_HOST").is_ok() {
+        return true;
+    }
+    let client = MetadataClient::new();
```
SurferJeffAtGoogle, line 91: In a comment, please explain how/why the strategy for detecting gce.

auth/src/metadata.rs
```diff
@@ -0,0 +1,139 @@
...
+/// The result of requesting a token from the metadata service.
+#[derive(Deserialize)]
+pub struct Token {
+    pub access_token: String,
+    pub expires_in: i64,
+}
+
+/// Fetches a [AccessToken] from the metadata service with the provided scopes. If an
+/// account is not provided the value will be set to `default`.
+pub async fn access_token(account: Option<&str>, scopes: Vec<String>) -> Result<Token> {
```
SurferJeffAtGoogle, line 119: Please add a verb to the function name.

auth/src/metadata.rs
```diff
@@ -0,0 +1,139 @@
...
+            res.status().as_str()
+        )));
+    }
+    let content = res.text().await?;
+    Ok(content)
+}
+
+/// Checks the environment to determine if code is executing in a Google Cloud
+/// environment.
+pub async fn on_gce() -> bool {
```
SurferJeffAtGoogle, line 87: Would you mind a longer name like `im_running_on_gce()`?

auth/src/oauth2.rs
```diff
@@ -0,0 +1,63 @@
...
+    pub(crate) exp: Option<i64>,
+    pub(crate) iat: Option<i64>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) typ: Option<&'a str>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub(crate) sub: Option<&'a str>,
+}
+
+impl JwsClaims<'_> {
+    pub(crate) fn encode(&mut self) -> Result<String> {
```
SurferJeffAtGoogle, line 35: I'm very surprised that the `encode()` method mutates the struct.

auth/src/source.rs
```diff
@@ -0,0 +1,474 @@
...
+
+/// Options for building a [ServiceAccountKeySource].
+pub struct ServiceAccountKeySourceBuilder<'a> {
+    file_path: Option<PathBuf>,
+    contents: Option<&'a [u8]>,
+
+    scopes: Vec<String>,
+}
+
+impl ServiceAccountKeySourceBuilder<'_> {
```
SurferJeffAtGoogle, line 52: I propose we write all new code without builders; which means dropping the builders from this file too.

I have clearly documented the negative effects of writing builders in an earlier conversation.  If there are benefits I haven't considered, please share them.

auth/src/source.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+}
+
+/// A [Source] derived from a Service Account Key.
+pub struct ServiceAccountKeySource {
+    file: ServiceAccountKeyFile,
+    scopes: Vec<String>,
+}
+
+/// A representation of a Service Account File.
```
SurferJeffAtGoogle, line 81: If there is a public document describing the format of the key file, please link to it here.

auth/src/source.rs
```diff
@@ -0,0 +1,474 @@
...
+    }
+    pub fn from_file_contents(contents: &[u8]) -> ServiceAccountKeySourceBuilder {
+        ServiceAccountKeySourceBuilder {
+            file_path: None,
+            contents: Some(contents),
+            scopes: Vec::new(),
+        }
+    }
+
+    async fn _access_token(&self) -> Result<AccessToken> {
```
SurferJeffAtGoogle, line 110: Add a function comment.

auth/src/source.rs
```diff
@@ -0,0 +1,474 @@
...
+            .await
+            .map_err(|_| Error::Other("unable to decode response".into()))?;
+
+        Ok(AccessToken {
+            value: token_response.access_token,
+            expires: Some(Utc::now() + Duration::seconds(token_response.expires_in)),
+        })
+    }
+
+    fn signer(&self) -> Result<Box<dyn Signer>> {
```
SurferJeffAtGoogle, line 140: Add comment for function.

auth/src/source.rs
```diff
@@ -0,0 +1,474 @@
...
+        let pk = rustls_pemfile::read_one(&mut self.file.private_key.as_bytes())?
+            .ok_or_else(|| Error::Other("".into()))?;
+        let pk = match pk {
+            Item::RSAKey(item) => item,
+            Item::PKCS8Key(item) => item,
+            _ => return Err(Error::Other("expected key in a different format".into())),
```
SurferJeffAtGoogle, line 146: In the error message, list the format provided and the ones that would have been accepted.

auth/src/source.rs
SurferJeffAtGoogle: Add a verb to the function name.

auth/src/source.rs
SurferJeffAtGoogle: Add a verb to the function name.

auth/src/source.rs
```diff
@@ -0,0 +1,474 @@
...
+        };
+        rustls::sign::RsaSigningKey::new(&rustls::PrivateKey(pk))
+            .map_err(|_| Error::Other("unable to create signer".into()))?
+            .choose_scheme(&[rustls::SignatureScheme::RSA_PKCS1_SHA256])
+            .ok_or_else(|| Error::Other("invalid signing scheme".into()))
+    }
+
+    fn payload(&self, signer: Box<dyn Signer>) -> Result<String> {
```
SurferJeffAtGoogle, line 154: Add a function comment and a verb to the function name.

auth/src/source.rs
```diff
@@ -0,0 +1,474 @@
...
+            .ok_or_else(|| Error::Other("".into()))?;
+        let pk = match pk {
+            Item::RSAKey(item) => item,
+            Item::PKCS8Key(item) => item,
+            _ => return Err(Error::Other("expected key in a different format".into())),
+        };
+        rustls::sign::RsaSigningKey::new(&rustls::PrivateKey(pk))
+            .map_err(|_| Error::Other("unable to create signer".into()))?
+            .choose_scheme(&[rustls::SignatureScheme::RSA_PKCS1_SHA256])
+            .ok_or_else(|| Error::Other("invalid signing scheme".into()))
```
SurferJeffAtGoogle, line 151: In the error message, tell the reader what you expected and what you found.

auth/src/source.rs
```diff
@@ -0,0 +1,474 @@
...
+
+/// Options for building a [UserSource].
+pub struct UserSourceBuilder<'a> {
+    file_path: Option<PathBuf>,
+    contents: Option<&'a [u8]>,
+
+    scopes: Vec<String>,
+}
+
+impl UserSourceBuilder<'_> {
```
SurferJeffAtGoogle, line 211: Apply my comments from ServiceAccountKeySource to the other sources.

auth/src/source.rs
```diff
@@ -0,0 +1,474 @@
...
+            source: Box::new(NoOpSource {}),
+        }
+    }
+}
+
+#[async_trait]
+impl Source for RefresherSource {
+    async fn token(&self) -> Result<AccessToken> {
+        let mut cur_token = self.current_token.lock().await;
+        if cur_token.needs_refresh() {
```
SurferJeffAtGoogle, line 394: The condition looks inverted.  Should this be `if !cure_token.needs_refresh() {`?

SurferJeffAtGoogle: I expect either that I'm misinterpreting the logic or a test to fail.

auth/Cargo.toml
```diff
@@ -0,0 +1,23 @@
+[package]
+name = "google-cloud-auth"
+version = "0.1.0"
+authors = ["Google LLC"]
+description = "An auth library for Cloud Google APIs."
+license = "Apache-2.0"
+repository = "https://github.com/googleapis/google-cloud-rust/auth"
```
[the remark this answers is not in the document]
codyoss: Good catch!

auth/src/metadata.rs
codyoss: named `fetch_access_token`

auth/src/oauth2.rs
codyoss: It does so to provide good defaults as this struct will be directly encoded into a string. It could probably just consume self if you think that is better.

auth/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
+    #[error("unable to read file")]
+    Io(#[from] std::io::Error),
+    #[error("unable to deserialize value")]
+    Serde(#[from] serde_json::Error),
+    #[error("unable to process request")]
+    Http(#[from] reqwest::Error),
+    #[error("{0}")]
+    Other(String),
```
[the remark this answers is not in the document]
codyoss: Do you think error should just be something like the following instead:

```
#[derive(Debug)]
pub struct Error(Box<dyn std::error::Error>);

impl std::fmt::Display for Error {
    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
        write!(f, "{}", self.0)
    }
}
```

I don't think it should really be a public enum right now. Leaks too much imo.

auth/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+            }
+        };
+        Ok(source)
+    }
+
+    /// Returns the path to a gcloud user credential.
+    fn well_known_file() -> Result<PathBuf> {
+        let mut path = PathBuf::new();
+        if cfg!(windows) {
+            if let Ok(appdata) = std::env::var(WINDOWS_APPDATA) {
```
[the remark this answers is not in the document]
codyoss: ack

auth/src/lib.rs
```diff
@@ -0,0 +1,223 @@
...
+                path.push(appdata);
+            } else {
+                return Err(Error::Other("unable to find APPDATA".into()));
+            }
+            path.push("gcloud");
+        } else if let Ok(home) = std::env::var(UNIX_HOME) {
```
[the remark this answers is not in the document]
codyoss: done.

auth/src/source.rs
codyoss: done.

auth/src/source.rs
codyoss: done.

auth/src/source.rs
codyoss: done.

auth/src/source.rs
codyoss: When I re-named I implied the wrong thing. Updated name and tests

auth/src/lib.rs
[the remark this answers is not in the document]
SurferJeffAtGoogle: We agree to drop string types and model our Error after reqwest's error:
https://docs.rs/reqwest/0.7.2/reqwest/struct.Error.html

auth/src/lib.rs
```diff
@@ -0,0 +1,231 @@
...
+            expiresish > now
+        } else {
+            false
+        }
+    }
+}
+
+/// Configuration for various authentication flows.
+#[non_exhaustive]
+pub struct CredentialConfig {
```
SurferJeffAtGoogle, line 78: I now concur we should use builders because:
1. We can't add public members to structs without taking a major version number upgrade.
2. Having a separate builder type with a `.build()` method means we validate the parameters exactly once at build time, instead of validating it at the beginning of each function call.

codyoss: Done.

auth/src/lib.rs
```diff
@@ -0,0 +1,313 @@
...
+        };
+        Ok(source)
+    }
+
+    /// Returns the path to a gcloud user credential.
+    fn well_known_file() -> Result<PathBuf> {
+        let mut path = PathBuf::new();
+        if cfg!(windows) {
+            let appdata = std::env::var(WINDOWS_APPDATA_ENV)
+                .map_err(|_| Error::new("unable to find APPDATA"))?;
```
SurferJeffAtGoogle, line 261: So callers don't have to compare strings, I still think we need an enum in the error type.  We talked about following the pattern in std::io::Error, with ErrorKind:
```
#[non_exhaustive]
pub enum ErrorKind 
```

codyoss: Done.

SurferJeffAtGoogle (approved):

PR #4, opened 2021-12-23
feat(discovery-generator): add generator and gen storage

codyoss:
This needs a more iteration before it is actually ready for review and will also require the auth PR to land first. Opening this up as a draft just for some visibility.


codyoss:
Going to close this PR and raise just the generated surface first, we can double back and make the generator itself better later. I would like to land something the is usable and give the community time to give us some feedback.

PR #5, opened 2022-01-20
feat(storage): add generated storage client

codyoss:
For time sake I think it will be easier to first review the output
of the storage client and then come back and clean up the generator
after.


storage/src/bytes.rs
```diff
@@ -0,0 +1,312 @@
...
+use bytes::BytesMut;
+use reqwest::Response;
+use tokio::fs::File;
+use tokio::io::AsyncReadExt;
+
+use super::{Error, Result};
+pub use bytes::Bytes;
+
+#[derive(Debug)]
+pub struct BytesReader {
```
codyoss, line 14: Note this file was manually written to be a helper struct for working with file uploads and storage downloads. It is not generated.

storage/src/model.rs
```diff
@@ -0,0 +1,3967 @@
...
+    /// Creates a builder to more easily construct the [Bucket] struct.
+    pub fn builder() -> BucketBuilder {
+        Default::default()
+    }
+}
+
+#[non_exhaustive]
+#[derive(Clone, Debug, Default)]
+/// A builder used to more easily construct the [Bucket] struct.
+pub struct BucketBuilder {
```
codyoss: These builders can never fail to build. Instead of builders should we just have setters on the original structs? Would save a lot of generated code. Thoughts?

storage/src/lib.rs
```diff
@@ -0,0 +1,4984 @@
...
+    /// Pays buckets.
+    pub fn user_project(mut self, value: impl Into<String>) -> Self {
+        self.url_params
+            .insert("user_project".into(), vec![value.into()]);
+        self
+    }
+
+    pub async fn upload(
+        mut self,
+        media: impl Into<BytesReader>,
```
SurferJeffAtGoogle: `BytesReader` will limit what programmers can pass in here.  They can pass in a `Bytes`, `Response`, or `File`.  If they have some other bytes source that isn't one of these three then they have to read all the bytes and pass it in as Bytes, which requires reading the entire stream into memory.  The size of the content to upload may exceed the size of RAM.

Instead, could this function take a `Box<tokio::io::AsyncRead>` parameter?  Then, no matter what their byte source, they can implement `AsyncRead` for it if they haven't already.

I think that would also obviate some of the code in `Bytes.rs`.  `tokio::fs::File` already implements `AsyncRead`.  We could provide helper functions to convert `reqwest::Response` and `Bytes` to `AsyncRead`s.

storage/src/lib.rs
```diff
@@ -0,0 +1,4984 @@
...
+        let form = reqwest::multipart::Form::new()
+            .part(
+                "body",
+                reqwest::multipart::Part::bytes(body)
+                    .mime_str("application/json")
+                    .map_err(Error::wrap)?,
+            )
+            .part(
+                "media",
+                reqwest::multipart::Part::bytes(media.into().read_all().await?.as_ref().to_owned())
```
SurferJeffAtGoogle: I'm surprised to see `.read_all()` here.  What if the caller wants to upload a 300GB file?

SurferJeffAtGoogle (changes requested):
I spent about an hour studying the APIs provided by `reqwest`, `tokio`, and `Bytes`.  It's very tricky to fit them all together.

storage/src/lib.rs
codyoss: This is the exact reason we eventually need to support chucked uploads. Then smaller buffers can be uploaded over several requests. The way this method is structured we could bolt on an optional call param called "chunk_size". If set it would do a buffered chuncked upload instead of the single shot upload as shown here. I still need to do more investigation into this space as well.

storage/src/lib.rs
codyoss: Let me stew on this some more over the weekend. This is for sure the hardest part of the API to get right. I agree that it feel like we could do something here to be even more permissive. I think a convince type is still nice for the common cases regardless though. Implementing AsyncRead on something is not super intuitive, to me at least, so having some helpers seems like a nice to have. 

If this was just an `Box<AsyncRead>` is makes it quite hard to use for types like `Vec<u8>`, `String`, or a `std::fs::File` which may be common for input read from another source.

codyoss: Maybe this type could just have another internal varient and impl a From<Box< AsyncRead >>?

storage/src/lib.rs
SurferJeffAtGoogle: 👍

storage/src/lib.rs
SurferJeffAtGoogle: > Maybe this type could just have another internal varient and impl a From<Box< AsyncRead >>?

I think we'll end up with a lot less code if accept a `Box<AsyncRead>` instead of adding another internal variant.

SurferJeffAtGoogle: For `Vec<u8>`, `String`, and the like, we could provide user a wrapper
```
let readCursor = AsyncCursorReader::from(std::io::Cursor);
```

For `std::fs::File`, I'm not sure.  It would be doing synchronous reads from an asynchronous function.  Not sure how/if we should support that.

codyoss: I did some playing around this weekend and I think an <'a>AsyncRead + Unpin works well here. This seems to be the common construct for tokio::io::copy and the likes. As long as we have some good examples I think that works. I will update code. I still question on the download side though. I will start a new thread for that.

storage/src/lib.rs
```diff
@@ -0,0 +1,5006 @@
...
+            .map_err(Error::wrap)?;
+        if !res.status().is_success() {
+            let error: ApiErrorReply = res.json().await.map_err(Error::wrap)?;
+            return Err(Error::wrap(error.into_inner()));
+        }
+        let res: model::Object = res.json().await.map_err(Error::wrap)?;
+        Ok(res)
+    }
+
+    pub async fn download(self) -> Result<Vec<u8>> {
```
codyoss: @SurferJeffAtGoogle We need a better return type here, and I don't know if AsyncRead is good enough. I feel like we should return a concrete type folks can work with easier. Maybe this is where something like the BytesReader type I made make sense? Thoughts?

codyoss: Maybe a [BufReader](https://docs.rs/tokio/latest/tokio/io/struct.BufReader.html)?

storage/src/model.rs
SurferJeffAtGoogle: Sounds good but not urgent either way.  Maybe this is a question the community can help answer later.

storage/src/lib.rs
SurferJeffAtGoogle: Sounds good.

storage/src/lib.rs
codyoss: What would you suggest for a return value here?

SurferJeffAtGoogle: Return a `Box<AsyncRead>`.

storage/src/bytes.rs
```diff
@@ -0,0 +1,326 @@
...
+    inner: Inner,
+}
+
+impl BytesReader {
```
SurferJeffAtGoogle, line 32: Is `BytesReader` now obsolete?

storage/src/lib.rs
```diff
@@ -0,0 +1,5012 @@
...
+
+    pub async fn upload<'a, R>(mut self, reader: &'a mut R) -> Result<model::Object>
+    where
+        R: tokio::io::AsyncRead + Unpin,
+    {
+        let client = self.client.inner;
+        let tok = client.cred.access_token().await.map_err(Error::wrap)?;
+        let body = serde_json::to_vec(&self.request).map_err(Error::wrap)?;
+        let mut buf: Vec<u8> = vec![];
+        tokio::io::copy(reader, &mut buf)
```
SurferJeffAtGoogle: It looks like the the entire payload is loaded into memory.  Is my understanding correct?

storage/src/lib.rs
codyoss: I am a little concerned about the usability of that type I guess. Unless the user also pulls in [AsyncReadExt](https://docs.rs/tokio/1.15.0/tokio/io/trait.AsyncReadExt.html) that type is quite hard to use imo. I will move to this for now but I think this needs more disscussion.

storage/src/bytes.rs
codyoss: Sure, I can remove. I was keeping it around incase we wanted to have a helper type for downloads.

codyoss: Actually I think we still need this type, but maybe not publicly exposed? The response itself is not an AsyncRead, so if that is what we want we need to have a type impl that. Also we will need to store intermediate buffers it we do something like this... The API that is exposed to use will return Bytes and we would only be able to fill the users buffer up to the capacity I think. This is kinda tricky, might be a good pair activity if you have the time.

codyoss: Looking at the Java client it appears they have two type of download methods that take arguments instead of returning type: https://github.com/googleapis/java-storage/blob/1cfdf751f0b8e68a2f7f49b81d730b6fc11535d6/google-cloud-storage/src/main/java/com/google/cloud/storage/Blob.java#L226-L262.

Will investigate a few others as well.

codyoss: In Go: The apiary just returns a raw HTTP response, but the veneer returns a stream
C++: returns a stream
C#: accepts stream
Java: accepts a stream or path
Node: accepts input
PHP: accepts input
Python: accepts input
Ruby: accepts input

Found by prior knowledge and quickly looking at Cloud sample: https://cloud.google.com/storage/docs/downloading-objects#storage-download-object-csharp

storage/src/lib.rs
SurferJeffAtGoogle: > I am a little concerned about the usability of that type I guess. Unless the user also pulls in [AsyncReadExt](https://docs.rs/tokio/1.15.0/tokio/io/trait.AsyncReadExt.html) that type is quite hard to use imo. I will move to this for now but I think this needs more disscussion.

I was 100% assuming that the caller would import `AsyncReadExt`.  I couldn't imagine using `AsyncRead` directly.

storage/src/lib.rs
SurferJeffAtGoogle: For this uploading and downloading, I'm surprised that reqwest doesn't accept the standard `AsyncRead` and `AsyncWrite` types.

I don't think we can solve this problem really well without feedback from the larger community, so I propose a **new strategy**:  Expose all the reqwest types in our APIs.  It's the easiest thing to do, no?  Let our early adopters tell us if it works for them or not.

storage/src/lib.rs
codyoss: I think I found a better API for this, and that is accepting an AsyncWrite instead. I will make this change and update the examples so you can LMK what you think

storage/src/lib.rs
codyoss: I think what we now have for downloads and uploads should be good to get some feedback. I would prefer not to expose types from other libraries, especially if they are not 1.0 yet. It would be unfortunate if another crates type broke  our compatibility.

SurferJeffAtGoogle: Ok.  I'll accept this code if you open a bug saying that upload() must be fixed to not load the entire payload into memory.

storage/src/lib.rs
SurferJeffAtGoogle: Ok.  I'll accept this code if you open a bug saying that download() must not store the entire payload in memory.

codyoss: Ack, opened #7

storage/src/lib.rs
codyoss: Ack, opened #6

storage/src/model.rs
codyoss: KK, will leave for now.

PR #8, opened 2022-02-02
chore: use rustls for reqwest

lucacasonato:
This commit changes the auth library to use the rustls feature of
reqwest instead of native-tls. This is done because rustls is already
included as a dependency of google-cloud-auth anyway. There is no point
in including multiple TLS libraries.

By extension this removes the dependency on OpenSSL.

This reduces the Cargo.lock file form 1217 lines to 1070 lines :-)

### DIFF
--- a/auth/Cargo.toml
+++ b/auth/Cargo.toml
@@ -10,15 +10,15 @@ categories = ["authentication", "web-programming"]
 edition = "2021"
 
 [dependencies]
+async-trait = "0.1"
+backoff = { version="0.4", features = ["tokio"] }
 base64 = "0.13"
 chrono = { version = "0.4", features = ["serde"] }
-reqwest = {version = "0.11", features = ["json"] }
+http = "0.2"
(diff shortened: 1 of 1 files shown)


auth/Cargo.toml
```diff
@@ -10,15 +10,15 @@ categories = ["authentication", "web-programming"]
 edition = "2021"
 
 [dependencies]
+async-trait = "0.1"
+backoff = { version="0.4", features = ["tokio"] }
 base64 = "0.13"
 chrono = { version = "0.4", features = ["serde"] }
-reqwest = {version = "0.11", features = ["json"] }
+http = "0.2"
+reqwest = { version = "0.11", default-features = false, features = ["json", "rustls-tls"] }
```
lucacasonato, line 18: This line is the only actual change. Other lines are just from alpha sorting the dependency list.

coryan:
Thanks for the PR and sorry for the lack of attention.  We are restarting development on this project and closing all stale PRs.

PR #17, opened 2024-10-29
ci: initial builds to support golang development

coryan:
Create the initial CI builds to support golang development.
Add a small function for the client generator to bootstrap these
builds.


.github/workflows/gclientgen.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - name: Setup Go ${{ matrix.go-version }}
+      uses: actions/setup-go@v5
+      with:
+        go-version: ${{ matrix.go-version }}
+        cache-dependency-path: |
+          go.work.sum
+          gclientgen/go.sum
+    - name: Display Go version
+      run: go version
+    - run: go -C gclientgen get ./...
```
codyoss, line 23: Could you use `go mod download` instead here

.github/workflows/gclientgen.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - run: go -C gclientgen test ./...
```
codyoss, line 24: Since we are in a Go workspace we can remove -C flag as the Go command is aware of all declared modules in the Workspace and will run tests when using the `./...` pattern.

gclientgen/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/googleapis/gclientgen
```
codyoss, line 1:
```suggestion
module github.com/googleapis/google-cloud-rust/gclientgen
```

gclientgen/language/internal/symbol_names_test.go
```diff
@@ -0,0 +1,14 @@
+package symbol_names
```
codyoss, line 1:
```suggestion
package internal
```

codyoss (changes requested):
Just a couple small things.

gclientgen/go.mod
coryan: Done.

.github/workflows/gclientgen.yaml
coryan: That does not work for me, maybe because I did something wrong elsewhere?

```
go test ./...
pattern ./...: directory prefix . does not contain modules listed in go.work or their selected dependencies
```

gclientgen/language/internal/symbol_names_test.go
coryan: Done.

.github/workflows/gclientgen.yaml
julieqiu: I don't think there is a need for a `go.work` file for this module

.github/workflows/gclientgen.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - name: Run `staticcheck`
+      uses: dominikh/staticcheck-action@v1.3.0
+      with:
+          install-go: false
+          cache-key: ${{ matrix.go-version }}
+          working-directory: gclientgen
+    - name: Detect Changes
+      run: git diff --exit-code
+
+  vuln:
```
julieqiu, line 56: I would also add a step to run https://pkg.go.dev/mvdan.cc/unparam

.github/workflows/gclientgen.yaml
```diff
@@ -0,0 +1,75 @@
...
+      with:
+        go-version: ${{ matrix.go-version }}
+        cache-dependency-path: |
+          go.work.sum
+          gclientgen/go.sum
+    - name: Display Go version
+      run: go version
+    - run: |
+        go install golang.org/x/vuln/cmd/govulncheck@latest
+        govulncheck -C gclientgen ./...
```
julieqiu, line 75: you don't need `-C` here

.github/workflows/gclientgen.yaml
```diff
@@ -0,0 +1,75 @@
...
+      uses: actions/setup-go@v5
+      with:
+        go-version: ${{ matrix.go-version }}
+        cache-dependency-path: |
+          go.work.sum
+          gclientgen/go.sum
+    - name: Display Go version
+      run: go version
+    - run: go -C gclientgen mod tidy
+    - run: go mod download
```
julieqiu, line 44: just `go mod tidy` should be sufficient (you don't need `go mod download`)

.github/workflows/gclientgen.yaml
```diff
@@ -0,0 +1,75 @@
...
+    - run: go -C gclientgen fmt ./...
```
julieqiu, line 45: I don't think you need the `-C gclientgen` here

gclientgen/go.mod
```diff
@@ -0,0 +1,5 @@
+module github.com/googleapis/google-cloud-rust/gclientgen
```
julieqiu, line 1: What about just `github.com/googleapis/google-cloud-rust/generator` for the name?

.github/workflows/gclientgen.yaml
codyoss: That is fine too, in that case I think you could add a default working directory for the job. Something like

```
defaults:
      run:
        working-directory: ./generator
```

.github/workflows/gclientgen.yaml
coryan: Done.

.github/workflows/gclientgen.yaml
coryan: Done.

.github/workflows/gclientgen.yaml
coryan: Done.

.github/workflows/gclientgen.yaml
coryan: Done.

gclientgen/go.mod
coryan: Done.

.github/workflows/generator.yaml
```diff
@@ -0,0 +1,81 @@
...
+    - uses: actions/checkout@v4
+    - name: Setup Go ${{ matrix.go-version }}
+      uses: actions/setup-go@v5
+      with:
+        go-version: ${{ matrix.go-version }}
+        cache-dependency-path: |
+          generator/go.sum
+    - name: Display Go version
+      run: go version
+    - run: go mod download
```
julieqiu, line 24: use `go mod tidy` instead (in general I almost never run` go mod download`)

julieqiu (approved):

.github/workflows/generator.yaml
coryan: Done.

codyoss (approved):

PR #20, opened 2024-10-29
feat: add a prototype  generator

codyoss:
This prototype is meant to be run as a protoc plugin and yield a HTTP/JSON client library. It is missing many features and has many todos. Currently this is only tested by one "golden" style test that validates generated output matches cached testdata.

PR #24, opened 2024-10-29
chore: move package to internal/language

julieqiu:
Go packages typically live in the `internal/` directory, unless the code needs to be exported (as opposed to the other way around).


codyoss:
I removed this code in #20 as it was mostly a placeholder to check CI. I will close this PR

PR #29, opened 2024-10-30
chore: move src/ -> internal/

julieqiu:
Move code from `src/` to `internal/`, since the code is only meant for use within this module, per https://go.dev/doc/go1.4#internalpackages.


codyoss (approved):
Thanks!

Issue #32, opened 2024-10-30
Construct protoc expected output


codyoss:
Right now the generator does output code, but it does not give protoc the information back on this generated code.

https://github.com/googleapis/google-cloud-rust/blob/77720dee3887853cc65bf31c90ce887b6e629976/generator/src/genclient/genclient.go#L86-L89

coryan:
Fixed in 192

PR #41, opened 2024-10-30
chore: convert all todos to real bugs

codyoss:
I did leave some breadcrumbs in the code still so we know where some of these changes should be made, but every todo now has a bug.

### DIFF
--- a/generator/src/genclient/genclient.go
+++ b/generator/src/genclient/genclient.go
@@ -84,8 +84,7 @@ func (r *GenerateRequest) outDir() string {
 
 // Output of generation.
 type Output struct {
-	// TODO(codyoss): needs enough info to pass information back to protoc
-	// on what was generated.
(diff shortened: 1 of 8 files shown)


generator/src/genclient/genclient.go
```diff
@@ -84,8 +84,7 @@ func (r *GenerateRequest) outDir() string {
 
 // Output of generation.
 type Output struct {
-	// TODO(codyoss): needs enough info to pass information back to protoc
-	// on what was generated.
+	// TODO(codyoss): https://github.com/googleapis/google-cloud-rust/issues/32
```
coryan, line 87: I usually write TODO(#32) and maybe a short summary. Your call.

coryan (approved):

generator/src/genclient/genclient.go
codyoss: oops, had auto merge on this. Ack for next time

PR #46, opened 2024-10-30
impl: bootstrap Timestamp type

coryan:


types/src/timestamp.rs
```diff
@@ -0,0 +1,204 @@
...
+    fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        let odt = time::OffsetDateTime::parse(&value, &Rfc3339).map_err(E::custom)?;
+        let nanos_since_epoch = odt.unix_timestamp_nanos();
+        let seconds = (nanos_since_epoch / NS) as i64;
+        let nanos = (nanos_since_epoch % NS) as i32;
+        Ok(Self::Value {
+            seconds: seconds,
```
codyoss, line 99: nit: clippy warning


```suggestion
            seconds,
```

types/src/timestamp.rs
```diff
@@ -0,0 +1,204 @@
...
+            nanos: nanos,
```
codyoss, line 100: nit: clippy warning


```suggestion
            nanos,
```

types/src/timestamp.rs
```diff
@@ -0,0 +1,204 @@
...
+
+    fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
+        formatter.write_str("a string with a timestamp in RFC 3339 format")
+    }
+
+    fn visit_str<E>(self, value: &str) -> Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        let odt = time::OffsetDateTime::parse(&value, &Rfc3339).map_err(E::custom)?;
```
codyoss, line 94: nit: clippy warning


```suggestion
        let odt = time::OffsetDateTime::parse(value, &Rfc3339).map_err(E::custom)?;
```

codyoss (approved):
LGTM, maybe we should run clippy in CI? Not sure other code is compliant either though.

coryan (commented):
Enabling clippy sounds like a good idea.  We can first scope it to just `types` and add `google-cloud-auth` later.

PR #49, opened 2024-10-30
impl: bootstrap well-known `Duration` type

coryan:


codyoss (approved):

types/src/duration.rs
```diff
@@ -0,0 +1,317 @@
...
+            seconds: want_seconds,
+            nanos: want_nanos,
+        };
+        assert_eq!(want, got);
+    }
+
+    #[test_case(0, 0, "0s" ; "zero")]
+    #[test_case(12, 0, "12s"; "round positive seconds")]
+    #[test_case(12, 123, "12.000000123s"; "positive seconds and nanos")]
+    #[test_case(12, 123_000, "12.000123000s"; "positive seconds and micros")]
```
devoncarew, line 232: Here, and a few cases below, should we be instead expecting no trailing zeros? `12.000123000s` => `12.000123s`. I don't think it's incorrect to have trailing `0`s but suspect you'd want to encode as `12.1s` instead of `12.100000000s`.

coryan: Maybe, I think both are legal encodings, though the trailing zeroes are ugly. Let me open a bug.

PR #61, opened 2024-11-01
impl(generator): support more field types

coryan:
Support more field types, including several string formats.

Remove code duplication between array and scalar field handling.

Make it easier to write smaller unit tests.

### DIFF
--- a/generator/internal/genclient/translator/openapi/openapi.go
+++ b/generator/internal/genclient/translator/openapi/openapi.go
@@ -18,6 +18,7 @@ package openapi
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/googleapis/google-cloud-rust/generator/internal/genclient"
 	"github.com/googleapis/google-cloud-rust/generator/internal/genclient/language"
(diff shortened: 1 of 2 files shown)


generator/internal/genclient/translator/openapi/openapi.go
```diff
@@ -132,66 +140,145 @@ func makeMessageFields(messageName string, message *base.Schema) ([]*genclient.F
...
+			TypezID:       typezID,
+			Optional:      true,
+		}, nil
+	}
+	return nil, fmt.Errorf("cannot build any AllOf schema for field %s.%s", messageName, name)
+}
+
+func makeStringType(messageName, name, format string, optional bool, field *base.Schema) (*genclient.Field, error) {
+	switch format {
+	case "":
```
julieqiu, line 193: should this be case "string"?

coryan: No.  A normal string has no `"format"` field. Integers as strings have  `{"type": "string", "format": "int64"}`.

I sure don't like how OpenAPI represents primitive types 🤷

PR #69, opened 2024-11-01
impl: better name conversions

### LINKED ISSUE #35: bug in case conversion lib
The current case converter is not working as intended for some fields, for example: data_crc32c. Also most camel things in this file should be pascal case. Either need to find a new lib or just write these functions ourselves.

https://github.com/googleapis/google-cloud-rust/blob/77720dee3887853cc65bf31c90ce887b6e629976/generator/cmd/protoc-gen-gclient/testdata/rust/golden/model.rs#L460

### LINKED ISSUE #36: language codecs should register reserved words
Each language has a bunch of case-sensitive "words" that are keywords/reserved in the language spec. We need to register these so we can do name sanitation and possibly identifier mangling do we don't generate invalid code.

coryan:
This PR always uses the per-language Codec to perform any "case conversion". We
need to involve the Codec because each language has a different number of
reserved words that must be escaped, and the escaping is also language
specific.

I also fixed a number of places where we used "camel case" when we meant
`PascalCase` (note the starting upper case, as opposed to `camelCase`).

Almost incidentally, I fixed the conversion for `data_crc32c` (and other names
already in `snake_case` form). Previously that got converted to
`data_crc_32_c`.

Fixes #35 and fixes #36.

### DIFF
--- a/generator/internal/genclient/genclient.go
+++ b/generator/internal/genclient/genclient.go
@@ -60,7 +60,7 @@ type LanguageCodec interface {
 	// An example return value might be
 	// `&Pair{Key: "secretId", Value: "req.SecretId()"}`
 	QueryParams(m *Method, state *APIState) []*Pair
-	// ToSnake converts a symbol name to `snake_case``, applying any mangling
+	// ToSnake converts a symbol name to `snake_case`, applying any mangling
 	// required by the language, e.g., to avoid clashes with reserved words.
 	ToSnake(string) string
 	// ToPascal converts a symbol name to `PascalCase`, applying any mangling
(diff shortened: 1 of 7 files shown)


generator/internal/genclient/genclient.go
```diff
@@ -60,6 +60,16 @@ type LanguageCodec interface {
 	// An example return value might be
 	// `&Pair{Key: "secretId", Value: "req.SecretId()"}`
 	QueryParams(m *Method, state *APIState) []*Pair
+	// ToSnake converts a symbol name to `snake_case``, applying any mangling
```
julieqiu, line 63: extra "`"

julieqiu (approved):

generator/internal/genclient/genclient.go
coryan: Fixed, thanks.

PR #70, opened 2024-11-01
cmd/protoc-gen-gclient: factor out captureInput

julieqiu:
Move logic for captureInput to a separate function and document it.

### DIFF
--- a/generator/cmd/protoc-gen-gclient/main_test.go
+++ b/generator/cmd/protoc-gen-gclient/main_test.go
@@ -31,7 +31,7 @@ func TestRun_Rust(t *testing.T) {
 		outDir      = "../../testdata/rust/gclient/golden"
 		templateDir = "../../templates"
 	)
-	if err := run(inputPath, outDir, templateDir); err != nil {
+	if err := run(os.Stdin, os.Stdout, inputPath, outDir, templateDir); err != nil {
 		t.Fatal(err)
 	}
 }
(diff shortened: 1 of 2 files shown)


generator/cmd/protoc-gen-gclient/main_test.go
```diff
@@ -26,7 +26,7 @@ func TestMain(m *testing.M) {
 }
 
 func TestRun_Rust(t *testing.T) {
-	if err := run("testdata/rust/rust.bin", "testdata/rust/golden", "../../templates"); err != nil {
+	if err := run(os.Stdin, os.Stdout, "testdata/rust/rust.bin", "testdata/rust/golden", "../../templates"); err != nil {
```
codyoss, line 29: nit: do we suspect to every use values different from stdin/stdout? I am curious why we should make them configurable?

codyoss (commented):
Nice, thank you!

generator/cmd/protoc-gen-gclient/main_test.go
julieqiu: I found it surprising that `run` was reading/writing to a location that I didn't tell it to, so the reason was to make it explicit. 

I also imagine that at some point we will want to write [`outBytes`](https://github.com/googleapis/google-cloud-rust/pull/70/files#diff-847456702a1cf33aeec38dbdda1288d610dc4e8f6c926eb67dcc6a74124d22caR97) to somewhere other than stdout?

codyoss (approved):

generator/cmd/protoc-gen-gclient/main_test.go
codyoss: I think that is just the contract for protoc. But if you think it adds value I am fine with it. If we ever add a test to validate output you are right it would be useful to capture.

PR #71, opened 2024-11-01
impl(generator): handle OpenAPI map types

### LINKED ISSUE #62: Support map fields in OpenAPI prototype
Map types and `.google.protobuf.Any` are represented as anonymous messages in Google's OpenAPI discovery docs.

coryan:
OpenAPI does not really have `map<..>` types like Protobuf, but it has `object`
fields sometimes with restrictions on the value type. If one squints, `object`
is equivalent to `google.protobuf.Any`. And an JSON object where the values can
only be `int32` is the same as `map<string, int32>`.

Fixes #62

### DIFF
--- a/generator/internal/genclient/translator/openapi/openapi.go
+++ b/generator/internal/genclient/translator/openapi/openapi.go
@@ -184,19 +184,55 @@ func (t *Translator) makeObjectField(messageName, name string, field *base.Schem
 	if len(field.AllOf) != 0 {
 		return t.makeObjectFieldAllOf(messageName, name, field)
 	}
-	// TODO(#62) - this is an Any or a map<string, T>, needs a TypezID
-	return &genclient.Field{
-		Name:          name,
-		Documentation: field.Description,
-		Typez:         genclient.MESSAGE_TYPE,
-		Optional:      true,
-	}, nil
+	if field.AdditionalProperties != nil && field.AdditionalProperties.IsA() {
+		// This indicates we have a map<K, T> field. In OpenAPI, these are
+		// simply JSON objects, maybe with a restrictive value type.
+		schema, err := field.AdditionalProperties.A.BuildSchema()
+		if err != nil {
+			return nil, fmt.Errorf("cannot build schema for field %s.%s, error=%q", messageName, name, err)
+		}
+
+		if len(schema.Type) == 0 {
+			// Untyped message fields are .google.protobuf.Any
+			return &genclient.Field{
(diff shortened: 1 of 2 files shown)


generator/internal/genclient/translator/openapi/openapi.go
```diff
@@ -241,6 +288,44 @@ func (t *Translator) makeObjectFieldAllOf(messageName, name string, field *base.
...
+	message := t.state.MessageByID[id]
+	if message == nil {
+		// The map was not found, insert the type.
+		key := &genclient.Field{
+			Name:    "$key",
+			ID:      id + "$key",
+			Typez:   genclient.STRING_TYPE,
+			TypezID: "string",
+		}
+		new := &genclient.Message{
```
julieqiu, line 313: nit: `new` is a special keyword in go, so I tend to avoid it as a variable name

generator/internal/genclient/translator/openapi/openapi_test.go
```diff
@@ -204,6 +204,176 @@ func TestArrayTypes(t *testing.T) {
 	})
 }
 
+func TestSimpleObject(t *testing.T) {
+	const messageWithBasicTypes = `
```
julieqiu, line 208: We might want to move all of this and similar testdata to files at some point

julieqiu (approved):

generator/internal/genclient/translator/openapi/openapi.go
```diff
@@ -184,19 +184,55 @@ func (t *Translator) makeObjectField(messageName, name string, field *base.Schem
...
-		Documentation: field.Description,
-		Typez:         genclient.MESSAGE_TYPE,
-		Optional:      true,
-	}, nil
+	if field.AdditionalProperties != nil && field.AdditionalProperties.IsA() {
+		// This indicates we have a map<K, T> field. In OpenAPI, these are
+		// simply JSON objects, maybe with a restrictive value type.
+		schema, err := field.AdditionalProperties.A.BuildSchema()
+		if err != nil {
+			return nil, fmt.Errorf("cannot build schema for field %s.%s, error=%q", messageName, name, err)
```
codyoss, line 192: nit: if you want to return the underlying error we usually wrap those in Go. This can be achieved with `%w`

```suggestion
			return nil, fmt.Errorf("cannot build schema for field %s.%s: %w", messageName, name, err)
```

codyoss (approved):

PR #78, opened 2024-11-03
impl(generator): handle nested messages and enums in Rust

### LINKED ISSUE #40: better handling of nested messages
Today we generate nested messages flatly and namespace types by prepending onto the identifier. We could do better with mods.

coryan:
The mapping nested messages and enums uses a public module within the crate.
That requires a new "snake case" accessor for the message name, as Rust uses
snake case for modules.

We also need a boolean to find out if the module is needed at all.

And we need to get a different attribute for the fully qualified name of each
message and enum, as we must refer to the qualified name of the generated
struct when using the type.

Fixes #40. This PR is more bulky than I like, it snowballed on me. Let me know if you want me to break it into smaller PRs.

### DIFF
--- a/generator/internal/genclient/language/internal/rust/rust_test.go
+++ b/generator/internal/genclient/language/internal/rust/rust_test.go
@@ -86,3 +86,70 @@ func TestToPascal(t *testing.T) {
 		}
 	}
 }
+
+func TestMessageNames(t *testing.T) {
+	message := &genclient.Message{
+		Name: "Replication",
+		ID:   "..Replication",
+		Fields: []*genclient.Field{
+			{
+				Name:     "automatic",
+				Typez:    genclient.MESSAGE_TYPE,
+				TypezID:  "..Automatic",
+				Optional: true,
+				Repeated: false,
+			},
+		},
+	}
+	nested := &genclient.Message{
+		Name: "Automatic",
+		ID:   "..Replication.Automatic",
+	}
+
+	api := genclient.NewTestAPI([]*genclient.Message{message, nested}, []*genclient.Enum{}, []*genclient.Service{})
+
+	c := &Codec{}
+	if got := c.MessageName(message, api.State); got != "Replication" {
+		t.Errorf("mismatched message name, want=Replication, got=%s", got)
+	}
+	if got := c.FQMessageName(message, api.State); got != "crate::Replication" {
+		t.Errorf("mismatched message name, want=crate::Replication, got=%s", got)
+	}
+
+	if got := c.MessageName(nested, api.State); got != "Automatic" {
+		t.Errorf("mismatched message name, want=Automatic, got=%s", got)
+	}
+	if got := c.FQMessageName(nested, api.State); got != "crate::replication::Automatic" {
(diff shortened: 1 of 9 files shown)


generator/internal/genclient/templatedata.go
```diff
@@ -202,11 +202,30 @@ func (m *message) Enums() []*enum {
...
-// NameToCamel converts a Name to CamelCase.
 func (m *message) Name() string {
 	return m.c.MessageName(m.s, m.state)
 }
 
+func (m *message) QualifiedName() string {
+	return m.c.FQMessageName(m.s, m.state)
+}
+
+func (m *message) NameSnakeCase() string {
```
codyoss, line 213: comment: all of these different case top level case conversion functions make me wonder if there is a better way to handle this and/or there is not enough flexibility at the template layer. I originally chose mustache because it is simple and does not allow you to code in the template itself. As time goes on we should monitor to make sure this is still the right choice. We could switch over to Go's template engine in the future to provide much more extensibility in the templates.

Nothing to change for now, but something we should think more about.

generator/internal/genclient/language/internal/rust/rust_test.go
```diff
@@ -86,3 +86,70 @@ func TestToPascal(t *testing.T) {
...
+	nested := &genclient.Message{
+		Name: "Automatic",
+		ID:   "..Replication.Automatic",
+	}
+
+	api := genclient.NewTestAPI([]*genclient.Message{message, nested}, []*genclient.Enum{}, []*genclient.Service{})
+
+	c := &Codec{}
+	if got := c.MessageName(message, api.State); got != "Replication" {
+		t.Errorf("mismatched message name, want=Replication, got=%s", got)
```
codyoss, line 113: nit: goes convention for testing is backwards from most languages. I would usually expect these strings to be written as "got Replication, want %s"

generator/testdata/rust/gclient/golden/model.rs
```diff
@@ -163,39 +163,43 @@ pub struct SecretVersion {
...
+    pub struct State(i32);
 
-    // Not specified. This value is unused and invalid.
-    pub const SecretVersion_StateUnspecified: SecretVersion_State = SecretVersion_State(0);
+    impl State {
 
-    // The [SecretVersion][google.cloud.secretmanager.v1.SecretVersion] may be
-    // accessed.
-    pub const SecretVersion_Enabled: SecretVersion_State = SecretVersion_State(1);
+        // Not specified. This value is unused and invalid.
```
codyoss, line 183: Looks like we are missing an extra `/` in the template.

codyoss (approved):
Nice, just a couple small things.

generator/internal/genclient/language/internal/rust/rust_test.go
coryan: Fixed.

generator/internal/genclient/templatedata.go
coryan: Ack and agree.

generator/testdata/rust/gclient/golden/model.rs
coryan: Agree, created #84

PR #79, opened 2024-11-03
impl: move testdata

julieqiu:
Move all testdata to one location.


generator/cmd/protoc-gen-gclient/testdata/rust/golden/model.rs
coryan, line 1: I would have expected this file to move to a new location, but it is just marked as deleted?

generator/cmd/protoc-gen-gclient/testdata/rust/golden/lib.rs
coryan, line 1: I would have expected this file to move to a new location, but it is just deleted?

julieqiu: Good catch, thanks

coryan (approved):

PR #81, opened 2024-11-04
impl(internal/genclient/translator/protobuf): simplify protobuf translator

julieqiu:
The Translator struct isn't necessary, so it is removed to simplify the code in generator/internal/genclient/translator/protobuf

### DIFF
--- a/generator/internal/genclient/translator/protobuf/protobuf.go
+++ b/generator/internal/genclient/translator/protobuf/protobuf.go
@@ -15,6 +15,7 @@
 package protobuf
 
 import (
+	"fmt"
 	"log/slog"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


generator/internal/genclient/translator/protobuf/protobuf.go
```diff
@@ -190,24 +171,25 @@ func normalizeTypes(in *descriptorpb.FieldDescriptorProto, field *genclient.Fiel
 	}
 }
 
-func (t *Translator) processMessage(m *descriptorpb.DescriptorProto, mFQN string, parent *genclient.Message) *genclient.Message {
-	message := &genclient.Message{}
-	t.state.MessageByID[mFQN] = message
-	message.Name = m.GetName()
-	message.ID = mFQN
-	message.Parent = parent
+func processMessage(state *genclient.APIState, m *descriptorpb.DescriptorProto, mFQN string, parent *genclient.Message) *genclient.Message {
```
coryan, line 174: I don't claim to know what is idiomatic in Go, but this is basically making the functions methods on the `state` parameter...

coryan (approved):
Almost all the cleanups look great to me.

The removal of the Translator object is maybe six of one half a doze of the other... we can always change it back.

generator/internal/genclient/translator/protobuf/protobuf.go
julieqiu: I wanted to do away with the `Translator` struct, since it didn't seem to be adding anything and removing made things easier to reason about

PR #82, opened 2024-11-04
impl(internal/gclient/translator/protobuf): use a map in normalizeTypes

julieqiu:
Use a map instead of a switch statement for readability.


coryan (approved):
FYI, in GitHub land it is more common / idiomatic to work out of a fork vs. working out of branches in the main repository.

PR #96, opened 2024-11-05
impl(generator/protobuf): message fields are always optional

coryan:
Normalize message fields to always be optional. This will make other parts of
the code easier to implement: we always wrap optional fields with `Option<T>`.

Motivated by #95


codyoss (approved):
LGTM. Should we add optional wrappers just in the template directly you think instead of resolving that in the codec layer?

coryan:
> LGTM. Should we add optional wrappers just in the template directly you think instead of resolving that in the codec layer?

I think this may answer your question. In a couple of PRs the codec will do this:

https://github.com/coryan/google-cloud-rust/blob/1d7edd54d11207ae757122fea4be1f8deb3eedd7/generator/internal/genclient/language/internal/rust/rust.go#L99-L106

PR #99, opened 2024-11-05
feat: add Rust oneof support

codyoss:
- Added support for one-ofs as mods related to a message
- refactored tests to source files from testdata
- removed some generated TODOs to make code easier to generate

### DIFF
--- a/generator/internal/genclient/language/internal/rust/rust.go
+++ b/generator/internal/genclient/language/internal/rust/rust.go
@@ -90,7 +90,7 @@ func ScalarFieldType(f *genclient.Field) string {
 
 func (c *Codec) FieldType(f *genclient.Field, state *genclient.APIState) string {
 	value := c.fieldType(f, state)
-	if f.IsExplicitOneOf {
+	if f.IsOneOf {
 		return c.wrapOneOfField(f, value)
 	}
 	return value
--- a/generator/internal/genclient/language/internal/rust/rust.go
+++ b/generator/internal/genclient/language/internal/rust/rust.go
@@ -89,6 +89,14 @@ func ScalarFieldType(f *genclient.Field) string {
 }
 
 func (c *Codec) FieldType(f *genclient.Field, state *genclient.APIState) string {
+	value := c.fieldType(f, state)
+	if f.IsExplicitOneOf {
+		return c.wrapOneOfField(f, value)
+	}
+	return value
+}
+
+func (c *Codec) fieldType(f *genclient.Field, state *genclient.APIState) string {
 	if f.Typez == genclient.MESSAGE_TYPE {
 		m, ok := state.MessageByID[f.TypezID]
 		if !ok {
@@ -101,7 +109,8 @@ func (c *Codec) FieldType(f *genclient.Field, state *genclient.APIState) string 
 			return "Option<std::collections::HashMap<" + key + "," + val + ">>"
 		}
 		if strings.HasPrefix(m.ID, ".google.protobuf.") {
-			return "Option<String> /* TODO(#77) - handle " + f.TypezID + " */"
+			// TODO(#77): Better handling of well-known types
+			return "Option<String>"
 		}
 		return "Option<" + c.FQMessageName(m, state) + ">"
 	} else if f.Typez == genclient.ENUM_TYPE {
@@ -116,6 +125,17 @@ func (c *Codec) FieldType(f *genclient.Field, state *genclient.APIState) string 
 		return ""
 	}
 	return ScalarFieldType(f)
+
+}
+
+func (c *Codec) wrapOneOfField(f *genclient.Field, value string) string {
+	if f.Typez == genclient.MESSAGE_TYPE {
+		// Remove optional wrapper for fields of one-of
+		value = value[7 : len(value)-1]
+		return fmt.Sprintf("(%s)", value)
+	}
+	return fmt.Sprintf("{ %s: %s }", c.ToSnake(f.Name), value)
+
 }
 
 func (c *Codec) TemplateDir() string {
@@ -172,6 +192,10 @@ func (c *Codec) EnumValueName(e *genclient.EnumValue, _ *genclient.APIState) str
 
 func (c *Codec) FQEnumValueName(v *genclient.EnumValue, state *genclient.APIState) string {
 	return fmt.Sprintf("%s::%s::%s", c.enumScopeName(v.Parent), c.ToSnake(v.Parent.Name), c.EnumValueName(v, state))
+}
+
+func (c *Codec) OneOfType(o *genclient.OneOf, _ *genclient.APIState) string {
+	return c.messageScopeName(o.Parent) + "::" + c.ToPascal(o.Name)
 }
 
 func (c *Codec) BodyAccessor(m *genclient.Method, state *genclient.APIState) string {
--- a/generator/internal/genclient/language/internal/rust/rust.go
+++ b/generator/internal/genclient/language/internal/rust/rust.go
@@ -130,7 +130,8 @@ func (c *Codec) fieldType(f *genclient.Field, state *genclient.APIState) string 
 
 func (c *Codec) wrapOneOfField(f *genclient.Field, value string) string {
 	if f.Typez == genclient.MESSAGE_TYPE {
-		// Remove optional wrapper for fields of one-of
+		// Remove `Option<...>` wrapper for fields of one-of
+		//               ^7
 		value = value[7 : len(value)-1]
 		return fmt.Sprintf("(%s)", value)
 	}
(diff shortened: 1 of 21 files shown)


generator/internal/genclient/translator/protobuf/protobuf_test.go
```diff
@@ -28,59 +30,7 @@ import (
 )
 
 func TestScalar(t *testing.T) {
-	contents := `
```
coryan, line 31: Could you do this in a refactoring PR? It is harder to grok new code and cleanups in the same change.

generator/internal/genclient/translator/protobuf/testdata/oneofs.proto
```diff
@@ -0,0 +1,19 @@
...
+        // An int choice
+        int64 field_two = 2;
+    }
+
+    // Optional is oneof in proto
+    optional string field_three = 3;
+
+    // A normal field
+    int32 field_four = 4;
+}
```
coryan, line 19: Missing newline at end of file.

coryan: PS: we need a linter.

generator/templates/rust/model.mustache
```diff
@@ -7,13 +7,20 @@
...
     {{/DocLines}}
     pub {{NameToSnake}}: {{{FieldType}}},
-    {{/Fields}}
+    {{/BasicFields}}
+    {{#ExplicitOneOfs}}
+
+    {{#DocLines}}
+    /// {{{.}}}
+    {{/DocLines}}
+    pub {{NameToSnake}}: Option<{{{FieldType}}}>,
```
coryan, line 22: I have a pending PR (#96) that would (I think) simplify this kind of code.  IMHO, the `genclient.Field` should always have `Optional: true` for message and oneof fields.  I can do the cleanup after this PR or we can do it in the opposite order, your call.

generator/templates/rust/oneof.mustache
```diff
@@ -0,0 +1,15 @@
...
+#[serde(rename_all = "camelCase", untagged)]
+#[non_exhaustive]
+pub enum {{NameToPascal}} {
+    {{#Fields}}
+    {{#DocLines}}
+    /// {{{.}}}
+    {{/DocLines}}
+    {{NameToPascal}}{{{FieldType}}},
+    {{/Fields}}
+}
```
coryan, line 15: Missing newline at EOF.

generator/internal/genclient/translator/protobuf/testdata/scalar.proto
```diff
@@ -0,0 +1,50 @@
+syntax = "proto3";
+package test;
+
+// A test message.
+message Fake {
+	// A singular field tag = 1
```
coryan, line 6: nit: here and in the other files, the proto files need formatting now.

generator/internal/genclient/translator/protobuf/testdata/scalar.proto
```diff
@@ -0,0 +1,50 @@
+syntax = "proto3";
```
coryan, line 1: nit: almost certainly this will need some copyright boilerplate.  All files that can have the copyright boilerplate are supposed to have it.

generator/internal/genclient/translator/protobuf/testdata/comments.proto
```diff
@@ -0,0 +1,47 @@
...
+  // Yes, this also has a field.
+  string name = 1;
+
+  // Some enum.
+  //
+  // Line 1.
+  // Line 2.
+  enum Status {
+    // The first enum value description.
+	//
```
coryan, line 30: nit: formatting.

generator/internal/genclient/translator/protobuf/protobuf_test.go
```diff
@@ -329,60 +240,69 @@ message Fake {
 				Typez:         genclient.BYTES_TYPE,
 			},
 		},
+		OneOfs: []*genclient.OneOf{
```
coryan, line 243: This is a bit strange.  I get that Protobuf chooses to represent optional fields as one-of fields with a single branch. Do we need to use that representation in our own data structures?

generator/internal/genclient/translator/protobuf/protobuf.go
```diff
@@ -207,17 +207,29 @@ func processMessage(state *genclient.APIState, m *descriptorpb.DescriptorProto,
...
 		field := &genclient.Field{
-			Name:     mf.GetName(),
-			ID:       mFQN + "." + mf.GetName(),
-			JSONName: mf.GetJsonName(),
-			Optional: mf.Proto3Optional != nil && *mf.Proto3Optional,
-			Repeated: mf.Label != nil && *mf.Label == descriptorpb.FieldDescriptorProto_LABEL_REPEATED,
+			Name:            mf.GetName(),
+			ID:              mFQN + "." + mf.GetName(),
+			JSONName:        mf.GetJsonName(),
+			Optional:        mf.Proto3Optional != nil && *mf.Proto3Optional,
```
coryan, line 223: Maybe time to refactor that expression to a variable?

coryan (changes requested):

generator/internal/genclient/translator/protobuf/testdata/scalar.proto
codyoss: ack

generator/internal/genclient/translator/protobuf/testdata/scalar.proto
codyoss: ack

generator/internal/genclient/translator/protobuf/testdata/comments.proto
codyoss: ack

generator/internal/genclient/translator/protobuf/protobuf_test.go
codyoss: yeah, that is a good idea: https://github.com/googleapis/google-cloud-rust/pull/101

generator/internal/genclient/language/internal/rust/rust.go
```diff
@@ -116,6 +125,17 @@ func (c *Codec) FieldType(f *genclient.Field, state *genclient.APIState) string
 		return ""
 	}
 	return ScalarFieldType(f)
+
+}
+
+func (c *Codec) wrapOneOfField(f *genclient.Field, value string) string {
+	if f.Typez == genclient.MESSAGE_TYPE {
+		// Remove optional wrapper for fields of one-of
+		value = value[7 : len(value)-1]
```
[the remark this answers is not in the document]
codyoss: Tried to clear this up in the comment better, but yes. If you prefer a const I can do that too though.

generator/internal/genclient/translator/protobuf/protobuf_test.go
codyoss: So the reason this is important in proto is because it is safe to move optinal into a one-of without a breaking change, because of how this is implemented by protobuf. That is the reason I kept optionals as one-ofs. As written for Rust, this would be a breaking change which I think is fine as long as we know we are making this choice. Other languages, like Go, also implement oneofs in a way where we can't freely move in and out of oneofs. That said other langauges may chose differently than we have with Rust. That is why I left in this information. But if you think we should simpify this in our model and say only explicit oneofs are oneofs I am fine with that, there are just tradeoffs.

generator/internal/genclient/translator/protobuf/protobuf.go
codyoss: done

generator/templates/rust/oneof.mustache
coryan: TIL

coryan (approved):

generator/internal/genclient/translator/protobuf/protobuf_test.go
codyoss: Hmmm, I see your point. I will remove optionals from one-of and say compatibility as an exercise to the impl of a given language.

generator/internal/genclient/language/internal/rust/rust.go
```diff
@@ -101,7 +109,8 @@ func (c *Codec) FieldType(f *genclient.Field, state *genclient.APIState) string
 			return "Option<std::collections::HashMap<" + key + "," + val + ">>"
 		}
 		if strings.HasPrefix(m.ID, ".google.protobuf.") {
-			return "Option<String> /* TODO(#77) - handle " + f.TypezID + " */"
+			// TODO(#77): Better handling of well-known types
```
coryan, line 112: BTW, this was a good idea.  My idea of putting TODOs in the generated code ... not so smart.

coryan (approved):

Issue #77, opened 2024-11-03
Implement well-known types for Rust


coryan:
We want to use custom types (e.g. [this](https://github.com/googleapis/google-cloud-rust/blob/bbddb19cd7a6edae0303a47971b68053a0b4afe6/types/src/timestamp.rs#L48)) for well-known messages.

These include:

- [x] `google.protobuf.Timestamp`
- [x] `google.protobuf.Duration`
- [x] `google.protobuf.Any`
- [x] `google.protobuf.FieldMask`
- [x] `google.protobuf.Empty`
- [ ] `google.protobuf.Struct`
- [x] #265

coryan:
We have enough well-known types to generate the majority of the Google Cloud APIs.  We can add the remaining ones as time allows.

codyoss:
Dupe of #37

PR #100, opened 2024-11-05
impl(types): initial version of `Any`

coryan:
Introduce a Rust type to support the well-known type `.google.protobuf.Any`.
This will probably need more work for a full implementation, but I think it is
enough to unblock development of the generator.

Part of the work for #77

### DIFF
--- a/types/src/any.rs
+++ b/types/src/any.rs
@@ -129,7 +129,7 @@ impl Any {
             .map_err(Self::map_de_str)?;
         let r#type = object["@type"]
             .as_str()
-            .ok_or_else(|| "missing @type field or fild".to_string())
+            .ok_or_else(|| "@type field is missing or is not a string".to_string())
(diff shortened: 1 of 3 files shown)


types/src/any.rs
```diff
@@ -0,0 +1,230 @@
...
+        T: serde::de::DeserializeOwned,
+    {
+        let value = &self.0;
+        let object = value
+            .as_object()
+            .ok_or_else(|| "expected Object value inside Any".to_string())
+            .map_err(Self::map_de_str)?;
+        let r#type = object["@type"]
+            .as_str()
+            .ok_or_else(|| "missing @type field or fild".to_string())
```
codyoss, line 132: nit: needs cleanup

codyoss (approved):

types/src/any.rs
coryan: Thanks, fixed.

PR #102, opened 2024-11-05
impl(generator/protobuf): cleanup map fields

coryan:
Map fields are represented as repeated fields in Protobuf. I think this is
leaking their representation on the wire (a sequence of objects, with `key` and
`value` fields). For our purposes the field is not repeated, and can never be
(`repeated map<K, V> foo = 1;` is not a legal Protobuf field).

Part of the work for #95

### DIFF
--- a/generator/internal/genclient/translator/protobuf/protobuf.go
+++ b/generator/internal/genclient/translator/protobuf/protobuf.go
@@ -167,7 +167,7 @@ var descriptorpbToTypez = map[descriptorpb.FieldDescriptorProto_Type]genclient.T
 	descriptorpb.FieldDescriptorProto_TYPE_ENUM:     genclient.ENUM_TYPE,
 }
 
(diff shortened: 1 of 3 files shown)


generator/internal/genclient/translator/protobuf/protobuf.go
```diff
@@ -182,11 +182,20 @@ func normalizeTypes(in *descriptorpb.FieldDescriptorProto, field *genclient.Fiel
 		// Repeated fields are not optional, they can be empty, but always have
 		// presence.
 		field.Optional = !field.Repeated
+		message, ok := state.MessageByID[field.TypezID]
```
codyoss, line 185: nit: may be reduced to `if message, ok := state.MessageByID[field.TypezID]; ok {`

codyoss (approved):

generator/internal/genclient/translator/protobuf/protobuf.go
coryan: Thanks, done.

PR #104, opened 2024-11-05
impl(generator/rust): support repeated fields

### LINKED ISSUE #95: Rust generator uses  `Vec<T>` for repeated fields

coryan:
The Rust generator now uses `Vec<T>` for repeated fields.

Fixes #95

### DIFF
--- a/generator/testdata/rust/openapi/golden/model.rs
+++ b/generator/testdata/rust/openapi/golden/model.rs
@@ -10,7 +10,7 @@ use serde::{Deserialize, Serialize};
 pub struct ListLocationsResponse {
 
     /// A list of locations that matches the specified filter in the request.
-    pub locations: ,
+    pub locations: Vec<>,
 
     /// The standard List next-page token.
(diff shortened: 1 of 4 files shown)


generator/testdata/rust/openapi/golden/model.rs
```diff
@@ -10,7 +10,7 @@ use serde::{Deserialize, Serialize};
 pub struct ListLocationsResponse {
 
     /// A list of locations that matches the specified filter in the request.
-    pub locations: ,
+    pub locations: Vec<>,
```
codyoss, line 13: nit: this looks like it improved, but something is still missing. Well know cloud location.

codyoss (approved):

generator/testdata/rust/openapi/golden/model.rs
coryan: I think the merge with #103 will fix those too.

PR #106, opened 2024-11-05
fix(internal/genclient): use service config name

### LINKED ISSUE #38: extract the service name from proto or service config
https://github.com/googleapis/google-cloud-rust/blob/77720dee3887853cc65bf31c90ce887b6e629976/generator/src/genclient/translator/protobuf/protobuf.go#L65-L66

julieqiu:
Fixes #38

### DIFF
--- a/generator/internal/genclient/translator/protobuf/protobuf.go
+++ b/generator/internal/genclient/translator/protobuf/protobuf.go
@@ -55,29 +55,30 @@ type Options struct {
 
 // Translate translates proto representation into a [genclienGenerateRequest].
 func Translate(req *pluginpb.CodeGeneratorRequest, opts *Options) (*genclient.GenerateRequest, error) {
-	api := makeAPI(req)
-
 	codec, err := language.NewCodec(opts.Language)
 	if err != nil {
 		return nil, err
 	}
+	var name string
+	if opts.ServiceConfig != nil {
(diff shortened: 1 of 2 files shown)


generator/internal/genclient/translator/protobuf/protobuf.go
```diff
@@ -55,29 +55,30 @@ type Options struct {
...
 func Translate(req *pluginpb.CodeGeneratorRequest, opts *Options) (*genclient.GenerateRequest, error) {
-	api := makeAPI(req)
-
 	codec, err := language.NewCodec(opts.Language)
 	if err != nil {
 		return nil, err
 	}
+	var name string
+	if opts.ServiceConfig != nil {
+		name = opts.ServiceConfig.Name
```
coryan, line 64: I think this is an improvement, but I am not sure it is exactly what we want.  For our test service that would be "secretmanager.googleapis.com".  Or maybe we want this, but we strip out the `.googleapis.com` to create crate names and the such.

No action required.

coryan (approved):

generator/internal/genclient/translator/protobuf/protobuf.go
julieqiu: Ack that's a good point.

PR #110, opened 2024-11-06
impl(generator/rust): generate basic crate scaffold

coryan:
This generates the basic scaffold to make the generated code "compilable". The
contents of the files is known to be incomplete, this PR just creates enough
structure for us to start finding out build problems.


codyoss (approved):
Nice!

PR #116, opened 2024-11-06
impl(generator): MOAR service config information

coryan:
With this PR the OpenAPI variant also reads the service config file. We also
capture more information from the service config file: the API title and
summary are useful to generate acceptable README files.

### DIFF
--- a/generator/cmd/openapi/main.go
+++ b/generator/cmd/openapi/main.go
@@ -24,23 +24,33 @@ import (
 	"github.com/googleapis/google-cloud-rust/generator/internal/genclient/translator/openapi"
 )
 
+var (
+	language      = flag.String("language", "", "the generated language")
+	output        = flag.String("out", "output", "the path to the output directory")
+	templateDir   = flag.String("template-dir", "templates/", "the path to the template directory")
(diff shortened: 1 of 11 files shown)


generator/cmd/openapi/main.go
```diff
@@ -24,23 +24,33 @@ import (
 	"github.com/googleapis/google-cloud-rust/generator/internal/genclient/translator/openapi"
 )
 
+var (
+	language      = flag.String("language", "", "the generated language")
```
codyoss, line 28: nit: any reason for moving these out of the function call? They are now globally accessible -- to this package.

codyoss (approved):

generator/cmd/openapi/main.go
coryan: I was just following the style in:

https://github.com/googleapis/google-cloud-rust/blob/61b9d3bbac5530e4321ac19fe7d2760db82e31db/generator/devtools/cmd/generate/main.go#L27-L33

I am not sure what is right, I will leave it to you and @julieqiu to set a standard.

PR #119, opened 2024-11-06
impl(generator/rust): remove more trailing spaces

coryan:
Removing a few more trailing spaces from the generated code would allow us to
put the golden files in the CI build. I am sure eventually we will need to run
`cargo fmt` as part of the code generation. On the other hand, this is an easy
change and it unblocks faster testing on the generated code.


generator/templates/rust/src/lib.rs.mustache
```diff
@@ -42,7 +42,7 @@ impl Client {
     {{#Services}}
 
     {{#DocLines}}
-    /// {{{.}}}
+    {{{.}}}
```
codyoss, line 45: future: maybe we just have this return a string instead of slice of strings.

codyoss (approved):

PR #122, opened 2024-11-07
test(generator): compile Rust golden files

coryan:
Hardcode a few things to get the code compiling.

I want to add these files to the Rust CI builds, and that includes enforcing
any formatting for them. I could have fixed the generator to correct some of
the formatting problems, but the last few nits would have complicated the
templates too much.

Whenever the generator gets moved to its own repository: we should remove this
step. That will complicate the tooling for generator developers.

I left a number of TODO entries (pointing to bugs).  I think the changes in the CI scripts are necessary if we want to compile the golden files, but they may also be a reason to **not** merge this PR. I thought the discussion would be easier with a visible PR.

### DIFF
--- a/generator/cmd/protoc-gen-gclient/main_test.go
+++ b/generator/cmd/protoc-gen-gclient/main_test.go
@@ -15,8 +15,10 @@
 package main
 
 import (
+	"errors"
 	"flag"
 	"os"
+	"os/exec"
 	"testing"
 )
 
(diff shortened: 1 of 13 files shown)


generator/templates/rust/src/lib.rs.mustache
```diff
@@ -13,6 +13,9 @@ WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 See the License for the specific language governing permissions and
 limitations under the License.
 }}
+{{! TODO(#120) - re-enable dead_code warnings once OpenAPI generates some functions }}
+#![allow(dead_code)]
```
codyoss, line 17: nit: I think this may likely always be needed unless we write tests for every generated method, right?

generator/cmd/protoc-gen-gclient/main_test.go
```diff
@@ -34,4 +36,12 @@ func TestRun_Rust(t *testing.T) {
 	if err := run(os.Stdin, os.Stdout, inputPath, outDir, templateDir); err != nil {
 		t.Fatal(err)
 	}
+
+	cmd := exec.Command("cargo", "fmt", "--manifest-path", outDir+"/Cargo.toml")
```
codyoss, line 40: nit: Maybe we should expose a post-processing hook in the Codec that takes the outDir and can manipulate the sources as desired? Feels like a smell to call this from a Go test imo.

generator/templates/rust/src/lib.rs.mustache
coryan: No, once the define some public methods that use all the private stuff we will be fine.  The issue here is that the OpenAPI translator does not create any public APIs yet.

codyoss (approved):
ooops, meant to approve this the first time. I think this is fine for now but could be re-visited later.

PR #124, opened 2024-11-07
test(generator): use `protoc` to parse test files

coryan:
I could not get `bufbuild/protocompile` to handle all the complex options we
use in our service proto files. Since we use files to keep the test input, we
can just call `protoc` to parse the file.

Motivated by #120: I will need to do some refactoring in protobuf.go, and I would prefer having some unit tests before doing that.

### DIFF
--- a/generator/internal/genclient/translator/protobuf/protobuf_test.go
+++ b/generator/internal/genclient/translator/protobuf/protobuf_test.go
@@ -15,17 +15,16 @@
 package protobuf
 
 import (
-	"context"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"testing"
 
-	"github.com/bufbuild/protocompile"
 	"github.com/google/go-cmp/cmp"
 	"github.com/google/go-cmp/cmp/cmpopts"
 	"github.com/googleapis/google-cloud-rust/generator/internal/genclient"
 	"google.golang.org/genproto/googleapis/api/serviceconfig"
-	"google.golang.org/protobuf/reflect/protodesc"
+	"google.golang.org/protobuf/proto"
 	"google.golang.org/protobuf/types/descriptorpb"
 	"google.golang.org/protobuf/types/pluginpb"
 )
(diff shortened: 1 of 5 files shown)


generator/internal/genclient/translator/protobuf/protobuf_test.go
```diff
@@ -485,32 +485,49 @@ func TestMapFields(t *testing.T) {
...
-		filename: string(contents),
-	})
-	compiler := protocompile.Compiler{
-		Resolver:       &protocompile.SourceResolver{Accessor: accessor},
-		MaxParallelism: 1,
-		SourceInfoMode: protocompile.SourceInfoStandard,
+	defer os.Remove(tempFile.Name())
+
+	var stderr bytes.Buffer
+	cmd := exec.Command("protoc",
```
codyoss, line 495: In the README for the generator can we provide a link to where to download this. I am fine doing this here but it then requires/assumes protoc in the host environment to run tests.

generator/internal/genclient/translator/protobuf/protobuf_test.go
```diff
@@ -485,32 +485,49 @@ func TestMapFields(t *testing.T) {
...
+		"--proto_path", "../../../../testdata/googleapis",
+		"--include_imports",
+		"--include_source_info",
+		"--retain_options",
+		"--descriptor_set_out", tempFile.Name(),
+		filepath.Join("testdata", filename))
+	cmd.Stderr = &stderr
+	err = cmd.Run()
+	if err != nil {
+		t.Errorf("protoc error: %s", stderr.String())
```
codyoss, line 506: nit:
```suggestion
		t.Logf("protoc error: %s", stderr.String())
```

codyoss (approved):

generator/internal/genclient/translator/protobuf/protobuf_test.go
coryan: Done.

PR #127, opened 2024-11-07
impl(generator/protobuf): new representation for HTTP RPCs

coryan:
This introduces an (at the moment) alternative representation for HTTP RPCs. I
cannot produce the previous representation from the OpenAPI specification, and
it seems overly difficult to use. In a pending PR I will change the Codecs to
use `PathInfo` and then remove `genclient.HTTPInfo`.

Motivated by #120


codyoss (approved):
This makes sense to me. The previous data model was very coupled to proto.

Issue #132, opened 2024-11-08
Generate fluent modifiers for Rust `struct`


coryan:
@codyoss explains the problem well in:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=ace8701471ccbc16583f62329e809ee5

We expect request and response messages to gain fields over time, and we do not want to break applications when that happens. The only safe way to do this is to create messages with fluent initializers.  Applications would need to say:

```rust
  let request = Request::default().set_parent("projects/my-project").set_filter("only the good stuff");
  client.MakeRequest(request)....
```

The fields will remain public, applications can access them and change them if they prefer to do so:

```rust
  let request = Request::default() ... ...;
  request.another_thing = "blah";
  let request = request.and_more_stuff(42);
  assert_eq!(42, request.and_more_stuff); 
  let response = client.MakeRequest(request)
```

Issue #129, opened 2024-11-07
Query parameters can be more than strings


coryan:
At least with the OpenAPI specification, some query parameters can be integers:

https://github.com/googleapis/google-cloud-rust/blob/920f876f910115830123e6f45c1b56f418ac6484/generator/testdata/openapi/secretmanager_openapi_v1.json#L133-L141

Or `google.protobuf.FieldMask`:

https://github.com/googleapis/google-cloud-rust/blob/920f876f910115830123e6f45c1b56f418ac6484/generator/testdata/openapi/secretmanager_openapi_v1.json#L751-L761

Seems like we need a more generic treatment than `req.${param}.as_str()`.  I tried `format!("{}", req.${param)` but that fails for the field mask.

PR #134, opened 2024-11-08
fix: make types usable outside the crate

coryan:
`FieldMask` had private fields when they should have been public. We were
missing setters for a number of fields.

I refactored some of the tests to be integration tests, so we can verify the
intended usage outside the crate.

Motivated by #132 and #129


codyoss (approved):
I like the idea of integration tests, nice!

Issue #128, opened 2024-11-07
Support optional query parameters for Rust


coryan:
The generated code for query parameters looks like so:

https://github.com/googleapis/google-cloud-rust/blob/920f876f910115830123e6f45c1b56f418ac6484/generator/testdata/rust/gclient/golden/src/lib.rs#L70

This does not work if the parameter is an `Option<T>`.  We need something different that works for required and optional fields.

PR #135, opened 2024-11-08
impl: helpers for query parameters

coryan:
I want to keep the generated code for query parameters relatively simple, and
at the same time I want to support optional query parameters, required query
parameters, query parameters of different types, including some well-known
types.

This introduces a new crate[^1] which initially just has a helper function to
convert all the expected types to strings. The integration tests show how this
helper would be used in generated code.

[^1]: curiously named `gax` for no reason I can think of

Part of the work for #128 and #129

### DIFF
--- a/gax/src/query_parameter.rs
+++ b/gax/src/query_parameter.rs
@@ -1,0 +1,162 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub type Error = serde_json::Error;
+type Result = serde_json::Result<String>;
+
+pub fn format<T>(
+    name: &'static str,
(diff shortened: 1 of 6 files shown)


gax/src/query_parameter.rs
```diff
@@ -0,0 +1,162 @@
...
+
+impl RequiredQueryParameter for f64 {
+    fn format(&self) -> Result {
+        Ok(format!("{self}"))
+    }
+}
+
+impl RequiredQueryParameter for types::Duration {
+    fn format(&self) -> Result {
+        Ok(serde_json::to_value(self)?.as_str().unwrap().to_string())
```
codyoss, line 95: Since we are already calling unwrap here and we control the input type, could we simplify this and remove the need to return a result at all? I think we can guarantee that  our types will always serialize.

coryan: That `unwrap` is just turning `Option<&str>` into `&str`.  I am using secret knowledge about how `Duration` always serializes to a JSON string.  Maybe a cleaner approach would be to provide a function in `Duration` to format as a string, and then use that here (and in the serializer).

To your larger point: I am not sure it is true that we can always serialize the values we have. For example:  Timestamps can only be (safely) serialized in a certain range:

https://github.com/protocolbuffers/protobuf/blob/d0aefcba0ea1dd814c12915685eda9e953df49bb/src/google/protobuf/timestamp.proto#L53-L55

We currently do not enforce this restriction, but I can see us returning an error if the timestamp cannot be formatted correctly because it is out of range.

codyoss (approved):

gax/src/query_parameter.rs
codyoss: hmm that is fair, proceed!

PR #138, opened 2024-11-08
impl(generator/openapi): generate service and RPCs

### LINKED ISSUE #120: Implement Client functions for the OpenAPI-based generator

### LINKED ISSUE #128: Support optional query parameters for Rust

### LINKED ISSUE #129: Query parameters can be more than strings

coryan:
Use the information from the `paths` section to find out what RPCs are
available. They all go into the same service, because there is nothing obvious
to use.

We create synthetic request objects because the OpenAPI spec does not seem to
have them for all RPCs. We inject the require path and query parameters into
the request object, whether that is synthetic or included by the specification.

Fixes #120 and fixes #128 and fixes #129

### DIFF
--- a/generator/testdata/rust/openapi/golden/src/model.rs
+++ b/generator/testdata/rust/openapi/golden/src/model.rs
@@ -149,6 +149,25 @@ pub struct Secret {
     /// SecretVersions added afterwards. They do not apply
     /// retroactively to existing SecretVersions.
     pub customer_managed_encryption: Option<crate::model::CustomerManagedEncryption>,
+
+    ///
+    pub project: String,
+
+    /// Required. This must be unique within the project.
+    ///
+    /// A secret ID is a string with a maximum length of 255 characters and can
+    /// contain uppercase and lowercase letters, numerals, and the hyphen (`-`) and
+    /// underscore (`_`) characters.
+    pub secret_id: String,
+
+    ///
+    pub location: String,
+
+    ///
+    pub secret: String,
+
+    /// Required. Specifies the fields to be updated.
+    pub update_mask: gax_placeholder::FieldMask,
 }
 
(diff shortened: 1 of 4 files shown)


generator/testdata/rust/openapi/golden/src/model.rs
```diff
@@ -149,6 +149,25 @@ pub struct Secret {
     /// SecretVersions added afterwards. They do not apply
     /// retroactively to existing SecretVersions.
     pub customer_managed_encryption: Option<crate::model::CustomerManagedEncryption>,
+
+    ///
```
codyoss, line 153: Do we have an issue for this?

codyoss (approved):

generator/testdata/rust/openapi/golden/src/model.rs
coryan: We do now, #147.

PR #146, opened 2024-11-11
fix(generator/protobuf): support top-level enums in proto files

codyoss:


coryan (commented):
I was wondering about these. Thanks!

coryan (approved):

Issue #140, opened 2024-11-09
Support path parameters in Rust `Option<>` fields


coryan:
Some path parameters are of the form `a.b`:

https://github.com/googleapis/google-cloud-rust/blob/c9f74adf12932f910aa69c167a9ca5ec0fe258fd/generator/testdata/googleapis/google/cloud/secretmanager/v1/service.proto#L88

In `Rust` the first component is an `Option<Secret>`, as all message fields are optional.  But when used as a path parameter the field is **required**.  We need the generator to support these.  I think it can be implemented with some help from `gax`.

coryan:
We need more than #141 to fix this.

PR #149, opened 2024-11-11
feat(gax): support optional path parameters

coryan:
Sometimes path parameters appear inside optional fields.  Consider:

https://github.com/googleapis/google-cloud-rust/blob/c9f74adf12932f910aa69c167a9ca5ec0fe258fd/generator/testdata/googleapis/google/cloud/secretmanager/v1/service.proto#L88

That requests the `.name` field from the `.secret` field in the request.  But
`.secret` is mapped to `Option<Secret>` because it is a message. We need to say
something like `.secret.ok_or_else(...)?.name`. This PR introduces some helpers
to simplify the generator. The generator always calls the
`PathParameter::required()` function. The function expands to either `self` or
`self.ok_or_else()` depending on whether the parameter is an `Option<_>` or not.

A future PR will take advantage of these helpers in the generated code.

Part of the work for #140

### DIFF
--- a/gax/src/path_parameter.rs
+++ b/gax/src/path_parameter.rs
@@ -1,0 +1,72 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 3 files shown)


gax/src/path_parameter.rs
```diff
@@ -0,0 +1,72 @@
...
+    T: crate::request_parameter::RequestParameter,
+{
+    type P = T;
+    fn required<'a>(&'a self, _: &str) -> std::result::Result<&'a Self::P, Error> {
+        Ok(self)
+    }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum Error {
```
codyoss, line 39: nit: non_exhaustive

codyoss (approved):

Issue #150, opened 2024-11-11
Generate rust code for the full `secretmanager/v1/service.proto` file


coryan:
We currently generate a subset of the API, we need multiple changes to generate the full API.

Known problems include:

- [ ] #156
- [ ] #157
       - For example:  `model::GaxPlaceholderEmpty`.
- [ ] #158

coryan:
Closing. I broke down the remaining work into smaller tasks.

PR #151, opened 2024-11-11
impl(generator/protobuf): skip imported messages

coryan:
Imported messages should not be generated, so we exclude them from the
`api.Messages` list. They do need to be in the `api.State.MessageByID` index,
as these messages may be referenced by other messages or services.

Part of the work for #150


codyoss (approved):
Nice!

PR #154, opened 2024-11-11
feat(generator/rust): use gax helpers for path parameters

### LINKED ISSUE #140: Support path parameters in Rust `Option<>` fields

coryan:
Fixes #140 for realsies

### DIFF
--- a/generator/internal/genclient/language/internal/rust/rust.go
+++ b/generator/internal/genclient/language/internal/rust/rust.go
@@ -231,6 +231,34 @@ func (c *Codec) HTTPPathFmt(m *genclient.PathInfo, state *genclient.APIState) st
 	return fmt
 }
 
+// Returns a Rust expression to access (and if needed validatre) each path parameter.
+//
+// In most cases the parameter is a simple string in the form `name`. In those
+// cases the field *must* be a thing that can be formatted to a string, and
+// a simple "req.name" expression will work file.
+//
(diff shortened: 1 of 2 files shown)


generator/internal/genclient/language/internal/rust/rust.go
```diff
@@ -231,11 +231,26 @@ func (c *Codec) HTTPPathFmt(m *genclient.PathInfo, state *genclient.APIState) st
 	return fmt
 }
 
+func (c *Codec) unwrapFieldPath(components []string, requestAccess string) (string, string) {
```
codyoss, line 234: nit: can you add a doc comment to this method to describe what is being returned.

codyoss (approved):

generator/internal/genclient/language/internal/rust/rust.go
coryan: Good idea, done.

PR #155, opened 2024-11-11
test(generator/protobuf): larger golden files

coryan:
Part of the work for #150


codyoss (approved):

codyoss:
Note this is the second copy of these protos checked into this repo now. This one without Iam for simplicity I assume in generation. We should eventually remove this copy.

coryan:
> Note this is the second copy of these protos checked into this repo now. This one without Iam for simplicity I assume in generation. We should eventually remove this copy.

Ack, and agreed.

Issue #158, opened 2024-11-11
Cannot use the external `google.iam.v1.*` types in Rust


coryan:
We need to create the crate for these types, and inject the dependency in the Cargo.toml file

PR #166, opened 2024-11-12
impl(generator): configurable packages for Rust codec

coryan:
Part of the work for #158

### DIFF
--- a/generator/internal/genclient/language/internal/rust/rust.go
+++ b/generator/internal/genclient/language/internal/rust/rust.go
@@ -17,6 +17,7 @@ package rust
 import (
 	"fmt"
 	"log/slog"
+	"sort"
 	"strings"
 	"unicode"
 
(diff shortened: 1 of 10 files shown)


generator/internal/genclient/language/internal/rust/rust.go
```diff
@@ -369,6 +431,25 @@ func (*Codec) FormatDocComments(documentation string) []string {
 	return ss
 }
 
+func (c *Codec) RequiredPackages() []string {
+	lines := []string{}
```
codyoss, line 435: nit: you don't need to allocate slices in Go because of the append idiom

```suggestion
	var lines []string
```

codyoss (approved):

generator/internal/genclient/language/internal/rust/rust.go
coryan: I am going to merge as-is to unblock other work.  Could you take a few minutes to answer a basic golang question?

In this case there may be no `append()` call.  Is it safe to return a `nil` lines if that is the case?  I will send a follow up PR if that is true.

Issue #175, opened 2024-11-13
Create common error types for rust


codyoss:
- [x] Definitions for error_details
- [x] Definitions for status
- [x] Core error type
- [x] Standard HTTP error
- [ ] Update templates to use new error types
- [x] Add integration tests for converting to HTTP errors and getting rpc::Status

Issue #98, opened 2024-11-05
Rust messages need a typename to serialize `.google.protobuf.Any`


coryan:
The JSON encoding of `.google.protobuf.Any` includes the name of the type:

https://github.com/protocolbuffers/protobuf/blob/10c20cefd1f724031760b2d0db725968f2fee41c/src/google/protobuf/any.proto#L112-L116

To serialize arbitrary messages from an `Any` into JSON we would need to add this field, which means we need a generic way to get the type name.  That suggests some trait that all generated messages must implement.

To deserialize arbitrary messages from an `Any` we are supposed to validate the type using this field.  Though we can just ignore this field to start.

coryan:
extra fun: I am not sure we get the right type names from something like an OpenAPI v3 discovery doc.

PR #177, opened 2024-11-13
feat(gax): add representation for google.rpc.Status

codyoss:
Absent from the tests are dealing with the `Other` variant. This is due to some known issues with our implmentation of Any and the handling of `@type`. Eventually we may need custom serde handling here, but for now this works.

Updates: #175
Updates: #98


gax/src/error/rpc/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+
+/// The [Status] type defines a logical error model that is suitable for
+/// different programming environments, including REST APIs and RPC APIs. It is
+/// used by [gRPC](https://github.com/grpc). Each [Status] message contains
+/// three pieces of data: error code, error message, and error details.
+///
+/// You can find out more about this error model and how to work with it in the
+/// [API Design Guide](https://cloud.google.com/apis/design/errors).
+#[derive(Clone, Debug, Default, PartialEq, serde::Deserialize, serde::Serialize)]
+#[serde(rename_all = "camelCase")]
```
coryan, line 26: Should we add `#[non_exhaustive]`?

gax/src/error/rpc/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+mod error_details;
+pub use error_details::*;
+
+/// The [Status] type defines a logical error model that is suitable for
+/// different programming environments, including REST APIs and RPC APIs. It is
+/// used by [gRPC](https://github.com/grpc). Each [Status] message contains
```
coryan, line 20: I would omit the reference to gRPC here, since we are planning to use this for mostly HTTP+JSON clients?

gax/src/error/rpc/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+/// three pieces of data: error code, error message, and error details.
+///
+/// You can find out more about this error model and how to work with it in the
+/// [API Design Guide](https://cloud.google.com/apis/design/errors).
+#[derive(Clone, Debug, Default, PartialEq, serde::Deserialize, serde::Serialize)]
```
coryan, line 25: FYI: in hand-crated code I think we case `use serde::Deserialize` with abandon.  It is only in generated code where we have to preemptively protect against name clashes.

gax/src/error/rpc/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+#[serde(rename_all = "camelCase")]
+pub struct Status {
+    pub code: i32,
```
coryan, line 28: We should have a list of constants (or an enum?) for the well-known values of this code.

coryan (approved):
A few minor things that you can fix and then merge.

I wonder if we should have an integration test showing how to use this class from outside the crate, since it is intended to be a public API.

gax/src/error/rpc/mod.rs
codyoss: made an enum

PR #182, opened 2024-11-15
impl(generator): full service config parser

coryan:
Parsing via yaml -> jsonproto -> proto works for enums as strings, and
for well-know protos and for a number of other corner cases.

Part of the work for #158

### DIFF
--- a/generator/internal/genclient/parser/openapi/openapi.go
+++ b/generator/internal/genclient/parser/openapi/openapi.go
@@ -151,7 +151,10 @@ func makeServices(api *genclient.API, serviceConfig *serviceconfig.Service, mode
 	if serviceConfig != nil {
 		for _, api := range serviceConfig.Apis {
 			serviceName = api.Name
-			break
+			// Keep searching after well-known mixin services.
+			if !strings.HasPrefix(api.Name, "google.cloud.location.Location") && !strings.HasPrefix(api.Name, "google.iam.v1.IAMPolicy") {
+				break
+			}
 		}
 	}
 	service := &genclient.Service{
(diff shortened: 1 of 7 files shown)


generator/internal/genclient/parser/openapi/openapi.go
```diff
@@ -151,7 +151,10 @@ func makeServices(api *genclient.API, serviceConfig *serviceconfig.Service, mode
 	if serviceConfig != nil {
 		for _, api := range serviceConfig.Apis {
 			serviceName = api.Name
-			break
+			// Keep searching after well-known mixin services.
+			if !strings.HasPrefix(api.Name, "google.cloud.location.Location") && !strings.HasPrefix(api.Name, "google.iam.v1.IAMPolicy") {
```
codyoss, line 155: The other mixin we will need to handle is `google.longrunning.Operations`

generator/internal/genclient/service_config.go
```diff
@@ -18,8 +18,10 @@ import (
 	"fmt"
 	"os"
 
+	"github.com/ghodss/yaml"
 	"google.golang.org/genproto/googleapis/api/serviceconfig"
-	"gopkg.in/yaml.v3"
+	"google.golang.org/protobuf/encoding/protojson"
+	_ "google.golang.org/protobuf/types/known/wrapperspb"
```
codyoss, line 24: What is this needed for? If it is needed for something could you add a comment.

generator/internal/genclient/parser/openapi/openapi.go
coryan: Fixed.

generator/internal/genclient/service_config.go
coryan: Not needed, I neglected to remove this after adding it during troubleshooting. Removed.

Issue #188, opened 2024-11-15
Automate common library maintenance tasks


coryan:
We will need to repeatedly run the generator to keep the protos up to date.  The generator should gain some subcommands so we can run things like:

- [x] Generate a library and save the command-line flags used to generate it (I am thinking a `.sidekick.toml` file).
- [x] Refresh / regenerate a client library using the saved command-line arguments.
- [x] #261
   - For example, the language is `rust` in the `google-cloud-rust` directory.
   - Accept per-directory overrides for tests and other edge cases.
- [x] #263
   - `sidekick` should download the specified version in that case, and provide the path to each directory.
   - the generated directory should use the global setting for this.
- [x] #270
- [x] #271

coryan:
The `generate` and `refresh` commands work.  `generate` saves the options in a `.sidekick.toml` file, that `refresh` can use to do its job.

coryan:
This is now done.

PR #193, opened 2024-11-15
refactor(generator): prepare for subcommands

coryan:
Split the code a bit so it is easier to introduce subcommands.

Part of the work for #188

### DIFF
--- a/generator/cmd/generate.go
+++ b/generator/cmd/generate.go
@@ -1,0 +1,78 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


generator/cmd/generate.go
```diff
@@ -0,0 +1,78 @@
...
+import (
+	"flag"
+	"fmt"
+	"strings"
+
+	"github.com/googleapis/google-cloud-rust/generator/internal/genclient"
+)
+
+func Generate(args []string) error {
+	fs := flag.NewFlagSet("generate", flag.ExitOnError)
```
codyoss, line 26: This is fine for now, and maybe will continue to be in the future. In a previous life I have used [cobra](https://github.com/spf13/cobra) to help manage building a nice cli. Just thought I would share.

codyoss (approved):

generator/cmd/generate.go
coryan: Ack.  I think that is the thing that @julieqiu was advising against.  Please invite me when you discuss and I will bring popcorn.

PR #194, opened 2024-11-15
refactor(generator): test using command-line args

coryan:
Change the main program test to use command-line arguments. That makes
the test more comprehensive, and it helps me document the commands in
the README file.

Part of the work for #188

### DIFF
--- a/generator/README.md
+++ b/generator/README.md
@@ -48,8 +48,6 @@ go run ./cmd -project-root=.. generate \
   -codec-option package:gax=package=gax,path=gax,feature=sdk_client
 ```
 
-[Secret Manager]: https://cloud.google.com/secret-manager/
-
 ## Installing `protoc`: the Protobuf Compiler
 
 The unit tests use `protoc` to parse text `.proto` files. You will need this
@@ -73,3 +71,4 @@ we suggest two approaches:
 
 [grpc tutorial]: https://grpc.io/docs/protoc-installation/
 [protobuf-latest]: https://github.com/protocolbuffers/protobuf/releases/latest
+[secret manager]: https://cloud.google.com/secret-manager/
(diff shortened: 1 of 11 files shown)


generator/cmd/sidekick.go
```diff
@@ -28,6 +28,14 @@ func main() {
 }
 
 func root() error {
+	var projectRoot = flag.String("project-root", "", "the root of the output project")
```
codyoss, line 31: nit: eventually I would move this tree of files into a directory called sidekick. This makes it easier for people to install the program with a good name. If you run `go install github.com/googleapis/google-cloud-rust/generator/cmd` today it will install a binary on your machine named `cmd`.

generator/README.md
```diff
@@ -1,31 +1,51 @@
...
 ```bash
-go install ./cmd/protoc-gen-gclient
-
-protoc -I testdata/googleapis \
-    --gclient_out=. \
-    --gclient_opt=capture-input=true,language=rust \
-    testdata/googleapis/google/cloud/secretmanager/v1/resources.proto \
-    testdata/googleapis/google/cloud/secretmanager/v1/service.proto
+cd generator
+go run ./cmd -project-root=.. generate \
```
codyoss, line 17: nit: consider using the fully qualified install path so that this command can be run from anywhere. Also, command may change from suggestion based on another comment.

```suggestion
go run github.com/googleapis/google-cloud-rust/generator/cmd@latest -project-root=.. generate \
```

codyoss (approved):

generator/cmd/sidekick.go
coryan: Ack will do.

coryan: (in a future PR)

generator/README.md
coryan: Done.

PR #195, opened 2024-11-15
feat(generator): new refresh command

coryan:
Part of the work for #188


coryan:
I will do the rename in the next PR.

generator/testdata/rust/gclient/golden/iam/v1/.sidekick.toml
```diff
@@ -0,0 +1,15 @@
+[General]
+SourceFormat = 'protobuf'
```
codyoss, line 2: nit: I think we add some toml struct tags so casing is a little more consistent. For examples see: https://pkg.go.dev/github.com/pelletier/go-toml/v2#example-Marshal-Commented

codyoss (approved):

generator/testdata/rust/gclient/golden/iam/v1/.sidekick.toml
coryan: Good idea. I made the command-line options and TOML entries have the same names too.

PR #207, opened 2024-11-18
feat(generator/rust): support complex query parameters

### LINKED ISSUE #181: Decide what to do about `GetIamPolicy()` and `body: *`
@codyoss says:

nit: per https://google.aip.dev/127 Gets don't have a body

----

Ack, though:

https://github.com/googleapis/googleapis/blob/14e5759e6036ea42178a7e9cf2490d72cb6813a2/google/iam/v1/iam_policy.proto#L76-L80

Maybe we need to special case `GetIamPolicy()`?

_Originally posted by @coryan in https://github.com/googleapis/google-cloud-rust/pull/180#discussion_r1843128247_

coryan:
With this change we support complex query parameters, such as repeated fields
and object fields.

For object fields, we first serialize the object to JSON, and then recursively
add any fields.  That is relatively slow, but it is also relatively rare.

Fixes #181


generator/testdata/rust/gclient/golden/secretmanager/src/lib.rs
```diff
@@ -301,22 +273,18 @@ impl SecretManagerService {
...
         req: crate::model::ListSecretVersionsRequest,
     ) -> Result<crate::model::ListSecretVersionsResponse> {
-        let query_parameters = [
-            gax::query_parameter::format("pageSize", &req.page_size).map_err(Error::other)?,
-            gax::query_parameter::format("pageToken", &req.page_token).map_err(Error::other)?,
-            gax::query_parameter::format("filter", &req.filter).map_err(Error::other)?,
-        ];
         let client = self.client.inner.clone();
-        let res = client
+        let builder = client
```
codyoss, line 277: Should we make this mut and remove the following lets?

gax/src/query_parameter.rs
```diff
@@ -14,44 +14,91 @@
...
-        .transpose()
+    QueryParameter::add(parameter, builder, name)
 }
 
 /// [QueryParameter] is a trait representing types that can be used as a query
 /// parameter.
-///
 pub trait QueryParameter {
-    fn format(&self) -> Option<Result<String>>;
+    fn add(&self, builder: reqwest::RequestBuilder, name: &str) -> Result<reqwest::RequestBuilder>;
```
codyoss, line 47: Seeing the generated code makes me wonder if we should have a query builder type. The shape here is trying to be a method on RequestBuilder, but of course we can't really do that. Just a thought for the future...

codyoss (approved):

gax/src/query_parameter.rs
coryan: Ack.

generator/testdata/rust/gclient/golden/secretmanager/src/lib.rs
coryan: Something like this does not work:

```rust
        let mut builder = client
            .http_client
            .get(format!("{}/v1/{}/secrets", self.base_path, req.parent,))
            .query(&[("alt", "json")]);
        gax::query_parameter::add(builder, "pageSize", &req.page_size).map_err(Error::other)?;
        gax::query_parameter::add(builder, "pageToken", &req.page_token)
            .map_err(Error::other)?;
        gax::query_parameter::add(builder, "filter", &req.filter).map_err(Error::other)?;
```

The problem is that `builder` gets moved on the first `::add` call.  We could try to pass a `&mut RequestBuilder`, but the `.query()` method expects a value:

https://docs.rs/reqwest/latest/reqwest/struct.RequestBuilder.html#method.query

Open to other ideas if you have them, but I think we can merge this PR and clean this up later?

PR #213, opened 2024-11-19
fix(wkt): handle empty fieldmask

coryan:
Clippy told me my test was not doing what I tought it did, and when I
fixed the test I found a bug.


codyoss (approved):
The power of a strong type system and tools!

PR #217, opened 2024-11-20
fix(generator/openapi): iterate on synthetic messages

coryan:
We create synthetic request messages because OpenAPI does not always
include them. That would be fine, except that sometimes they do include
the `*Request` message and we need to modify it to add back the path
parameters and inject the body parameter.

I was (incorrectly) modifying the body parameter in some cases.

### DIFF
--- a/generator/internal/genclient/parser/openapi/openapi.go
+++ b/generator/internal/genclient/parser/openapi/openapi.go
@@ -215,17 +215,13 @@ func makeMethods(api *genclient.API, model *libopenapi.DocumentModel[v3.Document
 			if op.Operation == nil {
 				continue
 			}
-			requestMessage, err := makeRequestMessage(api, op.Operation, pattern)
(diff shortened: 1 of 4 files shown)


generator/internal/genclient/parser/openapi/openapi.go
```diff
@@ -269,7 +265,7 @@ func makePathTemplate(template string) []genclient.PathSegment {
 	return segments
 }
 
-func makeRequestMessage(api *genclient.API, operation *v3.Operation, template string) (*genclient.Message, error) {
+func makeRequestMessage(api *genclient.API, operation *v3.Operation, template string) (*genclient.Message, string, error) {
```
codyoss, line 268: nit: could you add a comment to this function as it is no longer clear what is being returned by reading the signature.

codyoss (approved):

generator/internal/genclient/parser/openapi/openapi.go
coryan: Done.

PR #219, opened 2024-11-20
test: all RPCs for secret manager + protobuf

coryan:
Test all the RPCs for Secret Manager over Protobuf. I do not believe we
should do this for each client, but at least one client should run the
full set.

We will also want at least one integration test for LROs. More
generally, one for AIP affecting the generator behavior.


codyoss (approved):
Nice! It is good to see what writing the code looks like currently. I think there is some usability stuff we can clean up with some tricks in future PRs.

Issue #209, opened 2024-11-19
generate some Go code


codyoss:
We should make sure we are not too tightly coupling the data model to Rust. We should add a test to generate correctish Go code as well.

PR #220, opened 2024-11-20
feat(generator): add Go iam service generation

codyoss:
It compiles, did not test running it.

- Added lightweight wkt package for Go
- Added Codec fn for appending imports
- Added Go codec opts for mapping imports

Updates: #209


generator/internal/genclient/genclient.go
```diff
@@ -62,7 +62,7 @@ type LanguageCodec interface {
 	// variable, or enum value that holds it.
 	EnumValueName(e *EnumValue, state *APIState) string
 	// The fully qualified enum value name, as used when using the constant,
-	// variable, or enum value that hodls it.
```
coryan, line 65: I wonder why `typos` missed that 🤷

generator/internal/genclient/language/internal/golang/golang.go
```diff
@@ -25,22 +25,34 @@ import (
...
-		case "package-name-override":
+		switch {
+		case key == "package-name-override":
 			codec.PackageNameOverride = definition
-			continue
-		case "go-package-name":
+		case key == "go-package-name":
 			codec.GoPackageName = definition
-			continue
+		case strings.HasPrefix(key, "import-mapping"):
```
coryan, line 40: TIL.

generator/internal/genclient/model.go
```diff
@@ -78,6 +78,8 @@ type APIState struct {
 	MessageByID map[string]*Message
 	// EnumByID returns a message that is associated with the API.
 	EnumByID map[string]*Enum
+	// Package is the proto package name.
+	Package string
```
coryan, line 82: With OpenAPI specifications there is no package name.  For the Rust codec I chose to return either `api.Name` or (if set) `codec.PackageNameOverride`.  I think we should do something consistent across codecs for this.

Separately: I think APIState does not carry its weight.  We may want to refactor those fields in `API`.

generator/internal/genclient/templatedata.go
```diff
@@ -145,6 +149,10 @@ func (s *service) NameToCamel() string {
 	return s.c.ToCamel(s.s.Name)
 }
 
+func (s *service) Name() string {
+	return s.s.Name
+}
```
coryan, line 154: Should this go through the Codec to keep the right style?  All languages use `PascalCase` for these (and Protobuf too), so maybe we can postpone that change until we run into a language that insists on `snake_case` for classes too.  Okay, I answered my own question, leave as-is.

generator/templates/go/client.go.mustache
```diff
@@ -74,23 +77,23 @@ func (c *Client) token(ctx context.Context) (string, error) {
...
+func (c *Client) {{Name}}() *{{Name}}{
+    return &{{Name}}{client: c, baseURL: "https://{{DefaultHost}}"}
 }
 
 {{#Methods}}
 {{#DocLines}}
 // {{{.}}}
 {{/DocLines}}
-func (s *{{ServiceNameToCamel}}) {{NameToCamel}}(ctx context.Context, req *{{InputTypeName}}) (*{{OutputTypeName}}, error) {
+func (s *{{Name}}) {{NameToCamel}}(ctx context.Context, req *{{InputTypeName}}) (*{{OutputTypeName}}, error) {
```
coryan, line 96: Hmm... maybe `Name` is slightly confusing here, should that be `ServiceName` so we don't confuse it with the name of the method?

generator/internal/genclient/parser/protobuf/protobuf.go
```diff
@@ -197,6 +197,7 @@ func MakeAPI(serviceConfig *serviceconfig.Service, req *pluginpb.CodeGeneratorRe
 		ServiceByID: make(map[string]*genclient.Service),
 		MessageByID: make(map[string]*genclient.Message),
 		EnumByID:    make(map[string]*genclient.Enum),
+		Package:     req.GetSourceFileDescriptors()[0].GetPackage(),
```
coryan, line 200: Wouldn't this crash if there are no `GetSourceFileDescriptors()` ?  

And what if there are multiple package names?

coryan (approved):
The thing about consistency between Rust and Golang worries me.  If you think I shouldn't then go ahead, merge and we can discuss tomorrow or later in the week.  If you think I am right, maybe fix it and ask me for a second review.

codyoss:
@coryan I will fixup a couple things here and then wait to merge until we have time to talk about your concerns.

generator/internal/genclient/templatedata.go
codyoss: changed to service name to read better in template.

generator/internal/genclient/parser/protobuf/protobuf.go
codyoss: This is guaranteed to be set, assuming we sent in some protos. Added a comment to explain this.

generator/internal/genclient/model.go
codyoss: removed. Was not needed after all.

codyoss:
@coryan was able to factor out the controversial bit, will be merging.

coryan (approved):

PR #229, opened 2024-11-20
fix(generator/rust): handle sneaky field names

coryan:
Specially with OpenAPI specifications the field name may have gone
through a `snake_case` to `camelCase` translation. In Rust we want to
use `snake_case` for the fields, but that roundtrip is lossy.

I think the bad field name in Rust is undesirable, but it works. The bad
name is the serialized JSON does not work at all. This PR fixes that
problem.

Part of the work for #226

### DIFF
--- a/generator/testdata/rust/gclient/golden/secretmanager/src/model.rs
+++ b/generator/testdata/rust/gclient/golden/secretmanager/src/model.rs
@@ -866,6 +866,7 @@ pub struct SecretPayload {
     /// The CRC32C value is encoded as a Int64 for compatibility, and can be
     /// safely downconverted to uint32 in languages that support this type.
     /// https://cloud.google.com/apis/design/design_patterns#integer_types
+    #[serde(rename = "dataCrc32c")]
     #[serde_as(as = "Option<serde_with::DisplayFromStr>")]
     pub data_crc32c: Option<i64>,
 }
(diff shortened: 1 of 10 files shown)


generator/testdata/rust/openapi/golden/src/model.rs
```diff
@@ -2733,6 +2734,7 @@ pub struct GetIamPolicyRequest {
     /// To learn which resources support conditions in their IAM policies, see the
     /// [IAM
     /// documentation](https://cloud.google.com/iam/help/conditions/resource-policies).
+    #[serde(rename = "options.requestedPolicyVersion")]
```
codyoss, line 2737: This should be requested_policy_version, right?

generator/testdata/rust/gclient/golden/secretmanager/src/model.rs
```diff
@@ -864,6 +864,7 @@ pub struct SecretPayload {
     /// The CRC32C value is encoded as a Int64 for compatibility, and can be
     /// safely downconverted to uint32 in languages that support this type.
     /// https://cloud.google.com/apis/design/design_patterns#integer_types
+    #[serde(rename = "dataCrc32c")]
```
codyoss, line 867: This should be data_crc32c: https://github.com/googleapis/google-cloud-go/blob/47bb902e48695e5f0c46f7bee00d72ae5f794e1e/secretmanager/apiv1/secretmanagerpb/resources.pb.go#L973C97-L973C108

generator/testdata/rust/openapi/golden/src/model.rs
coryan: I am going to punt on this.  It is a query parameter, and not serialized at all.

codyoss (approved):

PR #231, opened 2024-11-20
feat(generator/rust): embed an auth cred

codyoss:
For now just embed an auth client so we can get a better feel about how we want to construct a client. This will change in the future.

A few traits needed to be removed, for now, because our auth Cred type does not implment them. We can fix this later if we care to.


Cargo.lock
```diff
@@ -736,6 +736,7 @@ version = "0.1.0"
 dependencies = [
  "bytes",
  "gax",
+ "google-cloud-auth",
```
coryan, line 739: Not for this PR: we should start using `gcp-sdk-auth` ASAP.

coryan (approved):

PR #236, opened 2024-11-20
test: expand secret manager integration tests

coryan:
Expand the integration tests using the client generated from the OpenAPI
specification. With this PR we can verify we can create, update, list,
get, and delete secrets (without locations). We can also get their IAM
policies, change the IAM policy, and test the IAM permissions.

For secret versions, verify we can create, enable, disable, access,
list, and delete them.

Finally verify we can list and get details about locations.

Part of the work for #226


codyoss (approved):
Nice!

Issue #139, opened 2024-11-08
Refactor repetitive code from the generated Rust client


coryan:
Every generated RPC repeats some code.  Other than one type (generics!) this:

https://github.com/googleapis/google-cloud-rust/blob/c9f74adf12932f910aa69c167a9ca5ec0fe258fd/generator/testdata/rust/gclient/golden/src/lib.rs#L70-L87

is basically identically to this:

https://github.com/googleapis/google-cloud-rust/blob/c9f74adf12932f910aa69c167a9ca5ec0fe258fd/generator/testdata/rust/gclient/golden/src/lib.rs#L100-L116

We should be able to refactor that code to a common private function.  In fact, we may be able to refactor that to the `gax` crate.

PR #241, opened 2024-11-20
refactor(generator/rust): factor out some common execute code

codyoss:
Had to create a `NoBody` type to pass along to satisfy generics. This type is not exported. 

Fixes: #139

### DIFF
--- a/generator/templates/rust/crate/src/lib.rs.mustache
+++ b/generator/templates/rust/crate/src/lib.rs.mustache
@@ -106,6 +106,10 @@ impl Client {
     }
     {{/Services}}
 }
+
+#[derive(serde::Serialize)]
+#[allow(dead_code)]
+struct NoBody {}
 {{/HasServices}}
 
 {{#Services}}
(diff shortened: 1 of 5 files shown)


codyoss:
@coryan The spell check is a false positive. Not sure how to get around that...

coryan:
This worked for me:

```
[type.mustache]
extend-glob = ["*.mustache"]

[type.mustache.extend-words]
ser = "ser"
```

generator/templates/rust/crate/src/lib.rs.mustache
```diff
@@ -106,6 +106,10 @@ impl Client {
     }
     {{/Services}}
 }
+
+#[derive(serde::Serialize)]
+#[allow(dead_code)]
+struct NoBody {}
```
coryan, line 112: If I was clever I would suggest a way to do this with the unit type...

generator/templates/rust/crate/src/lib.rs.mustache
```diff
@@ -137,23 +141,36 @@ impl {{NameToPascal}} {
...
-            .map_err(Error::io)?;
-        if !res.status().is_success() {
-            let status = res.status().as_u16();
-            let headers = gax::error::convert_headers(res.headers());
-            let body = res.bytes().await.map_err(Error::io)?;
+        self.execute(builder, {{#HasBody}}Some(req{{BodyAccessor}}){{/HasBody}}{{^HasBody}}None::<NoBody>{{/HasBody}}).await
+    }
+    {{/Methods}}
+
+    async fn execute<I: serde::ser::Serialize, O: serde::de::DeserializeOwned>(
```
coryan, line 148: For a future PR: how do we deal with services that have an `execute()` RPC?

coryan (approved):

generator/templates/rust/crate/src/lib.rs.mustache
codyoss: Yeah, I could not figure something like that out...

generator/templates/rust/crate/src/lib.rs.mustache
codyoss: ack, cross that when we get there. I am sure we can mangle or throw an underscore in front or something.

PR #250, opened 2024-11-21
ci: configure GCB to run integration tests

coryan:
This creates all the GCB (Google Cloud Build) configuration needed to
run the integration tests. There is a lot more Terraform than I would
like, but at least that can easily grow as we start needing more
resources.

The triggers for these builds are not created yet.

### DIFF
--- a/.gcb/builds/main.tf
+++ b/.gcb/builds/main.tf
@@ -1,0 +1,56 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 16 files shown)


.gcb/builds/main.tf
```diff
@@ -0,0 +1,56 @@
...
+
+# Create the service account needed for GCB and grant it the necessary
+# permissions.
+module "grants" {
+  source  = "./grants"
+  project = var.project
+}
+
+# Enable Cloud Build and create triggers.
+# module "triggers" {
```
codyoss, line 50: Is this coming in a future PR?

codyoss (approved):

Issue #235, opened 2024-11-20
Default `*Client` name for OpenAPI is not very ergonomic


coryan:
The factory function for secret manager comes out as `.google_cloud_secretmanager_v_1_secret_manager_service()` and the class as the `PascalCase` version of that. Yuck.

PR #251, opened 2024-11-21
refactor: have per-service constructors for Rust

codyoss:


src/integration-tests/src/secret_manager/openapi.rs
```diff
@@ -23,9 +23,7 @@ pub async fn run() -> Result<()> {
         .map(char::from)
         .collect();
 
-    let client = smo::Client::new()
-        .await?
-        .google_cloud_secretmanager_v_1_secret_manager_service();
+    let client = smo::GoogleCloudSecretmanagerV1SecretManagerServiceClient::new().await?;
```
coryan, line 26: No change needed in this PR, but I think #235 is really relevant.

coryan (approved):

src/integration-tests/src/secret_manager/openapi.rs
codyoss: Yeah, I think we need a codec option for this if we can't programmatically come up with a good name for openAPI clients. We have this in a couple languages today as a part of the per-language service config.

PR #259, opened 2024-11-21
ci(gcb): use pre-existing resources

coryan:
Creating project-wide grants in Terraform conflicts with the internal
Google Cloud project management tools. I moved the service account and
project-wide IAM permissions there, and use Terraform only for narrower
resources and permissions.


coryan:
The golang stuff will get fixed in #260

codyoss (approved):

PR #262, opened 2024-11-22
impl(generator): global configuration for refresh

### LINKED ISSUE #261: Support a repo-wide set of settings for all directories

coryan:
This change introduces support for an optional per-repository
configuration file, and starts using that file in the `refresh`
subcommand.

The motivation is to set the googleapis location in a single place and
to minimize the number of arguments that we need to provide in
`generate`.

Once the googleapis location is centralized, we can also start
downloading it from GitHub.

Fixes #261

### DIFF
--- a/generator/sidekick/config.go
+++ b/generator/sidekick/config.go
@@ -37,21 +37,18 @@ type GeneralConfig struct {
 }
 
 func LoadRootConfig(filename string) (*Config, error) {
-	var config Config
+	config := &Config{
+		Codec:  map[string]string{},
+		Source: map[string]string{},
+	}
 	if contents, err := os.ReadFile(filename); err == nil {
 		err = toml.Unmarshal(contents, &config)
 		if err != nil {
 			return nil, fmt.Errorf("error reading top-level configuration: %w", err)
 		}
 	}
-	if config.Codec == nil {
-		config.Codec = map[string]string{}
-	}
-	if config.Source == nil {
-		config.Source = map[string]string{}
-	}
 	// Ignore errors reading the top-level file.
-	return &config, nil
+	return config, nil
 }
 
 func MergeConfig(rootConfig *Config, filename string) (*Config, error) {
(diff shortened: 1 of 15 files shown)


generator/sidekick/config.go
```diff
@@ -14,6 +14,13 @@
...
 
+import (
+	"fmt"
+	"os"
+
+	toml "github.com/pelletier/go-toml/v2"
+)
+
 type Config struct {
 	General GeneralConfig `toml:"general"`
```
julieqiu, line 25: What's the difference between Config and General Config?

generator/sidekick/config.go
```diff
@@ -22,9 +29,77 @@ type Config struct {
...
+
+func LoadRootConfig(filename string) (*Config, error) {
+	var config Config
+	if contents, err := os.ReadFile(filename); err == nil {
+		err = toml.Unmarshal(contents, &config)
+		if err != nil {
+			return nil, fmt.Errorf("error reading top-level configuration: %w", err)
+		}
+	}
+	if config.Codec == nil {
```
julieqiu, line 47: nit: I'm not sure why these need to be initialized here as opposed to just leaving them nil, but if you need to I would do it on line 40

```
config := &Config{
   Codec: map[string]string{},
   Source: map[string]string{}
}
```

julieqiu (approved):

generator/sidekick/config.go
coryan: I am terrible at naming.

There are things that affect (or or at least most) parsers and codecs.  They go into `GeneralConfig`.  There are things that are only codec-specific (they go into the `codec` section).

Happy to rename all these things if you have better suggestions.

generator/sidekick/config.go
coryan: Moved the initialization. Also took the opportunity to add some tests.

Is it idiomatic to leave maps `nil`? And then one checks them before writing to them?

generator/sidekick/config.go
coryan: Also added some comments for "future us".

PR #264, opened 2024-11-22
impl(generator): helper functions for downloads

coryan:
To refresh all the directories we will need to download snapshots of
googleapis/googleapis from GitHub. This PR introduces a number of
helper functions to perform these downloads and keep a local cache in
$HOME.

The cache is content based, that is, the SHA256 of the complete tarball
is the index into the cache. The download fails if the contents do not
match the specified SHA256, and if the SHA256 is already in the cache
the functions skip the download.

Part of the work for #263

### DIFF
--- a/generator/sidekick/downloads_cache.go
+++ b/generator/sidekick/downloads_cache.go
@@ -23,6 +23,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/walle/targz"
 )
@@ -45,7 +46,7 @@ func makeGoogleapisRoot(rootConfig *Config) (string, error) {
 	if !ok {
 		return "", fmt.Errorf("using an https:// URL for googleapis-root requires setting googleapis-sha256")
 	}
-	cacheDir, err := downloadsCacheDir(rootConfig)
+	cacheDir, err := getCacheDir(rootConfig)
 	if err != nil {
 		return "", err
 	}
@@ -81,7 +82,12 @@ func downloadGoogleapisRoot(target, source, sha256 string) error {
 		return nil
 	}
 	var err error
-	for range 3 {
+	backoff := 10 * time.Second
+	for i := range 3 {
+		if i != 0 {
+			time.Sleep(backoff)
+			backoff = 2 * backoff
+		}
 		if err = downloadAttempt(target, source, sha256); err == nil {
 			return nil
 		}
@@ -150,7 +156,7 @@ func isDirectory(name string) bool {
 	return true
 }
 
-func downloadsCacheDir(rootConfig *Config) (string, error) {
+func getCacheDir(rootConfig *Config) (string, error) {
 	cacheDir, ok := rootConfig.Source["cachedir"]
 	if !ok {
 		var err error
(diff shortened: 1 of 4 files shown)


generator/sidekick/downloads_cache.go
```diff
@@ -0,0 +1,166 @@
...
+
+func downloadsCacheDir(rootConfig *Config) (string, error) {
+	cacheDir, ok := rootConfig.Source["cachedir"]
+	if !ok {
+		var err error
+		if cacheDir, err = os.UserCacheDir(); err != nil {
+			return "", err
+		}
+	}
+	return path.Join(cacheDir, "sidekick", "downloads"), nil
```
sai-sunder-s, line 161: should this path be added to the rootConfig now, in case it needs to be used again later?

generator/sidekick/downloads_cache.go
```diff
@@ -0,0 +1,166 @@
...
+	if err != nil {
+		return false
+	}
+	if !stat.IsDir() {
+		return false
+	}
+	return true
+}
+
+func downloadsCacheDir(rootConfig *Config) (string, error) {
```
sai-sunder-s, line 153: Rename to getCacheDir? Since this function is not doing the actual download?

generator/sidekick/downloads_cache.go
```diff
@@ -0,0 +1,166 @@
...
+	return "googleapis-" + filepath.Base(strings.TrimSuffix(googleapisRoot, ".tar.gz"))
+}
+
+func downloadGoogleapisRoot(target, source, sha256 string) error {
+	if fileExists(target) {
+		return nil
+	}
+	var err error
+	for range 3 {
+		if err = downloadAttempt(target, source, sha256); err == nil {
```
sai-sunder-s, line 85: should there be an exponential backoff here?

generator/sidekick/downloads_cache_test.go
```diff
@@ -0,0 +1,308 @@
...
+	server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.URL.Path != simulatedPath {
+			t.Errorf("Expected to request '%s', got: %s", simulatedPath, r.URL.Path)
+		}
+		w.WriteHeader(http.StatusOK)
+		w.Write(tarball.Contents)
+	}))
+	defer server.Close()
+
+	rootConfig := &Config{
```
sai-sunder-s, line 88: Just for my understanding, how is this rootConfig created? How is the sha256 available ahead?

generator/sidekick/downloads_cache.go
coryan: Good idea.

generator/sidekick/downloads_cache.go
coryan: Done.

generator/sidekick/downloads_cache.go
coryan: I think we should treat `rootConfig` as read-only.  It is going to be shared by multiple goroutines.  Also, the function is called just once. I think.  The code is factored out to make the call site more readable.

generator/sidekick/downloads_cache_test.go
coryan: Eventually we will have a `.sidekick.toml` file in the top-level directory of the monorepo. The URL and SHA256 is stored in that file.  At first we will update those manually, eventually `sidekick upgrade` will update both.

What other magic happens on those upgrades is up for debate.

coryan (commented):
PTAL

sai-sunder-s (approved):

PR #272, opened 2024-11-23
fix(generator/protobuf): trim comments only once

coryan:
For nested messages the protobuf parser was trimming excess whitespace
twice (or multiple times for deeply nested messages).

I created some additional tests and golden files in the process of
troubleshooting in the process of debugging this problem. They seem
useful to keep around.

I manually fixed `src/gax/src/error/rpc/generated/mod.rs`. Automatically
generating this file is a few PRs away, but we can clean it up now that
the generator is fixed.


generator/internal/genclient/language/internal/rust/rust_test.go
```diff
@@ -910,6 +910,32 @@ Maybe they wanted to show some JSON:
 		"/// }",
 		"/// ```",
 	}
+
+	c := &Codec{}
+	got := c.FormatDocComments(input)
+	if diff := cmp.Diff(want, got); len(diff) > 0 {
```
julieqiu, line 916: nit: this usually written this as `diff != ""`

julieqiu (approved):

generator/internal/genclient/language/internal/rust/rust_test.go
coryan: Fixed here and elsewhere in the PR. Will send a separate PR to clean up existing tests.

PR #290, opened 2024-11-26
refactor(generator): factor out default host to const

codyoss:


generator/internal/genclient/templatedata.go
```diff
@@ -79,6 +79,13 @@ func (t templateData) Imports() []string {
 	return t.c.Imports()
 }
 
+func (t templateData) DefaultHost() string {
```
coryan, line 82: Implicitly this means a single `API` can only have one `DefaultHost`.  I guess that must be true since we are also restricting these to a single service config yaml per API.  Do you mind adding a comment here explaining our reasoning? Future readers may not be aware.

coryan (approved):

generator/internal/genclient/templatedata.go
codyoss: done.

PR #292, opened 2024-11-26
refactor(generator): add helpers for processing services/methods

codyoss:
These will soon be used by mixin code as well.

### DIFF
--- a/generator/internal/genclient/model.go
+++ b/generator/internal/genclient/model.go
@@ -74,6 +74,8 @@ type API struct {
 type APIState struct {
 	// ServiceByID returns a service that is associated with the API.
 	ServiceByID map[string]*Service
+	// MethodByID returns a method that is associated with the API.
+	MethodByID map[string]*Method
 	// MessageByID returns a message that is associated with the API.
 	MessageByID map[string]*Message
 	// EnumByID returns a message that is associated with the API.
--- a/generator/internal/genclient/parser/openapi/openapi.go
+++ b/generator/internal/genclient/parser/openapi/openapi.go
@@ -166,7 +166,8 @@ func makeServices(api *genclient.API, model *libopenapi.DocumentModel[v3.Documen
 	if model.Model.Paths == nil {
 		return nil
 	}
-	methods, err := makeMethods(api, model, packageName)
+	sID := fmt.Sprintf(".%s.%s", packageName, serviceName)
+	methods, err := makeMethods(api, model, packageName, sID)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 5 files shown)


generator/internal/genclient/model.go
```diff
@@ -74,6 +74,8 @@ type API struct {
 type APIState struct {
 	// ServiceByID returns a service that is associated with the API.
 	ServiceByID map[string]*Service
+	// MethodByID returns a method that is associated with the API.
```
coryan, line 77: Unrelated, but since we are in that type: do we need `APIState`?  Seems like it is always tied to one `API` instance, and it is (like `API`) immutable once the parser finishes with it.

Should we just merge the two structs? (in a future PR)

generator/internal/genclient/parser/protobuf/protobuf.go
```diff
@@ -377,6 +365,34 @@ func normalizeTypes(state *genclient.APIState, in *descriptorpb.FieldDescriptorP
...
+		return nil
+	}
+	method := &genclient.Method{
+		ID:           mFQN,
+		PathInfo:     pathInfo,
+		Name:         m.GetName(),
+		InputTypeID:  m.GetInputType(),
+		OutputTypeID: m.GetOutputType(),
+	}
+	state.MethodByID[mFQN] = method
```
coryan, line 392: I would expect an analogous change in `openapi.go` and `openapi_test.go`.

coryan (changes requested):
I believe we should make the analogous changes to `openapi.go` at the same time.

generator/internal/genclient/parser/protobuf/protobuf.go
codyoss: done, openapi already had the ids in the test.

generator/internal/genclient/model.go
codyoss: In my mind the main reason this is seperate today is because it is what is needed to be passed into a lot of the codec functions. For that reason having a smaller type for lookups in nice vs passing in the full API everytime. Thoughts?

generator/internal/genclient/parser/openapi/openapi.go
```diff
@@ -247,6 +248,7 @@ func makeMethods(api *genclient.API, model *libopenapi.DocumentModel[v3.Document
 				OutputTypeID:  responseMessage.ID,
 				PathInfo:      pathInfo,
 			}
+			api.State.MethodByID[m.ID] = m
```
coryan, line 251: `m.ID` does not have the same format as the Protobuf methods. Those include the service ID too.  Probably does not matter (maybe for documentation links? not sure).

generator/internal/genclient/model.go
coryan: You mean it has a narrower interface?  You are right, that is a Good Thing :tm:   But that argues for passing an interface to the codec methods, instead of an specific implementation.

We should not do design via review comments.  I think we have enough for a future discussion.

coryan (approved):

generator/internal/genclient/parser/openapi/openapi.go
codyoss: made them align, thanks

Issue #284, opened 2024-11-25
Use generated code in `src/generated/cloud/secretmanager/v1/`


coryan:
The code does not need to be perfect or ready for release, just needs to be automatically generated and easy to refresh.

- [ ] Use generated code in `src/generated/type`
- [ ] Use generated code in `src/generated/iam/v1`
- [ ] Use generated code in `src/generated/cloud/location`

PR #298, opened 2024-11-27
feat(generator): ignore some packages

coryan:
Most generated packages will use `gax`, `wkt`, and `auth`. It makes sense to
configure these packages in the top-level `.sidekick.toml` file. But we also
need to remove `gax` from that list when generating
`src/gax/.../generated/mod.rs`. So far, the `.sidekick.toml` configurations are
additive, or at most "replace the previous value", but now we want to *remove*
a value.

With this change, one can configure a package to be "ignored". The top-level
`.sidekick.toml` file will list the most common packages, and we will ignore
these in the few places where it makes sense.

That makes the generation of most libraries relatively ergonomic.

Part of the work for #284

### DIFF
--- a/generator/internal/genclient/language/internal/rust/rust.go
+++ b/generator/internal/genclient/language/internal/rust/rust.go
@@ -46,7 +46,7 @@ func NewCodec(copts *genclient.CodecOptions) (*Codec, error) {
 		case "generate-module":
 			value, err := strconv.ParseBool(definition)
 			if err != nil {
-				return nil, fmt.Errorf("cannot convert `generate-module` value (%s) to boolean: %w", definition, err)
+				return nil, fmt.Errorf("cannot convert `generate-module` value %q to boolean: %w", definition, err)
 			}
 			codec.GenerateModule = value
 			continue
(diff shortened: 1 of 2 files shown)


generator/internal/genclient/language/internal/rust/rust.go
```diff
@@ -86,11 +86,17 @@ func NewCodec(copts *genclient.CodecOptions) (*Codec, error) {
 				specificationPackages = append(specificationPackages, s[1])
 			case "feature":
 				pkg.Features = append(pkg.Features, strings.Split(s[1], ",")...)
+			case "ignore":
+				value, err := strconv.ParseBool(s[1])
+				if err != nil {
+					return nil, fmt.Errorf("cannot convert `ignore` value in %s to boolean: %w", definition, err)
```
codyoss, line 92: nit: in other error message you would have written `value (%s) to boolean`. Also maybe the parenthesis are fine but you could also do `%q` for a quoted string

codyoss (approved):

generator/internal/genclient/language/internal/rust/rust.go
coryan: Done.

PR #313, opened 2024-11-27
ci: verify generated files are in sync

coryan:
Introduce a new build that generates all the clients and detects any
differences between the commited and generated code.

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -57,3 +57,47 @@ jobs:
       - run: cargo clippy -- --deny warnings
       - run: cargo fmt
       - run: git diff --exit-code
+  regenerate:
+    # Verifies the generated code has not been tampered with. Or maybe that the
+    # code requires no tampering.
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        rust-version: ['1.81']
+        go-version: ['1.23']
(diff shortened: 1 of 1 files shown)


.github/workflows/sdk.yaml
```diff
@@ -57,3 +57,47 @@ jobs:
...
+          cache-dependency-path: |
+            generator/go.sum
+      - name: Install protoc
+        run: |
+          set -e
+          curl -fSSL -o /tmp/protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v28.3/protoc-28.3-linux-x86_64.zip
+          cd /usr/local
+          sudo unzip -x /tmp/protoc.zip
+          protoc --version
+      - name: Install goimports
```
julieqiu, line 94: It does't look like you run `goimports` anywhere. Is this being used somewhere?

julieqiu (approved):

.github/workflows/sdk.yaml
coryan: You are right, it is used in `sidekick_test.go` but we are not running that here. Removed.

https://github.com/googleapis/google-cloud-rust/blob/3aa85ba410fd6888effd693f222cfcf047241b42/generator/sidekick/sidekick_test.go#L270

PR #331, opened 2024-11-29
impl(internal/language): merge rust and golang into one package

julieqiu:
Merge rust and golang to a single language package. This reduces nesting and also makes it easier to share code between these packages in future PRs.


coryan (approved):
LGTM.  We can always break this up again if it turns out ${futureLang} requires a separate package or something.

PR #332, opened 2024-11-29
impl(internal/language): move NewTestAPI

julieqiu:
Now that golang and rust are in the same package, move genclient.NewTestAPI to a `_test.go` file in package language and unexport it, since that is the open place this function is used.


coryan (approved):
Nice.

PR #341, opened 2024-11-30
impl(internal): split up package genclient

julieqiu:
Refactor the genclient package by splitting it into other existing packages. It mainly contains helper functions that make more sense to live in those packages.


generator/internal/parser/common.go
coryan, line 1: nit: maybe this file should be called  `parser/parser.go` to match `language/language.go`.  Could be done in a future PR.

coryan (approved):

PR #344, opened 2024-11-30
impl(internal/sidekick):  construct TemplateData using values instead of methods

julieqiu:
The TemplateData struct serves primarily as a data container for passing information to Mustache templates. Since the values it provides are either static or derived from other data, and do not rely on complex logic tied to the struct's internal state, convert these methods into fields to make that purpose more explicit.


generator/internal/sidekick/templatedata.go
```diff
@@ -481,3 +246,52 @@ func mapSlice[T, R any](s []T, f func(T) R) []R {
...
+		NameToSnakeNoMangling: c.ToSnakeNoMangling(oneOf.Name),
+		FieldType:             c.OneOfType(oneOf, state),
+		DocLines:              c.FormatDocComments(oneOf.Documentation),
+		Fields: mapSlice(oneOf.Fields, func(field *api.Field) *Field {
+			return newField(field, c, state)
+		}),
+	}
+}
+
+// Constructor function for c.Field
```
coryan, line 263: Why comments in some of these functions but not the others?

coryan (approved):
I must confess I do not understand the motivation here, but SGTM.  If you want a reviewer with ~more~ better opinions, maybe wait for @codyoss ?

generator/internal/sidekick/templatedata.go
julieqiu: stray comments, removed.

coryan (approved):

codyoss (approved):

Issue #92, opened 2024-11-05
Protect all code blocks with `norust`


coryan:
We need to parse the CommonMark comments and protect all code blocks with `norust`.

## Background

The Protobuf and OpenAPI v3 documentation comments are in Markdown format, they are supposed to be ([aip.dev/192](https://google.aip.dev/192)) in [CommonMark](https://commonmark.org/) format, though sometimes they deviate from the AIP.

The generator converts these Protobuf comments into Rustdoc comments.

Rust tries to compile all code blocks unless protected by a `norust` mark. 

Block quotes appear in many [different forms](https://spec.commonmark.org/0.31.2/#block-quotes) in the CommonMark spec.

In #57 I fixed fenced code blocks (that is, those protected by backticks `` ``` ``), then I thought of other ways to specify code quotes (indented multiple spaces).

## Overview

We need to parse the CommonMark comments, apply changes in the syntax tree, and re-emit Markdown.  Probably in the Codec's `FormatDocComments()` function.  As a starting point, consider this Golang library: https://github.com/yuin/goldmark

## Alternatives Considered

We could try to perform this cleanup using heuristics. Shudder. 😨

coryan:
I think the hacks in #346 are enough to make this issue non-blocking for a release.

coryan:
There is a code block where we could do better:

https://docs.rs/gcp-sdk-type/0.1.0-rc2/gcp_sdk_type/model/struct.PhoneNumber.html

mpeddada1:
@coryan Yup! I think this can be closed now that https://github.com/googleapis/google-cloud-rust/pull/693 is merged.

PR #346, opened 2024-11-30
impl(generator): better blockquote handling

coryan:
Rust treats any blockquote as a Rust code sample, unless it has the
`norust` tag. Furthermore, `cargo test` will try to compile these code
samples. This is all very good, but we generate the documentation
comments from the Probobuf (or OpenAPI) comments, which may and do
include blockquotes that are not Rust code.

This change handles a few more blockquote types. We really need to parse
the Markdown comments to handle all possible cases, but this can handle
the most urgent cases.

Arguably this is partial progress on #92

### DIFF
--- a/generator/testdata/rust/gclient/golden/iam/v1/src/model.rs
+++ b/generator/testdata/rust/gclient/golden/iam/v1/src/model.rs
@@ -437,41 +437,43 @@ impl Binding {
 ///
 /// Example Policy with multiple AuditConfigs:
 ///
+/// ```norust
+/// {
+///   "audit_configs": [
 ///     {
-///       "audit_configs": [
+///       "service": "allServices",
+///       "audit_log_configs": [
 ///         {
-///           "service": "allServices",
-///           "audit_log_configs": [
-///             {
-///               "log_type": "DATA_READ",
-///               "exempted_members": [
-///                 "user:jose@example.com"
-///               ]
-///             },
-///             {
-///               "log_type": "DATA_WRITE"
-///             },
-///             {
-///               "log_type": "ADMIN_READ"
-///             }
+///           "log_type": "DATA_READ",
+///           "exempted_members": [
+///             "user:jose@example.com"
 ///           ]
 ///         },
 ///         {
-///           "service": "sampleservice.googleapis.com",
-///           "audit_log_configs": [
-///             {
-///               "log_type": "DATA_READ"
-///             },
-///             {
-///               "log_type": "DATA_WRITE",
-///               "exempted_members": [
-///                 "user:aliya@example.com"
-///               ]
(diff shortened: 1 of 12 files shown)


src/generated/type/src/model.rs
```diff
@@ -458,17 +482,27 @@ pub struct Decimal {
     ///
     /// The ENBF grammar is:
     ///
-    ///     DecimalString =
-    ///       [Sign] Significand [Exponent];
+    /// ```norust
+    /// DecimalString =
+    ///   [Sign] Significand [Exponent];
+    /// ```
```
coryan, line 488: I think my hack is working correctly in this case, creating multiple blockquotes:

https://spec.commonmark.org/0.31.2/#example-242

However, this may not have been the intent upstream. We may want to fix the upstream code to use explicit `> ` markers.

src/generated/type/src/model.rs
```diff
@@ -763,11 +805,17 @@ impl Money {
 /// For instance, in Java this would be:
 ///
 ///    com.google.type.PhoneNumber wireProto =
-///        com.google.type.PhoneNumber.newBuilder().build();
+/// ```norust
```
coryan, line 808: I think my hack is not doing well in this case.  Blockquotes require a blank line before they start and my hack does not require this. The code upstream is only indented 3 spaces, so about 1/2 of the lines appear to be blockquotes.

I am going to let it stand.  The true fix is to use a Markdown parser, as suggested in #92.

generator/testdata/rust/gclient/golden/iam/v1/src/model.rs
```diff
@@ -437,41 +437,43 @@ impl Binding {
 ///
 /// Example Policy with multiple AuditConfigs:
 ///
+/// ```norust
```
coryan, line 440: These diffs are much easier to grok if you turn off whitespace differences.

src/gax/src/error/rpc/generated/mod.rs
```diff
@@ -335,21 +339,27 @@ pub mod bad_request {
...
-        ///         }
+        /// ```norust
+        /// message CreateContactRequest {
+        ///   message EmailAddress {
+        ///     enum Type {
+        ///       TYPE_UNSPECIFIED = 0;
+        ///       HOME = 1;
+        ///       WORK = 2;
+        ///     }
+        /// ```
```
coryan, line 350: As noted elsewhere, ideally these two blockquotes would be merged, but the code (maybe by accident) is doing the right thing:

https://spec.commonmark.org/0.31.2/#example-242

The comments upstream (in the protos) need fixing.

julieqiu (approved):

PR #349, opened 2024-12-02
impl(internal/api): merge packages api and parser

julieqiu:
Merge packages api and parser into a single package, since the purpose of the package parser is to provide constructor functions for the `API` model in the api package.


coryan (approved):
I see your point about the parsers being complex constructors in disguise.

I think the separation between "these are the components that parse text files (.proto, .json) into the AST" vs. "these are the components that define the AST" is useful. It may be even more useful as we add more parsers (e.g. apiary discovery docs).

I don't have a strong enough opinion though.

Please think it over for a minute and if you still like this merge then go for it.

julieqiu:
> I see your point about the parsers being complex constructors in disguise.
> 
> I think the separation between "these are the components that parse text files (.proto, .json) into the AST" vs. "these are the components that define the AST" is useful. It may be even more useful as we add more parsers (e.g. apiary discovery docs).
> 
> I don't have a strong enough opinion though.
> 
> Please think it over for a minute and if you still like this merge then go for it.

I also don't have a strong opinion here and can see it either way. Will abandon this change for now, and we can merge it in the future if useful.

Issue #297, opened 2024-11-27
automatic pagination as defined by AIP-4233


codyoss:
- [x] Create a Stream adapter
- [x] Support iterating over items
- [x] Test if we need to declare Send/Sync bounds
- [x] Integrate into generated clients, behind a feature flag.

See: https://google.aip.dev/client-libraries/4233
and
See: https://google.aip.dev/158

PR #352, opened 2024-12-02
feat(gax): add Stream auto pagination

codyoss:
This addes a new Paginator type that can be exposed by List RPCs for auto paginating over async streams. This feature is not yet stable in rust so we should have users opt-in to using this unstable feature. When Streams are added to the stdlib we will likely want to make a breaking change to use the traits defined there.

This currently only auto-paginates over pages. We will need to expand the trait of PageableResponse to also describe items to iterate over to paginate over those. But this seemed like enough complexity for one PR.

Updates: #297

### DIFF
--- a/src/gax/src/lib.rs
+++ b/src/gax/src/lib.rs
@@ -62,3 +62,8 @@ mod request_parameter;
 
 /// The core error types used by generated clients.
 pub mod error;
+
+/// Defines some types and traits to convert and use List RPCs as a Stream.
+/// Async streams are not yet stable, so neither is the use of this feature.
+#[cfg(feature = "stream")]
+pub mod paginator;
--- a/src/gax/src/paginator.rs
+++ b/src/gax/src/paginator.rs
@@ -1,0 +1,228 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use futures::stream::unfold;
+use futures::{Stream, StreamExt};
+use pin_project::pin_project;
+use std::future::Future;
+use std::pin::Pin;
+
+/// Describes a type that can be iterated over asyncly when used with [Paginator].
+pub trait PageableResponse {
+    fn next_page_token(&self) -> String;
+}
+
+/// An adapter that converts list RPCs as defined by [AIP-4233](https://google.aip.dev/client-libraries/4233)
+/// into a [futures::Stream] that can be iterated over in an async fashion.
+#[pin_project]
+pub struct Paginator<T, E> {
+    #[pin]
+    stream: Pin<Box<dyn Stream<Item = Result<T, E>>>>,
+}
+
+impl<T, E> Paginator<T, E>
+where
+    T: PageableResponse,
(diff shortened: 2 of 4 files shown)


src/gax/src/paginator.rs
```diff
@@ -0,0 +1,228 @@
...
+
+impl<T, E> Paginator<T, E>
+where
+    T: PageableResponse,
+{
+    /// Creates a new [Paginator] given the initial page token and a function
+    /// to fetch the next [PageableResponse].
+    pub fn new<F>(seed_token: String, execute: impl Fn(String) -> F + Clone + 'static) -> Self
+    where
+        // TODO(codyoss): This, and other types, may also need send here in practice.
```
coryan, line 42: Please convert this to a GitHub issue. 

Can we test if this requires `Send` and/or `Sync`?

src/gax/src/lib.rs
```diff
@@ -62,3 +62,8 @@ mod request_parameter;
 
 /// The core error types used by generated clients.
 pub mod error;
+
+/// Defines some types and traits to convert and use List RPCs as a Stream.
+/// Async streams are not yet stable, so neither is the use of this feature.
+#[cfg(feature = "stream")]
```
coryan, line 68: Will this be a feature in every generated library then?

src/gax/src/paginator.rs
```diff
@@ -0,0 +1,228 @@
...
+        }
+    }
+
+    /// Returns the next mutation of the wrapped stream.
+    pub fn next(&mut self) -> futures::stream::Next<'_, Self> {
+        StreamExt::next(self)
+    }
+}
+
+enum IterState {
```
coryan, line 78: This looks like a [std::ops::ControlFlow<String,()>](https://doc.rust-lang.org/beta/std/ops/enum.ControlFlow.html) Should we use that instead?

src/gax/src/paginator.rs
```diff
@@ -0,0 +1,228 @@
...
+{
+    /// Creates a new [Paginator] given the initial page token and a function
+    /// to fetch the next [PageableResponse].
+    pub fn new<F>(seed_token: String, execute: impl Fn(String) -> F + Clone + 'static) -> Self
+    where
+        // TODO(codyoss): This, and other types, may also need send here in practice.
+        F: Future<Output = Result<T, E>> + 'static,
+    {
+        let stream = unfold(IterState::Token(seed_token), move |state| {
+            let execute = execute.clone();
```
coryan, line 46: Why do we need to clone this? Feels like we should be able to borrow the function to call it, that is why it is not a `FnOnce`, right?

src/gax/src/paginator.rs
```diff
@@ -0,0 +1,228 @@
...
+    }
+
+    struct InnerClient {
+        data: Arc<Mutex<Vec<TestResponse>>>,
+    }
+
+    impl Client {
+        fn build_outgoing_list_request(&self) {
+            // placeholder for building an http request
+        }
```
coryan, line 132: I am not sure this adds enough value for a test.

src/gax/src/paginator.rs
```diff
@@ -0,0 +1,228 @@
...
+            },
+        ];
+
+        let client = Client {
+            inner: Arc::new(InnerClient {
+                data: Arc::new(Mutex::new(responses)),
+            }),
+        };
+        let mut resps = vec![];
+        let mut stream = client.list_rpc_stream(TestRequest::default());
```
coryan, line 197: Seems like you are missing a test where the initial token is not the empty string.

coryan (approved):
Some minor stuff.  I have questions, but they are really for my education, not intended to block you.

src/gax/src/paginator.rs
codyoss: I believe this is needed because Fn does not implement copy and we move this this function  to the async block below for the call. There may be a better way to call this, but this is where I ended up after a well fought battle with the compiler. If you see away around this I am open, just could not figure out anything better.

PR #355, opened 2024-12-02
feat(generator/rust): generate client trait

coryan:
We want to generate a trait for the clients. Eventually we will have
more than one implementation of this trait (e.g. logging, tracing,
retry), and customers will be able to mock the clients.

Part of the work for #222


generator/testdata/rust/gclient/golden/iam/v1/src/client.rs
```diff
@@ -0,0 +1,89 @@
...
+/// attached.
+///
+/// # Notes
+///
+/// Application developers may use this trait to mock the  clients.
+///
+/// Services gain new RPCs routinely. Consequently, this trait gains new methods
+/// too. To avoid breaking applications the trait provides a default
+/// implementation for each method. These implementations return an error.
+#[async_trait::async_trait]
```
codyoss, line 53: Is this still needed for our use case? See: https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#should-i-still-use-the-async_trait-macro

Should we be using the trait_variant?

codyoss: If we do need to use this, I wonder if this should be pub(crate) by default for our own needs and expose it with a feature for clients?

src/generated/type/src/client.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2024 Google LLC
```
codyoss, line 1: This should likely not exist?

generator/testdata/rust/gclient/golden/iam/v1/src/lib.rs
```diff
@@ -21,6 +21,8 @@ use gax::error::{Error, HttpError};
 use google_cloud_auth::{Credential, CredentialConfig};
 use std::sync::Arc;
 
+pub mod client;
```
codyoss, line 24: Is the client implementing the trait going to be in a separate PR?

coryan: yes, but I can make it a single PR if you think that is easier to grok.

src/generated/type/src/client.rs
coryan: Yes, but all `.rs.mustache` files are generated unconditionally.  I could not think of a simple trick to avoid the code generation.  Maybe the Codec should return a list of input templates -> output file names?

generator/testdata/rust/gclient/golden/iam/v1/src/client.rs
coryan: > Is this still needed for our use case? See: https://blog.rust-lang.org/2023/12/21/async-fn-rpit-in-traits.html#should-i-still-use-the-async_trait-macro
> 
> Should we be using the trait_variant?

We are building against 1.70 (that was based on Tokio's MSRV), and `trait_variant` requires 1.75:

https://crates.io/crates/trait-variant

note that it is also 0.1.x

We can change the MSRV, or switch to using explicit futures.  The latter may be a more stable approach, and since it is generated code, the robots won't complain about extra typing.

src/generated/type/src/client.rs
codyoss: Fair, maybe just open an issue for now.

codyoss (approved):

generator/testdata/rust/gclient/golden/iam/v1/src/client.rs
codyoss: oops, I did not notice 1.70 for now. ack

src/generated/type/src/client.rs
coryan: Tracking in #357

generator/testdata/rust/gclient/golden/iam/v1/src/client.rs
codyoss: > https://crates.io/crates/trait-variant
>
> note that it is also 0.1.x

True, but at least the language team owns this which gives it more cred in my mind.

generator/testdata/rust/gclient/golden/iam/v1/src/lib.rs
coryan: See #356 if you are interested.

PR #364, opened 2024-12-05
feat: Introducing Credential Trait and AuthError

sai-sunder-s:
Create a credential Trait that can be used by the old and new auth libs.


coryan (changes requested):

src/auth/src/token.rs
```diff
@@ -0,0 +1,23 @@
...
+// limitations under the License.
+
+use std::collections::HashMap;
+
+#[derive(Debug)]
+pub struct Token {
+    pub token: String,
+    pub token_type: String,
+    pub expires_in: Option<i64>,
+    pub metadata: Option<HashMap<String, String>>,
```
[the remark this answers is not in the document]
sai-sunder-s: This is to hold miscellaneous things that are present in a token response. For example, OAuth2's token endpoint returns list of scopes that the user granted (different from the requested scopes). A calling application may want to see this info to control their application flow.

src/auth/src/credentials.rs
```diff
@@ -0,0 +1,27 @@
...
+use std::collections::HashMap;
+
+use async_trait::async_trait;
+
+pub type Result<T> = std::result::Result<T, crate::errors::AuthError>;
+
+#[async_trait]
+pub trait Credential: Send + Sync {
+    async fn get_token(&mut self) -> Result<crate::token::Token>;
+    async fn get_headers(&mut self) -> Result<HashMap<String, String>>;
```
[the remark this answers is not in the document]
sai-sunder-s: A vector sounds good. We will have non-standard header names, so Vec seems easier.

src/auth/src/errors.rs
```diff
@@ -12,57 +12,58 @@
...
     }
 
+    /// Returns `true` if the error is retryable; otherwise returns `false`.
     pub fn is_retryable(&self) -> bool {
         self.is_retryable
     }
 }
 
-impl std::error::Error for AuthError {}
+impl Error for AuthError {}
```
coryan, line 55: If we bother to capture the source error, we should make it available via the standard trait?

```suggestion
impl Error for AuthError {
  fn source(&self) -> Option<Box<dyn Error + 'static>> {
    Some(self.source)
  }
}
```

src/auth/src/errors.rs
```diff
@@ -12,57 +12,58 @@
...
 
-impl std::fmt::Display for AuthError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+impl Display for AuthError {
+    /// Formats the error message to include retryability and source.
+    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
         write!(f, "Retryable:{}, Source:{}", self.is_retryable, self.source)
     }
 }
```
coryan, line 62: This looks more like the Debug trait than the Display trait... Consider:

```suggestion
impl Display for AuthError {
    /// Formats the error message to include retryability and source.
    fn fmt(&self, f: &mut Formatter<'_>) -> Result {
       let msg = if self.is_retryable { "but future attempts may succeed" } else { "and future attempts will not succeed" };
        write!(f, "cannot create access token, {}, source:{}", msg, self.source)
    }
}
```

src/auth/src/token.rs
```diff
@@ -12,12 +12,20 @@
...
     pub token: String,
+
+    /// The type of the token.  Common types include "Bearer".
     pub token_type: String,
-    pub expires_in: Option<i64>,
-    pub metadata: Option<HashMap<String, String>>,
+
+    /// The instant at which the token expires. If `None`, the token does not
+    /// expire (or its expiration is unknown).
+    pub expires_in: Option<std::time::Instant>,
```
coryan, line 26:
```suggestion
    pub expires_at: Option<std::time::Instant>,
```

src/auth/src/credentials.rs
```diff
@@ -12,16 +12,31 @@
...
-#[async_trait]
+/// Represents an auth credential used to obtain auth tokens.
+/// Implementors of this trait provide a way to asynchronously retrieve tokens
+/// and construct auth headers.
 pub trait Credential: Send + Sync {
-    async fn get_token(&mut self) -> Result<crate::token::Token>;
-    async fn get_headers(&mut self) -> Result<HashMap<String, String>>;
-    fn get_quota_project_id(&self) -> Result<String>;
-    fn get_universe_domain(&self) -> Result<String>;
+    /// Asynchronously retrieves a token.
```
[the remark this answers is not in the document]
coryan: Maybe define the token once at the trait definition?

src/auth/src/credentials.rs
```diff
@@ -12,16 +12,31 @@
...
+    ///
+    /// This function returns a `Future` that resolves to a `Result` containing
+    /// either the `Token` or an `AuthError` if an error occurred during
```
coryan, line 26: I think when referencing `Token` you want to say `[crate::token::Token]` so rustdoc will cross-reference the thing. Ditto for the error.  You may be able to just say `[Token]`, rustdoc may be smart and understand what thing you are referring to, try it out, I am not sure.

coryan (changes requested):
Let me know if these reviews are getting nit picky, we can open issues to improve things later.

Please write the documentation thinking about folks that are not experts in authentication and authorization.

src/auth/src/credentials.rs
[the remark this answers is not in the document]
sai-sunder-s: Sorry misinterpreted your 2nd comment. Changed to Vector of HeaderName, HeaderValue.

It maybe easier for you if I create as a headermap looks like? You can call .extend to combine the headermap you have on the client lib side with the auth headermap. WDYT?

sai-sunder-s:
> Let me know if these reviews are getting nit picky, we can open issues to improve things later.
> 
> Please write the documentation thinking about folks that are not experts in authentication and authorization.

I think we can address the documentation related comments later. Let us focus on the basic framework first. Created https://github.com/googleapis/google-cloud-rust/issues/365 to track this. Will leave the doc related ones open for now.

I am happy to work through all the other comments!

src/auth/src/credentials.rs
```diff
@@ -12,16 +12,31 @@
...
+    /// token retrieval.
+    fn get_token(&mut self) -> impl Future<Output = Result<crate::token::Token>> + Send;
+
+    /// Asynchronously retrieves auth headers.
+    ///
+    /// This function returns a `Future` that resolves to a `Result` containing
+    /// either a vector of key-value pairs representing the headers or an
+    /// `AuthError` if an error occurred during header construction.
+    fn get_headers(&mut self) -> impl Future<Output = Result<Vec<(String, String)>>> + Send;
```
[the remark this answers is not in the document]
sai-sunder-s: Fixed. Sorry misinterpreted the comment. Yes we should definitely use HeaderValue for auth headers.

coryan (approved):

src/auth/src/credentials.rs
```diff
@@ -0,0 +1,45 @@
...
+    /// either the `Token` or an `AuthError` if an error occurred during
+    /// token retrieval.
+    fn get_token(&mut self) -> impl Future<Output = Result<crate::token::Token>> + Send;
+
+    /// Asynchronously retrieves auth headers.
+    ///
+    /// This function returns a `Future` that resolves to a `Result` containing
+    /// either a vector of key-value pairs representing the headers or an
+    /// `AuthError` if an error occurred during header construction.
+    fn get_headers(
```
codyoss, line 36: nit: this seems associated with a token rather than the credential itself right?

src/auth/src/credentials.rs
```diff
@@ -0,0 +1,45 @@
...
+        &mut self,
+    ) -> impl Future<Output = Result<Vec<(HeaderName, HeaderValue)>>> + Send;
+
+    /// Retrieves the quota project ID associated with the credential, if any
+    fn get_quota_project_id(&self) -> Option<String>;
```
codyoss, line 41: I don't think there is harm in just making all of these functions futures today. You never know when we may start reading a file or make a network request to MDS to fetch this value in a lazy manner that forces the use of futures.

src/auth/src/credentials.rs
sai-sunder-s: It is broader than just token. Right now possible headers are:
- Authorization
- x-goog-api-key
- x-goog-user-project
- x-goog-api-client
- google-byoid-sdk

Future header:
x-allowed-locations

src/auth/src/credentials.rs
sai-sunder-s: Good point. Done.

src/auth/src/errors.rs
```diff
@@ -12,57 +12,58 @@
...
-type BoxError = Box<dyn std::error::Error + Send + Sync>;
+use std::error::Error;
+use std::fmt::{Display, Formatter, Result};
 
+type BoxError = Box<dyn Error + Send + Sync>;
+
+/// Represents an auth error.  This error type indicates issues
+/// encountered during the auth process.
 #[derive(Debug)]
 pub struct AuthError {
```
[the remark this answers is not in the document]
sai-sunder-s: @coryan

src/auth/src/credentials.rs
sai-sunder-s: On second thought, I cannot think of a use case where someone want to know the quota project or token alone. The `get_headers()` should suffice for the client library to consume auth lib I believe. 

We still need `get_universe_domain()` for the client lib to determine which universe to use to build the urls.

So I am removing `get_token` and `get_quota_project` for now. We can add them later if we find a need for it.

LMK if either of you think client lib will need those 2 methods. @codyoss and @coryan

coryan (changes requested):

src/auth/src/credentials.rs
codyoss: > is that what you were thinking about?

Yes, I think these can be consolidated.

src/auth/src/token.rs
```diff
@@ -0,0 +1,31 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Represents an auth token.
+#[derive(Debug)]
+pub struct Token {
+    /// The actual token string.  This is the value used in Authorization header.
```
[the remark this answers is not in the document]
sai-sunder-s: Access/ID tokens always use "Authorization" header right, irrespective of platform? (even though ID token is not really authz)

src/auth/src/credentials.rs
sai-sunder-s: Yes, I had same thought. The comment I put this in got outdated. Here it is

> On second thought, I cannot think of a use case where someone want to know the quota project or token alone. The get_headers() should suffice for the client library to consume auth lib I believe.
> 
> We still need get_universe_domain() for the client lib to determine which universe to use to build the urls.
> 
> So I am removing get_token and get_quota_project for now. We can add them later if we find a need for it.
> 
> LMK if either of you think client lib will need those 2 methods. @codyoss and @coryan

src/auth/src/credentials.rs
coryan: > On second thought, I cannot think of a use case where someone want to know the quota project or token alone.

gRPC for C++ required the access token (and nothing else) for per-call credentials.

I think we can live with `get_headers()` for now, but caching may be harder for you 🤷

src/auth/src/token.rs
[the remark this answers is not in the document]
coryan: I think gRPC uses something else for per-call credentials.

src/auth/src/token.rs
```diff
@@ -23,7 +23,7 @@ pub struct Token {
 
     /// The instant at which the token expires. If `None`, the token does not
     /// expire (or its expiration is unknown).
-    pub expires_at: Option<std::time::Instant>,
+    pub expires_at: Option<time::OffsetDateTime>,
```
coryan, line 26: And then I noticed that the time crate is `0.3.*` and therefore not suitable for a public interface.  Feel free to merge as-is, just open a tracking bug.

@codyoss any thoughts?

coryan (approved):

src/auth/src/credentials.rs
sai-sunder-s: grpc doesn't need quota project?

Not sure I understand why caching is difficult for `get_headers()`

src/auth/src/token.rs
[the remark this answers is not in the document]
sai-sunder-s: I see. that means we would still need `get_token()` so that grpc lib can add whatever header name it wants.
But let us just keep `get_headers()` for now as being discussed in the other comment. We can revisit adding `get_token()` when needed.

src/auth/src/credentials.rs
coryan: Because it does not have an expiration time, so one does not know when to refresh?

sai-sunder-s: When building the headers, we will get the token from the token provider and the token provider will internally kick-off an async refresh if the token is about to expire?

Caller doesn't have to trigger a refresh.

src/auth/src/token.rs
sai-sunder-s: Filed https://github.com/googleapis/google-cloud-rust/issues/387

src/auth/src/credentials.rs
sai-sunder-s: @coryan

coryan: I see. SGTM. :shipit:

PR #366, opened 2024-12-06
refactor(generator): make execute and associated fn

codyoss:
In preparation for passing function states around for iterators our execution of RPCs needs to be completely owned and not have references to `self`. This means passing in all the state that is needed to the execute function.

Updates: #297

### DIFF
--- a/generator/testdata/rust/gclient/golden/iam/v1/src/transport.rs
+++ b/generator/testdata/rust/gclient/golden/iam/v1/src/transport.rs
@@ -85,20 +85,22 @@ impl Iampolicy {
         })
     }
 
+    async fn fetch_token(&self) -> Result<String> {
+        let tok = self
+            .inner
+            .cred
+            .access_token()
+            .await
+            .map_err(Error::authentication)?;
+        Ok(tok.value)
+    }
+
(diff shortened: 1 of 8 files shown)


generator/testdata/rust/gclient/golden/iam/v1/src/transport.rs
```diff
@@ -85,20 +85,22 @@ impl Iampolicy {
...
+            .cred
+            .access_token()
+            .await
+            .map_err(Error::authentication)?;
+        Ok(tok.value)
+    }
+
     async fn execute<I: serde::ser::Serialize, O: serde::de::DeserializeOwned>(
-        &self,
+        access_token: String,
```
coryan, line 99: I am not sure how you are planning to use `execute()` in pagination, but I assume we want to create the token before requesting each page. So we need to clone `self.inner.cred` and not just the token used in the initial request.

coryan (approved):

generator/testdata/rust/gclient/golden/iam/v1/src/transport.rs
codyoss: The code will request a token from the function passed to the paginator, so I think this will be covered.

Issue #362, opened 2024-12-05
New home for OpenAPI validation library


coryan:
* The OpenAPI integration tests live in `generator/testdata`
* When we remove the generator/ directory they will be gone and the integration tests won't compile.
* The generator should not be running integration tests.
* We need a new location for these libraries.  Somewhere that does not get shipped to [crates.io](http://crates.io/).
* This new location has to work with `sidekick refreshall`.
* We picked `src/generated/openapi-validation`
* We may need to add a flag into the generator to mark crates as "not for publication".

Close this issue once the new library is generated and the integration tests are using it.

PR #369, opened 2024-12-06
feat(generator): new option to prevent publication

coryan:
Some languages have facilities or idioms to prevent accidental publication of packages. This is useful when creating test-only packages or developing new SDKs that are not ready, either because the service or the generator are evolving.

Part of the work for #362


generator/templates/rust/crate/Cargo.toml.mustache
```diff
@@ -28,6 +28,9 @@ license.workspace    = true
 repository.workspace = true
 keywords.workspace   = true
 categories.workspace = true
+{{#NotForPublication}}
+publish              = ["not-for-publication"]
```
codyoss, line 32: nit: maybe simply false here: https://doc.rust-lang.org/cargo/reference/manifest.html#the-publish-field

codyoss (approved):

generator/templates/rust/crate/Cargo.toml.mustache
coryan: D'oh! I should keep reading next time.

PR #376, opened 2024-12-08
ci: add initial publish workflow

### LINKED ISSUE #353: We should run `cargo publish --dry-run` in the CI builds
We need to verify the packages can be created and pass the necessary tests.

SchahinRohani:
Implement a publish dry-run for the google-cloud-auth package to automate testing before actual release.

Closes #353

### DIFF
--- a/.github/workflows/publish.yaml
+++ b/.github/workflows/publish.yaml
@@ -1,0 +1,37 @@
+# Copyright 2024 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Publish Rust SDK
+on: [push, pull_request]
+jobs:
+  google-cloud-auth:
+    runs-on: ubuntu-24.04
+    strategy:
+      matrix:
+        rust-version: ['1.83']
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo
+          key: ${{ github.job }}-${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+      - name: Setup Rust ${{ matrix.rust-version }}
+        run: rustup toolchain install ${{ matrix.rust-version }}
+      - name: Display Cargo version
+        run: cargo version
+      - name: Display rustc version
+        run: rustc --version
+      - name: Dry-run cargo publish
+        run: cargo publish --dry-run -p google-cloud-auth


coryan:
Thank you for your contribution.

Unfortunately, this is not a good place to start.  We are not planning to (ever) publish `google-cloud-auth` in the `/auth` directory. We are writing a new version in `gcp-sdk-auth` (under `/src/auth`) and it is nowhere near the point when it would be suitable for publication.

A better starting point might be `gcp-sdk-wkt`, but even then we are reluctant to accept external contributions for the CI system.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,37 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+name: Publish Rust SDK
+on: [push, pull_request]
+jobs:
+  google-cloud-auth:
```
coryan, line 18: The name should be something like `packaging-dry-run` that is neutral w.r.t. to the package being tested.

.github/workflows/publish.yaml
```diff
@@ -0,0 +1,37 @@
...
+            ~/.cargo
+          key: ${{ github.job }}-${{ runner.os }}-cargo-${{ hashFiles('Cargo.lock') }}
+      - name: Setup Rust ${{ matrix.rust-version }}
+        run: rustup toolchain install ${{ matrix.rust-version }}
+      - name: Display Cargo version
+        run: cargo version
+      - name: Display rustc version
+        run: rustc --version
+      - name: Dry-run cargo publish
+        run: cargo publish --dry-run -p google-cloud-auth
```
coryan, line 37: This may be a better start, but what we really want is "all the packages that are not explicitly disabled".

```suggestion
        run: cargo publish --dry-run -p gcp-sdk-wkt
```

SchahinRohani:
@coryan Thank you for your feedback!

A little bit of context: I am working on a `Google Cloud Storage` integration for [nativelink](https://github.com/TraceMachina/nativelink), therefore i was researching on existing solutions and came across this project. And of course it looks like a well-thought and clean approach. 
I would be glad if i can contribute sometimes, so i can use that sdk in the future.

.github/workflows/publish.yaml
SchahinRohani: I was thinking of simply calling it `packages` .

```yaml
jobs:
  packages:
    runs-on: ubuntu-24.04
    strategy:
      matrix:
        packages: ['auth', 'wkt', 'gax']
```

This would result to a ci title on push for the package wkt to:  `Publish Rust SDK / packages (wkt) (push)`.

.github/workflows/publish.yaml
SchahinRohani: This part would result to:

```yaml
    strategy:
      matrix:
        packages: ['auth', 'wkt', 'gax']
        .
        .
        .
      - name: Dry-run cargo publish
        run: cargo publish --dry-run -p gcp-sdk-${{ matrix.packages }}
```

coryan:
> @coryan Thank you for your feedback!
> 
> A little bit of context: I am working on a `Google Cloud Storage` integration for [nativelink](https://github.com/TraceMachina/nativelink),

Thanks for that context.  For what is worth: we will eventually release a crate for Google Cloud Storage, but we cannot give you a timeline, other than "not soon".  You may want to consider existing OSS crates that support Google Cloud Storage: https://crates.io/search?q=gcs

> therefore i was researching on existing solutions and came across this project. And of course it looks like a well-thought and clean approach.

Thanks for saying that.

> I would be glad if i can contribute sometimes, so i can use that sdk in the future.

And we would welcome your contributions, you probably picked an area that is too immature at the moment.

For example, running `cargo package -o gcp-sdk-gax` will fail locally, and after we fix that problem: `publish --dry-run` will fail until version 0.1.0 of `gcp-sdk-wkt` is published.

We also anticipating dozens, maybe a couple of hundred crates to be published.  Listing all of them explicitly will get tedious rather quickly.  We would prefer something like `-p 'gcp-sdk-*'`, but as I said that does not work yet.

I think we should close this PR and wait until the project has matured enough to get a CI going.

PR #377, opened 2024-12-08
cleanup: published dependencies need versions

coryan:
Any package we intend to publish should include a version for each of its
dependencies. This PR marks a few more crates as `publish = false`, and marks
all the dependencies with a specific version.


coryan:
Part of the work for #353

codyoss (approved):

Issue #370, opened 2024-12-06
Generate clients using traits+transport+tracing+client structure


coryan:
As discussed (outside GitHub) we will change the generator to generated code more like this:

```rust
struct Client {
  inner: Box<dyn traits::dynamic::Client>, // Can change the implementation dynamically
}

impl traits::Client for Client { /* RPCs go here */ }
```

Applications create these clients via a builder (which will be the subject of a separate feature request), which initializes the inner client using internal crate types. The (non-dynamic) traits are public and finesse around the `Send` requirements. Applications can use these to create mocks:

```rust
pub mod traits {
  pub trait Client {
    fn create_foo(&self, req: CreateFooRequest) 
        -> impl Future<Output = Result<Foo>> + Send;
  }
}
```
 
The dynamic traits are not public and use the `async-trait` crate:

```rust
pub mod traits {
    pub(crate) mod dynamic {
        #[async_trait::async_trait]
        pub trait Client: Send + Sync {
            async fn create_foo(&self, req: CreateFooRequest) -> Result<Foo>;
        }

        #[async_trait::async_trait]
        impl<T: super::Client> Client for T {
            async fn create_foo(&self, req: CreateFooRequest) -> Result<Foo> {
                let request = T::create_foo(self, req).await?;
                Ok(request)
            }
        }
    }
}
```

----

The builder will add a tracing layer if requested:

```rust
pub mod builder {
    use super::*;
    use crate::gax5;

    #[derive(Clone, Debug, Default)]
    pub struct FooService {
        tracing: bool,
        // This serves as a proxy for all the configuration optiosn, retry loops, timeouts, etc.
        endpoint: Option<String>,
    }

    impl FooService {
        pub fn with_endpoint<T: Into<String>>(mut self, value: T) -> Self {
            self.endpoint = Some(value.into());
            self
        }

        pub fn with_tracing(mut self) -> Self {
            self.tracing = true;
            self
        }

        pub fn build(self) -> client::FooService {
            return client::FooService::new(self.build_inner());
        }

        fn build_inner(self) -> Box<dyn traits::dynamic::FooService> {
            if Self::tracing_enabled(self.tracing) {
                let client = StaticBuilder::new(self.endpoint).build_with_tracing();
                return Box::new(client);
            }
            let client = StaticBuilder::new(self.endpoint).build();
            return Box::new(client);
        }

        fn tracing_enabled(tracing: bool) -> bool {
            if tracing {
                return true;
            }
[...]

PR #380, opened 2024-12-09
feat(generator/rust): introduce client types

coryan:
This introduces the Client type we think most applications will use. It will
be dynamically configured, but in this version it is rather trivial and not
very interesting.

Part of the work for #370


generator/templates/rust/crate/src/client.rs.mustache
```diff
@@ -0,0 +1,71 @@
...
+}
+
+impl crate::traits::{{NameToPascal}} for {{NameToPascal}} {
+    {{#Methods}}
+    {{#DocLines}}
+    {{{.}}}
+    {{/DocLines}}
+    async fn {{NameToSnake}}(&self, req: {{InputTypeName}}) -> Result<{{OutputTypeName}}> {
+        let response = self.inner.{{NameToSnake}}(req).await?;
+        Ok(response)
```
codyoss, line 66: nit: Can this be simplified to a oneliner?  `self.inner.{{NameToSnake}}(req).await`

codyoss (approved):

generator/templates/rust/crate/src/client.rs.mustache
coryan: Good idea. Done.

PR #386, opened 2024-12-09
feat(generator/rust): add tracing decorator

### LINKED ISSUE #370: Generate clients using traits+transport+tracing+client structure

coryan:
Introduce a decorator with Tokio tracing support. This can be used for both logging and remote tracing.

Fixes #370


codyoss (commented):
Nice!

codyoss (approved):

PR #388, opened 2024-12-09
 cleanup: refactor `InnerClient` to gax

coryan:
We had repeated code in all the generated libraries. This change
refactors `crate::InnerClient` to `gax::http_client::ReqwestClient`.
This also enables future improvements, such as retry loops and timeouts
implemented in a single place, and (maybe) better tests for the gax
components.


src/gax/src/http_client/mod.rs
```diff
@@ -0,0 +1,138 @@
...
+        })
+    }
+
+    pub fn builder(&self, method: reqwest::Method, path: String) -> reqwest::RequestBuilder {
+        self.inner
+            .request(method, format!("{}{path}", &self.endpoint))
+    }
+
+    pub async fn execute<I: serde::ser::Serialize, O: serde::de::DeserializeOwned>(
+        &self,
```
coryan, line 51: @codyoss this version of `execute()` may have wrecked your plans for pagination.  If you can point me to some WIP with how you were planning to use it the version that consumed a `token: String`, I can make sure to adapt my current code.

Or even some pseudo-code about why consuming a `self` was not a great idea.

codyoss: The issue comes when putting any code within a higher order function in rust. When you do this you need to move variables to new code stacks. I was having issues making the compiler happy if self was ever mentioned in the function for this reason -- which does make sense I suppose. If you think about it the list RPCs will return a type out-side the scope of the call to the client, so the execute function passed to the paginator could live longer than the client itself. So it all has to do with lifetimes. You could replicate this with any lambda I assume, or a func that takes one of the Fn traits.

```
    // build up a request
        gax::paginator::Paginator::new(tok, move |token| {
            // do some non-async things
            async move {
                // call the execute function that refs self and see a compile error.
                self.execute(...)
            }
        })
```

codyoss: It could be I was holding it wrong here too. Would be happy to get some more input if I am simply missing something. All the moving done with the higher order functions was tripping me up a bit.

coryan: PTAL.  In generated code there is an extra layer of indirection (the `client::Foo` and `traits::dyntraits::Foo`), but I think that will make things faster (clones are cheaper), and not affect the lifetime analysis.

src/gax/tests/paginator.rs
```diff
@@ -0,0 +1,174 @@
...
+use gcp_sdk_gax::http_client::*;
+use gcp_sdk_gax::paginator::{PageableResponse, Paginator};
+use std::collections::HashMap;
+use warp::http::Response;
+use warp::Filter;
+
+type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
+
+#[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+async fn test_paginator() -> Result<()> {
```
codyoss, line 24: Thank you for adding these tests that I should have in the first place!

src/gax/src/lib.rs
```diff
@@ -65,6 +65,7 @@ pub mod path_parameter;
 
 /// Implementation details for [query_parameter](::crate::query_parameter) and
 /// [path_parameter](::crate::path_parameter).
+#[cfg(feature = "unstable-sdk-client")]
```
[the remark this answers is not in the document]
coryan: Done. Though it is a private module, so no documentation is generated in any case.

src/generated/cloud/location/src/transport.rs
```diff
@@ -14,79 +14,32 @@
...
-    http_client: reqwest::Client,
-    cred: Credential,
-    endpoint: String,
-}
-
-#[derive(serde::Serialize)]
-#[allow(dead_code)]
-struct NoBody {}
+use crate::Result;
+// Sometimes there are no RPCs with query parameters and this goes unused.
```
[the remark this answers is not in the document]
coryan: Done.

src/gax/tests/paginator.rs
coryan: No worries.  I think this is cool infrastructure to have for future tests.

Issue #365, opened 2024-12-06
Add/Update Documentation for Auth Library


sai-sunder-s:
The auth library code currently does not include good documentation. Address it in a future PR.

coryan:
I don't like the names, but I think this is obsolete.

PR #390, opened 2024-12-10
docs(auth): close some documentation gaps

coryan:
Part of the work for #365


src/auth/src/token.rs
```diff
@@ -12,20 +12,27 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// Represents an auth token.
+/// Represents an authentication token.
```
sai-sunder-s, line 15: I am still not sold on the idea of calling "authentication".  Access token is for AuthZ and ID Token (from OIDC) is for AuthN. By calling "auth" token, we cover both.

Checkout the last section of https://www.microsoft.com/en-us/security/business/security-101/what-is-openid-connect-oidc
> OIDC was built on top of OAuth 2.0 to add authentication. The OAuth 2.0 protocol was developed first and then OIDC was added to enhance its capabilities. The difference between the two is that OAuth 2.0 provides authorization, while OIDC provides authentication. OAuth 2.0 is what allows users to gain access to a relying party, using their account with an OpenID provider, and OIDC is what allows the OpenID provider to pass along a user profile to the relying party. OIDC also allows organizations to offer their users single sign-on.

coryan: It might well be be that OAuth 2.0 provides Authz.  But the token itself does not.  

Authz happens on the server-side **after** the access token is used to establish identity (that is Authn).  It cannot possibly work any other way: 

* when the library (or MDS) creates an access token we do not know what resource it will be used with, so no Authz at that time.
* the service only knows what principals have access to what resources, not what tokens have access to what resources , so the token by itself cannot grant Authz .
* Consequently, the token can only be used to establish what principal is issuing the request: that is the definition of Authn.

In any case, if I did not convince you, I will drop this change.

sai-sunder-s: Actual "Authentication" is done by the IDP at the time of issuing the token. 

> the token can only be used to establish what principal is issuing the request

The primary purpose of access token is to say what scopes the principal has access to, not just who it is. It says what APIs the holder of the token has access to on behalf of the principal. i.e. User U may have access to Blob B but if during the consent flow, they do not give access to gcloud for Storage API, then when gcloud uses the token to access B, it is going to fail because the user did not give permission to use the storage API on its behalf. So the access token carries some authorization details as well, in addition to whatever IAM does to check if U has permission on B.

> not what tokens have access to what resources , so the token by itself cannot grant Authz

Like mentioned above, even though token doesn't say access to what resources, it says access to which scopes/ APIs.

Also, in the comment below we say "Authorization" header. If  a user is going to read through them, it is going to confuse them why an "Authentication" token is used for "Authorization". If the user is not going to care the difference in terminology, we might as well use the term "auth" to cover both.

Even google's endpoints and scopes just use "auth".
eg: oauth2.googleapis.com/auth for getting access token
https://www.googleapis.com/auth/cloud-platform scope

coryan: Reverted.  You convinced me that access tokens encode both Authn and Authz aspects in them.

OTOH, you folks really need to come up with a better terminology than "auth token". That is meaningless and confusing for folks like me that spend a fraction of our time on Authn / Authz.  And frankly, I wish that fraction was smaller.

sai-sunder-s: Also needs update in other places.

coryan: Maybe. Let's merge all the other changes and we can bikeshed on this one line later?

src/auth/src/credentials.rs
```diff
@@ -15,24 +15,53 @@
 use http::header::{HeaderName, HeaderValue};
 use std::future::Future;
 
-pub type Result<T> = std::result::Result<T, crate::errors::CredentialError>;
+type Result<T> = std::result::Result<T, crate::errors::CredentialError>;
 
-/// Represents an auth credential used to obtain auth tokens.
-/// Implementors of this trait provide a way to asynchronously retrieve tokens
-/// and construct auth headers.
+/// Represents a [Credential] used to obtain authentication
```
sai-sunder-s, line 20: Update "Authentication" here and next line

src/auth/src/credentials.rs
```diff
@@ -15,24 +15,53 @@
...
     ///
-    /// This function returns a `Future` that resolves to a `Result` containing
-    /// either the `Token` or an `AuthError` if an error occurred during
-    /// token retrieval.
+    /// Returns a [Token][crate::token::Token] for the current credentials.
+    /// The underlying implementation refreshes the token as needed.
     fn get_token(&mut self) -> impl Future<Output = Result<crate::token::Token>> + Send;
 
-    /// Asynchronously retrieves auth headers.
+    /// Asynchronously constructs the authentication headers.
```
sai-sunder-s, line 58: Update "Authentication" here

src/auth/src/credentials.rs
```diff
@@ -15,24 +15,53 @@
...
     ///
-    /// This function returns a `Future` that resolves to a `Result` containing
-    /// either a vector of key-value pairs representing the headers or an
-    /// `AuthError` if an error occurred during header construction.
+    /// Different authentication tokens are sent via different headers. The
```
sai-sunder-s, line 60: Update "Authentication" here

sai-sunder-s (approved):

src/auth/src/token.rs
sai-sunder-s: Yes. Feel free to open an issue and we can handle this name change later in all the places.

coryan: Done #395

PR #391, opened 2024-12-10
feat(generator): new cmd to update googleapis-root

### LINKED ISSUE #271: Update the googleapis SHA and regenerate all the client libraries.

coryan:
The `update` command updates the `googleapis-root` SHA and then
refreshes all the generated code.

Fixes #271

### DIFF
--- a/generator/internal/sidekick/sidekick.go
+++ b/generator/internal/sidekick/sidekick.go
@@ -61,19 +61,14 @@ func runSidekick(cmdLine *CommandLine) error {
 
 	switch cmdLine.Command {
 	case "generate":
-		if err := generate(config, cmdLine); err != nil {
-			return err
-		}
+		return generate(config, cmdLine)
 	case "refresh":
-		if err := refresh(config, cmdLine, cmdLine.Output); err != nil {
-			return err
-		}
+		return refresh(config, cmdLine, cmdLine.Output)
 	case "refresh-all", "refreshall":
-		if err := refreshAll(config, cmdLine); err != nil {
-			return err
-		}
+		return refreshAll(config, cmdLine)
+	case "update":
+		return update(config, cmdLine)
 	default:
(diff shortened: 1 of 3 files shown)


coryan:
The generator breakage is unrelated and will be fixed by #392

generator/internal/sidekick/update.go
```diff
@@ -0,0 +1,152 @@
...
+	}
+	return refreshAll(rootConfig, cmdLine)
+}
+
+func updateRootConfig(rootConfig *Config) error {
+	gitHubApi, ok := rootConfig.Source["github-api"]
+	if !ok {
+		gitHubApi = defaultGitHubApi
+	}
+	gitHub, ok := rootConfig.Source["github"]
```
codyoss, line 53: Note: I think this is fine for now. But in the future we may want to also honor an environment variable to source a local github token if we notice we are running into quota issues. There are some good libs out there that make all this pretty easy that I have used before.

generator/internal/sidekick/sidekick.go
```diff
@@ -61,19 +61,14 @@ func runSidekick(cmdLine *CommandLine) error {
...
-		if err := refresh(config, cmdLine, cmdLine.Output); err != nil {
-			return err
-		}
+		return refresh(config, cmdLine, cmdLine.Output)
 	case "refresh-all", "refreshall":
-		if err := refreshAll(config, cmdLine); err != nil {
-			return err
-		}
+		return refreshAll(config, cmdLine)
+	case "update":
```
codyoss, line 69: Can we file a bug to add help text for these different commands?

codyoss (approved):

generator/internal/sidekick/sidekick.go
coryan: Done (#393)

PR #396, opened 2024-12-10
feat(gax): support full `Status` message

### LINKED ISSUE #255: update gax::error::Error to support status
When the error is HTTP based the status type can have an extra field called status which is a rpc.Code. See https://google.aip.dev/193#http11json-representation

coryan:
We were missing the `status` field from the JSON encoding.

Fixes #255


coryan:
Hopefully this is what we meant in #255.

PR #398, opened 2024-12-11
impl(internal/language): remove unused CopyrightYear

julieqiu:


coryan (changes requested):
I think this is used, or will be, as soon as we start generating code in 2025.  It is saved here:

https://github.com/googleapis/google-cloud-rust/blob/b6089b583e8c2bb65c9319e059ee7a496e2ed355/src/generated/type/.sidekick.toml#L20

And makes the copyright notices stable, no more spurious changes just because in January we regenerated all the libraries.

julieqiu:
> I think this is used, or will be, as soon as we start generating code in 2025. It is saved here:

I think that is this line of code:
https://github.com/googleapis/google-cloud-rust/blob/main/generator/internal/sidekick/generate.go#L42

which is set with this type:
https://github.com/googleapis/google-cloud-rust/blob/main/generator/internal/sidekick/config.go#L28

But I don't think it actually uses language.Codec

coryan:
st/blob/main/generator/internal/sidekick/config.go#L28
> 
> But I don't think it actually uses language.Codec

You are absolutely right.  The fix is something like #405.  There may be more opportunity for code cleanups after that PR, I did not look that hard once I got the code to do what I think it should do.

PR #410, opened 2024-12-12
cleanup: simplify library generation common case

coryan:
Most libraries will depend on `iam_v1` and `location`. It is easier to
add an exception to ignore them, for example:

`-codec-option package:location=ignore=true`

than to list them in each library that uses IAM, Location, and (soon?)
Longrunning operations.


codyoss (approved):
From my very rough and fast searching in googleapis it seemed about 50/50 for each in the cloud folder. I am fine either way though.

PR #414, opened 2024-12-12
feat(generator): embed the templates

### LINKED ISSUE #413: [sidekick] Consider making templates embeddable
If I build sidekick into a binary, I still need to provide it with a templates folder for it to function. While it may be useful to override the templates location for development, it probably makes more sense for the templates to be bundled with the binary so it is self-contained. See [embed](https://pkg.go.dev/embed) for reference.

coryan:
With this change the templates are embedded into the generator code, so
`go install sidekick` will not need the templates to live in a
well-known directory.

It also opens the door for languages where the set of output files are
dynamic (e.g., Java or C++), will allow us to omit some files (e.g. Rust
when there are no services), and may allow us to change the processing
order.

Fixes #413


generator/internal/language/codec.go
```diff
@@ -20,8 +20,11 @@ import "github.com/googleapis/google-cloud-rust/generator/internal/api"
 // representations. This is used to manipulate the data the is fed into
 // templates to generate clients.
 type Codec interface {
-	// TemplateDir returns the directory containing the templates.
-	TemplateDir() string
+	// TemplatesProvider returns a function that loads templates from whatever
+	// filesystem the Codec is using.
+	TemplatesProvider() func(string) (string, error)
```
codyoss, line 25: nit: by the description I can't tell what the input/output string are here. Can this be rephrased/elaborated? Even just some parameter var name may be enough. Or sometimes when dealing with higher order functions I assign them to a type so I can better document them.

codyoss:
@coryan Could you rebase this? There seem to be extra changes from another PR here.

codyoss (approved):

PR #422, opened 2024-12-13
cleanup(gax): simpler request parameter formatting

### LINKED ISSUE #136: Cleanup Duration-as-a-query-parameter serialization
That `unwrap` is just turning `Option<&str>` into `&str`.  I am using secret knowledge about how `Duration` always serializes to a JSON string.  Maybe a cleaner approach would be to provide a function in `Duration` to format as a string, and then use that here (and in the serializer).

To your larger point: I am not sure it is true that we can always serialize the values we have. For example:  Timestamps can only be (safely) serialized in a certain range:

https://github.com/protocolbuffers/protobuf/blob/d0aefcba0ea1dd814c12915685eda9e953df49bb/src/google/protobuf/timestamp.proto#L53-L55

We currently do not enforce this restriction, but I can see us returning an error if the timestamp cannot be formatted correctly because it is out of range.

_Originally posted by @coryan in https://github.com/googleapis/google-cloud-rust/pull/135#discussion_r1835013713_

coryan:
Change the `Duration` and `Timestamp` well known types so it is easier
to format them as strings (their prescribed JSON formatting). Take
advantage of this change in the implementation of request parameters.

Fixes #136

### DIFF
--- a/src/wkt/src/duration.rs
+++ b/src/wkt/src/duration.rs
@@ -62,6 +62,14 @@ pub enum DurationError {
     /// The sign of the seconds component does not match the sign of the nanoseconds component.
     #[error("if seconds and nanoseconds are not zero, they must have the same sign")]
     MismatchedSigns(),
+
+    /// Cannot serialize the duration.
+    #[error("cannot serialize the duration")]
+    Serializate(),
+
+    /// Cannot deserialize the duration.
+    #[error("cannot deserialize the duration: {0:?}")]
+    Deserialize(String),
 }
 
 type Error = DurationError;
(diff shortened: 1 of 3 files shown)


src/wkt/src/duration.rs
```diff
@@ -154,13 +162,57 @@ impl Duration {
...
+            return format!("{sign}{}s", self.seconds.abs());
+        }
+        if self.seconds == 0 {
+            return format!("{sign}0.{}", self.nanos.abs());
+        }
+        format!("{sign}{}.{:09}s", self.seconds.abs(), self.nanos.abs())
+    }
+
+    /// Parses a duration from the JSON representation.
+    pub fn from_json(value: &str) -> Result<Self, DurationError> {
```
codyoss, line 183: Should this be a tryfrom?

codyoss (approved):

src/wkt/src/duration.rs
coryan: Maybe: #423

PR #425, opened 2024-12-13
impl(gax): support for per-request options

coryan:
This introduces a mechanism to pass arbitrary options into the gax
`ReqwestClient`. I use `UserAgentPrefix` to show how these would work,
but I anticipate many more options (timeouts, retry loop policies, HTTP
version strings, proxy configuration?). And in the hand-crafted
libraries we will have even more options, many of them service specific.

Also introduced an echo server to simplify testing of the
`ReqwestClient`.


coryan:
We are following a different path

Issue #442, opened 2024-12-16
Bootstrap `gcp-sdk-auth`


dbolduc:
We want to swap out the auth implementation from `auth` -> `src/auth`. Before doing this the bare minimum we need is ADC support for user credentials and MDS credentials.

Initially we do not need any configuration options or helpful things like token caching. We can fill these in later.

The work, broken down into PRs, is roughly:

- [x] Move existing `Credential` trait into `mod traits`, `mod dynamic::traits`, add Credential struct
- [x] Add `TokenProvider` interface
- [x] Add `UserCredential` above `TokenProvider`
- [x] Add `UserCredentialTokenProvider`
- [x] Add `UserCredential` ctor from json
- [x] Add ADC File path function (does not need to respect env var yet)
- [x] Add `create_access_token_credential()`
  - This thing loads and parses ADC json
  - For now, if no file -> error
  - If `type != "authorized_user"` -> error
  - return a `UserCredential`
- [x] Add `MdsCredential` implementation
- [x] In `create_access_token_credential()`, if no file -> `MdsCredential`
- [x] Refactor `gax` to use `gcp-sdk-auth` instead of `google-cloud-auth`

PR #455, opened 2024-12-16
impl(auth): introduce `TokenProvider`, `UserCredentials`

dbolduc:
Part of the work for #442 

Introduce the (internal) `TokenProvider`, and (internal) `UserCredential` types.

My intention is to mock the `TokenProvider` in the test cases. It looks like unit tests for internal classes belong in the same file as the implementation, so that is what I did for `user_credential.rs`.

Q: Should unit tests return `Result`s?

I decided to cut off the PR early (without implementing `get_headers()`) because I have probably made too many mistakes already.

Note that `CredentialError` is using a `Box`, so we cannot clone it. That prevents me from using code like this in my test. Maybe it should use an `Arc`?

```rs
struct TestTokenProvider {
  token: Result<Token>,
}
```

### DIFF
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -28,3 +28,7 @@ async-trait = "0.1.83"
 thiserror   = "2"
 http        = "1.2.0"
 time        = "0.3.37"
+
+[dev-dependencies]
+test-case = "3.3.1"
+tokio     = { version = "1.41.1", features = ["macros"] }
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -30,5 +30,6 @@ http        = "1.2.0"
 time        = "0.3.37"
 
 [dev-dependencies]
+mockall   = "0.13.1"
 test-case = "3.3.1"
-tokio     = { version = "1.41.1", features = ["macros"] }
+tokio     = { version = "1.42", features = ["macros", "test-util"] }
--- a/src/auth/src/credentials/user_credential.rs
+++ b/src/auth/src/credentials/user_credential.rs
@@ -20,12 +20,18 @@ use http::header::{HeaderName, HeaderValue};
 
 /// Data model for a UserCredential
 #[allow(dead_code)] // TODO(#442) - implementation in progress
-pub(crate) struct UserCredential {
-    token_provider: Box<dyn TokenProvider>,
+pub(crate) struct UserCredential<T>
+where
+    T: TokenProvider,
+{
+    token_provider: T,
 }
 
 #[async_trait::async_trait]
-impl Credential for UserCredential {
+impl<T> Credential for UserCredential<T>
+where
+    T: TokenProvider,
+{
     async fn get_token(&mut self) -> Result<Token> {
         self.token_provider.get_token().await
     }
@@ -43,32 +49,7 @@ impl Credential for UserCredential {
 #[cfg(test)]
 mod test {
     use super::*;
-
-    struct TestTokenProvider {
-        token: Token,
-    }
-
-    #[async_trait::async_trait]
-    impl TokenProvider for TestTokenProvider {
-        async fn get_token(&mut self) -> Result<Token> {
-            Ok(self.token.clone())
-        }
-    }
-
-    struct TestErrorProvider {
-        is_retryable: bool,
-        message: String,
-    }
-
-    #[async_trait::async_trait]
-    impl TokenProvider for TestErrorProvider {
-        async fn get_token(&mut self) -> Result<Token> {
-            Err(CredentialError::new(
-                self.is_retryable,
-                BoxError::from(self.message.clone()),
-            ))
-        }
-    }
+    use crate::token::test::MockTokenProvider;
 
     #[tokio::test]
     async fn get_token_success() {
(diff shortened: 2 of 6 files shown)


src/auth/Cargo.toml
```diff
@@ -28,3 +28,7 @@ async-trait = "0.1.83"
 thiserror   = "2"
 http        = "1.2.0"
 time        = "0.3.37"
+
+[dev-dependencies]
+test-case = "3.3.1"
+tokio     = { version = "1.41.1", features = ["macros"] }
```
coryan, line 34: I think we moved to 1.42 elsewhere in the project.  You may also want the `test-util` feature.

src/auth/src/credentials/user_credential.rs
```diff
@@ -0,0 +1,103 @@
...
+    async fn get_universe_domain(&mut self) -> Option<String> {
+        Some("googleapis.com".to_string())
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    struct TestTokenProvider {
```
dbolduc, line 47: also, I think other Credential types will want to reuse this thing in their unit tests. Does it belong in a

```
// file = token.rs
#[cfg(test)]
pub(crate) mod test {
  struct TestTokenProvider { ... }
}
```

coryan:
> Note that `CredentialError` is using a `Box`, so we cannot clone it. That prevents me from using code like this in my test. Maybe it should use an `Arc`?


You can ask for `Box<dyn Error + Clone + ...>` if cloning is going to be common.  Alternatively, you can have a `Vec<Result<>>` if your mocks and return the last one (by move, so no cloning or copying is needed).

src/auth/src/credentials/user_credential.rs
```diff
@@ -0,0 +1,103 @@
...
+use crate::credentials::Result;
+use crate::errors::{BoxError, CredentialError};
+use crate::token::{Token, TokenProvider};
+use http::header::{HeaderName, HeaderValue};
+
+/// Data model for a UserCredential
+#[allow(dead_code)] // TODO(#442) - implementation in progress
+pub(crate) struct UserCredential {
+    token_provider: Box<dyn TokenProvider>,
+}
```
coryan, line 25: Do you expect to dynamically change the type of token provider here?  If not, consider:

```suggestion
pub(crate) struct UserCredential<T> where T: impl TokenProvider + /* whatever else you need */ {
    token_provider: T,
}
```

You can still mock, you can still decorate this with tracing (with `T == TracingProvider<U>`), and you can avoid all the dynamic allocations.

src/auth/src/credentials/user_credential.rs
coryan: That sounds like a good idea, if you find this use-case is common enough.  I think you should use mocks instead of your own classes, see:

https://docs.rs/mockall/latest/mockall/

src/auth/src/token.rs
```diff
@@ -12,8 +12,10 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::credentials::Result;
```
coryan, line 15: This suggests that maybe `Result` should be at the `crate::` level.

src/auth/src/token.rs
```diff
@@ -36,3 +38,9 @@ pub struct Token {
     /// This might include information like granted scopes or other claims.
     pub metadata: Option<std::collections::HashMap<String, String>>,
 }
+
+#[async_trait::async_trait]
+#[allow(dead_code)] // TODO(#442) - implementation in progress
+pub(crate) trait TokenProvider: Send + Sync {
```
coryan, line 44: `pub(crate)` is fine for now.  You folks will need to decide if external applications can implement their own `TokenProvider`.  If you do, they this should be `pub` instead of `pub(create)`.

coryan (changes requested):

src/auth/src/token.rs
dbolduc: Good idea, let's start with this while I work on the other stuff. #457

src/auth/src/token.rs
dbolduc: Ack. I don't think we want the thing public. If we change our minds, we can always go private -> public later.

src/auth/Cargo.toml
dbolduc: Updated.

> I think we moved to 1.42 elsewhere in the project.

Everywhere except the `Cargo.toml` I copied. :smile: 

https://github.com/googleapis/google-cloud-rust/blob/e33a1bb86ae33d4c5ef65aa85f88560705b26229/src/gax/Cargo.toml#L47

src/auth/src/credentials/user_credential.rs
dbolduc: > Do you expect to dynamically change the type of token provider here?

No, I expect we will make a concrete implementation of `TokenProvider`, feed it to this `UserCredential` struct, and never change it.

> If not, consider...

Adopted.

src/auth/src/credentials/user_credential.rs
dbolduc: Thanks. I used `mockall`. I know that I want `mockall` as a dev-dependency only. (We only use it to test our internal code).

I think that means I can't skip this boilerplate by using the `#[automock]` attribute.

https://github.com/googleapis/google-cloud-rust/blob/6ba4912d9df20ddd56a44cfe3bbedd9a3a2aeaaf/src/auth/src/token.rs#L52-L59

coryan (approved):

PR #456, opened 2024-12-16
chore: prepare for 0.1.0-rc2

coryan:
Missed the hand-crafted libraries (wkt, gax, auth) in the *-rc1
versions.

### DIFF
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name                 = "gcp-sdk-auth"
-version              = "0.0.0"
+version              = "0.1.0-rc2"
 description          = "Google Cloud SDK for Rust"
 edition.workspace    = true
 authors.workspace    = true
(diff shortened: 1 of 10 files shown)


src/auth/Cargo.toml
```diff
@@ -14,7 +14,7 @@
 
 [package]
 name                 = "gcp-sdk-auth"
-version              = "0.0.0"
+version              = "0.1.0-rc2"
```
dbolduc, line 17: I am not sure this should this be included. No other packages depend on it.

I think you want https://github.com/googleapis/google-cloud-rust/blob/30a8efaf921c105ab78650c476aecddb280bc6a0/auth/Cargo.toml#L17

coryan: > No other packages depend on it.

Yup.

I am trying to discover what will be the process for releases.  I expect all these "core" libraries will have a release [^1] in the future.  So I would *like* to have `gax`, `auth`, and `wkt` already released, with all having the same version.  We will need to have multiple versions in the future. I just want to introduce that complexity as-needed.

[^1]: that is, an existing crate version with the code matching some tag in the  GitHub repo.

codyoss (approved):

PR #457, opened 2024-12-16
refactor(auth): move `Result` definition

dbolduc:
Re: https://github.com/googleapis/google-cloud-rust/pull/455#discussion_r1887541999

### DIFF
--- a/src/auth/src/lib.rs
+++ b/src/auth/src/lib.rs
@@ -41,3 +41,5 @@ pub mod credentials;
 ///
 /// [Tokens]: https://cloud.google.com/docs/authentication#token
 pub mod token;
+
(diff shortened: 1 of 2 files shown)


src/auth/src/lib.rs
```diff
@@ -41,3 +41,5 @@ pub mod credentials;
 ///
 /// [Tokens]: https://cloud.google.com/docs/authentication#token
 pub mod token;
+
+pub(crate) type Result<T> = std::result::Result<T, crate::errors::CredentialError>;
```
coryan, line 45: Consider a comment, in case this becomes public.  Something like:

https://docs.rs/reqwest/latest/reqwest/type.Result.html

dbolduc: Done.

Issue #27, opened 2024-10-30
implement proto annotation we don't process today 


codyoss:
- proto.GetExtension(m, annotations.E_FieldBehavior)
- proto.GetExtension(m, annotations.E_FieldInfo)
- proto.GetExtension(m, extendedops.E_OperationService) -- and all other associated annotations
- proto.GetExtension(m, longrunning.E_OperationInfo)
- proto.GetExtension(m, annotations.E_OauthScopes)
- proto.GetExtension(m, annotations.E_Routing)
- proto.GetExtension(m, annotations.E_ApiVersion)

codyoss:
Prioritizing field_behavior first here as we need it to better generate our proto based models.

PR #462, opened 2024-12-17
feat(sidekick): source field behavior annotations in proto clients 

codyoss:
This relates to https://google.aip.dev/203. If a field is not labeled
required we should consider it optional. This field behavior annotations
are traditionally only one request messages which in turn makes almost
everything optional on response messages. This change should make clients
generated from proto and openapi have closer req/resp models.

Updates: https://github.com/googleapis/google-cloud-rust/issues/27

### DIFF
--- a/generator/internal/parser/protobuf.go
+++ b/generator/internal/parser/protobuf.go
@@ -482,7 +482,7 @@ func processMessage(state *api.APIState, m *descriptorpb.DescriptorProto, mFQN, 
 			Name:     mf.GetName(),
 			ID:       mFQN + "." + mf.GetName(),
 			JSONName: mf.GetJsonName(),
-			Optional: isProtoOptional,
+			Optional: isProtoOptional && !isFieldRequired(mf),
 			Repeated: mf.Label != nil && *mf.Label == descriptorpb.FieldDescriptorProto_LABEL_REPEATED,
 			IsOneOf:  mf.OneofIndex != nil && !isProtoOptional,
 		}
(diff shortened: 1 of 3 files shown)


generator/testdata/rust/gclient/golden/iam/v1/src/model.rs
```diff
@@ -26,7 +26,7 @@ pub struct SetIamPolicyRequest {
 
     /// REQUIRED: The resource for which the policy is being specified.
     /// See the operation documentation for the appropriate value for this field.
-    pub resource: String,
+    pub resource: Option<String>,
```
coryan, line 29: required put optional?

src/generated/type/src/model.rs
```diff
@@ -1184,39 +1184,39 @@ impl PostalAddress {
 #[non_exhaustive]
 pub struct Quaternion {
     /// The x component.
-    pub x: f64,
+    pub x: Option<f64>,
```
coryan, line 1187: I am not sure this is the code we want.  Seems like the annotations are creating more trouble than it is worth.

Maybe we should discuss what specific problem the non-optional fields caused with pagination?

generator/testdata/rust/gclient/golden/iam/v1/src/model.rs
codyoss: this was a bug.

src/generated/type/src/model.rs
codyoss: Yes, I was honoring the logic that states we should treat things as optional if not specified, but I think this was too far reaching as it should only apply to the request messages -- from my understanding. When I reduced the logic to ensure the required annotation was being strictly honored it produced no diffs. May look at this again in the future but I agree it is best to just close this for now.

PR #464, opened 2024-12-17
impl(auth): get_headers for user credentials

dbolduc:
Part of the work for #442 

Implement the `get_headers()` function (based on the `get_token()` function) for user credentials.

Note that we set the header as sensitive, which is what `reqwest::RequestBuilder::bearer_auth(...)` does: https://docs.rs/reqwest/0.12.9/src/reqwest/async_impl/request.rs.html#270. (I could not find a public ctor that took `is_sensitive`, so we do it in two lines).

---

Is there a better way to set test expectations on a container?

TIL: `map_err()`. Check me on my error propagation.

### DIFF
--- a/src/auth/src/credentials/user_credential.rs
+++ b/src/auth/src/credentials/user_credential.rs
@@ -41,8 +41,7 @@ where
         let mut value = HeaderValue::from_str(&format!("{} {}", token.token_type, token.token))
             .map_err(|e| CredentialError::new(false, e.into()))?;
         value.set_sensitive(true);
-        let headers = vec![(AUTHORIZATION, value)];
-        Ok(headers)
+        Ok(vec![(AUTHORIZATION, value)])
     }
 
     async fn get_universe_domain(&mut self) -> Option<String> {
@@ -92,6 +91,13 @@ mod test {
 
     #[tokio::test]
     async fn get_headers_success() {
+        #[derive(Debug, PartialEq)]
+        struct HV {
+            header: String,
+            value: String,
+            is_sensitive: bool,
+        }
+
         let token = Token {
             token: "test-token".to_string(),
             token_type: "Bearer".to_string(),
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/user_credential.rs
```diff
@@ -37,8 +37,12 @@ where
...
 
     async fn get_headers(&mut self) -> Result<Vec<(HeaderName, HeaderValue)>> {
-        // TODO(#442) - implementation in progress
-        Err(CredentialError::new(false, BoxError::from("unimplemented")))
+        let token = self.get_token().await?;
+        let mut value = HeaderValue::from_str(&format!("{} {}", token.token_type, token.token))
+            .map_err(|e| CredentialError::new(false, e.into()))?;
+        value.set_sensitive(true);
+        let headers = vec![(AUTHORIZATION, value)];
+        Ok(headers)
```
coryan, line 45: I would probably just write:

```suggestion
        Ok(vec![(AUTHORIZATION, value)])
```

But both seem fine.

src/auth/src/credentials/user_credential.rs
```diff
@@ -78,11 +82,46 @@ mod test {
...
+        mock.expect_get_token().times(1).return_once(|| Ok(token));
+
+        let mut uc = UserCredential {
+            token_provider: mock,
+        };
+        let actual = uc.get_headers().await.unwrap();
+        assert_eq!(actual.len(), 1);
+        assert_eq!(actual[0].0, AUTHORIZATION);
+        assert_eq!(actual[0].1, "Bearer test-token");
+        assert!(actual[0].1.is_sensitive());
```
coryan, line 112: Consider something like:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=af731026c583586602b056c07e2b44b4

You may need to transform the `HeaderMap` into something with strings...

coryan (approved):
The tests could use some cleanups, the rest is just nitpicky.

src/auth/src/credentials/user_credential.rs
```diff
@@ -78,11 +82,46 @@ mod test {
...
+    }
+
+    #[tokio::test]
+    async fn get_headers_failure() {
+        let mut mock = MockTokenProvider::new();
+        mock.expect_get_token()
+            .times(1)
+            .return_once(|| Err(CredentialError::new(false, Box::from("fail"))));
```
sai-sunder-s, line 120: consider adding a test where retryable is returned true from the provider.

src/auth/src/credentials/user_credential.rs
dbolduc: Fixed.

src/auth/src/credentials/user_credential.rs
dbolduc: > You may need to transform the `HeaderMap` into something with strings...

Done.

The conversion into the struct is a bit gross (aesthetically), but I find the code easy to read, so I am happy with it.

src/auth/src/credentials/user_credential.rs
dbolduc: No thanks. I am not too concerned about accidentally manipulating the error before returning it here:

https://github.com/googleapis/google-cloud-rust/blob/60bbda3220e13784278e6b33c286df3b2fa91122/src/auth/src/credentials/user_credential.rs#L40

Issue #469, opened 2024-12-18
Change generated clients to use per-request builders


coryan:
# Proposal

Change the generated clients so they can be used as follows:

```rust
async fn app1(client: &secretmanager_v1::client::Client) -> Result<()> {
  let response = client.create_secret() // no parameters
      .set_parent("projects/my-project/locations/us-central1")
      .set_secret_id("my-secret")        // set parameters directly
      .send()  // sends the request
      .await?; // always asynchronous
  println!("response = {response}");
}

async fn app2(client: &secretmanager_v1::client::Client) -> Result<()> {
  let response = client.create_secret() // no parameters
      .set_parent("projects/my-project/locations/us-central1")
      .set_secret_id("my-secret")        // set parameters directly
      .with_timeout(std::time::Duration::from_secs(30)) // set options too!
      .with_user_agent("app2/3.4.5")
      .with_retry_policy(/* details omitted */)
      .send()  // sends the request
      .await?; // always asynchronous
  println!("response = {response}");
}
```

The implementation will change along these lines:

```rust
impl SecretManagerService {
  fn create_secret() -> crate::builder::CreateSecret;
}
```

In a future change this function may consume all required parameters for the `CreateSecretRequest`. In this change we will keep this simple.

Note that each function returns a different builder. All these builders can be implemented with a generic struct.  The `send()` function finalizes them and makes the request:

```rust
// In the `crate::builder` module:
pub type CreateSecret = RequestBuilder<crate::model::CreateSecretRequest>;

impl CreateSecret {
  pub async fn send(self) -> Result<crate::model::Secret> {
      // call the right function in the implementation.
      self.client.create_request(self.request, self,options).await
  }
}

struct RequestBuilder<T> {
    // The request message
    request: T,
    // Any options
    options: /*TBD*/,
    // The mockable client
    client: Arc<dyn crate::traits::dyntraits::SecretManagerService>,
}
```

This approach works well with pagination and LROs: the streaming call can be another finalization call on the builder:

```rust
// In the `crate::builder` module:
pub type ListSecrets = RequestBuilder<crate::model::ListSecretsRequest>;

impl ListSecrets {
  pub async fn send(self) -> Result<crate::model::Secret> { /* as before */ }

  #[cfg(feature = "unstable-stream")]
  pub fn by_pages(self)
    -> gax::paginators::Paginator<crate::model::ListSecretsResponse, crate::error> {
  }

  #[cfg(feature = "unstable-stream")]
[...]

Issue #461, opened 2024-12-17
Use `IntoIterator<>` for `Vec` and `HashMap` setters


coryan:
As suggested in #460, we can use `IntoIterator<>` to accept more types for `Vec<>` and `HashMap<>` fields.

PR #475, opened 2024-12-18
feat(generator): use per-request builder

coryan:
The client methods now return a builder. The application fills out the parameters, and maybe some options and then `.send()` the builder to make the request.

Part of the work for #469


generator/internal/language/templates/rust/crate/src/builders.rs.mustache
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+}
+
+{{#Methods}}
+/// The request builder for a {{ServiceNameToPascal}}::{{NameToSnake}} call.
+pub struct {{NameToPascal}}({{ServiceNameToPascal}}RequestBuilder<{{InputTypeName}}>);
+
+impl {{NameToPascal}} {
+    pub(crate) fn new(stub: Arc<dyn crate::traits::dyntraits::{{ServiceNameToPascal}}>) -> Self {
```
codyoss, line 51: Should this just take the concrete client instead?

generator/internal/sidekick/templatedata.go
```diff
@@ -155,13 +161,27 @@ func newTemplateData(model *api.API, c language.Codec) *TemplateData {
 	// Delay this until the Codec had a chance to compute what packages are
 	// used.
 	data.RequiredPackages = c.RequiredPackages()
+
+	messagesByID := map[string]*Message{}
```
codyoss, line 165: Is this different than model.State.MessageByID?

generator/internal/sidekick/templatedata.go
```diff
@@ -213,10 +233,11 @@ func newMessage(m *api.Message, c language.Codec, state *api.APIState) *Message
 		DocLines:           c.FormatDocComments(m.Documentation, state),
 		IsMap:              m.IsMap,
 		IsPageableResponse: m.IsPageableResponse,
+		ID:                 m.ID,
 	}
 }
 
-func newMethod(m *api.Method, c language.Codec, state *api.APIState) *Method {
+func newMethod(s *api.Service, m *api.Method, c language.Codec, state *api.APIState) *Method {
```
codyoss, line 240: Within our model many of the elements have a field Parent. I wonder if any of this code is simplified if api.Method has such a field that ties back to the service.

src/integration-tests/src/secret_manager/openapi.rs
```diff
@@ -39,22 +39,20 @@ pub async fn run(tracing: bool) -> Result<()> {
...
-                    smo::model::Secret::default()
-                        .set_replication(
-                            smo::model::Replication::default()
-                                .set_automatic(smo::model::Automatic::default()),
-                        )
-                        .set_labels(
-                            [("integration-test", "true")]
-                                .map(|(k, v)| (k.to_string(), v.to_string())),
-                        ),
+        .create_secret()
```
codyoss, line 42: This is more aesthetically pleasing :)

src/integration-tests/src/secret_manager/openapi.rs
```diff
@@ -39,22 +39,20 @@ pub async fn run(tracing: bool) -> Result<()> {
...
+        .set_project(&project_id)
+        .set_secret_id(&secret_id)
+        .set_request_body(
+            smo::model::Secret::default()
```
codyoss, line 46: Nothing to do with the PR, but I wonder if all these messages should have a new() method that just calls default. I like fewer characters personally.

src/integration-tests/src/secret_manager/openapi.rs
```diff
@@ -39,22 +39,20 @@ pub async fn run(tracing: bool) -> Result<()> {
...
+                .set_replication(
+                    smo::model::Replication::default()
+                        .set_automatic(smo::model::Automatic::default()),
+                )
+                .set_labels(
+                    [("integration-test", "true")].map(|(k, v)| (k.to_string(), v.to_string())),
```
codyoss, line 52: Again nothing to do with this PR, but I wonder if this setter could be broadened so the code is valid without the mapping.

codyoss (approved):
Thank you for this. I think this improves the readability of standard usage a good bit!

generator/internal/sidekick/templatedata.go
coryan: Yes, these are `*sidekick.Message`  not the `*api.Message`.

src/integration-tests/src/secret_manager/openapi.rs
coryan: Ack.  I tend to agree.

src/integration-tests/src/secret_manager/openapi.rs
coryan: Maybe if it consumed `IntoIterator` ?  We have #461 already, also created #477.

src/integration-tests/src/secret_manager/openapi.rs
coryan: #478 to track that work.

generator/internal/language/templates/rust/crate/src/builders.rs.mustache
coryan: I made the change locally.  It feels like a weird dependency cycle to have the `crate::builder::Foo*` objects depend on `crate::client::Foo` and vice-versa.  Leaving as-is, we can change it later since these are all `pub(crate)` methods.

generator/internal/sidekick/templatedata.go
coryan: Will do this in #479, the cleanup looked too big for post-review changes.

PR #483, opened 2024-12-19
feat(gax): use `@type` to tag `ErrorDetails`

### LINKED ISSUE #291: Use `@type` field to distinguish error detail types
The deserialization for error details assumes that each error detail type can be determined from its structure:

https://github.com/googleapis/google-cloud-rust/blob/a4cdb98178a8a358f0ece80dd4677ed8a5af035c/src/gax/src/error/rpc/mod.rs#L282-L287

The encoding is an `Any`:

https://github.com/googleapis/googleapis/blob/99a5e296c891faa79784687422b9d555bcf54a3f/google/rpc/status.proto#L48

That means there should be a `@type` field with the name of the type.  With some custom deserialization we should be able to extract the correct Rust type, without depending on the message structure being different enough.

coryan:
The error details are encoded as any. Protobuf encodes the type in the
`@type` field, and we can use it to disambiguate what error detail we
are dealing with.

Fixes #291


codyoss (approved):
Nice!

PR #484, opened 2024-12-19
feat(rpc): generate library

coryan:


coryan:
I will refactor `gax` to use this library in a followup PR.

src/generated/rpc/README.md
```diff
@@ -0,0 +1,3 @@
+# Google Cloud SDK for Rust - Google RPC Types
+
+Defines RPC types.
```
dbolduc, line 3: Aside: is this generated and maintained by sidekick? If so, should it have a comment saying so?

src/generated/rpc/Cargo.toml
```diff
@@ -0,0 +1,38 @@
...
+version              = "0.1.0-rc2"
+description          = "Google Cloud SDK for Rust - Google RPC Types"
+edition.workspace    = true
+authors.workspace    = true
+license.workspace    = true
+repository.workspace = true
+keywords.workspace   = true
+categories.workspace = true
+
+[dependencies]
```
dbolduc, line 28: not saying it is worth doing, but seems like the dependencies could be pruned for packages without services

dbolduc (approved):

src/generated/rpc/Cargo.toml
coryan: I think it is worth doing #492

src/generated/rpc/README.md
coryan: #493

PR #485, opened 2024-12-19
feat(generator/rust): top-level enums

coryan:
We need to generate top-level enums, that is, those that are not nested
within a message.


generator/internal/language/templates/rust/crate/src/model.rs.mustache
```diff
@@ -25,3 +25,6 @@ limitations under the License.
 {{#Messages}}
 {{> message}}
 {{/Messages}}
+{{#Enums}}
```
codyoss, line 28: So interesting I could have sworn we had this at some point. But I guess not :P

PR #487, opened 2024-12-19
feat(generator): fewer files on nosvc crates

coryan:
A small number of generated libraries have no services. This will create
fewer files for them.


codyoss (approved):
Hooray for a negative commit :)

PR #488, opened 2024-12-19
impl(auth): add helper for retryable errors

dbolduc:
Part of the work for #442 

We will use this function to tell us which errors are retryable. I think they apply broadly (not just to user credentials), so they go in `errors.rs`.

### DIFF
--- a/src/auth/src/errors.rs
+++ b/src/auth/src/errors.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use http::StatusCode;
 use std::error::Error;
 use std::fmt::{Display, Formatter, Result};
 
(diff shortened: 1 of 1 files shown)


src/auth/src/errors.rs
```diff
@@ -88,3 +89,34 @@ impl Display for CredentialError {
 pub enum InnerAuthError {
     // TODO(#389) - define error types here
 }
+
+#[allow(dead_code)] // TODO(#442) - implementation in progress
+pub(crate) fn is_retryable(c: StatusCode) -> bool {
+    c == StatusCode::INTERNAL_SERVER_ERROR
```
coryan, line 95: `INTERNAL_SERVER_ERROR` gave me pause... why is Okay to retry those?  If you have a good reason, just write a comment so future readers are not confused.

sai-sunder-s: Server could have crashed or thrown exception for something not related to the specific request itself right?

If server thinks our request is bad and we should not retry, it should send us 4xx.

src/auth/src/errors.rs
```diff
@@ -88,3 +89,34 @@ impl Display for CredentialError {
...
+        || c == StatusCode::SERVICE_UNAVAILABLE
+        || c == StatusCode::REQUEST_TIMEOUT
+        || c == StatusCode::TOO_MANY_REQUESTS
+}
```
coryan, line 99: For clarity consider:

```suggestion
pub(crate) fn is_retryable(c: StatusCode) -> bool {
   match c {
       StatusCode::INTERNAL_SERVER_ERROR | StatusCode::SERVICE_UNAVAILABLE | StatusCode::REQUEST_TIMEOUT | StatusCode::TOO_MANY_REQUESTS => true,
        _ => false,
      }
}
```

coryan (approved):

sai-sunder-s (approved):

src/auth/src/errors.rs
dbolduc: Done, plus a comment about `500`

PR #500, opened 2024-12-20
feat(gax): expose ItemPaginator for paging over items

codyoss:
Fixes: #297


generator/internal/language/rust.go
```diff
@@ -1046,12 +1047,30 @@ func (c *RustCodec) Validate(api *api.API) error {
...
-	HasFeatures bool
-	Features    []string
+	HasFeatures       bool
+	Features          []string
+	RemoveTypeWrapper mustache.LambdaFunc
+}
+
+// removeTypeWrapper removes a container type around another. For example it
+// would transform `Vec<String>` to `String`.
+func removeTypeWrapper(text string, render mustache.RenderFunc) (string, error) {
```
codyoss, line 1057: Lmk if you don't like this approach. This seemed like an easier way to modify information the codec already spits out.

codyoss: This also might be a way to register some global cross language modifiers in the future...

coryan: See below, but maybe you have some ideas for future changes?

generator/internal/language/templates/rust/common/message.mustache
```diff
@@ -67,6 +67,14 @@ impl {{Name}} {
 
 #[cfg(feature = "unstable-stream")]
 impl gax::paginator::PageableResponse for {{Name}} {
+    {{#PageableItem}}
+    type PageItem = {{#RemoveTypeWrapper}}{{{FieldType}}}{{/RemoveTypeWrapper}};
```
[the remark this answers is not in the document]
codyoss: Sounds good, I will update the PR and go with your approach. I did not know if this was a general enough thing that we should add a language agnostic thing for this. 

In the future we should come back and audit some of these types of things. Maybe fieldtype should always be the primitive thing for instance and we have an additional repeated flag that signals we should wrap it in a vec for the message.... just thinking out loud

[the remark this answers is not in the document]
codyoss: done

coryan (approved):
I thought I had approved this yesterday. :shipit:

PR #504, opened 2024-12-20
feat(longrunning): generate library

coryan:
Also get the top-level .sidekick.toml file to use this library wherever it is needed.


coryan:
I will merge #508 and #509 first, but this is read for review if you skip those commits.

codyoss (approved):

PR #506, opened 2024-12-20
chore: update googleapis circa 2024-12-20

coryan:


src/gax/src/error/rpc/mod.rs
```diff
@@ -453,13 +453,15 @@ mod test {
...
+        // TODO(#505) - change this test so it does not require updates as new
+        //     fields appear.
         let got = serde_json::to_value(&status).unwrap();
         let want = json!({
             "code": 12,
             "message": "test",
             "status": "UNIMPLEMENTED",
             "details": [
-                {"@type": "google.rpc.BadRequest", "fieldViolations": [{"field": "field", "description": "desc"}]},
+                {"@type": "google.rpc.BadRequest", "fieldViolations": [{"field": "field", "description": "desc", "reason": "", "localizedMessage": null}]},
```
dbolduc, line 464: Aside: I am guessing `Option<T>` should be annotated with:

```rs
#[serde(skip_serializing_if = "Option::is_none")]
```

so it doesn't show up as `null`.

dbolduc (approved):

src/gax/src/error/rpc/mod.rs
coryan: #507

PR #515, opened 2024-12-22
feat(gax): per-attempt timeouts

### LINKED ISSUE #435: Support timeouts on `gax::http_client::ReqwestClient`
We need RPCs to have a configurable (maybe optional) timeout.

- The generated code should be able to pass the default timeout value to the `ReqwestClient`.
- The application should be able to override (on each request) whatever defaults the generated code has.

coryan:
Support setting per-attempt timeouts on the request builders, and use
these timeouts when sending the RPC.

Fixes #435

### DIFF
--- a/src/gax/Cargo.toml
+++ b/src/gax/Cargo.toml
@@ -35,6 +35,7 @@ serde       = "1.0.214"
 serde_json  = "1.0.133"
 serde_with  = { version = "3.11.0", default-features = false, features = ["base64", "macros"] }
 thiserror   = "2.0.3"
+tokio       = { version = "1.42", features = ["time"] }
 auth        = { version = "0.1.0", path = "../../auth", package = "google-cloud-auth", optional = true }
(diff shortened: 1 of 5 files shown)


src/gax/Cargo.toml
```diff
@@ -35,6 +35,7 @@ serde       = "1.0.214"
 serde_json  = "1.0.133"
 serde_with  = { version = "3.11.0", default-features = false, features = ["base64", "macros"] }
 thiserror   = "2.0.3"
+tokio       = { version = "1.42", features = ["time"] }
```
codyoss, line 38: I think this is only need for tests currently. Can we keep this out of the main deps? Maybe we are already tightly coupled to tokio with reqwest -- but if we aren't and it is not too much trouble it may be good to try to stay open to other runtimes.

codyoss (approved):

src/gax/Cargo.toml
coryan: Fixed.

PR #525, opened 2024-12-22
test(gax): add missing tests

coryan:


src/gax/src/error/rpc/mod.rs
```diff
@@ -614,9 +615,62 @@ mod test {
...
+        let want = Code::try_from(input).unwrap();
+        let serialized = serde_json::to_value(&want)?;
+        let got = serde_json::from_value::<Code>(serialized)?;
+        assert_eq!(got, want);
+        Ok(())
+    }
+
+    #[test]
+    fn code_try_from_string_error() {
+        let err = Code::try_from("INVALID-NOT-A-CODE");
```
coryan, line 652: I wonder if this shoudl return `Code::Unknown` as we do for integer conversion.

codyoss (approved):

Issue #437, opened 2024-12-16
Support retry loops on `gax::http_client::ReqwestClient`


coryan:
We need RPCs to have a configurable (maybe optional) retry loop.

- The generated code should be able to provide a default retry policy (what is retryable? for how long / how many times?) and backoff policy (how long to backoff for?)
- The application should be able to override (on each request) the default retry policy and/or default backoff policies.

PR #537, opened 2024-12-24
feat(gax): introduce retry policies

coryan:
These policies will be used to control (1) what errors are retryable,
and (2) how many retry attempts we make. The backoff policy (how long
to wait between retries) is orthogonal and will happen on a separate
PR.

The policies can be composed, so customers could override the error
messages that need retry, and reuse the deadline or error count
policies.

Customers can provide their own policies, for example, they may
introduce policies where some kind of error is tolerated more than
others.

Budgets (how many retry requests are sent at a time), are also
orthogonal and the subject of future PRs.

Part of the work for #437

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -630,6 +630,7 @@ dependencies = [
  "gcp-sdk-wkt",
  "google-cloud-auth",
  "http",
+ "mockall",
  "pin-project",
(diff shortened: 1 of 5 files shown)


coryan:
Closing for now.  I need to rewrite some things.

PR #538, opened 2024-12-24
feat(generator/rust): require path parameters

### LINKED ISSUE #536: Make path parameters required for RPC construction
It would be nice if we could also read the field_behavior annotation for but because it is considered backwards compatible to remove a required annotation this is not possible. See https://google.aip.dev/203#backwards-compatibility for more details.

Related: https://github.com/googleapis/google-cloud-rust/discussions/460

coryan:
Change the generated code to require any arguments that will be part of
the path parameters. These include optional fields (such as objects)
where it is a sub-field of the parameter that goes into the path
parameter.

Part of the work for #536

### DIFF
--- a/generator/testdata/rust/gclient/golden/iam/v1/src/client.rs
+++ b/generator/testdata/rust/gclient/golden/iam/v1/src/client.rs
@@ -97,15 +97,33 @@ impl IAMPolicy {
     /// existing policy.
     ///
     /// Can return `NOT_FOUND`, `INVALID_ARGUMENT`, and `PERMISSION_DENIED` errors.
-    pub fn set_iam_policy(&self) -> crate::builders::SetIamPolicy {
+    pub fn set_iam_policy<
+        IntoResource,
+    >(
+        &self,
+        resource: IntoResource,
+    ) -> crate::builders::SetIamPolicy
+    where
+        IntoResource: Into<String>,
+    {
         crate::builders::SetIamPolicy::new(self.inner.clone())
+            .set_resource ( resource.into() )
     }
 
     /// Gets the access control policy for a resource.
     /// Returns an empty policy if the resource exists and does not have a policy
     /// set.
-    pub fn get_iam_policy(&self) -> crate::builders::GetIamPolicy {
+    pub fn get_iam_policy<
+        IntoResource,
+    >(
+        &self,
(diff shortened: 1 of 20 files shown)


generator/testdata/rust/gclient/golden/iam/v1/src/client.rs
```diff
@@ -97,15 +97,33 @@ impl IAMPolicy {
     /// existing policy.
     ///
     /// Can return `NOT_FOUND`, `INVALID_ARGUMENT`, and `PERMISSION_DENIED` errors.
-    pub fn set_iam_policy(&self) -> crate::builders::SetIamPolicy {
+    pub fn set_iam_policy<
+        IntoResource,
+    >(
+        &self,
+        resource: IntoResource,
```
codyoss, line 104: This signature is a little hard to read imo with the indirection of the where clause. How would you fell about:

```
pub fn set_iam_policy(&self, resource: impl Into<String>) -> crate::builders::SetIamPolicy { ... }
```

src/integration-tests/src/secret_manager/openapi_locational.rs
```diff
@@ -212,11 +189,12 @@ async fn run_secret_versions(
 
     println!("\nTesting get_secret_version_by_project_and_location_and_secret_and_version()");
     let get = client
-        .get_secret_version_by_project_and_location_and_secret_and_version()
-        .set_project(project_id)
-        .set_location(location_id)
-        .set_secret(secret_id)
-        .set_version(version_id)
+        .get_secret_version_by_project_and_location_and_secret_and_version(
```
codyoss, line 192: nice, I like this a lot better!

codyoss (approved):
LGTM. Left some food for thought on the method signatures.

generator/testdata/rust/gclient/golden/iam/v1/src/client.rs
coryan: Done.

src/integration-tests/src/secret_manager/openapi_locational.rs
coryan: Most client methods are fine.  It is less than ideal that some require fields are not required by the method...

PR #540, opened 2024-12-25
impl(sidekick): Implemented command pattern for CLI actions and flags

### LINKED ISSUE #393: `sidekick` commands should have help messages
As the number of commands grows we may need some help to tell what commands do what.

----
Can we file a bug to add help text for these different commands?

_Originally posted by @codyoss in https://github.com/googleapis/google-cloud-rust/pull/391#discussion_r1878328859_

dangazineu:
This resolves #393 by structuring commands and their flags under an encapsulated data structure. 

Commands are hierarchical, and inherit flags from their parents. Flag parsing happens only once, within the context of the command that matches the arguments in the command line. If no exact match is found, flags are not parsed, and the program exits with an error with a short message explaining how to get help for the closest command.

i.e: 
- `sidekick bad param` exits with a message explaining how to run `sidekick help`. 
- `sidekick generate bad param`  exits with a message explaining how to run `sidekick help generate`. 

The same behavior applies for the `help` command itself:
- `sidekick help bad param` exits with a message explaining how to run `sidekick help`. 
- `sidekick help generate bad param`  exits with a message explaining how to run `sidekick help generate`. 

The main functional difference is that now the full command path needs to be included **before** any flags (i.e: `sidekick -project-root .. generate` becomes `sidekick generate -project-root ..`). That is because the command needs to be identified before the flags are parsed, and only the remaining arguments after the last matching command are given to the flag parser. 

Things to consider for the future:
- I didn't want to restructure the existing commands as part of this change. As all the generator commands (`generate`, `refresh` and `refresh-all`) share the same flags, I have registered all the original flags directly at the `sidekick` command, so every command will inherit them. It might make sense to group these commands together, under the same parent.
- `cmdline.go` implementation is pretty light after this change, and because flags are parsed directly under the variable that they are assigned to, it may not make sense to continue using the `CommandLine` struct anymore. 
- `package sidekick` is getting crowded and may benefit from some refactoring, I have run into a couple name clashes already (had to change the original `runCommand` to `execCommand`).


generator/internal/sidekick/sidekick.go
```diff
@@ -17,57 +17,94 @@ package sidekick
...
-	cmdLine, err := parseArgs()
-	if err != nil {
-		return err
+	if len(os.Args) < 2 {
+		_ = CmdSidekick.PrintUsage()
+		return fmt.Errorf("no command given")
 	}
-	return runSidekick(cmdLine)
+
+	return runSidekick(os.Args[1:])
```
julieqiu, line 61: pass these into the Run function 
```
func Run(arg ...string) error { 
  ...
}
```

generator/internal/sidekick/command.go
```diff
@@ -0,0 +1,256 @@
...
+	if parent == nil {
+		panic("command must have a parent")
+	}
+	c := newCommand(usageLine, shortDescription, parent)
+	c.longDescription = longDescription
+	c.action = action
+	return c
+}
+
+func newCommand(usageLine string, shortDescription string, parent *Command) *Command {
```
coryan, line 201: Maybe "newRootCommand" is a better name?  Having two functions (`newCommand` and `NewCommand`) seems too subtle, and making the function private does not have much of an effect, as all the commands are in the same package.

generator/internal/sidekick/sidekick.go
```diff
@@ -17,57 +17,94 @@ package sidekick
...
-	"maps"
 	"os"
+	"strings"
+	//"maps"
+	//"os"
 )
 
+var CmdSidekick = newCommand(
+	"sidekick",
+	"sidekick is a tool for automating code generation.",
```
coryan, line 28: s/automating code generation/automating SDK generation/ I am aiming for more than just generating the code, the goal is to generate the SDK (code, examples, documentation, README, package files, etc. etc.)

coryan (commented):
I think this is an improvement on what we have, and I would gladly approve it (modulo some small nits and changes listed below).

The package-globals give me pause, and maybe we should consult some golang expert (@julieqiu  or @codyoss ) in case there is a cleaner solution.

generator/internal/sidekick/config.go
```diff
@@ -37,6 +38,27 @@ type GeneralConfig struct {
 	ServiceConfig       string `toml:"service-config,omitempty"`
 }
 
+func loadConfig(cmdLine *CommandLine) (*Config, error) {
+	// Load the top-level configuration file. If there are any errors loading
+	// the file just run with the defaults.
```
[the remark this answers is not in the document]
dangazineu: made this a bit more descriptive, since the function is doing quite a bit

.github/workflows/sdk.yaml
```diff
@@ -121,7 +121,7 @@ jobs:
           protoc --version
       - name: Regenerate all the code
         working-directory: generator
-        run: go run ./cmd/sidekick -project-root .. refreshall
+        run: go run ./cmd/sidekick refreshall -project-root ..
```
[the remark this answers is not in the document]
dangazineu: my bad, I completely forgot those, updated now

generator/internal/sidekick/command.go
dangazineu: Yeah, I agree there is no benefit in making it private. I do think that as the app grows, it would make sense to move each command to its own package (like the Go CLI does). And in that case only `CmdSidekick` would have access to the private constructor. 

I didn't name it `newRootCommand` because I reuse the function for every command that is created, but have additional checks and logic in the exported function. I don't think it would be too bad to merge these two functions together and hardcode a check that only allows a command named "sidekick" to have no parent.

generator/internal/sidekick/cmdline.go
```diff
@@ -35,72 +28,14 @@ type CommandLine struct {
...
-		SpecificationSource: *source,
-		ServiceConfig:       *serviceConfig,
-		Source:              sourceOpts,
-		Language:            *language,
-		Output:              *output,
-		Codec:               codecOpts,
-		DryRun:              *dryrun,
-	}, nil
-}
+var (
```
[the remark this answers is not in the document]
dangazineu: I didn't want to make too many changes at once, but I agree that this is weird and worth refactoring. As it stands right now, `CommandLine` isn't adding much value.

I don't like the globals either, but in the Go CLI, they followed a similar structure, and just named the flags more explicitly for global use (i.e: `FlagProjectRoot`, `FlagGenerateDryRun`, etc). We won't even be able to rely on the package visibility: if the app grows to a point where commands get their own packages, we'll have to export these flags so they are accessible to the commands that need them.

We could switch our flag declaration from the pointer constructor to the one that allows for encapsulated parsing into local variables (as it was implemented before). But I think that will require more boilerplate code if we want to reuse flag declaration across multiple commands.

generator/internal/sidekick/command.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+	c.parent = parent
+
+	c.flags.Usage = func() {
+		_ = c.PrintUsage()
+	}
+	return c
+}
+
+var usageTemplate = `
```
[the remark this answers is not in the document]
dangazineu: TBH, this didn't occur to me as I was doing it, but now that you mentioned I quite like the idea. I mostly just reverse engineered it from the Go implementation, but I think I have simplified the logic to a point that it may make more sense to just write the code for it rather than use a template (especially because we only have a single template). I'll give it a try and report back here.

generator/internal/sidekick/cmdline.go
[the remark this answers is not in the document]
coryan: SGTM.  If we go with the code (largely) as-is: please open an issue to clean up this bit around the globals  later.

generator/internal/sidekick/command.go
```diff
@@ -0,0 +1,256 @@
...
+	if i := strings.LastIndex(name, " "); i >= 0 {
+		name = name[i+1:]
+	}
+	return name
+}
+
+func (c *Command) Names() []string {
+	names := []string{c.Name()}
+	names = append(names, c.altNames...)
+	return names
```
julieqiu, line 81: return append([]string{c.Name()}, c.altNames...)

generator/internal/sidekick/command.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+func (c *Command) Runnable() bool {
```
julieqiu, line 84: It doesn't look like this is used anywhere

generator/internal/sidekick/command.go
```diff
@@ -0,0 +1,256 @@
...
+	"os"
+	"slices"
+	"strings"
+	"text/template"
+)
+
+// A Command is an implementation of a sidekick command
+// like sidekick generate
+type Command struct {
+	action           func(rootConfig *Config, cmdLine *CommandLine) error
```
julieqiu, line 29: nit: I normally see this in Go named as `run`

generator/internal/sidekick/command.go
```diff
@@ -0,0 +1,256 @@
...
+	usageLine        string
+	altNames         []string
+	shortDescription string
+	longDescription  string
+	flags            *flag.FlagSet
+	commands         []*Command
+	parent           *Command
+}
+
+func (c *Command) UsageLine() string {
```
julieqiu, line 39: what's the reason for introducing this function? It looks like this is only ever called within this package, so calling c.usageLine directly should be sufficient

generator/internal/sidekick/command.go
```diff
@@ -0,0 +1,256 @@
...
+	"fmt"
+	"os"
+	"slices"
+	"strings"
+	"text/template"
+)
+
+// A Command is an implementation of a sidekick command
+// like sidekick generate
+type Command struct {
```
julieqiu, line 28: It looks like `Command` is only ever used inside the internal/sidekick package, so I would unexport it

generator/internal/sidekick/command.go
```diff
@@ -0,0 +1,256 @@
...
+
+func (c *Command) LongDescription() string {
+	if c.longDescription != "" {
+		return c.longDescription
+	} else {
+		return c.shortDescription
+	}
+}
+
+func (c *Command) ShortDescription() string {
```
julieqiu, line 51: same comment as above re: UsageLine - I think this is only ever called within this package and c.shortDescription can be called directly

In general, I would only make things exported if they are used outside the package

generator/internal/sidekick/sidekick.go
```diff
@@ -17,57 +17,88 @@ package sidekick
...
-		},
-		Source: maps.Clone(cmdLine.Source),
-		Codec:  maps.Clone(cmdLine.Codec),
-	}
-	config, err := mergeConfigs(rootConfig, argsConfig)
-	if err != nil {
-		return err
+}
+
+func NotFoundError(bestMatch *Command, allArgs []string, unusedArgs []string, msg string) error {
```
julieqiu, line 79: maybe `constructNotFoundError`?

generator/internal/sidekick/generate.go
```diff
@@ -24,6 +24,19 @@ import (
 	toml "github.com/pelletier/go-toml/v2"
 )
 
+var CmdGenerate = NewCommand(
```
julieqiu, line 27: I would construct these within an `init` command:
https://go.dev/doc/effective_go#init

generator/internal/sidekick/command.go
dangazineu: The command struct already has a `run` function that calls this after some validation.

generator/internal/sidekick/sidekick_test.go
```diff
@@ -44,7 +44,7 @@ var (
 func TestRustFromOpenAPI(t *testing.T) {
 	var outDir = fmt.Sprintf("%s/rust/openapi/golden", testdataDir)
 	cmdLine := &CommandLine{
-		Command:             "generate",
+		Command:             []string{},
```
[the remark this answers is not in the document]
dangazineu: Because with this refactor, the CommandLine construction now happens after the command path is interpreted, and its value is ignored during the command execution -- but available, if a command needs it. 
The easiest path to test behavior is what we are doing here, passing a CommandLine struct directly to the runCommand function. Alternatively, we could pass an array of strings to `sidekick.Run` function, but that is more verbose.

generator/internal/sidekick/command.go
dangazineu: I've merged the two constructors into one, and added a check to only allow the "sidekick" command to have no parent. I think the result is much cleaner now.

generator/internal/sidekick/command.go
dangazineu: removed

generator/internal/sidekick/command.go
dangazineu: I was trying to encapsulate field ownership to prevent unwanted change, but I get that at most what I get it package private, which doesn't make much difference. Removed the redundant functions.

generator/internal/sidekick/command.go
dangazineu: Made Command and all its fields and functions into package private

generator/internal/sidekick/sidekick.go
dangazineu: used newNotFoundError, just to stay consistent with the newCommand function.

generator/internal/sidekick/generate.go
dangazineu: I moved all commands to init functions within their source files. The only command I kept out of an init function was cmdSidekick, since that one needs to be  visible within the package, and I didn't see a reason to declare it outside the init function, and construct it within the init function. Let me know if you think that would make sense to change too.

generator/internal/sidekick/sidekick.go
```diff
@@ -17,57 +17,95 @@ package sidekick
...
+	"sidekick is a tool for automating SDK generation.",
+	``,
+	nil, // nil parent is only allowed for the root command
+	nil).
+	addFlagString(&flagProjectRoot, "project-root", "", "the root of the output project").
+	addFlagString(&format, "specification-format", "", "the specification format. Protobuf or OpenAPI v3.").
+	addFlagString(&source, "specification-source", "", "the path to the input data").
+	addFlagString(&serviceConfig, "service-config", "", "path to service config").
+	addFlagString(&output, "output", "", "the path within project-root to put generated files").
+	addFlagString(&flagLanguage, "language", "rust", "the generated language").
```
dangazineu, line 35: I made `rust` the default value for `-language` flag because unparam wasn't happy with this value being always empty. I couldn't manage to fix it with `//nolint:unparam`, and didn't think it was worth removing the parameter just to make it happy. For the time being, `rust` is probably the right value for this flag anyways :)

generator/internal/sidekick/command.go
```diff
@@ -0,0 +1,240 @@
...
+	}
+
+	c.flags.Usage = func() {
+		c.printUsage()
+	}
+
+	return c
+}
+
+// printUsage prints the usage of the command to os.Stdout, following the same logic as the usageTemplate, but using standard fmt.Println statements instead.
```
coryan: nit: that comment line is longer than usual, consider breaking it.

generator/internal/sidekick/command.go
```diff
@@ -0,0 +1,240 @@
...
+
+	// first prints the entire Long Description for the given command
+	if len(c.longDescription) > 0 {
+		fmt.Println(c.longDescription)
+	} else {
+		fmt.Println(c.shortDescription)
+	}
+	fmt.Printf("\n")
+
+	// Then prints the usage line, skipping over <command> if there are no sub-commands, and [flags] if there are no flags
```
coryan, line 200: nit: some of the comments here start with uppercase, and some are not.  I think generally we have been using full sentences, with a period and everything.

generator/internal/sidekick/sidekick.go
```diff
@@ -17,57 +17,95 @@ package sidekick
...
+		msg,
+		validHelp)
+}
+
+func runCommand(cmd *command, cmdLine *CommandLine) error {
 	if cmdLine.ProjectRoot != "" {
 		cwd, err := os.Getwd()
 		if err != nil {
-			return err
+			return fmt.Errorf("could not get current working directory: %v", err)
```
coryan, line 96: I was told `%w` is how one wraps errors:

```suggestion
			return fmt.Errorf("could not get current working directory: %w", err)
```

You need to fix this in multiple places.

https://stackoverflow.com/questions/61283248/format-errors-in-go-s-v-or-w

generator/internal/sidekick/sidekick.go
```diff
@@ -17,57 +17,95 @@ package sidekick
...
 		}
-		defer os.Chdir(cwd)
-		if err := os.Chdir(cmdLine.ProjectRoot); err != nil {
-			return err
+		defer func(dir string) {
+			_ = os.Chdir(dir)
+		}(cwd)
+		if err = os.Chdir(cmdLine.ProjectRoot); err != nil {
+			return fmt.Errorf("could not change to project root [%s]: %v", cmdLine.ProjectRoot, err)
```
coryan, line 102: %w

generator/internal/sidekick/sidekick.go
```diff
@@ -17,57 +17,95 @@ package sidekick
...
-		General: GeneralConfig{
-			Language: cmdLine.Language,
-		},
-		Source: maps.Clone(cmdLine.Source),
-		Codec:  maps.Clone(cmdLine.Codec),
-	}
-	config, err := mergeConfigs(rootConfig, argsConfig)
-	if err != nil {
-		return err
+		return fmt.Errorf("could not load configuration: %v", err)
```
coryan, line 107: %w

generator/internal/sidekick/sidekick_test.go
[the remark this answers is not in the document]
coryan: Ack.  I think my question was less interesting: I am seeing other tests in this file use `Command: []string{"generate"}` and wonder why this test does not do the same.  Should they all be simplified like this test, or should they all use the longer incantation?

If there is a reason why this test should be different, then that deserves a comment.

coryan (approved):
A few nits and minor things, please fix and then merge.

generator/internal/sidekick/sidekick_test.go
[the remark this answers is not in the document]
dangazineu: Oh, I see... I missed the other tests in this file. The cmd name in the args is not useful in any of them, so it's just ignored. I have removed this value from the other tests too, so they're all following the same pattern.

generator/internal/sidekick/command.go
dangazineu: I removed some of these as they're not that useful anymore. I had carried them over from the previous template-based implementation. The new impl is easy enough to follow without comments.

generator/internal/sidekick/sidekick.go
dangazineu: Thanks! I didn't know that. Fixed.

generator/internal/sidekick/generate.go
julieqiu: sgtm - we can always move things around as a follow up

generator/internal/sidekick/command.go
julieqiu: I don't think I see the need for the `run` function, so I would recommend deleting that and renaming this to run. If you think it is necessary though feel free to leave it, and we can always change things in a follow up.

julieqiu (approved):

Issue #389, opened 2024-12-10
More detailed errors in `*-auth` crate


coryan:
This is a placeholder to reference in a TODO.  I expect that we will want more specific errors from the client library as it grows.

dbolduc:
We should revisit the error handling in `user_credential.rs` with higher scrutiny. The I/O errors should probably be retried according to: https://github.com/googleapis/google-cloud-rust/pull/489#pullrequestreview-2518327781

dbolduc:
This issue was about expanding an `ErrorKind`-like enum. We got rid of the enum.

PR #541, opened 2024-12-26
impl(auth): fetch service account info from mds

harkamaljot:
Part of the work for https://github.com/googleapis/google-cloud-rust/issues/442

Add an implementation for fetching service account info from metadata service. It is a required step before MDS credentials can be fetched for a given account.

Error handling
I am aware of https://github.com/googleapis/google-cloud-rust/issues/389. We should probably start tracking error kinds. I didn't do it in this PR.

Unit tests
I made a fake server similar to https://github.com/googleapis/google-cloud-rust/pull/489 and tested the cases for fetching service account info for both valid and invalid cases.

Integration tests?
There are none. I did manually tested the cloud on my cloudtop vm and verified I was able to fetch back service account info.


src/auth/src/credentials/mds_credential.rs
```diff
@@ -0,0 +1,364 @@
...
+use reqwest::Client;
+use std::collections::HashMap;
+use std::env;
+
+const METADATA_FLAVOR_VALUE: &str = "Google";
+const METADATA_FLAVOR: &str = "metadata-flavor";
+
+lazy_static! {
+    // Use lazy_static to initialize the metadata URLs.
+    static ref _METADATA_ROOT: String = format!(
```
coryan, line 32: Why a learning `_` in this variable name?

src/auth/src/credentials/mds_credential.rs
```diff
@@ -0,0 +1,364 @@
...
+            .headers(headers)
+            .send()
+            .await
+            .map_err(|e| CredentialError::new(false, e.into()))?;
+
+        response
+            .json::<ServiceAccountInfo>()
+            .await
+            .map_err(|e| CredentialError::new(false, e.into()))
+    }
```
coryan, line 116: All of this seems unrelated to the initial goal of getting a working MDS token.

harkamaljot: for fetching an mds token, first we need to fetch service account info specifically email and scopes, so therefore this call needs to happen first. As this was my first rust change request, so I wanted to keep PR requests small.

src/auth/src/credentials/mds_credential.rs
harkamaljot: Initially I labelled it thinking of it as a private variable like in python, but I realized that is invalid in rust, so updated it.

src/auth/src/credentials/mds_credential.rs
coryan: I don't know what you mean.  This works:

curl -H"metadata-flavor:Google" http://metadata.google.internal./computeMetadata/v1/instance/service-accounts/default/token

We do not need the email and/or scopes. Whatever are the defaults are good enough for us.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -0,0 +1,332 @@
...
+    use crate::token::test::MockTokenProvider;
+    use axum::response::IntoResponse;
+    use reqwest::StatusCode;
+    use serde_json::Value;
+    use tokio::task::JoinHandle;
+
+    #[test]
+    fn metadata_root_default() {
+        env::remove_var("GCE_METADATA_ROOT"); // Ensure default is used
+        env::remove_var("GCE_METADATA_HOST");
```
coryan, line 138: `std::env::remove_var` is `unsafe`:

https://doc.rust-lang.org/std/env/fn.remove_var.html

You need to make sure the test is serialized, and we should call those functions form `unsafe` blocks. For inspiration see:

https://github.com/googleapis/google-cloud-rust/blob/30257cf2dc4ffa12a4dc5015457745f890534a9d/src/gax/src/http_client/mod.rs#L162-L169 


PS: I think we must be missing a trick in your CI if we are allowing unsafe functions to get called.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -0,0 +1,332 @@
...
+
+    fn handle_token_factory(
+        response_code: StatusCode,
+        response_headers: HeaderMap,
+        response_body: Value,
+    ) -> impl IntoResponse {
+        (response_code, response_headers, response_body.to_string()).into_response()
+    }
+
+    // Starts a server running locally. Returns an (endpoint, path, handler) pair.
```
coryan, line 246:
```suggestion
    // Starts a server running locally. Returns an (endpoint, path, server) tuple.
```

coryan: Also, it seems like you never use `path`, so maybe just remove it?

src/auth/src/credentials/mds_credential.rs
```diff
@@ -0,0 +1,332 @@
...
+}
+
+#[allow(dead_code)] // TODO(#442) - implementation in progress
+#[derive(serde::Deserialize, serde::Serialize, PartialEq, Debug, Clone)]
+struct ServiceAccountInfo {
+    email: String,
+    scopes: Option<Vec<String>>,
+    aliases: Option<Vec<String>>,
+}
+#[allow(dead_code)] // TODO(#442) - implementation in progress
```
coryan, line 76: nit: insert blank line.

coryan (changes requested):
Please fix the stuff around unsafe functions. 

Since you already wrote the code for service accounts, just leave it there.  And then try to focus on the absolute minimal stuff to get a working token with the default SA and the default scopes.  We are not going to use anything else for a long time.

src/auth/src/credentials/mds_credential.rs
harkamaljot: Right now, I am following the python code here: https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/compute_engine/credentials.py#L126 where I see first we fetch service account email and then we make the call to mds server with the email. The default call is also works, but will it be okay if we can review it as given I already have the pr out for it?

src/auth/src/credentials/mds_credential.rs
coryan: Aha, these functions are not marked unsafe until the 2024 edition of rust.  They **are** unsafe already, just the marker is missing.  That explains why the code compiled.

src/auth/src/credentials/mds_credential.rs
harkamaljot: Sounds good. I read your later comment, let me send another pull request for bare minimum for mds creds.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -0,0 +1,330 @@
...
+const METADATA_FLAVOR: &str = "metadata-flavor";
+#[allow(dead_code)] // TODO(#442) - implementation in progress
+static METADATA_ROOT: LazyLock<String> = LazyLock::new(|| {
+    format!(
+        "http://{}/computeMetadata/v1",
+        env::var("GCE_METADATA_HOST").unwrap_or_else(|_| {
+            env::var("GCE_METADATA_ROOT").unwrap_or_else(|_| "metadata.google.internal".to_string())
+        })
+    )
+});
```
coryan, line 37: This function seems complex enough to merit testing. 

I wonder if we need it though, are these environment variables used for anything other than testing?  If not, can we make the configuration-via-environment-variables go away?

If we do need this behavior, then (1) move the closure to a function, (2) test all the interesting cases for the function, not just the one where both values are not set.

coryan (changes requested):

src/auth/src/credentials/mds_credential.rs
harkamaljot: basically python library uses these env variables to pick up the address of the metadata server, which I assumed can be different from the default value see: https://googleapis.dev/python/google-auth/latest/reference/google.auth.environment_vars.html  and here's the corresponding code: https://github.com/googleapis/google-auth-library-python/blob/main/google/auth/compute_engine/_metadata.py#L45. Given it is just the MVP, I think the better approach would be remove these env variables and add it later when we need them.

coryan (approved):

PR #546, opened 2024-12-30
impl(auth): Implement fetching token from mds server

harkamaljot:
Part of the work for https://github.com/googleapis/google-cloud-rust/issues/442

Add an implementation for fetching token from mds server

Error handling
I am aware of https://github.com/googleapis/google-cloud-rust/issues/389. We should probably start tracking error kinds. I didn't do it in this PR.

Unit tests

Integration tests?
There are none. I did manually tested the cloud on my gce vm and verified I was able to fetch back token

### DIFF
--- a/src/auth/src/credentials/mds_credential.rs
+++ b/src/auth/src/credentials/mds_credential.rs
@@ -121,28 +121,28 @@ impl MDSAccessTokenProvider {
 #[allow(dead_code)]
 impl TokenProvider for MDSAccessTokenProvider {
     async fn get_token(&mut self) -> Result<Token> {
-        let request = Client::new();
-        let path: String = format!("{}/instance/service-accounts/default/token", self.endpoint);
-        let mut headers = HeaderMap::new();
-        headers.insert(
-            METADATA_FLAVOR,
-            HeaderValue::from_static(METADATA_FLAVOR_VALUE),
-        );
-        let url = reqwest::Url::parse(path.as_str())
-            .map_err(|e| CredentialError::new(false, e.into()))?;
+        let client = Client::new();
+        let request = client
+            .get(format!(
+                "{}/instance/service-accounts/default/token",
+                self.endpoint
+            ))
+            .header(
+                METADATA_FLAVOR,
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/mds_credential.rs
```diff
@@ -111,7 +121,46 @@ impl MDSAccessTokenProvider {
...
+                .map_err(|e| CredentialError::new(false, e.into()))?;
+            return Err(CredentialError::new(
+                is_retryable(status),
+                Box::from(format!("Failed to fetch token. {body}")),
+            ));
+        }
+        let response = response
+            .json::<MDSTokenResponse>()
+            .await
+            .map_err(|e| CredentialError::new(false, e.into()))?;
```
coryan, line 154: This is an I/O error *after* reading the HTTP status code.  That seems like a retryable error to me.  MDS is there, it responds, it wants to send data in the right path, the network just blipped before we could get all the data.

coryan (changes requested):

src/auth/src/credentials/mds_credential.rs
harkamaljot: I think this can be both a data format issue and an I/O issue here. However, most likely issue will be still I/O bound, so updating it to true here.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -65,9 +67,17 @@ struct ServiceAccountInfo {
     aliases: Option<Vec<String>>,
 }
 
+#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
```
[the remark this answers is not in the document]
harkamaljot: I tried using serde_json::json!() however that doesn't work directly as I would have manually create a json like this
```
json!({
        "field1": my_struct.field1,
        "field2": my_struct.field2,
    });
```
or I can also do a mapping like this
```
and_then(|text| serde_json::from_str::<Value>(&text)
```
however both of them seem inconvenient compared to serde::Serialize, so keeping the trait here.

coryan (approved):

Issue #423, opened 2024-12-13
Consider more from/tryfrom conversions for Duration and Timestamp


coryan:
We may want to consider conversions from and to the [chrono](https://crates.io/crates/chrono), maybe direct mappings to and from `tokio::time`, and (as [suggested](https://github.com/googleapis/google-cloud-rust/pull/422#discussion_r1884254470) by @codyoss) to and from strings.

mpeddada1:
> maybe direct mappings to and from `tokio::time`

@coryan Adding TryFrom conversion between `tokio::time` and `Duration` results in a conflict with [this function](https://github.com/googleapis/google-cloud-rust/blob/88109cb6fbc0ddbfd5a35d7c2ea6c5587585a5c8/src/wkt/src/duration.rs#L225) with the following error message: `conflicting implementations of trait TryFrom<std::time::Duration> for type duration::Duration`. 
Since `tokio::time::Duration` is re-export of `std::time::Duration`, I wonder if having a separate conversion might be redundant?

coryan:
> Since `tokio::time::Duration` is re-export of `std::time::Duration`, I wonder if having a separate conversion might be redundant?

Agree.

mpeddada1:
Thanks for confirming! Closing this issue for now given the implementation of from/tryfrom conversions for `chrono` and String representations in https://github.com/googleapis/google-cloud-rust/pull/542 and https://github.com/googleapis/google-cloud-rust/pull/550.

coryan:
The bug says "Duration **and** Timestamp".  We do not have anything for [chrono::DateTime<Utc>](https://docs.rs/chrono/latest/chrono/struct.DateTime.html) do we?

mpeddada1:
Ah my bad, missed this. Let me put up a PR for the Timestamp conversions as well.

PR #550, opened 2024-12-30
feat: add try/from trait for conversion between chrono and Duration

mpeddada1:
For #423.

Add TryFrom trait for conversion between [chrono::time](https://crates.io/crates/chrono) and `Duration`.

### DIFF
--- a/src/wkt/Cargo.toml
+++ b/src/wkt/Cargo.toml
@@ -39,4 +39,4 @@ bytes      = { version = "1.8.0", features = ["serde"] }
 [dev-dependencies]
 test-case = "3.3.1"
 bytes     = { version = "1.8.0", features = ["serde"] }
-wkt       = { path = ".", package = "gcp-sdk-wkt", features = ["time", "chrono"] }
+wkt       = { path = ".", package = "gcp-sdk-wkt", features = ["chrono", "time"] }
--- a/src/wkt/Cargo.toml
+++ b/src/wkt/Cargo.toml
@@ -24,8 +24,8 @@ keywords.workspace   = true
 categories.workspace = true
 
 [features]
-time = []
 chrono = []
+time = []
 
 [dependencies]
 serde      = { version = "1.0.214", features = ["serde_derive"] }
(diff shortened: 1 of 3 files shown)


src/wkt/src/duration.rs
```diff
@@ -272,6 +272,28 @@ impl std::convert::From<Duration> for time::Duration {
...
+/// Converts from [Duration] to [chrono::Duration].
+#[cfg(feature = "chrono")]
+impl std::convert::TryFrom<Duration> for chrono::Duration {
+    type Error = DurationError;
+    fn try_from(value: Duration) -> Result<Self, Self::Error> {
+        if value.nanos < 0 {
+            return Err(Error::OutOfRange());
+        }
+        Ok(Self::new(value.seconds(), value.nanos() as u32).unwrap())
+    }
```
coryan, line 294: I think you can safely handle the `value.nanos < 0` case using:

```suggestion
        Ok(Self::seconds(value.seconds()) + Self::nanoseconds(value.nanos()))
    }
```

Which suggests you do not need a `TryFrom` at all, and a simple `From` will do.

src/wkt/Cargo.toml
```diff
@@ -24,17 +24,19 @@ keywords.workspace   = true
...
 [features]
+chrono = []
 time = []
 
 [dependencies]
 serde      = { version = "1.0.214", features = ["serde_derive"] }
 serde_with = { version = "3.11.0", default-features = false, features = ["base64"] }
 serde_json = "1.0.134"
 time       = { version = "0.3.36", features = ["formatting", "parsing"] }
+chrono     = { version = "0.4.39" }
```
coryan, line 35: I think you can say

```suggestion
chrono     = { version = "0.4.39", optional = true }
```

and in the features section:

```
chrono = ["dep:chrono"]
```

coryan (changes requested):

src/wkt/src/duration.rs
mpeddada1: That's a good idea!

src/wkt/Cargo.toml
mpeddada1: Sounds good, made chrono an optional dependency.

coryan (approved):

PR #552, opened 2024-12-30
impl(auth): helper function for ADC path

dbolduc:
Adds a helper function (to be used shortly) that returns the path to look up ADC.

Note that the path changes depending on the OS.

~I do not think our CI exercises the Windows code path. I successfully ran a `cargo test` on my windows machine, if that means anything.~

**EDIT:** Good thing we have Windows coverage thanks to #554.

This is the first use of `scoped_env::ScopedEnv`. I found it helpful: https://docs.rs/scoped-env/latest/scoped_env/index.html

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -15,7 +15,6 @@
 pub(crate) mod mds_credential;
 pub(crate) mod user_credential;
 
-use crate::errors::CredentialError;
 use crate::Result;
 use http::header::{HeaderName, HeaderValue};
 use std::future::Future;
@@ -171,9 +170,9 @@ pub mod traits {
 ///
 /// [AIP-4110]: https://google.aip.dev/auth/4110
 #[allow(dead_code)] // TODO(#442) - implementation in progress
-fn adc_path() -> Result<String> {
+fn adc_path() -> Option<String> {
     if let Ok(e) = std::env::var("GOOGLE_APPLICATION_CREDENTIALS") {
-        return Ok(e);
+        return Some(e);
     }
     adc_well_known_path()
 }
@@ -182,35 +181,28 @@ fn adc_path() -> Result<String> {
 ///
 /// [AIP-4113]: https://google.aip.dev/auth/4113
 #[cfg(target_os = "windows")]
-fn adc_well_known_path() -> Result<String> {
+fn adc_well_known_path() -> Option<String> {
     if let Ok(root) = std::env::var("APPDATA") {
-        return Ok(root + "/gcloud/application_default_credentials.json");
+        return Some(root + "/gcloud/application_default_credentials.json");
     }
-    Err(CredentialError::new(
-        false,
-        Box::from("Error loading ADC: %APPDATA% environment variable not set."),
-    ))
+    None
 }
 
 /// The well-known path to ADC on Linux and Mac, as specified in [AIP-4113].
 ///
 /// [AIP-4113]: https://google.aip.dev/auth/4113
 #[cfg(not(target_os = "windows"))]
-fn adc_well_known_path() -> Result<String> {
+fn adc_well_known_path() -> Option<String> {
     if let Ok(root) = std::env::var("HOME") {
-        return Ok(root + "/.config/gcloud/application_default_credentials.json");
+        return Some(root + "/.config/gcloud/application_default_credentials.json");
     }
-    Err(CredentialError::new(
-        false,
-        Box::from("Error loading ADC: ${HOME} environment variable not set."),
-    ))
+    None
 }
 
 #[cfg(test)]
 mod test {
     use super::*;
     use scoped_env::ScopedEnv;
-    use std::error::Error;
 
     #[cfg(target_os = "windows")]
     #[test]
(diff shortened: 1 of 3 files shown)


src/auth/src/credentials.rs
```diff
@@ -165,3 +165,80 @@ pub mod traits {
...
+    }
+    adc_well_known_path()
+}
+
+/// The well-known path to ADC on Windows, as specified in [aip/4113].
+///
+/// [aip/4113]: https://google.aip.dev/auth/4113
+#[cfg(target_os = "windows")]
+fn adc_well_known_path() -> String {
+    let root = std::env::var("APPDATA").unwrap();
```
coryan, line 185: This will panic and crash the application.  I do not believe that is the behavior we want.  It should be something like "there is no ADC file".

src/auth/src/credentials.rs
```diff
@@ -165,3 +165,80 @@ pub mod traits {
...
+fn adc_path() -> String {
+    if let Ok(e) = std::env::var("GOOGLE_APPLICATION_CREDENTIALS") {
+        return e;
+    }
+    adc_well_known_path()
+}
+
+/// The well-known path to ADC on Windows, as specified in [aip/4113].
+///
+/// [aip/4113]: https://google.aip.dev/auth/4113
```
coryan, line 182: Elsewhere we use `[AIP-4113]` for these links.  I think I like these better, but maybe we should stay consistent. Unless there is some precedent we should be following?

coryan (approved):

src/auth/src/credentials.rs
dbolduc: Changed to `[AIP-XXXX]`

src/auth/src/credentials.rs
dbolduc: Yeah, good catch. I assumed I would have a `HOME` / `APPDATA`.

Fixed by propagating errors. I kind of hate the error type though. I used `CredentialError`, but I don't think I want that. I think I want a new type that makes my life easier in the credential factory functions... Let me think about it.

coryan: `Option<T>` may do the trick.  It is made for "I wanted a T but it is not set / not found".

coryan: And btw, `HOME` is often missing when running inside a docker container.

dbolduc: > `Option<T>` may do the trick.

Yeah, thanks, I prefer this.

If we don't have ADC it is not an error, we are supposed to fallback to MDS credentials. So returning an error here did not make much sense.

src/auth/src/credentials.rs
```diff
@@ -165,3 +165,107 @@ pub mod traits {
...
+
+/// The well-known path to ADC on Linux and Mac, as specified in [AIP-4113].
+///
+/// [AIP-4113]: https://google.aip.dev/auth/4113
+#[cfg(not(target_os = "windows"))]
+fn adc_well_known_path() -> Option<String> {
+    if let Ok(root) = std::env::var("HOME") {
+        return Some(root + "/.config/gcloud/application_default_credentials.json");
+    }
+    None
```
coryan, line 199: Not sure if this is better, but you could write:

```suggestion
    std::env::var("HOME").ok().map(|root| root + "/.config/gcloud/application_default_credentials.json")
```

src/auth/src/credentials.rs
```diff
@@ -165,3 +165,107 @@ pub mod traits {
...
+        );
+    }
+
+    #[cfg(target_os = "windows")]
+    #[test]
+    #[serial_test::serial]
+    fn adc_well_known_path_windows_no_appdata() {
+        let _creds = ScopedEnv::remove("GOOGLE_APPLICATION_CREDENTIALS");
+        let _appdata = ScopedEnv::remove("APPDATA");
+        assert!(adc_well_known_path().is_none());
```
coryan, line 229: Here and below, consider:

```suggestion
        assert_eq!(adc_well_known_path(), None);
```

which should give you better error messages on failure.

src/auth/src/credentials.rs
```diff
@@ -165,3 +165,107 @@ pub mod traits {
...
+    #[cfg(not(target_os = "windows"))]
+    #[test]
+    #[serial_test::serial]
+    fn adc_well_known_path_posix() {
+        let _creds = ScopedEnv::remove("GOOGLE_APPLICATION_CREDENTIALS");
+        let _home = ScopedEnv::set("HOME", "/home/foo");
+        assert_eq!(
+            adc_well_known_path().unwrap(),
+            "/home/foo/.config/gcloud/application_default_credentials.json"
+        );
```
coryan, line 242: maybe: 

```suggestion
        assert_eq!(
            adc_well_known_path().as_deref(),
            Some("/home/foo/.config/gcloud/application_default_credentials.json")
        );
```

coryan (commented):
A few nits, none of these are blocking.

src/auth/src/credentials.rs
coryan: Another alternative is something like:

```rust
let root = std::env::var("HOME").ok()?;
Some(root + "/. blah blah")
```

The `?` thing works for `Result<>` and `Option<>` (I would say something about monads, but I left my smugness at home).

dbolduc: Done.

src/auth/src/credentials.rs
dbolduc: Fixed

src/auth/src/credentials.rs
dbolduc: Done

coryan (approved):

PR #554, opened 2024-12-30
ci: enable Windows and macOS builds

coryan:


.github/workflows/sdk.yaml
```diff
@@ -127,3 +127,41 @@ jobs:
         # committed code that is an error. All the inputs should be pinned,
         # including the generator version and the googleapis SHA.
       - run: git diff --exit-code
+  windows:
```
dbolduc, line 130: comment: These seem `matrix`-able, as the steps are identical to the `build` job

... but maybe it is foolish to expect the steps to remain identical.

dbolduc (approved):

.github/workflows/sdk.yaml
coryan: Tempting... We would want to skip the 1.70 builds on Windows and macOS, and the build names would get messier.... I am going to punt for now.

PR #556, opened 2024-12-30
cleanup(auth): minor code simplification

coryan:


dbolduc (approved):
Thanks, I have yet to acquire good taste in this language.

coryan:
> Thanks, I have yet to acquire good taste in this language.

I am not sure I have good taste, in this language or any other.  It just seems to me that -- in general -- using standard library features to express something *should* be more readable. Not because the standard library(ies) are perfect, but because they are well-known.

PR #559, opened 2024-12-30
feat: add try/from traits for conversion between Timestamp and String/chrono::DateTime

### LINKED ISSUE #423: Consider more from/tryfrom conversions for Duration and Timestamp

mpeddada1:
Fixes #423 

- Replaces use of `to_json` and `from_json` with TryFrom traits for conversion between `Timestamp` and String representations. 
- Adds TryFrom traits for conversion between `chrono::DateTime` and `Timestamp`. **Note** that additional tests have been added to a separate `tests/timestamp.rs` following the convention set by existing tests that verify conversion between `Timestamp` and `time::OffsetDateTime`.

### DIFF
--- a/src/wkt/tests/timestamp.rs
+++ b/src/wkt/tests/timestamp.rs
@@ -99,3 +99,21 @@ fn convert_to_time() -> Result {
     assert_eq!(got, want);
     Ok(())
 }
+
+#[test]
+fn convert_from_chrono_time() -> Result {
+    let ts = chrono::DateTime::from_timestamp(123, 456789012).unwrap();
+    let got = Timestamp::try_from(ts)?;
+    let want = Timestamp::new(123, 456789012)?;
+    assert_eq!(got, want);
+    Ok(())
+}
+
+#[test]
+fn convert_to_chrono_time() -> Result {
+    let ts = Timestamp::new(123, 456789012)?;
+    let got = chrono::DateTime::try_from(ts)?;
+    let want = chrono::DateTime::from_timestamp(123, 456789012).unwrap();
+    assert_eq!(got, want);
+    Ok(())
+}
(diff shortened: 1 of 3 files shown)


src/wkt/tests/timestamp.rs
```diff
@@ -99,3 +99,21 @@ fn convert_to_time() -> Result {
     assert_eq!(got, want);
     Ok(())
 }
+
```
mpeddada1, line 102: Do we prefer having these tests be in `wkt/src/timestamp.rs`?

coryan: I think so, but maybe I am wrong.

The tests in this file can **only** use the public API for `Timestamp` and related trait implementations.  That makes them better tests of the behavior of the type, as they cannot use any private functions or the type internals.

coryan (approved):

src/wkt/tests/timestamp.rs
mpeddada1: Thanks for the context, from reading https://doc.rust-lang.org/book/ch11-03-test-organization.html#:~:text=The%20purpose%20of%20unit%20tests,code%20that%20they're%20testing, it looks like the tests/ file is reserved for integration tests? 

_Unit tests are small and more focused, testing one module in isolation at a time, and can test private interfaces. Integration tests are entirely external to your library and use your code in the same way any other external code would, using only the public interface and potentially exercising multiple modules per test._

We have a few other tests (like `convert_from_time`) that might belong in the `wkt/src/timestamp.rs`. I'll create a follow up to move them.

mpeddada1: Create https://github.com/googleapis/google-cloud-rust/issues/573 for tracking

PR #564, opened 2024-12-31
cleanup: simplify generated code

coryan:
We were using an alias of an alias for in the generated code. Just use
the type directly.

Motivated by #437

### DIFF
--- a/generator/testdata/rust/protobuf/golden/iam/v1/src/lib.rs
+++ b/generator/testdata/rust/protobuf/golden/iam/v1/src/lib.rs
@@ -17,8 +17,7 @@
 /// The messages and enums that are part of this client library.
 pub mod model;
 
-/// Common error returned by the RPCs in this client library.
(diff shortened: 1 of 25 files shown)


generator/testdata/rust/protobuf/golden/iam/v1/src/lib.rs
```diff
@@ -55,8 +54,3 @@ pub(crate) mod info {
     }
 }
 
-/// A `Result` alias where the `Err` case is an [Error].
-pub type Result<T> = std::result::Result<T, Error>;
```
codyoss, line 59: I do understand not wanting a chain of aliases, but maybe those aliases in gax should just not exist. I think the downside of this change is there is no longer a top-level links to these important types in our crates documentation -- outside of method signatures. Also, for result specifically it does feel a little odd to return another crates result type.

codyoss (approved):

PR #565, opened 2025-01-01
WIP - feat(gax): implement retry loop

coryan:
Related to #437


src/gax/tests/retry_policy.rs
```diff
@@ -0,0 +1,49 @@
+// Copyright 2024 Google LLC
```
coryan, line 1: This should be part of the retry policy PR

src/gax/src/options/mod.rs
```diff
@@ -105,6 +105,26 @@ impl RequestOptions {
...
+        &self,
+    ) -> Option<Box<dyn crate::backoff_policy::BackoffPolicy>> {
+        self.backoff_policy_provider.as_ref().map(|f| f())
+    }
+
+    /// Returns the request retry throttler, if any.
+    #[cfg(feature = "unstable-sdk-client")]
+    pub(crate) fn retry_throttler(&self) -> Option<RetryThrottlerOption> {
+        self.retry_throttler.as_ref().cloned()
+    }
```
coryan, line 127: This should go away or move to the ReqwestClient

src/gax/src/http_client/mod.rs
```diff
@@ -60,12 +66,108 @@ impl ReqwestClient {
...
+        builder: reqwest::RequestBuilder,
+        options: &crate::options::RequestOptions,
+        retry_policy: Box<dyn crate::retry_policy::RetryPolicy>,
+    ) -> Result<O> {
+        let throttler = options
+            .retry_throttler()
+            .unwrap_or_else(|| self.retry_throttler.clone());
+        let mut backoff = options
+            .make_backoff_policy()
+            .unwrap_or(Box::new(ExponentialBackoff::default()));
```
coryan, line 94: This should use the policy from the client config if available

src/gax/src/http_client/mod.rs
```diff
@@ -60,12 +66,108 @@ impl ReqwestClient {
                 reqwest::header::HeaderValue::from_str(user_agent).map_err(Error::other)?,
             );
         }
-        if let Some(timeout) = options.attempt_timeout() {
-            builder = builder.timeout(*timeout);
-        }
         if let Some(body) = body {
             builder = builder.json(&body);
         }
+        let retry_policy = options.make_retry_policy();
```
coryan, line 72: This should fallback on the option from the `ReqwestClient` if available

src/gax/src/http_client/mod.rs
```diff
@@ -12,16 +12,21 @@
...
 use crate::Result;
 use auth::Credential;
+use std::sync::{Mutex, PoisonError};
 
 #[derive(Clone)]
 pub struct ReqwestClient {
     inner: reqwest::Client,
     cred: Credential,
     endpoint: String,
+    retry_throttler: RetryThrottlerOption,
```
coryan, line 29: Need to add the retry and backoff policy providers here.

src/gax/tests/retry_policy.rs
coryan: Done.

src/gax/src/http_client/mod.rs
```diff
@@ -54,18 +66,103 @@ impl ReqwestClient {
...
+                        throttler
+                            .lock()
+                            .expect("retry throttler lock is poisoned")
+                            .on_success();
+                    }
+                    return Ok(r);
+                }
+                Err(e) => {
+                    let delay = backoff.on_failure();
+                    let flow = retry_policy.on_error(/* TODO */ true, e);
```
coryan, line 121: This should be an argument in `RequestOptions` and set by the generated code.

src/gax/src/http_client/mod.rs
```diff
@@ -54,18 +66,103 @@ impl ReqwestClient {
...
+                .lock()
+                .expect("retry throttler lock is poisoned")
+                .on_retry_failure(&retry_flow);
+        }
+        match retry_flow {
+            RetryFlow::Permanent(e) | RetryFlow::Exhausted(e) => {
+                return Err(e);
+            }
+            RetryFlow::Continue(_e) => {
+                tokio::time::sleep(backoff_delay).await;
```
coryan, line 148: This is not tested.  We need to verify the sleep happens.

src/gax/src/http_client/mod.rs
```diff
@@ -54,18 +66,103 @@ impl ReqwestClient {
...
+                    if !first_attempt {
+                        throttler
+                            .lock()
+                            .expect("retry throttler lock is poisoned")
+                            .on_success();
+                    }
+                    return Ok(r);
+                }
+                Err(e) => {
+                    let delay = backoff.on_failure();
```
coryan, line 120: What happens if `retry_policy.remaining_time()` is going to be zero after this long?  Maybe we should return immediately, as it makes no sense to sleep only to fail on wake up?

coryan, line 120: I do not think the integration tests validate that the backoff policy is called.

src/gax/src/http_client/mod.rs
```diff
@@ -54,18 +66,103 @@ impl ReqwestClient {
...
+            };
+            let response = if accept {
+                self.request_attempt(builder, options, remaining_time).await
+            } else {
+                Err(Error::other("TODO change to a better error".to_string()))
+            };
+            match response {
+                Ok(r) => {
+                    if !first_attempt {
+                        throttler
```
coryan, line 112: This is not tested.  We do not have any tests that succeeds on retry attempts.

coryan (commented):
LGTM

PR #572, opened 2025-01-02
feat(gax): introduce `RetryPolicy`

coryan:
This introduces a trait and some implementations to drive retry loops.
The policy controls for how long the retry loop runs, and what errors
are considered retryable.

Part of the work for #437.  #565  maybe give you a more complete picture of where this is going.

### DIFF
--- a/src/gax/src/retry_policy.rs
+++ b/src/gax/src/retry_policy.rs
@@ -545,13 +545,14 @@ where
     P: RetryPolicy,
 {
     fn on_error(&self, idempotent: bool, error: Error) -> RetryFlow {
-        let exhausted = match self.attempt_count.lock() {
-            Err(_) => true,
-            Ok(guard) => {
-                let count = guard.get().saturating_add(1);
-                guard.set(count);
-                count > self.maximum_attempts
-            }
+        let exhausted = {
+            let guard = self
+                .attempt_count
+                .lock()
+                .expect("retry policy attempt counter is poisoned");
+            let count = guard.get().saturating_add(1);
+            guard.set(count);
+            count > self.maximum_attempts
         };
         match self.inner.on_error(idempotent, error) {
             RetryFlow::Permanent(e) => RetryFlow::Permanent(e),
--- a/src/gax/src/retry_policy.rs
+++ b/src/gax/src/retry_policy.rs
@@ -26,11 +26,13 @@
 //!
 //! # Example:
 //! ```
-//! # use std::sync::Arc;
 //! # use gcp_sdk_gax::retry_policy::*;
-//! # use gcp_sdk_gax::options::RequestOptionsBuilder;
-//! fn customize_retry_policy(builder: impl RequestOptionsBuilder) -> impl RequestOptionsBuilder {
-//!     builder.with_retry_policy(LimitedAttemptCount::provider(5))
+//! # use gcp_sdk_gax::options;
+//! fn customize_retry_policy(config: options::ClientConfig) -> options::ClientConfig {
+//!     // Retry for at most 10 seconds or at most 5 attempts: whichever limit
+//!     // is reached first stops the retry loop.
+//!     let d = std::time::Duration::from_secs(10);
+//!     config.set_retry_policy(Aip194Strict.with_time_limit(d).with_attempt_limit(5))
 //! }
 //! ```
 //!
@@ -38,9 +40,7 @@
 
 use crate::error::rpc::Status;
 use crate::error::{Error, HttpError};
-use std::cell::Cell;
 use std::sync::Arc;
-use std::sync::Mutex;
 
 /// The result of a retry policy decision.
 ///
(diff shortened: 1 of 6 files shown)


src/gax/src/retry_policy.rs
```diff
@@ -0,0 +1,954 @@
...
+/// [Continue][std::ops::ControlFlow::Continue]. If the caller
+/// should stop retrying, the policy returns
+/// [Break][std::ops::ControlFlow::Break].
+///
+/// In both variants the result includes an error. This is useful when retry
+/// policies are composed. The inner policy returns `Continue` based on the
+/// error type, and the outer policy may return `Break` based on the number
+/// errors, or the elapsed time.
+#[derive(Debug)]
+pub enum RetryFlow {
```
codyoss, line 57: nit: I wonder if this should be called RetryFlowFailure or something as all of these are the sad path of a retry loop.

src/gax/src/retry_policy.rs
```diff
@@ -0,0 +1,954 @@
...
+/// no method provider is available. The application may also set a provider in
+/// effect for a client method. Note that some methods issue multiple RPCs, for
+/// example, pagination helpers and LRO helpers. The provider is used to create
+/// a retry policy for each one of the RPCs.
+///
+/// Retry policy providers are passed between async functions, so they must be
+/// `Send` and `Sync`.  They must also implement [Debug][std::fmt::Debug]
+/// because they are logged as part of the overall request / response logs.
+pub trait RetryPolicyProvider: Send + Sync + std::fmt::Debug {
+    fn make(&self) -> Box<dyn RetryPolicy>;
```
codyoss, line 127: Something about this does not feel right to me. But maybe I need to look at more examples. Have you considered a Extension trait pattern instead? That way you don't need to specify these extra bounds and callers could work with returned concrete types. Some examples I have seen:

- https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html
- https://docs.rs/tokio/latest/tokio/io/trait.AsyncBufReadExt.html

src/gax/src/retry_policy.rs
coryan: Maybe?  This was inspired by `std::ops::ControlFlow`, but I am terrible at naming.

With that said, is not all sad path: `RetryFlow::Continue` is the happy case.

src/gax/src/retry_policy.rs
```diff
@@ -0,0 +1,953 @@
...
+/// Implementations of this trait determine if errors are retryable, and for how
+/// long the retry loop may continue.
+pub trait RetryPolicy: Send + Sync {
+    /// Query the retry policy after an error.
+    ///
+    /// # Parameters
+    /// * `idempotent` - if `true` assume the operation is idempotent. Many more
+    ///   errors are retryable on idempotent operations.
+    /// * `error` - the last error when attempting the request.
+    fn on_error(&self, idempotent: bool, error: Error) -> RetryFlow;
```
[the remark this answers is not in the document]
coryan: > If something knows it is not idempotent it should not configure a retry in the first place, right? 

I think retry loops still apply.  Some errors are retryable even for non-idempotent operations, e.g. if we failed to create the access token, then we never sent the request, and we can retry safely.

It also gives customers flexibility: they can set a default retry policy for the client and then in the policy they can decide what to do about non-idempotent requests.  In some applications / services even POST operations are safe (with a request id, or with preconditions, or maybe the application just does not care).

src/gax/src/retry_policy.rs
coryan: There is some stuff I dislike about this too, like why do I need to pass a provider to the request options when I am just going to use the retry policy the one time?  But your comment went over my head a bit. 

We *could* make the retry loop generic on the concrete retry policy (and backoff policy, and throttler).

But that means the `ClientConfig` and `RequestOptions` would  need to be generic too.  And then that implies the `*Builder` needs to be generic, because we want applications to say:

```rust
  // pseudo code, I probably will get some names wrong ...
  let config = ClientConfig::default().with_retry_policy(default_retry).with_backoff_policy(default_backoff);

  let client = SecretManager::new(config).await?;

  let response = client.create_secret(..).with_retry_policy(blah).with_backoff_policy(blah blah).send().await?
```

We would also need to figure out how to combine the default retry policy in the client, with the retry policy in the request options.  What if they are of different types?

**TL;DR;** I think you are onto something here, but I do not know how to implement something like that.

src/gax/src/retry_policy.rs
```diff
@@ -89,63 +89,119 @@ impl RetryFlow {
 ///
 /// Implementations of this trait determine if errors are retryable, and for how
 /// long the retry loop may continue.
-pub trait RetryPolicy: Send + Sync {
+pub trait RetryPolicy: Send + Sync + std::fmt::Debug {
```
codyoss, line 92: I wonder if this trait should have a now method for clock manipulation. We could provide a default.

codyoss (approved):

src/gax/src/retry_policy.rs
coryan: I started making this change the the tests looked more complicated.  Maybe I am holding this wrong. We can try again in a future PR.

PR #578, opened 2025-01-02
feat: (WIP) implement PathParameter trait for Enums

dangazineu:
Still need to figure a way to test this :)

### DIFF
--- a/generator/internal/language/templates/rust/crate/src/enum.mustache
+++ b/generator/internal/language/templates/rust/crate/src/enum.mustache
@@ -15,15 +15,8 @@ limitations under the License.
 }}
 {{> ../../common/enum}}
 
-{{! TODO: before merging, make the following code conditional on the need for the enum to be used as a path parameter }}
-
-impl gax::path_parameter::PathParameter for {{Name}} {
-
-    type P = {{Name}};
-
-    fn required<'a>(&'a self, _: &str) -> std::result::Result<&'a Self::P, gax::path_parameter::Error> {
-        Ok(self)
+impl gax::request_parameter::RequestParameter for {{Name}} {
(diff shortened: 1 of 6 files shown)


generator/internal/language/templates/rust/crate/src/enum.mustache
```diff
@@ -14,3 +14,16 @@ See the License for the specific language governing permissions and
...
+impl gax::path_parameter::PathParameter for {{Name}} {
+
+    type P = {{Name}};
+
+    fn required<'a>(&'a self, _: &str) -> std::result::Result<&'a Self::P, gax::path_parameter::Error> {
+        Ok(self)
+    }
+
+
+}
```
coryan, line 29: You can save yourself some trouble, I think this would make enums work as both query and path parameters:

```suggestion
impl gax::request_parameter::RequestParameter for {{Name}} {
    fn format(&self) -> gax::request_parameter::Result {
        Ok(self.0.clone())
    }
}
```

coryan (commented):
To test this, consider writing an integration test like:

https://github.com/googleapis/google-cloud-rust/blob/5201d89f54616288b91a2710f76bdf706233698e/src/gax/tests/path_parameters.rs#L60

Just simulate the generated code as part of the test.  There is a separate file to test these as query parameters.

generator/internal/language/templates/rust/crate/src/enum.mustache
dangazineu: Neat! This will require making the `gax::request_parameter` and `gax::request_parameter::RequestParameter` as public though.

I'll look into that test and update the PR later today.

dangazineu:
I'll close this, as the direction I am going has changed quite a bit, and there is no reason to keep these commits in the history.

PR #585, opened 2025-01-03
feat(auth): fallback to MDS credentials

dbolduc:
Part of the work for #442 

If the ADC path is not known, or the file does not exist, fallback to MDS credentials.

If the ADC path is found, but malformed in some way, we do not fallback, we return the error.

We are desperate for some integration tests. The unit tests just verify that some credentials are successfully made. They cannot tell the difference between MDS and User Credentials. It also wouldn't make sense to try to use the token providers in the unit tests.

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -215,24 +215,9 @@ pub mod traits {
 /// [gcloud auth application-default]: https://cloud.google.com/sdk/gcloud/reference/auth/application-default
 /// [gke-link]: https://cloud.google.com/kubernetes-engine
 pub async fn create_access_token_credential() -> Result<Credential> {
-    let (adc_path_kind, adc_path) = match adc_path() {
-        Some((adc_path_kind, adc_path)) => (adc_path_kind, adc_path),
-        None => return Ok(mds_credential::new()),
-    };
-    let contents = match std::fs::read_to_string(&adc_path) {
-        Ok(contents) => contents,
-        Err(e) if e.kind() == std::io::ErrorKind::NotFound => match adc_path_kind {
-            AdcPathKind::WellKnown => return Ok(mds_credential::new()),
-            AdcPathKind::FromEnv => {
-                return Err(CredentialError::new(
-                    false,
-                    Box::from(format!(
-                        "Failed to load Application Default Credentials (ADC) from {adc_path}. Check that the `GOOGLE_APPLICATION_CREDENTIALS` environment variable points to a valid file."
-                    )),
-                ))
-            }
-        },
-        Err(e) => return Err(CredentialError::new(false, e.into())),
+    let contents = match load_adc()? {
+        AdcContents::Contents(contents) => contents,
+        AdcContents::FallbackToMds => return Ok(mds_credential::new()),
     };
     let js: serde_json::Value =
         serde_json::from_str(&contents).map_err(|e| CredentialError::new(false, e.into()))?;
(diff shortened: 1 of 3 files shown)


src/auth/src/credentials.rs
```diff
@@ -215,22 +215,17 @@ pub mod traits {
...
+    let adc_path = match adc_path() {
+        Some(path) => path,
+        None => {
+            return Ok(mds_credential::new());
         }
-    })?;
+    };
+    let contents = match std::fs::read_to_string(adc_path) {
+        Ok(contents) => contents,
+        Err(e) if e.kind() == std::io::ErrorKind::NotFound => return Ok(mds_credential::new()),
```
coryan, line 226: So if `GOOGLE_APPLICATION_CREDENTIALS` points to a file that does not exist we fall back on MDS?  I think this:

https://google.aip.dev/auth/4110#expected-behavior

Says you should return an error.  But I find that diagram and the text very confusing.

dbolduc: You are right. I did not read the spec closely.

I considered doing away with `fn adc_path()`, but instead decided to have it also return whether or not we are using the well known path.

I don't have a unit test for falling back to MDS if the well known path is not found. It seems like we don't have too much control over that. I could write a test that checks if the well known path is a file, and sees if we fallback based on that. Do you think that test is worth having?

https://doc.rust-lang.org/std/path/struct.Path.html#method.is_file

src/auth/src/credentials.rs
```diff
@@ -253,13 +256,20 @@ pub async fn create_access_token_credential() -> Result<Credential> {
     }
 }
 
+#[derive(Debug, PartialEq)]
+enum AdcPathKind {
+    FromEnv,
+    WellKnown,
+}
+
```
coryan, line 264: I think you want something like:

```suggestion
#[derive(Debug, PartialEq)]
enum AdcPath {
    FromEnv(String),
    WellKnown(String)
}
```

and then return `fn adc_path() -> Option<AdcPath>` or you want:

```suggestion
#[derive(Debug, PartialEq)]
enum AdcPath {
    FromEnv(String),
    WellKnown(String),
    None
}
```

and then `adc_path() -> AdcPathKind`

src/auth/tests/credentials.rs
```diff
@@ -22,24 +22,22 @@ mod test {
...
-        let msg = err.source().unwrap().to_string();
-        assert!(msg.contains("Unable to find Application Default Credentials"));
+        create_access_token_credential().await.unwrap();
     }
 
     #[tokio::test]
     #[serial_test::serial]
-    async fn create_access_token_credential_no_adc_file_fallback() {
-        // TODO(#442) - We should (successfully) fall back to MDS Credentials. But for now we are temporarily returning an error.
+    async fn create_access_token_credential_errors_if_adc_env_is_not_a_file() {
```
coryan, line 34: Seems like you should be able to add a test like:

```rust
#[tokio::test]
#[serial_test::serial]
async fn create_access_token_credential_no_well_known_adc() -> std::result::Result<(), Box<dyn std::error::Error>>  {
  let dir = testfile::tempdir()?;
  let _e1 = ScopedEnv::remove("GOOGLE_APPLICATION_CREDENTIALS");
  let _e2 = ScopedEnv::set("HOME", format!("{}/blah blah", tempdir.path()); 
  let _e3 = ScopedEnv::set("APPDATA", format!({}/blah blah", tempdir.path());

  let e = create_access_token_credential().await;
  assert!(e.is_err());

  Ok(())
}
```

coryan (changes requested):
If you accept my suggestion on `AdcPathKind` then all the other code changes (for the better I think).  So I did not review it.

src/auth/tests/credentials.rs
dbolduc: Yeah, that is clever. I did not think of overriding `HOME` / `APPDATA`. Added a test. Thanks.

src/auth/src/credentials.rs
```diff
@@ -253,13 +241,56 @@ pub async fn create_access_token_credential() -> Result<Credential> {
...
+    match adc_path() {
+        AdcPath::None => Ok(AdcContents::FallbackToMds),
+        AdcPath::FromEnv(path) => {
+            match std::fs::read_to_string(&path) {
+                Ok(contents) => Ok(AdcContents::Contents(contents)),
+                Err(e) if e.kind() == std::io::ErrorKind::NotFound => Err(CredentialError::new(
+                    false,
+                    Box::from(format!(
+                        "Failed to load Application Default Credentials (ADC) from {path}. Check that the `GOOGLE_APPLICATION_CREDENTIALS` environment variable points to a valid file."
+                    )))),
```
coryan, line 267: nit: maybe move this to a function, so it reads easier?

```suggestion
                Err(e) if e.kind() == std::io::ErrorKind::NotFound => Err(path_not_found(path)),
```

src/auth/src/credentials.rs
```diff
@@ -253,13 +241,56 @@ pub async fn create_access_token_credential() -> Result<Credential> {
...
 /// The path to Application Default Credentials (ADC), as specified in [AIP-4110].
 ///
 /// [AIP-4110]: https://google.aip.dev/auth/4110
-fn adc_path() -> Option<String> {
+fn adc_path() -> AdcPath {
     std::env::var("GOOGLE_APPLICATION_CREDENTIALS")
         .ok()
-        .or_else(adc_well_known_path)
+        .map_or_else(
+            || match adc_well_known_path() {
```
coryan, line 288: 🫤  maybe a bit too big for a `.map_or_else()`

coryan (approved):
Some nits that you can fix at your discretion.  :shipit:

src/auth/src/credentials.rs
dbolduc: Done

src/auth/src/credentials.rs
dbolduc: Rearranged.

PR #587, opened 2025-01-03
feat(gax): introduce `BackoffPolicy`

coryan:
Introduce `BackoffPolicy` a trait that defines how long the retry
loop waits before making a new attempt.  Provide the most common
implementation: truncated exponential backoff with (full) jitter.

Part of the work for #437

### DIFF
--- a/src/gax/src/backoff_policy.rs
+++ b/src/gax/src/backoff_policy.rs
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2025 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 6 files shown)


src/gax/src/backoff_policy.rs
```diff
@@ -0,0 +1,463 @@
+// Copyright 2024 Google LLC
```
dbolduc, line 1: drive by: 2025

src/gax/tests/backoff_policy.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2024 Google LLC
```
dbolduc, line 1: same

PR #588, opened 2025-01-04
WIP - prototype enums as path parameters

coryan:


src/gax/tests/path_parameters.rs
```diff
@@ -56,33 +61,66 @@ impl FakePayload {
...
+/// Constants representing the known values of the enum
+pub mod fake_enum_parameter {
+    pub const RED: &str = "RED";
+    pub const GREEN: &str = "GREEN";
+}
+
+impl gcp_sdk_gax::path_parameter::PathParameter for FakeEnumParameter {
+    type P = String;
+    fn required<'a>(&'a self, _: &str) -> gcp_sdk_gax::path_parameter::Result<&'a Self::P> {
+        Ok(&self.0)
```
coryan, line 90: @dangazineu this is the only interesting line (I think).

src/gax/src/path_parameter.rs
```diff
@@ -12,14 +12,61 @@
...
 
+//! Defines traits and helpers to serialize path parameters.
+//!
+//! Path parameters in the Google APIs are always required, but they may need
+//! to be source from fields that are `Option<T>`, or they may be nested fields.
+//! In the later case, the containing submessage is always `Option<M>`.
+//!
+//! We could change the generator to issue different code depending on whether
+//! the parameter is `Option<T>` or not. In the first case we would have the
+//! generator write:
```
coryan, line 23: @dangazineu in hindsight, I wonder if changing the generator and moving the complexity there is a better solution.

dangazineu: IMO, the option where the generator produces a series of chained `unwrap_or_else` calls, despite being more verbose, seems more readable than chaining calls to `PathParamether::require`. 

A third option, that I don't know how to express in Rust yet, would be to embed the `require` logic in the types that hold optional fields; I think that passing the field name to a helper function seems odd. The holding type already knows the name of their fields, and could produce the error with the field name as needed. 

```rust
format!("/v1/foos/{}"
     request.require_field()? 
)
```
OR
```rust
format!("/v1/foos/{}"
     request.field_or_else(|| gax::path_parameter::missing())? 
)
```

coryan: > IMO, the option where the generator produces a series of chained `unwrap_or_else` calls, despite being more verbose, seems more readable than chaining calls to `PathParamether::require`.

Fair.  Though I do not expect a lot of people to read generated code.

> A third option, that I don't know how to express in Rust yet, would be to embed the `require` logic in the types that hold optional fields;

Interesting.... I think something along the lines of:

```rust
impl FooRequest {
  // For a field that is Option<T>
  pub(crate) fn get_field(&self) -> Result<&T, gax::error::Error> {
    self.field.as_ref().ok_or_else(|| gax::error::Error::other(
        format!("field {} is required in a request, but was missing", "field")))
  }

  // For a field that is T
  pub(crate) fn get_field(&self) -> Result<&T, gax::error::Error> {
    Ok(&self.field)
  }

  ... ... .. other stuff ...
```

I am suggesting `pub(crate)` because these ad-hoc functions should not be used by customers.  It would be nice if we could avoid generating this code unless the field is a path parameter.

> I think that passing the field name to a helper function seems odd. The holding type already knows the name of their fields, and could produce the error with the field name as needed.

Well, yes.  But using that knowledge in the mustache templates is really hard (maybe impossible).  So you would be coding this in the Rust codec.  Which also seems less than ideal: it may require creating more Rust-specific fields in the mustache data model.

With all of this said: I encourage you to explore the idea of using the generator + mustache templates to generate this code.  I don't think we can find out what solution works best by just looking at the screen.

PR #589, opened 2025-01-04
feat(auth): implement `std::fmt::Debug` trait

coryan:
Make the credentials easier to troubleshoot. Implementing the `Debug`
traits will help us log their state at runtime, including what specific
token provider is in use.

In some cases we need to censor some fields because they contain
sensitive information. We may consider logging a hash of the field, or
maybe the first N characters of the field and a hash. I decided to be
conservative for now.


dbolduc (approved):
Thanks

PR #590, opened 2025-01-04
impl(auth): Credentials have clone

dbolduc:
Part of the work for #442 

The `gax::http_client::ReqwestClient` has `Clone`. We want it to hold `auth::credential::Credential`. So `Credential` needs `Clone` as well.

We definitely want the interfaces in the `traits::Credential` to take `&self` instead of `&mut self`. We want to be able to call these things in `ReqwestClient::execute(&self, ...)`.

## On changing the `TokenProvider` trait...

This PR also changes the `TokenProvider`'s traits to use `&self` instead of `&mut self`. This is not necessarily what we want in the long term, but it unblocks us for now.

When we introduce caching, we will be mutating some aspect of the token provider. We could have the caching layer hold an interior mutable cache.

I think more likely though, we will want the `*Credential` types to hold an interior mutable token provider.

We can solve the problem when we design the token cache. Note that these APIs are all private outside of the crate, so it is not a big deal.

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -56,6 +56,14 @@ use std::sync::Arc;
 /// [Google Kubernetes Engine]: https://cloud.google.com/kubernetes-engine
 #[derive(Clone, Debug)]
 pub struct Credential {
+    // We use an `Arc` to hold the inner implementation.
+    //
+    // Credentials may be shared across threads (`Send + Sync`), so an `Rc`
+    // will not do.
+    //
+    // They also need to derive `Clone`, as the
+    // `gax::http_client::ReqwestClient`s which hold them derive `Clone`. So a
+    // `Box` will not do.
     inner: Arc<dyn traits::dynamic::Credential>,
 }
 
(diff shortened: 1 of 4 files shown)


src/auth/src/credentials.rs
```diff
@@ -53,23 +54,21 @@ use std::future::Future;
 /// [Metadata Service]: https://cloud.google.com/compute/docs/metadata/overview
 /// [Google Compute Engine]: https://cloud.google.com/products/compute
 /// [Google Kubernetes Engine]: https://cloud.google.com/kubernetes-engine
-#[derive(Debug)]
+#[derive(Clone, Debug)]
 pub struct Credential {
-    inner: Box<dyn traits::dynamic::Credential>,
+    inner: Arc<dyn traits::dynamic::Credential>,
```
coryan, line 59: Future you may wonder why `Arc` and not `Rc` or `Box` consider a comment (I think `Rc` is out because you need `Send + Sync`, and `Box` because you need `Clone`).

coryan (approved):

src/auth/src/credentials.rs
dbolduc: Done.

PR #592, opened 2025-01-05
feat(auth): credentials for tests

coryan:
Tests connecting to a local (usually temporary) server use test
credentials. The generated tokens are not used. We could use mocks, but
those are fairly tedious to set up when token creation is uninteresting.

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -327,6 +327,8 @@ mod test {
     use super::*;
     use scoped_env::ScopedEnv;
     use std::error::Error;
+    use traits::Credential;
+    type TestResult = std::result::Result<(), Box<dyn std::error::Error>>;
 
     #[derive(Clone, Debug)]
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials.rs
```diff
@@ -315,12 +315,46 @@ fn adc_well_known_path() -> Option<String> {
         .map(|root| root + "/.config/gcloud/application_default_credentials.json")
 }
 
+#[cfg(test)]
```
dbolduc, line 318: I do not think this is accessible from other crates. This is what I did:

https://github.com/googleapis/google-cloud-rust/blob/66a07fe9b269a42eb00b5aebcefb221f4abfed11/src/auth/src/credentials.rs#L318-L337

Should we go with my PR?

coryan: Your PR seems like a better idea.

Issue #593, opened 2025-01-05
Auth credentials should be mockable


dbolduc:
`Credential`s are not mockable from outside the `gcp-sdk-auth` crate AFAICT.

We have the proper traits:

https://github.com/googleapis/google-cloud-rust/blob/de1bb37d5328e1ba64732b5d50c653cb8c94f631/src/auth/src/credentials.rs#L84

But there is no public ctor to actually make a `Credential` from the `traits::Credential`.

https://github.com/googleapis/google-cloud-rust/blob/de1bb37d5328e1ba64732b5d50c653cb8c94f631/src/auth/src/credentials.rs#L67

PR #594, opened 2025-01-05
impl: swap out auth implementation in gax

### LINKED ISSUE #442: Bootstrap `gcp-sdk-auth`

dbolduc:
Fixes #442

Swap out the `gax` implementation of auth from `google-cloud-auth` to `gcp-sdk-auth`.

The GCB build runs integration tests using the `MDSCredentials` written by @harkamaljot.

Also introduces fake credentials in `gcp_sdk_auth::credentials::testing`. I think we need a public module. I don't think this thing gets exported to other crates if we gate it with a `#[cfg(test)]`.

I would like to fix the mocking story for credentials, and use them in the auth unit test. But I will defer that. It is tracked in #593

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -315,6 +315,58 @@ fn adc_well_known_path() -> Option<String> {
         .map(|root| root + "/.config/gcloud/application_default_credentials.json")
 }
 
+/// A module providing invalid credentials where authentication does not matter.
+///
+/// These credentials are a convenient way to avoid errors from loading
+/// Application Default Credentials in tests.
+///
+/// This module is mainly relevant to other `gcp-sdk-*` crates, but some
+/// external developers (i.e. consumers, not developers of `google-cloud-rust`)
+/// may find it useful.
+pub mod testing {
+    use crate::credentials::traits::dynamic::Credential as CredentialTrait;
+    use crate::credentials::Credential;
+    use crate::token::Token;
+    use crate::Result;
+    use http::header::{HeaderName, HeaderValue, AUTHORIZATION};
+    use std::sync::Arc;
+
+    /// A simple credentials implementation to use in tests where authentication does not matter.
+    ///
+    /// Always returns a "Bearer" token, with "test-only-token" as the value.
+    pub fn test_credentials() -> Credential {
+        Credential {
+            inner: Arc::from(TestCredential {}),
+        }
+    }
+
+    #[derive(Debug)]
+    struct TestCredential {}
+
+    #[async_trait::async_trait]
+    impl CredentialTrait for TestCredential {
+        async fn get_token(&self) -> Result<Token> {
+            Ok(Token {
+                token: "test-only-token".to_string(),
+                token_type: "Bearer".to_string(),
+                expires_at: None,
+                metadata: None,
+            })
+        }
+
+        async fn get_headers(&self) -> Result<Vec<(HeaderName, HeaderValue)>> {
+            Ok(vec![(
+                AUTHORIZATION,
+                HeaderValue::from_static("Bearer: test-only-token"),
+            )])
+        }
+
+        async fn get_universe_domain(&self) -> Option<String> {
+            Some("googleapis.com".to_string())
+        }
+    }
+}
+
 #[cfg(test)]
 mod test {
     use super::*;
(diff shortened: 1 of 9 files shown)


src/auth/src/credentials.rs
```diff
@@ -315,6 +315,58 @@ fn adc_well_known_path() -> Option<String> {
...
+    ///
+    /// Always returns a "Bearer" token, with "test-only-token" as the value.
+    pub fn test_credentials() -> Credential {
+        Credential {
+            inner: Arc::from(TestCredential {}),
+        }
+    }
+
+    #[derive(Debug)]
+    struct TestCredential {}
```
coryan, line 344: nit: 

```suggestion
    struct TestCredential;
```

src/gax/src/http_client/mod.rs
```diff
@@ -15,7 +15,8 @@
 use crate::error::Error;
 use crate::error::HttpError;
 use crate::Result;
-use auth::Credential;
+use auth::credentials::traits::Credential as CredentialTrait;
```
coryan, line 18: For a future PR (maybe): in the generated code we use a module for the traits because we want to avoid name clashes.  You do not have that problem and (maybe) using `CredentialTrait` would be easier to manage.

src/gax/src/http_client/mod.rs
```diff
@@ -30,7 +31,9 @@ impl ReqwestClient {
         let cred = if let Some(c) = config.cred {
             c
         } else {
-            ClientConfig::default_credential().await?
+            create_access_token_credential()
+                .await
+                .map_err(Error::authentication)?
```
coryan, line 36: For a future PR.  the `gax::Error` class should (maybe) only consume `CredentialsError` in `Error::authentication`.

src/gax/Cargo.toml
```diff
@@ -44,11 +44,12 @@ echo-server = { path = "echo-server" }
 # This is a workaround to integration test features of this crate. Open issue
 # https://github.com/rust-lang/cargo/issues/2911.
 gax         = { path = ".", package = "gcp-sdk-gax", features = ["unstable-sdk-client", "unstable-stream"] }
+async-trait = "0.1.83"
```
coryan, line 47: Do we need this?  It seems like it is not used.

coryan (approved):
Nothing seems blocking, but maybe clean these up before merging.

nit: there is still a TODO(#442) in src/auth/src/errors.rs  we fix that or carry it over in a separate bug?

dbolduc:
> nit: there is still a TODO(https://github.com/googleapis/google-cloud-rust/issues/442) in src/auth/src/errors.rs we fix that or carry it over in a separate bug?

oops, #595

src/auth/src/credentials.rs
dbolduc: Fixed.

src/gax/src/http_client/mod.rs
dbolduc: That seems like a good idea. I will rename in a follow up.

src/gax/src/http_client/mod.rs
dbolduc: Ack

src/gax/Cargo.toml
dbolduc: Nope, removed.

PR #596, opened 2025-01-05
cleanup: remove `google-cloud-auth` crate

### LINKED ISSUE #15: conflicting crate name
for better or worse https://crates.io/crates/google-cloud-auth has the same name as the auth/ member of this workspace.

Not expressing an opinion about whether having 'official' or 'community' crates is better, but you won't be able to publish or have people consume this effort with a conflicting name.

### LINKED ISSUE #21: Fix auth library integration tests to work in GHA
One of the auth library tests is really an integration test, it depends on things in the environment:

https://github.com/googleapis/google-cloud-rust/blob/1e1ed307e39c16c18cfaf3895ec16d4138349353/auth/src/lib.rs#L354-L362

This test breaks under GHA.  I do not know exactly what the problem might be.  I am going to skip the test when GCE is not detected.  That will unblock other development, and we can fix auth later.

coryan:
Now that we use `gcp-sdk-auth`, we can remove stale references to
`google-cloud-auth` from tests and golden files. And then we can remove
the code.

Fixes #15, and fixes #21.


.sidekick.toml
```diff
@@ -33,9 +33,9 @@ version = "0.1.0-rc2"
 'package:serde'      = 'force-used=true,package=serde,version=1.0.216,feature=serde_derive'
 'package:serde_with' = 'force-used=true,package=serde_with,version=3.12.0,default-features=false,feature=base64,feature=macros,feature=std'
 # These are used by crates with services.
-'package:async-trait' = 'required-by-services=true,package=async-trait,version=0.1.83'
+'package:async-trait' = 'required-by-services=true,package=async-trait,version=0.1.84'
```
dbolduc, line 36: Q: were these necessary?

dbolduc (approved):

.sidekick.toml
coryan: No. Kind of. Maybe. I don't know 😵‍💫 

I did a `cargo update` to get all the dependencies updated (since a lot of them get removed / changed by removing `/auth`), and that requires updating the generator configuration.

PR #608, opened 2025-01-06
feat(gax): introduce `RetryThrottler`

coryan:
Introduce the `RetryThrottler` trait and a couple of implementations.
This is used to limit the number of retries when too many retries are
failing. That can improve the recovery time under large failure or
overload conditions. Unlike backoff or retry policies, a single
RetryThrottler is used for all the requests in a client, or even in
multiple clients.

### DIFF
--- a/src/gax/src/lib.rs
+++ b/src/gax/src/lib.rs
@@ -79,3 +79,4 @@ pub mod http_client;
 pub mod backoff_policy;
 pub mod options;
 pub mod retry_policy;
+pub mod retry_throttler;
--- a/src/gax/src/retry_throttler.rs
+++ b/src/gax/src/retry_throttler.rs
@@ -1,0 +1,479 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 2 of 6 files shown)


src/gax/src/retry_throttler.rs
```diff
@@ -0,0 +1,479 @@
...
+    /// # use gcp_sdk_gax::retry_throttler::*;
+    /// fn configure_throttler() -> Result<http_client::ClientConfig> {
+    ///     let throttler = AdaptiveThrottler::new(2.0)?;
+    ///     Ok(http_client::ClientConfig::default()
+    ///           .set_retry_throttler(throttler))
+    /// }
+    /// ```
+    pub fn new(factor: f64) -> Result<Self> {
+        if factor < 0.0 {
+            return Err(Error::other(format!("factor ({factor}must be >= 0.0")));
```
codyoss, line 149: I wonder if we should have a validation type of error in the future.

src/gax/src/lib.rs
```diff
@@ -79,3 +79,4 @@ pub mod http_client;
 pub mod backoff_policy;
 pub mod options;
 pub mod retry_policy;
+pub mod retry_throttler;
```
codyoss, line 82: nit: I wonder if we should have a retry mod to group all of these cool things you have been building.

codyoss (approved):

src/gax/src/lib.rs
coryan: 🤔  maybe, I think this is the last one, but we can reconsider after I merge the complete loop.

coryan:
part of the work for #437

PR #616, opened 2025-01-06
feat(gax): support passing idempotency as a request option

coryan:
Part of the work for #437

### DIFF
--- a/src/gax/src/options/mod.rs
+++ b/src/gax/src/options/mod.rs
@@ -49,18 +49,27 @@ pub struct RequestOptions {
 }
 
 impl RequestOptions {
-    /// Treat the RPC underlying this method as idempotent.
-    pub fn set_idempotent(&mut self) {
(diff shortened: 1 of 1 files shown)


src/gax/src/options/mod.rs
```diff
@@ -48,6 +49,22 @@ pub struct RequestOptions {
 }
 
 impl RequestOptions {
+    /// Treat the RPC underlying this method as idempotent.
+    pub fn set_idempotent(&mut self) {
```
codyoss, line 53: Should this and the following RPC be combined and just accept a bool? Also, should we mention what setting this means for the request, i.e. how it relates retrying?

codyoss (approved):

src/gax/src/options/mod.rs
coryan: Done.

PR #618, opened 2025-01-06
fix(auth): Credential API

dbolduc:
Part of the work for #593 

`Credential` need not get its public interfaces from `CredentialTrait`.

We want to define
```rs
impl<T> std::convert::From<T> -> Credential
where T: CredentialTrait;
```

But if `Credential` implements `CredentialTrait`, this will conflict with the generic `impl<T> From<T> -> T`. (let `T = Credential` to see the problem).

The fix is to manually define the public interface for `Credential`.

This has some benefits:
1. Consumers do not need to `use auth::credentials::CredentialTrait` to have access to the methods.
1. The interface should read nicer. It is a simple `async fn`, instead of a `fn ... -> impl Future<Output = ...>`

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -67,17 +67,17 @@ pub struct Credential {
     inner: Arc<dyn dynamic::CredentialTrait>,
 }
 
-impl CredentialTrait for Credential {
-    fn get_token(&self) -> impl Future<Output = Result<crate::token::Token>> + Send {
-        self.inner.get_token()
+impl Credential {
+    pub async fn get_token(&self) -> Result<crate::token::Token> {
+        self.inner.get_token().await
     }
 
-    fn get_headers(&self) -> impl Future<Output = Result<Vec<(HeaderName, HeaderValue)>>> + Send {
-        self.inner.get_headers()
+    pub async fn get_headers(&self) -> Result<Vec<(HeaderName, HeaderValue)>> {
+        self.inner.get_headers().await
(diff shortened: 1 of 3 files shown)


coryan (approved):
I think this is fine, but be aware that consumers may be unable to mock `auth::Credential` (there is no trait to mock now).  They can only mock the `CredentialTrait`.  I think that is fine, but give it a thought.

dbolduc:
> be aware that consumers may be unable to mock `auth::Credential` (there is no trait to mock now). They can only mock the `CredentialTrait`.

Yeah. I thought only traits were mock-able.

I thought even if `auth::Credential` implements the `CredentialTrait`, there is no way to actually make a concrete `Credential` from something that mocks `CredentialTrait` unless we expose some new ctor / conversion function. (Hence #614 and subsequently this PR.)

Am I missing something?

coryan:
> 
> Am I missing something?

I could have a function:

```rust
fn foo(creds: impl CredentialTrait) -> Magic { .... }
```

I can feed this function anything that implements the credential traits. Both the real `auth::Credential` type and a mock. Or a completely weird implementation that the customer dreams up.

I am not sure that is meaningfully different from 

```rust
foo(Credentials::from(/* thing that implements CredentialTrait */))
```

dbolduc:
> > Am I missing something?
> 
> I could have a function: ...

Ah, thanks. I am going to go forward with this. It seems like the most straight forward way to do things.

Issue #438, opened 2024-12-16
AIP-151: Support helpers for LROs


coryan:
Many Google APIs use longrunning operations ([aip.dev/151](https://google.aip.dev/151)). The client libraries should have a helper function that turns the LRO into a stream which automatically polls (with backoff and some timeout), returns the intermediate metadata results, and on completion extracts the result and/or error.

- [x] #561 
- [x] #617 
- [x] #640
- [x] #683
- [x] #688
- [x] #684
- [x] generator creates `poller()` helper function
- [x] Add integration test using `poller()` to `src/integration-test`
- [x] Add polling backoff policy to gax
- [x] Add polling policy to gax
- [x] #714
- [x] #713
- [x] #716
- [x] #715

coryan:
Renamed for consistency with other AIP-related issues

PR #637, opened 2025-01-07
fix: serialization of `oneof` fields

### LINKED ISSUE #634: Bad enum generated for `oneof` field of string type
This code:

https://github.com/googleapis/google-cloud-rust/blob/5cd390d9d4bc0c204d67e836e577b31a9e7fc271/src/generated/cloud/workflows/v1/src/model.rs#L369

serializes to:

```
"{\"workflow\":{\"state\":\"\",\"callLogLevel\":\"\",\"sourceContents\":{\"source_contents\":\"# Test only workflow\\nmain:\\n    steps:\\n        - sayHello:\\n            return: Hello World\\n\"}}}"
```

Note the nested (and incorrectly spelled)  `"sourceContents": {"source_contents"`

The code comes from:

https://github.com/googleapis/googleapis/blob/2884323d6cc362b0b417e9bcc1c823243de63549/google/cloud/workflows/v1/workflows.proto#L227-L230

coryan:
First, they were always serialized, even when not present.  That took a
small change in the mustache template.

Second, fields of primitive types (such as strings) were serialized as
a nested JSON object.  That took a small change in the Rust codec.

Last, we did not have any test to verify serde behaves as we think it
does. It is not that serde would be broken, is that we may be using it
incorrectly.

Fixes #634


codyoss (approved):

coryan:
For the record, this was motivated by #438

PR #639, opened 2025-01-07
fix(wkt): extract `Empty` from `Any`

coryan:
`Duration`, `Timestamp`, and `FieldMask` need special code to extract
from `Any`. But `Empty` does not, the exception to the exception I
guess.


codyoss (approved):
TIL

coryan:
For the record, this was motivated by #438

PR #642, opened 2025-01-08
feat(wkt): validate message typenames

### LINKED ISSUE #98: Rust messages need a typename to serialize `.google.protobuf.Any`

coryan:
Fixes #98

### DIFF
--- a/src/wkt/Cargo.toml
+++ b/src/wkt/Cargo.toml
@@ -29,7 +29,7 @@ time   = []
 
 [dependencies]
 serde      = { version = "1.0.217", features = ["serde_derive"] }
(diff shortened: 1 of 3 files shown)


codyoss (approved):
Nice!

PR #648, opened 2025-01-08
cleanup(gax): move `convert_headers()`

coryan:
The function to convert HTTP headers into something we can use in
`HttpError` was used only in `http_client/mod.rs`, so move it there and
make it private.

In the process, I noticed we were printing sensitive headers, so I
changed the code to avoid doing so. I am not sure if any incoming
headers are labeled sensitive, but it does not hurt to prevent
embarrassing leaks.

I had to move some test code, which motivated some changes to it:
instead of testing that we can create an `HttpError` from a response,
verify that the `ReqwestClient` does this and verify its output.

### DIFF
--- a/src/gax/src/error/http_error.rs
+++ b/src/gax/src/error/http_error.rs
@@ -73,20 +73,6 @@ impl std::fmt::Display for HttpError {
 
 impl std::error::Error for HttpError {}
 
-#[cfg(feature = "unstable-sdk-client")]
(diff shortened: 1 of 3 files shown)


coryan:
The golang dependency is fixed in https://github.com/googleapis/google-cloud-rust/pull/649

codyoss (approved):

PR #654, opened 2025-01-09
cleanup(gax): remove unused function

coryan:
`StatusDetails::default()` is not used anywhere, and I think it is
unlikely to be used.  The `StatusDetails` should only be initialized
if once has a valid `Any` or a valid error detail type.

### DIFF
--- a/src/gax/src/error/rpc/mod.rs
+++ b/src/gax/src/error/rpc/mod.rs
@@ -381,12 +381,6 @@ pub enum StatusDetails {
     Other(wkt::Any),
 }
 
-impl Default for StatusDetails {
-    fn default() -> Self {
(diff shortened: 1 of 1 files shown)


codyoss (approved):
The only place where this is potentially nice to have is there is a new details starting to be returned tmr that we don't yet support. I *assume with default things would searilze to other for awhile instead of being a hard error.

coryan:
> The only place where this is potentially nice to have is there is a new details starting to be returned tmr that we don't yet support. I *assume with default things would searilze to other for awhile instead of being a hard error.

Yeah, any new detail should get represented as a `StatusDetails::Other`, and the `wkt::Any` contained therein could be used to get the values out (or at least print something vaguely useful).

Issue #651, opened 2025-01-09
User Credential needs to support Quota Project


sai-sunder-s:
When reading the user credential json, `quota_project_id` field should be read as well. If this value is provided, then it should be added to the `x-goog-user-project` header during the `get_headers()` call.

PR #656, opened 2025-01-09
refactor(auth): prepare to support quota project in UC

dbolduc:
Motivated by #651 

We will want `AuthorizedUser` to hold the `quota_project_id`. It does not belong to the `UserTokenProvider`. So going directly from `serde_json::Value` -> `UserTokenProvider` doesn't make sense.

Add a test for missing optional fields in the user credentials json.

And some minor cleanups around local variable names.

### DIFF
--- a/src/auth/src/credentials/user_credential.rs
+++ b/src/auth/src/credentials/user_credential.rs
@@ -26,7 +26,14 @@ use time::OffsetDateTime;
 const OAUTH2_ENDPOINT: &str = "https://oauth2.googleapis.com/token";
 
 pub(crate) fn creds_from(js: serde_json::Value) -> Result<Credential> {
-    let token_provider = UserTokenProvider::from_json(js)?;
+    let au = serde_json::from_value::<AuthorizedUser>(js)
+        .map_err(|e| CredentialError::new(false, e.into()))?;
+    let token_provider = UserTokenProvider {
+        client_id: au.client_id,
+        client_secret: au.client_secret,
+        refresh_token: au.refresh_token,
+        endpoint: OAUTH2_ENDPOINT.to_string(),
+    };
 
     Ok(Credential {
         inner: Arc::new(UserCredential { token_provider }),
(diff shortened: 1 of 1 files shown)


sai-sunder-s (approved):

src/auth/src/credentials/user_credential.rs
```diff
@@ -26,7 +26,14 @@ use time::OffsetDateTime;
 const OAUTH2_ENDPOINT: &str = "https://oauth2.googleapis.com/token";
 
 pub(crate) fn creds_from(js: serde_json::Value) -> Result<Credential> {
-    let token_provider = UserTokenProvider::from_json(js)?;
+    let au = serde_json::from_value::<AuthorizedUser>(js)
+        .map_err(|e| CredentialError::new(false, e.into()))?;
```
coryan, line 30: drive-by: I think if if you had `CredentialError::non_retryable<T: Into<...>>(e: T)` you could write:

```rust
.map_err(CredentialError::non_retryable)
```

You could make it `pub(crate)` if you don't want to grow your public API.

src/auth/src/credentials/user_credential.rs
```diff
@@ -247,7 +260,9 @@ mod test {
             let mut json = json_full.clone();
             // Remove a required field from the JSON
             json[required_field].take();
-            UserTokenProvider::from_json(json).err().unwrap();
+            serde_json::from_value::<AuthorizedUser>(json)
```
coryan, line 263: maybe

```rust
  assert!(serde_json::from_value::<AuthorizedUser>(json).is_err(), "{json:?}");
```

is marginally more readable, it tells me you are asserting something vs. indirectly asserting something because the error did not fail.

coryan (commented):
I know you merged already, none of these comments are important.

src/auth/src/credentials/user_credential.rs
dbolduc: Yeah, that is much nicer. I tried this once when I was 3-weeks worse at Rust. Now I can make it work.

PR #661, opened 2025-01-09
feat(auth): support quota project in UserCredentials

### LINKED ISSUE #651: User Credential needs to support Quota Project

dbolduc:
Fixes #651

Parse `quota_project_id` if it is part of an Authorized User Credential in ADC.

Include the quota project in requests as the `x-goog-user-project` header.

### DIFF
--- a/src/auth/src/credentials/user_credential.rs
+++ b/src/auth/src/credentials/user_credential.rs
@@ -13,8 +13,7 @@
 // limitations under the License.
 
 use crate::credentials::dynamic::CredentialTrait;
-use crate::credentials::Credential;
-use crate::credentials::Result;
+use crate::credentials::{Credential, Result, QUOTA_PROJECT_KEY};
 use crate::errors::{is_retryable, CredentialError};
 use crate::token::{Token, TokenProvider};
 use http::header::{HeaderName, HeaderValue, AUTHORIZATION, CONTENT_TYPE};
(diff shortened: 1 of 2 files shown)


src/auth/src/credentials/user_credential.rs
```diff
@@ -134,7 +137,15 @@ where
         let mut value = HeaderValue::from_str(&format!("{} {}", token.token_type, token.token))
             .map_err(|e| CredentialError::new(false, e.into()))?;
         value.set_sensitive(true);
-        Ok(vec![(AUTHORIZATION, value)])
+        let mut headers = vec![(AUTHORIZATION, value)];
+        if let Some(project) = &self.quota_project_id {
+            headers.push((
+                HeaderName::from_static(QUOTA_PROJECT_KEY),
+                HeaderValue::from_str(project)
+                    .map_err(|e| CredentialError::new(false, e.into()))?,
```
coryan, line 145: Maybe as a future improvement: you could do this once when the credentials are created. Or cache the results on the first attempt and keep them from that point.

dbolduc: I considered that, but I did not think there were any savings. If we cache the results, we still need to copy them.

The only thing I see is that we could reserve capacity in the vector, knowing we will add extra headers. That seemed negligible to me. :shrug:

Maybe I am missing something?

dbolduc: (Aside) note to self: I should clean up the `map_err(...)` after #662

coryan (approved):

src/auth/src/credentials/user_credential.rs
coryan: I think you are right.

PR #662, opened 2025-01-09
refactor(auth): simplify error creation

dbolduc:
Introduce crate private error helpers, as suggested by @coryan in https://github.com/googleapis/google-cloud-rust/pull/656/files#r1909410273 and probably other places.

There is a different issue to rethink whether some of these errors are retryable or not. I am not dealing with that in this PR. I want this PR to be a pure refactor.


coryan (approved):

src/auth/src/credentials/user_credential.rs
```diff
@@ -81,15 +81,12 @@ impl TokenProvider for UserTokenProvider {
         let resp = builder
             .send()
             .await
-            .map_err(|e| CredentialError::new(false, e.into()))?;
+            .map_err(CredentialError::non_retryable)?;
```
sai-sunder-s, line 84: why is this error not retryable?

if it is a connection error or something like that, it should be possible to retry?

src/auth/src/credentials/user_credential.rs
```diff
@@ -81,15 +81,12 @@ impl TokenProvider for UserTokenProvider {
...
 
         // Process the response
         if !resp.status().is_success() {
             let status = resp.status();
-            let body = resp
-                .text()
-                .await
-                .map_err(|e| CredentialError::new(false, e.into()))?;
+            let body = resp.text().await.map_err(CredentialError::non_retryable)?;
```
sai-sunder-s, line 89: It could be a 500 or other retryable error here?

src/auth/src/credentials/user_credential.rs
dbolduc: Made retryable in #700

PR #669, opened 2025-01-12
impl(language): remove Codec interface

julieqiu:
As explained in https://github.com/googleapis/google-cloud-rust/pull/667, this PR removes the `Codec` interface.

This change introduces some duplicate code in `GenerateClient` and between `GoTemplateData` and `RustTemplateData`, which will be cleaned up in follow-up PRs.

For now, `GoTemplateData` and `RustTemplateData` and their respective constructor functions are identical, differing only in the `Go` and `Rust` prefixes in their names.


generator/internal/language/client.go
```diff
@@ -33,17 +33,49 @@ func (p *mustacheProvider) Get(name string) (string, error) {
...
+		if err := codec.Validate(model); err != nil {
+			return err
+		}
+		data := newRustTemplateData(model, codec)
+		context = append(context, data)
+		if languageContext := codec.AdditionalContext(model); languageContext != nil {
+			context = append(context, languageContext)
+		}
+		provider = codec.TemplatesProvider()
+		generatedFiles = codec.GeneratedFiles()
```
coryan, line 57: Maybe this is already part of the cleanup PRs, but it feels that this should become:

```
     case "rust":
        return GenerateRust(outdir, options, model)
```

maybe with the code starting at line 78 refactored to a helper function?

coryan (approved):

generator/internal/language/client.go
julieqiu: yup, will do in a follow up PR

PR #671, opened 2025-01-12
impl(language): delete unused Go methods

julieqiu:
Now that the Go codec does not need to satisfy the Codec interface, many
methods can now be deleted.


generator/internal/language/golang.go
```diff
@@ -149,12 +134,8 @@ func (c *goCodec) fieldType(f *api.Field, state *api.APIState) string {
 	return out
 }
 
-func (c *goCodec) primitiveFieldType(f *api.Field, state *api.APIState) string {
-	return c.fieldType(f, state)
-}
-
 func (c *goCodec) asQueryParameter(f *api.Field, _ *api.APIState) string {
-	return fmt.Sprintf("req.%s.to_str()", c.toCamel(f.Name))
+	return fmt.Sprintf("req.%s.to_str()", strcase.ToLowerCamel(f.Name))
```
coryan, line 138: unrelated to this PR: protobuf fields named `package`, or `switch` may be a problem, they should be escaped to avoid a conflict. But this is not a new problem.

coryan (approved):

PR #677, opened 2025-01-13
impl(language): convert Rust codec methods to functions

julieqiu:
Convert several Rust codec methods to functions, as a step towards removing the Rust codec.

### DIFF
--- a/generator/internal/language/client.go
+++ b/generator/internal/language/client.go
@@ -41,14 +41,14 @@ func GenerateClient(model *api.API, language, outdir string, options map[string]
 	)
 	switch language {
 	case "rust":
-		codec, err := newRustCodec(outdir, options)
+		codec, err := newRustCodec(options)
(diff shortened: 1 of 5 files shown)


coryan (approved):
Some of the functions have lots of arguments, I assume we are Okay with that?

And as you probably noticed, you will need to rebase. Thanks for letting me merge first.

julieqiu:
> Some of the functions have lots of arguments, I assume we are Okay with that?

It is possible we may have to create a new struct later, but I don't think rustCodec is the right one as it is. I also wonder if some of the functions themselves need to be rewritten (for example, some of the validate functions actually set fields, which seems unexpected). It was easier to me to just put all the arguments in the functions as is for now, and think about that next. 
 
> 
> And as you probably noticed, you will need to rebase. Thanks for letting me merge first.

🫡

PR #685, opened 2025-01-13
feat(lro): new crate for LRO helper support

### LINKED ISSUE #683: new crate (`gcp-sdk-lro`) with support types for LROs

coryan:
Introduce a new crate to support LRO helpers. The key abstraction is
`gcp_sdk_lro::Poller`: a type to poll LROs and automatically decode the
metadata and response types.

Fixes #683

### DIFF
--- a/src/lro/src/lib.rs
+++ b/src/lro/src/lib.rs
@@ -273,4 +273,56 @@ mod test {
         let p2 = poller.poll().await;
         assert!(p2.is_none(), "{p2:?}");
     }
+
+    #[tokio::test(flavor = "multi_thread")]
+    async fn poll_basic_stream() {
+        let start = || async move {
+            let any = wkt::Any::try_from(&wkt::Timestamp::clamp(123, 0))
+                .map_err(|e| Error::other(format!("unexpected error in Any::try_from {e}")))?;
+            let op = longrunning::model::Operation::default()
+                .set_name("test-only-name")
+                .set_metadata(any);
+            let op = TestOperation::new(op);
+            Ok::<TestOperation, Error>(op)
+        };
+
+        let query = |_: String| async move {
+            let any = wkt::Any::try_from(&wkt::Duration::clamp(234, 0))
+                .map_err(|e| Error::other(format!("unexpected error in Any::try_from {e}")))?;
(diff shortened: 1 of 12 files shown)


src/lro/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+
+    fn name(&self) -> String;
+    fn done(&self) -> bool;
+    fn metadata(&self) -> Option<&wkt::Any>;
+    fn response(&self) -> Option<&wkt::Any>;
+    fn error(&self) -> Option<&rpc::model::Status>;
+}
+
+/// Implements [Operation] using [longrunning::model::Operation].
+pub struct TypedOperation<R, M> {
```
codyoss, line 67: nit: could this be called Operation, and the trait be renamed to something like `LongrunningOperation`?

src/lro/src/lib.rs
```diff
@@ -0,0 +1,392 @@
...
+///   are captured by this function.
+/// - `SF` - the type of future returned by `S`.
+/// - `Q` - the query closure. Queries the status of the LRO created by `start`.
+///   It receives the name of the operation as its only input parameter. It
+///   should have captured any stubs and request options.
+/// - `QF` - the type of future returned by `Q`.
+pub struct PollerImpl<O, S, SF, Q, QF>
+where
+    O: Operation,
+    S: FnOnce() -> SF + Send + Sync,
```
codyoss, line 141: I think the reason for making this take a closure is mainly so we can honor any request options -- is that correct? If so, should we have an API similar to our clients where they can just be explicitly passed in? That way you would not need to capture any state.

What I am trying to tease out is: Is this more of a helper utility or is this a lightweight veneer over LROs that anyone could pickup and use. If the former I think this is fine, but if the later I would say this would be harder for someone to pickup and use.

codyoss (commented):
A couple of questions. Overall I am fine with this approach, but want to make sure I follow.

src/lro/src/lib.rs
codyoss: Thanks for the explanation, that makes sense to me.

codyoss (approved):
Thanks for the changes. I think hidding more of the details helped me understand more.

PR #700, opened 2025-01-14
impl(auth): more retrying?

dbolduc:
Do a pass on retrying, mainly in User Credentials.

Retry I/O errors: `builder.send()`, `response.text()`

My read of `reqwest::Response::json<>` is that we should retry if `reqwest::Response::bytes()` fails, but we should not retry if the deserialization into json fails.

https://docs.rs/reqwest/0.12.12/src/reqwest/async_impl/response.rs.html#268-270

### DIFF
--- a/src/auth/src/credentials/user_credential.rs
+++ b/src/auth/src/credentials/user_credential.rs
@@ -78,24 +78,24 @@ impl TokenProvider for UserTokenProvider {
             .request(Method::POST, self.endpoint.as_str())
             .header(CONTENT_TYPE, header)
             .json(&req);
-        let resp = builder
-            .send()
-            .await
-            .map_err(CredentialError::non_retryable)?;
+        let resp = builder.send().await.map_err(CredentialError::retryable)?;
 
         // Process the response
         if !resp.status().is_success() {
             let status = resp.status();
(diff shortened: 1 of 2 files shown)


src/auth/src/credentials/user_credential.rs
```diff
@@ -78,24 +78,24 @@ impl TokenProvider for UserTokenProvider {
...
             ));
         }
-        let response = resp
-            .json::<Oauth2RefreshResponse>()
-            .await
-            .map_err(CredentialError::non_retryable)?;
+        let response = resp.json::<Oauth2RefreshResponse>().await.map_err(|e| {
+            let retryable = !e.is_decode();
+            CredentialError::new(retryable, e.into())
+        })?;
```
coryan, line 98: I am not sure if this helps, but you could implement the `From` trait:

```rust
impl std::convert::From<reqwest::Error> for CredentialError {
  fn from(value: reqwest::Error) -> Self {
    // Do the needful
  }
}
```


And then you can write (I think) `.map_err(CredentialError::from)`.  The `from` function is also easier to test.  But whether it works in all these cases? 🤷

coryan (approved):

PR #711, opened 2025-01-16
ci(gcb): use `sccache` and GCS to speed up build

### LINKED ISSUE #288: Enable `sccache` for the integration tests
The integration tests on GCB (Google Cloud Build) take about 4 minutes to build, and about 5 seconds to run. That is about 3 minutes longer than the builds on GHA (GitHub Actions).

I think the difference is due to caching: GHA builds keep a cache, but we don't have one in GCB. We can use [sccache](https://github.com/mozilla/sccache) to keep a content-based build for GCB. 

Right now the difference is not very big, but it will get bigger as we add more clients to the project.

coryan:
Fixes #288 The time is around 3m vs. 5m

### DIFF
--- a/.gcb/builds/grants/main.tf
+++ b/.gcb/builds/grants/main.tf
@@ -16,6 +16,10 @@ variable "project" {
   type = string
 }
 
+variable "build_cache" {
+  type = string
+}
+
 data "google_project" "project" {
 }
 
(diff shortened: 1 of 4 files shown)


codyoss (approved):
TIL sccache

PR #723, opened 2025-01-17
feat(lro): use polling policy on success

coryan:
We want to stop the polling loop if it has been running for too long, or
it has made too many polling attempts.

Part of the work for #713

### DIFF
--- a/src/lro/src/details.rs
+++ b/src/lro/src/details.rs
@@ -49,7 +49,20 @@ where
             let state = polling_policy.on_error(loop_start, attempt_count, e);
             handle_polling_error(state, operation_name)
         }
-        Ok(op) => handle_common(op),
+        Ok(op) => {
+            let (name, result) = handle_common(op);
+            match &result {
+                PollingResult::Completed(_) => (name, result),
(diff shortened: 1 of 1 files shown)


src/lro/src/details.rs
```diff
@@ -49,7 +49,20 @@ where
...
+            match &result {
+                PollingResult::Completed(_) => (name, result),
+                PollingResult::InProgress(_) => {
+                    match polling_policy.on_in_progress(loop_start, attempt_count, &operation_name)
+                    {
+                        None => (name, result),
+                        Some(e) => (None, PollingResult::Completed(Err(e))),
+                    }
+                }
+                PollingResult::PollingError(_) => panic!(),
```
codyoss, line 63: I know this can't be hit with how the code is currently written, but panics still scare me a little. If we are going to panic, can you put a message in the panic so we know where this is coming from if it is ever hit.

codyoss (approved):

src/lro/src/details.rs
coryan: Done.  Good call.

PR #725, opened 2025-01-17
impl(config): move sidekick.Config to separate package

julieqiu:
Refactor the logic for creating and parsing the `sidekick.toml` config file into a separate package. This change will prevent a dependency loop in a future PR when introducing a `Config` struct to the `sample` package for testing.


generator/internal/config/config.go
```diff
@@ -0,0 +1,273 @@
+// Copyright 2024 Google LLC
```
coryan, line 1: nit: it is 2025....

coryan: Ah, you moved the file, but git thinks it is not a move for whatever reason.  I think 2024 is fine in this case.

coryan (approved):

Issue #655, opened 2025-01-09
Improve PathInfo to link the field and messages involved


coryan:
In `api.PathInfo` we just name the fields used as variables.  `sidekick` could link the fields (as a `*api.Field`) and messages involved.  Recursively for variables with dot separated subfields (like `field.subfield`).

The codecs can use this information to create cleaner code.

Thanks to @dangazineu for this suggestion.

Issue #653, opened 2025-01-09
Skip the `templatedata.go` intermediate step


coryan:
`sidekick` parses Protobuf or OpenAPI (and whatever comes next) and creates a `api.API` model of the service specification.

Then we ask the target language Codec (currently for Rust or Go) a bunch of questions and create the model in `templatedata.go`.  And then we use the same mustache templates that the Codec creates to generate the code.

Maybe a cleaner approach would be to pass the `api.API` model directly to the Codec and let it call the mustache templates.  The Codec can create its own enriched[^1] model of the API if needed.

@dangazineu suggested the good parts of this issue request, the bad ideas came from somebody else (not me, of course).

[^1]: I think that is what the cool kids call them.

dangazineu:
This was mostly done as part of #667 and #669. We still have the problem that the enriched language types re-implementations of the `api` types, which discourages advancing our model.  For example, #655 calls for `PathInfo` to have references to the fields and messages that each path segment requires, but these are useless in the mustache templates, as they don't know the language-specific naming conventions. So any improvements made to `PathInfo` would have to be replicated for each language implementation. 

We could move `model.go` into the `language` package, and have each language template contribute their own methods to the `api` types directly, but that seems like a hack and bound to get out of hand quickly.

I think the solution is probably to invert the control between the language types and the model, so instead of using the model to construct the language types, we pass the language-specific logic to the model structs as adapter interfaces, and use the model types directly in the templates.

coryan:
> and use the model types directly in the templates.

We talked about this with Julie.  That is the plan.  Maybe with some annotations introduced by each language.

For now, you can change `rusttemplate.go` to add the annotations in `PathInfo` that are needed.  At least it is easier now that one needs to worry about only one language at a time.

Issue #610, opened 2025-01-06
Support enum types as query parameters


coryan:
Some services use enum types as query parameters:

https://github.com/googleapis/googleapis/blob/3cef095370b3176f67542e527ffde81e019df091/google/cloud/webrisk/v1/webrisk.proto#L210

https://github.com/googleapis/googleapis/blob/3cef095370b3176f67542e527ffde81e019df091/google/cloud/webrisk/v1/webrisk.proto#L268

Related to #482

Issue #482, opened 2024-12-19
Generate valid Rust code for the showcase protos


coryan:
sidekick does not produce valid Rust for the [gapic-showcase](https://github.com/googleapis/gapic-showcase) protos.  There are a number of problems:

- [x] #792
- [x] #793
- [ ] We do not support streaming RPCs.
- [x] We are missing the dependency for `google.longrunning`
- [x] We require HTTP annotations on all RPCs, but streaming RPCs often do not have them.
- [x] #610.
- [x] Handle enum values as path parameters.
- [x] Handle complex path parameters such as [this](https://github.com/googleapis/gapic-showcase/blob/8b39504dce40abd80b0967dee1e7c2da34cd5c6e/schema/google/showcase/v1beta1/compliance.proto#L72)

I think we should try to get more real services to work, but checking what does not work from time to time may help us prioritize problems.

coryan:
I am going to close this, we have an specific bug for streaming RPCs (#2318).

PR #729, opened 2025-01-17
cleanup(generator): annotate `api.Field` types

coryan:
Start a cleanup of the `rusttemplate.go` file. The idea is to minimize
duplication with the `api.*` types. Instead we attach a number of
annotations to `api.Field` that represent the Rust-specific thing.

Other languages will attach different annotation objects. The type of
the annotation is `any` to support this case.

Part of the work for #653, helps with #655, and from there #610 and #482

### DIFF
--- a/generator/internal/language/templates/rust/common/message.mustache
+++ b/generator/internal/language/templates/rust/common/message.mustache
@@ -24,41 +24,38 @@ limitations under the License.
 pub struct {{Name}} {
     {{#BasicFields}}
 
-    {{#DocLines}}
+    {{#Codec.DocLines}}
     {{{.}}}
-    {{/DocLines}}
-    {{#FieldAttributes}}
+    {{/Codec.DocLines}}
+    {{#Codec.Attributes}}
     {{{.}}}
-    {{/FieldAttributes}}
-    pub {{NameToSnake}}: {{{FieldType}}},
+    {{/Codec.Attributes}}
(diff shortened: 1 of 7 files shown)


coryan:
@codyoss @dangazineu you both are interested in this.... What I am proposing here is to (slowly) refactor the generator to use `api.*` directly, with some annotations that will be language specific.

As @dangazineu has pointed out, then we can apply transformations and analysis in the `api.*` model, e.g., field the fields that are needed for path parameters, or the type of a query parameter (and not just its name).

If this is a terrible idea, we can stop here.  Otherwise, I can merge this and then move on to messages, methods, services, etc.

generator/internal/language/templates/rust/common/message.mustache
```diff
@@ -24,41 +24,38 @@ limitations under the License.
 pub struct {{Name}} {
     {{#BasicFields}}
 
-    {{#DocLines}}
+    {{#Codec.DocLines}}
```
dangazineu, line 27: neat!

dangazineu:
I like the direction this is going. I think for now, `any` as the type of the Codec makes sense, but I wonder if in the future we'd like to have a well defined set of things we expect to be language-specific. The only reason I think this may be useful is if we want to extract the generator out of the rust repo (so it supports more languages) but keep the templates here (so the language-specific changes happen atomically).

julieqiu (approved):

PR #739, opened 2025-01-18
cleanup: error handling for query parameters

### LINKED ISSUE #610: Support enum types as query parameters

coryan:
Only the conversion from a message type to a `serde_json::Value` can
result in an error. If we convert first, then `gax::query_parameter`
can work without errors, can operate on values (vs. references).

Fixes #610

Normally I try to avoid mixing generator and gax changes in the same PR, but this would require 3 or 4 PRs otherwise.


generator/internal/language/rust.go
```diff
@@ -540,9 +540,9 @@ func rustAddQueryParameter(f *api.Field) string {
 		// few requests use nested objects as query parameters. Furthermore,
 		// the conversion is skipped if the object field is `None`.`
 		if f.Optional || f.Repeated {
-			return fmt.Sprintf(`let builder = req.%s.iter().try_fold(builder, |builder, p| { use gax::query_parameter::QueryParameter; serde_json::to_value(p).map_err(Error::serde)?.add(builder, "%s").map_err(Error::other) })?;`, rustToSnake(f.Name), f.JSONName)
+			return fmt.Sprintf(`let builder = req.%s.as_ref().map(|p| serde_json::to_value(p).map_err(Error::serde) ).transpose()?.into_iter().fold(builder, |builder, v| { use gax::query_parameter::QueryParameter; v.add(builder, "%s") });`, rustToSnake(f.Name), f.JSONName)
```
codyoss, line 543: TIL: transpose

codyoss (approved):
Nice, I like the cleaner API!

PR #744, opened 2025-01-19
feat: Enhanced the model so path parameters and the fields they reference can be used together in the template

dangazineu:
This is a proposed solution for #655. I introduced a visitor interface to the model, and used it to initialize elements after the parser is finished (because not all elements are available during parsing) .

My goal was to keep as much of the data structure and logic either in the model, or in the templates.


generator/internal/api/model.go
```diff
@@ -226,6 +435,101 @@ type Message struct {
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (i InitializationVisitor) VisitMessage(m *Message) error {
+	m.Elements = make(map[string]*MessageElement)
+	for _, f := range m.Fields {
+		if f.Typez == MESSAGE_TYPE {
```
dangazineu, line 464: It is not clear to me why this is the case, but not all messages are available in the `API.Messages` slice, and some are in `API.Fields` instead.

generator/testdata/rust/protobuf/golden/iam/v1/src/transport.rs
```diff
@@ -47,8 +47,12 @@ impl crate::traits::IAMPolicy for IAMPolicy {
     ) -> Result<crate::model::Policy> {
         let options = options.set_default_idempotency(reqwest::Method::POST.is_idempotent());
         let builder = self.inner.builder(
-            reqwest::Method::POST, format!("/v1/{}:setIamPolicy"
-               , req.resource
+            reqwest::Method::POST,
+            format!("/v1/{}:setIamPolicy"
+            ,
+                req
+                            .resource
```
dangazineu, line 54: I couldn't find a way to skip white spaces in the template. Any suggestion? 
At least this only affects the test data, `cargo fmt` takes care of this for the generated code.

src/generated/cloud/secretmanager/v1/src/transport.rs
```diff
@@ -135,8 +135,11 @@ impl crate::traits::SecretManagerService for SecretManagerService {
                 reqwest::Method::PATCH,
                 format!(
                     "/v1/{}",
-                    gax::path_parameter::PathParameter::required(&req.secret, "secret")
-                        .map_err(Error::other)?
+                    req.secret
+                        .as_ref()
+                        .ok_or_else(|| gax::error::Error::other(
+                            "field \"secret\" is required in a request, but was missing"
```
dangazineu, line 141: I'll revisit this code to include the name of the request in the error message

generator/internal/api/model.go
coryan: The API.Messages slice contains all the *top-level* (non-nested) messages we need to generate. Some messages are imported (and do not need to be generated), and some are children of the top-level messages (and need recursing via `api.Message.Messages`:

https://github.com/googleapis/google-cloud-rust/blob/d1348119749a52e9bdc48267eca3d1746d330250/generator/internal/api/model.go#L216

generator/internal/language/templates/rust/crate/src/transport.rs.mustache
```diff
@@ -56,10 +56,21 @@ impl crate::traits::{{NameToPascal}} for {{NameToPascal}} {
...
+                {{#Components}}
+                    {{#Reference}}
+                        {{#Optional}}
+                            .{{Codec.FieldName}}.as_ref().ok_or_else(|| gax::error::Error::other("field \"{{Codec.FieldName}}\" is required in a request, but was missing"))?
+                        {{/Optional}}
+                        {{^Optional}}
+                            .{{Codec.FieldName}}
+                        {{/Optional}}
+                    {{/Reference}}
+                {{/Components}}
```
coryan, line 72: Mustache preserves the indentation of your sections, so don't use indentation unless you want this to be in the generated code.  This may be closer to what we want:

```suggestion
                req
                {{#Components}}
                {{#Reference}}
                    {{#Optional}}
                    .{{Codec.FieldName}}.as_ref().ok_or_else(|| gax::error::Error::other("field \"{{Codec.FieldName}}\" is required in a request, but was missing"))?
                    {{/Optional}}
                    {{^Optional}}
                    .{{Codec.FieldName}}
                    {{/Optional}}
                {{/Reference}}
                {{/Components}}
```

generator/internal/language/templates/rust/crate/src/transport.rs.mustache
```diff
@@ -56,10 +56,21 @@ impl crate::traits::{{NameToPascal}} for {{NameToPascal}} {
...
-               , {{{.}}}
-               {{/HTTPPathArgs}}
+            reqwest::Method::{{HTTPMethod}},
+            format!("{{#HTTPPathInfo.PathTemplate}}{{#Literal}}/{{Value}}{{/Literal}}{{#FieldPath}}/{}{{/FieldPath}}{{#Verb}}:{{Value}}{{/Verb}}{{/HTTPPathInfo.PathTemplate}}"
+            {{#HTTPPathInfo.PathTemplate}}{{#FieldPath}},
+                req
+                {{#Components}}
+                    {{#Reference}}
+                        {{#Optional}}
+                            .{{Codec.FieldName}}.as_ref().ok_or_else(|| gax::error::Error::other("field \"{{Codec.FieldName}}\" is required in a request, but was missing"))?
```
coryan, line 66: We can use a helper function to make this code smaller:

```suggestion
                            .{{Codec.FieldName}}
                                .as_ref()
                                .ok_or_else(|| gax::path_parameter::missing("{{Codec.FieldName}}")?
```

generator/testdata/rust/protobuf/golden/iam/v1/src/transport.rs
coryan: See above.

generator/internal/language/templates/rust/crate/src/transport.rs.mustache
coryan, line 72: Consider:

```suggestion
            reqwest::Method::{{HTTPMethod}},
            format!("{{#HTTPPathInfo.PathTemplate}}{{#Literal}}/{{Value}}{{/Literal}}{{#FieldPath}}/{}{{/FieldPath}}{{#Verb}}:{{Value}}{{/Verb}}{{/HTTPPathInfo.PathTemplate}}"{{#HTTPPathInfo.PathTemplate}} 
                 {{#FieldPath}}
                 , req{{#Components}}{{#Reference}}.{{Codec.FieldName}}
                     {{#Optional}}
                    .as_ref().ok_or_else(|| gax::path_parameters::missing("{{Codec.FieldName}}"))?
                     {{/Optional}}{{/Reference}}{{/Components}}
```

generator/internal/api/model.go
```diff
@@ -102,10 +198,18 @@ type Method struct {
 	Name string
 	// ID is a unique identifier.
 	ID string
-	// InputType is the input to the Method
+	// InputTypeID is the ID of the input to the Method, to be used with the API state to retrieve the message.
 	InputTypeID string
-	// OutputType is the output of the Method
+
+	//InputType is the input to the Method, it is only present after the Method struct has been initialized.
```
coryan, line 204:
```suggestion
	// InputType is the input to the Method, it is only present after the InitializationVisitor runs.
```

generator/internal/api/model.go
```diff
@@ -144,6 +276,47 @@ type PathInfo struct {
...
+			return err
+		}
+	}
+	return nil
+}
+
+func (i InitializationVisitor) VisitPathInfo(p *PathInfo) error {
+	for _, segment := range p.PathTemplate {
+		if segment.FieldPath != nil {
+			// Every FieldPath starts pointing to the root input type Message
```
coryan, line 302: Consider

```suggestion
		if segment.FieldPath == nil {
		    continue
		}
	        // Every FieldPath starts pointing to the root input type Message
```

To save ourselves som indentation.

generator/internal/api/model.go
```diff
@@ -41,6 +46,56 @@ const (
...
+}
+
+func (n NoOpVisitor) VisitPathSegment(s *PathSegment) error {
+	return nil
+}
+
+// InitializationVisitor is a Visitor used to initialize the API elements.
+type InitializationVisitor struct {
+	NoOpVisitor
+}
```
coryan, line 97: Consider  refactoring `InitializationVisitor` to its own file. It may be easier to grok what it does if we put it all in one place.

generator/internal/api/model.go
```diff
@@ -41,6 +46,56 @@ const (
...
+
+type Visitor interface {
+	VisitAPI(a *API) error
+	VisitMessage(m *Message) error
+	VisitService(s *Service) error
+	VisitMethod(m *Method) error
+	VisitMessageElement(m *MessageElement) error
+	VisitPathInfo(p *PathInfo) error
+	VisitPathSegment(s *PathSegment) error
+}
```
coryan, line 61: Consider refactoring to its own file.

generator/internal/api/model.go
```diff
@@ -226,6 +435,101 @@ type Message struct {
 	IsPageableResponse bool
 	// PageableItem is the field to be paginated over.
 	PageableItem *Field
+
+	// The API that this message belongs to.
+	API *API
+
+	// Elements is a map of all the elements in the message, keyed by their name.
+	// This field is only available after the message is initialized.
```
coryan, line 443: I find "initialized" a bit confusing.  What do you think of "cross-referenced" ?  And then it would be "CrossReferenceVisitor" instead of "InitializationVisitor"?

generator/internal/api/model.go
coryan: Maybe you should traverse the `API.State.MessagesByID` structure instead.

generator/internal/parser/protobuf.go
```diff
@@ -354,7 +354,7 @@ func makeAPIForProtobuf(serviceConfig *serviceconfig.Service, req *pluginpb.Code
 		result.Name = strings.TrimSuffix(serviceConfig.Name, ".googleapis.com")
 	}
 	updateMethodPagination(result)
-	return result
+	return result, result.Accept(&api.InitializationVisitor{})
```
coryan, line 357: I think this is too early to call `Accept`.  The codecs also introduce the well-known types:

https://github.com/googleapis/google-cloud-rust/blob/d1348119749a52e9bdc48267eca3d1746d330250/generator/internal/language/rust.go#L274-L278

And that happens much later:

https://github.com/googleapis/google-cloud-rust/blob/d1348119749a52e9bdc48267eca3d1746d330250/generator/internal/language/rusttemplate.go#L170

I think you want to run this visitor after the well-known types are loaded.

generator/internal/language/rust.go
```diff
@@ -653,73 +653,6 @@ func rustBodyAccessor(m *api.Method) string {
 	return "." + rustToSnake(m.PathInfo.BodyFieldPath)
 }
 
-func rustHTTPPathFmt(m *api.PathInfo) string {
```
coryan, line 656: Nice to see this go, but the golang code still looks like does most of the work in the codec?

coryan (commented):
First pass of reviews.  Don't worry too much about my comments about the mustache template files, I think they may help, but meh!  The structure of the generator code is what we need to make sure lands correctly.

generator/internal/api/model.go
dangazineu: I think I got it now, let me confirm:

1 - Every message that is declared at the root of a proto file will be in `API.Messages`, and `Message.Parent` will be `nil`; 

2 - When a message is declared inline within another message, it will be in the `Message.Messages` list, and its `Message.Parent` value will point to the message where it was declared. e.g.: 

https://github.com/googleapis/google-cloud-rust/blob/d1348119749a52e9bdc48267eca3d1746d330250/generator/testdata/googleapis/google/cloud/secretmanager/v1/resources.proto#L262-L266

3 - When a message declared elsewhere (either 1 or 2) is referenced within another message, it will be in the `Message.Fields` list, with `Field.Typez == MESSAGE_TYPE`. The `Message.Parent` field of that message will not be affected by this. e.g.:

https://github.com/googleapis/google-cloud-rust/blob/d1348119749a52e9bdc48267eca3d1746d330250/generator/testdata/googleapis/google/api/http.proto#L29-L33

4 - Every message, regardless of where it is declared, will be present in `API.State.MessagesByID`;

5 - A similar rule applies to enums and oneofs, with the different that they are not recursive, so their parent will always either be the message where they were declared, or `nil` if they were declared at the root of a proto.

If what I wrote above is true, I have to revise the implementation of the visitor navigation, as I may be leaving some elements unvisited, and changing some element's parents incorrectly. Should probably add some tests while I'm at it :)

generator/internal/language/templates/rust/crate/src/transport.rs.mustache
dangazineu: Nice, I'll add that!

generator/internal/api/model.go
dangazineu: I'll add a single `visitors.go` file to the api package for all the interfaces and the two visitor implementations I added.

generator/internal/api/model.go
coryan: I think (1), (2), (4) are correct.  Minor  corrections on (3) and (5):

> 3 - When a message declared elsewhere (either 1 or 2) is referenced within another message, it will be in the Message.Fields list,

A message may also appear *only* as a method input or output field.  Consider mixins, where we import request and response messages from an external API.  Services with this mixin gain this method (amongst others): `ListLocations(google.cloud.location.ListLocationRequest) return (google.cloud.location.ListLocationResponse)`

The request and response messages here only appear in the method definition and in the `MessagesByID`.  You cannot find them by navigating from `API.Messages`.


> 5 - A similar rule applies to enums and oneofs, with the different that they are not recursive, so their parent will always either be the message where they were declared, or nil if they were declared **at the root of a proto**.

Oneofs may also appear only in messages defined by imported messages (this is the case for the `longrunning` mixin).  The best way to find them may be to navigate via `MessageByID`.

I think in theory a mixin may introduce an enum without a parent, though in practice none of the standard mixins does this.

coryan: We should probably document this as comments in the model.go file.

dangazineu: Thank you! That clarifies a lot! I'll update the code later later today and document this as suggested.

generator/internal/api/model.go
dangazineu: Changed the name and documentation

generator/internal/api/model.go
dangazineu: I've updated the model documentation with this info, and the code to reflect the proper type hierarchy.

generator/internal/language/templates/rust/crate/src/transport.rs.mustache
dangazineu: good catch! updated

generator/internal/parser/protobuf.go
dangazineu: Moved. I think we could refactor this later, to avoid repetition. For now I am running the visitor in both language templates (not worse than before though, as I was running it in two parsers).

generator/internal/language/rust.go
dangazineu: Yeah, I want to clean that up too. But I think I'll do it in another PR, WDYT?

generator/internal/language/templates/rust/crate/src/transport.rs.mustache
dangazineu: I couldn't figure out how to add the function directly under `gax::path_parameter`, so I made it `gax::path_parameter::errors::missing`.

src/gax/src/path_parameter.rs
```diff
@@ -35,10 +36,15 @@ where
     }
 }
 
-#[derive(thiserror::Error, Debug)]
```
dangazineu, line 38: This was conflicting with the gax::error::Error import, and seemed OK to replace.

generator/internal/api/model.go
```diff
@@ -240,6 +369,9 @@ type Enum struct {
 	Values []*EnumValue
 	// Parent returns the ancestor of this node, if any.
 	Parent *Message
+
```
coryan, line 372: nit: why the blank line when the other fields do not have it?

generator/internal/api/visitors.go
```diff
@@ -0,0 +1,316 @@
+// Copyright 2024 Google LLC
```
coryan, line 1: s/2024/2025/

generator/internal/api/visitors.go
```diff
@@ -0,0 +1,316 @@
...
+func (n NoOpVisitor) VisitPathInfo(p *PathInfo) error {
+	return nil
+}
+
+func (n NoOpVisitor) VisitPathSegment(s *PathSegment) error {
+	return nil
+}
+
+// CrossReferencingVisitor is a Visitor used cross-reference API elements.
+// This visitor needs to be invoked through the Traverse method, as it needs to keep track of the API being visited.
```
coryan, line 248:
```suggestion
// CrossReferencingVisitor populates the cross-references between API elements.
// This visitor needs to be invoked through the Traverse method, as it needs to keep track of the API being visited.
```

generator/internal/api/visitors.go
```diff
@@ -0,0 +1,316 @@
...
+type Element interface {
+	Accept(v *Visitor) error
+}
+
+// Visitor is an interface for visiting API elements, with each visitable element having its own Visit* method.
+//
+// Each node is visited before its children. Parent/child hierarchy is documented in api.API.
+//
+// Traversing the API with a visitor, even the NoOpVisitor, has the side effect of assigning the *.Parent and *.API
+// fields to the visited elements. This side effect is idempotent.
```
coryan, line 29: That is too surprising for a  NoOpVisitor.

generator/internal/api/visitors.go
```diff
@@ -0,0 +1,316 @@
...
+//
+// Note that as of now, not all API elements are visitable.
+// This was done mostly because there hasn't been a use case for it yet.
+type Visitor interface {
```
coryan, line 33: Do we have a second and third visitor in mind (beyond `CrossReferencingVisitor`)?  If not, I think the visitor pattern is overkill for this problem.  We are not getting enough reuse from the abstraction, and abstractions have costs.  it may be easier to replace this with a function (or series of functions) that operate on `API` and its related types.

generator/internal/api/model.go
```diff
@@ -226,6 +297,64 @@ type Message struct {
 	IsPageableResponse bool
 	// PageableItem is the field to be paginated over.
 	PageableItem *Field
+
+	// The API that this message belongs to.
+	API *API
+
+	// Elements is a map of all the elements in the message, keyed by their name.
+	// This field is only available after the message has been visited by the CrossReferencingVisitor.
+	Elements map[string]*MessageElement
```
coryan, line 306: Consider  `ElementsByName map[string]*MessageElement`.  That seems more consistent with the existing `API.State.*ByID` maps.

generator/internal/api/visitors.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+	for _, o := range m.OneOfs {
+		m.Elements[o.Name] = &MessageElement{OneOf: o, Parent: m}
+	}
+	return nil
+}
+
+func (v CrossReferencingVisitor) VisitPathInfo(p *PathInfo) error {
+	for _, segment := range p.PathTemplate {
+		if segment.FieldPath != nil {
```
coryan, line 297: Consider:

```suggestion
		if segment.FieldPath == nil {
		    continue
		}
```

In my opinion, deep indentation is harder to read.

generator/internal/api/visitors.go
```diff
@@ -0,0 +1,316 @@
...
+	for _, m := range a.Messages {
+		m.API = a
+		err = m.Accept(v)
+		if err != nil {
+			return err
+		}
+	}
+	// Then we visit the mixin types, as the easiest way to distinguish
+	// them from the local types is that their API field should remain nil.
+	for _, m := range a.State.MessageByID {
```
coryan, line 61: This is visiting the local messages twice.  probably fine if all visitors are idempotent, but maybe a bit wasteful?

generator/internal/api/visitors.go
```diff
@@ -0,0 +1,316 @@
...
+	}
+
+	for _, e := range a.Enums {
+		e.API = a
+		err = e.Accept(v)
+		if err != nil {
+			return err
+		}
+	}
+	for _, e := range a.State.EnumByID {
```
coryan, line 77: Ditto, this is visiting all local enums twice.

generator/internal/api/visitors.go
```diff
@@ -0,0 +1,316 @@
...
+			// Every FieldPath starts pointing to the root input type Message
+			ref := &MessageElement{
+				Message: p.Method.InputType,
+			}
+			for _, component := range segment.FieldPath.Components {
+				if ref.Message == nil {
+					return fmt.Errorf("could not initialize FieldPath (%s), as the component (%s) does not map to any field in the enclosing type: %v", segment.FieldPath.String(), component.Identifier, ref)
+				}
+				enclosingType := ref.Message
+				ref = enclosingType.Elements[component.Identifier]
```
coryan, line 307: Implicitly, this assumes that `Elements` was populated, for all relevant types.  Which I am not sure is true. Is it  possible that a path info references a message that has not been visited yet?  Maybe it is not possible, but if so (1) you need to explain that somewhere, (2) it is too clever and we should make it obvious why the code works.

It seems that two phases for your indexing, one to populate all `Elements` and a second to populate all the `PathInfo` elements would be easier to understand.

generator/internal/language/rust.go
coryan: A separate PR SGTM.

generator/internal/language/rusttemplate.go
```diff
@@ -149,6 +148,23 @@ type RustEnumValue struct {
 	EnumType string
 }
 
+type RustCodecVisitor struct {
```
coryan, line 151: Ah, here is your second visitor... We are getting closer to having the abstraction pay off [^1].


[^1]: my rule of thumb (stolen from a book I think) is that you need 3 non-trivial use-cases to break even, and 4 to actually gain something.

generator/internal/parser/openapi.go
```diff
@@ -242,10 +243,10 @@ func makePathTemplate(template string) []api.PathSegment {
 			}
 			if strings.HasPrefix(element, "{") && strings.HasSuffix(element, "}") {
 				element = element[1 : len(element)-1]
-				segments = append(segments, api.PathSegment{FieldPath: &element})
+				segments = append(segments, api.PathSegment{FieldPath: &api.FieldPath{Components: []*api.FieldPathComponent{{Identifier: element}}}})
```
coryan, line 246: I think go experts would break this down in more lines for readability. Add some trailing commas and `go fmt` will do the expected thing.

```suggestion
				segments = append(segments, api.PathSegment{
				     FieldPath: &api.FieldPath{
				         Components: []*api.FieldPathComponent{
				             {Identifier: element},
				         },
				      },
				  })
```

generator/internal/parser/openapi.go
```diff
@@ -125,6 +125,7 @@ func makeAPIForOpenAPI(serviceConfig *serviceconfig.Service, model *libopenapi.D
 		return nil, err
 	}
 	updateMethodPagination(result)
+
```
coryan, line 128: nit: do we need this change?

generator/internal/parser/protobuf.go
```diff
@@ -46,7 +46,7 @@ func ParseProtobuf(source, serviceConfigFile string, options map[string]string)
 		}
 		serviceConfig = cfg
 	}
-	return makeAPIForProtobuf(serviceConfig, request), nil
+	return makeAPIForProtobuf(serviceConfig, request)
```
coryan, line 49: Seems like you can revert all the changes in this file?

src/gax/src/path_parameter.rs
coryan: Rust style is to avoid "stringly-typed" errors.  If we return the (existing) `Error`  enum the callers can detect the error type:

```rust
  let response = client.make_request( /* blah blah */).send().await;
  if let Some(e) = response.err() {
    if let Some(missing) = e.as_inner::<gax::path_parameter::Error>() {
    }
  }
```

src/gax/src/path_parameter.rs
```diff
@@ -35,10 +36,15 @@ where
...
-pub enum Error {
-    #[error("missing required parameter {0}")]
-    MissingRequiredParameter(String),
+pub mod errors {
+    use crate::error::Error;
+
+    pub fn missing(name: &str) -> Error {
```
coryan, line 42: I think we explicitly exclude the `path_parameter` module from the documentation, but I would still write a Rustdoc comment.

src/gax/src/path_parameter.rs
coryan, line 1: In general, I try to avoid mixing changes to `gax` and the generator in the same PR.  I think we need to build the muscle of "How do I work changes through when the generator is no longer part of the same repository".

coryan: I am not going to block this PR on that.

generator/internal/language/templates/rust/crate/src/transport.rs.mustache
```diff
@@ -56,10 +56,10 @@ impl crate::traits::{{NameToPascal}} for {{NameToPascal}} {
     ) -> Result<{{OutputTypeName}}> {
         let options = options.set_default_idempotency(reqwest::Method::{{HTTPMethod}}.is_idempotent());
         let builder = self.inner.builder(
-            reqwest::Method::{{HTTPMethod}}, format!("{{HTTPPathFmt}}"
-               {{#HTTPPathArgs}}
-               , {{{.}}}
-               {{/HTTPPathArgs}}
+            reqwest::Method::{{HTTPMethod}},
+            format!("{{#HTTPPathInfo.PathTemplate}}{{#Literal}}/{{Value}}{{/Literal}}{{#FieldPath}}/{}{{/FieldPath}}{{#Verb}}:{{Value}}{{/Verb}}{{/HTTPPathInfo.PathTemplate}}"{{#HTTPPathInfo.PathTemplate}}{{#FieldPath}}
```
coryan, line 60: I have to say, I am not sure if having the code here is the template is any better than having it in the codec... At least  I could write unit tests for the codec.

I like the rest of the changes, but this is hard to read.  Let's go with this and maybe experiment with moving all of this to the Codec on a separate PR?

coryan (changes requested):
The path info cross-referencing is assuming an earlier phase cross referencing elements.  That needs fixing or an explanation.

I think the visitor abstraction is not carrying its own weight, but maybe you can convince me of upcoming changes that will need it?

The final structure of the mustache template is scary, we can review the alternatives in a future PR.

Lots of nits here and there.

generator/internal/language/templates/rust/crate/src/transport.rs.mustache
dangazineu: I think this could be reimplemented as a Codec in the HTTPPathInfo. I was trying to keep the `HTTPPathFmt` and `HTTPPathArgs` logic side by side, as they depend on each other, but I agree that building the rust string template in mustache reduces readability.

generator/internal/language/rusttemplate.go
dangazineu: I think we could use this pattern for all language templates going forward. It reduces duplication, as you don't have to navigate to the elements you want to decorate, just override the method for the element you want. 
I didn't want to go all-in in this PR, but I think we could remove some code from the `rusttemplate.go` and `gotemplate.go` using this.

generator/internal/api/visitors.go
dangazineu: it loops through, but skips the local elements as they have `Message.API = nil` (guaranteed by the previous loop).

generator/internal/api/visitors.go
dangazineu: That's guaranteed by the traversal order of `API.Accept(v Visitor)`. It visits local and mixin messages and enums before the methods. I believe this is a generic approach, not only to this specific visitor implementation, but in general. One issue I caught as I responded to this comment is that it is currently visiting messages before enums, and it is probably better to reverse this, as messages reference enums, but not the other way around. 

Services are visited last, so all elements they reference have already been visited by the time they are (and FieldPaths are contained within the Service hierarchy). 

This approach does not solve for cross-referencing within the same type: i.e: a message referencing another message. But we don't have that problem now, so I didn't want to overcomplicate the traversal logic.

generator/internal/api/model.go
```diff
@@ -41,7 +46,27 @@ const (
...
+//   - Nested types are Enum and Message types defined inline within an enclosing Message.
+//     Nested types have their respective Enum.Parent and Message.Parent fields set to the enclosing Message.
+//   - Local types are Enum and Message types defined within the API being evaluated.
+//     Local types have their respective Enum.API and Message.API fields set to the API instance.
+//   - Mixin types are Enum and Message types defined outside the API, but referenced by it.
+//     Mixin types have their respective Enum.API and Message.API fields set to nil.
+//
+// The structure of the API represents the hierarchy of its types as follows:
+//
+//   - API.Messages contains only local, named messages. These are reusable message types defined in the API.
```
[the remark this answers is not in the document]
dangazineu: I think I had a slight misunderstanding on what it meant for a message to be declared inline within a proto. I thought it was a "complex field", but I realized that they are actually named and reusable. I've updated the documentation to reflect that.

generator/internal/api/visitors.go
dangazineu: We could (should?) move this logic to the open api and protobuf parsers. It just felt like a harmless shortcut to have it implemented in the visitor first, since it is idempotent and required only by visitors. If we ever need to traverse mixins *before* local types, we will need to change this, as the only way to distinguish them today is to traverse the local types first.

generator/internal/api/visitors.go
dangazineu: I left that comment somewhere else, but just adding here as well for context. I think visitors are a much easier way to annotate API elements with language-specific logic, possibly replacing the *template.go files in the future. As we add more languages, the savings would become more evident.

generator/internal/language/rusttemplate.go
dangazineu: I'll resolve this to keep the conversation in a single place

src/gax/src/path_parameter.rs
dangazineu: I am considering reverting that change back to the original proposal, with the error message directly in the template as a first iteration of this PR. The additional changes in gax will require some rust knowledge that I don't have yet, and I'd rather take my time on that without blocking this PR.

generator/internal/language/templates/rust/crate/src/transport.rs.mustache
```diff
@@ -56,11 +56,11 @@ impl crate::traits::{{NameToPascal}} for {{NameToPascal}} {
     ) -> Result<{{OutputTypeName}}> {
         let options = options.set_default_idempotency(reqwest::Method::{{HTTPMethod}}.is_idempotent());
         let builder = self.inner.builder(
-            reqwest::Method::{{HTTPMethod}}, format!("{{HTTPPathFmt}}"
-               {{#HTTPPathArgs}}
-               , {{{.}}}
-               {{/HTTPPathArgs}}
-            ))
+            reqwest::Method::{{HTTPMethod}},{{#HTTPPathInfo}}
```
dangazineu, line 59: We can simplify the templates even further, if we loop through `api.*` elements, instead of the language-specific ones. The fields used in this level, for example, could be attached directly under the `api.Method.Codec` instead of the `api.HTTPPathInfo`.

generator/internal/language/templates/rust/crate/src/transport.rs.mustache
dangazineu: I've updated this code to take the fully formed format string from the codec, and added helper methods to the codec so the template doesn't have to iterate through irrelevant segments, it got a bit cleaner.

generator/internal/api/visitors.go
coryan: > I think visitors are a much easier way to annotate API elements with language-specific logic,

Whether this is "much easier" or not is what we are discussing, no?  Abstractions have costs.  This one allows you to write "call this function on every element of type X" and skip the iteration[^1].  Essentially saving a number of for loops at the cost of "I am not sure how these functions get called".

> possibly replacing the *template.go files in the future.

I can see how that would work.  We would write a visitor (maybe more than one) for each language, run it over the model and it would annotate each message, method, field, etc. with the language specific elements.

> As we add more languages, the savings would become more evident.

Admittedly, it would help us structure the code. An implementor of a new language would write the `Visit*()` functions and be done.  I would not need to learn about how to traverse the `API` data structure.

We would still need the same language specific annotation types, we would still need to run the same computations on each element type.

Overall, I remain skeptical about this particular visitor helping us in the long term.  It may be a good idea, but it is too early to tell.  Certainly too early to introduce it without more design.

My vote is to rip it out.  We can do it on this PR, or I volunteer to do it in a second pass.  FWIW, I think the cross-referencing is useful (and better than what we had before), and the code for handling HTTPInfo is also better.

We can revisit (pun intended) this decision once we have completed the cleanups to remove the templatedata.go duplication, and we have a second language implemented.

[^1]: side note: this is hardly an implementation of the [Visitor pattern](https://en.wikipedia.org/wiki/Visitor_pattern).  Visitors are about double dispatch, about adding dynamic polymorphic behavior to classes without having to add functions to the whole class hierarchy, and about avoiding unsafe downcasting.  Here the types are known.  The only thing the visitors do is consolidate the loops.

coryan (changes requested):

generator/internal/api/visitors.go
dangazineu: The side effect itself isn't done by the visitor, it is done by the `Accept` methods in the model. Regardless of which visitor you provide, the first time you invoke `API.Accept`, all parent hierarchy will be set. 

The correct implementation here is for the `makeAPIFor*` functions to build the hierarchy correctly, including all the pointers (minus the cross-references). I can fix that as part of this PR to make sure the visitor implementation doesn't have any unintended side effect.

generator/internal/api/visitors.go
dangazineu: Sounds good. I think we could achieve the same result without a visitor with not much change involved, I'll take a look in the coming days.

dangazineu:
I'll close this for now, as most of the history in this PR is about the visitor implementation, which won't make into the final solution. I also won't be able to work on this for a few days, so when I get back to this I'll just start from the latest at head.

Issue #37, opened 2024-10-30
handle well-known proto types


codyoss:
This should maybe be multiple bugs. We need to figure out how we handle reading common proto types. Some examples are, but not limited to:

- [x] google.protobuf.Timestamp
- [x] google.protobuf.Duration
- [x] google.protobuf.Any
- [x] google.protobuf.Empty
- [x] google.protobuf.FieldMask
- [x] google.protobuf.DoubleValue
- [x] google.protobuf.FloatValue
- [x] google.protobuf.Int64Value
- [x] google.protobuf.UInt64Value
- [x] google.protobuf.Int32Value
- [x] google.protobuf.UInt32Value
- [x] google.protobuf.BoolValue
- [x] google.protobuf.StringValue
- [x] google.protobuf.BytesValue
- [x] google.protobuf.Struct
- [x] google.protobuf.ListValue
- [x] google.protobuf.Value
- [x] google.protobuf.NullValue
- [x] google.protobuf.Api
- [x] google.protobuf.Type
- [x] google.protobuf.Enum
- [x] google.longrunning.Operation
- [x] google.rpc.Status
- [x] google.protobuf.DescriptorProto - used by BigQuery storage (for writes), which requires bidi streams.

coryan:
Moving out of the milestone, the remaining types will be needed with other services, but not secret manager.

coryan:
There are only two remaining, none of them will be needed for HTTP + JSON based client libraries.

PR #754, opened 2025-01-20
feat(wkt): implement `Struct` and friends

coryan:
Part of the work for #37

### DIFF
--- a/src/wkt/src/lib.rs
+++ b/src/wkt/src/lib.rs
@@ -33,6 +33,8 @@ mod field_mask;
 pub use crate::field_mask::*;
 mod timestamp;
 pub use crate::timestamp::*;
+mod rstruct;
+pub use crate::rstruct::*;
 mod wrappers;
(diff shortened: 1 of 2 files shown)


codyoss (approved):
The other type we may eventually need here is:https://protobuf.dev/reference/php/api-docs/Google/Protobuf/NullValue.html. This is at least generated in Go's structpb package: https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#NullValue

coryan:
> The other type we may eventually need here is:https://protobuf.dev/reference/php/api-docs/Google/Protobuf/NullValue.html. This is at least generated in Go's structpb package: https://pkg.go.dev/google.golang.org/protobuf/types/known/structpb#NullValue

I left it out because it is not referenced anywhere in googleapis.  If we ever need it, we can add it, though exactly how it would be represented in Rust escapes me at the moment.

PR #756, opened 2025-01-20
feat(generator/rust): avoid name clashes in `model`

### LINKED ISSUE #755: Avoid name clashes with Rust's prelude
Rust automatically `use`'s a number of symbols from the `std` library, including `Option`, and `String`.  This can clash with service, message, and enum names generated from protobuf files.

We have been ignoring this problem because the clashes are uncommon. This is no longer tenable:

To implement `google.protobuf.Api`, `google.protobuf.Type` and other names used by `google/api` (see #748) we will need to introduce `Option`:

https://github.com/protocolbuffers/protobuf/blob/e3ac4c9342e5393e3fa974a420fc1bbc68e7397a/src/google/protobuf/type.proto#L172

That will clash with `std::option::Option`. It is time to disable the prelude in `model.rs` and fully qualify all the types we use from `std::*`.

coryan:
Rust automatically imports a number of names, including `Option` and
`String`. Sometimes these names clash with the names of generated types.
To avoid this problem we can disable the automatic imports, and generate
fully qualified names where we use any imported names.

Fixes #755


codyoss (approved):
We can do this. But I wonder if we should just obfuscate in these cases. Any downstream user that * imports these modules would run into the same issue -- or even just tries to import the exact type without a qualifier or rename. We can't protect against things that may be added in future rust editions predules, but these feel common enough in Rust that it may be good not to use the same name.

PR #774, opened 2025-01-21
feat(generator/rust): handle lowercase enum values

### LINKED ISSUE #772: Some protobuf enum values are not `SNAKE_CASE`
We use Rust constants for enum values. These must be styled as `SNAKE_CASE` .  Protobuf typically uses the same style, but sometimes not:

https://github.com/googleapis/googleapis/blob/318818b22ec2bd44ebe43fe662418b7dff032abf/google/cloud/sql/v1/cloud_sql_resources.proto#L1731

We need to use the Rust style for the string *names*, but preserve the lowercase string *values*.

coryan:
Some enum values do not follow the Protobuf naming conventions. We need
to normalize them or Rust complains.

Fixes #772


codyoss (approved):
This seems fine, but it makes we wonder if we should just do this during parsing -- have a sort of normalization at that point so we don't have to handle this in every language.

coryan:
> This seems fine, but it makes we wonder if we should just do this during parsing -- have a sort of normalization at that point so we don't have to handle this in every language.

I am not sure... other languages may have different conventions, and the mapping from `whatever` -> `SCREAMING_SNAKE_CASE` -> `$lang style` may be lossy.  Maybe experience will show that you are right, but I would like to keep my options open for now.

PR #789, opened 2025-01-22
fix(auth): sync claim/token times in SA creds

### LINKED ISSUE #733: Mismatched expiration time in service account credentials
This is wrong:

https://github.com/googleapis/google-cloud-rust/blob/d1348119749a52e9bdc48267eca3d1746d330250/src/auth/src/credentials/service_account_credential.rs#L89

The expiration time should match the expiration time in the credentials, which is set to `now - fudge + default`.  And there are two separate calls to `utc_now()` which may return different values.

dbolduc:
Fixes #733 

Make sure the token's `expires_at` time matches the `exp` field in the claim.

We drop the `JwsClaimsBuilder` and do any defaulting in `service_account_credential.rs`. Note that the `exp` and `iat` fields are required, so they are probably better represented as `OffsetDateTime` than `Option<OffsetDateTime>`.

### DIFF
--- a/src/auth/src/credentials/service_account_credential.rs
+++ b/src/auth/src/credentials/service_account_credential.rs
@@ -13,7 +13,9 @@
 // limitations under the License.
 
 use crate::credentials::dynamic::CredentialTrait;
-use crate::credentials::util::jws::{JwsClaimsBuilder, JwsHeader, DEFAULT_TOKEN_TIMEOUT};
+use crate::credentials::util::jws::{
+    JwsClaims, JwsHeader, CLOCK_SKEW_FUDGE, DEFAULT_TOKEN_TIMEOUT,
+};
 use crate::credentials::Result;
 use crate::errors::CredentialError;
 use crate::token::{Token, TokenProvider};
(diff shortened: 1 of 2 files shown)


src/auth/src/credentials/util/jws.rs
```diff
@@ -14,7 +14,6 @@
 
 use crate::credentials::CredentialError;
 use crate::credentials::Result;
-use derive_builder::Builder;
```
coryan, line 17: Thanks. I think there is an unused dependency in `src/auth/Cargo.toml` now.

src/auth/src/credentials/service_account_credential.rs
```diff
@@ -66,11 +68,17 @@ impl TokenProvider for ServiceAccountTokenProvider {
...
+        let exp = now + DEFAULT_TOKEN_TIMEOUT;
+        let claims = JwsClaims {
+            iss: self.service_account_info.client_email.clone(),
+            scope: Some(DEFAULT_SCOPES.map(|s| s.to_string()).to_vec()),
+            aud: None,
+            exp,
+            iat: now,
+            typ: None,
+            sub: None,
+        };
```
coryan, line 81: Alternatively, you could create a `JwsClaims::new(now: OffsetDateTime) -> Self ` function.

src/auth/src/credentials/util/jws.rs
```diff
@@ -29,32 +28,28 @@ pub(crate) const CLOCK_SKEW_FUDGE: Duration = Duration::from_secs(10);
 pub(crate) const DEFAULT_TOKEN_TIMEOUT: Duration = Duration::from_secs(3600);
 
 /// JSON Web Signature for a token.
-#[derive(Clone, Serialize, Default, Builder)]
-#[builder(setter(into, strip_option), default)]
-pub struct JwsClaims<'a> {
-    pub iss: &'a str,
+#[derive(Serialize)]
+pub struct JwsClaims {
+    pub iss: String,
```
coryan, line 33: I like the simplicity of using `String`.  It is more expensive, but this function will be called like once per hour so we should not complicate things too much in search of small performance improvements.

src/auth/src/credentials/util/jws.rs
```diff
@@ -116,44 +112,30 @@ mod tests {
         )
         .unwrap();
 
-        // 5 seconds is like 640KiB, good enough for everybody
```
coryan, line 119: 🎉

coryan (commented):
Please remove any unnecessary dependencies and then ship it!

src/auth/src/credentials/util/jws.rs
dbolduc: Ack

src/auth/src/credentials/util/jws.rs
dbolduc: Still used elsewhere:

https://github.com/googleapis/google-cloud-rust/blob/31c8a5eb590cf8908453ddaa14fc35cfb32220fb/src/auth/src/credentials/service_account_credential.rs#L40

sai-sunder-s (approved):

PR #798, opened 2025-01-23
cleanup(gax): remove unused code

coryan:
Most of the code we used to require path parameters is gone. With that
code removed, the `request_parameter` module served no purpose, so I
removed that too.


codyoss (approved):
yea deleting code!

Issue #381, opened 2024-12-09
Handle comments that look like HTML tags but aren't


coryan:
Consider this:

https://github.com/googleapis/google-cloud-rust/blob/2c05b4823fbcf6364820907e2709a1502a39f89d/src/gax/src/error/rpc/generated/mod.rs#L199-L200

Rustdoc treats `<ip ...>` and `<Google ...>` as HTML tags.  This creates two problems:

1. rustdoc warngs complains that they are not closed, and
    ```
    warning: unclosed HTML tag `ip`
    --> src/gax/src/error/rpc/generated/mod.rs:198:9
        |
    198 | /         /// The subject on which the quota check failed.
    199 | |         /// For example, "clientip:<ip address of client>" or "project:<Google
    200 | |         /// developer project id>".
        | |___________________________________^
    ```
1. rustdoc removes them from the output:
    ![image](https://github.com/user-attachments/assets/bb325ca1-bcc4-40f3-bfbc-40feadce4831)

Either we change all the cases of placeholders using `<... >`, or we escape these placeholders in the Rust code to keep these in the documentation output.

To my knowledge, the only relevant API ( [aip/192](https://google.aip.dev/192) says nothing about such "almost looks like a tag" problems.

coryan:
Maybe all we need to do is disable this warning for generated code:

https://doc.rust-lang.org/rustdoc/lints.html#invalid_html_tags

coryan:
FWIW, in #409 I disabled the warnings about these things.  The generated documentation skips the thing that looks like a tag, so the problems from the description remains.

PR #801, opened 2025-01-23
fix: escape html tags for placeholders in comments

mpeddada1:
Fixes https://github.com/googleapis/google-cloud-rust/issues/381

This PR adds escaping for placeholders in comments. 

Running `cargo doc --open` results in the placeholders being retained as seen below:
Screenshot 2025-01-22 at 10 33 23 PM

Additional notes: tried making the changes consistent by escaping *all* urls with `&lt;` and `&gt;` but this resulted in several warnings when `cargo doc` was called.


src/generated/api/src/model.rs
```diff
@@ -7157,7 +7157,7 @@ impl wkt::message::Message for SystemParameter {
 #[non_exhaustive]
 pub struct Usage {
     /// Requirements that must be satisfied before a consumer project can use the
-    /// service. Each requirement is of the form <service.name>/<requirement-id>;
+    /// service. Each requirement is of the form <service.name>/&lt;requirement-id&gt;;
```
mpeddada1, line 7160: Hm interestingly, even though <service.name> isn't escaped, it still renders as:

Screenshot 2025-01-22 at 10 52 36 PM

generator/internal/language/rust.go
```diff
@@ -1035,6 +1038,33 @@ func processParagraph(node ast.Node, links map[string]bool, documentationBytes [
 	return results
 }
 
+func fetchPlaceholders(node ast.Node, documentationBytes []byte) map[string]string {
```
coryan, line 1041: placeholders is not a great name.  Maybe `fetchHtmlTags` or `fetchHtmlSubstitutions`?

generator/internal/language/rust.go
```diff
@@ -1035,6 +1038,33 @@ func processParagraph(node ast.Node, links map[string]bool, documentationBytes [
...
+				if !commentUrlRegex.MatchString(original) && !strings.HasPrefix(original, "<a href=") && !strings.HasPrefix(original, "</a>") {
+					htmlMapping[original] = html.EscapeString(original)
+				}
+			}
+		}
+	}
+	return htmlMapping
+}
+
+func escapeHTMLTags(htmlTags map[string]string, line string) string {
```
coryan, line 1060:
```suggestion
func escapeHTMLTags(htmlSubstitutions map[string]string, line string) string {
```

coryan (commented):
After seeing the effect of these changes, I wonder if my original bug report was a good idea.  Maybe we need more fixes upstream, or maybe we need to be far more surgical in what HTML-like things are replaced.

src/generated/spanner/admin/instance/v1/src/model.rs
```diff
@@ -1608,9 +1608,9 @@ pub struct CreateInstanceConfigRequest {
 
     /// Required. The `InstanceConfig` proto of the configuration to create.
     /// `instance_config.name` must be
-    /// `<parent>/instanceConfigs/<instance_config_id>`.
+    /// `&lt;parent&gt;/instanceConfigs/<instance_config_id>`.
```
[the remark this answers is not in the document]
mpeddada1: Hm that's true, we might need to additional logic to ensure that text within `` doesn't get processed like the others.

generator/internal/language/rust.go
```diff
@@ -898,8 +898,8 @@ func extractProtoLinks(line string, links map[string]bool) {
 	}
 }
 
-// Encloses standalone URLs with angled brackets.
-func escapeUrls(line string) string {
+// Encloses standalone URLs with angled brackets and escape placeholders.
+func escapeUrlsAndPlaceholders(placeholders map[string]string, line string) string {
```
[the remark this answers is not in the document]
mpeddada1: That's much better, thank you.

generator/internal/language/rust.go
mpeddada1: Done, good point.

generator/internal/language/rust.go
mpeddada1: Done, renamed to `htmlSubstitutions`.

src/generated/api/src/model.rs
```diff
@@ -3103,8 +3103,8 @@ pub struct Page {
     #[serde(skip_serializing_if = "std::string::String::is_empty")]
     pub name: std::string::String,
 
-    /// The Markdown content of the page. You can use <code>&#40;== include {path}
-    /// ==&#41;</code> to include content from a Markdown file. The content can be
+    /// The Markdown content of the page. You can use &lt;code&gt;&#40;== include {path}
```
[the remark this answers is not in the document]
coryan: I think your PR is doing the right thing here FWIW.

src/generated/spanner/admin/instance/v1/src/model.rs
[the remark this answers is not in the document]
coryan: Still not quite right here?

generator/testdata/rust/protobuf/golden/module/rpc/mod.rs
```diff
@@ -235,8 +235,8 @@ pub mod quota_failure {
     pub struct Violation {
 
         /// The subject on which the quota check failed.
-        /// For example, "clientip:<ip address of client>" or "project:<Google
-        /// developer project id>".
+        /// For example, "clientip:&lt;ip address of client&gt;" or "project:&lt;Google
```
coryan, line 238: Can you use `\<` instead of `&lt;` ?  No big deal if you cannot, I just find the backslash escaping more readable, but both render the same I assume.

src/generated/spanner/admin/instance/v1/src/model.rs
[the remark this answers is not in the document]
mpeddada1: Yes, not quite there yet. Looking into making the search more robust. I realized that collecting the tags and doing a replace all doesn't work in this case especially.

src/generated/api/src/model.rs
```diff
@@ -3098,8 +3098,8 @@ pub struct Page {
     #[serde(skip_serializing_if = "std::string::String::is_empty")]
     pub name: std::string::String,
 
-    /// The Markdown content of the page. You can use <code>&#40;== include {path}
-    /// ==&#41;</code> to include content from a Markdown file. The content can be
+    /// The Markdown content of the page. You can use \<code\>&#40;== include {path}
```
coryan, line 3101: This change is good, but we should also fix this upstream: it should be using a code span (in markdown), instead of `<code> ... </code>`.

generator/internal/language/rust.go
```diff
@@ -934,7 +933,50 @@ func extractProtoLinks(line string, links map[string]bool) {
...
+
+					}
+				}
+			}
+		}
+	}
+	return escapedString
+}
+
+func isWithinCodeSpan(node ast.Node) bool {
```
coryan, line 970: nice.

generator/internal/language/rust.go
```diff
@@ -934,7 +933,50 @@ func extractProtoLinks(line string, links map[string]bool) {
...
+							end := int(segment.Stop) - line.Start
+							escapedHTML := strings.Replace(segmentContent, "<", "\\<", 1)
+							escapedHTML = strings.Replace(escapedHTML, ">", "\\>", 1)
+							escapedString = strings.ReplaceAll(escapedString, string(lineContent[start:end]), escapedHTML)
+						}
+
+					}
+				}
+			}
+		}
```
coryan, line 965: Consider:

```suggestion
		if child.Kind() != ast.KindRawHTML {
		  continue
		}
		rawHTML := child.(*ast.RawHTML)
		if isWithinCodeSpan(node) {
		    continue
		}
		for i := 0; i < rawHTML.Segments.Len(); i++ {
			segment := rawHTML.Segments.At(i)
			segmentContent := string(segment.Value(documentationBytes))
			if segment.Start < line.Start || (segment.Start >= line.Stop) {
			    continue
			}
			if strings.HasPrefix(segmentContent, "<br />") || strings.HasPrefix(segmentContent, "<a href=") || strings.HasSuffix(segmentContent, "</a>") {
			    continue
			}
			start := int(segment.Start) - line.Start
			end := int(segment.Stop) - line.Start
			escapedHTML := strings.Replace(segmentContent, "<", "\\<", 1)
			escapedHTML = strings.Replace(escapedHTML, ">", "\\>", 1)
			escapedString = strings.ReplaceAll(escapedString, string(lineContent[start:end]), escapedHTML)
		}
```

coryan (approved):
Please merge this change (assuming you are happy with it, I am).

Then try removing this:

https://github.com/googleapis/google-cloud-rust/blob/25149ad9f5cfdef32822a29710a519a49e37807b/generator/internal/language/templates/rust/common/model.rs.mustache#L21C1-L22C38

Regenerate all the code and see if more warnings appear.  If so, document them so we can fix them in a new pass.  Otherwise, just leave the warnings turned on.

generator/internal/language/rust.go
mpeddada1: Done, thanks!

src/generated/api/src/model.rs
mpeddada1: Makes sense, will up a change for this.

PR #802, opened 2025-01-23
feat(generator): support self-referential fields

### LINKED ISSUE #586: Support message types that are self referential
Rust does not like self-referential types, at least without indirection. We will need to Box types that are self referential as such. Example

```proto
message Node {
  Node next = 1;
  Node prev = 2;
  string val = 3;
}
```

```rust
pub struct Node {
  pub next: Option<Box<Node>>,
  pub prev: Option<Box<Node>>,
  pub val: String,
}
```

coryan:
Message fields may refer to the same message, directly or indirectly.
With this change the fields gain a `Recursive` attribute, set to `true`
if the field directly (or indirectly) references the containing message.

The Rust Codec uses this attribute to emit `Option<Box<T>>` instead of
`Option<T>` for such fields.

Fixes #586

### DIFF
--- a/generator/internal/language/rust.go
+++ b/generator/internal/language/rust.go
@@ -506,10 +506,43 @@ func rustFieldType(f *api.Field, state *api.APIState, primitive bool, modulePath
 	if !primitive && f.Repeated {
 		return fmt.Sprintf("std::vec::Vec<%s>", rustBaseFieldType(f, state, modulePath, sourceSpecificationPackageName, packageMapping))
 	}
+	if !primitive && f.Recursive {
+		base := rustBaseFieldType(f, state, modulePath, sourceSpecificationPackageName, packageMapping)
+		if f.Optional {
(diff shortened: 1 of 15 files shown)


generator/internal/language/rust.go
```diff
@@ -506,12 +506,45 @@ func rustFieldType(f *api.Field, state *api.APIState, primitive bool, modulePath
 	if !primitive && f.Repeated {
 		return fmt.Sprintf("std::vec::Vec<%s>", rustBaseFieldType(f, state, modulePath, sourceSpecificationPackageName, packageMapping))
 	}
+	if !primitive && f.Recursive {
```
codyoss, line 509: nit: all of the top level if checks check `!primitive`. Maybe this should be factored out.

codyoss (approved):

generator/internal/language/rust.go
coryan: Done.

PR #811, opened 2025-01-24
impl(sample): consolidate secretmanager testdata

julieqiu:
Consolidate test data used across multiple tests that rely on Secret Manager. The testdata is centralized within the sample package to enable reuse, reduce duplication, and more close reflect the Secret Manager data.


generator/internal/sample/serviceconfig.go
```diff
@@ -0,0 +1,84 @@
+// Copyright 2024 Google LLC
```
coryan, line 1: nit: if this is a new file then it should be 2025.  My guess is you moved it but GitHub gets confused when you move and change a file "too much".

generator/internal/sample/api.go
```diff
@@ -0,0 +1,341 @@
+// Copyright 2024 Google LLC
```
coryan, line 1: nit: 2025

generator/internal/parser/openapi_test.go
```diff
@@ -606,7 +593,7 @@ func TestOpenAPI_MakeAPI(t *testing.T) {
 				Name:          "locations",
 				JSONName:      "locations",
 				Documentation: "A list of locations that matches the specified filter in the request.",
-				Typez:         api.MESSAGE_TYPE,
+				Typez:         11,
```
coryan, line 596: Why replace this with the literal instead of the constant?

coryan (approved):
A few minor changes.  Please address and then merge.

generator/internal/parser/openapi_test.go
julieqiu: This is a typo. Replaced throughout the file.

generator/internal/sample/serviceconfig.go
julieqiu: Done.

PR #818, opened 2025-01-24
feat(functions/v1): generate library

### LINKED ISSUE #817: Generate library for `google/cloud/functions/v1`

coryan:
Fixes #817


codyoss (approved):
Approved. Looking at the APIs between this and v1 I don't really see anything you can't do in v2 that you can do in v1. It makes we wonder if we should generate this one -- at least to start.

coryan:
Let's not generate it then. Closing.

PR #836, opened 2025-01-24
feat(bigtableadmin/v2): generate library

### LINKED ISSUE #712: Generate client library for `google/bigtable/admin`
This is blocked by

- [x] #666
- [x] #781  
- [x] #586 

The generated code fails to compile:

```
cargo new --lib --vcs none src/generated/bigtableadmin/v2
taplo fmt Cargo.tom
go -C generator/ run ./cmd/sidekick generate -project-root .. -specification-source google/bigtable/admin/v2 -service-config google/bigtable/admin/v2/bigtableadmin_v2.yaml -output src/generated/bigtableadmin/v2
cargo fmt
```

Then `cargo build` gives you:

```
error[E0428]: the name `GetIamPolicy` is defined multiple times
    --> src/generated/bigtableadmin/v2/src/builders.rs:2803:1
     |
1032 | pub struct GetIamPolicy(BigtableInstanceAdminRequestBuilder<iam_v1::model::GetIamPolicyRequest>);
     | ------------------------------------------------------------------------------------------------- previous definition of the type `GetIamPolicy` here
...
2803 | pub struct GetIamPolicy(BigtableTableAdminRequestBuilder<iam_v1::model::GetIamPolicyRequest>);
     | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `GetIamPolicy` redefined here
     |
     = note: `GetIamPolicy` must be defined only once in the type namespace of this module
```

and

```
error[E0391]: cycle detected when computing when `model::r#type::Aggregate` needs drop
    --> src/generated/bigtableadmin/v2/src/model.rs:6793:5
     |
6793 |     pub struct Aggregate {
     |     ^^^^^^^^^^^^^^^^^^^^
     |
     = note: ...which immediately requires computing when `model::r#type::Aggregate` needs drop again
note: cycle used when computing when `model::ColumnFamily` needs drop
    --> src/generated/bigtableadmin/v2/src/model.rs:5502:1
     |
5502 | pub struct ColumnFamily {
     | ^^^^^^^^^^^^^^^^^^^^^^^
     = note: see https://rustc-dev-guide.rust-lang.org/overview.html#queries and https://rustc-dev-guide.rust-lang.org/query.html for more information
```

coryan:
Fixes #712


Cargo.toml
```diff
@@ -20,6 +20,7 @@ members = [
   "src/base",
   "src/gax",
   "src/generated/api",
+  "src/generated/bigtableadmin/v2",
```
dbolduc, line 23: this is surprising to me. I expected `src/generated/bigtable/admin/v2`

dbolduc: Is it based on the endpoint?

If it is based on the `*.yaml`... I think there are services with multiple yamls of the same name in different directories.

dbolduc: > services with multiple yamls of the same name in different directories

It is `monitoring`, which I think is next on your list anyway.

```
google/monitoring/dashboard/v1/monitoring.yaml
google/monitoring/metricsscope/v1/monitoring.yaml
google/monitoring/v3/monitoring.yaml
```

coryan: The directory can change without much problem.  Customers never see it.  I am agnostic about its name so I just changed it.

The interesting part is the package name.  It is  `${prefix}-bigtable-admin-v2`, which is based on the protobuf package name.

.typos.toml
```diff
@@ -30,6 +30,9 @@ ser = "ser"
 # from google/api:
 advices    = "advices"    # In the code, too late to fix.
 readablity = "readablity"
+# src/generated/bigtableadmin/v2, this may be a typo or not, but in any case,
```
codyoss, line 33: nit: this path is now out of dat

codyoss (approved):

.typos.toml
coryan: Fixed.

Cargo.toml
dbolduc: > The directory can change without much problem. Customers never see it.

Ah, ok.

PR #852, opened 2025-01-26
feat(generator/rust): implicit LROs mixin

### LINKED ISSUE #781: Missing LRO mixin in some APIs
I am not sure if this is a bug in cloud build, or `sidekick` needs to support services that have LROs but do not have the LRO mixin.

- [x] `google/devtools/cloudbuild/v1`
- [x] `google/bigtable/admin/v2`

----

As noted in #780 the code emitted for `google/devtools/cloudbuild/v1` does not compile.  `sidekick` is trying to generate the LRO helpers, since there are functions that return `google.longrunning.Operation` and have the right annotations:

```proto
  rpc CreateBuild(CreateBuildRequest) returns (google.longrunning.Operation) {
  ... ... ... snip snip ... ... ...
    option (google.longrunning.operation_info) = {
      response_type: "Build"
      metadata_type: "BuildOperationMetadata"
    };
  }
```
 https://github.com/googleapis/googleapis/blob/e9a4c38a81933108eaa6ac96c7ead31e253c8c64/google/devtools/cloudbuild/v1/cloudbuild.proto#L97-L117

However, the service config does not import the mixin:

```yaml
type: google.api.Service
config_version: 3
name: cloudbuild.googleapis.com
title: Cloud Build API


apis:
- name: google.devtools.cloudbuild.v1.CloudBuild
```

https://github.com/googleapis/googleapis/blob/e9a4c38a81933108eaa6ac96c7ead31e253c8c64/google/devtools/cloudbuild/v1/cloudbuild_v1.yaml#L1-L15

Without the mixin, there is no `get_operation()` function or builder defined. The generated code assumes these exist. One possible "solution" is to disable the LRO helpers if the mixin does not exist, but I am not sure the generated code would be that useful in that case.

Another solution might be to force the mixin when we find LRO-like functions.

coryan:
Fixes #781 (yes, second time)

If you are wondering how this can possible work: for services with `/v1` in the path, the default rules inherited from the mixin work just fine.  

I wonder if we should mixin *all* the methods.  Thoughts?

### DIFF
--- a/generator/internal/parser/mixin.go
+++ b/generator/internal/parser/mixin.go
@@ -72,6 +72,12 @@ func loadMixins(serviceConfig *serviceconfig.Service, withLongrunning bool) (mix
 		}
 	}
 	enabledMixinMethods = loadMixinMethods(serviceConfig)
+	if withLongrunning {
+		// We prefer using the `http.rules` section from the service config, but
+		// if we must implement the longrunning mixin, we must also implement
+		// the GetOperation method.
+		enabledMixinMethods[".google.longrunning.Operations.GetOperation"] = true
+	}
 	return enabledMixinMethods, files
 }
 
--- a/generator/internal/parser/mixin_test.go
+++ b/generator/internal/parser/mixin_test.go
@@ -1,0 +1,81 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package parser
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/googleapis/google-cloud-rust/generator/internal/sample"
+	"google.golang.org/genproto/googleapis/api/annotations"
+)
+
+func TestProtobuf_ForceLongrunning(t *testing.T) {
+	sc := sample.ServiceConfig()
+	sc.Http = &annotations.Http{
+		Rules: []*annotations.HttpRule{
+			{
+				Selector: "google.longrunning.Operations.CancelOperation",
+				Pattern: &annotations.HttpRule_Post{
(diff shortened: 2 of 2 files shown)


codyoss (commented):
This seems to directly go against what is outlined in the [AIP](https://google.aip.dev/client-libraries/4234). The relevant section:

```
To be generated into a client library, a mixin service must be declared under [apis] in a 
[google.api.Service]. Furthermore, only the RPCs with [google.api.http] bindings declared in 
the [http] configuration of the same [google.api.Service] can be generated (more details on 
these bindings in [AIP-127]). If a mixin service RPC does not have a [google.api.http]
rule declared in the [google.api.Service], it must not be generated.
```

---

I think the key difference between the level of support is just that these methods are on the clients. I think one way we could work around this is still generate these, but only have them as crate visible. Thoughts?

coryan:
> I think the key difference between the level of support is just that these methods are on the clients. I think one way we could work around this is still generate these, but only have them as crate visible. Thoughts?

Then applications cannot manually poll the LRO.  They could only use the helpers (or use the client in `gcp-sdk-longrunning` directly).  That seems bad.

I think this is either a bug in the AIP, or a bug in the service definition file.  I am willing to workaround such bugs.

codyoss:
> Then applications cannot manually poll the LRO. They could only use the helpers (or use the client in gcp-sdk-longrunning directly). That seems bad.

You are correct. Although, if we are going with the approach in the PR as written we would only be exposing Get -- which is what our helper type provides anyways. Said another way, we only need to expose/generate this because it is a detail of our implementation of LRO helpers. To me having useful helpers for LROs and exposing them directly on the client are two seperate things -- this AIP only touches on the later.

> I think this is either a bug in the AIP, or a bug in the service definition file. I am willing to workaround such bugs.

I would not call this a bug in the API, because it is basically saying only generate methods which are publicly documented as being a part of the service. I could see the argument for this being a usability bug of the service though.

The thing I worry about generating extra methods is not about it working today but what happens when the next iterations of mixins happens and it make changes to the system assuming all generators follow the rules. I could see a world where a change that is fine for every other generator happens to be a breaking change for us. Will this happen -- unlikely, but I think that is what these AIPs are trying to protect us from; breaking changes.

With all that said I don't want to block on this, but this is why I am hesitant.

codyoss (approved):

coryan:
> > I think this is either a bug in the AIP, or a bug in the service definition file. I am willing to workaround such bugs.
> 
> I would not call this a bug in the API, because it is basically saying only generate methods which are publicly documented
> as being a part of the service. I could see the argument for this being a usability bug of the service though.

Ack.

>  but this is why I am hesitant.

Fair enough.

Issue #679, opened 2025-01-13
Auth implement getting token for service account credentials


harkamaljot:
We want to allow the user to fetch token for service account credentials.

The work, broken down into PRs, is roughly:

- [x] Implement ServiceAccountTokenProvider for fetching token with just issuer, audience and private key support RSA256.
- [ ] Support ES256 algorithm for private key
- [ ] Support passing in additional claims when creating a self signed jwt
- [ ] Supporting passing in scopes when making a self signed jwt

coryan:
Is this done?

harkamaljot:
Yes this is done, closing this task.

PR #867, opened 2025-01-28
cleanup(auth): censor SA private fields

dbolduc:
Part of the work for #679 

censor private fields, like we do for user credentials.

### DIFF
--- a/src/auth/src/credentials/service_account_credential.rs
+++ b/src/auth/src/credentials/service_account_credential.rs
@@ -52,7 +52,7 @@ impl std::fmt::Debug for ServiceAccountInfo {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         f.debug_struct("ServiceAccountInfo")
             .field("client_email", &self.client_email)
-            .field("private_key_id", &"[censored]")
+            .field("private_key_id", &self.private_key_id)
             .field("private_key", &"[censored]")
             .field("project_id", &self.project_id)
             .field("universe_domain", &self.universe_domain)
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/service_account_credential.rs
```diff
@@ -48,6 +48,18 @@ pub(crate) struct ServiceAccountInfo {
     universe_domain: String,
 }
 
+impl std::fmt::Debug for ServiceAccountInfo {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ServiceAccountInfo")
+            .field("client_email", &self.client_email)
+            .field("private_key_id", &"[censored]")
```
coryan, line 55: I am not sure if you need to censor this.  I would hope that the ID of the key does not reveal anything about the key contents, or allow you to do anything nefarious.  And it would be useful to know what key was involved when troubleshooting something.

Maybe merge as-is and find out if it really needs censoring?

src/auth/src/credentials/service_account_credential.rs
```diff
@@ -48,6 +48,18 @@ pub(crate) struct ServiceAccountInfo {
...
+            .field("private_key", &"[censored]")
```
coryan, line 56: For a future improvement: maybe printing the first X digits or a SHA would be useful for troubleshooting.

coryan (approved):

src/auth/src/credentials/service_account_credential.rs
dbolduc: > first X digits

In this case, it will be the BEGIN PRIVATE KEY nonsense

> SHA would be useful

I could see that. #872

src/auth/src/credentials/service_account_credential.rs
dbolduc: I think you are right that the ID of the private key does not need to be censored. Fixed.

PR #873, opened 2025-01-28
cleanup(auth): visibility + error helpers

dbolduc:


src/auth/src/credentials/service_account_credential.rs
```diff
@@ -40,7 +40,7 @@ const DEFAULT_SCOPES: [&str; 1] = ["https://www.googleapis.com/auth/cloud-platfo
 #[allow(dead_code)]
 #[derive(serde::Deserialize, Debug, Builder)]
 #[builder(setter(into))]
-pub(crate) struct ServiceAccountInfo {
+struct ServiceAccountInfo {
```
coryan, line 43: These are still `pub(crate)` because of:

https://github.com/googleapis/google-cloud-rust/blob/33a8148e4c7a38a0e5132cd7518ac6eda0434acd/src/auth/src/credentials.rs#L16

Right?  What difference does it make to make them public?  Curious, maybe it goes into the TIL list.

If you really intend to make this public: the `builder` thing is at 0.20.  I do not think we should have public APIs that depend on unstable APIs (unless protected by a feature or something).

dbolduc: > If you really intend to make this public:

It should not be public.

> These are still `pub(crate)` because of:

I missed this line. The mod isn't even used in `credentials.rs` at the moment. I am surprised that doesn't emit some warning.

Anyways, I will connect the pieces then revisit this cleanup. Closing.

Issue #863, opened 2025-01-27
Resolve warning about hyperlinks in documentation.


mpeddada1:
Follow up to https://github.com/googleapis/google-cloud-rust/pull/801

Calling `cargo doc` results in the following warning for hyperlinks/html tags that span multiple lines:

![Image](https://github.com/user-attachments/assets/45b0f69e-741f-4ce7-90f8-068de84c94fe)

mpeddada1:
The issue appears to be due to the use of backticks in https://github.com/googleapis/google-cloud-rust/blame/19effa146d0e968e62544143fb8a0f0cc59120db/src/generated/type/src/model.rs#L748. Running `cargo doc` with ""  works without warnings:
```
<a href="http://www.unoosa.org/pdf/icg/2012/template/WGS_84.pdf">WGS84
/// standard</a>. Values must be within normalized ranges.
```

Issue #763, opened 2025-01-21
Use goldmark's linkify to automatically process links


coryan:
Maybe we should use the `Linkify` extension to automatically link URLs

https://pkg.go.dev/github.com/yuin/goldmark#readme-linkify-extension

PR #876, opened 2025-01-28
feat: enable warnings for invalid html tags

mpeddada1:
Fixes https://github.com/googleapis/google-cloud-rust/issues/863.

This PR also fixes how hyperlinks are rendered (by surrounding them with quotes (") instead of backticks (`) which addresses the warnings we saw in #863. 

Verified with `cargo doc --open`

Screenshot 2025-01-28 at 1 15 44 PM


generator/internal/language/rust.go
```diff
@@ -1060,6 +1060,10 @@ func escapeUrls(line string) string {
 			// Skip adding <> if the url is already surrounded by angled brackets.
 			escapedLine.WriteString(line[lastIndex:match[1]])
 			lastIndex = match[1]
+		} else if strings.Contains(line[lastIndex:match[0]], "<a href") {
```
coryan, line 1063: This is good, but it is growing in complexity.  Next time we should consider #763 .

coryan (approved):
nit: I may have separated this PR in two.

generator/internal/language/rust.go
mpeddada1: Thanks! I didn't realize goldmark had an extension for this. Might be more robust than manually processing every type of url. Will look into this in a follow up PR.

PR #882, opened 2025-01-28
fix(auth): SA cred missing claims

dbolduc:
Part of the work for #679 

We were missing the ...
- `kid` claim, which is equal to the `private_key_id`,  if there is one.
- `sub` claim which is the subject aka principal aka SA email.

Found by local integration testing.

Verify the values  of the claims we set in a unit test.

### DIFF
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -41,11 +41,12 @@ derive_builder = "0.20.2"
 [dev-dependencies]
 axum        = "0.8.1"
 mockall     = "0.13.1"
+rand        = "0.8.5"
+regex       = "1.11.1"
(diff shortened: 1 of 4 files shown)


src/auth/Cargo.toml
```diff
@@ -41,11 +41,12 @@ derive_builder = "0.20.2"
 [dev-dependencies]
 axum        = "0.8.1"
 mockall     = "0.13.1"
+rand        = "0.8.5"
```
coryan, line 44: For a separate PR, I think we can ask `taplo` to sort these.

coryan (approved):
BTW, there is no group auto-assigned to review PRs.  Folks may miss your PRs unless you explicitly assign reviewers.  We should fix that too.

Issue #635, opened 2025-01-07
Make enum field setters easier to use


coryan:
The setters for enum fields are too hard to use.  Consider:

https://github.com/googleapis/google-cloud-rust/blob/5cd390d9d4bc0c204d67e836e577b31a9e7fc271/src/generated/cloud/workflows/v1/src/model.rs#L202-L206

That consumes the enum field. We would need to initialize these with:

```rust
let workflow = wf::model::Workflow::default()
                .set_call_log_level(
                    wf::model::workflow::CallLogLevel::default()
                        .set_value(wf::model::workflow::call_log_level::LOG_ERRORS_ONLY),
                )
```

Ugh.... maybe we can implement `From<&str>` for the enums, so we can write:

```rust
let workflow = wf::model::Workflow::default()
                .set_call_log_level(wf::model::workflow::call_log_level::LOG_ERRORS_ONLY))
```

fiadliel:
I think the best thing here (for the user) is that `set_call_log_level` takes a `CallLogLevel` value. An `Into<CallLogLevel>` is possible of course, but here having a `From<&str>` feels very easy to mix up with an incorrect enum (for the user).

So then, what about if the enums generated by the code are not `&str`, but actually are values of `CallLogLevel`?

Two approaches here:

1. The value inside `CallLogLevel` could be a `Cow<'static, str>`, which lets you generate const values for each enum, these can actually return a `CallLogLevel` value.
2. Use an actual rust enum (a more usual approach for rust code)? But I see you already decided against that in https://github.com/googleapis/google-cloud-rust/pull/88

So to be more concrete:

```rust
pub struct CallLogLevel(std::borrow::Cow<'static, str>);

impl CallLogLevel {
    pub const fn new(v: &'static str) -> Self {
        Self(std::borrow::Cow::Borrowed(v))
    }
}

pub mod call_log_level {
  pub const LOG_ERRORS_ONLY: CallLogLevel = CallLogLevel::new("LOG_ERRORS_ONLY");
}
```

coryan:
> So then, what about if the enums generated by the code are not `&str`, but actually are values of `CallLogLevel`?

I like that idea.  Safer and we can still support unexpected enum values.

> Use an actual rust enum (a more usual approach for rust code)? But I see you already decided against that in https://github.com/googleapis/google-cloud-rust/pull/88

[...]

PR #884, opened 2025-01-28
feat: simplify enum setters

mpeddada1:
Fixes https://github.com/googleapis/google-cloud-rust/issues/635

Previous usage:

```
wf::model::Workflow::default()
                .set_labels([("integration-test", "true")])
                .set_call_log_level(
                    wf::model::workflow::CallLogLevel::default()
                        .set_value(wf::model::workflow::call_log_level::LOG_ERRORS_ONLY),
                )
                .set_state(
                    wf::model::workflow::State::default()
                        .set_value(wf::model::workflow::state::UNAVAILABLE),
                )
```

After moving away from using string slices:

```
wf::model::Workflow::default()
      .set_labels([("integration-test", "true")])
      .set_call_log_level(CallLogLevel::new("LOG_ERRORS_ONLY"))
      .set_state(wf::model::workflow::State::new("UNAVAILABLE"))
```


src/integration-tests/src/workflows.rs
```diff
@@ -155,14 +150,8 @@ main:
...
-                .set_call_log_level(
-                    wf::model::workflow::CallLogLevel::default()
-                        .set_value(wf::model::workflow::call_log_level::LOG_ERRORS_ONLY),
-                )
-                .set_state(
-                    wf::model::workflow::State::default()
-                        .set_value(wf::model::workflow::state::UNAVAILABLE),
-                )
+                .set_call_log_level(CallLogLevel::new("LOG_ERRORS_ONLY"))
+                .set_state(wf::model::workflow::State::new("UNAVAILABLE"))
```
coryan, line 154: I would prefer:

```suggestion
                .set_call_log_level(wf::model::call_log_level::LOG_ERRORS_ONLY)
                .set_state(wf::model::workflow::state::UNAVAILABLE)
```

or something similar, is that possible?  We want to use the constants to avoid typos.

generator/internal/language/templates/rust/common/enum.mustache
```diff
@@ -18,13 +18,12 @@ limitations under the License.
...
+pub struct {{Codec.Name}}(std::borrow::Cow<'static, str>);
 
 impl {{Codec.Name}} {
-    /// Sets the enum value.
-    pub fn set_value<T: std::convert::Into<std::string::String>>(mut self, v: T) -> Self {
-        self.0 = v.into();
-        self
+    /// Creates a new {{Codec.Name}} instance.
+    pub const fn new(v: &'static str) -> Self {
+        Self(std::borrow::Cow::Borrowed(v))
```
coryan, line 26: We also need a way to construct these from a `std::string::String`, in case a customer built a string dynamically, something like:

```suggestion
    /// Creates a new {{Codec.Name}} instance.
    pub const fn new(v: &'static str) -> Self {
        Self(std::borrow::Cow::Borrowed(v))
   }
   ... the existing code ...
}

impl std::convert::From<std::string::String> for {{Codec.Name}} {
  fn from(value: std::string::String) -> Self {
    Self(std::borrow::Cow::Owned(value))
  }
}
```

coryan (changes requested):

src/integration-tests/src/workflows.rs
mpeddada1: Good point, done.

generator/internal/language/templates/rust/common/enum.mustache
mpeddada1: Ah missed this, thank you. Added unit testing to verify this for a generated enum under `src/wkt/generated/` but open to suggestions on a better location.

coryan (approved):
Great change, thanks!
Please make sure the commit message recommends using the constants instead of string literals.

PR #892, opened 2025-01-29
feat(generator): improve client documentation

### LINKED ISSUE #626: Improve documentation for `client::*` structs
Some ideas...

image

coryan:
Fixes #626

### DIFF
--- a/generator/testdata/rust/protobuf/golden/iam/v1/src/client.rs
+++ b/generator/testdata/rust/protobuf/golden/iam/v1/src/client.rs
@@ -18,15 +18,9 @@
 use crate::Result;
 use std::sync::Arc;
 
-/// An implementation of [crate::stubs::IAMPolicy] to make requests with.
+/// Implements a client for the .
 ///
-/// `IAMPolicy` has various configuration parameters, but the defaults
-/// are set to work with most applications.
-///
-/// `IAMPolicy` holds a connection pool internally, it is advised to
-/// create one and the reuse it.  You do not need to wrap `IAMPolicy` in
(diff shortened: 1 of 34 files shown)


generator/testdata/rust/protobuf/golden/iam/v1/src/client.rs
```diff
@@ -18,15 +18,9 @@
 use crate::Result;
 use std::sync::Arc;
 
-/// An implementation of [crate::stubs::IAMPolicy] to make requests with.
+/// Implements a client for the .
```
mpeddada1, line 21: @coryan It looks like we might be missing the API name here?

coryan: Good catch.  The title is missing, yes.  Looks like this is an existing problem:

https://github.com/googleapis/google-cloud-rust/blob/86e412ac88540aa58505de7a2023b74b310745a1/generator/testdata/rust/protobuf/golden/iam/v1/README.md?plain=1#L1

The root cause is that these test files (in golden/iam/v1) are generated without a service config file:

https://github.com/googleapis/google-cloud-rust/blob/86e412ac88540aa58505de7a2023b74b310745a1/generator/internal/sidekick/sidekick_test.go#L90-L96

I am not worried.  This file is never exposed to customers or shipped with packages.

mpeddada1: Ah got it, thanks for looking into this!

mpeddada1 (approved):

Issue #806, opened 2025-01-23
Bootstrap E2E Integration Tests for Auth component


dbolduc:
## Background

We want integration tests for `gcp-sdk-auth` which. These will be triggered on every PR / post-merge like the existing `integration` triggers in the `rust-sdk-testing` project. They will just run in the `rust-auth-testing` project.

## End goal

At the end of this exercise, we should have:
- infrastructure as code, managed by terraform, for as much of the test resources as we can
- `auth-integration` builds triggered on PR / post-merge
- one integration test that validates service account credentials

## Tasks

- [x] Add placeholder GCB build yaml. It probably just runs `cargo build`.
- [x] Bootstrap auth terraform
- [x] Introduce auth testing resources / SAs managed by terraform
- [x] Configure auth GCB (triggers, repository links) via terraform
- [x] ADC support for SA creds
- [x] Add auth-integration-tests package with placeholder SA credential test
- [x] Run the SA credential integration test from our GCB build yaml
- [x] Write a real integration test for SA credentials
- [x] README.md for running integration tests in a project other than `rust-auth-testing`
- [x] Add build cache

PR #901, opened 2025-01-30
feat(auth): SA creds

dbolduc:
Part of the work for #679 and #806 

Support service account ADC.

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 pub(crate) mod mds_credential;
-pub(crate) mod service_account_credential;
+mod service_account_credential;
 pub(crate) mod user_credential;
 pub(crate) mod util;
 
@@ -263,6 +263,7 @@ pub async fn create_access_token_credential() -> Result<Credential> {
         )?;
     match cred_type {
         "authorized_user" => user_credential::creds_from(js),
+        "service_account" => service_account_credential::creds_from(js),
         _ => Err(CredentialError::non_retryable(format!(
             "Unimplemented credential type: {cred_type}"
         ))),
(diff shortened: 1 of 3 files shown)


src/auth/src/credentials.rs
```diff
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 pub(crate) mod mds_credential;
-pub(crate) mod service_account_credential;
+mod service_account_credential;
```
sai-sunder-s, line 16: We kept these at crate level visibility because we wanted to avoid exposing exact credential types unless required.

Is there a reason to make this change?

src/auth/src/credentials/service_account_credential.rs
```diff
@@ -54,18 +66,16 @@ impl std::fmt::Debug for ServiceAccountInfo {
     }
 }
 
-#[allow(dead_code)] // TODO(#679) - implementation in progress
 #[derive(Debug)]
-pub(crate) struct ServiceAccountCredential<T>
```
sai-sunder-s, line 59: why should we make these structs visible outside?

src/auth/src/credentials.rs
dbolduc: Now it is private. `pub mod ...` would make it externally visible. It did not need to be `pub(crate)` because only this file needs it.

src/auth/src/credentials/service_account_credential.rs
dbolduc: Ditto, I made it so `credentials.rs` cannot use this type directly.

src/auth/src/credentials.rs
sai-sunder-s: Got it.

sai-sunder-s (approved):

PR #902, opened 2025-01-30
ci(auth): add SA integration test

dbolduc:
Part of the work for #806 

Add a package for `auth-integration-tests`. Right now it is one test, guarded by the `run-integration-tests` feature.

Update auth's `cloudbuild.yaml` to run the tests in this new package.

Add a test that verifies SA credentials. Hopefully the code makes it clear what we are doing. I will wait to refactor until we have another test.

I will take a TODO to abstract the project ID, and add a README for running the integration tests locally.

### DIFF
--- a/src/auth/.gcb/integration.yaml
+++ b/src/auth/.gcb/integration.yaml
@@ -16,12 +16,12 @@ options:
   dynamic_substitutions: true
   substitutionOption: 'ALLOW_LOOSE'
   logging: CLOUD_LOGGING_ONLY
+  env:
+    - GOOGLE_CLOUD_PROJECT=${PROJECT_ID}
 serviceAccount: 'projects/${PROJECT_ID}/serviceAccounts/integration-test-runner@${PROJECT_ID}.iam.gserviceaccount.com'
 steps:
   - name: 'rust:1.84-bookworm'
     script: |
       #!/usr/bin/env bash
       set -e
-      # TODO(#806) - For now, we just build the code. Eventually we will run
-      # real integration tests.
-      cargo build -p gcp-sdk-auth
+      cargo test --features run-integration-tests -p auth-integration-tests
--- a/src/auth/integration-tests/tests/driver.rs
+++ b/src/auth/integration-tests/tests/driver.rs
@@ -1,0 +1,76 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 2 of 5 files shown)


dbolduc:
Neat, this is what I was hoping for:

```
running 1 test
test driver::service_account ... FAILED
failures:
---- driver::service_account stdout ----
thread 'driver::service_account' panicked at src/auth/integration-tests/tests/driver.rs:47:60:
called `Result::unwrap()` on an `Err` value: CredentialError { is_retryable: false, source: "Unimplemented credential type: service_account" }
failures:
    driver::service_account
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.21s
```

src/auth/integration-tests/tests/driver.rs
```diff
@@ -0,0 +1,75 @@
...
+    use gax::error::Error;
+    use gax::options::ClientConfig as Config;
+    use scoped_env::ScopedEnv;
+    use secretmanager::client::SecretManagerService;
+
+    type Result<T> = std::result::Result<T, gax::error::Error>;
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[serial_test::serial]
+    async fn service_account() -> Result<()> {
```
coryan, line 27: nit: in the other integration tests we keep the bulk of the code outside the `#[cfg(all(test, feature = ...]` to always compile and lint the code, even if we do not run it.  Consider doing the same here.

src/auth/.gcb/integration.yaml
```diff
@@ -16,12 +16,12 @@ options:
   dynamic_substitutions: true
   substitutionOption: 'ALLOW_LOOSE'
   logging: CLOUD_LOGGING_ONLY
+  env:
+    - GOOGLE_CLOUD_PROJECT=${PROJECT_ID}
```
coryan, line 20: Consider adding a cache, these builds can get slow .

https://github.com/googleapis/google-cloud-rust/blob/ce2b7195a0ee696cbd0eeb570b8c7ec78a822554/.gcb/integration.yaml#L23

and

https://github.com/googleapis/google-cloud-rust/blob/ce2b7195a0ee696cbd0eeb570b8c7ec78a822554/.gcb/integration.yaml#L29C1-L36C40

dbolduc: My plan was to add the build cache after the build is set up.

There is a new line item in #806 to track the task.

coryan (approved):

src/auth/integration-tests/tests/driver.rs
dbolduc: Ack, thanks. I did not understand why the other tests were written that way.

I started making this change. The change is significant enough to require another review.

I will merge this commit first in the name of incremental progress.

PR #921, opened 2025-01-31
fix(generator/golang): fix testing and regressions

coryan:
I broke the golang generator in #895: it found no templates and
therefore generated no code and nothing changed in the golden files.
Then I introduced regressions in #905 and #912. This restores the
testing and fixes the golden files.


julieqiu:
Is there a test we could add to prevent this in the future? Otherwise LGTM

julieqiu (approved):

coryan:
Good catch, included tests for Go and Rust.

PR #924, opened 2025-01-31
ci(auth): terraform for SA integration tests

dbolduc:
Part of the work for #806 

Add the terraform configuration needed to set up the service account integration test.

I think it is cleaner to keep the setup for each test in its own module. (vs. splitting it across `resources/` and `grants/`).

I am considering flattening the structure of `resources/` and `grants/`.

### DIFF
--- a/src/auth/.gcb/builds/service_account_test/main.tf
+++ b/src/auth/.gcb/builds/service_account_test/main.tf
@@ -1,0 +1,83 @@
+# Copyright 2025 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+variable "project" {}
+variable "test_runner" {}
+
+data "google_project" "project" {
+}
+
+# The service account we use as our principal for testing service account credentials.
+data "google_service_account" "test-sa-creds-principal" {
+  account_id = "test-sa-creds"
+}
+
+# Generate a key for the test service account credentials.
+resource "google_service_account_key" "test-sa-creds-principal-key" {
+  service_account_id = data.google_service_account.test-sa-creds-principal.name
+}
+
+# This secret stores the ADC json for the principal testing service account credentials.
+resource "google_secret_manager_secret" "test-sa-creds-json-secret" {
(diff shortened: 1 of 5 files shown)


src/auth/.gcb/builds/service_account_test/main.tf
```diff
@@ -0,0 +1,83 @@
...
+# The "secret" that will be accessed by the principal testing service account
+# credentials.
+#
+# Note that this is not really a "secret", in that we are not trying to hide its
+# contents.
+#
+# In order to validate our credential types, we need a GCP resource we can set
+# fine-grained ACL on. We have picked Secret Manager secrets for this purpose.
+#
+resource "google_secret_manager_secret" "test-sa-creds-secret" {
```
coryan, line 62: This is hard to reason about.  For a future PR: considering we have a lot more client libraries now: do you think you can use a different resource type?  Even if we do not have a library today that fits your needs: we can prioritize that library type to make things easier.  Maybe an IAM account?  Or a project / folder?

src/auth/.gcb/builds/service_account_test/main.tf
```diff
@@ -0,0 +1,83 @@
...
+  service_account_id = data.google_service_account.test-sa-creds-principal.name
+}
+
+# This secret stores the ADC json for the principal testing service account credentials.
+resource "google_secret_manager_secret" "test-sa-creds-json-secret" {
+  secret_id   = "test-sa-creds-json"
+  replication {
+    auto {}
+  }
+}
```
coryan, line 37: I assume you are going to rotate this secret somehow?

coryan (approved):

src/auth/.gcb/builds/service_account_test/main.tf
dbolduc: Tracked in #926

src/auth/.gcb/builds/service_account_test/main.tf
dbolduc: > do you think you can use a different resource type? 

https://cloud.google.com/iam/docs/resource-types-with-policies

Any of these work (as long as there is a `google_<resource>_iam_member` in terraform).

> This is hard to reason about.

Let me try clarifying this in the `README.md` before switching course.

Spoilers, this is what I have written:

## Test Design

For access token credentials, there are integration tests for each type of
principal (service account, authorized user, etc.).

For each principal we have:
- a secret in [SecretManager] containing the [ADC] JSON for this principal
  - the test runner service account can access this
  - so can any owners in the GCP project
- a secret in [SecretManager] containing test data
  - the test runner service account **cannot** access this
  - only the principal under test can access this secret

The steps in the test are:
1. The principal running the build pulls the ADC JSON from SecretManager.
1. We create a credential object from the ADC JSON.
1. We create a SecretManager client using these credentials.
1. We use this client to access the principal-specific secret.

[ADC]: https://cloud.google.com/docs/authentication/application-default-credentials
[SecretManager]: https://cloud.google.com/security/products/secret-manager

PR #928, opened 2025-01-31
refactor(generator): use annotations for the model

### LINKED ISSUE #653: Skip the `templatedata.go` intermediate step

coryan:
Fixes #653


coryan:
I realized I should break down #927 into smaller PRs.

julieqiu (approved):

PR #935, opened 2025-02-01
feat(logging/type): generate library

coryan:
This is a library without any services. Its types are used by other
libraries such as the upcoming `google/logging/v2` and some
libraries in `google/api/*`.

```
go -C generator/ run ./cmd/sidekick generate \
    -project-root .. \
    -specification-source google/logging/type \
    -output src/generated/logging/type \
    -codec-option package-name-override=gcp-sdk-logging-type
```


src/generated/logging/type/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Code generated by sidekick. DO NOT EDIT.
+
+[package]
+name                 = "gcp-sdk-logging-type"
+version              = "0.1.0"
+description          = "Google Cloud Client Libraries for Rust - "
```
coryan, line 20: I don't like how this comes up here and in the README.md file.  There is no serviceconfig.yaml for this directory, so we don't have a canonical source for this name.  We can do a few things:

1) Leave as-is, it is not a very important crate and people will rarely see this icky title.
2) Add an option to the Rust codec to override the title.  May be useful in other cases.
3) Create a service config yaml, that only the Rust SDK would use.
4) Combine (1) and (2), merge as-is to unblock things, and add the option (and a value) later.

My vote is for (4).

coryan:
> I think this is fine, but I wonder if we want a general strategy on how to handle these unversioned types and where to put them. Some libraries have multiple packages worth of such protos.
> 
> An example of where this could get to be a lot is the schema structures in aiplatform: https://github.com/googleapis/googleapis/tree/master/google/cloud/aiplatform/v1/schema.

Ack.  I am hoping there are not that many of these. Created #983 for the particular case of aiplatform/v1/schema.

PR #961, opened 2025-02-03
test(generator/go): add an integration test for golang generation

devoncarew:
- add an integration test for golang generation

When looking at adding a Dart backend, I'd copied the Go one (from a few days ago), but noticed that it was no longer working; the templates directory location had changed as well as several of the mustache variable names. This PR adds an integration test for the golang generator as a github workflow. It's not 100% functional, as the generation code doesn't `go build`:

```
# secretmanager
./client.go:1112:5: undefined: Metadata
./client.go:1147:17: undefined: Time
./client.go:1168:17: undefined: Time
./client.go:1171:10: undefined: Duration
./client.go:1213:24: undefined: Duration
...
```

but perhaps this test is useful as a start? The fixes are probably pretty straightforward for someone w/ more go experience.

### DIFF
--- a/.github/workflows/generator_go.yaml
+++ b/.github/workflows/generator_go.yaml
@@ -14,14 +14,12 @@
 
 name: "Client Generator: Go"
 permissions: read-all
-
 # Run CI on pushes to the main branch and on PRs against main.
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
-    branches: [ main ]
-
+    branches: [main]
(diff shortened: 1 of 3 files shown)


devoncarew:
From the CI failures it looks like I need to update the goldens for `generator/testdata/go/protobuf/golden/iam/v1/go.mod`. I can do that, but I'm wondering if I should use that api as the one to use for an integration test?

devoncarew:
closing for now -

Issue #976, opened 2025-02-04
Streaming API (from unstable-stream feature) doesn't set page_token in requests


fiadliel:
#### Environment details

  - Programming language: Rust
  - OS: MacOS
  - Language runtime version: 1.84.1
  - Package version: d5faa567e78a5fee9282c9e0265ff3c40e700556

#### Steps to reproduce

  1. Enable the `unstable-stream` feature.
  1. Run a query that requires paging to get all results (I was using `ArtifactRegistry::list_versions`, call `.stream`, and collect results.
  2. Look at the requests being sent; there is no `nextPage` parameter; also, the calls never stop, as the first batch of data is always returned.

I think the issue is due to setting the token on a cloned request that is never used, rather than the builder used to send the next query.

*Suggestion:* maybe scattering some `#[must_use]` on the builder setters could help here in future.

PR #977, opened 2025-02-04
fix(generator/rust): set next_page token when streaming

fiadliel:
Modify the builder to set the page token when iterating over a stream, rather than modifying a cloned request.

Part of the work for #976

### DIFF
--- a/generator/internal/rust/templates/crate/src/builders.rs.mustache
+++ b/generator/internal/rust/templates/crate/src/builders.rs.mustache
@@ -90,8 +90,8 @@ pub mod {{Codec.ModuleName}} {
         pub async fn stream(self) -> gax::paginator::Paginator<{{OutputType.Codec.QualifiedName}}, gax::error::Error> {
             let token = gax::paginator::extract_token(&self.0.request.page_token);
             let execute = move |token: String| {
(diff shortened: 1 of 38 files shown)


coryan:
Thanks for the PR!  It all looks good. I would like to add an integration test before declaring success on #976 (which I would be happy to do myself, I am not asking your to do it). Thoughts?

coryan (approved):

fiadliel:
> Thanks for the PR! It all looks good. I would like to add an integration test before declaring success on #976 (which I would be happy to do myself, I am not asking your to do it). Thoughts?

Well, it's a good idea. I'm happy to do it either. I'm not sure what needs to be done here other than check that one response returns `nextPageToken`, and then that is set in the next request. 

It might be worth considering a runtime assert that two successive `pageToken` values are different? I can't think of a situation where that should be supported, and it would usually have bad outcomes.

dbolduc (approved):

coryan:
Thanks again for the PR.  I like your idea of checking the token values.  We should be able to test that with a mock, which seems less likely to flake.  I do think we need one test against production to make sure this delivers the values we want (without turning into a test of the service).  I will write those in a jiffy.

PR #985, opened 2025-02-04
ci(auth): reorganize terraform for local use

dbolduc:
Motivated by #806 

Isolate GCB configuration into its own module. This makes it easier to ignore the GCB set up, which we do not need when setting up integration test resources in a local test project.

I wanted to rename `triggers` -> `cloudbuild` but renaming modules in terraform is annoying. (It is [doable](https://developer.hashicorp.com/terraform/language/modules/develop/refactoring#renaming-a-module-call), but probably not worth doing).


coryan (changes requested):
I don't think this will help you do what you want to do.  You still have `backend.tf` in the same directory, with a hardcoded bucket.  If you run this with a different project terraform will *destroy* all the resources in the old project and then create a new set of resources in the new project.

The backend acts as a globally unique database describing the state of your infrastructure.

If you want to automate resource creation for developer projects you need to either:

- List all the developer projects and iterate over them, calling all the modules for each (permissions might be a problem for this approach), or
- Refactor all of this to a reusable module, then use that module with a fixed backend for "production" and then each developer can use a local backend (not great) or their own backend (a bit of a hack, but could be done)


Then again, I may misunderstand how you intend people to use this.

dbolduc:
> You still have `backend.tf`

My workaround was to override the backend locally.

I added the instructions to this PR. Hopefully that clarifies things:

https://github.com/googleapis/google-cloud-rust/blob/acd2ac711cb3d6cf724963cdd1a1964bb0f39415/src/auth/integration-tests/README.md#in-your-own-test-project

PTAL

coryan (approved):

PR #992, opened 2025-02-05
feat(servicecontrol/v1): generate library

### LINKED ISSUE #932: Generate library for `google/api/servicecontrol/v1/servicecontrol.yaml`
Blocked on

- [x] #935 

The generation fails, we are missing some common type:

```
go -C generator/ run ./cmd/sidekick rust-generate -project-root .. -service-config google/api/servicecontrol/v1/servicecontrol.yaml
```

```
2025/02/01 11:17:34 INFO Preparing cargo workspace to get new package
2025/02/01 11:17:34 INFO Generating new library code and adding it to git
/Users/coryan/.cargo/bin/cargo fmt: exit status 1
error: expected `,`, or `}`, found `.`
   --> /Users/coryan/rust-generate/src/generated/api/servicecontrol/v1/src/model.rs:819:25
    |
805 | pub struct LogEntry {
    |            -------- while parsing this struct
...
819 |     pub severity: google.logging.type::LogSeverity,
```

coryan:
Fixes #932


src/generated/api/servicecontrol/v1/Cargo.toml
```diff
@@ -0,0 +1,41 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Code generated by sidekick. DO NOT EDIT.
+
+[package]
+name                 = "google-cloud-api-servicecontrol-v1"
```
codyoss, line 18: This is fine, but if we decide to flatten out the usage of devtools from crate naming we should do the same for this directory as well.

codyoss (approved):

src/generated/api/servicecontrol/v1/Cargo.toml
coryan: Yup.  There is also `google/cloud/privacy/dlp` (see #917).  I think the question is how many of these directories with multiple services we have.  We can use manual overrides if there are just a handful.  If there are "lots" then we may want to add some code to handle them.

PR #1011, opened 2025-02-05
feat(generator): have mustache fail if there are unresolved template variables

devoncarew:
- have mustache fail if there are unresolved template variables

I found a flag on the cbroglie/mustache library to fast fail if there are variable references in the templates that don't resolve; likely useful to enable?


codyoss (approved):

coryan (approved):

coryan:
Thanks! Cool find.

PR #1012, opened 2025-02-05
chore(ci): have github workflows run fewer times

devoncarew:
- have github workflows run fewer times (just on PRs, or, pushes to main)

I noticed that most checks where running twice per PR; this should just run a workflow once per PR (or PR change), and once when that PR lands in main.

### DIFF
--- a/.github/workflows/generator.yaml
+++ b/.github/workflows/generator.yaml
@@ -13,9 +13,12 @@
 # limitations under the License.
 
 name: Client Generator
-on: [push, pull_request]
-permissions:
-  contents: read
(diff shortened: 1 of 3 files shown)


coryan (approved):
Thanks!

Eventually we will have release branches too, but we can add those to the workflows when we know what they will be named 😀

devoncarew:
> Eventually we will have release branches too, but we can add those to the workflows when we know what they will be named 😀

I _think_ you could just not specify the list of branches for the `pull_request` section? I'm cargo-culting this from the dart-lang repo's workflows.

PR #1013, opened 2025-02-05
docs: add set up guides

coryan:
Part of the work for #574

### DIFF
--- a/guide/src/setting_up_rust_on_cloud_shell.md
+++ b/guide/src/setting_up_rust_on_cloud_shell.md
@@ -55,11 +55,13 @@ Cloud Shell is a great environment to run small examples and tests.
 ```
 
 <!-- markdownlint-disable MD029 -->
+
 6. Run your program, replacing `[PROJECT ID]` with the id of your project:
 
    ```shell
    cargo run [PROJECT ID]
    ```
+
 <!-- markdownlint-enable MD029 -->
 
 [cloud shell]: https://cloud.google.com/shell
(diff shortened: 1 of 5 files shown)


guide/src/setting_up_rust_on_cloud_shell.md
```diff
@@ -0,0 +1,79 @@
...
+
+   ```shell
+   cargo add tokio --features macros
+   ```
+
+1. Edit your project to use the Secret Manager client library:
+
+   ```shell
+   cat >src/main.rs <<_EOF_
+   #[tokio::main]
```
codyoss, line 55: nit: Could this be written in rust and included with an anchor similar to what you did in the init client guide? (Same goes for the other inlined code)

codyoss (approved):

guide/src/setting_up_rust_on_cloud_shell.md
coryan: I like that idea.  The obvious thing renders all wrong:

```
1. Bla bla

    ```rust,ignore,noplayground
    {{#include ../samples/src/bin/getting_started.rs:all}}
    ```
```

Renders to:

![image](https://github.com/user-attachments/assets/f3a78199-2e27-49ef-a8d4-f4d6cd88ff2e)


What I did breaks the itemized list.  I think the tradeoff makes sense (the code is compiled and looks nicer), but let me know what you think.

![image](https://github.com/user-attachments/assets/337b9cb4-40e6-4659-a8d2-55eed3109bc9)

codyoss: > What I did breaks the itemized list. I think the tradeoff makes sense (the code is compiled and looks nicer), but let me know what you think.

LGTM

Issue #987, opened 2025-02-04
Start generating crates with "google-cloud-" prefix


codyoss:
We want to change the namespace in which we generate crates from "gcp-sdk-" to "google-cloud-".

- [x] Preserve original names for all existing released packages
- [x] Update the default prefix
- [x] Update any existing yet to be released crates
- [x] Update gax crate
- [x] Update auth crate
- [x] Update longrunning crate
- [x] Migrate released crates to new namespace
- [x] Delist/Turndown gcp-sdk crates

codyoss:
Crates generated from the devtools folder include `devtools` in the name. We may want to fix this -- maybe as its own issue.

coryan:
I created a capstone release for all the `gcp-sdk-*` crates that points to the new crate names.  I left `gcp-sdk` (without suffix) as-is because I think we do not want to release the namespace. We can change our mind about that later.

PR #1030, opened 2025-02-06
feat: update most existing generated crate names

codyoss:
The notable exclusion here is longrunning which has a conflict we are working through.

Updates: #987


coryan (approved):
Curious why not all of the packages.  This is an improvement, so let's ship it!

codyoss:
> Curious why not all of the packages. This is an improvement, so let's ship it!

Longrunning was the only generated library that I did not update here due to the existing naming conflict. Still need to vet handwritten crates. Expect that in a followup PR.

PR #1037, opened 2025-02-06
feat: update handwritten crate names to use google-cloud prefix

codyoss:
This notably leaves out our gax and auth libraries as there are existing crates with those names already today.

Updates: #987


.github/workflows/sdk.yaml
```diff
@@ -70,8 +70,8 @@ jobs:
           cargo tarpaulin --out xml \
             --package gcp-sdk-auth \
             --package gcp-sdk-gax \
-            --package gcp-sdk-lro \
-            --package gcp-sdk-wkt
+            --package google-cloud-lro \
```
coryan, line 73: Why not rename `gcp-sdk-gax` too?

src/gax/Cargo.toml
```diff
@@ -13,46 +13,46 @@
 # limitations under the License.
 
 [package]
-name                 = "gcp-sdk-gax"
-version              = "0.2.0"
-description          = "Google Cloud Client Libraries for Rust"
+authors.workspace    = true
```
coryan, line 16: nit: I think it is less distracting to keep the `*.workspace` entries at the end 🤷

coryan (approved):

.github/workflows/sdk.yaml
codyoss: It exists. I want to make sure we have a plan in place with the other project first: https://crates.io/crates/google-cloud-gax

src/gax/Cargo.toml
codyoss: I will double check but I did not intentionally change this. I wonder if it was my editor or CLI...

codyoss: restored, I think my editors built in formatted like things alphabetical.

.github/workflows/sdk.yaml
coryan: Oh, I had missed that one. Yes, we should wait.

PR #1075, opened 2025-02-08
docs: add user guide chapter on long-running ops

### LINKED ISSUE #1052: How to start and wait for a long-running operation

coryan:
This chapter covers how to (1) start a LRO (long-running operation) and
then manually wait for completion, (2) start a LRO and just await a
future that is satisfied when the LRO completes, and (3) start a LRO and
poll a stream-like object to get intermediate results and any
intermediate updates.

I used the Speech-to-Text service for this example because it has a free
tier, it is easy to set up, and does not require cleaning up resources
after running the examples.

The code snippets are compiled and run as part of the integration tests.

Fixes #1052


guide/src/working_with_long_running_operations.md
```diff
@@ -0,0 +1,321 @@
...
+{{#rustdoc_include ../samples/src/lro.rs:manual-match-success}}
+```
+
+Note that extraction of the value may fail if the type does not match what the
+service sent.
+
+All types in Google Cloud may add fields and branches in the future. While this
+is unlikely for a common type such as `Operation`, it happens frequently for
+most service messages. The Google Cloud Client Libraries for Rust mark all
+strucs and enums as `#[non_exhaustive]` to signal that such changes are
```
codyoss, line 243:
```suggestion
structs and enums as `#[non_exhaustive]` to signal that such changes are
```

guide/samples/src/lro.rs
```diff
@@ -0,0 +1,216 @@
...
+    // ANCHOR_END: call-manual
+
+    Ok(())
+}
+// ANCHOR_END: start
+
+// ANCHOR: automatic
+pub async fn automatic(project_id: &str) -> crate::Result<()> {
+    // ANCHOR: automatic-use
+    use speech::Poller;
```
codyoss, line 71: off topic: I am not sure it is a good idea or not, but I know some crates have a prelude of there own. If we end up needing a bunch of other imports to make functionality work, we could consider having something similar.

codyoss (approved):

PR #1117, opened 2025-02-11
feat(wkt): support `Any` inside an `Any`

### LINKED ISSUE #1090: Support `wkt::Any` as metadata type for LROs
I guess the `Any` type should also implement `Message` so we can put anys inside anys.  Weird, but apparently used by at least one service:

https://github.com/googleapis/googleapis/blob/f6c71177d0c954717065ab9d17e995590ce74b7c/google/cloud/redis/cluster/v1/cloud_redis_cluster.proto#L115-L118
https://github.com/googleapis/googleapis/blob/master/google/cloud/redis/cluster/v1/cloud_redis_cluster.proto#L117

coryan:
Apparently some of our APIs like to store an `Any` object inside another `Any` object.

Fixes #1090

### DIFF
--- a/src/wkt/src/message.rs
+++ b/src/wkt/src/message.rs
@@ -14,12 +14,84 @@
 
 //! Define traits required of all messages.
 
+pub(crate) type Map = serde_json::Map<String, serde_json::Value>;
+use crate::AnyError as Error;
+
 /// A trait that must be implemented by all messages.
 ///
 /// Messages sent to and received from Google Cloud services may be wrapped in
 /// [Any][crate::any::Any]. `Any` uses a `@type` field to encoding the type
 /// name and then validates extraction and insertion against this type.
-pub trait Message {
+pub trait Message: serde::ser::Serialize + serde::de::DeserializeOwned {
     /// The typename of this message.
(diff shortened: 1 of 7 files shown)


src/wkt/src/message.rs
```diff
@@ -14,12 +14,83 @@
...
+pub(crate) type Map = serde_json::Map<String, serde_json::Value>;
+use crate::AnyError as Error;
+
 /// A trait that must be implemented by all messages.
 ///
 /// Messages sent to and received from Google Cloud services may be wrapped in
 /// [Any][crate::any::Any]. `Any` uses a `@type` field to encoding the type
 /// name and then validates extraction and insertion against this type.
-pub trait Message {
+pub trait Message: serde::ser::Serialize + serde::de::DeserializeOwned {
```
codyoss, line 25: Instead of adding these to the trait can more pub(crate) fns just be added like you did for `to_json_object`?

codyoss (approved):

src/wkt/src/message.rs
codyoss: Works for me!

PR #1130, opened 2025-02-11
feat(securitycenter/v2): generate library

codyoss:
Fixes: #1129


coryan (approved):

coryan:
Pro-tip: avoid sequential names when generating libraries. Maybe generate a library where there is some space between them.

Issue #1122, opened 2025-02-11
google/cloud/retail/v2/retail_v2.yaml


codyoss:
```
go -C generator/ run ./cmd/sidekick rust-generate \                                                                                  
    -project-root .. \
    -service-config google/cloud/retail/v2/retail_v2.yaml
```

Output:

```
error: module has the same name as its containing module
    --> src/generated/cloud/retail/v2/src/model.rs:8003:1
     |
8003 | / pub mod model {
8004 | |     #[allow(unused_imports)]
8005 | |     use super::*;
...    |
8384 | |     }
8385 | | }
     | |_^
     |
     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#module_inception
     = note: `-D clippy::module-inception` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(clippy::module_inception)]`

error: could not compile `google-cloud-retail-v2` (lib) due to 1 previous error
```

PR #1145, opened 2025-02-11
feat(sidekick): add message-name-override flag

codyoss:
Add a flag to specify overriding of messages names. This can be useful for when there are conflicts within a given programming language.

Updates: #1122

### DIFF
--- a/generator/internal/api/overrides.go
+++ b/generator/internal/api/overrides.go
@@ -1,0 +1,23 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+func ApplyNameOverrides(model *API, msgNameOverrides map[string]string) {
+	for _, msg := range model.State.MessageByID {
+		if override, ok := msgNameOverrides[msg.ID]; ok {
+			msg.Name = override
+		}
+	}
+}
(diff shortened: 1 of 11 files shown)


coryan (changes requested):
I think we need a couple more tests and some documentation.

## Tests

1. It would be useful to have an example of this in action. 
   * Maybe expand sidekick_test.go to use this feature with an existing proto in testdata?
   * OTOH, sidekick_test.go is getting long, maybe split it or add a new test.
1. I think this works for cross-reference links in the documentation, but it would be nice to have a test.
   * It is more or less obvious that something like `[Foo][google.cloud.blah.v1.Foo]` will "work" even if `Foo` is renamed to `Bar` (though "works" means: the link will point to `Bar` but the documentation will read all wrong).
   * OTOH, `[Foo][]` is less obvious to me.

## Docs

I think this is subtle enough that we should write the alternatives in a doc, or a longer bug report.

1. I wonder if/how this works when cross referencing messages from another library if the source library has a name override.
    * I think it does not work, but we could also add the override in the downstream library and that fixes it?
    * Or these renames are supposed to be part of the global file?
    * We should document our thinking on that.
1. I thought of alternatives, these are worse (I think):
   * Renaming the `model` module seems more disruptive and creates more surprising behavior for customers.
   * Removing the `model` module altogether gives us problems with the existing modules (`client` and `builders` may conflict, or `Result` and `Error` from the top-level).
1. One alternative that (maybe) is not terrible is to rename *just* the module so instead of `pub mod model {` we generate `pub mod modelz {`.  
   * That seems like it would be less general (only works or makes sense for Rust)
   *  It is less disruptive: the Protobuf documentation mostly would match maybe?
   * It still requires global renaming to work.

Maybe my alternative is no good.

codyoss:
> I think this is subtle enough that we should write the alternatives in a doc, or a longer bug report.

That is a valid critique -- I will create a spin-off issue around this problem space to write this down. Reading your feedback, I still think the rough idea in this PR is the solution I will purpose. I will write some more detailed tests though to see if that changes my mind or if there are limitations I am not currently thinking off -- the cross referencing stuff is something I did not consider.

codyoss:
Closing as after some discussion I believe we will take a different approach.

PR #1154, opened 2025-02-12
feat(wkt): Support jiff From conversions for Duration and Timestamp

### LINKED ISSUE #1153: Support conversions to/from datatypes in jiff time handling library
jiff is a promising time (and timezone) handling crate based on the design of the Temporal JavaScript API. It is planned for a long-term 1.0 version to be released later in the year, see https://github.com/BurntSushi/jiff/releases/tag/0.2.0

I expect it to be a very important part of the library ecosystem, so it would be nice to get basic support for it, at least with relevant `From`/`TryFrom` implementations. The types I'm thinking of here are `google_cloud_wkt::Duration` and `google_cloud_wkt::Timestamp`. Due to rules on where trait implementations can be placed, probably the easiest place is to add them in this repository under a feature flag.

fiadliel:
jiff is a promising time handling crate based on the design of the Temporal JavaScript API. It is planned for a long-term 1.0 version to be released later in the year.

This commit adds support for converting to and from `jiff::SignedDuration` and `jiff::Timestamp` for version 0.2.

To ease the transition between version 0.2 and 1.0, I've named the feature `jiff-0_2`, so that support for multiple versions could coexist. But if preferred, I can rename the feature to just `jiff`.

Fixes #1153

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2471,6 +2471,7 @@ dependencies = [
  "bytes",
  "chrono",
  "google-cloud-wkt",
+ "jiff",
  "serde",
  "serde_json",
  "serde_with",
@@ -2902,6 +2903,18 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "d75a2a4b1b190afb6f5425f10f6a8f959d2ea0b9c2b1d79553551850539e4674"
 
 [[package]]
+name = "jiff"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "ba926fdd8e5b5e7f9700355b0831d8c416afe94b014b1023424037a187c9c582"
+dependencies = [
+ "log",
+ "portable-atomic",
+ "portable-atomic-util",
+ "serde",
+]
+
+[[package]]
 name = "jobserver"
 version = "0.1.32"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


coryan:
Generally this looks good.  I assume the motivation for naming the feature `jiff-0_2` is to provide (eventually) `jiff-0_3` and `jiff` for a future 1.0 release?  I would like to avoid an implicit commitment to support multiple versions.  Though I guess we can always remove supported versions while our own version is 0.x (we would need to bump the `0.x` to 0.${x + 1}`, of course).
Thoughts?

PS: not asking you to change anything yet, just trying to understand motivations and tradeoffs.

fiadliel:
Updated the feature name as `jiff`.

Also updated the `Timestamp` tests to verify that the jiff and wkt values both represent the same point in time (otherwise, it proves that the first value in the test case was converted to the other, but not that the test case values were wisely chosen).

coryan:
I think we are going to wait until `jiff` gains more traction before taking a dependency on it.  It looks very promising, and once it reaches 1.0 I think the case is easier.  For now, we want to be conservative on adding extra deps to our public API (even if protected by feature flags).

I hope you do not feel like you wasted a lot of time on this.  We are still refining our thinking about how to manage deps for Rust.

fiadliel:
No hard feelings here - it's fair. Also given the beta nature of both these libraries _at this point in time_, it's not going to inconvenience a lot of users.

I'll close this PR for now, but you can use it as reference code in the future if you want.

PR #1185, opened 2025-02-13
fix(generator/rust): inconsistent `HasServices`

coryan:
Found this while trying to generate osconfig/agentendpoint/v1, which failed for other reasons, but also uncovered this bug.


codyoss (approved):
Good find!

PR #1186, opened 2025-02-13
feat(apikeys/v2): generate library

### LINKED ISSUE #1184: Generate library for `google/api/apikeys/v2/apikeys_v2.yaml`

coryan:
Fixes #1184


codyoss (approved):

codyoss:
I will try to get around to doing the flattening change today for some crates. This name in particular is not super great. But I will fix it later

coryan:
> I will try to get around to doing the flattening change today for some crates. This name in particular is not super great. But I will fix it later

Thanks. Remember that we can override specific names too.  Not desirable if we have to change 50 names, but Okay if we have to change < 10 (maybe?)

Issue #1041, opened 2025-02-07
Generate library for `google/cloud/dataplex/v1/dataplex_v1.yaml`


coryan:
Blocked by:

- [x] #1042
- [x] we tweaked the documentation upstream, so we do not need #1081

Issue #1007, opened 2025-02-05
Generate library for `google/cloud/osconfig/v1/osconfig_v1.yaml`


coryan:
Blocked by documentation problems.

- [ ] #763 

----
```
warning: this URL is not a hyperlink
    --> src/generated/cloud/osconfig/v1/src/model.rs:3627:31
     |
3627 |                     /// href="https://en.wikipedia.org/wiki/Shebang_(Unix)"
     |                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
     |
     = note: bare URLs are not automatically turned into clickable links
     = note: `#[warn(rustdoc::bare_urls)]` on by default
help: use an automatic link instead
     |
3627 |                     /// href="<https://en.wikipedia.org/wiki/Shebang>_(Unix)"
     |                               +                                     +

warning: unopened HTML tag `a`
    --> src/generated/cloud/osconfig/v1/src/model.rs:3628:56
     |
3628 |                     /// class="external"\>shebang lines</a>.
     |                                                        ^^^^
     |
     = note: `#[warn(rustdoc::invalid_html_tags)]` on by default

warning: `google-cloud-osconfig-v1` (lib doc) generated 2 warnings (run `cargo fix --lib -p google-cloud-osconfig-v1` to apply 1 suggestion)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 33.91s
   Generated /Users/coryan/rust-generate/target/doc/google_cloud_osconfig_v1/index.html
```

Issue #828, opened 2025-01-24
Generate library for Vertex `google/cloud/aiplatform/v1`


coryan:
Generated with:
```
go -C generator/ run ./cmd/sidekick generate \
    -project-root .. \
    -specification-source google/cloud/aiplatform/v1 \
    -service-config google/cloud/aiplatform/v1/aiplatform_v1.yaml \
    -output src/generated/cloud/aiplatform/v1

git add src/generated/cloud/aiplatform/v1
cargo fmt && cargo build && cargo test && cargo doc
```

The build fails with:

```
error[E0412]: cannot find type `RRF` in this scope
     --> src/generated/cloud/aiplatform/v1/src/model.rs:35355:40
      |
35340 |         pub struct Rrf {
      |         -------------- similarly named struct `Rrf` defined here
...
35355 |         impl wkt::message::Message for RRF {
      |                                        ^^^ help: a struct with a similar name exists: `Rrf`

error[E0609]: no field `order_by_data_item` on type `SearchDataItemsRequest`
   --> src/generated/cloud/aiplatform/v1/src/transport.rs:434:64
    |
434 |         let builder = builder.query(&[("orderByDataItem", &req.order_by_data_item)]);
    |                                                                ^^^^^^^^^^^^^^^^^^ unknown field
    |
    = note: available fields are: `dataset`, `saved_query`, `data_labeling_job`, `data_item_filter`, `annotations_filter` ... and 7 others

error[E0609]: no field `order_by_annotation` on type `SearchDataItemsRequest`
   --> src/generated/cloud/aiplatform/v1/src/transport.rs:436:14
    |
436 |             .order_by_annotation
    |              ^^^^^^^^^^^^^^^^^^^ unknown field
    |
    = note: available fields are: `dataset`, `saved_query`, `data_labeling_job`, `data_item_filter`, `annotations_filter` ... and 7 others

warning: unused import: `gax::query_parameter::QueryParameter`
   --> src/generated/cloud/aiplatform/v1/src/transport.rs:442:21
    |
442 |                 use gax::query_parameter::QueryParameter;
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: `#[warn(unused_imports)]` on by default

warning: unused import: `gax::query_parameter::QueryParameter`
   --> src/generated/cloud/aiplatform/v1/src/transport.rs:459:21
    |
459 |                 use gax::query_parameter::QueryParameter;
[...]

PR #1195, opened 2025-02-13
feat(generator/rust): option for rustdoc warnings

coryan:
Use an option to control the list of disabled `rustdoc` warnings. This
will let us add a few more options for generated libraries that need
them, and unblock generation of the libraries while we fix the upstream
markdown.

Use the flag in some (but not all) golden files to show that it works. That
changed the golden files a little bit, as expected.

Motivated by #1041 and #1007 and #828


codyoss (approved):
Good idea!

Issue #1210, opened 2025-02-13
Implement token cache for auth


dbolduc:
We will add a `TokenProvider` that caches tokens.

I am too lazy to detail the spec for this token cache. There are internal docs, and this suggestion: https://github.com/googleapis/google-cloud-rust/discussions/981#discussion-7915320. (Thanks, @fiadliel).

To build upon the discussion:

We will limit a thundering herd to one outgoing request from the cache.

For the initial implementation, we will have one token cache per credential. When we modify `Credential::get_token()` to accept options (e.g. custom scopes/audience), we can introduce a map for the cache then.

For refreshing, we will have calls to `get_token()` handle background refreshing (think `tokio::spawn`) when tokens are about to expire. It will not be scheduled. The application can call into `get_token()` periodically to avoid latency hits.

---

Work
- [x] `CredentialError` can be cloned
- [x] Add token cache without background refresh
- [x] Introduce background refreshing
- [x] Include token cache layer in existing credentials

coryan:
Can we close this?

PR #1213, opened 2025-02-13
feat(auth): clone for `CredentialError`

dbolduc:
Part of the work for #1210 

In the token cache, if we have N waiters for a single request, and that request errors, we want to give that error to each of the waiters. That means we need our errors to be clone-able.

We do not want to use an `Arc` instead of a `Box` to hold the error `source`. Some `source`s do not fit in an `Arc`.

We can stream the error message, and copy that into a new `CredentialError`.

If the error message is the same, the `CredentialError` is the same, as far as I am concerned.

### DIFF
--- a/src/auth/src/errors.rs
+++ b/src/auth/src/errors.rs
@@ -75,6 +75,13 @@ impl std::error::Error for CredentialError {
     }
 }
 
+impl Clone for CredentialError {
+    fn clone(&self) -> CredentialError {
+        let source = format!("{}", self.source);
+        CredentialError::new(self.is_retryable, source.into())
+    }
+}
+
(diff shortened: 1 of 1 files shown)


coryan (commented):
> If the error message is the same, the `CredentialError` is the same, as far as I am concerned.

That is not very idiomatic.  In Rust, the error type is what matters in the code, the message is for the humans.  Can we change `CredentialError` to require that the source be `Clone` too?

Changing this:

https://github.com/googleapis/google-cloud-rust/blob/148805e4085e4b2b8b9191e8da6723149ee71ba5/src/auth/src/errors.rs#L19

to

```
pub(crate) type BoxError = Box<dyn Error + Send + Sync + Clone>;
```

??

dbolduc:
Closing.

PR #1288, opened 2025-02-14
docs: link to user-guide for LRO documentation

coryan:
I think that makes the reference docs more focused, and saves us build
time because we have fewer sample code snippets.


dangazineu:
I don't mind the change, if the improvement in build time is significant--especially if we think of this as a temporary efficiency gain while the SDK in under heavy development. But I think in general, the verbose samples are a better (and copy-pasteable) user experience than being taken to a general abstract description of the concept.

dangazineu (approved):

coryan:
I need to rebase before merging. I will wait until all the other libraries are in before doing either.

Issue #1324, opened 2025-02-15
near-term work for the dart generator


devoncarew:
I have a rough plan for near-term work for the dart generator and wanted to open an issue to get some feedback.

I need some place to generate more than just few golden packages, and, would like to make it easier for other people to run the generator (or for the CI to do it, ...).

My rough plan is to add a new command to the sidekick tool (`dart`? `dart-generate`?). That may do a few things, but would at least be able to generate a package to a well-known location based on a spec ID (e.g. `'google/cloud/apihub/v1'`). It may also be able to re-generate, based on the spec ID and the info in an associated toml file.

That would always emit to something like `generator/testdata/dart/pkgs`. `pkgs/` may or may not include some hand-written packages, like for well known types.

coryan:
> I need some place to generate more than just few golden packages, and, would like to make it easier for other people to run the generator (or for the CI to do it, ...).

We could create a `dart/` top-level directory and then this should work:

```
go -C generator/ run ./cmd/sidekick refreshall -project-root ../dart
```

it should find all the subdirectories of `dart/` that have a `.sidekick.toml` file and run the generator in that directory, with the configuration for that directory.  We can share common configuration in a `dart/.sidekick.toml` file to minimize duplication.

For the golden packages, we should create a new `sidekick_data_test.go` inspired by `generator/internal/sidekick
/sidekick_test.go`.  You can regenerate all the golde files using:

```
go -C generator/ test ./...
```

> My rough plan is to add a new command to the sidekick tool (`dart`? `dart-generate`?). 

I think `refreshall` would work well as you are updating the generator.  We can certainly add something like `dart-generate` to simplify the generation of *new* libraries, but that seems premature.  I expect you will be improving the generator and updating the libraries first?

> That would always emit to something like `generator/testdata/dart/pkgs`. `pkgs/` may or may not include some hand-written packages, like for well known types.

I would rather keep `testdata` for things that are not hand-crafted.  Let's give you a real directory `/dart` or `/generator/dart`  to put your work.   We can always remove it once it grows large that we can justify a new repository.

coryan:
I got things working in #1329. Once that gets merged you can regenerate[^2] the Dart code using:

```
go -C generator/ run ./cmd/sidekick refreshall -project-root dart
```

Hopefully I structured the code in a way that makes sense, but you can change the directory structure later. We also need to add a small CI for the Dart generated code.  Likewise, you can run this to refresh  all the generator golden files:

```
[...]

PR #1327, opened 2025-02-16
fix(generator): default to root config language

coryan:
The intention is to use the language configured in the top-level
`.sidekick.toml` file, with an opportunity to override via the
command-line or with a local `.sidekick.toml` file.

The actual implementation always overrode the language to `rust` because
it provided a default value for the command-line argument.

Motivated by #1324


dangazineu (approved):
Might want to remove all those duplicated `language='rust'` entries under `src/generated` as well.

coryan:
> Might want to remove all those duplicated `language='rust'` entries under `src/generated` as well.

Will do, created #1330 so I don't forget.

PR #1341, opened 2025-02-17
feat(auth): introduce `CredentialType` enum

moricho:
This PR introduces `CredentialType` enum and replaces string-based credential type matching with it for type-safe credential handling

from: https://github.com/googleapis/google-cloud-rust/issues/1342

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -201,6 +201,41 @@ pub(crate) mod dynamic {
     }
 }
 
+/// Represents the type of credential as specified in the credentials file.
+#[derive(Debug, PartialEq)]
+pub enum CredentialType {
+    /// User account credentials, typically from gcloud auth
+    AuthorizedUser,
+    /// Service account key file credentials
+    ServiceAccount,
+    /// External account credentials (e.g., Workload Identity Federation)
+    ExternalAccount,
+}
+
+impl CredentialType {
+    fn from_str(type_str: &str) -> Result<Self> {
+        match type_str {
+            "authorized_user" => Ok(Self::AuthorizedUser),
+            "service_account" => Ok(Self::ServiceAccount),
+            "external_account" => Ok(Self::ExternalAccount),
+            _ => Err(CredentialError::non_retryable(format!(
+                "Unimplemented credential type: {type_str}"
+            ))),
+        }
+    }
+}
+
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials.rs
```diff
@@ -201,6 +201,41 @@ pub(crate) mod dynamic {
...
+    ExternalAccount,
+}
+
+impl CredentialType {
+    fn from_str(type_str: &str) -> Result<Self> {
+        match type_str {
+            "authorized_user" => Ok(Self::AuthorizedUser),
+            "service_account" => Ok(Self::ServiceAccount),
+            "external_account" => Ok(Self::ExternalAccount),
+            _ => Err(CredentialError::non_retryable(format!(
```
dbolduc, line 221: I broke this API yesterday. Sorry. It is now `non_retryable_from_str(...)`. You will need to rebase this PR, and make the code change.

src/auth/src/credentials.rs
```diff
@@ -201,6 +201,41 @@ pub(crate) mod dynamic {
     }
 }
 
+/// Represents the type of credential as specified in the credentials file.
+#[derive(Debug, PartialEq)]
+pub enum CredentialType {
```
dbolduc, line 206: Hm, this is public, but we are not exposing any way to go from a `Credential` -> `CredentialType` anywhere. Can we make it `pub(crate)` for now? Why should it be `pub`?

FYI: I think our plan in the long run is to have metadata like `CredentialType` available, at least in the error messages, so users can debug auth configuration errors.

src/auth/src/credentials.rs
```diff
@@ -201,6 +201,41 @@ pub(crate) mod dynamic {
...
+    AuthorizedUser,
+    /// Service account key file credentials
+    ServiceAccount,
+    /// External account credentials (e.g., Workload Identity Federation)
+    ExternalAccount,
+}
+
+impl CredentialType {
+    fn from_str(type_str: &str) -> Result<Self> {
+        match type_str {
```
dbolduc, line 217: comment: We have a credential type (MDS credential) that is NOT in the ADC json. So there is no string for it.

That is fine, just an FYI.

dbolduc:
Hey @moricho, are you still working on this?

dbolduc:
Closing, no activity in ~a month.

PR #1355, opened 2025-02-19
impl(auth): introduce token cache

dbolduc:
Part of the work for #1210 

Add a token cache that refreshes tokens when they expire.

We will add background refreshing in a follow up PR, because this is already a decent amount of code.

## On the thundering herd unit tests...

I struggled to avoid races. I wanted to release a token only after all waiters were ready. But there is a delta between a task spawning and the next line of code in a task executing. I felt I had to introduce some sort of timing. No uses of `tokio::sync::Notify`, `std::sync::Condvar` could save me.

I settled for releasing a token every 50 ms. We will only flake if each of 100 lines of code take >= 50ms to get scheduled. And in practice the test only takes 50ms. I am open to other suggestions, but totally fine with this.

### DIFF
--- a/src/auth/src/token_cache.rs
+++ b/src/auth/src/token_cache.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::errors::CredentialError;
 use crate::token::{Token, TokenProvider};
 use crate::Result;
 use std::sync::{Arc, RwLock};
@@ -38,14 +39,6 @@ where
     inner: Arc<T>,
 }
 
-// Returns true if we are holding an error, or a token that has expired.
-fn is_invalid(token: &Result<Token>) -> bool {
-    match token {
-        Ok(t) => t.expires_at.is_some_and(|e| e <= Instant::now().into_std()),
-        Err(_) => true,
-    }
-}
-
 // We manually implement the `Clone` trait because the Rust compiler will
 // squawk if `T` is not `Clone`, even though we only hold an `Arc<T>`. :shrug:
 impl<T: TokenProvider> Clone for TokenCache<T> {
@@ -59,60 +52,92 @@ impl<T: TokenProvider> Clone for TokenCache<T> {
     }
 }
 
+fn initial_token() -> Result<Token> {
+    Err(CredentialError::retryable_from_str("No token or error in the cache. This should never happen. Something has gone wrong. Open an issue at https://github.com/googleapis/google-cloud-rust/issues/new?template=bug_report.md."))
+}
+
 impl<T: TokenProvider> TokenCache<T> {
     // TODO(#1210) - implementation in progress
     #[allow(dead_code)]
     pub fn new(inner: T) -> TokenCache<T> {
         TokenCache {
-            token: Arc::new(RwLock::new(Err(crate::errors::CredentialError::retryable_from_str("No token in the cache. This should never happen. Something has gone wrong. Open an issue at https://github.com/googleapis/google-cloud-rust/issues/new?template=bug_report.md.")))),
+            token: Arc::new(RwLock::new(initial_token())),
             refresh_in_progress: Arc::new(Mutex::new(())),
             refresh_notify: Arc::new(Notify::new()),
             inner: Arc::new(inner),
         }
     }
 
-    // Clones the current token, in a thread-safe manner. Releases the lock on return.
-    async fn current_token(&self) -> Result<Token> {
-        self.token.read().unwrap().clone()
+    fn state(&self) -> TokenState {
+        let token = self.token.read().unwrap();
+        match &*token {
+            Ok(t) => match t.expires_at {
+                None => TokenState::Valid(t.clone()),
+                Some(e) => {
+                    let now = Instant::now().into_std();
+                    if e <= now {
+                        TokenState::Invalid(self.try_refresh_lock())
+                    } else {
+                        TokenState::Valid(t.clone())
+                    }
+                }
+            },
+            Err(_) => TokenState::Invalid(self.try_refresh_lock()),
+        }
     }
+
+    fn try_refresh_lock(&self) -> RefreshLock {
+        self.refresh_in_progress.clone().try_lock_owned()
+    }
+}
+
+type RefreshLock = std::result::Result<tokio::sync::OwnedMutexGuard<()>, tokio::sync::TryLockError>;
+
+enum TokenState {
+    Invalid(RefreshLock),
+    Valid(Token),
 }
 
 #[async_trait::async_trait]
 impl<T: TokenProvider + 'static> TokenProvider for TokenCache<T> {
     async fn get_token(&self) -> Result<Token> {
-        let token = self.current_token().await;
-
-        if !is_invalid(&token) {
-            return token;
-        }
-
-        match self.refresh_in_progress.try_lock() {
-            // Check if there are any outstanding refreshes...
-            Ok(guard) => {
-                // No refreshes. We should start one.
-                let token = self.inner.get_token().await;
-
-                // Store the token, or an updated error.
-                *self.token.write().unwrap() = token.clone();
-
-                // The refresh is complete. Release the refresh guard.
-                drop(guard);
-
-                // Notify any and all waiters.
-                self.refresh_notify.notify_waiters();
-
-                // Return here without asking for the token lock again.
-                return token;
+        match self.state() {
+            TokenState::Valid(token) => {
+                // We have a valid token, and it is not expiring any time soon.
+                // Return it immediately.
+                Ok(token)
             }
-            Err(_) => {
-                // There is already a refresh. We will await its result.
-                self.refresh_notify.notified().await;
+
+            TokenState::Invalid(refresh_guard) => {
+                match refresh_guard {
+                    // Check if there are any outstanding refreshes...
+                    Ok(guard) => {
+                        // No refreshes. We should start one.
+                        let token = self.inner.get_token().await;
+
+                        // Store the token, or an updated error.
+                        *self.token.write().unwrap() = token.clone();
+
+                        // The refresh is complete. Release the refresh guard.
+                        drop(guard);
+
+                        // Notify any and all waiters.
+                        self.refresh_notify.notify_waiters();
+
+                        // Return here without asking for the token lock again.
+                        token
+                    }
+                    Err(_) => {
+                        // There is already a refresh. We will await its result.
+                        self.refresh_notify.notified().await;
+
+                        // The refresh operation has completed. We should have a new
+                        // error/token. Return it.
+                        self.token.read().unwrap().clone()
+                    }
+                }
             }
         }
-
-        // The refresh operation has completed. We should have a new
-        // error/token. Return it.
-        self.current_token().await
     }
 }
 
--- a/src/auth/src/token_cache.rs
+++ b/src/auth/src/token_cache.rs
@@ -16,7 +16,7 @@ use crate::errors::CredentialError;
 use crate::token::{Token, TokenProvider};
 use crate::Result;
 use std::sync::{Arc, RwLock};
-use tokio::sync::{Mutex, Notify};
+use tokio::sync::Mutex;
 // Using tokio's wrapper makes the cache testable without relying on clock times.
 use tokio::time::Instant;
 
@@ -32,8 +32,6 @@ where
 
     // Tracks if a refresh is ongoing. If the lock is held, there is a refresh.
     refresh_in_progress: Arc<Mutex<()>>,
-    // Allows us to await the result of a refresh in multiple threads.
-    refresh_notify: Arc<Notify>,
 
     // The token provider. This thing does the refreshing.
     inner: Arc<T>,
@@ -46,7 +44,6 @@ impl<T: TokenProvider> Clone for TokenCache<T> {
         TokenCache {
             token: self.token.clone(),
             refresh_in_progress: self.refresh_in_progress.clone(),
-            refresh_notify: self.refresh_notify.clone(),
             inner: self.inner.clone(),
         }
     }
@@ -63,7 +60,6 @@ impl<T: TokenProvider> TokenCache<T> {
         TokenCache {
             token: Arc::new(RwLock::new(initial_token())),
             refresh_in_progress: Arc::new(Mutex::new(())),
-            refresh_notify: Arc::new(Notify::new()),
             inner: Arc::new(inner),
         }
     }
@@ -111,25 +107,22 @@ impl<T: TokenProvider + 'static> TokenProvider for TokenCache<T> {
             TokenState::Invalid(refresh_guard) => {
                 match refresh_guard {
                     // Check if there are any outstanding refreshes...
-                    Ok(guard) => {
+                    Ok(_guard) => {
                         // No refreshes. We should start one.
                         let token = self.inner.get_token().await;
 
                         // Store the token, or an updated error.
                         *self.token.write().unwrap() = token.clone();
 
-                        // The refresh is complete. Release the refresh guard.
-                        drop(guard);
-
-                        // Notify any and all waiters.
-                        self.refresh_notify.notify_waiters();
-
                         // Return here without asking for the token lock again.
                         token
                     }
                     Err(_) => {
-                        // There is already a refresh. We will await its result.
-                        self.refresh_notify.notified().await;
+                        // There is already a refresh. When the lock becomes
+                        // available, we know the refresh has completed. Once
+                        // we acquire the lock, we can release it immediately.
+                        #[allow(unused_must_use)]
+                        self.refresh_in_progress.lock().await;
 
                         // The refresh operation has completed. We should have a new
                         // error/token. Return it.
@@ -148,6 +141,7 @@ mod test {
     use crate::token::test::MockTokenProvider;
     use std::sync::{Arc, Mutex};
     use std::time::Duration;
+    use tokio::sync::Notify;
 
     static TOKEN_VALID_DURATION: Duration = Duration::from_secs(3600);
 
@@ -268,8 +262,8 @@ mod test {
         assert!(cache.get_token().await.is_err());
     }
 
-    #[tokio::test(start_paused = true)]
-    async fn token_cache_lock_contention() {
+    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+    async fn token_cache_with_token_lock_contention() {
         let now = Instant::now();
 
         let token = Token {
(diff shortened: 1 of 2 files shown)


src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,351 @@
...
+    refresh_in_progress: Arc<Mutex<()>>,
+    // Allows us to await the result of a refresh in multiple threads.
+    refresh_notify: Arc<Notify>,
+
+    // The token provider. This thing does the refreshing.
+    inner: Arc<T>,
+}
+
+// Returns true if we are holding an error, or a token that has expired.
+fn invalid(token: &Result<Token>) -> bool {
```
sai-sunder-s, line 42: can we call it is_invalid or something like that or is it not the rusty way?

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,351 @@
...
+    }
+}
+
+// We manually implement the `Clone` trait because the Rust compiler will
+// squawk if `T` is not `Clone`, even though we only hold an `Arc<T>`. :shrug:
+impl<T: TokenProvider> Clone for TokenCache<T> {
+    fn clone(&self) -> TokenCache<T> {
+        TokenCache {
+            token: self.token.clone(),
+            refresh_in_progress: self.refresh_in_progress.clone(),
```
sai-sunder-s, line 55: is the task doing the refresh also cloned? If not, this cache will permanently be in a refreshing state (if cloned at the time when a refresh was happening)

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,351 @@
...
+        }
+    }
+}
+
+impl<T: TokenProvider> TokenCache<T> {
+    // TODO(#1210) - implementation in progress
+    #[allow(dead_code)]
+    pub fn new(inner: T) -> TokenCache<T> {
+        TokenCache {
+            token: Arc::new(Mutex::new(Err(crate::errors::CredentialError::retryable_from_str("No token in the cache. This should never happen. Something has gone wrong. Open an issue at https://github.com/googleapis/google-cloud-rust/issues/new?template=bug_report.md.")))),
```
sai-sunder-s, line 67: "This should NEVER happen" seems extreme. What if an unretryable error happens, like maybe MDS is unreachable or oauth server is throwing 500s due to some bug or something like that.

sai-sunder-s: nvrm. looks like we will throw an exception at the time of token fetch in such cases and we will not write anything to this token field.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,351 @@
...
+            refresh_in_progress: Arc::new(Mutex::new(())),
+            refresh_notify: Arc::new(Notify::new()),
+            inner: Arc::new(inner),
+        }
+    }
+
+    // Clones the current token, in a thread-safe manner. Releases the lock on return.
+    async fn current_token(&self) -> Result<Token> {
+        self.token.lock().await.clone()
```
sai-sunder-s, line 76: Does rust have concept of read lock vs write lock? We would need to lock only when the token is getting updated I believe?

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,351 @@
...
+            .times(1)
+            .return_once(|| Ok(refresh_clone));
+
+        // fetch an initial token
+        let cache = TokenCache::new(mock);
+        let actual = cache.get_token().await.unwrap();
+        assert_eq!(actual, initial);
+
+        // wait long enough for the token to be expired
+        let sleep = TOKEN_VALID_DURATION;
```
sai-sunder-s, line 204: we are sleeping for an hour in the test?

I guess I am missing something here.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,351 @@
...
+            expires_at: Some(Instant::now().into_std()),
+            metadata: None,
+        };
+
+        let tp = FakeTokenProvider::new(Ok(token.clone()));
+
+        let cache = TokenCache::new(tp.clone());
+
+        // Spawn N tasks, all asking for a token at once.
+        let tasks = (0..100)
```
sai-sunder-s, line 294: why cannot we have this thundering herd try to get a token that is already expired? Then atmost 1 refresh should happen?

src/auth/src/token_cache.rs
dbolduc: We are cloning the `Arc`. The stuff in the `Arc` doesn't change.

src/auth/src/token_cache.rs
dbolduc: `tokio::time::Instant` is like a wrapper over `std::time::Instant` that lets us fake the time in a test.

So the code is more like "pretend to sleep for an hour".

https://docs.rs/tokio/latest/tokio/time/fn.advance.html

src/auth/src/token_cache.rs
dbolduc: Alternatively, I could use an expired token, and not address a failure case that can't happen. :shrug: 

Thoughts?

src/auth/src/token_cache.rs
dbolduc: I think you are noticing that I had tests for  `initial_token_*` and `expired_token_*`, but I only have a unit test for `initial_thundering_herd_*`. There is no `expired_thundering_herd_*`.

The secret is that these are the same case, given how I wrote the code. (where `is_invalid(token) == true`), so I didn't want to write the same test again.

(We need both the `initial_token_*` and `expired_token_*` to actually test that `is_invalid(token) == initial || expired` .)

I changed the test name to be `invalid_token_thundering_herd_*`. (I will add `stale_token_thundering_herd_*` tests later).

Let me know if you were asking something else.

src/auth/src/token_cache.rs
dbolduc: Yes, good catch! Fixed.

I had `RwLock` initially, but changed it when I was flailing about trying to fix my unit tests. We do not need `tokio::sync::Mutex` for this because we do not need to hold the lock over an `await` boundary.

I also added another unit test to exercise the read lock contention.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,389 @@
...
+
+        if !is_invalid(&token) {
+            return token;
+        }
+
+        match self.refresh_in_progress.try_lock() {
+            // Check if there are any outstanding refreshes...
+            Ok(guard) => {
+                // No refreshes. We should start one.
+                let token = self.inner.get_token().await;
```
coryan, line 93: You seem to have a race condition here.  You grab and release the token, then check if there is a refresh.  Some other task / thread could have updated the token after like 83 and before like 89. That is extra work that you could eliminate.

coryan (changes requested):
I think you have a race condition.  But I might be dead wrong.

src/auth/src/token_cache.rs
dbolduc: I agree.

dbolduc: I think my problem is that the state of the token cache is split across `refresh_in_progress` and `token`. And I try to update them one at a time (transitioning through an unwanted state).

I need something new to hold state and manage the transitions between states.

src/auth/src/token_cache.rs
sai-sunder-s: thats awesome!

src/auth/src/token_cache.rs
sai-sunder-s: So lets say thread A is doing refresh and has a lock on the mutex in refresh_in_progress. A clone is called when this is happening, what will be the state of the mutex in the clone?

src/auth/src/token_cache.rs
dbolduc: Ok, I think there is a simpler fix. I just need to hold the token lock when I try the refresh lock.

That will prevent a refresh from writing the token behind my back.

Thanks.

src/auth/src/token_cache.rs
dbolduc: It would still be locked in the clone. Clones of `Arc<Mutex<T>>` just share the underlying `Mutex<T>`.

Note that `tokio::sync::Mutex` cannot be cloned: https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#trait-implementations.

src/auth/src/token_cache.rs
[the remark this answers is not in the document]
dbolduc: Done.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,389 @@
...
+        }
+    }
+}
+
+impl<T: TokenProvider> TokenCache<T> {
+    // TODO(#1210) - implementation in progress
+    #[allow(dead_code)]
+    pub fn new(inner: T) -> TokenCache<T> {
+        TokenCache {
+            token: Arc::new(RwLock::new(Err(crate::errors::CredentialError::retryable_from_str("No token in the cache. This should never happen. Something has gone wrong. Open an issue at https://github.com/googleapis/google-cloud-rust/issues/new?template=bug_report.md.")))),
```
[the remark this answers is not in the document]
dbolduc: Done.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,389 @@
...
+        self.token.read().unwrap().clone()
+    }
+}
+
+#[async_trait::async_trait]
+impl<T: TokenProvider + 'static> TokenProvider for TokenCache<T> {
+    async fn get_token(&self) -> Result<Token> {
+        let token = self.current_token().await;
+
+        if !is_invalid(&token) {
```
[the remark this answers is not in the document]
dbolduc: I revamped this. We will use an enum with values `Valid`, `Invalid`, and (in a future PR) `Stale`

src/auth/src/token_cache.rs
dbolduc: Fixed this problem.

I think the one other thing that needs scrutiny is the `notify` business.

Maybe I fail to acquire the lock guard. Then we could release the lock and notify all waiters:

https://github.com/googleapis/google-cloud-rust/blob/d044da84ccb96aff14eedc837b0a4f85eda1c90b/src/auth/src/token_cache.rs#L121-L125

before we await the notify:

https://github.com/googleapis/google-cloud-rust/blob/d044da84ccb96aff14eedc837b0a4f85eda1c90b/src/auth/src/token_cache.rs#L132

Hm. I think this is a problem.

dbolduc: Thinking out loud...

If I fail to acquire the refresh lock, instead of waiting for some notify business, I could just demand the refresh lock.

When I am given the lock, it means the refresh operation has completed. Then I just return whatever is in my cache.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,414 @@
...
+
+                        // Notify any and all waiters.
+                        self.refresh_notify.notify_waiters();
+
+                        // Return here without asking for the token lock again.
+                        token
+                    }
+                    Err(_) => {
+                        // There is already a refresh. We will await its result.
+                        self.refresh_notify.notified().await;
```
fiadliel, line 132: Hope it's okay for me to comment, I'm concerned here; is there any guarantee that this happens before `notify_waiters` above? I think in theory, the guarded code could run, call `notify_waiters`, and then `notified` here has no waker.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,414 @@
...
+            token: self.token.clone(),
+            refresh_in_progress: self.refresh_in_progress.clone(),
+            refresh_notify: self.refresh_notify.clone(),
+            inner: self.inner.clone(),
+        }
+    }
+}
+
+fn initial_token() -> Result<Token> {
+    Err(CredentialError::retryable_from_str("No token or error in the cache. This should never happen. Something has gone wrong. Open an issue at https://github.com/googleapis/google-cloud-rust/issues/new?template=bug_report.md."))
```
coryan, line 56: Consider:

```suggestion
    Err(CredentialError::retryable_from_str("No token or error in the cache. This is unexpected behavior and probably indicates a bug in the client library code. Please open an issue at https://github.com/googleapis/google-cloud-rust/issues/new?template=bug_report.md."))
```

You should also consider:

- Use a different representation, such as `Option<Result<Token>>` 
- Use your own enum, such as:
  ```rust
  enum CachedToken {
    InitialState, // or None? I am not good at naming.
    LastFetchFailed(Error),
    Ok(Token),
  }
  ```

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,414 @@
...
+}
+
+impl<T: TokenProvider> TokenCache<T> {
+    // TODO(#1210) - implementation in progress
+    #[allow(dead_code)]
+    pub fn new(inner: T) -> TokenCache<T> {
+        TokenCache {
+            token: Arc::new(RwLock::new(initial_token())),
+            refresh_in_progress: Arc::new(Mutex::new(())),
+            refresh_notify: Arc::new(Notify::new()),
```
coryan, line 66: I think this is getting sufficiently complicated that it needs a design doc, maybe one or two pages or a detailed GitHub issue would do.

dbolduc: What do you want clarified in the design doc? Is it more the specification or the implementation?

coryan: The specification would help.  I think a description of why you are using whatever synchronization primitives you are using would help too.

- What alternatives exist?
- Why did we pick the ones we did?

src/auth/src/token_cache.rs
dbolduc: `Option<Result<Token>>`

I don't think this helps, we should never be in the `Option::is_none` state when we are returning a result. So an error value / expired token seem preferable.

src/auth/src/token_cache.rs
dbolduc: I fixed this.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,474 @@
...
+
+    fn state(&self) -> TokenState {
+        let token = self.token.read().unwrap();
+        match &*token {
+            Ok(t) => match t.expires_at {
+                None => TokenState::Valid(t.clone()),
+                Some(e) => {
+                    let now = Instant::now().into_std();
+                    if e <= now {
+                        TokenState::Invalid(self.try_refresh_lock())
```
coryan, line 75: Assume thread T1 this is trying to acquire the refresh lock (1).  Observe that T1 is are already holding a read lock on `self.token`.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,474 @@
...
+impl<T: TokenProvider + 'static> TokenProvider for TokenCache<T> {
+    async fn get_token(&self) -> Result<Token> {
+        match self.state() {
+            TokenState::Valid(token) => {
+                // We have a valid token, and it is not expiring any time soon.
+                // Return it immediately.
+                Ok(token)
+            }
+
+            TokenState::Invalid(refresh_guard) => {
```
coryan, line 107: Assume T2 is holding the refresh lock.... and that is why T1 cannot acquire it.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,474 @@
...
+                match refresh_guard {
+                    // Check if there are any outstanding refreshes...
+                    Ok(_guard) => {
+                        // No refreshes. We should start one.
+                        let token = self.inner.get_token().await;
+
+                        // Store the token, or an updated error.
+                        *self.token.write().unwrap() = token.clone();
```
coryan, line 115: Now T2 is trying to acquire the write lock on `self.token`.  It cannot because T1 is holding a read lock... 

I think you have a deadlock.

dbolduc: The read lock in T1 is released when we return from `state()`. No?

coryan: `T1` is in the body of `state()` it has not returned.  It is trying to acquire the `refresh_lock`...

dbolduc: > It is trying to acquire the refresh_lock...

key word: trying.

Note that we call `try_lock_owned()`, not `lock()`

https://docs.rs/tokio/latest/tokio/sync/struct.Mutex.html#method.try_lock_owned

> `T1` is in the body of `state()`

If `T2` is holding the refresh lock is held, we return a `TryLockError` and exit `self.state()`

coryan (commented):
I think this works, but it is too subtle for my taste (and far from "fearless concurrency").  A higher-level abstraction, such as:

https://docs.rs/tokio/latest/tokio/sync/index.html#watch-channel

Seems easier to reason about.  You could do something like "borrow the current value, but if it is expired start a background task and wait for the next value".

fiadliel:
There's an alternative approach here that you might consider - fundamentally it uses a `RwLock<Option<Arc<Notify>>>`. If the value is `None` and the token is out of date, readers race to set the value; losing readers wait to be notified, while the winner gets a new token and does the notification.

Some code:

The structure (here I've split out the `Arc` into a parent structure):
```rust
struct TokenCacheInner<T>
where
    T: TokenProvider,
{
    // The cached token, or the last seen error.
    token_lock: RwLock<Option<Result<Token>>>,

    // Data used to coordinate token refreshes.
    refresh_data_lock: RwLock<Option<Arc<RefreshData>>>,

    // The token provider. This thing does the refreshing.
    inner: T,
}
```

A structure representing whether this particular task will update or wait:
```rust
#[derive(Debug)]
enum FetchState {
    WaitForUpdate(Arc<RefreshData>),
    FetchAndNotifyWaiters(Arc<RefreshData>),
}
```

A function that decides whether to update or wait:
```rust
fn maybe_start_refresh(refresh_data_lock: &RwLock<Option<Arc<RefreshData>>>) -> FetchState {
    if let Some(refresh_data) = refresh_data_lock.read().as_ref() {
        // A refresh is already in progress.
        return FetchState::WaitForUpdate(refresh_data.clone());
    }

    let mut refresh_data_rw = refresh_data_lock.write();

    match refresh_data_rw.as_ref() {
        // Another task started a refresh first.
        Some(refresh_data) => FetchState::WaitForUpdate(refresh_data.clone()),
        // Still no refresh in progress, start one.
        None => {
            let new_refresh_data = Arc::new(RefreshData::new());
            refresh_data_rw.replace(new_refresh_data.clone());
            FetchState::FetchAndNotifyWaiters(new_refresh_data)
        }
    }
}
```

A structure representing refresh data. Currently just a `Notify` object, but could contain other data, e.g. refresh start time, for timeouts/etc.
```rust
#[derive(Debug)]
struct RefreshData {
    // The notify object to wake up waiters.
    notify: Notify,
}
```

And the `get_token` call combines this:

```rust
async fn get_token(&self) -> Result<Token> {
    if let Some(Ok(cur_token)) = self.token_lock.read().as_ref() {
        if cur_token
            .expires_at
            .is_none_or(|expires_at| expires_at > Instant::now().into_std())
        {
            return Ok(cur_token.clone());
        }
    }

    match maybe_start_refresh(&self.refresh_data_lock) {
        FetchState::WaitForUpdate(refresh_data) => {
            refresh_data.notify.notified().await;
            self.token_lock
                .read()
                .as_ref()
                .cloned()
                .expect("Token should be present after refresh")
        }

        FetchState::FetchAndNotifyWaiters(refresh_data) => {
            let new_token = self.inner.get_token().await;
            self.token_lock.write().replace(new_token.clone());
            self.refresh_data_lock.write().take();
            refresh_data.notify.notify_waiters();
            new_token
        }
    }
}
```

Note: this code uses the `RwLock` from `parking_lot`.

dbolduc:
@fiadliel - I like your suggestion. Thanks. I think we will likely do something similar to this.

We decided we want the token cache to have an internal retry loop. (We cannot just leave retrying up to the clients. Clients for different services could share a `Credential`, but not retry loops).

I like how `RefreshData` will let us hold information like a deadline, in case we need to cut off operations.

I think I want to keep my `TokenState` struct, conceptually, but modify it to incorporate your `FetchState`, `maybe_start_refresh(...)`. It would look like:

```rs
enum TokenState {
    PermanentError(CredentialError),
    Invalid(FetchState),
    Stale(Token, FetchState),
    Valid(Token),
}
```

I think we want to hold the token read lock while we check the refresh lock. Otherwise, we might have a refresh that completes in between releasing the token read lock and seeing if there are active refreshes. Basically, [this problem](https://github.com/googleapis/google-cloud-rust/pull/1355#discussion_r1962309900), which coryan noticed.

The `Stale` case I was saving for another PR, but it is basically:
- return a `Token` if we are in the `WaitForUpdate` state
- `tokio::spawn` a background refresh if we are in the `FetchAndNotify` state.

Thanks for collaborating on this! Your expertise is always appreciated.

fiadliel:
> I think we want to hold the token read lock while we check the refresh lock. Otherwise, we might have a refresh that completes in between releasing the token read lock and seeing if there are active refreshes. Basically, [this problem](https://github.com/googleapis/google-cloud-rust/pull/1355#discussion_r1962309900), which coryan noticed.

The race condition looks real, though the window where this might happen should be small (microseconds?). Multiple locks could fix it alright, just be careful about possible deadlocks, especially where the write lock is obtained.

> The Stale case I was saving for another PR

I like this. The only thought I have is to make sure the code acts correctly if the background refresh is still ongoing when the token becomes invalid (it could be delayed indefinitely in the case of a suspended VM).

dbolduc:
Update: I think we are going to go with a simpler design where we spawn a background task (I assume per token) that wakes up just before the token expires, and performs a refresh.

We may consider killing the background task if the token has not been asked for in the last hour, or something.

So I am going to close this PR.

fiadliel:
> I think we are going to go with a simpler design where we spawn a background task (I assume per token) that wakes up just before the token expires, and performs a refresh.

@dbolduc Remember the throttled Cloud Run scenario where you are not guaranteed CPU unless there's an active request in progress. So background threads are not guaranteed to have progressed. That's my only worry; otherwise whatever design works for you!

Issue #1077, opened 2025-02-09
Support multiple rustls crypto providers


Dig-Doug:
**Is your feature request related to a problem? Please describe.**
No.

**Describe the solution you'd like**
`rustls` provides multiple [crypto providers](https://github.com/rustls/rustls/tree/main?tab=readme-ov-file#built-in-providers). However, it looks like you're directly using one [here](https://github.com/googleapis/google-cloud-rust/blob/9feaaa2af160258c4268398fbf2d24681eace4d0/src/auth/src/credentials/service_account_credential.rs#L129).

**Describe alternatives you've considered**
Is it possible to drop the direct usage of `aws-lc-rs` so it can be configured by the caller (and not included in the dependency tree)?

coryan:
This reminded me, we also need something in the user guide: https://github.com/googleapis/google-cloud-rust/issues/1078

dbolduc:
We are also going to need to add a `default-crypto-provider` feature to `gax` and each of the N clients. The feature will be in their default feature set.

This is unfortunate, but it is the only way to both
1. have a default crypto provider in a client.
2. have the ability to disable the default crypto provider in a client.

dbolduc:
@Dig-Doug - we interpreted this request as: "do not force me to pull in `aws-lc-rs`".

We changed the default crypto provider to `ring`. We would rather have the mandatory dependency on `ring` than maintain a `default-crypto-provider` feature in each of our client crates (there are hundreds).

If this does not work for you, please let us know why.

Dig-Doug:
Regarding:

> We would rather have the mandatory dependency on ring than maintain a default-crypto-provider feature in each of our client crates (there are hundreds).

Maybe you could avoid that issue by:
- Add a `runtime-interfaces` crate that defines interfaces and provides a way to register an implementation
- Add a `runtime` crate that implements the interfaces and registers them. Users would have to depend on the crate explicitly and could also configure it
- All the service clients would depend on `runtime-interfaces`, so there's no direct dependency to `rustls` to configure

houseme:
It is suggested to configure aws-lc-rs as an optional dependency so that users can choose whether to include the library as needed, thereby enhancing the flexibility and configurability of the project.

PR #1359, opened 2025-02-21
fix(auth): feature gate aws_lc_rs from rustls

Sh4d1:
Part of the work for #1077 

It avoids bringing in the entire aws_lc_rs crate when ring is used for the crypto provider.

Maybe aws_lc_rs needs to be disabled by default, but this why it won't be breaking change.

### DIFF
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -25,8 +25,7 @@ keywords.workspace   = true
 categories.workspace = true
 
 [features]
-default = ["default-crypto-provider"]
-
+default                 = ["default-crypto-provider"]
 default-crypto-provider = ["rustls/aws_lc_rs"]
 
 [dependencies]
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -25,11 +25,9 @@ keywords.workspace   = true
 categories.workspace = true
 
 [features]
-default = ["aws_lc_rs"]
+default = ["default-crypto-provider"]
 
-aws-lc-rs = ["aws_lc_rs"]
-aws_lc_rs = ["rustls/aws_lc_rs"]
-ring      = []
+default-crypto-provider = ["rustls/aws_lc_rs"]
 
 [dependencies]
 async-trait    = "0.1"
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -24,6 +24,13 @@ repository.workspace = true
 keywords.workspace   = true
 categories.workspace = true
 
+[features]
+default = ["aws_lc_rs"]
+
+aws-lc-rs = ["aws_lc_rs"]
+aws_lc_rs = ["rustls/aws_lc_rs"]
+ring      = []
+
 [dependencies]
 async-trait    = "0.1"
 http           = "1"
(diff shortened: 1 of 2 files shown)


src/auth/Cargo.toml
```diff
@@ -24,6 +24,13 @@ repository.workspace = true
 keywords.workspace   = true
 categories.workspace = true
 
+[features]
+default = ["aws_lc_rs"]
+
+aws-lc-rs = ["aws_lc_rs"]
+aws_lc_rs = ["rustls/aws_lc_rs"]
+ring      = []
```
dbolduc, line 32: What about having a default, which we do not necessarily name as `aws_lc_rs`. And if the application cares to have something other than our default, they can just disable that feature, and provide their own.

I don't want to have N features for N different crypto providers.

```suggestion
default = ["default-crypto-provider"]

default-crypto-provider = ["rustls/aws_lc_rs"]
```

src/auth/src/credentials/service_account_credential.rs
```diff
@@ -129,6 +129,12 @@ impl TokenProvider for ServiceAccountTokenProvider {
 impl ServiceAccountTokenProvider {
     // Creates a signer using the private key stored in the service account file.
     fn signer(&self, private_key: &String) -> Result<Box<dyn Signer>> {
+        #[cfg(not(feature = "aws_lc_rs"))]
+        let key_provider = CryptoProvider::get_default()
+                    .map(|p| p.key_provider)
+                    .ok_or_else(|| CredentialError::non_retryable_from_str("No default crypto provider found. Please enable one of the following features: `aws_lc_rs`"))?;
+
+        #[cfg(feature = "aws_lc_rs")]
```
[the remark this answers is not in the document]
dbolduc: FYI: My colleague will yell at me for having a long error message. He would ask me to factor it out into a function like:

```rs
fn no_crypto_provider_error() -> CredentialError {
  CredentialError::non_retryable_from_str("No default crypto provider found. Please (1) install a global CryptoProvider, or (2) enable the feature `default-crypto-provider` to have the library pick one for you.")
}

// etc...
        let key_provider = CryptoProvider::get_default()
                    .map(|p| p.key_provider)
                    .ok_or_else(no_crypto_provider_error)?;
```

dbolduc (changes requested):
Thanks for the PR!

I am unsure about the exact features. What do you think of my suggestion?

## On unit testing...

We also have several tests that depend on there being a default crypto provider.

- `get_service_account_token_invalid_key_failure`
- `get_service_account_token_pkcs1_key_failure`
- `get_service_account_token_pkcs8_key_success`
- `signer_failure`
- there may be more... I think the auth integration test would also fail in this configuration.

I think we should feature gate these tests. I think we also want a test that verifies we return an error if the feature is not enabled.

I would accept this PR without these feature gates, though. I would be happy to add the gates and a new unit test, and make sure they are exercised somehow in the CI.

Let me know how you want to proceed.

Sh4d1:
Hey sorry, been pretty busy and clearly this got out of my head! I guess the quicker for you is to push your commits here, I can address the comments but that would be more towards the end of the week!

src/auth/Cargo.toml
```diff
@@ -24,6 +24,11 @@ repository.workspace = true
 keywords.workspace   = true
 categories.workspace = true
 
+[features]
+default = ["default-crypto-provider"]
+
```
coryan, line 29: nit: why the extra blank line?

coryan (approved):

dbolduc (approved):

PR #1376, opened 2025-02-25
test(auth): api key integration  test

### LINKED ISSUE #1371: Support API Key credentials
https://cloud.google.com/api-keys/docs/overview

- [x] Add API Key implementation
- [ ] Add API Key integration test
- [ ] Add terraform for managing API Key integration test resources

dbolduc:
Fixes #1371 

Add an integration test for API  Keys.

The test uses a long lived API key, stored in secret manager. These keys can take on the order of minutes to be ready after being created by the `apikeys` service. So let's avoid that latency in our tests.

https://github.com/googleapis/google-cloud-cpp/blob/1f054ae244e473fe3911419ddec3358e4dbeeccf/examples/api_key.cc#L141-L160

### DIFF
--- a/src/auth/integration-tests/tests/driver.rs
+++ b/src/auth/integration-tests/tests/driver.rs
@@ -14,9 +14,13 @@
 
 #[cfg(all(test, feature = "run-integration-tests"))]
 mod driver {
-    use auth::credentials::create_access_token_credential;
+    use auth::credentials::{
+        create_access_token_credential, create_api_key_credential, ApiKeyOptions,
+    };
     use gax::error::Error;
     use gax::options::ClientConfig as Config;
+    use language::client::LanguageService;
+    use language::model::Document;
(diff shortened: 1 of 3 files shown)


src/auth/integration-tests/tests/driver.rs
```diff
@@ -76,4 +80,44 @@ mod driver {
...
+            .access_secret_version(format!(
+                "projects/{}/secrets/test-api-key/versions/latest",
+                project
+            ))
+            .send()
+            .await?;
+        let api_key = response
+            .payload
+            .expect("missing payload in test-api-key response")
+            .data;
```
coryan, line 103: FWIW, you can expose secrets to GCB:

https://cloud.google.com/build/docs/securing-builds/use-secrets

sai-sunder-s (approved):

PR #1388, opened 2025-02-27
feat(wkt): implement  `NullValue`

### LINKED ISSUE #37: handle well-known proto types

coryan:
The `google.protobuf.NullValue` well-known type is represented as a
unit-like struct. It only has one value, and it is a distinct type.

It can be stored inside a `wkt::Any` and freely converts to and from
`serde_json::Value::NullValue`. Recall that `wkt::Value` is a type alias
for `serde_json::Value`, meaning it works well with the value type.

The type also works as a member of oneof fields, their only use (AFAICT)
in Google Cloud services.

Fixes #37

### DIFF
--- a/src/wkt/src/rstruct.rs
+++ b/src/wkt/src/rstruct.rs
@@ -12,22 +12,31 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Implement the well-known types for structured, yet dynamically typed
+//! messages. These types are a representation of JSON objects, lists, and
+//! values as Protobuf messages. We have taken some (allowed) liberty in their
+//! representation for Rust. We map them directly to the [serde_json] types,
+//! except for `NullValue` where there is no corresponding type in serde.
+//! 
+//! Services specified using Protobuf files may use `google.protobuf.Struct`,
+//! `google.protobuf.Value`, `google.protobuf.ListValue`, and/or 
+//! `google.protobuf.NullValue` as part of their interface specification.
+
 /// Protobuf (and consequently the Google Cloud APIs) use `Struct` to represent
-/// JSON objects. We need a type that might be referenced from the generated
-/// code.
+/// JSON objects. We need a type that can be referenced from the generated code.
 pub type Struct = serde_json::Map<String, serde_json::Value>;
 
 /// Protobuf (and consequently the Google Cloud APIs) use `Value` to represent
-/// JSON values. We need a type that might be referenced from the generated
-/// code.
+/// JSON values. We need a type that can be referenced from the generated code.
 pub type Value = serde_json::Value;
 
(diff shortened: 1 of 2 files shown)


coryan:
The one missed line is a false-negative in the coverage analysis.

src/wkt/src/rstruct.rs
```diff
@@ -110,14 +114,94 @@ impl crate::message::Message for ListValue {
     }
 }
 
+impl crate::message::Message for NullValue {
+    fn typename() -> &'static str {
+        "type.googleapis.com/google.protobuf.Value"
```
dbolduc, line 119: I think I am missing the nuance here on why `ListValue` is `google.protobuf.ListValue` but things like `BoolValue` / `NullValue` are just `google.protobuf.Value`.

I am also not seeing why we need the special `NullValue` when we do not have a special `BoolValue`.

src/wkt/tests/null_value.rs
```diff
@@ -0,0 +1,149 @@
...
+        });
+        let got = serde_json::from_value::<Value>(input)?;
+        let want = Value::new()
+            .set_test_name("test_serialize")
+            .set_number(123.5);
+        assert_eq!(got, want);
+        Ok(())
+    }
+
+    // A test message, inspired by `google.firestore.v1.ValueType`.
```
dbolduc, line 94: Ah, thanks. I was confused as to why we were checking the duration / number cases.

although nit: `s/ValueType/Value/`

dbolduc (approved):
Left a comment, coming from a place of ignorance.

src/wkt/src/rstruct.rs
dbolduc: So why isn't the typename s/Value/NullValue/ on this line?

Issue #1350, opened 2025-02-18
Add unit tests for auth failures in gax layer


dbolduc:
We only test the success cases. We should make sure auth errors are propagated to the caller.

nakayamaryo0731:
Hi! I'd like to contribute to this repository and I'm interested in tackling this issue. If no one else is currently working on it, I'd be happy to get started. Please let me know if there's anything specific I should keep in mind or any guidance you can provide. Thank you!

dbolduc:
Hey @nakayamaryo0731 - this task is all yours.

## Setup

https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-set-up-development-environment.md

## Background

We have one auth test in `google-cloud-gax`:

https://github.com/googleapis/google-cloud-rust/blob/b1ba21bbd001bc20e02d3565a7e0dec572bfddbc/src/gax/tests/auth.rs#L39

We use mock credentials, which return `Ok` credentials:

https://github.com/googleapis/google-cloud-rust/blob/b1ba21bbd001bc20e02d3565a7e0dec572bfddbc/src/gax/tests/auth.rs#L47

## Task

We want to add two tests that return a `CredentialError`:
1. `auth_error_retryable() { ... }`
  - this one should have >1 calls to `get_headers()`
1. `auth_error_non_retryable() { ... }`
  - this one should have exactly 1 call to `get_headers()`

> anything specific I should keep in mind

It looks like all auth errors are currently retried, by default. This is not what we want. We need more intricate logic here:

https://github.com/googleapis/google-cloud-rust/blob/6ced131d63a7395b5307a2fbbc55add46900e47e/src/gax/src/retry_policy.rs#L236-L240

You can run only the auth tests in `google-cloud-gax` with:

```
cargo test -p google-cloud-gax --test auth
```

PR #1394, opened 2025-02-27
feat(auth): enhance retry logic for authentication errors

nakayamaryo0731:
## Description
Issue: https://github.com/googleapis/google-cloud-rust/issues/1350

I implemented the following two unit tests for CredentialError:

auth_error_retryable() { ... }
auth_error_non_retryable() { ... }
I tried to extend the existing retry_loop to implement the authentication loop process, but I couldn't get it to work properly. So, I implemented auth_retry_loop instead.

If you think it should be unified, please suggest an implementation approach.

## Test
`cargo test -p google-cloud-gax --test auth`

### DIFF
--- a/src/auth/src/errors.rs
+++ b/src/auth/src/errors.rs
@@ -63,7 +63,7 @@ impl CredentialError {
     /// # Arguments
     /// * `is_retryable` - A boolean indicating whether the error is retryable.
     /// * `message` - The underlying error that caused the auth failure.
-    pub(crate) fn from_str<T: Into<String>>(is_retryable: bool, message: T) -> Self {
+    pub fn from_str<T: Into<String>>(is_retryable: bool, message: T) -> Self {
         CredentialError::new(
             is_retryable,
             CredentialErrorImpl::SimpleMessage(message.into()),
--- a/src/gax/src/http_client/mod.rs
+++ b/src/gax/src/http_client/mod.rs
@@ -26,6 +26,8 @@ use crate::retry_policy::RetryPolicy;
 use crate::retry_throttler::RetryThrottlerWrapped;
 use crate::Result;
 use auth::credentials::{create_access_token_credential, Credential};
+use http::HeaderName;
+use http::HeaderValue;
 use std::sync::Arc;
 
 #[derive(Clone, Debug)]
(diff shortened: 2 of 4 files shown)


dbolduc (commented):
/gcbrun

src/auth/src/errors.rs
```diff
@@ -63,7 +63,7 @@ impl CredentialError {
     /// # Arguments
     /// * `is_retryable` - A boolean indicating whether the error is retryable.
     /// * `message` - The underlying error that caused the auth failure.
-    pub(crate) fn from_str<T: Into<String>>(is_retryable: bool, message: T) -> Self {
+    pub fn from_str<T: Into<String>>(is_retryable: bool, message: T) -> Self {
```
dbolduc, line 66: comment: As I suspected, this was a necessary change to use `CredentialError` outside of the `google-cloud-auth` crate. Thanks.

dbolduc (changes requested):
The tests look great! Thanks!

> If you think it should be unified, please suggest an implementation approach.

I do think we need to unify the loops. Let's move the call to get the auth headers inside `request_attempt()`. I think that should achieve what we want and avoid duplication.

https://github.com/googleapis/google-cloud-rust/blame/5bf2f639130cad781a7402acfcc9ba691b01f6a5/src/gax/src/http_client/mod.rs#L177

Something like:

```diff
@@ -258,6 +242,14 @@ impl ReqwestClient {
         {
             builder = builder.timeout(timeout);
         }
+        let auth_headers = self
+            .cred
+            .get_headers()
+            .await
+            .map_err(Error::authentication)?;
+        for header in auth_headers.into_iter() {
+            builder = builder.header(header.0, header.1);
+        }
         let response = builder.send().await.map_err(Error::io)?;
         if !response.status().is_success() {
             return Self::to_http_error(response).await;
```

And we might want to run all of the tests for gax, to make sure we are not breaking existing retry loop tests:

```
cargo test -p google-cloud-gax
```

nakayamaryo0731:
> The tests look great! Thanks!
> 
> > If you think it should be unified, please suggest an implementation approach.
> 
> I do think we need to unify the loops. Let's move the call to get the auth headers inside `request_attempt()`. I think that should achieve what we want and avoid duplication.
> 
> https://github.com/googleapis/google-cloud-rust/blob/5bf2f639130cad781a7402acfcc9ba691b01f6a5/src/gax/src/http_client/mod.rs#L177
> 
> Something like:
> 
> ```diff
> @@ -258,6 +242,14 @@ impl ReqwestClient {
>          {
>              builder = builder.timeout(timeout);
>          }
> +        let auth_headers = self
> +            .cred
> +            .get_headers()
> +            .await
> +            .map_err(Error::authentication)?;
> +        for header in auth_headers.into_iter() {
> +            builder = builder.header(header.0, header.1);
> +        }
>          let response = builder.send().await.map_err(Error::io)?;
>          if !response.status().is_success() {
>              return Self::to_http_error(response).await;
> ```
> 
> And we might want to run all of the tests for gax, to make sure we are not breaking existing retry loop tests:
> 
> ```
> cargo test -p google-cloud-gax
> ```

Thank you for the review!
I've applied your suggestions in the following commit:
[f08f72f](https://github.com/googleapis/google-cloud-rust/pull/1394/commits/f08f72f7f584c3ad8868a34609f80041381df296)

I also confirmed that the GAX tests pass by running:
```
cargo test -p google-cloud-gax
```

dbolduc (approved):
Looks great, thanks!

coryan (approved):

PR #1400, opened 2025-02-28
fix(wkt): no trailing zeroes in duration as JSON

### LINKED ISSUE #1397: Trailing zeroes in `wkt::Duration` JSON encoding
Here, and a few cases below, should we be instead expecting no trailing zeros? `12.000123000s` => `12.000123s`. I don't think it's incorrect to have trailing `0`s but suspect you'd want to encode as `12.1s` instead of `12.100000000s`.

_Originally posted by @devoncarew in https://github.com/googleapis/google-cloud-rust/pull/49#discussion_r1975826686_

coryan:
Fixes #1397

### DIFF
--- a/src/wkt/src/duration.rs
+++ b/src/wkt/src/duration.rs
@@ -187,14 +187,14 @@ impl std::convert::From<&Duration> for String {
         if duration.nanos == 0 {
             return format!("{sign}{}s", duration.seconds.abs());
         }
+        let ns = format!("{:09}", duration.nanos.abs());
         if duration.seconds == 0 {
-            let ns = format!("{:09}", duration.nanos.abs());
             return format!("{sign}0.{}s", ns.trim_end_matches('0'));
         }
         format!(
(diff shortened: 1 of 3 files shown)


src/wkt/src/duration.rs
```diff
@@ -187,14 +187,14 @@ impl std::convert::From<&Duration> for String {
         if duration.nanos == 0 {
             return format!("{sign}{}s", duration.seconds.abs());
         }
+        let ns = format!("{:09}", duration.nanos.abs());
         if duration.seconds == 0 {
-            let ns = format!("{:09}", duration.nanos.abs());
             return format!("{sign}0.{}s", ns.trim_end_matches('0'));
         }
```
dbolduc, line 193: looks like we can dispense with this `if`

dbolduc (approved):

src/wkt/src/duration.rs
coryan: Done.

Issue #1414, opened 2025-03-05
Implement a codec to generate code using Prost and Tonic


coryan:
To generate veneers and libraries that must use gRPC as their transport we will implement a new codec that:

1. Invokes the rust codec to generate any types or client code we need.
2. Generates a minimal `build.rs` file (and matching `Cargo.toml` file)
3. Invokes the Tonic + Prost generators using this minimal `build.rs` file.

coryan:
I am going to declare this done.  The generator can call Prost+Tonic to generate code, and can keep this code refreshed.

PR #1418, opened 2025-03-06
feat(generator): generate code using Tonic+Prost

coryan:
Sometimes we need to use Tonic and Prost to generate gRPC and Protobuf
code. This change introduces a new codec that calls the right code
generators for this case. For now one needs to manually integrate the
generated code into a larger crate.

Part of the work for #1414


generator/internal/rust_prost/rust_prost.go
```diff
@@ -0,0 +1,129 @@
...
+func buildRS(googleapisRoot, outdir string) error {
+	absolute, err := filepath.Abs(googleapisRoot)
+	if err != nil {
+		return err
+	}
+	cmd := exec.Command("cargo", "build", "--features", "_generate-protos")
+	cmd.Dir = outdir
+	cmd.Env = append(os.Environ(), fmt.Sprintf("GOOGLEAPIS_ROOT=%s", absolute))
+	return runAndCaptureErrors(cmd)
+
```
dbolduc, line 112: nit: weird newline

generator/internal/rust_prost/rust_prost.go
```diff
@@ -0,0 +1,129 @@
...
+type root struct {
+	CopyrightYear string
+	BoilerPlate   []string
+	PackageName   string
+	Files         []string
+}
+
+func Generate(model *api.API, outdir string, cfg *config.Config) error {
+	if cfg.General.SpecificationFormat != "protobuf" {
+		return fmt.Errorf("the `rust+prost` generator only supports `protobuf` as an specification source, outdir=%s", outdir)
```
dbolduc, line 47: nit: s/an specification/a specification/

generator/internal/rust_prost/templates/prost/Cargo.toml.mustache
```diff
@@ -0,0 +1,37 @@
...
+publish     = false
+build       = "build.rs"
+edition     = "2021"
+
+[build-dependencies]
+prost-build = { version = "0.13", optional = true }
+tonic-build = { version = "0.12", optional = true }
+
+[features]
+_generate-protos = ["dep:prost-build", "dep:tonic-build"]
```
dbolduc, line 34: just curious, why the leading underscore?

generator/testdata/rust/prost/golden/type/google.r#type.rs
```diff
@@ -0,0 +1,22 @@
+// This file is @generated by prost-build.
```
dbolduc, line 1: Do we have any control over whether this file gets a license?

dbolduc (approved):
Just curious, will we be checking in prost generated code to the repo?

generator/internal/rust_prost/rust_prost.go
coryan: Thanks, fixed.

generator/internal/rust_prost/templates/prost/Cargo.toml.mustache
coryan: https://github.com/rust-lang/cargo/issues/10882#issuecomment-1880844632

Features with leading underscores are hidden from the documentation. Seems like a neat idea to adopt for features we do not want to advertise. Either way would be mostly harmless in these crates, as they will never be published.

generator/testdata/rust/prost/golden/type/google.r#type.rs
coryan: I could not find any. I am not sure we are supposed to put a license in these files either.  We could add the license after the fact: #1421

coryan:
> Just curious, will we be checking in prost generated code to the repo?

Yes, for now. I do not think we can assume our customers will have `protoc` installed.  And (unlike in C++) Prost does not have super strict restrictions on generated code vs. runtime version compatibilities.

PR #1430, opened 2025-03-07
fix(auth): Read token_uri from user json

sai-sunder-s:
If token_uri is provided in the json, we should use that instead of a hardcoded value.

### DIFF
--- a/src/auth/src/credentials/user_credential.rs
+++ b/src/auth/src/credentials/user_credential.rs
@@ -27,11 +27,14 @@ const OAUTH2_ENDPOINT: &str = "https://oauth2.googleapis.com/token";
 pub(crate) fn creds_from(js: serde_json::Value) -> Result<Credential> {
     let au =
         serde_json::from_value::<AuthorizedUser>(js).map_err(CredentialError::non_retryable)?;
+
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/user_credential.rs
```diff
@@ -27,11 +27,14 @@ const OAUTH2_ENDPOINT: &str = "https://oauth2.googleapis.com/token";
 pub(crate) fn creds_from(js: serde_json::Value) -> Result<Credential> {
     let au =
         serde_json::from_value::<AuthorizedUser>(js).map_err(CredentialError::non_retryable)?;
+
+    let endpoint = au.token_uri.unwrap_or_else(|| OAUTH2_ENDPOINT.to_string());
+
```
coryan, line 32: I do not see any new tests exercising the new functionality. Did I miss them?

coryan (commented):
I would like to see a new test for the new functionality. Is it possible to write one?

src/auth/src/credentials/user_credential.rs
sai-sunder-s: Added

coryan (approved):

PR #1431, opened 2025-03-08
cleanup: faster local development

coryan:
Skip local development by skipping most of the generated crates. The CI
builds will still compile all the code, we just won't do it in our
local development environments.


dbolduc (approved):
Nice, thanks.

Issue #1437, opened 2025-03-10
Add options to MDSCredentials


harkamaljot:
In rust auth library we want MDSCredential to have the following options
1. service account email
2. scopes
3. quota project id
4. universe domain

When fetching a token service account email and scopes is used. 
While quota project id is used, when fetching the headers and universe domain if absent is queried from mds server otherwise is used whatever is supplied by the user.

Therefore this work can be split into 2 PR's
- [x] update MDSAccessTokenProvider to use service account email and scopes
- [ ] modify MDSCredential to support all the parameters, so that user can create a custom credential

harkamaljot:
This seems to be done, closing it out.

PR #1438, opened 2025-03-10
impl(auth): support non-default MDS credentials

harkamaljot:
Part of the work for https://github.com/googleapis/google-cloud-rust/issues/1437

Add an implementation for fetching token from mds server using scopes and service account email.

Unit tests

Integration tests?
There are none. I did manually tested the cloud on my gce vm and verified I was able to fetch back token


coryan (commented):
LGTM. @sai-sunder-s probably should sign off.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -365,10 +360,63 @@ mod test {
...
-        let (endpoint, _server) = start(StatusCode::OK, response_body, path.to_string()).await;
+
+        let (endpoint, _server) = start(HashMap::from([(
+            path.to_string(),
+            (StatusCode::OK, response_body),
+        )]))
+        .await;
         println!("endpoint = {endpoint}");
+        let tp = MDSAccessTokenProviderBuilder::default()
+            .scopes(vec!["scope1".to_string()])
```
sai-sunder-s, line 371: how do we test whether the scopes get passed to mds correctly?

src/auth/src/credentials/mds_credential.rs
```diff
@@ -390,10 +438,19 @@ mod test {
...
+        let (endpoint, _server) = start(HashMap::from([(
+            path.to_string(),
+            (StatusCode::OK, response_body),
+        )]))
+        .await;
         println!("endpoint = {endpoint}");
 
-        let tp = MDSAccessTokenProvider { endpoint: endpoint };
+        let tp = MDSAccessTokenProviderBuilder::default()
+            .scopes(vec!["scope1".to_string()])
```
sai-sunder-s, line 449: same here. how do we test if scopes are passed along

src/auth/src/credentials/mds_credential.rs
```diff
@@ -17,10 +17,9 @@ use crate::credentials::{Credential, Result};
 use crate::errors::{is_retryable, CredentialError};
 use crate::token::{Token, TokenProvider};
 use async_trait::async_trait;
+use derive_builder::Builder;
```
sai-sunder-s, line 20: this is not a stable version. Since its a public API, we should use some other crate with a stable version or handwrite the builder.

sai-sunder-s: Since you are not adding a dependency in this PR. I am ok if you want to create an issue and work on removing this dependency in a future PR.

src/auth/src/credentials/mds_credential.rs
harkamaljot: I updated the server code to include a assert on the parameters.

sai-sunder-s (approved):

PR #1448, opened 2025-03-10
feat(generator/rust)!: use integers for enums

### LINKED ISSUE #1379: Support enums encoded as numbers
## TL;DR;

We will change the enum implementation to use `i32` as its internal representation, and request integers from the service using the `$alt=json;enum-encoding=int` [system parameter](https://cloud.google.com/apis/docs/system-parameters)

## Background

We use HTTP+JSON as the transport for all client libraries that support it.

Some services (most veneers, and even some GAPICs) will require gRPC to support bidi streaming RPCs.

Adding an RPC should not be a breaking change.

Protobuf enums (like `enum { Red, Green, Blue }`) are "open". Adding a new enum value should not be a breaking change or require client-library upgrades to receive them.  Ideally applications can send new enum values, unknown when the client library was generated. They just need to read the right documentation to find out what these values are.

By default, over HTTP+JSON, enums are serialized as strings.  Unknown enum values (e.g. `Cyan`) are typically received as strings by the client library. And the client library receives **only** the string. The client library *can* send unknown integer values over HTTP+JSON. The service could send unknown enum values as integers, but this is unlikely to happen in practice.

The client can request that enums are serialized as integers over HTTP+JSON.

Over gRPC, enums are always serialized as integers. Unknown enum values are received as integers by the client library. And the client library receives **only** the integer.

Consider this scenario:

1. We have launched a client library for ServiceX.  
2. The library uses HTTP+JSON because there are no bidirectional streaming RPCs in the interface.  
3. Consequently, the representation of enums **must** be strings, otherwise we cannot handle unknown enum values sent by the service.  
4. We declare the library stable (1.0).  
5. The service then adds a streaming RPC.  
6. That requires switching the client library to start using gRPC as the transport.  
7. Now we have a problem:  
   1. If we use strings to represent enum values, then we cannot represent unknown integers received over gRPC.  
   2. If we use integers to represent enum values, then application code breaks because they may be using unknown string values already.

## Switching the transport is a breaking change

We can decide to change the enum representation based on the underlying transport, and bump the major version of the ServiceX library when switching to gRPC as the transport.  
**Cons:** 

* This is really a breaking change for customer code. The enum API changes so their code needs to change.  
* In the context of OSS libraries Google's [policy](https://opensource.google/documentation/policies/library-breaking-change) requires one year of support, meaning for one year we will fix bugs [if asked](https://screenshot.googleplex.com/AgFxq9EjF2ZKKJA).

**Pros:** 

* It is relatively easy to explain.

## Use only gRPC for streaming RPCs and "smart" enum values

We could use an enum representation that supports both unknown integers and unknown strings. We could make the requests for non-streaming RPCs using HTTP+JSON and streaming RPCs over gRPC.  
**Cons:**

* Trying to serialize an unknown value of the wrong type (i.e. an unknown string over gRPC) will need to return an error or panic.  
  * Since only *new* code will use the new streaming RPC, that is not likely to break existing code  
* Unknown values received from HTTP+JSON cannot be sent back over gRPC. That may break code unexpectedly.

**Pros:**

* No code breaking changes when we switch the default transport.

## Use the same transport for all RPCs with "smart" enum values

We could use an enum representation that supports both unknown integers and unknown strings and switch *all* RPCs to gRPC when the library gains a value.  
**Pros:**

* Unknown values received from the wire work well: they will always be sent back over the same transport  
* In practice, unknown values provided by the application are extremely rare.

**Cons:**

* Trying to serialize an unknown value of the wrong type (i.e. an unknown string over gRPC) will need to return an error or panic.

coryan:
Change the enums to be integer-based as opposed to string-based. This
changes the API, so it is a breaking change. We need to make this change
to support client libraries based on gRPC. Unknown enums are received
as integers in that case. We need to make the change for all client
libraries, as any of them may need to switch the underlying transport if
they gain streaming RPCs.

Fixes #1379

### DIFF
--- a/generator/internal/rust/rust.go
+++ b/generator/internal/rust/rust.go
@@ -58,7 +58,7 @@ var commentUrlRegex = regexp.MustCompile(
 		`(/[-a-zA-Z0-9@:%_\+.~#?&/={}\$]*)?`) // Accept just about anything on the query and URL fragments
 
 func Generate(model *api.API, outdir string, cfg *config.Config) error {
-	codec, err := newCodec(cfg.Codec)
+	codec, err := newCodec(cfg.General.SpecificationFormat == "protobuf", cfg.Codec)
 	if err != nil {
 		return err
 	}
@@ -68,15 +68,27 @@ func Generate(model *api.API, outdir string, cfg *config.Config) error {
 	return language.GenerateFromRoot(outdir, model, provider, generatedFiles)
 }
 
-func newCodec(options map[string]string) (*codec, error) {
+func newCodec(protobufSource bool, options map[string]string) (*codec, error) {
+	var sysParams []systemParameter
+	if protobufSource {
+		sysParams = append(sysParams, systemParameter{
+			Name: "$alt", Value: "json;enum-encoding=int",
+		})
+	} else {
+		sysParams = append(sysParams, systemParameter{
+			Name: "$alt", Value: "json",
+		})
+	}
+
 	year, _, _ := time.Now().Date()
 	codec := &codec{
-		generationYear: fmt.Sprintf("%04d", year),
-		modulePath:     "crate::model",
-		extraPackages:  []*packagez{},
-		packageMapping: map[string]*packagez{},
-		version:        "0.0.0",
-		releaseLevel:   "preview",
+		generationYear:   fmt.Sprintf("%04d", year),
+		modulePath:       "crate::model",
+		extraPackages:    []*packagez{},
+		packageMapping:   map[string]*packagez{},
+		version:          "0.0.0",
+		releaseLevel:     "preview",
+		systemParameters: sysParams,
 	}
 
 	for key, definition := range options {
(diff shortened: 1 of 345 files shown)


generator/internal/rust/rust.go
```diff
@@ -68,15 +68,27 @@ func Generate(model *api.API, outdir string, cfg *config.Config) error {
 	return language.GenerateFromRoot(outdir, model, provider, generatedFiles)
 }
 
-func newCodec(options map[string]string) (*codec, error) {
+func newCodec(protobufSource bool, options map[string]string) (*codec, error) {
+	var sysParams []systemParameter
+	if protobufSource {
+		sysParams = append(sysParams, systemParameter{
+			Name: "$alt", Value: "json;enum-encoding=int",
```
dbolduc, line 75: TIL, thanks @codyoss

dbolduc (approved):
It seems correct that `openapi/.../model.rs` is not changing, but I did not find exactly why.

enums must not be included as `enum` here for `openapi`?

https://github.com/googleapis/google-cloud-rust/blob/662a03be3fe9179514834498ab6e4fe2cdb67e48/generator/internal/rust/templates/common/model.rs.mustache#L34

coryan:
> It seems correct that `openapi/.../model.rs` is not changing, but I did not find exactly why.
> 
> enums must not be included as `enum` here for `openapi`?
> 
> https://github.com/googleapis/google-cloud-rust/blob/662a03be3fe9179514834498ab6e4fe2cdb67e48/generator/internal/rust/templates/common/model.rs.mustache#L34

That is because we do not convert anything in OpenAPI to enums.  At least, not yet: #94

Issue #720, opened 2025-01-17
Implement pagination helpers without streams


coryan:
As shows by the LRO changes, we can unconditionally implement a stream-like struct without having to implement the `future::Stream` trait, and then implement the stream trait conditionally.

That seems more friendly: the easy to use API is always available, and if you want streams (with all the `StreamExt` goodies) you can turn that on.

coryan:
Can we close this? If not, what remains?

PR #1466, opened 2025-03-11
feat(gax)!: stable paginator

dbolduc:
Part of the work for #720 

Makes the `Paginator` class part of gax. It is no longer gated by the `unstable-stream` feature. The `Paginator` picks up a `to_stream()` function which is gated by the `unstable-stream` feature. (Same goes for `ItemPaginator`).

We renamed the `stream()` function to `paginator()` on the paginated builders.

## `unstable-stream` feature in client crates

All downstream crates still have an `unstable-stream` feature. I think we should get rid of them, and let applications opt in via the dependency on gax. If we want to do this, I will send another PR.

We would need to update some of our docs, because we could no longer one-line things like:
```
   cargo add google-cloud-secretmanager-v1 --features unstable-stream
```

## On testing feature configurations...

We now testing gax with and without `unstable-stream`. (By removing both `dev.dependencies` on this feature).

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -42,9 +42,10 @@ jobs:
         run: cargo version
       - name: Display rustc version
         run: rustc --version
-      - name: Test gax with features disabled
-        run: cargo test --package google-cloud-gax
-      - run: cargo test --workspace
+      - name: Test with features disabled
+        run: cargo test --workspace --no-default-features
+      - name: Test with features enabled
+        run: cargo test --workspace --all-features
   coverage:
     runs-on: ubuntu-24.04
     strategy:
@@ -99,9 +100,6 @@ jobs:
       - name: Display rustc version
         run: rustc --version
       - run: cargo doc --workspace
-        env:
-          RUSTDOCFLAGS: "-D warnings"
-      - run: cargo doc --package google-cloud-gax
         env:
           RUSTDOCFLAGS: "-D warnings"
       - run: mdbook build guide
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -43,7 +43,7 @@ jobs:
       - name: Display rustc version
         run: rustc --version
       - name: Test with features disabled
-        run: cargo test --workspace --no-default-features
+        run: cargo test --workspace -p google-cloud-gax --no-default-features
       - name: Test with features enabled
         run: cargo test --workspace --all-features
   coverage:
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -42,10 +42,9 @@ jobs:
         run: cargo version
       - name: Display rustc version
         run: rustc --version
-      - name: Test with features disabled
-        run: cargo test --workspace -p google-cloud-gax --no-default-features
-      - name: Test with features enabled
-        run: cargo test --workspace --all-features
+      - name: Test gax with features disabled
+        run: cargo test --package google-cloud-gax
+      - run: cargo test --workspace
   coverage:
     runs-on: ubuntu-24.04
     strategy:
@@ -100,6 +99,9 @@ jobs:
       - name: Display rustc version
         run: rustc --version
       - run: cargo doc --workspace
+        env:
+          RUSTDOCFLAGS: "-D warnings"
+      - run: cargo doc --package google-cloud-gax
         env:
           RUSTDOCFLAGS: "-D warnings"
       - run: mdbook build guide
--- a/generator/internal/rust/templates/crate/src/builders.rs.mustache
+++ b/generator/internal/rust/templates/crate/src/builders.rs.mustache
@@ -86,7 +86,7 @@ pub mod {{Codec.ModuleName}} {
         {{#IsPageable}}
 
         /// Streams the responses back.
-        pub async fn stream(self) -> gax::paginator::Paginator<{{OutputType.Codec.QualifiedName}}, gax::error::Error> {
+        pub async fn paginator(self) -> gax::paginator::Paginator<{{OutputType.Codec.QualifiedName}}, gax::error::Error> {
             let token = gax::paginator::extract_token(&self.0.request.page_token);
             let execute = move |token: String| {
                 let mut builder = self.clone();
--- a/generator/internal/rust/templates/crate/src/builders.rs.mustache
+++ b/generator/internal/rust/templates/crate/src/builders.rs.mustache
@@ -86,7 +86,6 @@ pub mod {{Codec.ModuleName}} {
         {{#IsPageable}}
 
         /// Streams the responses back.
-        #[cfg(feature = "unstable-stream")]
         pub async fn stream(self) -> gax::paginator::Paginator<{{OutputType.Codec.QualifiedName}}, gax::error::Error> {
             let token = gax::paginator::extract_token(&self.0.request.page_token);
             let execute = move |token: String| {
--- a/src/auth/integration-tests/Cargo.toml
+++ b/src/auth/integration-tests/Cargo.toml
@@ -24,7 +24,7 @@ run-integration-tests = []
 
 [dependencies]
 auth          = { path = "../../../src/auth", package = "google-cloud-auth" }
-gax           = { path = "../../../src/gax", package = "google-cloud-gax" }
+gax           = { path = "../../../src/gax", package = "google-cloud-gax", features = ["unstable-stream"] }
 language      = { path = "../../../src/generated/cloud/language/v2", package = "google-cloud-language-v2" }
 scoped-env    = "2"
 secretmanager = { path = "../../../src/generated/cloud/secretmanager/v1", package = "google-cloud-secretmanager-v1" }
--- a/src/gax/Cargo.toml
+++ b/src/gax/Cargo.toml
@@ -28,9 +28,9 @@ repository.workspace = true
 [dependencies]
 base64      = "0.22"
 bytes       = "1"
-futures     = { version = "0.3", optional = true }
+futures     = { version = "0.3"}
 http        = "1"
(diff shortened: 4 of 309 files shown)


src/gax/src/paginator.rs
```diff
@@ -84,13 +86,26 @@ where
         ItemPaginator::new(self)
     }
 
+    /// Returns the next mutation of the wrapped stream.
+    ///
+    /// Enable the `unstable-stream` feature to interact with a [`futures::stream::Stream`].
+    ///
+    /// [`futures::stream::Stream`]: https://docs.rs/futures/latest/futures/stream/trait.Stream.html
+    pub async fn next_stable(&mut self) -> Option<Result<T, E>> {
```
coryan, line 92: Can we arrange things so this function is called `next()`?  If not, can we call it `next_page()`?

generator/internal/rust/templates/crate/src/builders.rs.mustache
```diff
@@ -86,7 +86,6 @@ pub mod {{Codec.ModuleName}} {
         {{#IsPageable}}
 
         /// Streams the responses back.
-        #[cfg(feature = "unstable-stream")]
         pub async fn stream(self) -> gax::paginator::Paginator<{{OutputType.Codec.QualifiedName}}, gax::error::Error> {
```
coryan, line 89: It seems like this function should be `paginator()` now, since it does not return a stream?   Or maybe we should have 2 functions: `by_page()` returns the paginator with the stable API, `by_item()` returns the thing with item-by-item.

src/auth/integration-tests/Cargo.toml
```diff
@@ -24,7 +24,7 @@ run-integration-tests = []
 
 [dependencies]
 auth          = { path = "../../../src/auth", package = "google-cloud-auth" }
-gax           = { path = "../../../src/gax", package = "google-cloud-gax" }
+gax           = { path = "../../../src/gax", package = "google-cloud-gax", features = ["unstable-stream"] }
```
coryan, line 27: This seems like an unwanted change.  If you are not enabling streaming for any of the client libraries used directly, why do you need to enable streaming for the indirect dependency?

coryan (commented):
I think you need a new integration test that does **not** enable `unstable-stream` and still runs the iterations.

src/auth/integration-tests/Cargo.toml
dbolduc: I enabled the `unstable-stream` feature `gax` anticipating that we would remove the `unstable-stream` feature from the client crates.

src/gax/src/paginator.rs
dbolduc: Calling this function `next()` seems to work.

generator/internal/rust/templates/crate/src/builders.rs.mustache
dbolduc: Hmm. Changed to `paginator()`. I think it is a stream, conceptually, so the old name was fine. But eh.

> Or maybe we should have 2 functions

I considered `stream_items()` and `stream_pages()`.

I think 99% of use cases will want to iterate over the items. Returning the `Paginator` requires the xtra call to `.items()`...

But I think I prefer that there is only one way for applications to write the code. There is less to keep track of mentally.

dbolduc:
> I think you need a new integration test that does not enable unstable-stream and still runs the iterations.

I went with gating the unit test for the feature:

https://github.com/googleapis/google-cloud-rust/blob/18ab34882e9119d1322ccf1a32678e9f0924f05a/.github/workflows/sdk.yaml#L45-L48

generator/internal/rust/templates/crate/src/builders.rs.mustache
dbolduc: I don't love that we are breaking folks with the name change. :shrug:. Do you care?

coryan: I don't love making breaking changes.  I would like having the wrong name in the 1.0 release even less, and it is a 0.x release for a reason.

src/gax/Cargo.toml
```diff
@@ -28,9 +28,9 @@ repository.workspace = true
 [dependencies]
 base64      = "0.22"
 bytes       = "1"
-futures     = { version = "0.3", optional = true }
+futures     = { version = "0.3" }
```
coryan, line 31: I finally understand what you did here: you are using streams in the implementation, and hiding the stream trait implementation under a feature.

That works, but it may be possible to improve the implementation to remove this dependency.  For example, the pollers do this:

https://github.com/googleapis/google-cloud-rust/blob/ff3abcc1b5753e53e9016718fbc27f2555f32eee/src/lro/src/lib.rs#L279-L295

coryan (commented):
I think this works, but we need to decide if we want to depend on `futures` (which is at 0.3) in our implementation.

src/gax/Cargo.toml
dbolduc: I thought unstable crates in the implementation were fair game. We also have `base64` and `rand` which are < 1.0. I take it you are more scared of `futures` breaking things?

FWIW: I originally went down the other path where the dep on `futures` is feature-gated behind `unstable-stream`. It would look something like this:

https://github.com/dbolduc/google-cloud-rust/blob/5d07ad8a3e6cef2bfc48934a8cfcf4b0f95df010/src/gax/src/paginator.rs#L41-L103

But then I realized I could save myself a ton of trouble by using streams in the implementation.

src/integration-tests/src/secret_manager/openapi.rs
```diff
@@ -311,7 +311,7 @@ async fn get_all_secret_names(
     project_id: &str,
 ) -> Result<Vec<String>> {
     let mut names = Vec::new();
-    let mut stream = client.list_secrets(project_id).stream().await;
+    let mut stream = client.list_secrets(project_id).paginator().await;
```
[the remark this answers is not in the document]
dbolduc: Fixed

src/integration-tests/src/secret_manager/protobuf.rs
```diff
@@ -304,7 +304,7 @@ async fn run_many_secret_versions(
     let mut stream = client
         .list_secret_versions(secret_name)
         .set_page_size(PAGE_SIZE)
-        .stream()
+        .paginator()
```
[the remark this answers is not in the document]
dbolduc: Fixed here and elsewhere

dbolduc:
> but we need to decide if we want to depend on `futures` (which is at 0.3) in our implementation.

Ack. I did not make the dep on `futures` optional in this PR. I want to get something in first, then we can think about refactoring the implementation.

If you think it is cleaner to not use `futures` in the implementation, I can look into that.

> I went with gating the unit test for the feature and testing the crates with all features or no features:

modification: I am only building `google-cloud-gax` with `--no-default-features` to hopefully keep the build a little smaller. We might need separate builds where all or no features are enabled.

I have a meeting item to discuss the build cache anyway.

.github/workflows/sdk.yaml
```diff
@@ -101,9 +102,6 @@ jobs:
       - run: cargo doc --workspace
         env:
           RUSTDOCFLAGS: "-D warnings"
-      - run: cargo doc --package google-cloud-gax
```
coryan: As I recall, `cargo doc`  can fail to build if you have some features enabled vs. disabled (e.g. if you link an element that is not present because the feature is disabled).   I believe we should keep this test.

src/gax/src/paginator.rs
```diff
@@ -397,4 +403,89 @@ mod tests {
...
+        let client = Client {
+            inner: Arc::new(InnerClient {
+                data: Arc::new(Mutex::new(responses)),
+            }),
+        };
+        let mut items = vec![];
+        let mut stream = client
+            .list_rpc_stream(TestRequest::default())
+            .items()
+            .to_stream();
```
coryan, line 480: codecov.io seems confused about the coverage for `to_stream()`.  is this code running?

src/integration-tests/tests/pagination.rs
coryan, line 1: Should we have a test for `to_stream()` in this file?

coryan (approved):
The code looks fine to me, but the test coverage results look weird.  I worry that maybe something in the features setting is making us miss testing?

coryan:
Congrats on the code removal btw.

.github/workflows/sdk.yaml
dbolduc: I reverted all changes to `.github/workflows/sdk.yaml`, but that is not obvious from the 9000 commits associated with this PR.

src/gax/src/paginator.rs
dbolduc: Hm, it should be exercised here:

https://github.com/googleapis/google-cloud-rust/blob/66824576babf09a168207a74695fdd1758bec70a/.github/workflows/sdk.yaml#L47

Locally:

```sh
$ cargo test --workspace | grep to_stream
test paginator::tests::test_item_paginator_to_stream ... ok
test paginator::tests::test_paginator_to_stream ... ok
```

src/integration-tests/tests/pagination.rs
dbolduc: Added. The crate depends on clients with the `unstable-stream` feature enabled. I will clean up the features/dependencies in a follow up PR.

coryan (approved):

src/gax/src/paginator.rs
coryan: FWIW, I believe the  coverage build uses a different set of options:

https://github.com/googleapis/google-cloud-rust/blob/66824576babf09a168207a74695fdd1758bec70a/.github/workflows/sdk.yaml#L69-L74

dbolduc: oh, d'uh. I misread the yaml.

PR #1467, opened 2025-03-11
feat(generator): options to skip some elements

coryan:
Sometimes we need to skip some elements in the generated code. The
current motivation is to skip conversion functions to and from Prost
types.

Part of the work for #1414

### DIFF
--- a/generator/internal/api/skip.go
+++ b/generator/internal/api/skip.go
@@ -1,0 +1,48 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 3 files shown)


generator/internal/api/skip.go
```diff
@@ -0,0 +1,48 @@
...
+	model.Services = slices.DeleteFunc(model.Services, func(x *Service) bool { return skippedIDs[x.ID] })
+	for _, service := range model.State.ServiceByID {
+		service.Methods = slices.DeleteFunc(service.Methods, func(x *Method) bool { return skippedIDs[x.ID] })
+	}
+}
+
+func skipMessageElements(message *Message, skippedIDs map[string]bool) {
+	for _, m := range message.Messages {
+		skipMessageElements(m, skippedIDs)
+	}
```
dbolduc, line 45: can we assume there are no cycles?

coryan: Yes.  The *names* may have cycles, but the AST (as the name implies) should not.

dbolduc (approved):

PR #1470, opened 2025-03-11
Prototype refactoring of gax to allow auth to use the retry policy

sai-sunder-s:
- This is just a prototype. If the approach looks good, will spend more time on it, chunk it up and send out few PRs.
- I did not update tests, so haven't built or tested this code. Have removed all errors from the prod code in gax, auth and retry. (generated code still has lots of errors)
- We may have to retire the auth retryable error concept and just use the logic for any rpc.


coryan:
We can close this, we decided to move the client out of `gax` and it is being done in #1490

PR #1472, opened 2025-03-11
feat(auth): support non-default MDS credentials

harkamaljot:
Part of the work for https://github.com/googleapis/google-cloud-rust/issues/1437

Added a builder for supporting non-default mds credentials, which include supplying quota-project-id when fetching headers. Further, allowing user to fetch universe_domain from mds, which can be used later on to make credentials.

Unit tests

Integration tests?
There are none.


coryan (commented):
Please keep the commit title (and consequently the PR title) under 50 characters:

https://stackoverflow.com/questions/2290016/git-commit-messages-50-72-formatting

It gets hard to look at the log (include the PR number that github attaches) if the titles are this long.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -13,37 +13,108 @@
...
     }
 }
 
+#[allow(dead_code)]
 #[derive(Debug)]
 struct MDSCredential<T>
 where
     T: TokenProvider,
 {
+    scopes: Option<Vec<String>>,
```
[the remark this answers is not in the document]
harkamaljot: I had kept it in, to allow the user to get the creds, but since it not needed now. Let me remove it.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -138,6 +136,16 @@ where
         }
         Ok(headers)
     }
+
+    async fn get_universe_domain(&self) -> Option<String> {
+        if self.universe_domain.is_some() {
+            return self.universe_domain.clone();
+        }
+        match self.get_universe_domain_from_mds().await {
```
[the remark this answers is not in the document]
harkamaljot: the problem with that is then we would have to change the getter interface to get_universe_domain(mut& self). Which seems wrong on an interface level as a getter is modifying state of the object. But definitely, agreed we should cache it as otherwise we would call mds each time. Do you think I should change the function to take in mut& self?

src/auth/src/credentials/mds_credential.rs
```diff
@@ -41,9 +44,74 @@ struct MDSCredential<T>
...
+            token_provider,
+        }
+    }
+}
+
+impl<T> MDSCredential<T>
+where
+    T: TokenProvider,
+{
+    async fn get_universe_domain_from_mds(&self) -> Result<String> {
```
coryan, line 80: We always ignore the error, should we consider returning `Option<String>` here?

src/auth/src/credentials/mds_credential.rs
```diff
@@ -41,9 +44,74 @@ struct MDSCredential<T>
...
 {
+    endpoint: String,
+    quota_project_id: Option<String>,
+    universe_domain: RwLock<Option<String>>,
     token_provider: T,
 }
 
+#[bon]
+impl MDSCredential<MDSAccessTokenProvider> {
+    #[builder]
```
coryan, line 55: AFAIK, this type is not public, so applications cannot use this builder.  Maybe I am not aware of how these types are intended to be used?

sai-sunder-s: We would have to make this type public.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -58,7 +126,29 @@ where
...
+    }
+
+    async fn get_universe_domain(&self) -> Option<String> {
+        {
+            let universe_domain = self.universe_domain.read().unwrap();
+            if universe_domain.is_some() {
+                return universe_domain.clone();
+            }
+        }
+        let universe_domain = match self.get_universe_domain_from_mds().await {
```
sai-sunder-s, line 146: It is possible that multiple threads kick off this mds request. You would have to put it behind a lock so that only one thread will call MDS.

sai-sunder-s (changes requested):

src/auth/src/credentials/mds_credential.rs
coryan: Seems to me like this is the same problem we have in the token cache.  Should we solve this once in one place? Or at least use the same solution?

src/auth/src/credentials/mds_credential.rs
coryan: AFAIK, you would have to make `struct MDSCredential<T>` (for all `T`) public.  That would also make all (public) traits implemented by this type public.  Is that what you want?

It also seems that we will have some overlap between this and any MDS client used for resource detection. Or for any other metadata (project ids for example).

Also: how do application developers discover that this has a builder? Do the `bon` macros expand to something obvious in the  reference docs?

coryan (commented):
This PR does not worry me, any PR that makes these APIs public would need more scrutiny.

src/auth/src/credentials/mds_credential.rs
sai-sunder-s: We need to keep refreshing tokens but for UD, it is one time thing. So we do not need the forever background task.

For this, maybe we could do a simplified form of that: 
1. Create a watch channel for "storing" universe domain: Give transmitter to background task and receiver is used by get_universe_domain.
2. Create another watch channel (maybe oneshot will work, need to investigate) for informing the background task. Lets calls this wakeup_channel.
3. Background task will wait to be woken up by a message in the wakeup_channel. Once woken up, it obtains the UD from MDS, pushes it onto the ud_channel and quits.
4. `get_universe_domain` will first check receiver of ud_channel. If a value was available in it, that is returned. If it did not find a value there, it will push a message to wakeup_channel and then wait on a changed notification on the ud_channel. Once the background task provides the ud value on the channel, it wakes up, reads and returns it.

src/auth/src/credentials/mds_credential.rs
sai-sunder-s: Not sure I understood "It also seems that we will have some overlap between this and any MDS client used for resource detection. Or for any other metadata (project ids for example)."

I will let Harkamal answer on the bon, I am not familiar with it.

Making the trait as public is going to be trouble. The credential trait uses [async_trait], so we cannot expose it I believe?

coryan: > Not sure I understood "It also seems that we will have some overlap between this and any MDS client used for resource detection. Or for any other metadata (project ids for example)."

I think eventually we will need a general-purpose MDS client to query other metadata.  Applications may want to dynamically discover the project id from the metadata service (for apps that are deployed to different projects for production vs. testing vs. development).  We will want to query MDS to perform resource detection in the context of OTel exporters.

> 
> I will let Harkamal answer on the bon, I am not familiar with it.
> 
> Making the trait as public is going to be trouble. The credential trait uses [async_trait], so we cannot expose it I believe?

It creates an unstable API.  It seems like we want to expose a different type for application developers to use.  It could be a type that just forwards calls to these internal types, but offers a stable API.

harkamaljot: what if I create a public method inside the module similar to the new method, called build that returns back MDSCredentialBuilder back. Since builder is public, so outside user can call methods on it, once they have a builder instance. Do you think this approach works @coryan and @sai-sunder-s ?

sai-sunder-s: Looking at the [design](https://docs.google.com/document/d/1_y0XhisC7S6TTxOTfatktcXMulD_5MWy6YxNiM3ouzk/edit?pli=1&tab=t.0#bookmark=id.m1277vvgbfq0), the plan was to hand write the builder and return the Credential object in the same was `create_access_token_credential` does. So we would only have to make the builder public, not the MDSCredential itself.

src/auth/src/credentials/mds_credential.rs
harkamaljot: thanks for the detailed explanation, I think oneshot will work here, just if the call happens concurrently then we would ignore error and thread would go onto waiting for the result to come back from the endpoint.

src/auth/src/credentials/mds_credential.rs
harkamaljot: got it, let me handwrite the builder here then.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -13,37 +13,164 @@
...
+        self.universe_domain = Some(universe_domain.into());
+        self
+    }
+
+    pub fn scopes<S: Into<String>>(mut self, scopes: Vec<S>) -> Self {
+        self.scopes = Some(scopes.into_iter().map(|s| s.into()).collect());
+        self
+    }
+
+    pub fn build(&self) -> Credential {
```
coryan, line 85: I think you want:

```suggestion
    pub fn build(self) -> Credential {
```

You can save some cloning I presume.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -13,37 +13,164 @@
...
     T: TokenProvider,
 {
+    quota_project_id: Option<String>,
+    ud_rx: watch::Receiver<UniverseDomainCache>,
+    wakeup_tx: Arc<Notify>,
     token_provider: T,
 }
 
+#[derive(Debug, Default)]
+pub struct MDSCredentialBuilder {
```
coryan, line 56: If the intent was to make this an API that applications can use from outside the crate, I think you are missing some changes.  The `mds_credentials` module is still `pub(crate)`, right?

https://github.com/googleapis/google-cloud-rust/blob/a778170abcd09bf216c5afdffe61b7a8a57f9643/src/auth/src/credentials.rs#L20

If you did intend to turn this into a public API, it needs (1) a test in the `tests/` directory to verify the code is usable from outside the crate, and (2) documentation: https://rust-lang.github.io/api-guidelines/documentation.html#c-crate-doc

The module needs to be documented.

The `struct` should have some comments describing what it is, why it is used, etc.  Each function should have a comment saying what it does.

You should also provide an example, e.g.:

https://github.com/googleapis/google-cloud-rust/blob/a778170abcd09bf216c5afdffe61b7a8a57f9643/guide/samples/src/lro.rs#L168-L213

src/auth/src/credentials/mds_credential.rs
```diff
@@ -13,37 +13,164 @@
...
-    Credential {
-        inner: Arc::new(MDSCredential { token_provider }),
-    }
+    MDSCredentialBuilder::default().build()
+}
+
+#[derive(Debug)]
+struct UniverseDomainCache {
+    universe_domain: Option<String>,
+    is_cached: bool,
```
coryan, line 41: I think you should move the caching to a separate PR.  It is hard to reason about two complex changes at the same time. 

With that said: can you add a comment explaining what `is_cached` does?  Why do we need it?

src/auth/src/credentials/mds_credential.rs
```diff
@@ -13,37 +13,164 @@
...
+        let universe_domain = response
+            .json::<UniverseDomainResponse>()
+            .await
+            .ok()?
+            .universe_domain;
+
+        Some(universe_domain)
+    }
+
+    async fn universe_domain_background_task(
```
coryan, line 154: Why is this a member of `MDSCredential<T>`? It does not use `Self`, and it seems to be called from outside the struct only?

src/auth/src/credentials/mds_credential.rs
```diff
@@ -13,37 +13,164 @@
...
+            inner: Arc::new(mdsc),
+        }
+    }
+}
+
+impl<T> MDSCredential<T>
+where
+    T: TokenProvider,
+{
+    async fn get_universe_domain_from_mds(endpoint: String) -> Option<String> {
```
coryan, line 127: This too? It does not seem to use `Self` or `T`?

coryan (changes requested):

coryan (changes requested):

guide/samples/src/credentials.rs
```diff
@@ -0,0 +1,34 @@
+// Copyright 2025 Google LLC
```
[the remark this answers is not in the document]
harkamaljot: Yeah, ideally we would want to have a chapter that discuss options in various credential types. let me move it out of this PR

guide/samples/src/credentials.rs
```diff
@@ -0,0 +1,34 @@
...
+// ANCHOR: use
+use google_cloud_auth::credentials::mds_credential::MDSCredentialBuilder;
+use google_cloud_auth::credentials::Credential;
+// ANCHOR_END: use
+
+// ANCHOR: build-mds-credentials
+pub fn build_mds_credentials() -> Credential {
+    let mds_endpoint = "http://metadata.google.internal/computeMetadata/v1";
+    let cloud_scopes = vec!["https://www.googleapis.com/auth/cloud-platform"];
+    let universe_domain = "googleapis.com";
```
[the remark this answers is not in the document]
harkamaljot: sure, in the guide PR. I will inline the params here, removing this file for now

src/auth/src/credentials/mds_credential.rs
```diff
@@ -28,22 +76,88 @@ const METADATA_FLAVOR: &str = "metadata-flavor";
...
+    scopes: Option<Vec<String>>,
+    universe_domain: Option<String>,
+}
+
+#[allow(dead_code)]
+impl MDSCredentialBuilder {
+    /// Sets the endpoint that this credential will use.
+    /// # Default
+    ///
+    /// if `endpoint` is not specified then  `http://metadata.google.internal/computeMetadata/v1` is used.
```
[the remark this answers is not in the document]
harkamaljot: @coryan just to confirm we should allow the user only to set ip address/host address of the server i.e. 
{http://metadata.google.internal} should be configurable by the user.

src/auth/src/credentials/mds_credential.rs
```diff
@@ -28,22 +76,88 @@ const METADATA_FLAVOR: &str = "metadata-flavor";
...
+    /// Builds and returns a `Credential` instance with the configured settings.
+    ///
+    /// # Returns
+    ///
+    /// A new `Credential` instance that will use Metadata Service.
+    pub fn build(self) -> Credential {
+        let endpoint = self.endpoint.clone().unwrap_or(METADATA_ROOT.to_string());
+
+        let token_provider = MDSAccessTokenProvider::builder()
+            .endpoint(endpoint.clone())
```
sai-sunder-s, line 146: why to clone endpoint again here? we are not using it after this point, we can just transfer ownership?

src/auth/src/credentials/mds_credential.rs
```diff
@@ -352,6 +487,54 @@ mod test {
         assert!(result.is_err());
     }
 
+    #[tokio::test]
+    async fn get_headers_success_with_quota_project() {
+        let scopes = vec!["scope1".to_string()];
```
sai-sunder-s, line 492: add one more scope to the vec to make sure we concatenate correctly

src/auth/src/credentials/mds_credential.rs
```diff
@@ -12,38 +12,154 @@
...
+    /// [quota project]: https://cloud.google.com/docs/quotas/quota-project
+    pub fn quota_project_id<S: Into<String>>(mut self, quota_project_id: S) -> Self {
+        self.quota_project_id = Some(quota_project_id.into());
+        self
+    }
+
+    /// Sets the universe_domain for this credential.
+    /// Client libraries use `universe_domain` to determine
+    /// the API endpoints to use for making requests.
+    /// If not set, then credentials use `googleapis.com`.
```
harkamaljot, line 125: this documentation will change, once I submit get_universe_domain function

src/auth/src/credentials/mds_credential.rs
```diff
@@ -12,38 +12,154 @@
...
+impl MDSCredentialBuilder {
+    /// Sets the endpoint for this credential.
+    /// If not set, the credentials use `http://metadata.google.internal/`.
+    pub fn endpoint<S: Into<String>>(mut self, endpoint: S) -> Self {
+        self.endpoint = Some(endpoint.into());
+        self
+    }
+
+    /// Set the [quota project] for this credential.
+    /// In some services, you can use a service account in
```
coryan, line 111: nit: here and everywhere else, separate short description from detailed description:

https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#documenting-components

```suggestion
    /// Set the [quota project] for this credential.
    ///
    /// In some services, you can use a service account in
```

src/auth/src/credentials/mds_credential.rs
```diff
@@ -12,38 +12,154 @@
...
+    /// one project for authentication and authorization, and charge
+    /// the usage to a different project. This may require that the
+    /// service account has serviceusage.services.use on the quota project.
```
coryan, line 114: nit:

```suggestion
    /// service account has `serviceusage.services.use` permissions on the quota project.
```

src/auth/src/credentials/mds_credential.rs
```diff
@@ -12,38 +12,154 @@
...
+    /// Metadata server issues tokens based on the requested scopes.
+    /// If no scopes are specified, the credential defaults to all
+    /// scopes configured for the [default service account] on the instance.
+    ///
+    /// [default service account]: https://cloud.google.com/iam/docs/service-account-types#default
+    /// [scopes]: https://developers.google.com/identity/protocols/oauth2/scopes
+    pub fn scopes<S: Into<String>>(mut self, scopes: Vec<S>) -> Self {
+        self.scopes = Some(scopes.into_iter().map(|s| s.into()).collect());
+        self
+    }
```
coryan, line 141: Generalize, you want something that supports `IntoIterator<T>`:

```suggestion
    pub fn scopes<I, S>(mut self, scopes: I) -> Self
    where:
        I: IntoIterator<Item = S>,
        S: Into<String>
     {
        self.scopes = Some(scopes.into_iter().map(|s| s.into()).collect());
        self
    }
```

src/auth/src/credentials/mds_credential.rs
```diff
@@ -12,38 +12,154 @@
...
+/// backed by the metadata service, some applications may need to:
+/// * Customize the metadata service credentials in some way
+/// * Bypass the [Application Default Credentials] lookup and only
+///    use the metadata server credentials
+/// * Use the credentials directly outside the client libraries
+///
+/// [Application Default Credentials]: https://cloud.google.com/docs/authentication/application-default-credentials
+/// [Metadata Service]: https://cloud.google.com/compute/docs/metadata/overview
+#[derive(Debug, Default)]
+pub struct MDSCredentialBuilder {
```
coryan, line 95: The name of this type repeats itself too much.  Applications would need to write:

```
use google_cloud_auth::credentials::mds_credentials::MDSCredentialBuilder;
```

That is saying credentials 3 times, and MDS twice.  Consider naming things such that this works:

```
use google_cloud_auth::credentials::mds::Builder;
```

or maybe this:

```
use google_cloud_auth::mds_credentials::Builder;
```

src/auth/src/credentials/mds_credential.rs
```diff
@@ -12,38 +12,154 @@
...
+    /// the usage to a different project. This may require that the
+    /// service account has serviceusage.services.use on the quota project.
+    ///
+    /// [quota project]: https://cloud.google.com/docs/quotas/quota-project
+    pub fn quota_project_id<S: Into<String>>(mut self, quota_project_id: S) -> Self {
+        self.quota_project_id = Some(quota_project_id.into());
+        self
+    }
+
+    /// Sets the universe_domain for this credential.
```
coryan, line 122: nit: I think

```suggestion
    /// Sets the universe domain for this credential.
```

src/auth/src/credentials/mds_credential.rs
```diff
@@ -12,38 +12,154 @@
...
+    /// Client libraries use `universe_domain` to determine
+    /// the API endpoints to use for making requests.
+    /// If not set, then credentials use `googleapis.com`.
```
coryan, line 125:
```suggestion
    /// If not set, then credentials use `${service}.googleapis.com`, otherwise they use `${service}.${universe_domain}.
```

src/auth/src/credentials/mds_credential.rs
coryan: Is there a page in cloud.google.com describing universe domains in any detail?

coryan (changes requested):

src/auth/src/credentials/mds_credential.rs
harkamaljot: updated it to use google_cloud_auth::credentials::mds::Builder;
the other credentials use the same naming convention of mentioning credential multiple times, so when updating those I will update them using the same convention used here

src/auth/src/credentials/mds.rs
```diff
@@ -352,6 +502,53 @@ mod test {
...
+                TokenQueryParams {
+                    scopes: Some(scopes.join(",")),
+                    recursive: None,
+                },
+            ),
+        )]))
+        .await;
+
+        let mdsc = Builder::default()
+            .scopes(scopes)
```
coryan, line 529: You should test the `IntoIterator` thing at least once, maybe here write:

```suggestion
            .scopes(["scope1", "scope2"])
```

src/auth/src/credentials/mds_credential.rs
harkamaljot: sorry there doesn't seem to any documentation existing for `cloud.google.com`. I asked around our TPC poc and asked in the TPC group as well. In case, I find anything I will update the comment with doc link.

src/auth/src/credentials/mds.rs
```diff
@@ -12,38 +12,163 @@
...
+//! Example usage:
+//!
+//! ```
+//! # use google_cloud_auth::credentials::mds::Builder;
+//! # use google_cloud_auth::credentials::Credential;
+//! # use google_cloud_auth::errors::CredentialError;
+//! # tokio_test::block_on(async {
+//! let credential: Credential = Builder::default().quota_project_id("my-quota-project").build();
+//! let token = credential.get_token().await?;
+//! println!("Token: {}", token.token);
```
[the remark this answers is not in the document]
harkamaljot: thanks

coryan:
I changed the title. Since it introduces a public API `feat` is a better conventional commit tag than `impl`.  And the important thing is what customers can do with the thing, not that a builder is involved.

sai-sunder-s (approved):

PR #1474, opened 2025-03-12
impl(firestore): helpers for `google.type.*`

coryan:
Add helper functions to convert between Prost and sidekick types in the
`google.type` namespace. If these types are common enough across veneers, we
can move the code to a common library and cover all the messages in the
package.

Part of the work for #1414


dbolduc (approved):

dbolduc:
> If these types are common enough across veneers, we
can move the code to a common library and cover all the messages in the
package.

Was about to ask. SGTM

PR #1476, opened 2025-03-12
prototype: refactor gax to pull out whole http_client

sai-sunder-s:


coryan:
I think we can close this.  The changes related to #1490 have the same effect.

Issue #1384, opened 2025-02-27
Prototype a client with gRPC+Protobuf


coryan:
We need to prototype a client using gRPC for the transport (and Protobuf for the data). The objectives of this prototype are:

- [x] Determine how code generated by sidekick, Protobuf, and gRPC can co-exist in the same crate
- [x] Determine how code generated by sidekick, Protobuf, and gRPC gets updated on googleapis SHA changes.
   - [x] We want consistent inputs to the generators
   - [x] We want easy updates ideally via sidekick
- [x] Determine how we convert between sidekick-generated types and Protobuf-generated types.
   - [x] We anticipate some veneers (or fully machine stamped) where we expose the sidekick types to the application
   - [x] We anticipate some veneers where not all the sidekick types are exposed or used.
- [x] Determine how the client surface differs (initialization? still use the same builders?) from the sidekick generated surface.
- [x] Determine how we support additional helpers to initialize complex types, like Firestore query objects.

The outcome of this will be a design doc.

This document may suggest changes to the purely generated clients to make wrapping of gRPC and veneers easier to write.

coryan:
The design doc is out for review. I think this is done.

PR #1484, opened 2025-03-13
feat(generator/rust): convert enum types

coryan:
Motivated by #1384


dbolduc (commented):
I am iffy on this. It seems cleaner to generate a `cnv` that just calls `value`. Why do you prefer this way?

dbolduc (approved):

dbolduc:
> Because in some cases we cannot generate the `.cnv()` at all.

Thanks for the explanation.

Issue #1468, opened 2025-03-11
Enable features in docs.rs


dbolduc:
Our docs do not currently have optional APIs that are feature-gated, such as the paginated `stream()` API.

For any feature with optional features that add APIs, we probably want a section in the `Cargo.toml` for:

```rs
[package.metadata.docs.rs]
all-features = true
```

See: https://docs.rs/about/metadata

We should also explicitly state in the method comments which feature enables the API.

dbolduc:
We actually do not want `all-features`, as some features control APIs that we do not want in our docs.

The only crates that have optional features gating APIs (at the moment) are `gax` and `wkt`.

coryan:
How does `all-features` work with `#[doc(hidden)]` ?  Maybe if we explicitly hide the things that are for `unstable` or `_foo` features then we avoid list explicit features in `Cargo.toml`.  OTOH, I am not sure which approach requires more maintenance work.

PR #1487, opened 2025-03-13
docs(gax): include streaming APIs

dbolduc:
Part of the work for #1468 

Paginator will (hopefully soon) pick up a `fn to_stream()` which is gated by the `unstable-stream` feature. We want this to appear in our docs.

While client crates currently have an `unstable-stream` feature which is gating their `stream()` methods, and preventing the methods from appearing in our docs, we plan to change the method to `paginator()` and make it always available. (Hence there are no changes to client crates).

See: https://docs.rs/about/metadata

---

Aside: Remind me where I can look at the generated documentation to make sure this change actually works...

### DIFF
--- a/src/gax/Cargo.toml
+++ b/src/gax/Cargo.toml
@@ -25,6 +25,12 @@ keywords.workspace   = true
 license.workspace    = true
 repository.workspace = true
 
+[package.metadata.docs.rs]
+# We want to generate documentation for streaming APIs, gated by this feature.
+# We do not want to generate documentation for `unstable-sdk-client`, which
+# gates internal tpes.
+features = ["unstable-stream"]
+
 [dependencies]
 base64      = "0.22"
 bytes       = "1"


src/gax/Cargo.toml
```diff
@@ -25,6 +25,12 @@ keywords.workspace   = true
 license.workspace    = true
 repository.workspace = true
 
+[package.metadata.docs.rs]
+# We want to generate documentation for streaming APIs, gated by this feature.
+# We do not want to generate documentation for `unstable-sdk-client`, which
```
coryan, line 30: Maybe it should be called `_unstable-sdk-client`...

coryan (approved):

src/gax/Cargo.toml
dbolduc: Ack, not for this PR.

Issue #1490, opened 2025-03-13
Refactor `gax::http_client` to a separate crate


coryan:
We want to use some types in `gax` from the `auth` library. That suggests moving the `ReqwestClient` implementation out of `gax`.

It will also reduce the number of things hidden under the `unstable-sdk-client` feature, with the view of eliminating that feature altogether.

Roughly speaking we need to:

- [x] Create a new crate (`google-cloud-http-client` ? SGTM), initially this will just re-export types from `gax`.
- [x] Change all the generated libraries to depend on and use this crate.
- [x] Change any hand-crafted libraries to depend on and use this crate.
- [x] Move the implementation from `gax` to the new crate

There are probably steps that we will discover as we make progress.

coryan:
Closing. The work as described in the title is done. Created #1542 for some additional work related to auth.

PR #1491, opened 2025-03-13
impl(http-client): placeholder to split `gax`

coryan:
Create a new library that will contain the `ReqwestClient` from `gax`.
In this commit, the library just redefines types from `gax`.

Part of the work for #1490


src/gax/src/lib.rs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-//! Google APIs helpers.
+//! Google APIs eXtensions for Rust.
```
dbolduc, line 15: :)

src/http-client/Cargo.toml
```diff
@@ -0,0 +1,28 @@
+# Copyright 2024 Google LLC
```
dbolduc, line 1: nit

src/http-client/src/lib.rs
```diff
@@ -0,0 +1,26 @@
...
+//! Implementation details for Google Cloud clients using HTTP as their
+//! transport.
+//!
+//! All the types, traits, and functions defined in this crate are **not**
+//! intended for general use. This crate will remain unstable for the
+//! foreseeable future, even if used in the implementation for stable client
+//! libraries. We (the Google Cloud Client Libraries for Rust team) control
+//! both and will change both if needed.
+
+pub use gax::http_client::ClientConfig;
```
dbolduc, line 24: This isn't moving from `gax::options`, right? We probably don't need to re-export it.

dbolduc (approved):
*accidental review comment*

sai-sunder-s (approved):

PR #1492, opened 2025-03-13
impl(auth)!: ring for default crypto provider

dbolduc:
Related to #1077 

The `ring` crate is less offensive than the `aws-lc-rs` crate. So we will depend on it by default.

We will not offer a way to exclude the `ring` crate in the dependency tree. That requires propagating a default feature to all N of our client crates. It is not worth it.


coryan (approved):

coryan (commented):
Do we have a test where we install a different provider?

dbolduc:
> Do we have a test where we install a different provider?

No, but let me add one in another PR. I have a test lying around.

See the "Testing?" section of this PR description: https://github.com/googleapis/google-cloud-rust/pull/899

PR #1496, opened 2025-03-14
test(firestore): integration tests for `Convert<>`

coryan:
We need to have integration tests for the generated `Convert<>`
implementations somewhere in the code. For reasons documented in the
comments, this is the only place where we can put them at the moment.

Part of the work for #1384


src/firestore/src/convert.rs
```diff
@@ -0,0 +1,227 @@
...
+            model::structured_query::unary_filter::OperandType::Field(Box::new(
+                model::structured_query::FieldReference::new().set_field_path("a.b.c"),
+            )),
+        );
+        let proto = google::firestore::v1::structured_query::UnaryFilter {
+            operand_type:
+                google::firestore::v1::structured_query::unary_filter::OperandType::Field(
+                    google::firestore::v1::structured_query::FieldReference {
+                        field_path: "a.b.c".into(),
+                        ..Default::default()
```
dbolduc, line 90: (so I can learn) how do I read `..Default::default()`? Like, what do the leading dots mean?

src/firestore/src/convert.rs
```diff
@@ -0,0 +1,227 @@
...
+
+        let got = sidekick.clone().cnv();
+        assert_eq!(got, proto);
+
+        let got = proto.clone().cnv();
+        assert_eq!(got, sidekick);
+    }
+
+    #[test]
+    fn test_enum_field() {
```
dbolduc, line 63: thank you for naming the tests after the field types and not the message names used within.

dbolduc (approved):

src/firestore/src/convert.rs
coryan: I have been cargo culting this syntax without reading what it means. I guess when I read this chapter the idea did not sink in:

https://doc.rust-lang.org/stable/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax

dbolduc: Ah, that is cool, thanks. TIL: Struct update syntax. This could have simplified a lot of my auth unit tests.

Also TIL: cargo culting. Unrelated to `cargo`.

PR #1497, opened 2025-03-14
refactor: move `ReqwestClient`

coryan:
Move the `ReqwestClient` implementation to `google-cloud-http-client`.

Part of the work for #1490


src/gax/src/error/rpc/mod.rs
```diff
@@ -48,6 +48,36 @@ pub struct Status {
     pub details: Vec<StatusDetails>,
```
dbolduc, line 48: do we want getters for the fields? It would be a break, but more consistent with other structs in this PR.

dbolduc (approved):

src/gax/src/error/rpc/mod.rs
coryan: Punting for now.

PR #1499, opened 2025-03-14
cleanup!: remove 'unstable-stream' feature from clients

### LINKED ISSUE #720: Implement pagination helpers without streams

dbolduc:
Fixes #720 

This feature is now present in the `lro` and `gax` crates only. Clients do not need it.

The golang code to detect whether a service has a paginated API seems useful, but we do not currently need it. So the code is deleted. We can find this PR in the future.


guide/samples/Cargo.toml
```diff
@@ -39,9 +39,7 @@ google-cloud-gax = { version = "0.20", path = "../../src/gax" }
 google-cloud-speech-v2 = { version = "0.2", path = "../../src/generated/cloud/speech/v2" }
 # ANCHOR_END: speech
 # ANCHOR: secretmanager
-google-cloud-secretmanager-v1 = { version = "0.2", features = [
-  "unstable-stream",
-], path = "../../src/generated/cloud/secretmanager/v1" }
+google-cloud-secretmanager-v1 = { version = "0.2", path = "../../src/generated/cloud/secretmanager/v1" }
```
coryan, line 42: Aside: I wish we could hide the `path = ....` part in generated user guide.

guide/src/setting_up_rust_on_cloud_shell.md
```diff
@@ -55,7 +55,7 @@ Cloud Shell is a great environment to run small examples and tests.
 1. Add the [Secret Manager] client library to the new project
 
    ```shell
-   cargo add google-cloud-secretmanager-v1 --features unstable-stream
+   cargo add google-cloud-secretmanager-v1
```
coryan, line 58: 🎉

coryan (approved):

guide/samples/Cargo.toml
dbolduc: the first bad idea that comes to mind: have some script that strips off the path and writes a new file. Then reference the new file in the doc.

```sh
# this works while our client crates do not have features we want to enable.
$ sed 's/^\(google-cloud-[^ ]*\) = { version = "\([^"]*\)".*/\1 = "\2"/' guide/samples/Cargo.toml > guide/samples/ScrubbedCargo.toml
```

```sh
$ cat guide/samples/ScrubbedCargo.toml
<omitted>

[dependencies]
tokio = { version = "1", features = ["full", "macros"] }
# ANCHOR: longrunning
google-cloud-longrunning = "0.22"
# ANCHOR_END: longrunning
# ANCHOR: gax
google-cloud-gax = "0.20"
# ANCHOR_END: gax
# ANCHOR: speech
google-cloud-speech-v2 = "0.2"
# ANCHOR_END: speech
# ANCHOR: secretmanager
google-cloud-secretmanager-v1 = "0.2"
# ANCHOR_END: secretmanager

<omitted>
```

PR #1504, opened 2025-03-14
feat(http-client): re-export parameter helpers

coryan:
The path parameter and query parameter helpers are only used for
HTTP+JSON based clients, so they belong in `google-cloud-http-client`.

Part of the work for #1490


dbolduc (approved):
Q: is the goal to have 0 documentation for the types in this crate?

coryan:
> Q: is the goal to have 0 documentation for the types in this crate?

I would not phrase it like that.  I would say that this crate will only have implementation details, and those should not be documented.

PR #1520, opened 2025-03-17
refactor: remove `TokenExtractor`

### LINKED ISSUE #1498: Remove `TokenExtractor` from gax
We have a `TokenExtractor` hidden behind the `unstable-sdk-client` feature in `gax`. We use this in our generated client crates, but don't really want to expose it in our public APIs.

It converts from a `String` or an `Option<String>` to a `String`.

Apparently the generator knows whether a field is optional or not. So we can handle this in the generator, and remove the public-but-secret helper from `gax`.

This should also let us drop the dependency of `firestore` on `unstable-sdk-client`. It also moves us closer to our desired goal of having internal-only crates, instead of internal-only features of public crates.

dbolduc:
Fixes #1498 

The claim that "I have simplified the code" is dubious. But at least I have a good understanding of how mustache + our Go types work in the generator. If we don't want this, I am happy to close it.

Maybe the code is harder to read if I use `{{Codec.FieldName}}` instead of `page_token` (for example) when I have previously confirmed that `{{Codec.FieldName}} == "page_token"`. :shrug:

### DIFF
--- a/generator/internal/rust/templates/common/message.mustache
+++ b/generator/internal/rust/templates/common/message.mustache
@@ -164,7 +164,7 @@ impl wkt::message::Message for {{Codec.Name}} {
     }
 }
 {{/Codec.HasSyntheticFields}}
-{{#IsPageableResponse}}
+{{#Pagination}}
 
 impl gax::paginator::PageableResponse for {{Name}} {
     {{#PageableItem}}
@@ -175,11 +175,22 @@ impl gax::paginator::PageableResponse for {{Name}} {
     }
 
     {{/PageableItem}}
+    {{#NextPageToken}}
     fn next_page_token(&self) -> std::string::String {
-        gax::paginator::extract_token(&self.next_page_token)
+        use std::clone::Clone;
+        {{^Optional}}
+        self.{{Codec.FieldName}}.clone()
+        {{/Optional}}
+        {{#Optional}}
+        match &self.{{Codec.FieldName}} {
+            std::option::Option::Some(v) => v.clone(),
+            std::option::Option::None => std::string::String::new(),
+        }
+        {{/Optional}}
     }
+    {{/NextPageToken}}
 }
-{{/IsPageableResponse}}
+{{/Pagination}}
 {{#Codec.HasNestedTypes}}
 
 /// Defines additional types related to {{Name}}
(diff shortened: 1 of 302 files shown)


generator/internal/rust/templates/common/message.mustache
```diff
@@ -175,11 +175,22 @@ impl gax::paginator::PageableResponse for {{Name}} {
...
-        gax::paginator::extract_token(&self.next_page_token)
+        use std::clone::Clone;
+        {{^Optional}}
+        self.{{Codec.FieldName}}.clone()
+        {{/Optional}}
+        {{#Optional}}
+        match &self.{{Codec.FieldName}} {
+            std::option::Option::Some(v) => v.clone(),
+            std::option::Option::None => std::string::String::new(),
+        }
```
coryan, line 188:
```suggestion
        self.{{Codec.FieldName}}.cloned().unwrap_or_default()
```

coryan: Though that may need a `use std::option::Option;`

generator/internal/rust/templates/crate/src/builders.rs.mustache
```diff
@@ -83,19 +83,28 @@ pub mod {{Codec.ModuleName}} {
...
 
         /// Streams the responses back.
         pub async fn paginator(self) -> gax::paginator::Paginator<{{OutputType.Codec.QualifiedName}}, gax::error::Error> {
-            let token = gax::paginator::extract_token(&self.0.request.page_token);
+            use std::clone::Clone;
+            {{#Optional}}
+            let token = match &self.0.request.{{Codec.FieldName}} {
+                std::option::Option::Some(v) => v.clone(),
+                std::option::Option::None => std::string::String::new(),
+            };
```
coryan, line 95: Ditto, this can be `.cloned().unwrap_or_default()`.

generator/testdata/rust/protobuf/golden/location/src/model.rs
```diff
@@ -134,7 +134,8 @@ impl gax::paginator::PageableResponse for ListLocationsResponse {
     }
 
     fn next_page_token(&self) -> std::string::String {
-        gax::paginator::extract_token(&self.next_page_token)
+        use std::clone::Clone;
+        self.next_page_token.clone()
```
coryan, line 138: I think you could write:

```suggestion
        std::clone::Clone::clone(&self.next_page_token)
```

But I am not sure that is better in any way.

coryan (changes requested):

generator/internal/rust/templates/common/message.mustache
dbolduc: Done.

generator/internal/rust/templates/crate/src/builders.rs.mustache
dbolduc: Done.

coryan (approved):

Issue #645, opened 2025-01-08
Support all the things in Any


coryan:
I am not sure if `Any` needs to support `enum` objects or if we would ever use `google.protobuf.*Value` into an Any.

They are less important than messages, so I will punt on them and this is a feature request to fix them someday, maybe, if we need to.

- [x] #1151
- [ ] `google.protobuf.*Value`
- [x] Enums - confirmed with the Protobuf team, only messages can be stored into  `Any`.

PR #1524, opened 2025-03-17
impl(wkt): `BoolValue` in `Any`

dbolduc:
Part of the work for #645

Implement `Message` for `BoolValue`, as suggested by @coryan.

If this looks right, I'll proceed with the other `*Value` types in `wrappers.rs`.

### DIFF
--- a/src/wkt/src/wrappers.rs
+++ b/src/wkt/src/wrappers.rs
@@ -110,3 +110,56 @@ pub type StringValue = String;
 ///
 /// The JSON representation for `BytesValue` is JSON string.
 pub type BytesValue = bytes::Bytes;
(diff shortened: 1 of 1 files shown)


coryan (approved):
This change LGTM.  I think we could use at least one test for "invalid usage results in an error".

As you expand this: keep an eye open for reuse, maybe via some common trait to implement `crate::message::Message`, or maybe via some macros, or a combination of these.

dbolduc:
> I think we could use at least one test for "invalid usage results in an error".

Added a test where we `Any::try_into_message::<BoolValue>()` and the thing is not technically a `BoolValue`.

PR #1525, opened 2025-03-17
impl(wkt): `Message` for several wrapper types

dbolduc:
Part of the work for #645 

Implement `Message` for the wrapper types (except for `BytesValue`, `UInt64Value`, and `Int64Value`, which require extra care to serialize/deserialize).

Use macros, because the code is repetitive.

### DIFF
--- a/src/wkt/src/wrappers.rs
+++ b/src/wkt/src/wrappers.rs
@@ -145,8 +145,6 @@ macro_rules! impl_message {
 
 impl_message!(DoubleValue);
 impl_message!(FloatValue);
-impl_message!(Int64Value);
-impl_message!(UInt64Value);
 impl_message!(Int32Value);
 impl_message!(UInt32Value);
 impl_message!(BoolValue);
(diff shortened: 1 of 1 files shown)


src/wkt/src/wrappers.rs
```diff
@@ -111,63 +111,98 @@ pub type StringValue = String;
...
     use crate::Any;
     type Result = std::result::Result<(), Box<dyn std::error::Error>>;
+    use test_case::test_case;
 
-    #[test]
-    fn test_bool_value() -> Result {
-        let input: BoolValue = true;
+    #[test_case(1234.5 as DoubleValue, "DoubleValue")]
+    #[test_case(9876.5 as FloatValue, "FloatValue")]
+    #[test_case(-1234567890123456789 as Int64Value, "Int64Value")]
```
coryan, line 165: I think `int64` and `uint64` values need to be encoded as strings.  `serde_json` tolerates numbers with more bits, but over-the-wire JSON requires these to be stringified.  You may want to split them to a different implementation, like you are (I am guessing) plan to do for `bytes::Bytes`.

coryan (changes requested):

src/wkt/src/wrappers.rs
dbolduc: > I think `int64` and `uint64` values need to be encoded as strings.

Yeah you are correct. Good catch and thanks.

I removed them from this PR. Will add them in a subsequent PR.

PR #1536, opened 2025-03-19
impl(wkt): `BytesValue` in `Any`

### LINKED ISSUE #645: Support all the things in Any

dbolduc:
Fixes #645 

https://protobuf.dev/programming-guides/json/ says:

> JSON value will be the data encoded as a string using standard base64 encoding with paddings. Either standard or URL-safe base64 encoding with/without paddings are accepted."

I think using `STANDARD` as the base64 generator makes the most sense.

### DIFF
--- a/src/wkt/src/wrappers.rs
+++ b/src/wkt/src/wrappers.rs
@@ -213,6 +213,34 @@ impl crate::message::Message for Int64Value {
     }
 }
 
+use base64::{engine::general_purpose::STANDARD, Engine};
+
+impl crate::message::Message for BytesValue {
+    fn typename() -> &'static str {
+        "type.googleapis.com/google.protobuf.BytesValue"
+    }
+    fn to_map(&self) -> Result<crate::message::Map, crate::AnyError>
+    where
+        Self: serde::ser::Serialize + Sized,
+    {
+        encode_string::<Self>(STANDARD.encode(self))
+    }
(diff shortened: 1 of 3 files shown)


src/wkt/src/wrappers.rs
```diff
@@ -213,17 +213,49 @@ impl crate::message::Message for Int64Value {
     }
 }
 
+use base64::{engine::general_purpose::STANDARD, Engine};
```
coryan, line 216: nit: I think we try to avoid `use` in the middle of a module. Sometimes we add them at the beginning of functions.  Now you made me thing about whether `use` declarations affect all the code or just the code after the declaration... 🤔

coryan (approved):

src/wkt/src/wrappers.rs
coryan: Obviously they are visible in the full module, like any other declaration.  Better move this to the top of the file.

dbolduc: Moved

Issue #829, opened 2025-01-24
Generate library for `google/cloud/bigquery/v2` (admin)


coryan:
## Blocked by:

- [x] #837  - if we always use Box<T> for messages in oneofs we solve the cycles found here
- [ ] #858
- [x] #859

----

## Generate

```
cargo new --lib --vcs none src/generated/cloud/bigquery/v2
taplo fmt Cargo.toml
go -C generator/ run ./cmd/sidekick generate     -project-root ..     -specification-source google/cloud/bigquery/v2     -service-config google/cloud/bigquery/v2/bigquery_v2.yaml -output src/generated/cloud/bigquery/v2
cargo fmt
```

The build fails.

## A number of duplicate enum values:

```
error[E0428]: the name `MINIMAL` is defined multiple times
    --> src/generated/cloud/bigquery/v2/src/model.rs:3925:9
     |
3922 |         pub const MINIMAL: &str = "minimal";
     |         ------------------------------------ previous definition of the value `MINIMAL` here
...
3925 |         pub const MINIMAL: &str = "MINIMAL";
     |         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `MINIMAL` redefined here
     |
     = note: `MINIMAL` must be defined only once in the value namespace of this module
```

# Another kind of cycle making infinitely-sized structs

```
error[E0072]: recursive types `StandardSqlDataType` and `SubType` have infinite size
     --> src/generated/cloud/bigquery/v2/src/model.rs:20289:1
      |
20289 | pub struct StandardSqlDataType {
      | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
...
20296 |     pub sub_type: std::option::Option<crate::model::standard_sql_data_type::SubType>,
      |                                       --------------------------------------------- recursive without indirection
...
20414 |     pub enum SubType {
      |     ^^^^^^^^^^^^^^^^
20415 |         /// The type of the array's elements, if type_kind = "ARRAY".
20416 |         ArrayElementType(crate::model::StandardSqlDataType),
      |                          --------------------------------- recursive without indirection
      |
help: insert some indirection (e.g., a `Box`, `Rc`, or `&`) to break the cycle
      |
20296 ~     pub sub_type: std::option::Option<Box<crate::model::standard_sql_data_type::SubType>>,
20297 | }
...
20415 |         /// The type of the array's elements, if type_kind = "ARRAY".
[...]

PR #1538, opened 2025-03-19
feat(generator/rust): skip duplicate enum values

coryan:
After converting the enum value names to Rust style we may find some
duplicates. With this change the generator can skip the dups. In all
practical cases the dups are harmless, they differ in case but have the
name numeric value. We warn if this assumption fails.

part of the work for #829

### DIFF
--- a/generator/internal/rust/annotate_test.go
+++ b/generator/internal/rust/annotate_test.go
@@ -289,14 +289,15 @@ func TestEnumAnnotations(t *testing.T) {
 	}
 	annotateModel(model, codec, "")
 
-	if diff := cmp.Diff(&enumAnnotation{
-		Name:             "TestEnum",
-		ModuleName:       "test_enum",
-		QualifiedName:    "crate::model::TestEnum",
-		RelativeName:     "TestEnum",
-		DocLines:         []string{"/// The enum is documented."},
-		DefaultValueName: "VALUE",
-	}, enum.Codec); diff != "" {
+	want := &enumAnnotation{
+		Name:          "TestEnum",
+		ModuleName:    "test_enum",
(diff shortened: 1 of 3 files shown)


generator/internal/rust/annotate_test.go
```diff
@@ -325,6 +325,57 @@ func TestEnumAnnotations(t *testing.T) {
...
+		Name:   "FULL",
+		ID:     ".test.v1.TestEnum.FULL",
+		Number: 1,
+	}
+	v2 := &api.EnumValue{
+		Name:   "partial",
+		ID:     ".test.v1.TestEnum.partial",
+		Number: 2,
+	}
+	// This does not happen in practice, but we want to verify the code can
```
dbolduc, line 345: This gave me pause. Maybe the generator should panic so we can go yell at the service team?

If we do receive the value 3, I guess we will just treat it as an UNKNOWN_VALUE:3 (or whatever). Which should be fine.

dbolduc (approved):

generator/internal/rust/annotate_test.go
coryan: > This gave me pause. Maybe the generator should panic so we can go yell at the service team?

I thought about that, but then we would stuck: `sidekick update` fails for all services, not just the offending one. We would need to hack something.

> If we do receive the value 3, I guess we will just treat it as an UNKNOWN_VALUE:3 (or whatever). Which should be fine.

Exactly. It is very unlikely to happen (https://google.aip.dev/126 says enum values **must** be SCREAMING_SNAKE_CASE, the deviations are service bugs).  And if it does happen we still can handle the incoming enums.

PR #1540, opened 2025-03-19
feat(bigquery/admin/v2): generate library

### LINKED ISSUE #829: Generate library for `google/cloud/bigquery/v2` (admin)

coryan:
Generate the "admin" library. This is basically the bigquery/v2 library with
some omissions.

I think we will want to include the `Query()` and `GetQueryResults()` RPCs in a
hand-crafted library, and probably using the BigQuery Storage APIs as the
underlying implementation. If I am wrong, we can always add these methods back.

Fixes #829


src/generated/cloud/bigquery/admin/v2/.sidekick.toml
```diff
@@ -0,0 +1,32 @@
...
+.google.cloud.bigquery.v2.QueryRequest,\
+.google.cloud.bigquery.v2.QueryResponse,\
+.google.cloud.bigquery.v2.JobService.GetQueryResults,\
+.google.cloud.bigquery.v2.GetQueryResultsRequest,\
+.google.cloud.bigquery.v2.GetQueryResultsResponse"""
+
+[codec]
+copyright-year = '2025'
+package-name-override = 'google-cloud-bigquery-admin-v2'
+not-for-publication = 'true'
```
dbolduc, line 32: Why?

Cargo.toml
```diff
@@ -69,6 +69,7 @@ members = [
   "src/generated/cloud/beyondcorp/appgateways/v1",
   "src/generated/cloud/beyondcorp/clientconnectorservices/v1",
   "src/generated/cloud/beyondcorp/clientgateways/v1",
+  "src/generated/cloud/bigquery/admin/v2",
```
dbolduc, line 72: I am iffy, but ok with making up an `admin` directory when it doesn't exist...

I am guessing the eventual veneer will be `google-cloud-bigquery` and you are trying to avoid confusion between that and `google-cloud-bigquery-v2`

src/generated/cloud/bigquery/admin/v2/.sidekick.toml
```diff
@@ -0,0 +1,32 @@
...
+.google.cloud.bigquery.v2.PostQueryRequest,\
+.google.cloud.bigquery.v2.QueryRequest,\
+.google.cloud.bigquery.v2.QueryResponse,\
+.google.cloud.bigquery.v2.JobService.GetQueryResults,\
+.google.cloud.bigquery.v2.GetQueryResultsRequest,\
+.google.cloud.bigquery.v2.GetQueryResultsResponse"""
+
+[codec]
+copyright-year = '2025'
+package-name-override = 'google-cloud-bigquery-admin-v2'
```
dbolduc, line 31: nit: is this actually needed? Isn't it the computed package name?

dbolduc (approved):
Approved, but check the `not-for-publication = 'true'` bit. Seems wrong to me, but maybe I don't know the plan.

src/generated/cloud/bigquery/admin/v2/.sidekick.toml
coryan: Because I am not sure if we want to remove more RPCs before publishing.

src/generated/cloud/bigquery/admin/v2/.sidekick.toml
coryan: The computed package name is `google-cloud-bigquery-v2`, it is based on the Protobuf package, not the directory.

dbolduc: Ack, thanks

src/generated/cloud/bigquery/admin/v2/.sidekick.toml
```diff
@@ -0,0 +1,32 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[general]
+specification-source = 'google/cloud/bigquery/v2'
+service-config = 'google/cloud/bigquery/v2/bigquery_v2.yaml'
+
+[source]
```
shollyman, line 19: This skiplist feels odd to me, but mostly because BQ has multiple (perhaps too many) ways of doing things.  Users can still insert and poll queries through the other RPCs in JobService, but the removal of GetQueryResults does remove the ability to read rows without using an additional API (bigquerystorage).

coryan: The notion is that an eventual veneer should provide the *only* way to retrieve query results.  The veneer can use BigQuery storage, or maybe switch between either.

The other rationale is that we can always add these methods back if we want to (or need to).  Whereas it is not possible to remove methods once this thing is released.

So far I am not hearing anything that tells me this skip list is a terrible idea.  Please do speak up if I am being stubborn.

Separately: is the skip list incomplete?  Should we remove more methods (or even the complete `JobService`) until the story with a veneer is clearer?

shollyman: I think if there's still some questions about the veneer shape then redacting all of JobService is the best option for the time being, rather than clawing back methods down the line.  It does limit user capabilities around data (import/export/copying), but it better targets the whole of the "compute" abstraction.  Also, there are some things like inline uploads which may complicate the surface.

coryan: Good idea, done.  I switched to excluding files because it was more succint doing so.

leahecole: I also agree with this skiplist feeling a bit odd. +1 to Seth's concerns, but also, how do you plan to query without surfacing the Query method at all? Will you be connecting to the service in another way?

coryan: > I also agree with this skiplist feeling a bit odd. +1 to Seth's concerns, 

Hopefully I addressed those by removing `JobService` completely.

> but also, how do you plan to query without surfacing the Query method at all? Will you be connecting to the service in another way?

Eventually we need to write a veneer. That will have the query APIs, and the details about how that is done will be hidden.  We do not need to expose them in this GAPIC to use them in the veneer.

coryan:
PTAL.

Cargo.toml
coryan: > I also feel iffy with making up an `admin` directory but I think the iffiness is because I want to hear more about why you are leaning in this direction.

Please ignore the directory, it really does not matter.  The only relevant thing is the crate name.

I am not sure there is much more to say: this is a crate that will **only** contain admin operations for BigQuery, using a subset of the full API.  The name is intended to match its contents.

> I think the iffiness for me is that for other services that have the notion of an admin API (thinking of Bigtable primarily) their admin API is a separate endpoint/service entirely, where BigQuery's is not.

That is true for some services (Bigtable as you say, Firestore, and Storage), but not others: Pub/Sub and Spanner use the same endpoint, Spanner and Storage with different gRPC services.

Pub/Sub traditionally used hand-crafted clients that only exposed the admin APIs, nowadays we try to automatically generate them with "selective GAPIC" or language specific features. I believe that is the model that better matches BigQuery.

With all of that said: the question is how do we organize the packages to offer a good SDK with minimal effort.  How we do that may vary by service.

For BigQuery, my plan was to create two crates, one with all the services and RPCs where the generator produces a good enough SDK and one with the services and RPCs where we need hand-crafted code to make the SDK usable.  I am open to different names for these crates, I thought `*-bigquery-admin-v2` and `*-bigquery` are good enough, but I am happy to hear other options.

dbolduc (approved):

Issue #1549, opened 2025-03-19
Create working generated library for Firestore


coryan:
We want a generated, but working, client library for Firestore.  We will not release this library (we want some helpers to make it more usable). We just want to have a end-to-end prototype of a client library using gRPC as its transport.  This will serve as an example for the other (partially) hand-crafted libraries we want to generate.

It should also help in converting the client libraries that must use gRPC as their transport from HTTP+JSON.

PR #1550, opened 2025-03-19
impl(firestore): reorganize generated code

coryan:
Move the code around so we can introduce the `client`, `builders`, and
other modules at the same level as `model`. We want the same (relative)
structure as in a purely generated client, this is the first step
towards that goal.

Part of the work for #1549


src/firestore/src/generated/gapic/mod.rs
```diff
@@ -0,0 +1,19 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// Code generated by sidekick. DO NOT EDIT.
+
+/// The messages and enums that are part of this client library.
+#[allow(clippy::module_inception)]
```
dbolduc, line 18: image

coryan: 🤣

PR #1554, opened 2025-03-20
impl(auth): introduce TokenCache

sai-sunder-s:
The token cache has a background task that keeps refreshing in the background before the token expires, irrespective of whether the token is used.

We can optimize in future to refresh only if the token is used in the past hour.

### DIFF
--- a/src/auth/src/token_cache.rs
+++ b/src/auth/src/token_cache.rs
@@ -334,7 +334,7 @@ mod test {
         rx.changed().await.unwrap();
 
         // Validate that the refresh loop tried getting new token almost immediately
-        assert!(Instant::now() < now + Duration::from_millis(50));
+        assert!(Instant::now() <= now + Duration::from_millis(500));
 
         let actual = rx.borrow().clone().unwrap().unwrap();
         assert_eq!(actual, token2.clone());
--- a/src/auth/src/token_cache.rs
+++ b/src/auth/src/token_cache.rs
@@ -15,8 +15,12 @@
 use crate::Result;
 use crate::token::{Token, TokenProvider};
 use tokio::sync::watch;
-use tokio::time::{Duration, Instant, sleep, sleep_until};
+use tokio::time::{Duration, Instant, sleep};
 
+// Different MDS(Metadata Service) backends have different policies to
+// determine when to refresh a token. Most MDS' refresh token 5 mins before
+// expiry, except for Serverless which refresh tokens 4 mins before
+// expiry. So we are using 4 mins as the staleness limit for our refresh logic.
 const NORMAL_REFRESH_SLACK: Duration = Duration::from_secs(240);
 const SHORT_REFRESH_SLACK: Duration = Duration::from_secs(10);
 
@@ -92,19 +96,23 @@ where
         match token_result {
             Ok(new_token) => {
                 if let Some(expiry) = new_token.expires_at {
-                    let time_until_expiry = expiry.duration_since(Instant::now().into_std());
+                    let time_until_expiry =
+                        expiry.checked_duration_since(Instant::now().into_std());
 
-                    if time_until_expiry > NORMAL_REFRESH_SLACK {
-                        let time_to_sleep =
-                            Instant::now() + time_until_expiry - NORMAL_REFRESH_SLACK;
-                        sleep_until(time_to_sleep).await;
-                    } else if time_until_expiry > SHORT_REFRESH_SLACK {
-                        // If expiry is less than 4 mins, try to refresh every 10 seconds
-                        // This is to handle cases where MDS **repeatedly** returns about to expire tokens
-                        sleep(SHORT_REFRESH_SLACK).await;
-                    } else {
-                        // We were given a token that is expired, or expires in less than 10 seconds.
-                        // We will immediately restart the loop, and fetch a new token.
+                    match time_until_expiry {
+                        None => {
+                            // We were given a token that is expired, or expires in less than 10 seconds.
+                            // We will immediately restart the loop, and fetch a new token.
+                        }
+                        Some(time_until_expiry) => {
+                            if time_until_expiry > NORMAL_REFRESH_SLACK {
+                                sleep(time_until_expiry - NORMAL_REFRESH_SLACK).await;
+                            } else if time_until_expiry > SHORT_REFRESH_SLACK {
+                                // If expiry is less than 4 mins, try to refresh every 10 seconds
+                                // This is to handle cases where MDS **repeatedly** returns about to expire tokens.
+                                sleep(SHORT_REFRESH_SLACK).await;
+                            }
+                        }
                     }
                 } else {
                     // If there is no expiry, the token is valid forever, so no need to refresh
--- a/src/auth/src/token_cache.rs
+++ b/src/auth/src/token_cache.rs
@@ -1,0 +1,530 @@
+use crate::token::{Token, TokenProvider};
+use std::sync::RwLock;
+use std::time::{Duration, Instant};
+use tokio::time::{sleep, sleep_until};
+use tokio::sync::watch;
+use crate::Result;
+
+
+const FIVE_MINUTES: Duration = Duration::from_secs(300);
+const TEN_SECONDS: Duration = Duration::from_secs(10); 
+
+#[derive(Debug)]
+pub(crate) struct TokenCache<T>
+where
+    T: TokenProvider + Send + Sync + 'static
+{
+    rx_token: watch::Receiver<Option<Result<Token>>>,
+    _phantom: std::marker::PhantomData<T>,
+}
+
+impl<T: TokenProvider> Clone for TokenCache<T> {
+    fn clone(&self) -> TokenCache<T> {
+        TokenCache {
+            rx_token: self.rx_token.clone(),
+            _phantom: std::marker::PhantomData,
+        }
+    }
+}
+
+impl<T> TokenCache<T>
+where
+    T: TokenProvider + Send + Sync + 'static
+{
+    pub(crate) fn new(inner: T) -> Self {
+        let (tx_token, rx_token) = watch::channel::<Option<Result<Token>>>(None);
+
+        tokio::spawn(async move {
+            refresh_task(inner, tx_token).await;
+        });
+
+        Self {
+            rx_token,
+            _phantom: std::marker::PhantomData,
+        }
+    }
+
+    async fn get_token_from_channel(&self, mut rx_token: watch::Receiver<Option<Result<Token>>>) -> Result<Token> {
+        rx_token.changed().await.unwrap();
+        let token_result = rx_token.borrow().clone();
+
+        if let Some(token_result) = token_result {
+            return token_result;
+        }
+        else {
+            panic!("There should always be a token or error in the channel after changed()")
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl<T> TokenProvider for TokenCache<T>
+    where 
+        T: TokenProvider + Send + Sync + 'static
+{
+    async fn get_token(&self) -> Result<Token> {
+        let mut rx = self.rx_token.clone();
+        let token_result = rx.borrow_and_update().clone();
+
+        if let Some(token_result) = token_result {
+            match token_result {
+                Ok(token) => match token.expires_at {
+                        None => return Ok(token),
+                        Some(e) => {
+                            if e < tokio::time::Instant::now().into_std() {
+                                // Expired token, wait for refresh
+                                return self.get_token_from_channel(rx).await
+                            } else {
+                                // valid token
+                                return Ok(token)
+                            }
+                        }
+                    },
+                // An error in the result is still a valid result to propagate to the client library
+                Err(e) => Err(e),
+            }
+        }
+        else {
+            self.get_token_from_channel(rx).await
+        }
+    }
+}
+
+async fn refresh_task<T>(
+    token_provider: T,
+    tx_token: watch::Sender<Option<Result<Token>>>,
+)
+where
+    T: TokenProvider + Send + Sync + 'static
+{
+    loop {
+        let token_result = token_provider.get_token().await;
+
+        let _ = tx_token.send(Some(token_result.clone()));
+
+        match token_result {
+            Ok(new_token) => {
+                if let Some(expiry) = new_token.expires_at {
+                    let time_until_expiry = expiry.duration_since(Instant::now());
+
+                    if time_until_expiry > FIVE_MINUTES {
+                        let time_to_sleep = Instant::now() + time_until_expiry - FIVE_MINUTES;
+                        sleep_until(time_to_sleep.into()).await;
+                    } else if time_until_expiry > TEN_SECONDS {
+                        let time_to_sleep = Instant::now() + time_until_expiry - TEN_SECONDS;
+                        sleep_until(time_to_sleep.into()).await;
+                    }
+                } else {
+                    // Handle case where expiry is not available
+                    // TODO: Validate that all auth backends provide expiry and make expiry not optional.
+                    sleep(FIVE_MINUTES).await;
+                }
+            }
+            Err(_) => {
+                // The retry policy has been used already by the inner token provider.
+                // If it ended in an error, just quit the background task.
+                break;
+            }
+        }
+    }
+}
+
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::errors::CredentialError;
+    use crate::token::test::MockTokenProvider;
+    use std::ops::{Add, Sub};
+    use std::sync::{Arc, Mutex};
+    use std::time::Duration;
+
+    static TOKEN_VALID_DURATION: Duration = Duration::from_secs(3600);
+
+    #[tokio::test]
+    async fn initial_token_success() {
+        let expected = Token {
+            token: "test-token".to_string(),
+            token_type: "Bearer".to_string(),
+            expires_at: None,
+            metadata: None,
+        };
+        let expected_clone = expected.clone();
+
+        let mut mock = MockTokenProvider::new();
+        mock.expect_get_token()
+            .times(1)
+            .return_once(|| Ok(expected_clone));
+
+        let cache = TokenCache::new(mock);
+        let actual = cache.get_token().await.unwrap();
+        assert_eq!(actual, expected);
+
+        // Verify that we use the cached token instead of making a new request
+        // to the mock token provider.
+        let actual = cache.get_token().await.unwrap();
+        assert_eq!(actual, expected);
+    }
+
+    #[tokio::test]
+    async fn initial_token_failure() {
+        let mut mock = MockTokenProvider::new();
+        mock.expect_get_token()
+            .times(1)
+            .returning(|| Err(CredentialError::non_retryable_from_str("fail")));
+
+        let cache = TokenCache::new(mock);
+        assert!(cache.get_token().await.is_err());
+
+        // Verify that a new request is made to the mock token provider when we
+        // don't have a valid token.
+        assert!(cache.get_token().await.is_err());
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn expired_token_success() {
+        let now = Instant::now();
+
+        let initial = Token {
+            token: "initial-token".to_string(),
+            token_type: "Bearer".to_string(),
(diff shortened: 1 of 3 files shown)


src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,538 @@
...
+            get_token_from_channel(rx).await
+        }
+    }
+}
+
+async fn get_token_from_channel(
+    mut rx_token: watch::Receiver<Option<Result<Token>>>,
+) -> Result<Token> {
+    // We have already checked the current token in the channel, wait for new one
+    rx_token.changed().await.unwrap();
```
dbolduc, line 97: https://docs.rs/tokio/latest/tokio/sync/watch/struct.Receiver.html#method.changed

```rs
// This method returns an error if and only if the `Sender` is dropped.
```

hmm.. I think we need to handle the error instead of using `unwrap()`, because there is a race.

Consider:

1. we get an expired token on L68

https://github.com/googleapis/google-cloud-rust/blob/3dd8abdebc54ea40d4173eaa451be10607e7c989/src/auth/src/token_cache.rs#L68

2. the background refresh task hits some error, and shuts down, dropping `tx_token`.

https://github.com/googleapis/google-cloud-rust/blob/3dd8abdebc54ea40d4173eaa451be10607e7c989/src/auth/src/token_cache.rs#L135

3. We wait for the next token on L77

https://github.com/googleapis/google-cloud-rust/blob/3dd8abdebc54ea40d4173eaa451be10607e7c989/src/auth/src/token_cache.rs#L77

I think the `unwrap()` will panic in this case.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,538 @@
...
+}
+
+// TODO 1552: Use the token cache in all creds
+#[allow(dead_code)]
+impl<T> TokenCache<T>
+where
+    T: TokenProvider + Send + Sync + 'static,
+{
+    pub(crate) fn new(inner: T) -> Self {
+        let (tx_token, rx_token) = watch::channel::<Option<Result<Token>>>(None);
```
dbolduc, line 48: comment: coryan was right about `watch::channel()` being a good construct for this use case.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,538 @@
...
+        // Wait for the N token requests to complete, verifying the returned token.
+        for task in tasks {
+            let actual = task.await.unwrap();
+            assert!(actual.is_ok(), "{}", actual.err().unwrap());
+            assert_eq!(actual.unwrap(), token);
+        }
+
+        let calls = tp.calls();
+        // Only one call to token provider should have been made
+        assert_eq!(calls, 1);
```
dbolduc, line 507: This assumes that the N tasks launch within the 50ms allotted. I would not rely on timing like that.

This is what I would do, but maybe you thought the expectation was comically loose? I thought it worked.

https://github.com/dbolduc/google-cloud-rust/blob/607315bca3477f8a478fe27e8ec1d8cc1faee6bd/src/auth/src/token_cache.rs#L465-L472

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,538 @@
...
+
+                    if time_until_expiry > FIVE_MINUTES {
+                        let time_to_sleep = Instant::now() + time_until_expiry - FIVE_MINUTES;
+                        sleep_until(time_to_sleep).await;
+                    } else if time_until_expiry > TEN_SECONDS {
+                        // If expiry is less than 5 mins, wait till last 10 seconds to refresh
+                        // This is to handle cases where MDS returns about to refresh tokens
+                        let time_to_sleep = Instant::now() + time_until_expiry - TEN_SECONDS;
+                        sleep_until(time_to_sleep).await;
+                    }
```
dbolduc, line 125: so if we get a token that expires in < 10 seconds, we immediately restart the loop and ask for another token. Makes sense.

Can you add a comment that makes this obvious?

```
} else {
  // We were given a token that is expired, or expires in less than 10 seconds. We will immediately restart the loop, and fetch a new token.
}
```

dbolduc: Observation: if a backend keeps giving us expired tokens, this background task will chew CPU.

Do we care?

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,538 @@
...
+}
+
+async fn refresh_task<T>(token_provider: T, tx_token: watch::Sender<Option<Result<Token>>>)
+where
+    T: TokenProvider + Send + Sync + 'static,
+{
+    loop {
+        let token_result = token_provider.get_token().await;
+
+        let _ = tx_token.send(Some(token_result.clone()));
```
dbolduc, line 110: Don't worry about it in this PR, but a future improvement to think about:

We might send a token to the channel with an expiry earlier than the last token sent to the channel. It would be nice to avoid this.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,538 @@
...
+#[allow(dead_code)]
+impl<T> TokenCache<T>
+where
+    T: TokenProvider + Send + Sync + 'static,
+{
+    pub(crate) fn new(inner: T) -> Self {
+        let (tx_token, rx_token) = watch::channel::<Option<Result<Token>>>(None);
+
+        tokio::spawn(async move {
+            refresh_task(inner, tx_token).await;
```
dbolduc, line 51: Observation: this starts fetching tokens at credential construction time. I think that is probably a good feature. The first token could be available without waiting.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,538 @@
...
+                        sleep_until(time_to_sleep).await;
+                    }
+                } else {
+                    // Handle case where expiry is not available
+                    // TODO 1553: Validate that all auth backends provide expiry and make expiry not optional.
+                    sleep(FIVE_MINUTES).await;
+                }
+            }
+            Err(_) => {
+                // The retry policy has been used already by the inner token provider.
```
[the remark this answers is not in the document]
dbolduc: We talked offline. The retry loop will be in the inner token provider, so the comment is accurate.

Resolving this...

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,538 @@
...
+                // If it ended in an error, just quit the background task.
```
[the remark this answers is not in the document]
dbolduc: We decided this is fine. We can avoid the case with reasonable defaults.

We will clearly document the expected behavior of credential refreshing so the auth library is predictable.

Resolving...

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,538 @@
...
+                    } else if time_until_expiry > TEN_SECONDS {
+                        // If expiry is less than 5 mins, wait till last 10 seconds to refresh
+                        // This is to handle cases where MDS returns about to refresh tokens
+                        let time_to_sleep = Instant::now() + time_until_expiry - TEN_SECONDS;
+                        sleep_until(time_to_sleep).await;
+                    }
+                } else {
+                    // Handle case where expiry is not available
+                    // TODO 1553: Validate that all auth backends provide expiry and make expiry not optional.
+                    sleep(FIVE_MINUTES).await;
```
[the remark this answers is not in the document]
sai-sunder-s: Good point. I missed that API key also uses same Token struct.

src/auth/src/token_cache.rs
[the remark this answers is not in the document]
sai-sunder-s: Synced offline. We will quit if retries are done.

src/auth/src/token_cache.rs
sai-sunder-s: Its an extreme edge case to keep receiving expired tokens I believe. And we would not be constantly retrying, we would have backoff in the retry policy, so it should be ok.

dbolduc: > we would have backoff in the retry policy

There is no backoff for successful requests.

src/auth/src/token_cache.rs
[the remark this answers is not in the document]
dbolduc: > Maybe it should not?

That is a good point.

If we end up with something like a `trait HeaderAdder` in gax, then we will just need `create_api_key_credentials()` and `create_access_token_credentials()` to return a type that implements that trait.

The API key creds would not need to know anything about `Token`s.

src/auth/src/token_cache.rs
dbolduc: This convinced me that I was wrong about there being a race condition with the code as written.

```rs
use tokio::sync::watch;
use tokio::time::{sleep, Duration};

#[derive(Clone)]
struct TokenCache {
    rx_token: watch::Receiver<String>,
}

impl TokenCache {
    fn new() -> Self {
        let (tx_token, rx_token) = watch::channel::<String>("expired".into());

        tokio::spawn(async move {
            refresh_task(tx_token).await;
        });

        Self {
            rx_token,
        }
    }

    async fn get_token(&self) -> String {
        let mut rx = self.rx_token.clone();
        let token = rx.borrow_and_update().clone();

        // Sleep for 2 seconds, to make sure this gets scheduled after the refresh task exits
        sleep(Duration::from_secs(2)).await;

        if token == "expired" {
            get_token_from_channel(rx).await
        } else {
            token
        }
    }
}

async fn get_token_from_channel(
    mut rx_token: watch::Receiver<String>,
) -> String {
    rx_token.changed().await.unwrap();
    rx_token.borrow().clone()
}

async fn refresh_task(tx_token: watch::Sender<String>)
{
    // sleep 1 second.
    sleep(Duration::from_secs(1)).await;

    // simulate an error
    let _ = tx_token.send("error".into());

    println!("refresh_task is dropped");
}

#[tokio::main]
async fn main() {
    // create the cache, spawning the refresh task.
    let c1 = TokenCache::new();
    let c2 = c1.clone();
    
    let h1 = tokio::spawn(async move { 
        let token = c1.get_token().await;
        println!("T1 token: {token}");
    });

    let h2 = tokio::spawn(async move {
        // stagger the second thread
        sleep(Duration::from_millis(500)).await;

        let token = c2.get_token().await;
        println!("T2 token: {token}");
    });

    h1.await.unwrap();
    h2.await.unwrap();
}
```

Output:
```
refresh_task is dropped
T1 token: error
T2 token: error
```

---

It looks like the cloned `rx_token`, has its own `changed()` state. I didn't realize that.

sai-sunder-s: Thank you for verifying. Yes every rx clone maintains its own `changed` state, but the `changed()` [documentation](https://github.com/tokio-rs/tokio/blob/0ec4d0db4d24af74cea40cf0ceed6b7a0a5ff183/tokio/src/sync/watch.rs#L712) states there will be error is sender is dropped.

My debugging also yielded similar result as yours. Was going to use thread::park() as a precaution. Lets keep code as it is right now.

src/auth/src/token_cache.rs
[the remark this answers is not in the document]
sai-sunder-s: Yeah we could do something like that.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,540 @@
...
+            Ok(new_token) => {
+                if let Some(expiry) = new_token.expires_at {
+                    let time_until_expiry = expiry.duration_since(Instant::now().into_std());
+
+                    if time_until_expiry > FIVE_MINUTES {
+                        let time_to_sleep = Instant::now() + time_until_expiry - FIVE_MINUTES;
+                        sleep_until(time_to_sleep).await;
+                    } else if time_until_expiry > TEN_SECONDS {
+                        // If expiry is less than 5 mins, wait till last 10 seconds to refresh
+                        // This is to handle cases where MDS returns about to refresh tokens
```
coryan, line 121: I think:

```suggestion
                        // This is to handle cases where MDS returns about to expire tokens
```

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,540 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::token::{Token, TokenProvider};
+use crate::Result;
+use tokio::sync::watch;
+use tokio::time::{sleep_until, Duration, Instant};
+
+const FIVE_MINUTES: Duration = Duration::from_secs(300);
+const TEN_SECONDS: Duration = Duration::from_secs(10);
```
coryan, line 21: More semantically meaningful names would be useful, such as:

```suggestion
const NORMAL_REFRESH_SLACK: Duration = Duration::from_secs(300);
const SHORT_REFRESH_SLACK: Duration = Duration::from_secs(10);
```

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,540 @@
...
+
+#[derive(Debug)]
+pub(crate) struct TokenCache<T>
+where
+    T: TokenProvider + Send + Sync + 'static,
+{
+    rx_token: watch::Receiver<Option<Result<Token>>>,
+    _phantom: std::marker::PhantomData<T>,
+}
```
coryan, line 30: I think you can simplify this to:

```suggestion
#[derive(Debug)]
pub(crate) struct TokenCache
{
    rx_token: watch::Receiver<Option<Result<Token>>>,
}
```

and then use the generic `T` parameter only in `new()`...

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,540 @@
...
+    }
+}
+
+// TODO(#1552): Use the token cache in all creds
+#[allow(dead_code)]
+impl<T> TokenCache<T>
+where
+    T: TokenProvider + Send + Sync + 'static,
+{
+    pub(crate) fn new(inner: T) -> Self {
```
coryan, line 47: If you agree with my simplification, this becomes:

```suggestion
impl TokenCache
{
    pub(crate) fn new<T>(inner: T) -> Self
    where
        T: TokenProvider + Send + Sync + 'static,
    {
```

src/auth/src/token_cache.rs
coryan: It happens from time to time. By accident if nothing else.

src/auth/src/token_cache.rs
coryan, line 121:
```suggestion
                        // This is to handle cases where MDS **repeatedly** returns about to refresh tokens
```

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,540 @@
...
+                if let Some(expiry) = new_token.expires_at {
+                    let time_until_expiry = expiry.duration_since(Instant::now().into_std());
+
+                    if time_until_expiry > FIVE_MINUTES {
+                        let time_to_sleep = Instant::now() + time_until_expiry - FIVE_MINUTES;
+                        sleep_until(time_to_sleep).await;
+                    } else if time_until_expiry > TEN_SECONDS {
+                        // If expiry is less than 5 mins, wait till last 10 seconds to refresh
+                        // This is to handle cases where MDS returns about to refresh tokens
+                        let time_to_sleep = Instant::now() + time_until_expiry - TEN_SECONDS;
```
coryan, line 122: This is very odd, if we normally think that starting 5 minutes earlier is a good idea, then why wait until there are only 10 seconds left?  maybe we should just sleep for 10 seconds in this case?  Or maybe do some kind of exponentially shorter sleeps (1/2 of the remaining time) if the expiration is within 5 minutes?

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,540 @@
...
+                        sleep_until(time_to_sleep).await;
+                    }
+                    else{
+                        // We were given a token that is expired, or expires in less than 10 seconds. We will immediately restart the loop, and fetch a new token.
+                    }
+                } else {
+                    // If there is no expiry, the token is valid forever, so no need to refresh
+                    // TODO(#1553): Validate that all auth backends provide expiry and make expiry not optional.
+                    break;
```
coryan, line 131: The code coverage says you are missing a test for this case, is that true?

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,540 @@
...
+            assert_eq!(actual.unwrap(), token);
+        }
+
+        let calls = tp.calls();
+        // Only one call to token provider should have been made
+        assert_eq!(calls, 1);
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
+    async fn no_initial_token_thundering_herd_failure_shares_error() {
```
coryan, line 513: I am not sure this is a useful test, it seems like you are testing `tokio::sync::watch` more than your own code?

coryan (changes requested):

src/auth/src/token_cache.rs
sai-sunder-s: That line is covered by `initial_token_success` case.

src/auth/src/token_cache.rs
sai-sunder-s: It also does test that we do not do any call to the token provider other than through the refresh task and also that multiple threads work fine with our implementation.

src/auth/src/token_cache.rs
sai-sunder-s: updated as per offline discussion.

coryan (approved):

coryan:
I fiddled with the title of this PR, it seemed too repetitive.

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,580 @@
...
+{
+    loop {
+        let token_result = token_provider.get_token().await;
+
+        let _ = tx_token.send(Some(token_result.clone()));
+
+        match token_result {
+            Ok(new_token) => {
+                if let Some(expiry) = new_token.expires_at {
+                    let time_until_expiry = expiry.duration_since(Instant::now().into_std());
```
coryan, line 95: I should have noticed this earlier. Here you  need: https://doc.rust-lang.org/std/time/struct.Instant.html#method.checked_duration_since

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,580 @@
...
+
+                    if time_until_expiry > NORMAL_REFRESH_SLACK {
+                        let time_to_sleep =
+                            Instant::now() + time_until_expiry - NORMAL_REFRESH_SLACK;
+                        sleep_until(time_to_sleep).await;
```
coryan, line 100: Simplify to:

```suggestion
                        sleep(time_until_expiry - NORMAL_REFRESH_SLACK).await;
```

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,580 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::Result;
+use crate::token::{Token, TokenProvider};
+use tokio::sync::watch;
+use tokio::time::{Duration, Instant, sleep, sleep_until};
+
+const NORMAL_REFRESH_SLACK: Duration = Duration::from_secs(240);
```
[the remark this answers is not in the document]
sai-sunder-s: Added comment

src/auth/src/token_cache.rs
```diff
@@ -0,0 +1,594 @@
...
+
+        // Wait to make sure the first token expires.
+        sleep(Duration::from_millis(10)).await;
+
+        tokio::spawn(async move {
+            refresh_task(mock, tx).await;
+        });
+
+        // Give the refresh task a chance to run
+        sleep(Duration::from_millis(100)).await;
```
coryan, line 338: These are race conditions and flakes waiting to happen. Can we write this test using `start_paused = true`?

coryan (changes requested):

src/auth/src/token_cache.rs
sai-sunder-s: Removed one of the sleeps. Added it while debugging the issue (turned out to be somehow windows cannot give an instant in the past but windows and mac can). 

I do not think start-pasued=true is of use here because we do not want forward time. We want to make sure not much time passes between refreshes i.e. we are running the loop in refresh task without any sleep.

coryan: Regardless, any test with a time in it is risky.  Maybe you will get lucky, but flaky tests are put out of their misery fast in this project.  I will not hesitate to remove or disable the test.

coryan (approved):

Issue #1539, opened 2025-03-19
Implement `google-cloud-gax-internal` design


coryan:
After some internal discussion we decided to rename `google-cloud-http-client` to `google-cloud-gax-internal`. Beyond renaming the library, we need to make some changes.

- [x] Move the `ReqwestClient` to a submodule (`http`?), making room for a `GrpcClient`. Re-export struct.
- [x] Change the generated code to use  `gaxi::http::ReqwestClient`
- [x] Remove the re-export
- [x] Add features (`_internal_http_client`, `_internal_grpc_client`, `_internal_common`) at first, do nothing with them
- [x] Change the CI builds to test the code with different combinations of these features
- [x] Change the generated code to enable the `_internal_http_client` feature
- [x] Hide `http::ReqwestClient` and `query_parameters` behind this feature
- [x] Hide `path_parameter` behind the `_internal_common` feature

PR #1555, opened 2025-03-20
impl(gax-internal): add placeholders for features

coryan:
gRPC-based clients will use a different subset of this library vs. the
HTTP+JSON-based clients. More importantly, they will use different
dependencies. We will want to enable these dependencies based on what
features are used. This PR introduces placeholders for the features, and
improves the CI to test the different combinations.

Part of the work for #1539

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -43,18 +43,19 @@ jobs:
       - name: Display rustc version
         run: rustc --version
       - name: Test gax with features disabled
-        run: cargo test --package google-cloud-gax
+        run: cargo clean && cargo test --package google-cloud-gax
       - name: Test gax-internal with no features
-        run: cargo test --package google-cloud-gax-internal --no-default-features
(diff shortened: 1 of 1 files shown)


.github/workflows/sdk.yaml
```diff
@@ -43,8 +43,19 @@ jobs:
       - name: Display rustc version
         run: rustc --version
       - name: Test gax with features disabled
-        run: cargo test --package google-cloud-gax
-      - run: cargo test --workspace
+        run: cargo clean && cargo test --package google-cloud-gax
+      - name: Test gax-internal with no features
+        run: cargo clean && cargo test --package google-cloud-gax-internal --no-default-features
```
dbolduc, line 48: As of now, there are no default features, so these cases are equivalent.

Consider removing the `--no-default-features` check until we add a default feature. (Which is hopefully never).

dbolduc: D'oh, I see now that one package is `google-cloud-gax` and the other is `google-cloud-gax-internal`. Ignore me.

Issue #1559, opened 2025-03-20
Update MSRV to 1.85


coryan:
While we do not have an exact date for the GA (or even the first release with stable APIs) we expect by the time we do release the 1.85 Rust version will not be very new.

Using 1.85 allows us to move to the 2024 edition, and use async closures and other goodies.

coryan:
Fixed in #1584

PR #1560, opened 2025-03-20
chore: update builds to Rust 1.85

coryan:
Build with Rust 1.85 and Rust 1.85.1. We will update the second as new
releases are created. I expect the first to be more stable.

Part of the work for #1559


coryan:
I think we should make the version names symbolic so the job names don't change on each upgrade. And then we can make the branch protections stable too.  I will fix that first.

dbolduc (approved):

Issue #1541, opened 2025-03-19
Review our APIs against the Rust API guidelines checklist


coryan:
We should compare our APIs against the [Rust API Guidelines Checklist](https://rust-lang.github.io/api-guidelines/checklist.html) and:

## Document any deviations that we do not want to fix and why

2. **Interoperability** *(crate interacts nicely with other library functionality)*

- [x] Data structures implement Serde's `Serialize`, `Deserialize` ([C-SERDE](https://rust-lang.github.io/api-guidelines/interoperability.html#c-serde))
  - The motivation is that types are "saveable".
  - Note that our message types have this property.
  - I cannot think of another type where this would be useful. e.g. saving a `ClientConfig` does not seem that useful. And if we do it, we would have to think about backwards compatibility for the serialization/deserialization. ew.

6. **Flexibility** *(crate supports diverse real-world use cases)*

- [x] Traits are object-safe if they may be useful as a trait object ([C-OBJECT](https://rust-lang.https://github.com/googleapis/googleapisgithub.io/api-guidelines/flexibility.html#c-object))
  - docs.rs will call out the traits that are not dyn-compatible. So we are at least predictable.
    - https://docs.rs/google-cloud-kms-v1/latest/google_cloud_kms_v1/stubs/trait.Autokey.html#dyn-compatibility
  - I do not think there is much utility to having these as trait objects though.
  - Applications can use the client structs, not the stub traits.

8. **Dependability** *(crate is unlikely to do the wrong thing)*

- [ ] Functions validate their arguments ([C-VALIDATE](https://rust-lang.github.io/api-guidelines/dependability.html#c-validate))
  - It is infeasible/impossible to verify that a request is valid for any particular service on the client side.
  - We could add newtypes for resources. e.g. a `Project` which knows its name is `"projects/<project>"`.
    - These expand the API surface, opening us up to more breaks
    - I think we would keep our request builders accepting strings for resource names.
    - I think we would not accept these new types, because that might be too restrictive.
    - We could add them later
  - Some fields can be validated on the client side.
    - Tracked in #520

10. **Future proofing** *(crate is free to improve without breaking users' code)*

- [x] Structs have private fields ([C-STRUCT-PRIVATE](https://rust-lang.github.io/api-guidelines/future-proofing.html#c-struct-private))
  - The flexibility gains in changing representations come at the cost of usability for the structs. We do not think this trade off is worth it (at least for our generated messages, many of which are used in responses from the server).
[...]

PR #1589, opened 2025-03-21
cleanup(wkt)!: hide `wkt::generated` module

dbolduc:
Noticed in #1541 

https://docs.rs/google-cloud-wkt/latest/google_cloud_wkt/generated/index.html

We re-export these types, and only ever call them by `wkt::Api` (for example).

We should not be exposing the `generated` module which is an implementation detail.

Technically a breaking change, because `wkt::generated::Api` is no longer a thing.


coryan (approved):
We need to bump the major version then, we can do that in a future PR if you prefer.

dbolduc:
> bump the major version

Done.

coryan (approved):

PR #1599, opened 2025-03-24
fix(generator): load mixin dependencies

### LINKED ISSUE #1598: unable to generate `google.ai.generativelanguage.v1`
I'm encountering an issue when trying to generate `google.ai.generativelanguage.v1`. My sidekick.toml file is:

```toml
[general]
specification-source = 'google/ai/generativelanguage/v1'
service-config = 'google/ai/generativelanguage/v1/generativelanguage_v1.yaml'
```

And the issue I'm seeing is:

```
error refreshing directory generated/google_cloud_generativelanguage:
  cannot find output type .google.protobuf.Empty for method .google.ai.generativelanguage.v1.ModelService.CancelOperation
```

This is from api/xref.go, `CrossReference()` - before the backends get a chance to run. I've tried to debug but wasn't able to make progress. It is interesting that it's not able to resolve `.google.protobuf.Empty`.

I don't believe that `generativelanguage` is one of the API we generally intend to generate, but it would be useful to me - it is one of the APIs that allows using an api key, and, there is a hand-written Dart implementation of it that we could compare w/ the generated output.

coryan:
Fixes #1598

At least I can generate the code for Rust, and it was failing before.

### DIFF
--- a/generator/internal/parser/protobuf_mixin_test.go
+++ b/generator/internal/parser/protobuf_mixin_test.go
@@ -208,6 +208,93 @@ func TestProtobuf_OperationMixin(t *testing.T) {
 	})
 }
 
+func TestProtobuf_OperationMixinNoEmpty(t *testing.T) {
+	var serviceConfig = &serviceconfig.Service{
+		Name:  "test.googleapis.com",
+		Title: "Test API",
+		Documentation: &serviceconfig.Documentation{
+			Summary:  "Used for testing generation.",
+			Overview: "Test Overview",
+			Rules: []*serviceconfig.DocumentationRule{
+				{
+					Selector:    "google.longrunning.Operations.GetOperation",
+					Description: "Custom docs.",
+				},
+				{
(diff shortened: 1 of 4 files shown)


generator/internal/parser/protobuf_mixin_test.go
```diff
@@ -208,6 +208,93 @@ func TestProtobuf_OperationMixin(t *testing.T) {
...
+					Selector: "google.longrunning.Operations.CancelOperation",
+					Pattern: &annotations.HttpRule_Delete{
+						Delete: "/v2/{name=operations/*}",
+					},
+					Body: "*",
+				},
+			},
+		},
+	}
+	test := makeAPIForProtobuf(serviceConfig, newTestCodeGeneratorRequest(t, "test_noempty_mixin.proto"))
```
dbolduc, line 256: comment for myself: This is where we generate the proto under `testdata/`

dbolduc (approved):

PR #1602, opened 2025-03-24
test(firestore): add basic integration tests

### LINKED ISSUE #1549: Create working generated library for Firestore

coryan:
Since this is the first gRPC-based client some integration tests seem
appropriate. I just tested basic CRUD stuff, pagination and delete are probably
the most interesting.

Fixes #1549

### DIFF
--- a/.gcb/builds/resources/main.tf
+++ b/.gcb/builds/resources/main.tf
@@ -16,6 +16,10 @@ variable "project" {
   type = string
 }
 
+variable "region" {
+  type = string
+}
+
 # Create a bucket to cache build artifacts.
 resource "google_storage_bucket" "build-cache" {
   name          = "${var.project}-build-cache"
@@ -44,6 +48,14 @@ resource "google_storage_bucket" "build-cache" {
       type = "Delete"
     }
   }
+}
+
+# Create a Firestore database for the integration tests.
+resource "google_firestore_database" "default" {
+  project     = var.project
+  name        = "(default)"
+  location_id = "us-central1"
+  type        = "FIRESTORE_NATIVE"
 }
 
 output "build-cache" {
--- a/src/integration-tests/src/firestore.rs
+++ b/src/integration-tests/src/firestore.rs
@@ -1,0 +1,162 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use firestore::Error;
+use firestore::Result;
+use firestore::client::Firestore;
+use firestore::model;
+use rand::{Rng, distr::Alphanumeric};
+
+pub const COLLECTION_ID_LENGTH: usize = 32;
+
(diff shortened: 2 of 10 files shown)


src/integration-tests/src/firestore.rs
```diff
@@ -0,0 +1,162 @@
...
+            &collection_id,
+        )
+        .paginator()
+        .await
+        .items();
+    while let Some(doc) = documents.next().await {
+        let doc = doc?;
+        println!("  ITEM = {doc:?}");
+    }
+    println!("SUCCESS on list_documents: {response:?}");
```
dbolduc, line 92: this is the `response` from `get_document()`. Intentional? or do we want:

```
    println!("SUCCESS on list_documents:");
```

src/integration-tests/src/firestore.rs
```diff
@@ -0,0 +1,162 @@
...
+
+    let response = client
+        .update_document(
+            model::Document::new()
+                .set_name(&document_name)
+                .set_fields([("greeting", model::Value::new().set_string_value("Goodbye."))]),
+        )
+        .set_update_mask(model::DocumentMask::new().set_field_paths(["greeting"]))
```
dbolduc, line 100: comment: it would be cool if the GAPIC computed a default update mask for the application.

and `.set_update_mask(...)` can be used to override it if necessary (hopefully not).

https://google.aip.dev/134

src/integration-tests/src/firestore.rs
```diff
@@ -0,0 +1,162 @@
...
+        if let Some(true) = doc.create_time.map(|v| v >= stale_deadline) {
+            continue;
+        }
+        if let Some(integration) = doc.fields.get("integration-test") {
+            if let Some(&true) = integration.get_boolean_value() {
+                stale_documents.push(doc.name);
+            }
+        }
+    }
+    let stale_documents = stale_documents;
```
dbolduc, line 149: ?

src/integration-tests/src/firestore.rs
```diff
@@ -0,0 +1,162 @@
...
+    let pending = stale_documents
+        .iter()
+        .map(|name| client.delete_document(name).send())
+        .collect::<Vec<_>>();
+
+    futures::future::join_all(pending)
+        .await
+        .into_iter()
+        .zip(stale_documents)
+        .for_each(|(r, name)| println!("{name} = {r:?}"));
```
dbolduc, line 159: maybe "Deleting stale document: {name} = {r:?}"

.gcb/builds/resources/main.tf
```diff
@@ -46,6 +50,14 @@ resource "google_storage_bucket" "build-cache" {
   }
 }
 
+# Create a Firestore database for the integration tests.
```
dbolduc, line 53: Do we need instructions for creating the database here:

https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-set-up-development-environment.md#one-time-set-up

dbolduc (approved):

src/integration-tests/src/firestore.rs
coryan: that removes the `mut`.

.gcb/builds/resources/main.tf
coryan: Done.

src/integration-tests/src/firestore.rs
dbolduc: Ack

PR #1605, opened 2025-03-24
cleanup!: hide request builder impls

dbolduc:
Noticed in #1541 

These types only exist to simplify the implementation. They do not need to be public.

https://docs.rs/google-cloud-kms-v1/latest/google_cloud_kms_v1/builders/autokey/struct.RequestBuilder.html

I am too lazy to call this a break. The only thing lost are the typenames. These things had no public fields/methods. Meh. :shrug:

### DIFF
--- a/.sidekick.toml
+++ b/.sidekick.toml
@@ -30,7 +30,7 @@ protobuf-subdir             = "src"
 [codec]
 # The default version for all crates. This can be overridden in the crate's
 # `.sidekick.toml` file.
-version = "0.3.0"
+version = "0.2.0"
 # The default release level for all crates.
 release-level = "preview"
 # Disable a number of warnings.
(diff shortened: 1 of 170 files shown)


.sidekick.toml
```diff
@@ -30,7 +30,7 @@ protobuf-subdir             = "src"
 [codec]
 # The default version for all crates. This can be overridden in the crate's
 # `.sidekick.toml` file.
-version = "0.2.0"
+version = "0.3.0"
```
coryan: The current release is 0.1.0. We already bumped the version number.  I do not think we need to bump it again.

There are also references to this version number starting on line 60.  If we decide to update the number here, we have to update those numbers too.

dbolduc: D'oh, I messed up my commits. This change was reverted in the "generated" commit.

there are no changes to `.sidekick.toml`.

coryan (approved):

.sidekick.toml
dbolduc: PTAL.

PR #1607, opened 2025-03-25
impl(auth): introduce caching for user credentials

sai-sunder-s:
For user credentials, use the token cache introduced in #1554

### DIFF
--- a/src/auth/src/credentials/user_credential.rs
+++ b/src/auth/src/credentials/user_credential.rs
@@ -524,9 +524,8 @@ mod test {
             refresh_token: "test-refresh-token".to_string(),
             endpoint: endpoint,
         };
-        let cached_token_provider = TokenCache::new(token_provider);
         let uc = UserCredential {
-            token_provider: cached_token_provider,
+            token_provider,
             quota_project_id: None,
         };
         let now = std::time::Instant::now();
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/user_credential.rs
```diff
@@ -512,8 +524,9 @@ mod test {
             refresh_token: "test-refresh-token".to_string(),
             endpoint: endpoint,
         };
+        let cached_token_provider = TokenCache::new(token_provider);
```
dbolduc, line 527: optional: I am not sure this is adding anything to the test other than distraction. I might revert these changes here and elsewhere.

src/auth/src/credentials/user_credential.rs
```diff
@@ -561,6 +575,13 @@ mod test {
         let token = cred.get_token().await?;
         assert_eq!(token.token, "test-access-token");
 
+        let token = cred.get_token().await?;
+        assert_eq!(token.token, "test-access-token");
+
+        // Test that the inner token provider was called only
```
dbolduc, line 581: nit: I might rename this test to `creds_from_json_custom_uri_with_caching()`

dbolduc (approved):

src/auth/src/credentials/user_credential.rs
sai-sunder-s: The value it adds is that we validate whether all types of success and error responses get propagated correctly out of the token cache.

(But that is something ideally tested in the token cache itself, which we are doing. So I agree, let me revert these to directly token provider)

Issue #1135, opened 2025-02-11
Consider renaming "polling policy" to "polling error policy"


coryan:
While writing the user guide I noticed that it is awkward to speak of the "polling backoff policy" and the "polling policy".

PR #1608, opened 2025-03-25
refactor!: rename polling policy

coryan:
The policy is all about handling polling errors, renaming to
`PollingErrorPolicy` makes the intent clearer, and more consistent with
the names of other policies.

Part of the work for #1135


coryan:
The backwards compat stuff is not well tested.  I will remove it in a couple of followup PRs.

Issue #1612, opened 2025-03-25
Implement common `grpc::Client`


coryan:
The generated code for gRPC-based clients has a lot of code that does not depend on the specific service: the generated client creates auth headers, implements the retry loop, adds the telemetry headers, etc.

We should refactor this code to `google-cloud-gax-internal`, where it can be reused and tested.

- [x] Need to convert `tonic::Status` errors to our types
- [x] Need to implement retry loop
- [x] Need to implement timeouts per RPC
- [x] Need to set `idempotency` from the generated code
- [x] `grpc::Client` should use `gax::Response`
- [x] Need to change the generated code to use this client
- [ ] #1881

dbolduc:
Closing, as the only outstanding work is blocked externally. It is tracked in its own issue now.

PR #1613, opened 2025-03-25
impl(gax-internal): introduce `grpc::Client`

coryan:
This type will be used by all gRPC-based client libraries and will implement
common functionality, such as timeouts and retry loops.

Part of the work for #1612


src/auth/src/credentials.rs
```diff
@@ -386,13 +386,41 @@ pub mod testing {
             None
         }
     }
+
+    /// A simple credentials implementation to use in tests.
+    ///
+    /// This always returns an error, the appliAlways returns a "Bearer" token, with "test-only-token" as the value.
```
dbolduc, line 392: check this.

dbolduc (approved):

src/auth/src/credentials.rs
coryan: Fixed.

PR #1621, opened 2025-03-26
impl(gax): introduce a generic retry loop

coryan:
Introduce a generic retry loop.


src/gax/src/retry_loop.rs
```diff
@@ -0,0 +1,187 @@
...
+        LoopState::Permanent(e) | LoopState::Exhausted(e) => Err(e),
+        LoopState::Continue(_e) => {
+            backoff(backoff_delay).await;
+            Ok(())
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
```
dbolduc, line 103: comment: At first, I thought that the tests should move to this interface, but I think its fine that the tests are in `src/gax-internal/tests/retry_loop.rs`

src/gax/src/lib.rs
```diff
@@ -54,3 +54,7 @@ pub mod polling_backoff_policy;
 pub mod polling_error_policy;
 pub mod retry_policy;
 pub mod retry_throttler;
+
+#[cfg(feature = "unstable-sdk-client")]
+#[doc(hidden)]
+pub mod retry_loop;
```
dbolduc, line 60: optional: say `internal` somewhere in addition to hiding this from the docs?

dbolduc: Also FWIW, `unstable-sdk-client` should already be hidden from the docs. It doesn't hurt to also have `#[doc(hidden)]` though

dbolduc (approved):

src/gax/src/retry_loop.rs
coryan: I think we should refactor the tests, but we can do that later.  For now, we have good coverage and the initial test is a good start.

src/gax/src/lib.rs
coryan: Ack on the documentation.
I changed the name to `retry_loop_internal`. This may belong on a separate crate, but I would like to finish all the refactoring around `auth` before figuring that out.

PR #1634, opened 2025-03-27
fix(wkt): non-negative timestamp nanos

coryan:
The timestamp nanos are supposed to be non-negative. For timestamps
before the epoch:

> Negative second values with fractions must still have non-negative
> nanos values that count forward in time.

I took the opportunity to add some missing test cases.


src/wkt/src/timestamp.rs
```diff
@@ -271,6 +271,9 @@ impl TryFrom<&str> for Timestamp {
         let nanos_since_epoch = odt.unix_timestamp_nanos();
         let seconds = (nanos_since_epoch / NS) as i64;
         let nanos = (nanos_since_epoch % NS) as i32;
+        if !(0..Self::NS).contains(&nanos) {
```
dbolduc, line 274: `if nanos < 0` might be more clear?

src/wkt/src/timestamp.rs
```diff
@@ -271,6 +271,9 @@ impl TryFrom<&str> for Timestamp {
...
+            return Timestamp::new(seconds - 1, Self::NS + nanos);
```
dbolduc, line 275: nit: `Self::NS` is just called `NS` above.

dbolduc (approved):

src/wkt/src/timestamp.rs
dbolduc: Oh, they are different types. Ignore me.

coryan: I got confused myself.

src/wkt/src/timestamp.rs
coryan: Fixed.

Issue #1609, opened 2025-03-25
Proposal: "seal" common traits


dbolduc:
## Background

Sealing traits is a way to prevent downstream crates from implementing traits, thus avoiding some breaks. https://rust-lang.github.io/api-guidelines/future-proofing.html#c-sealed

We have traits in common crates. Some of these are implementation conveniences for our GAPICs. Some of them provide public functions for customers to use.

In several cases, it is ideal for *our crates* (`google-cloud-*`) to implement these traits, but not for external applications. It would be ideal to enforce this somehow. This proposal discusses how.

### Category A

Public traits for applications to implement.

- `gax::backoff_policy::BackoffPolicy`
- `gax::retry_policy::RetryPolicy`
- etc.

Nothing happens to these. We do not want to seal them. We will just note that we cannot add a new method on these traits without providing a default implementation. That is ok.

- `lro::Poller`

This one is a bit more nuanced. If we seal this trait, then applications cannot create a fake `Poller` in their tests. Seems like we need to leave it open.

### Category B

Traits that are implementation details. No public APIs are exposed to customers via these traits. Customers would never need to implement the traits for their own types. Our client crates need to implement these traits.

- `wkt::message::Message`
- `wkt::prost::Convert`
- `gax::options::RequestBuilder`
- `gax::paginator::PageableResponse`

Most importantly, we should hide these, and any `impl`s for these with `#[doc(hidden)]`.

I think we should also move them into an `internal` module. (e.g. `wkt::internal::Message`).

Alternatives:
- We could choose to gate these behind an `internal` feature
  - We would have to gate the code that depends on the trait too. (e.g. `wkt::Any`). That seems unhelpful.
- We could also choose to make new internal crates (e.g. `wkt-internal`), like we do for gax.
  - I do not think this is worth it. I think no docs, plus `internal` in the name is good enough.

### Category C

Traits that offer public APIs for applications, that we do not want applications to implement. We want our crates to implement these.

- `gax::options::RequestOptionsBuilder`

We will want to seal this trait, so we can add new APIs to it. We can use an internal-only type to seal this. In this specific case, `RequestBuilder` is the perfect type to use.

```rs
pub trait RequestOptionsBuilder: internal::RequestBuilder { ... }

impl<T> RequestOptionsBuilder for T
where T: internal::RequestBuilder { ... }
```

[...]

PR #1637, opened 2025-03-27
cleanup(gax)!: seal `RequestOptionsBuilder`; hide `RequestBuilder`

dbolduc:
Part of the work for #1609 

This seals `RequestOptionsBuilder` (which is public), and hides `RequestBuilder` which was an implementation detail.

To hide `RequestBuilder`, we introduce an `internal` module which is not for applications to use. This is enforced via documentation.

Sealing the trait means that applications cannot implement it. We can add methods to the trait without fear of breaking applications.

`RequestOptionsBuilder` can only be implemented if an application utters `internal::RequestBuilder`, which they are not supposed to do.

---

A Builder still has its functions from the `RequestOptionsBuilder` blanket implementation:

image

But it does not have `request_options()` from `RequestBuilder`:

image

For comparison, here are the current docs: https://docs.rs/google-cloud-secretmanager-v1/latest/google_cloud_secretmanager_v1/builders/secret_manager_service/struct.ListLocations.html

### DIFF
--- a/generator/internal/rust/templates/crate/src/builders.rs.mustache
+++ b/generator/internal/rust/templates/crate/src/builders.rs.mustache
@@ -181,7 +181,8 @@ pub mod {{Codec.ModuleName}} {
         {{/InputType.OneOfs}}
     }
 
-    impl gax::options::RequestBuilder for {{Codec.BuilderName}} {
+    #[doc(hidden)]
+    impl gax::options::internal::RequestBuilder for {{Codec.BuilderName}} {
         fn request_options(&mut self) -> &mut gax::options::RequestOptions {
             &mut self.0.options
         }
--- a/src/gax/src/options/mod.rs
+++ b/src/gax/src/options/mod.rs
@@ -163,7 +163,7 @@ impl RequestOptions {
 /// These builders can be used to set the request parameters, e.g., the name of
 /// the resource targeted by the RPC, as well as any options affecting the
 /// request, such as additional headers or timeouts.
-pub trait RequestOptionsBuilder {
+pub trait RequestOptionsBuilder : internal::RequestBuilder {
     /// If `v` is `true`, treat the RPC underlying this method as idempotent.
     fn with_idempotency(self, v: bool) -> Self;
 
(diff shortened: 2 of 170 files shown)


src/gax/src/options/mod.rs
```diff
@@ -192,20 +192,25 @@ pub trait RequestOptionsBuilder {
...
-///
-/// This is an implementation detail, most applications have little need to
-/// worry about or use this trait.
-pub trait RequestBuilder {
-    fn request_options(&mut self) -> &mut RequestOptions;
+/// This module contains implementation details. It is not part of the public
+/// API. Types inside may be changed or removed without warnings. Applications
+///  should not use any types contained within.
+#[doc(hidden)]
+pub mod internal {
```
coryan, line 199: Does the module need to be public?  Could you say `mod internal` ?

generator/internal/rust/templates/crate/src/builders.rs.mustache
```diff
@@ -181,7 +181,8 @@ pub mod {{Codec.ModuleName}} {
         {{/InputType.OneOfs}}
     }
 
-    impl gax::options::RequestBuilder for {{Codec.BuilderName}} {
+    #[doc(hidden)]
```
coryan, line 184: Seems like this could be a separate PR 🤷‍♂️

coryan (commented):
Seems like we can do better than just hide `internal` from the docs?

src/gax/src/options/mod.rs
dbolduc: It needs to be public. All of the builders use it.

https://github.com/googleapis/google-cloud-rust/blob/d35df5e8ddb145c0ab0efe78c681ccbf11c11df2/generator/testdata/rust/openapi/golden/src/builders.rs#L106

generator/internal/rust/templates/crate/src/builders.rs.mustache
dbolduc: > Seems like this could be a separate PR 🤷‍♂️

I could have done a separate PR for sealing the `RequestOptionsBuilder` trait:

https://github.com/googleapis/google-cloud-rust/blob/e1fe15d02aa40c3158c8479ee320323a01e53f60/src/gax/src/options/mod.rs#L166

the rest of the PR is about hiding `RequestBuilder` (from the docs)

:shrug:

coryan (approved):

PR #1641, opened 2025-03-28
cleanup(auth): Remove deadcode attribute from token cache

### LINKED ISSUE #1552: Use token cache in all creds
Once token cache PR is merged, use the token cache in all credentials instead of always getting a new token.

sai-sunder-s:
Remove  `allow(dead_code)` from token cache because all credential types now use the cache. Fixes #1552.


dbolduc (approved):
FYI: This is a `cleanup`, not a `fix`.

Fixes will make it into our release notes, cleanups will not.

Issue #1548, opened 2025-03-19
Finalize design for client initialization


coryan:
Today clients can be initialized using either:

```rust
let client = Client::new().await?;
```

or

```rust
let config = gax::options::ClientConfig::new() /* maybe set some stuff */;
let client = Client::new_with_config(config).await?;
```

----

We should decide if this is the structure we want. Alternatives include:

### Minor changes to the function names

For example rename `new_with_config()` to `from_config()`.

### Treat ClientConfig as a builder?

Maybe something like this is possible:

```rust
use google_cloud_secret_manager_v1 as sm;
let builder = gax::options::ClientBuilder::new() /* maybe set some stuff */;
let client = builder.build::<sm::client:SecretManagerService>();
```

Or maybe:

Maybe something like this is possible:

```rust
let builder = gax::options::ClientBuilder::new() /* maybe set some stuff */;
let client :  sm::client:SecretManagerService = builder.build();
```

Or maybe:

```rust
// This creates a `gax::options::ClientBuilder<T>`
let builder = google_cloud_secret_manager_v1::client:SecretManagerService::builder()
   /* maybe configure some things */
   ;
let client : sm::client::SecretManagerService = builder.build().await?;
```

PR #1643, opened 2025-03-28
impl(gax): introduce the generic client builder

coryan:
Introduce some helpers to create clients via builders.

Part of the work for #1548


src/gax/src/client_builder.rs
```diff
@@ -0,0 +1,678 @@
...
+    ///     .build();
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn with_retry_throttler<V: Into<RetryThrottlerArg>>(mut self, v: V) -> Self {
+        self.config.retry_throttler = v.into().0;
+        self
+    }
+
+    /// Configure the polling backoff policy.
```
dbolduc, line 207: s/backoff/error/

src/gax/src/client_builder.rs
```diff
@@ -0,0 +1,678 @@
...
+    pub fn with_polling_error_policy<V: Into<PollingErrorPolicyArg>>(mut self, v: V) -> Self {
+        self.config.polling_error_policy = Some(v.into().0);
+        self
+    }
+
+    /// Configure the polling backoff policy.
+    ///
+    /// Some clients support long-running operations, the client libraries can
+    /// automatically poll these operations until they complete. The polling
+    /// backoff policy controls for how long the client waits between polling
```
dbolduc, line 238: s/for // ?

src/gax/src/client_builder.rs
```diff
@@ -0,0 +1,678 @@
...
+    ///
+    /// [tracing]: https://docs.rs/tracing/latest/tracing/
+    pub fn with_tracing(mut self) -> Self {
+        self.config.tracing = true;
+        self
+    }
+
+    /// Configure the authentication credentials.
+    ///
+    /// Most Google cloud services require authentication, though some services
```
dbolduc, line 106: nit: s/cloud/Cloud/

src/gax/src/client_builder.rs
```diff
@@ -0,0 +1,678 @@
...
+    ///
+    /// ```
+    /// # use google_cloud_gax::client_builder::examples;
+    /// use examples::Client; // Placeholder for examples
+    /// let client = Client::builder()
+    ///     .build();
+    /// ```
+    pub fn build<C>(self) -> C
+    where
+        F: Fn(internal::ClientConfig<Cr>) -> C,
```
dbolduc, line 66: Does this need to be `AsyncFn` ? and all the docs `.await`?

`Client::builder().build()` will call `async fn make_access_token_credential()` at some point.

dbolduc (approved):

src/gax-internal/tests/client_builder.rs
coryan, line 1: I decided to move all the integration tests here.  We need to know this works with the gRPC and Reqwest clients.

src/gax-internal/tests/client_builder.rs
```diff
@@ -0,0 +1,109 @@
...
+            pub fn builder() -> fake_client::Builder {
+                gax::client_builder::internal::new_builder(fake_client::Factory)
+            }
+
+            async fn new(config: gax::options::ClientConfig) -> gax::Result<Self> {
+                let inner = gaxi::grpc::Client::new(config, super::DEFAULT_ENDPOINT).await?;
+                Ok(Self { inner })
+            }
+        }
+        pub(self) mod fake_client {
```
coryan, line 49: I think this is enough to prevent usage, but I want to verify first.

coryan (commented):
Thanks for the review. I am still making changes, maybe we can discuss in detail on Monday.

src/gax-internal/tests/client_builder.rs
coryan: Yes, it works.

src/gax/src/client_builder.rs
coryan: You are absolutely right.  I thought I could just change the function trait later, but it turns out fixing it changes everything else.  This is all solved now.

PR #1656, opened 2025-03-31
cleanup: use `gax-internal` for `ClientOptions`

coryan:
The generated code uses an specific instantiation of `ClientOptions`,
add a definition to `gax-internal` and use that definition. A future
PR will cleanup the definition in `gax::options`. I could have done this
in one PR, but it seems easier to grok if the non-breaking and breaking
changes are separated.

Part of the work for #1548


src/gax-internal/src/options.rs
```diff
@@ -0,0 +1,57 @@
...
+// limitations under the License.
+
+pub use auth::credentials::Credential as Credentials;
+
+/// The client configuration for [crate::http::ReqwestClient] and [crate::grpc::Client].
+pub type ClientConfig = gax::client_builder::internal::ClientConfig<Credentials>;
+
+pub(crate) const LOGGING_VAR: &str = "GOOGLE_CLOUD_RUST_LOGGING";
+
+/// Returns true if the envirclient configuration enables tracing or th
```
dbolduc, line 22: "if the environment or the client configuration enables tracing"

dbolduc (approved):

src/gax-internal/src/options.rs
coryan: Oops, fixed.

PR #1660, opened 2025-03-31
feat(auth): support non-default Service account credentials

harkamaljot:
Add an implementation for service account credentials to allow the user to specify `scopes`, `aud` and `quota_project_id`.

Created a `Builder` struct in service_account.rs file to accept new optional parameters `scopes`, `aud` and `quota_project_id`. While `scopes`, `aud` are passed as JWT claims when a new token is created, `quota project` is passed as header .

Unit Tests
I updated the unit tests to test out the behavior for `scopes`, `aud` and `quota project`. I haven't added a test case where user specifies both  `scopes`, `aud` as that part is already handled and tested by service_account/jws.rs.


src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
+
+impl Builder {
+    /// Sets the [service account key] in [aip/4112] format for this credential.
+    ///
+    /// [aip/4112]: https://google.aip.dev/auth/4112
+    /// [service account key]: https://cloud.google.com/iam/docs/keys-create-delete#creating
+    pub fn service_account_key(mut self, service_account_key: ServiceAccountKey) -> Self {
+        self.service_account_key = service_account_key;
+        self
+    }
```
coryan, line 120: It seems that a builder without a service account key is useless. Should this be:

```suggestion
    /// Creates a new builder using `service_account_key` in [aip/4112] format for this credential.
    ///
    /// [aip/4112]: https://google.aip.dev/auth/4112
    /// [service account key]: https://cloud.google.com/iam/docs/keys-create-delete#creating
    pub fn new(service_account_key: ServiceAccountKey) -> Self {
        Self { service_account_key,  aud: None, scopes: None, quota_project_id: None }
    }
```

src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
+
+    /// Sets the audience for this credential.
+    ///
+    /// `aud` is a [JWT] claim specifying intended recipient(s) of the token,
+    /// that is, a service(s) or resource(s).
+    /// This cannot be used at the same time as the scopes claim.
```
coryan, line 126: This suggests we should change the implementation to be something like:

```
enum ServiceAccountRestrictions {
  Audience(String),
  Scopes(Vec<String>),
}

pub struct Builder {
  service_account_key: ServiceAccountKey,
  restrictions: Option<ServiceAccountRestrictions>,
}
```

And then you can change the implementation to:

```
impl Builder {
    pub fn aud<V: Into<String>>(mut self, v: V) -> Self {
        self.restrictions = ServiceAccountRestrictions::Audience(v.into());
        self
    }
}
```

The comments should mention that setting `aud()` overrides the value set in `scopes()`, but now it is impossible to set both.

src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
+    pub fn aud<S: Into<String>>(mut self, aud: S) -> Self {
+        self.aud = Some(aud.into());
+        self
+    }
+
+    /// Sets the [scopes] for this credentials.
+    ///
+    /// `scopes` is a [JWT] claim specifying requested permission(s) for the token.
+    /// This cannot be used at the same time as the aud claim.
+    /// Multiple scopes can be specified using a single space (" ") as delimiter.
```
coryan, line 139: It is sad that customers need to learn this weird thing. It would be nicer if we consumed `IntoIterator<Item = Into<String>>` and took care of the joining for them.

src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
+
+    /// Sets the [quota project] for this credential.
+    ///
+    /// In some services, you can use a service account in
+    /// one project for authentication and authorization, and charge
+    /// the usage to a different project. This may require that the
+    /// service account has `serviceusage.services.use` permissions on the quota project.
+    ///
+    /// [quota project]: https://cloud.google.com/docs/quotas/quota-project
+    pub fn quota_project_id<S: Into<String>>(mut self, quota_project_id: S) -> Self {
```
coryan, line 156: I should have noticed this earlier.  All the other builders use `with_` or `set_` . It is a bit sad that auth uses a different convention.

src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
-    })
+    let service_account_key =
+        serde_json::from_value::<ServiceAccountKey>(js).map_err(CredentialError::non_retryable)?;
+    Ok(Builder::default()
+        .service_account_key(service_account_key)
+        .build())
+}
+
+#[derive(Default)]
+pub struct Builder {
```
coryan, line 105: The builder type needs documentation. Consider including an example too:

```suggestion
/// Creates service account credentials.
///
/// # Example
/// ```
/// let key = ServiceAccountKey {

/// };
/// let credentials = Builder::new(key).with_aud("https://pubsub.googleapis.com").build();
/// ```
#[derive(Default)]
pub struct Builder {
```

src/auth/src/credentials/service_account.rs
```diff
@@ -12,15 +12,77 @@
...
+//! # use google_cloud_auth::credentials::Credential;
+//! # use google_cloud_auth::errors::CredentialError;
+//! # tokio_test::block_on(async {
+//! let service_account_key = serde_json::from_value::<ServiceAccountKey>(serde_json::json!({
+//! "client_email": "test-client-email",
+//! "private_key_id": "test-private-key-id",
+//! "private_key": "", // <-- Provide valid PKCS#8 PEM key here
+//! "project_id": "test-project-id",
+//! "universe_domain": "test-universe-domain",
+//! })).unwrap();
```
coryan, line 61: This seems like a rather indirect way to initialize the data structure.  A plain `ServiceAccountKey { ... }` seems clear enough.

src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
+        self.service_account_key = service_account_key;
+        self
+    }
+
+    /// Sets the audience for this credential.
+    ///
+    /// `aud` is a [JWT] claim specifying intended recipient(s) of the token,
+    /// that is, a service(s) or resource(s).
+    /// This cannot be used at the same time as the scopes claim.
+    /// The value should be https://{SERVICE}/. (e.g. `https://pubsub.googleapis.com/`)
```
coryan, line 127: So can it be a resource? Can we provide an example of an audience that is a resource? If not, we should change the  description.

src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
+    ///
+    /// [JWT]: https://google.aip.dev/auth/4111
+    pub fn aud<S: Into<String>>(mut self, aud: S) -> Self {
+        self.aud = Some(aud.into());
+        self
+    }
+
+    /// Sets the [scopes] for this credentials.
+    ///
+    /// `scopes` is a [JWT] claim specifying requested permission(s) for the token.
```
coryan, line 137: We should mention something about how scopes are orthogonal to IAM permissions.

src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
-/// for more details.
-#[derive(serde::Deserialize, Builder)]
-#[builder(setter(into))]
-struct ServiceAccountInfo {
-    client_email: String,
-    private_key_id: String,
-    private_key: String,
-    project_id: String,
-    universe_domain: String,
+/// A representation of a [service account key] in the format described by [aip/4112].
```
coryan, line 179: Here and elsewhere, is `aip/4112` the best link for this format?  Seems like the format of a service account key should be documented in IAM ? And it probably gets more attention there?

src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
+    pub client_email: String,
+    /// ID of the service account's private key.
+    pub private_key_id: String,
+    /// The PEM-encoded PKCS#8 private key string associated with the service account.
+    /// Begins with `-----BEGIN PRIVATE KEY-----`.
+    pub private_key: String,
+    /// The project id the service account belongs to.
+    pub project_id: String,
+    /// The universe domain this service account belongs to.
+    pub universe_domain: String,
```
coryan, line 202: Which of these fields are required? Can I create one of these service accounts with some fields set to empty strings?  Seems like this type is complex enough to require its own builder or at least validating constructors?

coryan (changes requested):
Much better, thanks.  I think the builder should be constructed with a valid service account key or not constructed at all.

Likewise, it seems like the service account key could use some thinking about how it is safely initialized.

src/auth/src/credentials/service_account.rs
```diff
@@ -12,15 +12,77 @@
...
+//! [Service Account] Credentials type.
+//!
+//! A service account is an account for an application or compute workload
+//! instead of an individual end user. The recommended practice is to use
+//! Google Default Credentials, which relies on the configuration of the Google
+//! Cloud system hosting your application ([Google Compute Engine (GCE)][gce-link],
+//! [Google Kubernetes Engine (GKE)][gke-link], [Cloud Run]) to authenticate
+//! your workload or application.  But sometimes you may need to create and
+//! download a [service account key], for example, to use a service account
+//! when running your application on a system that is not part of Google Cloud.
```
sai-sunder-s, line 24: If system is not running on google cloud, the recommended way is to use BYOID, not download SA keys.

Instead of explaining the best practice here, we could just link to https://cloud.google.com/iam/docs/best-practices-service-accounts#choose-when-to-use. @coryan is there expectation to have detailed description in the doc comments or can we just point to the public doc?

src/auth/src/credentials/service_account.rs
```diff
@@ -12,15 +12,77 @@
...
+//! format described by [aip/4112]. This string contains an ID for the service
+//! account, as well as the cryptographic materials (an RSA private key)
+//! required to authenticate the caller.
+//!
+//! Therefore, service account keys should be treated as any other secret
+//! with security implications. Think of them as unencrypted passwords. Do not
+//! store them where unauthorized persons or programs may read them.
+//!
+//! The credentials in this module use [Self-signed JWTs] to bypass the
+//! intermediate step of exchanging client assertions for OAuth tokens.
```
sai-sunder-s, line 39: You can remove this. Reader of the docs need not know about the old exchange flow.

src/auth/src/credentials/service_account.rs
sai-sunder-s: It is not too late to change. Other than this one, we currently have only one other builder I believe (and that may not even be released yet).

src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
-    universe_domain: String,
+/// A representation of a [service account key] in the format described by [aip/4112].
+///
+/// This type is typically created by
+/// deserializing the JSON key data, for example, when the service account key
+/// is obtained from [Cloud Secret Manager] or a similar service.
+/// This key can then be used to create a [Credential], by passing it to the [Builder].
+///
+/// [aip/4112]: https://google.aip.dev/auth/4112
+/// [Cloud Secret Manager]: https://cloud.google.com/secret-manager/docs
```
sai-sunder-s, line 187: Why do we have to mention about cloud secret manager

src/auth/src/credentials/service_account.rs
sai-sunder-s: Universe domain is optional. Need to validate the rest.

src/auth/src/credentials/service_account.rs
```diff
@@ -177,7 +332,14 @@ where
         let mut value = HeaderValue::from_str(&format!("{} {}", token.token_type, token.token))
             .map_err(CredentialError::non_retryable)?;
         value.set_sensitive(true);
-        Ok(vec![(AUTHORIZATION, value)])
+        let mut headers = vec![(AUTHORIZATION, value)];
+        if let Some(project) = &self.quota_project_id {
```
sai-sunder-s, line 336: Can you add a todo here to refactor the quota project logic. Right now we have these few lines to add the quota project in multiple places

src/auth/src/credentials/service_account.rs
```diff
@@ -32,33 +94,117 @@ use time::OffsetDateTime;
...
+    scopes: Option<String>,
+    quota_project_id: Option<String>,
+}
+
+impl Builder {
+    /// Sets the [service account key] in [aip/4112] format for this credential.
+    ///
+    /// [aip/4112]: https://google.aip.dev/auth/4112
+    /// [service account key]: https://cloud.google.com/iam/docs/keys-create-delete#creating
+    pub fn service_account_key(mut self, service_account_key: ServiceAccountKey) -> Self {
```
sai-sunder-s, line 117: The developer should be able to just give the json string as input. We should not expect them to build the ServiceAccountKey and give to the builder.

sai-sunder-s (changes requested):

src/auth/src/credentials/service_account.rs
sai-sunder-s: I am not aware of a scenario where a user has to build the Key. It would be sufficient to take the sa json as input and build this object. If the optional and required fields are defined appropriately, serde should take care of the validation for us I believe.

src/auth/src/credentials/service_account.rs
coryan: > @coryan is there expectation to have detailed description in the doc comments or can we just point to the public doc?

I think we should avoid too much detail for things that are explained elsewhere. But we should provide enough detail for people to (1) recall what the concept is, if they are familiar with it, or (2) get enough context to see if they need to read the more detailed description.  Here I would think something like:

```
//! A service account is an account for an application or compute workload
//! instead of an individual end user. The default credentials used by the
//! client libraries may be, and often are, associated with a service account.
//! Therefore, you can use service accounts by configuring your environment,
//! without any code changes.
//! Sometimes they application needs to use a [service account key] directly. 
//! The types in this module will help you in this case. For more information
//! on when service account keys are appropriate, consult the
//! relevant section in the [Best practices for using service accounts] guide.
//!
//! [Best practices for using service accounts]: https://cloud.google.com/iam/docs/best-practices-service-accounts#choose-when-to-use
```

src/auth/src/credentials/service_account.rs
harkamaljot: I would agree, as mds accepts a list so this is definitely confusing however the problem is that (https://google.aip.dev/auth/4112) mentioned that multiple scopes can be specified using single space (" ") as delimiter. So, to keep it consistent with aip I felt better to do the same thing. Should I change into IntoIterator?

src/auth/src/credentials/service_account.rs
harkamaljot: the reason for not adding a builder was because this type wouldn't be created by hand, it should be always created by serde deserialization. Let me update this so that this struct is private(make universe domain optional) and user passes in sa key as json, then serde can do the validation

src/auth/src/credentials/service_account.rs
harkamaljot: The reason I had linked cloud secret manager to give an example where the key can be stored/obtained. As otherwise mentioning about the key, without telling them where it can be obtained from didn't made sense to me.

src/auth/src/credentials/service_account.rs
harkamaljot: IAM docs talk about service account key format here: https://cloud.google.com/iam/docs/keys-create-delete#creating. Looking at the page, the format of the key is embedded in the creating discussion, given a user won't create the keys directly rather just pass in a json, definitely in places like module doc I can update it and reword it accordingly.

src/auth/src/credentials/service_account.rs
```diff
@@ -12,53 +12,240 @@
...
+    /// This can be overridden using [with_aud][Builder::with_aud]
+    /// or [with_scopes][Builder::with_scopes] methods.
+    ///
+    /// [cloud-platform]:https://cloud.google.com/compute/docs/access/service-accounts#scopes_best_practice
+    /// [service_account_key]: https://cloud.google.com/iam/docs/keys-create-delete#creating
+    pub fn new(service_account_key: Value) -> Self {
+        Self {
+            service_account_key,
+            restrictions: ServiceAccountRestrictions::Scopes(
+                DEFAULT_SCOPES.into_iter().map(|s| s.into()).collect(),
```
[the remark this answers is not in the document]
coryan: They are defined as a collection, but to be constant they need to be a collection of `&str`.  I would keep them defined as a single `&str` and then use:

```suggestion
                [ DEFAULT_SCOPE ].map(str::to_string).to_vec(),
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0ce8bc895907112d7fea6ada4f507da1

src/auth/src/credentials/service_account.rs
```diff
@@ -12,53 +12,242 @@
...
+        }
+    }
+
+    /// Sets the audience for this credential.
+    ///
+    /// `aud` is a [JWT] claim specifying intended recipient(s) of the token,
+    /// that is, a service(s).
+    /// Only one of audience or scopes can be specified for a credential.
+    /// Setting the audience will replace any previously configured scopes.
+    /// The value should be https://{SERVICE}/. (e.g. `https://pubsub.googleapis.com/`)
```
coryan, line 164: nit: 
```suggestion
    /// The value should be `https://{SERVICE}/`, e.g., `https://pubsub.googleapis.com/`.
```

src/auth/src/credentials/service_account.rs
```diff
@@ -12,53 +12,242 @@
...
+    ///
```
coryan, line 165: Consider a rustdoc example:

```suggestion
    ///
    /// # Example
    /// ```
    /// # use google_cloud_auth::credentials::service_account::Builder;
    /// fn example(service_account_key: &str) -> Result<Credential> {
    ///     Builder::new(service_account_key).with_aud("https://bigtable.googleapis.com/").build()
    /// }
```

or 

```
    /// ```
    /// # use google_cloud_auth::credentials::service_account::Builder;
    /// let service_account_key = "{ /* add details here */ }";
    /// let credentials = Builder::new(service_account_key).with_aud("https://bigtable.googleapis.com/").build()
```

src/auth/src/credentials/service_account.rs
```diff
@@ -12,53 +12,242 @@
...
+    /// Sets the [scopes] for this credential.
+    ///
+    /// `scopes` is a [JWT] claim specifying requested permission(s) for the token.
+    /// Only one of audience or scopes can be specified for a credential.
+    /// Setting the scopes will replace any previously configured audience.
+    ///
+    /// `scopes` define the *permissions being requested* for this specific session
+    /// when interacting with a service. For example, `https://www.googleapis.com/auth/devstorage.read_write`.
+    /// IAM permissions, on the other hand, define the *underlying capabilities*
+    /// the service account possesses within a system. For example, storage.buckets.delete.
```
coryan, line 181: nit:
```suggestion
    /// the service account possesses within a system. For example, `storage.buckets.delete`.
```

src/auth/src/credentials/service_account.rs
```diff
@@ -12,53 +12,242 @@
...
+    /// IAM permissions are typically more fine-grained than OAuth scopes.
+    ///
+    /// [JWT]: https://google.aip.dev/auth/4111
```
coryan, line 184: As before, a small example would be nice.

src/auth/src/credentials/service_account.rs
```diff
@@ -12,53 +12,242 @@
...
+        self
+    }
+
+    /// Sets the [scopes] for this credential.
+    ///
+    /// `scopes` is a [JWT] claim specifying requested permission(s) for the token.
+    /// Only one of audience or scopes can be specified for a credential.
+    /// Setting the scopes will replace any previously configured audience.
+    ///
+    /// `scopes` define the *permissions being requested* for this specific session
```
coryan, line 178: I think there is a cloud site page that describes when to use scopes vs. IAM permissions, might be useful to reference here.

It also seems like it would be useful to say that scopes can only apply as additional restrictions on the service account permissions.  At least I think that is how they work, if a service account does not have IAM permissions to write then setting the scope to `https://...read_write` won't gain you anything, right?

src/auth/src/credentials/service_account.rs
```diff
@@ -12,53 +12,242 @@
...
+    /// the usage to a different project. This requires that the
+    /// service account has `serviceusage.services.use` permissions on the quota project.
+    ///
+    /// [quota project]: https://cloud.google.com/docs/quotas/quota-project
+    pub fn with_quota_project_id<S: Into<String>>(mut self, quota_project_id: S) -> Self {
+        self.quota_project_id = Some(quota_project_id.into());
+        self
+    }
+
+    /// Returns a [Credential] instance with the configured settings.
```
coryan, line 209: Please say something about what kind of conditions may result in an error here.

coryan (approved):
A few suggestions to improve the docs and some code simplification ideas. Nothing blocking I think.

src/auth/src/credentials/service_account.rs
harkamaljot: yes that is correct, updated the explanation to reflect this.

sai-sunder-s (approved):

PR #1675, opened 2025-04-01
chore: update googleapis circa 2025-04-01

dbolduc:


src/generated/cloud/discoveryengine/v1/src/client.rs
```diff
@@ -2602,6 +2602,156 @@ impl SearchTuningService {
     }
 }
 
+/// Implements a client for the Discovery Engine API.
```
coryan, line 2605: I just realized that we use the same title for all services in the same API.  Something to think about.

coryan (approved):

PR #1677, opened 2025-04-01
ci: fix cargo doc for new library

dbolduc:
For some reason, `cargo doc` goes into modules that are `#[doc(hidden)]`.

This was making the `sidekick rust-generate` command fail because `crate::grpc::Client` does not exist in `gax-internal` when we only enable the HTTP feature:

https://github.com/googleapis/google-cloud-rust/blob/26c98334608cc879fe1457aa0688eb96324afaa2/src/gax-internal/src/options.rs#L17

Try for yourself with:

```sh
env RUSTDOCFLAGS="-D warnings" cargo doc -p google-cloud-tpu-v2
```

My fix is to not build docs for any dependencies in this command, which is sufficient, and a good change anyway. `cargo docs --no-deps` will still fail if we try to link types that don't exist in dependencies.

### DIFF
--- a/generator/internal/sidekick/rust_generate.go
+++ b/generator/internal/sidekick/rust_generate.go
@@ -95,7 +95,7 @@ func rust_generate(rootConfig *config.Config, cmdLine *CommandLine) error {
 		return err
 	}
 	slog.Info("Running `cargo doc` on new client library")
-	if err := runExternalCommand("env", "RUSTDOCFLAGS=-D warnings", "cargo", "doc", "--package", packagez); err != nil {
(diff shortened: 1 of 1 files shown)


coryan (approved):

coryan:
I think we should fix that comment to work when the features are disabled too.  This is why features are such a pain, you never know when you finished testing everything (docs, tests, build) with all the different feature combinations.

dbolduc:
> I think we should fix that comment to work when the features are disabled too.

Yeah, I agree. I am just surprised that `cargo doc` is parsing the comments inside these hidden modules.

PR #1681, opened 2025-04-01
test(integration-test): error detail deserialization

coryan:
I tested and this failed before the changes in #1680


src/integration-tests/tests/driver.rs
```diff
@@ -78,6 +78,16 @@ mod driver {
             .map_err(report)
     }
 
+    #[test_case(ta::client::TelcoAutomation::builder().with_tracing(); "with tracing enabled")]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn run_telco_automation(
```
dbolduc, line 83: Consider calling this `error_details`. That is what is under test. `telcoautomation` is an implementation detail.

src/integration-tests/src/lib.rs
```diff
@@ -16,6 +16,7 @@ use gax::error::Error;
 pub type Result<T> = std::result::Result<T, gax::error::Error>;
 pub mod firestore;
 pub mod secret_manager;
+pub mod telco_automation;
```
dbolduc, line 19: same goes for the module name.

dbolduc (approved):
Nice, thanks

src/integration-tests/tests/driver.rs
coryan: Done.

src/integration-tests/src/lib.rs
coryan: Done.

Issue #1654, opened 2025-03-31
User-guide section on how to write a tests when using a client


dbolduc:

dbolduc:
FYI: I plan to use `mockall` in the guide, even though its version is not `>= 1.0`. It seems to be the most popular mocking framework. I will link [this article](https://blog.logrocket.com/mocking-rust-mockall-alternatives/) discussing alternative frameworks.

PR #1683, opened 2025-04-01
docs: mocking errors

dbolduc:
Part of the work for #1654 

I felt this did not go without saying, so I said it.

### DIFF
--- a/guide/src/mock_a_client.md
+++ b/guide/src/mock_a_client.md
@@ -101,6 +101,18 @@ Finally, we are ready to call our function...
 {{#include ../samples/tests/mocking.rs:validate}}
 ```
 
+### Simulating errors
+
+Simulating errors is no different than simulating successes. We just need to
+modify the result returned by our mock.
+
+```rust,ignore
+{{#include ../samples/tests/mocking.rs:error}}
+```
+
+Note that a client built `from_stub()` does not have an internal retry loop. It
+returns all errors from the stub directly to the application.
+
 ______________________________________________________________________
 
 ## Full program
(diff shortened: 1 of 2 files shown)


guide/src/mock_a_client.md
```diff
@@ -101,6 +101,18 @@ Finally, we are ready to call our function...
...
+Simulating errors is no different than simulating successes. We just need to
+modify the result returned by our mock.
+
+```rust,ignore
+{{#include ../samples/tests/mocking.rs:error}}
+```
+
+Note that a client built `from_stub()` does not have an internal retry loop. It
+returns all errors from the stub directly to the application.
+
```
coryan, line 115: In other sections I link the full function for each sub-section separately. Consider doing the same here.

coryan (approved):

guide/src/mock_a_client.md
dbolduc: > Um, I think I will pass. The tests have shared code, so I do not think the sub-sections would actually be "full".

Strike this. I just realized we can have an anchor in multiple places, and `mdbook` will concatenate the sections. e.g.

```
// ANCHOR: foo
1
// ANCHOR_END: foo
2
3
4
// ANCHOR: foo
5
// ANCHOR_END: foo
```

yields:

```
1
5
```

So I am out of excuses. Will clean this up soon-ish.

dbolduc: Nevermind. The `mdbook` behavior I described is totally unreliable. :(

dbolduc: > totally unreliable

Lol, I finally understand...

`{{#include ... }}`  will not concatenate sections with the same marker.

`{{#rustdoc_include ... }}` will concatenate sections with the same marker.

I was switching between the two unreliably.

PR #1690, opened 2025-04-02
cleanup: flatten module files

coryan:
I find modules consisting of a single `mod.rs` file harder to navigate
than modules with a `${module-name}.rs` file. Having so many files
called `mod.rs` is confusing. I cannot avoid this for generated code,
or in some integration tests.


dbolduc (commented):
> I find modules consisting of a single `mod.rs` file harder to navigate
than modules with a `${module-name}.rs` file.

+1

dbolduc (approved):

PR #1696, opened 2025-04-03
impl: flip `gax` vs. `auth` dependency

### LINKED ISSUE #1542: Swap dependency between `gax` and `auth`
We want to use the retry loop policies in `auth`. But `gax` already depends on `auth` we need to break the cycle and revert the dependencies.

- [x] Copy `CredentialError` to `gax`
- [x] Make the  `pub(crate)` helpers in `auth::errors::CredentialError` into standalone functions.
- [ ] Break the dep from gax to auth.
- [ ] Add a dep from auth to gax
- [ ] Replace `CredentialError` in auth with the one in `gax`

coryan:
The retry policies in `gax` now use `CredentialError` from `gax`. Meanwhile,
`auth::errors::CredentialError` is now a use declaration from `gax`.  This
requires flipping the dependencies between `google-cloud-gax` and
`google-cloud-auth`.

Fixes #1542


src/gax/src/retry_policy.rs
```diff
@@ -725,31 +723,29 @@ mod tests {
         assert!(p.on_error(now, 0, idempotent, error).is_exhausted());
     }
 
-    fn from_status(status: Status) -> Error {
```
dbolduc, line 728: unrelated, but totally fine

dbolduc (approved):

sai-sunder-s (approved):

PR #1698, opened 2025-04-03
ci: better coverage for features

coryan:


.github/workflows/sdk.yaml
```diff
@@ -46,27 +46,77 @@ jobs:
...
+          set -e
+          rustup toolchain install ${{ fromJson(env.GHA_RUST_VERSIONS)[matrix.rust-version] }}
+          rustup default ${{ fromJson(env.GHA_RUST_VERSIONS)[matrix.rust-version] }}
+          rustup component add clippy
+      - run: rustup default ${{ fromJson(env.GHA_RUST_VERSIONS)[matrix.rust-version] }}
+      - name: Display Cargo version
+        run: cargo version
+      - name: Display rustc version
+        run: rustup show active-toolchain -v
+      - name: Test gax with features disabled
```
dbolduc, line 82: nit: "Test crates with features"

.github/workflows/sdk.yaml
```diff
@@ -46,27 +46,77 @@ jobs:
...
+          set -e
+          cargo clean
+          for sub in test doc; do
+            cargo "${sub}" --package google-cloud-wkt --no-default-features
+            cargo "${sub}" --package google-cloud-wkt --no-default-features --features prost
+            cargo "${sub}" --package google-cloud-wkt --no-default-features --features chrono
+            cargo "${sub}" --package google-cloud-wkt --no-default-features --features time
+            cargo "${sub}" --package google-cloud-wkt --all-features
+          done
+          cargo clippy --no-deps --package google-cloud-wkt --all-features -- --deny warnings
```
dbolduc, line 93: Can `clippy` go inside the loop?

coryan: Cannot think of a way to do so.  It needs the `-- --deny warnings` at the end, which the other `sub` commands do not accept.

It may be nicer if there was a way to get all the features of a package, but I cannot find an easy way to do so.

.github/workflows/sdk.yaml
coryan: Fixed.

PR #1718, opened 2025-04-04
cleanup(auth): update MDS credential builder methods

harkamaljot:
**Motivation:**

Align the MDS credential builder methods with the convention used by other builders in the module, primarily by adopting the `with_` prefix for configuration methods.

**Changes:**

* Refactored the MDS credential builder implementation. Renamed the following public methods to use the `with_` prefix for consistency:
    * `endpoint(value)` -> `with_endpoint(value)`
    * `quota_project_id(value)` -> `with_quota_project_id(value)`
    * `universe_domain(value)` -> `with_universe_domain(value)`
    * `scopes(value)` -> `with_scopes(value)`

### DIFF
--- a/src/auth/src/credentials/mds.rs
+++ b/src/auth/src/credentials/mds.rs
@@ -37,7 +37,7 @@
 //! # use google_cloud_auth::credentials::Credentials;
 //! # use google_cloud_auth::errors::CredentialError;
 //! # tokio_test::block_on(async {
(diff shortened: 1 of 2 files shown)


coryan (approved):
Thanks for the detailed commit message and release notes.

I do not think this is a breaking change, as there is no release that includes the previous methods. Consider updating the commit message before merging the changes.

PR #1722, opened 2025-04-05
docs: user guide for mocking LRO errors

### LINKED ISSUE #1661: User-guide section on how to test LROs

dbolduc:
Fixes #1661 

Describe how to simulate different types of LRO errors in tests in the user guide.

Looks like: [mocking_lros_error.pdf](https://github.com/user-attachments/files/19617740/mocking_lros_error.pdf)

### DIFF
--- a/guide/samples/tests/mocking_lros_error.rs
+++ b/guide/samples/tests/mocking_lros_error.rs
@@ -1,0 +1,217 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Examples showing how to simulate LROs in tests.
+
+// ANCHOR: all
+use gax::Result;
+use gax::error::Error;
+use google_cloud_gax as gax;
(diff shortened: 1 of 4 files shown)


guide/src/mocking_lros.md
```diff
@@ -157,6 +157,80 @@ response.
...
+will yield the simplest test.
+
+### Simulating an error starting an LRO
+
+The simplest way to simulate an error is to have the initial request fail with
+an error.
+
+```rust,ignore,noplayground
+{{#rustdoc_include ../samples/tests/mocking_lros_error.rs:expectation-initial}}
+```
```
coryan, line 177: This is subject to the retry policy. You should say something about returning an error that is not retryable.

guide/samples/tests/mocking_lros_error.rs
```diff
@@ -0,0 +1,217 @@
...
+        let mut mock = MockSpeech::new();
+        mock.expect_batch_recognize()
+            .once()
+            .in_sequence(&mut seq)
+            .returning(|_, _| make_partial_operation(25));
+        // ANCHOR: expectation-polling-error
+        mock.expect_get_operation()
+            .once()
+            .in_sequence(&mut seq)
+            .returning(|_, _| Err(Error::other("could not poll operation")));
```
coryan, line 201: I think it would be more interesting to return realistic errors here, like a `ServiceError` with `Code::Aborted` or something.  If you are writing tests that return errors, it is because your code handles realistic errors, we need to feed it realistic errors too. If not, we should explain that we are using `Error::other` to simplify the tests, and that real RPCs return something different.

dbolduc: Done.

guide/src/mocking_lros.md
dbolduc: The stubbed out clients do not have policies. I moved the blurb that says that up to the initial section of the error handling. (it was formerly tucked away in the last of the subsections).

coryan (approved):

PR #1732, opened 2025-04-07
test(wkt): prototype support for unknown fields

coryan:


src/wkt/tests/unknown.rs
```diff
@@ -0,0 +1,87 @@
...
+    #[non_exhaustive]
+    pub struct MessageWithUnknown {
+        #[serde(skip_serializing_if = "String::is_empty")]
+        pub field_0: String,
+
+        #[serde(skip_serializing_if = "Option::is_none")]
+        pub field_1: Option<i32>,
+
+        #[serde(flatten)]
+        _unknown_fields: Option<serde_json::Map<String, serde_json::Value>>,
```
dbolduc, line 69: I think you should just add the generator support now, then use a real message type in the test.

src/wkt/tests/unknown.rs
```diff
@@ -0,0 +1,87 @@
...
+//! This shows how the generator could handle unknown fields in the future.
+
+#[cfg(test)]
+mod test {
+    use serde_json::json;
+
+    #[test]
+    fn roundtrip_from_json() -> anyhow::Result<()> {
+        let input = json!({
+            "field0": "abc123",
```
dbolduc, line 26: might also want a test for a known field which is a message type, and that message has an unknown field.

src/wkt/tests/unknown.rs
coryan: It requires more work.  For `Any` we need to strip the `@type` field somewhere.

src/wkt/tests/unknown.rs
coryan: Done.

src/wkt/tests/unknown.rs
dbolduc: Ack.

dbolduc (approved):
I still prefer to test generated code instead of code that would have been generated on 2025-04-07.

PR #1736, opened 2025-04-07
cleanup(gax)!: prefer "CANCELLED"

dbolduc:
gRPC uses this spelling: https://grpc.io/docs/guides/status-codes/

and so does the `google.rpc.Code` proto: https://github.com/googleapis/googleapis/blob/07c7b132c79656e944a2cc97fb400532834eb6f0/google/rpc/code.proto#L41

(While many other protos in googleapis use the spelling with a single `L`, consistency with the `google.rpc.Code` proto seems most important.)


coryan (approved):
Good catch.

PR #1738, opened 2025-04-07
docs: update client builder

dbolduc:
Clients no longer have a public `new()`. They use `builder().build().await?`.

The link doesn't exist, because we haven't published the docs yet.

### DIFF
--- a/guide/src/initialize_a_client.md
+++ b/guide/src/initialize_a_client.md
@@ -48,8 +48,10 @@ As it is usual with Rust, you must declare the dependency in your
 {{#include ../samples/Cargo.toml:secretmanager}}
 ```
 
-The default initialization is designed to meet the requirements for most
-cases. You create a default client using `new()`:
+Clients are created by calling `build()` on a
+[`ClientBuilder`][gax-client-builder]. The following creates a client with
+default configuration, which is designed to meet requirements for most use
+cases.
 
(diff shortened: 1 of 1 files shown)


guide/src/initialize_a_client.md
```diff
@@ -48,8 +48,10 @@ As it is usual with Rust, you must declare the dependency in your
 {{#include ../samples/Cargo.toml:secretmanager}}
 ```
 
-The default initialization is designed to meet the requirements for most
-cases. You create a default client using `new()`:
+Clients are created by calling `build()` on a
+[`ClientBuilder`][gax-client-builder]. The following creates a client with
+default configuration, which is designed to meet requirements for most use
```
coryan, line 53: Consider:

```suggestion
Clients are created by calling `build()` on a
[`ClientBuilder`][gax-client-builder]. To obtain an appropriate `ClientBuilder` use
`Client::builder()`. The following creates a client with
default configuration, which is designed to meet requirements for most use
```

coryan (approved):
I think we should try to create a release ASAP.

guide/src/initialize_a_client.md
dbolduc: Reworded using some of your suggestion. You might not like what I went with.

coryan (approved):

Issue #1686, opened 2025-04-01
Auth: Refactor get_headers in Credentials


harkamaljot:
The get_headers() function contains nearly identical logic within the credential handling for:

* mds credentials
* service_account credentials
* user_account credentials

Refactor the common logic out of the individual get_headers methods into a shared location. This could be a helper function, a base class method, or a utility function.

chrischiedo:
I believe this was resolved with #1741. I think it can be closed?

cc: @coryan @sai-sunder-s

sai-sunder-s:
That is correct. Closing this. Thank you again for the change @chrischiedo !

PR #1741, opened 2025-04-08
cleanup: refactor headers in credentials

chrischiedo:
This PR is an effort to resolve issue #1686.

While working on this PR, I considered two approaches:

1. Creating a utility function that takes the necessary parameters and returns the required headers
2. Adding a default `headers` implementation to the `CredentialsTrait` trait that has the common logic that can then be shared across the credentials

Option 1 was more straight forward (and easier to implement). Option 2, on the other hand, is a bit more involving (requiring more changes to existing code). Therefore, for this PR, I went with Option 1.

**Note**: I have added a `utils` module, following the structure recommended in the Rust Book [here](https://doc.rust-lang.org/book/ch07-02-defining-modules-to-control-scope-and-privacy.html#defining-modules-to-control-scope-and-privacy). Currently we have only the `headers_util` utility function, but more utility functions could be added when/if needed.


src/auth/src/credentials/user.rs
```diff
@@ -13,10 +13,11 @@
 // limitations under the License.
 
 use crate::credentials::dynamic::CredentialsTrait;
-use crate::credentials::{Credentials, QUOTA_PROJECT_KEY, Result};
+use crate::credentials::{Credentials, Result};
 use crate::errors::{self, CredentialsError, is_retryable};
 use crate::token::{Token, TokenProvider};
 use crate::token_cache::TokenCache;
+use crate::utils::headers_util::build_headers;
```
coryan, line 20: I would flatten these modules to `crate::header_utils::build_headers`.  As long as the module is not public this is enough, and we can always change it.

coryan (commented):
Thanks for the PR! I started most builds and we left some comments. @harkamaljot or @sai-sunder-s may have further comments.

src/auth/src/utils.rs
```diff
@@ -0,0 +1,4 @@
+/// Utility functions to work with Google Cloud authentication [Credentials].
```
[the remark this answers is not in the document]
chrischiedo: Noted. Will fix shortly.

src/auth/src/lib.rs
```diff
@@ -52,3 +52,8 @@ pub(crate) mod token_cache;
 /// A `Result` alias where the `Err` case is
 /// `google_cloud_auth::errors::CredentialsError`.
 pub(crate) type Result<T> = std::result::Result<T, crate::errors::CredentialsError>;
+
+/// Utility functions to work with Google Cloud authentication [Credentials].
+///
+/// [Credentials]: https://cloud.google.com/docs/authentication#credentials
+pub mod utils;
```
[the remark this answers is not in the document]
chrischiedo: My initial thinking was that perhaps the module would end up containing other utility functions that might be useful outside the current crate, hence the `pub` visibility. But, I get your point. I'll change it to `pub(crate)`.

src/auth/src/credentials/user.rs
chrischiedo: This was actually my initial implementation before changing to the current one (for reasons similar to what I've described in my immediate previous comment). Will refactor accordingly. Curious to get thoughts from @harkamaljot and @sai-sunder-s.

sai-sunder-s: +1, we can flatten for now. We can always change it later if we decide to add more utils in future, since it is not public.

src/auth/src/credentials/user.rs
```diff
@@ -135,17 +136,13 @@ where
...
-                HeaderName::from_static(QUOTA_PROJECT_KEY),
-                HeaderValue::from_str(project).map_err(errors::non_retryable)?,
-            ));
-        }
-        Ok(headers)
+        build_headers(
+            &token,
+            &self.quota_project_id,
+            AUTHORIZATION,
+            |token| HeaderValue::from_str(&format!("{} {}", token.token_type, token.token))
```
sai-sunder-s, line 143: Can we add a method in header util that builds the Bearer value and pass in that method here? This same logic is repeated thrice for now and will be done more times in future.

chrischiedo: @sai-sunder-s: Yes, I think this can be refactored further. Let me give it a try.

chrischiedo:
@sai-sunder-s: Here's the updated PR. Lmk if this addresses your feedback/comment.

cc/ @coryan

sai-sunder-s (approved):

sai-sunder-s:
Thank you for the change!

Issue #1510, opened 2025-03-16
Design a solution to return metadata and the response


coryan:
In some use-cases it is useful to get the metadata (headers, client and server IP addresses, type of authentication used, etc.) along with the response. This can include important troubleshooting information.

We need a mechanism to return the metadata with the response.

The following considerations apply:

* What is the common case? Do we want most users to make the call, and then have to discard the metadata?
* What happens when we switch between gRPC and HTTP+JSON as transports? How do we represent metadata for both?
* Do we want to support [http::Extensions](https://docs.rs/http/latest/http/struct.Extensions.html): see [What are Rust's http Extensions](https://blog.adamchalmers.com/what-are-extensions/)
* What about streaming requests?

coryan:
An idea: the `.send()` method returns `Result<T>` where `T` is the return type, maybe a `.with_metadata()` method returns `Result<Wrapper<T>>` where `Wrapper<T>` is like `tonic::Response`: it includes `T` and the metadata headers

coryan:
I am going to open a separate bug to track the full implementation of this design (go/cloud-rust:getting-response-metadata).  We have implemented enough to stabilize the API.

PR #1745, opened 2025-04-08
feat(gax): a wrapper for responses with metadata

coryan:
Part of the work for #1510

### DIFF
--- a/src/gax/src/response.rs
+++ b/src/gax/src/response.rs
@@ -218,7 +218,7 @@ impl<T> Response<T> {
 /// Component parts of a response.
 ///
 /// The response parts, other than the body, consist of just headers. We
-/// anticipate addition fields in the future.
+/// anticipate the addition of new fields over time.
 ///
 /// The headers are used to return gRPC metadata, as well as (unsurprisingly)
 /// HTTP headers.
--- a/src/gax/src/response.rs
+++ b/src/gax/src/response.rs
@@ -300,8 +300,7 @@ mod test {
             http::header::CONTENT_TYPE,
             http::HeaderValue::from_static("application/json"),
         );
-        let parts = Parts::new()
-            .set_headers(headers.clone());
+        let parts = Parts::new().set_headers(headers.clone());
 
         let response = Response::from_parts(parts, "abc123".to_string());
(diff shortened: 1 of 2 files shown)


src/gax/src/response.rs
```diff
@@ -0,0 +1,332 @@
...
+//! # panic!()
+//! }
+//!
+//! # tokio_test::block_on(async {
+//! let response = make_google_service_rpc("my-project").await?;
+//! if let Some(date) = response.headers().get("Date") {
+//!     // do something with the date
+//! }
+//! let resource = response.body();
+//! // do something with
```
dbolduc, line 45: ...

src/gax/src/response.rs
```diff
@@ -0,0 +1,332 @@
...
+/// Represents a Google Cloud service response.
+///
+/// A response from a Google Cloud service consists of a body (potentially the
+/// unit type), and some metadata, currently just headers.
+///
+/// Typically you get a response as the result of making a request via some
+/// client in the Google Cloud client libraries for Rust. You may also create
+/// responses directly when mocking clients for your own tests.
+///
+/// # Examples
```
dbolduc, line 74: this is a dup of the above section. Remove

src/gax/src/response.rs
```diff
@@ -0,0 +1,332 @@
...
+    /// ```
+    pub fn into_body(self) -> T {
+        self.body
+    }
+}
+
+/// Component parts of a response.
+///
+/// The response parts, other than the body, consist of just headers. We
+/// anticipate addition fields in the future.
```
dbolduc, line 221: s/addition/the addition/

src/gax/src/response.rs
coryan: It is intentional. The generated documentation for the module and the type look too bare otherwise.  I think the duplication is consistent with the API checklist for Rust, and with other crates, e.g.: https://docs.rs/http/1.3.1/http/response/index.html

dbolduc: Apologies, I missed that these were different blocks:

https://github.com/googleapis/google-cloud-rust/blob/8cdec5d362548101c81481568dc8d42618aa33ef/src/gax/src/response.rs#L63-L65

PR #1764, opened 2025-04-10
Prototype(auth): Introduce builder for adc

sai-sunder-s:
Replacing `create_access_token_credentials()` function with a builder. Benefit is it allows user to add on more things like quota, scopes and more options like retry policy, backoff policy etc.

(A downside in introducing retry policy related options here is that SA creds do not do network calls, so these options are a no-op, but should be ok)


coryan:
Seems like #1797 was inspired by this. Closing this prototype, feel free to reopen if I made a mistake.

PR #1765, opened 2025-04-10
feat(gax-internal): return `gax::response::Response`

coryan:
Include the metadata (currently just headers) with each response. Change
the generator to only use the body.

Part of the work for #1510


dbolduc (approved):
All good, my comments are just random observations.

PR #1774, opened 2025-04-11
cleanup(generator): stop using golden files

coryan:
We have enough generated code that we can validate the generator directly with that code, the golden files are redundant.  I left the integration tests in place, but they generate the files to a temporary directory and just verify the generator does not crash and produces


dbolduc (approved):

devoncarew (commented):
👍

PR #1779, opened 2025-04-11
refactor: move `api_header` module to `gax-internal`

coryan:
This is only used in the implementation of the clients, no reason to have it in the public crate, even if we said it was internal / unstable.


src/gax/src/api_header.rs
```diff
@@ -12,82 +12,3 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// Generated libraries create one static instance of this struct and use it
```
dbolduc, line 15: Delete the file

dbolduc (approved):

src/gax/src/api_header.rs
coryan: Fixed, thanks.

Issue #34, opened 2024-10-30
handle nested path params


codyoss:
Current implementation is naive and assumes path arguments are always fields directly on the request object, but these can be deeply nested. 

https://github.com/googleapis/google-cloud-rust/blob/77720dee3887853cc65bf31c90ce887b6e629976/generator/src/genclient/language/internal/golang/golang.go#L137-L145

PR #1781, opened 2025-04-12
fix: field masks as query parameters

### LINKED ISSUE #736: Investigate correct encoding for FieldMask as query parameter
My best reading of the comments (since we do not have a spec) is that the encoding should include the `.paths` field:

> // Note that fields which are mapped to URL query parameters must have a
> // primitive type or a repeated primitive type or a non-repeated message type.
> // In the case of a repeated type, the parameter can be repeated in the URL
> // as `...?param=A&param=B`. In the case of a message type, **each field of the
> // message** is mapped to a separate parameter, such as
> // `...?foo.a=A&foo.b=B&foo.c=C`.

(emphasis mine, [original](https://github.com/googleapis/googleapis/blob/3776db131e34e42ec8d287203020cb4282166aa5/google/api/http.proto#L114-L119)).

I may be wrong, and we should just send the concatenated paths.

coryan:
Sadly, my reading of the spec was wrong. Query parameters cannot be sent
using their custom JSON serialization. They must send each path as any
other serialized object.

I expanded an existing integration test to test a field mask query
parameter with multiple fields, which immediately failed. Then I fixed
the code generator, its tests and then the integration test passed
again.

Fixes #736

### DIFF
--- a/generator/internal/rust/codec.go
+++ b/generator/internal/rust/codec.go
@@ -625,6 +625,16 @@ func addQueryParameter(f *api.Field) string {
 		}
 		return fmt.Sprintf(`let builder = builder.query(&[("%s", &req.%s.value())]);`, f.JSONName, fieldName)
 	case api.MESSAGE_TYPE:
+		if f.TypezID == ".google.protobuf.FieldMask" {
+			// `FieldMask` (and other well-known types) are special. Their JSON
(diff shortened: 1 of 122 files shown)


dbolduc (approved):

dbolduc:
I think fieldmasks are just one (common) case of the more general: https://github.com/googleapis/google-cloud-rust/issues/34

PR #1797, opened 2025-04-14
feat(auth): support non-default access token credentials

harkamaljot:
**Description**

This change introduces support for constructing credentials using explicitly provided source (JSON), allowing users more control beyond the standard Application Default Credentials (ADC) lookup flow. It adds a primary `credentials::Builder` as the main entry point.

A subsequent PR will remove the `create_access_token_credentials` function and migrate its usages to this new builder pattern.

Key implementation details:
* Introduced an `config_builder` macro to centralizes credential construction logic when a quota project id, scopes are given, reducing code duplication in `credential.rs`.
* Added a builder pattern in `credential.rs`. This builder accepts credential sources like JSON file or ADC, and allows users to explicitly set `scopes` and `quota_project_id`.
* Updated the `build()` method of the internal `mds::Builder` to return `Result<Credentials>` (instead of `Credentials`) for uniformity with other builders.  Although that is a breaking change but since there are no users of the builder directly outside auth crate right now, so not marking it as breaking change.

**Testing**

* Added unit tests in `credentials.rs` to verify that scopes are correctly configured via the macro.
* Verified `quota_project_id` `scopes` handling within unit tests in `credentials.rs`.
* Added an test (in `tests/`) that uses to validate that the `Builder` can be used from the outside module.


src/auth/src/credentials.rs
```diff
@@ -285,6 +485,63 @@ enum AdcContents {
...
+                })?
+                .as_str()
+                .ok_or_else(|| {
+                    errors::non_retryable_from_str(
+                        "Failed to parse Credentials Json. `type` field is not a string.",
+                    )
+                })?;
+            match cred_type {
+                "authorized_user" => {
+                    configure_and_build(user_account::Builder::new(json), quota_project_id, scopes)
```
coryan, line 530: I see why you introduced the `AccessTokenCredentialsBuilder` trait.  A few notes:

You can introduce the trait in this file and provide the `impl AccessTokenCredentialsBuilder for Blah` in this file too. That keeps the weirdness more local.

I am not sure you saved yourself that much code, yes, you can refactor to `configure_and_build()` but at the cost of a trait and an impl per builder.

You could write the code like this: 

```rust
  let b = user_account::Builder::new(json);
  let b = quota_project_id.into_iter().fold(b, |b, v| b.with_quota_project_id(v));
  let b = scopes.into_iter().fold(b, |b, v| b.with_scopes(v));
```

without any traits.  That seems like less code per credential type overall.  You can also use a macro to reduce the code even further. This is probably not going to work, but maybe gives you a start:

```rust
macro_rules! config_builder {
    ($t: ty) => { // Maybe you need some extra
        let b = $t::new(json);
        let b = quota_project_id.into_iter().fold(b, |b, v| b.with_quota_project_id(v));
        let b = scopes.into_iter().fold(b, |b, v| b.with_scopes(v));
        b
   }
}

// And then
// config_builder!(user_account::Builder)
```

coryan (changes requested):

src/auth/src/credentials.rs
```diff
@@ -206,6 +216,213 @@ pub(crate) mod dynamic {
...
+    /// Sets the [scopes] for these credentials.
+    ///
+    /// `scopes` define the *permissions being requested* for this specific access token
+    /// when interacting with a service. For example, `https://www.googleapis.com/auth/devstorage.read_write`.
+    /// IAM permissions, on the other hand, define the *underlying capabilities*
+    /// this credential possesses within a system. For example, `storage.buckets.delete`.
+    /// When a token generated with specific scopes is used, the request must be permitted
+    /// by both the this credential's underlying IAM permissions and the scopes requested
+    /// for the token. Therefore, scopes act as an additional restriction on what the token
+    /// can be used for.
```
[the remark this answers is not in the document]
harkamaljot: sure sounds good, will send another PR after this one, to update the description at other places.

src/auth/src/credentials.rs
```diff
@@ -206,6 +216,213 @@ pub(crate) mod dynamic {
...
+/// Common use cases where using ADC would is useful include:
+/// - Your application is deployed to a Google Cloud environment such as
+///   [Google Compute Engine (GCE)][gce-link],
+///   [Google Kubernetes Engine (GKE)][gke-link], or [Cloud Run]. Each of these
+///   deployment environments provides a default service account to the
+///   application, and offers mechanisms to change this default service account
+///   without any code changes to your application.
+/// - You are testing or developing the application on a workstation (physical or
+///   virtual). These credentials will use your preferences as set with
+///   [gcloud auth application-default]. These preferences can be your own GCP
```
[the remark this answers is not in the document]
harkamaljot: @coryan I'm assuming you meant "Google Cloud" instead of GCP here.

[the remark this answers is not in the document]
coryan: LOL, yes.

src/auth/src/credentials.rs
```diff
@@ -285,6 +485,63 @@ enum AdcContents {
...
+                "authorized_user" => {
+                    configure_and_build(user_account::Builder::new(json), quota_project_id, scopes)
+                }
+                "service_account" => configure_and_build(
+                    service_account::Builder::new(json),
+                    quota_project_id,
+                    scopes,
+                ),
+                _ => Err(errors::non_retryable_from_str(format!(
+                    "Unimplemented credentials type: {cred_type}"
```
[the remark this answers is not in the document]
harkamaljot: updated error message to `Invalid or unsupported credentials type found in JSON: {cred_type}` to reflect that credentials type can be invalid as well.

coryan (approved):

chrischiedo:
Sorry folks. I was just starting to review this PR (I know it's been open for 5 days -- it's only that I didn't get to it sooner). I see that it has just been approved. Can I still go ahead and leave my review comments at this point or should I raise those in a separate issue?

Your guidance will be highly appreciated. Thanks!

cc/ @coryan @harkamaljot

coryan:
Whatever is easier for you, we are a "improvement today is better than perfection tomorrow" kind of team, so I would not mind adding a second PR later.

chrischiedo:
Noted. Let me share my comments in the current PR -- not deal-breakers really, but I think they might still be helpful. Thanks for the quick response.

src/auth/src/credentials.rs
```diff
@@ -206,6 +207,213 @@ pub(crate) mod dynamic {
...
+/// # });
+/// ```
+///
+/// [ADC-link]: https://cloud.google.com/docs/authentication/application-default-credentials
+/// [AIP-4110]: https://google.aip.dev/auth/4110
+/// [Cloud Run]: https://cloud.google.com/run
+/// [gce-link]: https://cloud.google.com/products/compute
+/// [gcloud auth application-default]: https://cloud.google.com/sdk/gcloud/reference/auth/application-default
+/// [gke-link]: https://cloud.google.com/kubernetes-engine
+pub struct Builder {
```
chrischiedo, line 297: Not a big deal, but...

While this is a common naming convention, I think the alternative approach of being more descriptive about "what the builder builds" is better. Yes, the documentation is quite clear about what this `Builder` is all about (`A builder for constructing Credentials instances`), being able to glean that from the name of the type itself is much better. So, in this case, having `CredentialsBuilder` would be preferred over `Builder`.

You can check examples from the `tokio` and `reqwest` projects:
- [HistogramBuilder](https://github.com/tokio-rs/tokio/blob/1434b32b5a0df3b38a0d588485cd9a20a8e92a89/tokio/src/runtime/metrics/histogram.rs#L25) from tokio
- [ClientBuilder](https://github.com/seanmonstar/reqwest/blob/36fd0381ae1aa25e89475a90834cbd5ca0ca6404/src/blocking/client.rs#L78) from reqwest

But again, not really a deal-breaker per se.

src/auth/src/credentials.rs
```diff
@@ -206,6 +207,213 @@ pub(crate) mod dynamic {
...
+    credentials_source: CredentialsSource,
+    quota_project_id: Option<String>,
+    scopes: Option<Vec<String>>,
+}
+
+impl Default for Builder {
```
chrischiedo, line 303: Consider implementing `Debug` for Builder as well. This is considered good practice (source: [Rust API Guidelines Checklist](https://rust-lang.github.io/api-guidelines/checklist.html#rust-api-guidelines-checklist): Debuggability). The example builders from tokio and reqwest shared in a previous comment adhere to this.

src/auth/tests/credentials.rs
```diff
@@ -17,8 +17,8 @@ use google_cloud_auth::credentials::service_account::Builder as ServiceAccountBu
 use google_cloud_auth::credentials::testing::test_credentials;
 use google_cloud_auth::credentials::user_account::Builder as UserAccountCredentialBuilder;
 use google_cloud_auth::credentials::{
-    ApiKeyOptions, Credentials, CredentialsTrait, create_access_token_credentials,
-    create_api_key_credentials,
+    ApiKeyOptions, Builder as AccessTokenCredentialBuilder, Credentials, CredentialsTrait,
```
chrischiedo, line 20: The name aliasing here is interesting (`Builder as AccessTokenCredentialBuilder`). Would it still be necessary if we had `CredentialsBuilder` instead of just `Builder`?

chrischiedo (commented):
I've left a few high-level review comments. Otherwise, the PR looks good.

src/auth/src/credentials.rs
harkamaljot: Thanks for the feedback and the examples from tokio and reqwest. That's a valid point about the descriptiveness of CredentialsBuilder.

We actually started with that naming convention. However, we decided to use the simpler Builder name here because the struct is already located within the credentials module (.../src/auth/src/credentials/...). Using credentials::CredentialsBuilder felt somewhat redundant in context.

We've also established Builder as the naming convention for similar builders within this specific auth module (like the one in mds.rs) to maintain internal consistency.

While I acknowledge the clarity benefit of the explicit name, In this case, we opted for conciseness and local consistency, relying on the module path and documentation to provide context.

src/auth/tests/credentials.rs
harkamaljot: If the type were named `CredentialsBuilder`, it wouldn't necessarily require aliasing just to avoid name collisions when used alongside other differently named builders (like a potential ServiceAccountBuilder).

However, in this test file, we are importing builders for multiple credential types (user, MDS, service account). Since many of these follow the convention of being named simply `Builder` within their respective modules (e.g., mds::Builder, service_account::Builder), aliasing becomes essential here to prevent name collisions and to maintain clarity about which specific builder implementation is being used.

While many typical use cases might only involve importing one type of credential builder at a time, so this shouldn't be a problem there.

src/auth/src/credentials.rs
harkamaljot: Thanks for the suggestion added `Debug` here.

src/auth/src/credentials.rs
```diff
@@ -206,6 +207,214 @@ pub(crate) mod dynamic {
     }
 }
 
+#[derive(Debug, Clone)]
+enum CredentialsSource {
+    CredentialsJson(Value),
+    DefaultCredential,
```
sai-sunder-s, line 213: DefaultCredentials

sai-sunder-s (approved):

PR #1800, opened 2025-04-15
chore: simplify configuration setup and remove unnecessary iterator 

kasugamirai:


coryan:
Thanks again for the clean ups.  If you don't mind me asking, how do you find these? Do you use any tools or just read the code?

coryan:
Ah, I think `clippy --profile=test -- --deny warnings`, we should add that to our CI builds.

coryan (approved):

dbolduc (approved):

mrfaizal (approved):

PR #1801, opened 2025-04-15
cleanup: clippyfy all the tests

coryan:
`cargo clippy` does not lint the tests, `cargo clippy --profile=test` is
what we want to run.

### DIFF
--- a/src/firestore/src/status.rs
+++ b/src/firestore/src/status.rs
@@ -174,12 +174,10 @@ mod test {
                     description: "desc".into(),
                     ..Default::default()
                 }],
-                ..Default::default()
             }),
             Any::from_msg(&DebugInfo {
                 stack_entries: ["stack"].map(str::to_string).to_vec(),
                 detail: "detail".into(),
-                ..Default::default()
             }),
             Any::from_msg(&ErrorInfo {
                 reason: "reason".into(),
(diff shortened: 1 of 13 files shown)


src/gax/src/error/rpc.rs
```diff
@@ -850,13 +850,10 @@ mod test {
...
         let err = Code::try_from("INVALID-NOT-A-CODE");
-        match err {
-            Err(s) => assert!(
-                s.contains("INVALID-NOT-A-CODE"),
-                "expected invalid string in error {s}"
-            ),
-            Ok(v) => assert!(false, "expected error in try_from, got {v:?}"),
-        };
+        assert!(
+            matches!(&err, Err(s) if s.contains("INVALID-NOT-A-CODE")),
```
dbolduc, line 854: TIL: `matches!`

dbolduc (approved):

src/firestore/src/status.rs
```diff
@@ -174,12 +174,10 @@ mod test {
                     description: "desc".into(),
                     ..Default::default()
                 }],
-                ..Default::default()
```
dbolduc, line 177: Hilariously, removing these sets us up for test breaks if/when new fields are added to the protos.

Aside: seems like the prost messages should be marked `#[non_exhaustive]`.

Issue #1813, opened 2025-04-15
Generate scaffold for Cloud Storage


coryan:
We can generate all the admin (and unary) operations for the `.google.storage.v2.Storage` service. It will not be functional at first, but can give us a sense of what is missing.

To get a full client we need to:

- [x] Generate code for methods without HTTP annotations
- [x] Generate the scaffold with code skipped to get started
- [x] #1819 
- [x] Temporarily suppress documentation for client builders
- [x] Temporarily suppress documentation links in request builders
- [x] Write a basic client selectively exposing RPCs
- [x] Rename the crate to `google-cloud-storage-control`
- [x] Figure out a story for test coverage of veneers
- [x] Figure out a story for mocking veneers
  - This is not so bad for a trivial veneer. See [this commit](https://github.com/dbolduc/google-cloud-rust/commit/31c0c202c06ef9286f5f0b6f3f76209e38168842).
- [x] Rename builder module to `storage_control`
- [x] Rename client to `StorageControl`
- [x] Refactor the code to convert `google.rpc.Status` (maybe) handling some custom details, aka #1699

dbolduc:
> Figure out a story for test coverage of veneers

We have integration tests for every flavor of RPC (unary, paginated, LRO, unary returning empty, etc.).

We also have a unit test that uses a mock client.

The handwritten surface is incredibly thin. We copy the underlying GAPICs' signatures exactly, and just call through to them.

I think we are good on the test coverage.

PR #1821, opened 2025-04-16
impl(storage): add generated code

coryan:
Add the generated code and minimum supporting files to compile it.
Disable a number of warnings for now. Export `client`, `builder` and
other modules because they are needed for `cargo doc` to pass.

Part of the work for #1813


src/storage/src/generated/protos/storage/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+#
+# Code generated by sidekick. DO NOT EDIT.
+
+[package]
+name        = "google-cloud-storage-v2-code-generation"
+version     = "0.0.0"
+description = "Prost and Tonic generator for google-cloud-storage-v2"
+publish     = false
+build       = "build.rs"
+edition     = "2021"
```
dbolduc, line 23: non-blocking Q: do we have to use 2021? can it be 2024?

src/storage/src/lib.rs
```diff
@@ -23,3 +23,60 @@
...
+        pub mod v1 {
+            include!("generated/protos/storage/google.iam.v1.rs");
+        }
+    }
+    pub mod r#type {
+        include!("generated/protos/storage/google.r#type.rs");
+        include!("generated/convert/type/convert.rs");
+    }
+    pub mod rpc {
+        include!("generated/protos/storage/google.rpc.rs");
```
dbolduc, line 52: yucky

src/storage/src/generated/convert/type/.sidekick.toml
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[general]
+language             = 'rust'
+specification-source = 'google/type'
+
+[source]
+include-list = 'date.proto'
```
dbolduc, line 20: comment: Ah, this is the only proto that storage needs from `google/type/`.

https://github.com/googleapis/googleapis/blob/7e9433ad5d5f8615648f25deab17bef31a6fcc36/google/storage/v2/storage.proto#L30

src/storage/src/lib.rs
```diff
@@ -23,3 +23,60 @@
...
+// TODO(#1813) - fix the broken link to [here].
+#[allow(rustdoc::broken_intra_doc_links)]
+// TODO(#1813) - fix x-goog-request-params and this is not needed
+#[allow(clippy::op_ref)]
+#[allow(clippy::needless_borrow)]
+pub(crate) mod generated;
+
+pub use generated::gapic::builder;
+pub use generated::gapic::client;
+pub use generated::gapic::model;
```
dbolduc, line 39: this looks like a `feat:`

dbolduc (approved):
no changes, just comments and a question

src/storage/src/generated/protos/storage/Cargo.toml
coryan: Yes, we could change that.

src/storage/src/lib.rs
coryan: Agree.

PR #1833, opened 2025-04-18
refactor(gaxi): make fixed headers outside of loop

dbolduc:
Part of the work for #1612 

Move header creation outside of the retry loop.

The strings referenced by a `HeaderValue` (such as our `api_client_header`) are copied into a `HeaderMap` when a `HeaderMap` is made. The lifetime of the `HeaderMap` is easier for the compiler to reason about. (We do not want the inner call in the retry loop to have a `'static` lifetime, which I think is what was happening when we tried to hold `&'static str`s in it.)

Cleanups:
- This brings us to <= 7 parameters, so we can remove the clippy allows.
- Restrict visibility on implementation details. `execute()` is the only API we need to expose for our tests / client crates.

### DIFF
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -68,13 +68,19 @@ impl Client {
         Request: prost::Message + 'static + Clone,
         Response: prost::Message + Default + 'static,
     {
-        let headers =
-            Self::make_headers(&self.credentials, api_client_header, request_params).await?;
+        let headers = Self::make_headers(api_client_header, request_params).await?;
         match self.get_retry_policy(&options) {
             None => {
                 let mut inner = self.inner.clone();
                 Self::request_attempt::<Request, Response>(
-                    &mut inner, method, path, request, &options, None, headers,
+                    &mut inner,
+                    &self.credentials,
+                    method,
+                    path,
+                    request,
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/grpc.rs
```diff
@@ -67,48 +68,34 @@ impl Client {
         Request: prost::Message + 'static + Clone,
         Response: prost::Message + Default + 'static,
     {
+        let headers =
+            Self::make_headers(&self.credentials, api_client_header, request_params).await?;
```
coryan, line 72: The headers also include the credentials, and we need those to be created inside the loop.  Can we (maybe) split this in two: create fixed headers outside the loop, and then append the credential headers inside the loop?

coryan (changes requested):

src/gax-internal/src/grpc.rs
dbolduc: > we need those to be created inside the loop.

Yes, thanks for catching. :facepalm: Ugh, I got lost in the sauce on this refactor.

#1839 adds a test that would have caught this sooner.

dbolduc: Fixed

coryan (approved):

PR #1834, opened 2025-04-18
impl(gax-internal): fix tonic endpoints

dbolduc:
Part of the work for #1612 

`endpoint.connect()` would fail when trying to connect to `https://firestore.googleapis.com`. This was noticed when running the integration tests with a firestore client implemented by the `grpc::Client`.

Apparently we need to configure TLS on the endpoint. Note that:

(1) Our current firestore client (which passes the integration tests) uses `tonic::Endpoint::new()` (which @coryan notices is not part of `tonic`'s public API)

https://github.com/googleapis/google-cloud-rust/blob/cf54b210da52fd216042b57fb235e0edbb9cec47/src/firestore/src/generated/gapic/transport.rs#L74

(2) We are enabling the `_tls-any` feature via `tonic:tls-ring`, which gates the `tonic::transport::ClientTlsConfig` API.

### DIFF
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -175,7 +175,7 @@ impl Client {
     }
 
     async fn make_inner(endpoint: Option<String>, default_endpoint: &str) -> Result<InnerClient> {
-        let endpoint = tonic::transport::Endpoint::from_shared(
+        let endpoint = tonic::transport::Endpoint::new(
             endpoint.unwrap_or_else(|| default_endpoint.to_string()),
         )
         .map_err(Error::other)?;
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -175,10 +175,12 @@ impl Client {
     }
 
     async fn make_inner(endpoint: Option<String>, default_endpoint: &str) -> Result<InnerClient> {
-        let endpoint = tonic::transport::Endpoint::new(
-            endpoint.unwrap_or_else(|| default_endpoint.to_string()),
-        )
-        .map_err(Error::other)?;
+        use tonic::transport::{ClientTlsConfig, Endpoint};
+        let endpoint =
+            Endpoint::from_shared(endpoint.unwrap_or_else(|| default_endpoint.to_string()))
+                .map_err(Error::other)?
+                .tls_config(ClientTlsConfig::new().with_enabled_roots())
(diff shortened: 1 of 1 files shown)


coryan (approved):

src/gax-internal/src/grpc.rs
```diff
@@ -182,7 +182,7 @@ impl Client {
     }
 
     async fn make_inner(endpoint: Option<String>, default_endpoint: &str) -> Result<InnerClient> {
-        let endpoint = tonic::transport::Endpoint::from_shared(
+        let endpoint = tonic::transport::Endpoint::new(
```
coryan, line 185: `::new()` is an undocumented feature:

https://github.com/hyperium/tonic/blob/50253f1cb236feca3061488d542e37cf60ba4f65/tonic/src/transport/channel/endpoint.rs#L54-L57


We should not be using it. That is why I changed the code to use `from_shared()`, but maybe we need to do something more difficult?

coryan (changes requested):

src/gax-internal/src/grpc.rs
dbolduc: > `::new()` is an undocumented feature:

Ah, I missed this. Thanks.

> something more difficult?

Fixed by copying the implementation of `new()`.

(I am not going to update the instances of `Endpoint::new()` in generated code in this PR. Eventually the generated code will use the `grpc::Client` instead.)

src/gax-internal/src/grpc.rs
```diff
@@ -175,10 +175,12 @@ impl Client {
...
     async fn make_inner(endpoint: Option<String>, default_endpoint: &str) -> Result<InnerClient> {
-        let endpoint = tonic::transport::Endpoint::from_shared(
-            endpoint.unwrap_or_else(|| default_endpoint.to_string()),
-        )
-        .map_err(Error::other)?;
+        use tonic::transport::{ClientTlsConfig, Endpoint};
+        let endpoint =
+            Endpoint::from_shared(endpoint.unwrap_or_else(|| default_endpoint.to_string()))
+                .map_err(Error::other)?
+                .tls_config(ClientTlsConfig::new().with_enabled_roots())
```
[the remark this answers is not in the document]
dbolduc: > eventually... For gRPC and REST

Ack. In general, I worry about not being able to unify the options for the two different transports.

PR #1839, opened 2025-04-18
test(gaxi): auth in grpc client

dbolduc:
Motivated by #1612 

Make sure we call auth from inside the retry loop for the gRPC client, and bubble up the errors, and all that jazz.

These tests are basically the same as the `auth.rs` tests for the HTTP client. The reviewer is heavily advised to compare the test fixtures.

I am kicking the can on trying to unify gRPC vs. HTTP client tests. We would have to unify the interface for both servers and both clients. I am not sure it is worth doing yet.

While we're here, speed up the existing auth test for the HTTP client by using a snappier backoff policy.

### DIFF
--- a/src/gax-internal/tests/grpc_auth.rs
+++ b/src/gax-internal/tests/grpc_auth.rs
@@ -12,15 +12,17 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[cfg(all(test, feature = "_internal_http_client"))]
+//#[cfg(all(test, feature = "_internal_grpc_client"))]
+#[cfg(test)]
 mod test {
     use auth::credentials::{Credentials, CredentialsTrait};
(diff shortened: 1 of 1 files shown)


src/gax-internal/tests/grpc_auth.rs
coryan, line 1: nit: For a separate PR, I think we should rename the http tests too.  Or maybe use subdirectories if we can figure out how to do that.

coryan (approved):

src/gax-internal/tests/grpc_auth.rs
dbolduc: Ack.

dbolduc: https://github.com/googleapis/google-cloud-rust/pull/1841

PR #1845, opened 2025-04-18
refactor(gaxi): grpc request_attempt as member

dbolduc:
Part of the work for #1612 

Just pass along a `grpc::Client`. That is what we do in the HTTP client.

We are cloning the policies for the `self` that does a single request attempt. Sure, that is extra work, but it is negligible. These things are `Option<Arc<...>>`s. They are cheap to clone (we clone them for every single operation).

### DIFF
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -71,16 +71,8 @@ impl Client {
         let headers = Self::make_headers(api_client_header, request_params).await?;
         match self.get_retry_policy(&options) {
             None => {
-                let mut inner = self.inner.clone();
-                Self::request_attempt::<Request, Response>(
-                    &mut inner,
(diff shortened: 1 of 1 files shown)


coryan (approved):
> These things are Option<Arc<...>>s. They are cheap to clone (we clone them for every single operation).

Relatively speaking, yes.  `Arc` is *atomically* reference counted. That implies a lock or barrier, so about the cost of locking a mutex.  Less than the cost of a large data copy, more than the cost of accessing a few fields.  Look for "Latency numbers every programmer should know", such as:

https://gist.github.com/hellerbarde/2843375
https://static.googleusercontent.com/media/sre.google/en//static/pdf/rule-of-thumb-latency-numbers-letter.pdf

dbolduc:
Ack.

PR #1849, opened 2025-04-18
impl(gaxi): `Convert` for unit type

dbolduc:
Part of the work for #1612 

I suspect this will simplify generated code.


coryan:
Hmmm.. #1593 and #1604 ... it was less useful than I hoped.

PR #1855, opened 2025-04-19
cleanup(tests): simplify client configuration and response handling

kasugamirai:


src/storage/src/generated/protos/storage/google.storage.v2.rs
```diff
@@ -172,9 +172,7 @@ pub mod compose_object_request {
         #[prost(int64, tag = "2")]
         pub generation: i64,
         #[prost(message, optional, tag = "3")]
-        pub object_preconditions: ::core::option::Option<
-            source_object::ObjectPreconditions,
-        >,
+        pub object_preconditions: ::core::option::Option<source_object::ObjectPreconditions>,
```
coryan, line 175: This is generated code. In particular, this is generated by Prost. Please revert the changes to this file.

coryan (changes requested):
Thanks for the cleanups!  I think we should revert the changes to the generated code: while it may be possible to clean it up, we should avoid that temptation.

coryan (approved):

dbolduc (approved):

PR #1858, opened 2025-04-19
WIP - Draft-generator-per-service-features

coryan:


generator/internal/rust/templates/convert-prost/feature_gate.mustache
```diff
@@ -0,0 +1,16 @@
+{{!
```
dbolduc, line 1: why do we need the templates outside of `generator/internal/rust/templates/common/` ?

(genuine question, I don't know how this stuff works)

dbolduc: Are you planning to use `feature_gate` for the annotations from #1878?

dbolduc (approved):

generator/internal/rust/templates/convert-prost/feature_gate.mustache
coryan: > why do we need the templates outside of `generator/internal/rust/templates/common/` ?
> 
> (genuine question, I don't know how this stuff works)

Because golang's version of mustache searches in the same directory as the caller. For example:  `crates/src/message.mustache` searches for `crate/src/feature_gate.mustache` even though it "calls" `../common/message.mustache`.

But maybe there is a way to use absolute paths or something.  It would be a nice cleanup if it was possible (we also duplicate `message.mustache` and `enum.mustache` and a number of other things).

I will give it a try later and report back.

coryan: > Are you planning to use `feature_gate` for the annotations from #1878?

I am not sure what you mean.

dbolduc: > even though it "calls" `../common/message.mustache`.

This is why I was confused. I thought we were only coming from `common/`.

> I am not sure what you mean.

I was only trying to understand why we needed `feature_gate.mustache` in `crate/src/` (for example). And I noticed that this:

https://github.com/coryan/google-cloud-rust/blob/6d55adbfc3f5eb2a2fa9f34fbf54611f5e01d39a/generator/internal/rust/templates/crate/src/transport.rs.mustache#L22-L24

... is just a special case of `feature_gate.mustache`.

coryan: > ... is just a special case of `feature_gate.mustache`.

`feature_gate.mustache` assumes that the `.Codec` field has `.MultiFeatureGate` and `.FeatureGateOp` and `.FeatureGates`. None of those things are true for the top-level Codec.  We can make them true, but that would move the complexity to the `*.go` files and I am not sure that is a good tradeoff?

dbolduc: I am not saying that we should do it. Everything is fine.

> Are you planning to use `feature_gate` for the annotations from #1878?

^ in this comment, I was guessing as to why the `feature_gate.mustache` was needed in `crate/src`. And my guess was wrong.

PR #1860, opened 2025-04-20
refactor(generator): replace HasServices field with method for service

### LINKED ISSUE #1854: Prefer helper functions over duplicate fields in codec annotations
Sometimes our templates need to generate some extra code if a list is non-empty, for example:

https://github.com/googleapis/google-cloud-rust/blob/3db89bb5d543ddf4b3f642f9ceb28f46d98cb5be/generator/internal/rust/templates/common/README.md.mustache#L28-L33

So far, we have introduced an extra variable to the annotations to support such cases:

https://github.com/googleapis/google-cloud-rust/blob/3db89bb5d543ddf4b3f642f9ceb28f46d98cb5be/generator/internal/rust/annotate.go#L35

Turns out we can avoid the extra variable (and remove some opportunity for mistakes) by adding a helper function , in this case:

```go
func (a *modelAnnotations) HasServices() bool {
  return len(a.Services) > 0
}
```

dhvll:
Part of the work for #1854

Remove duplicate HasServices field in modelAnnotations

Changes made:
1. Removed the HasServices boolean field from the modelAnnotations struct
2. Added a HasServices() method that returns true if there are any services in the Services slice
3. Updated the call in generate.go to use the new method instead of the field

### DIFF
--- a/generator/internal/rust/annotate.go
+++ b/generator/internal/rust/annotate.go
@@ -32,7 +32,6 @@ type modelAnnotations struct {
 	PackageNamespace string
 	RequiredPackages []string
 	ExternPackages   []string
-	HasServices      bool
 	HasLROs          bool
 	CopyrightYear    string
 	BoilerPlate      []string
@@ -54,6 +53,11 @@ type modelAnnotations struct {
 	DisabledRustdocWarnings []string
 	// Sets the default system parameters
 	DefaultSystemParameters []systemParameter
+}
+
+// HasServices returns true if there are any services in the model
+func (m *modelAnnotations) HasServices() bool {
+	return len(m.Services) > 0
 }
 
 type serviceAnnotations struct {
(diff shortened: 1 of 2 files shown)


coryan:
Thanks for the PR!  I do not think this fixes #1860, as there are other such flags in the code: `HasServices` was intended as an example not an exhaustive list. Nevertheless, this is a good start, so I will happily merge the changes once the CI passes and another colleague reviews the code.

dhvll:
Thanks for the feedback. Is there any change which I can do?

coryan (approved):

dbolduc (approved):

coryan:
> Thanks for the feedback. Is there any change which I can do?

You would need to find any other `Has*` field in the `annotate.go` that correspond to arrays of elements and replace them with a function.  This is probably a good start:

```
git grep -E 'Has.*:.*len\(' -- *.go
generator/internal/dart/annotate.go:            HasServices:    len(model.Services) > 0,
generator/internal/dart/annotate.go:            HasDependencies:     len(packageDependencies) > 0,
generator/internal/dart/annotate.go:            HasDevDependencies:  len(devDependencies) > 0,
generator/internal/dart/annotate.go:            HasFields:         len(m.Fields) > 0,
generator/internal/dart/annotate.go:            HasToStringLines:  len(toStringLines) > 0,
generator/internal/golang/gotemplate.go:                HasServices:       len(model.Services) > 0,
generator/internal/rust/annotate.go:            HasServices:      len(servicesSubset) > 0,
```

Issue #1863, opened 2025-04-21
Introduce per-service crate features to reduce build times


coryan:
We want per-service [^1] features so application developers can compile a subset of large crates, such as `google-cloud-aiplatform-v1`.

[^1]: in this context a "service" is a "protobuf syntax service", which maps to a single `client::*` struct in our crates.

coryan:
I am going to pause here.  If anybody wants to take over and add some controls for the default feature, please go ahead.

coryan:
Split the thing that is left over to its own bug.

PR #1864, opened 2025-04-21
feat(generator/rust): per-service feature option

coryan:
Introduce an option to generate per-service features. Use this option
in the `Cargo.toml.mustache` template to introduce the features and
the default feature.

Part of the work for #1863

### DIFF
--- a/src/generated/cloud/aiplatform/v1/Cargo.toml
+++ b/src/generated/cloud/aiplatform/v1/Cargo.toml
@@ -26,6 +26,74 @@ keywords.workspace     = true
 categories.workspace   = true
 rust-version.workspace = true
 
+[features]
+default = [
+    "dataset_service",
+    "deployment_resource_pool_service",
+    "endpoint_service",
+    "evaluation_service",
+    "feature_online_store_admin_service",
+    "feature_online_store_service",
+    "feature_registry_service",
+    "featurestore_online_serving_service",
+    "featurestore_service",
+    "gen_ai_cache_service",
+    "gen_ai_tuning_service",
+    "index_endpoint_service",
+    "index_service",
+    "job_service",
+    "llm_utility_service",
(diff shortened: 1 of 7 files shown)


src/generated/cloud/aiplatform/v1/Cargo.toml
```diff
@@ -26,6 +26,74 @@ keywords.workspace     = true
...
+    "reasoning_engine_execution_service",
+    "reasoning_engine_service",
+    "schedule_service",
+    "specialist_pool_service",
+    "tensorboard_service",
+    "vertex_rag_data_service",
+    "vertex_rag_service",
+    "vizier_service",
+]
+dataset_service = []
```
dbolduc, line 64: forecasting: some, but not all of these will require gRPC. Will we micro-manage the features' dependencies?

I lean: nah.

src/generated/cloud/aiplatform/v1/Cargo.toml
```diff
@@ -26,6 +26,74 @@ keywords.workspace     = true
 categories.workspace   = true
 rust-version.workspace = true
 
+[features]
+default = [
```
dbolduc, line 30: Q: should we be opinionated about the default set?

e.g. if our telemetry indicated that 99.9% of requests were for either `prediction_service` or `gen_ai_tuning_service`, we'd probably prefer for those two features to be the extent of our default set.

We are allowed to add default features, but not allowed to remove default features.

(Dont let this block your PR though. Just something to think about before v1.0)

dbolduc (approved):

src/generated/cloud/aiplatform/v1/Cargo.toml
coryan: I think eventually: yes. But we can move forward without this.

PR #1869, opened 2025-04-21
impl(generator/rust): compute service dependencies

coryan:
Implement a function to compute the transitive closures of a service
dependencies, including all request messages, all response messages,
all LRO metadata and response messages, and the transitive deps of these
messages.

Part of the work for #1863

### DIFF
--- a/generator/internal/api/service_dependencies.go
+++ b/generator/internal/api/service_dependencies.go
@@ -49,8 +49,6 @@ func FindServiceDependencies(model *API, serviceID string) *ServiceDependencies 
 }
 
 type findServiceState struct {
-	// The message IDs that have already been visited
-	Visited map[string]bool
 	// The enums already included in these results
 	Enums map[string]bool
 	// The messages already included in these results
@@ -62,7 +60,6 @@ type findServiceState struct {
 
 func newFindServicesState(serviceID string, model *API) *findServiceState {
 	return &findServiceState{
-		Visited:   map[string]bool{},
 		Enums:     map[string]bool{},
 		Messages:  map[string]bool{},
 		model:     model,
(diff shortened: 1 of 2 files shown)


generator/internal/api/service_dependencies.go
```diff
@@ -0,0 +1,140 @@
...
+	model      *API
+	candidates []*Message
+	serviceID  string
+}
+
+func newFindServicesState(serviceID string, model *API) *findServiceState {
+	return &findServiceState{
+		Visited:   map[string]bool{},
+		Enums:     map[string]bool{},
+		Messages:  map[string]bool{},
```
dbolduc, line 67: I guess golang does not have sets either. :unamused:

generator/internal/api/service_dependencies.go
```diff
@@ -0,0 +1,140 @@
...
+
+type findServiceState struct {
+	// The message IDs that have already been visited
+	Visited map[string]bool
+	// The enums already included in these results
+	Enums map[string]bool
+	// The messages already included in these results
+	Messages   map[string]bool
+	model      *API
+	candidates []*Message
```
dbolduc, line 59: do they not have a deque?

generator/internal/api/service_dependencies.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+}
+
+func (state *findServiceState) addMessage(id string) {
+	state.addCandidate(id)
+	if m, ok := state.model.State.MessageByID[id]; ok {
+		if m.Parent != nil {
+			state.addCandidate(m.Parent.ID)
+		}
+		if !m.IsMap {
```
dbolduc, line 125: why aren't we handling maps? wouldn't we have to recurse over `m.Fields[0]`, `m.Fields[1]`?

generator/internal/api/service_dependencies.go
```diff
@@ -0,0 +1,140 @@
...
+	// The messages already included in these results
+	Messages   map[string]bool
+	model      *API
+	candidates []*Message
+	serviceID  string
+}
+
+func newFindServicesState(serviceID string, model *API) *findServiceState {
+	return &findServiceState{
+		Visited:   map[string]bool{},
```
dbolduc, line 65: Aren't `Visited` and `Messages` the same set?

coryan: Not for `.IsMap == true`. But I think we can change the code to set `Messages[]` for maps without harm.

generator/internal/api/service_dependencies.go
dbolduc: > a stack works

yes. FTR, my comment was more about preferring container functions like `PopBack()` over manually implementing a stack using an array, so the code is more readable.

PR #1874, opened 2025-04-21
refactor: use `grpc::Client`

dbolduc:
Most of the work for #1612

Implement our gRPC-based client crates using the `gaxi::grpc::Client`.

I think it is slightly more than a pure refactor as these clients now support timeouts.

We rely on the firestore integration tests to ensure correctness of the code.


dbolduc:
> nit: it also affects `storage`, so maybe just "refactor:" in the title?

Done.

coryan (approved):

PR #1876, opened 2025-04-21
cleanup(wkt): refactor `wkt::message::Message` trait

chrischiedo:
This PR is part of the work for #1609 (please check the issue for more/better context).

Creates a new `SerializableMessage` trait inside an internal module (with `pub(crate)` visibility), that is hidden from docs.

**Note**: Some alternative names considered for the new trait were: `MessageSerializer` and `MapSerializable`. I'm open to considering a better name if suggested.

cc/ @coryan

### DIFF
--- a/src/wkt/src/any.rs
+++ b/src/wkt/src/any.rs
@@ -101,7 +101,7 @@ impl Any {
     /// also supports serialization to JSON.
     pub fn try_from<T>(message: &T) -> Result<Self, Error>
     where
-        T: serde::ser::Serialize + crate::message::Message,
+        T: serde::ser::Serialize + crate::message::Message + crate::message::internal::SerializableMessage,
     {
         let value = message.to_map()?;
         Ok(Any(value))
@@ -110,7 +110,7 @@ impl Any {
     /// Extracts (if possible) a `T` value from the [Any].
     pub fn try_into_message<T>(&self) -> Result<T, Error>
     where
-        T: serde::de::DeserializeOwned + crate::message::Message,
+        T: serde::de::DeserializeOwned + crate::message::Message + crate::message::internal::SerializableMessage,
     {
         let map = &self.0;
         let r#type = map
@@ -136,10 +136,15 @@ impl crate::message::Message for Any {
     fn typename() -> &'static str {
         "type.googleapis.com/google.protobuf.Any"
     }
+}
+
+#[doc(hidden)]
+impl crate::message::internal::SerializableMessage for Any {
     fn to_map(&self) -> Result<crate::message::Map, AnyError> {
         use serde_json::Value;
+        use crate::message::Message;
         let map = [
-            ("@type", Value::String(Self::typename().into())),
+            ("@type", Value::String(<Self as Message>::typename().into())),
             ("value", Value::Object(self.0.clone())),
         ]
         .into_iter()
@@ -147,6 +152,7 @@ impl crate::message::Message for Any {
         .collect::<crate::message::Map>();
         Ok(map)
     }
+
     fn from_map(map: &crate::message::Map) -> Result<Self, AnyError> {
         crate::message::from_value(map)
     }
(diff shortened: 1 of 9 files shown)


coryan:
I think just splitting the message trait won't work. We would need to do something like `impl<T> SerializableMessage for T where T: Message` and that introduces ambiguity.  I need to think this over tomorrow morning, right now I cannot think straight.

chrischiedo:
> I think just splitting the message trait won't work. We would need to do something like `impl<T> SerializableMessage for T where T: Message` and that introduces ambiguity. I need to think this over tomorrow morning, right now I cannot think straight.

I actually realized that I had a minor error in the `impl_message_traits` macro definition (which was causing the build to fail).

I think having this implementation should work fine (I'm not sure how this would introduce ambiquity -- but I'll wait on your suggestions/review comments, if there are better alternatives):

```
impl<T> internal::SerializableMessage for T 
where 
    T: Message, 
{
    // No need to implement any methods here since both methods in SerializableMessage have default implementations.
    // The trait is automatically implemented for any type T that satisfies the Message bound.
}
```

The other approach I had considered was to "seal" the `Message` trait:

```
pub trait Message: internal::SerializableMessage { }
```
But I don't think that would be helpful in this context (happy to reconsider).

Also, in the sealed case, types in `wkt/src/generated/mod.rs` that implement `Message` would also need to implement `internal::SerializableMessage`. I'm not sure if generated files can/should be edited (commented with: `Code generated by sidekick. DO NOT EDIT`).

coryan:
> > I think just splitting the message trait won't work. We would need to do something like `impl<T> SerializableMessage for T where T: Message` and that introduces ambiguity. I need to think this over tomorrow morning, right now I cannot think straight.
> 
> I actually realized that I had a minor error in the `impl_message_traits` macro definition (which was causing the build to fail).

The builds still fail. Does `cargo test` work locally?

> I think having this implementation should work fine (I'm not sure how this would introduce ambiquity -- but I'll wait on your suggestions/review comments, if there are better alternatives):
> 
> ```
> impl<T> internal::SerializableMessage for T 
> where 
>     T: Message, 
> {
>     // No need to implement any methods here since both methods in SerializableMessage have default implementations.
>     // The trait is automatically implemented for any type T that satisfies the Message bound.
> }
> ```

We would probably want to implement this too:

```rust
impl internal::SerializableMessage for Duration {
 // The default implementation of the methods does not work, need to define custom versions..
}
```

And because `Duration` also implements `Message` we will have ambiguity.  That is my bad, we need something more interesting.

> 
> The other approach I had considered was to "seal" the `Message` trait:
> 
> ```
> pub trait Message: internal::SerializableMessage { }
> ```
> 
> But I don't think that would be helpful in this context (happy to reconsider).

I agree, we need to define `Message` outside this crate.

> Also, in the sealed case, types in `wkt/src/generated/mod.rs` that implement `Message` would also need to implement `internal::SerializableMessage`. I'm not sure if generated files can/should be edited (commented with: `Code generated by sidekick. DO NOT EDIT`).

The generator lives in the same repository. We can change the generator if needed and regenerate the code. But let's not worry about that just yet.

We should prototype a solution that avoid the ambiguity and then starting coding away.

chrischiedo:
> The builds still fail.

Okay. Let me investigate further.

> And because `Duration` also implements `Message` we will have ambiguity. That is my bad, we need something more interesting.

I get your point now. Please let me think through it and see if I can come up with a better solution.

> I agree, we need to define `Message` outside this crate.

Okay, let me think through this too.

> The generator lives in the same repository. We can change the generator if needed and regenerate the code. But let's not worry about that just yet.

This is noted. Thanks for the clarification.

> We should prototype a solution that avoid the ambiguity and then starting coding away.

Agreed.

In the meantime, let me convert the PR into a draft.

chrischiedo:
This PR is superseded by #1946. Closing.

PR #1877, opened 2025-04-21
cleanup(gax)!: hide default idempotency in `RequestOptions`

dbolduc:
Part of the work for #1541, noticed in #1612 

`RequestOptions::set_default_idempotency` is for the library only. We don't actually want customers to use it. (They use the builder's `with_idempotency()` method, which calls into `set_idempotency()`).

So hide the method in an internal module.

I considered modifying the HTTP client to accept idempotency as a flag, but decided against it. Maybe because I am lazy. I will keep passing it along via `options`.


coryan (changes requested):
We should bump the major version of `gax` since this is the first breaking change since the last release.

We may consider bumping all the patch version numbers for generated clients, since they now will require a newer version of `gax`.

dbolduc:
> We should bump the major version of `gax` since this is the first breaking change since the last release.

Done.

> We may consider bumping all the patch version numbers for generated clients, since they now will require a newer version of `gax`.

I see why we should do this. Annoyingly only crates with services need their version bumped. It might be time to start versioning things independently

dbolduc:
> We may consider bumping all the patch version numbers for generated clients, since they now will require a newer version of `gax`.

I bumped the minor version in anticipation of breaking changes.

coryan (approved):

Issue #1835, opened 2025-04-18
Extra slash in MDS URLs causes two calls per request due to redirect


fiadliel:
The auth code currently makes two calls for each request to MDS, due to an extra slash in the URL. You can see this in logging messages from reqwest:

```
redirecting 'http://metadata.google.internal//computeMetadata/v1/instance/service-accounts/default/?recursive=true' to 'http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/?recursive=true'
```

or

```
[...]

PR #1886, opened 2025-04-22
fix(auth): remove trailing slash from Metadata root

harkamaljot:
This PR removes the trailing slash from the default Metadata endpoint (`METADATA_ROOT `).

**Reason:**

* The trailing slash caused multiple calls to MDS, see: https://github.com/googleapis/google-cloud-rust/issues/1835

**Implementation Details:**

* Updated the `METADATA_ROOT` to remove trailing slash
* Added `url` crate to `dev-dependencies`.
* Added a unit test using the `url` crate to validate the format of the default URL(s): for fetching token and service account.

### DIFF
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -53,3 +53,4 @@ tempfile    = "3"
 test-case   = "3"
 tokio       = { version = "1", features = ["macros", "rt-multi-thread", "test-util"] }
 tokio-test  = "0.4"
+url = "2"
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -53,4 +53,4 @@ tempfile    = "3"
 test-case   = "3"
 tokio       = { version = "1", features = ["macros", "rt-multi-thread", "test-util"] }
 tokio-test  = "0.4"
-url = "2"
+url         = "2"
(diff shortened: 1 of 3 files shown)


src/auth/Cargo.toml
```diff
@@ -53,3 +53,4 @@ tempfile    = "3"
 test-case   = "3"
 tokio       = { version = "1", features = ["macros", "rt-multi-thread", "test-util"] }
 tokio-test  = "0.4"
+url = "2"
```
dbolduc, line 56: linter will probably complain

dbolduc (approved):

src/auth/src/credentials/mds.rs
```diff
@@ -217,10 +218,7 @@ struct MDSAccessTokenProvider {
 impl MDSAccessTokenProvider {
     async fn get_service_account_info(&self, client: &Client) -> Result<ServiceAccountInfo> {
         let request = client
-            .get(format!(
-                "{}/computeMetadata/v1/instance/service-accounts/default/",
```
sai-sunder-s, line 221: There is a trailing slash in the end here, but it is not present in the new format. was that intentional?

sai-sunder-s (changes requested):
Please update comments on builder's `with_endpoint` method to mention whether the user must provide a trailing slash or not in the endpoint.

src/auth/src/credentials/mds.rs
harkamaljot: yes because I wanted to be consistent i.e. slashes should be in the beginning so for instance now the token url is {MDS_URI}{MDS_DEFAULT_URI}/token. Also, slash won't be needed here so this should be a no-op.

sai-sunder-s (approved):

Issue #1767, opened 2025-04-10
Fix serialization / deserialization for floating point special values


coryan:
Serialization of floating point requires handling `"NaN"`, `"Infinity"`, and `"-Infinity"`:

https://protobuf.dev/programming-guides/json/

That probably means using a custom function (in wkt) as we do for int64.

- [x] Write the custom function and some tests in src/wkt
- [x] Modify the generate to use the custom function

PR #1893, opened 2025-04-22
impl(fix): add serialize and deserialize custom functions for inf, -inf, and nan floats

suzmue:
Support the special strings for positive infinity, negative infinity, and not-a-number as specified in https://protobuf.dev/programming-guides/json/.

For #1767


src/wkt/src/wrappers.rs
```diff
@@ -215,6 +221,47 @@ impl crate::message::Message for Int64Value {
...
+        match val {
+            // Decode special strings, see https://protobuf.dev/programming-guides/json/.
+            serde_json::Value::String(string) => match string.as_str() {
+                "Infinity" => Ok(f32::INFINITY),
+                "-Infinity" => Ok(f32::NEG_INFINITY),
+                "NaN" => Ok(f32::NAN),
+                _ => crate::message::from_value::<Self>(map),
+            },
+            _ => crate::message::from_value::<Self>(map),
+        }
```
suzmue, line 261: Would love a recommendation of a better way to write this. In this way I have main code path duplicated which I think could be improved.

src/wkt/src/wrappers.rs
```diff
@@ -286,6 +332,31 @@ mod test {
         Ok(())
     }
 
+    #[test_case(f32::INFINITY as FloatValue, "Infinity", "FloatValue")]
+    #[test_case(f32::NEG_INFINITY as FloatValue, "-Infinity", "FloatValue")]
+    #[test_case(f32::NAN as FloatValue, "NaN", "FloatValue")]
+    #[test_case(9876.5 as FloatValue, 9876.5, "FloatValue")]
+    fn test_wrapper_float<V>(input: FloatValue, value: V, typename: &str) -> Result
```
suzmue, line 339: To reuse this function for DoubleValue, it looks like we could us the num::Float trait instead of FloatValue: https://docs.rs/num/latest/num/trait.Float.html. But we would need to add a dependency, I think. Is there a better way to do this?

coryan: I think you can define your own trait in this file.  Argh, my computer is rebooting... got to go.

coryan (commented):
I would start with plain serde serialization / deserialization functions:

```rust
    fn float_serialize<S>(x: &f32, s: S) -> std::result::Result<S::Ok, S::Error>
    where
        S: serde::ser::Serializer,
    {
        if x.is_nan() {
            return s.serialize_str("NaN");
        }
        s.serialize_f32(*x)
    }

    struct FloatVisitor;

    fn float_deserialize<'de, D>(deserializer: D) -> std::result::Result<f32, D::Error>
    where
        D: serde::de::Deserializer<'de>,
    {
        deserializer.deserialize_any(FloatVisitor)
    }

    impl serde::de::Visitor<'_> for FloatVisitor {
        type Value = f32;

        fn visit_str<E>(self, value: &str) -> std::result::Result<Self::Value, E> 
        where E: serde::de::Error {
            match value {
                "NaN" => Ok(f32::NAN),
                "Infinity" => Ok(f32::INFINITY),
                _ => Err(serde::de::Error::invalid_value(
                    serde::de::Unexpected::Other(&value.to_string()),
                &"a valid ProtoJSON string for f32 (NaN, Infinity, -Infinity)")),
            }
        }

        fn visit_f32<E>(self, value: f32) -> std::result::Result<Self::Value, E> 
        where E: serde::de::Error {
            Ok(value)
        }

        fn expecting(&self, formatter: &mut std::fmt::Formatter) -> std::fmt::Result {
            formatter.write_str("a 32-bit floating point in ProtoJSON format")
        }
    }
```

That looks like a lot of code. OTOTH,  (I *think* you need to check my math!) we can reuse that for `impl Mesage for FloatValue` and also to implement the serialization of each field.  For the second part, I would take inspiration in:

https://github.com/googleapis/google-cloud-rust/blob/e8b3dede996e018a33f7d6ce15b7094bd49bcc47/src/generated/api/types/src/model.rs#L2938-L2943

Note how we can add attributes to a field choosing the serialization function. In this case I would say:

```rust
    #[derive(Clone, Debug, Default, serde::Serialize, serde::Deserialize)]
    #[serde(default, rename_all = "camelCase")]
    pub struct Example {
        #[serde(serialize_with = "float_serialize", deserialize_with = "float_deserialize")]
        with_nan: f32,

        normal: f32,
    }
```

But I am not sure if that approach works well with fields of type `Vec<f32>` and `HashMap<_, f32>`. You would need to learn more about how `serde_with(as = ` can be customized.

Keep in mind that my code does not have tests or work for `Infinity` and probably has many other problems.

src/wkt/src/float.rs
```diff
@@ -0,0 +1,128 @@
...
+
+    fn visit_f32<E>(self, value: f32) -> std::result::Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        Ok(value)
+    }
+
+    // Serialized f32 with the serde_json serializer is then f64 when we try to deserialize.
+    fn visit_f64<E>(self, value: f64) -> std::result::Result<Self::Value, E>
```
suzmue, line 68: CL currently a WIP, but noticed while testing that roundtrip with serde_json::value::Serializer did not work because all floats were represented as f64. Do you think that this should work or I should just rewrite the test around it? @coryan

coryan: > CL currently a WIP, but noticed while testing that roundtrip with serde_json::value::Serializer did not work because all floats were represented as f64. 

TIL.  I suspect for JSON that is true.  `serde` supports other serialization formats, where receiving `f32` and `f64` might be possible.

> Do you think that this should work or I should just rewrite the test around it? @coryan

I think we have two options:

- Write the code with `visit_f64`.  If this fails with another serialization format, we treat that as a bug when it happens and fix it then.
- Write the code with both `visit_f32` and `visit_f64`. If you find an easy way to test the f32 path (that requires some research), do so. If not, leave a comment and we can add the testing when we learn more about serde.


PS: Consider marking the PR as "draft" to avoid drive-by reviews?

suzmue: TY for "draft" tip. Done.

Will TAL at serde a bit more to see if I could easily use a different format for testing f32

src/wkt/src/wrappers.rs
suzmue: Ack, will keep this in mind when expanding to doubles :)

src/wkt/src/float.rs
```diff
@@ -0,0 +1,137 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// A helper to serialize `f32` to ProtoJSON format.
+// TODO(#1767): use float_serialize from generated code.
+#[allow(dead_code)]
+fn float_serialize<S>(x: &f32, s: S) -> std::result::Result<S::Ok, S::Error>
```
coryan, line 18: I think:

```suggestion
pub fn float_serialize<S>(x: &f32, s: S) -> std::result::Result<S::Ok, S::Error>
```

should work.

src/wkt/src/float.rs
```diff
@@ -0,0 +1,137 @@
...
+    if x.is_nan() {
+        return s.serialize_str("NaN");
+    }
+    if x.is_infinite() {
+        if x.is_sign_negative() {
+            return s.serialize_str("-Infinity");
+        }
+        return s.serialize_str("Infinity");
+    }
+    s.serialize_f32(*x)
```
coryan, line 32: Consider:

```suggestion
    // Handle special strings, see https://protobuf.dev/programming-guides/json/.
    match x {
      x if x.is_nan() => s.serialize_str("NaN"),
      x if x.is_infinite() && x.is_sign_negative => s.serialize_str("-Infinity"),
      x if x.is_infinite() => s.serialize_str("Infinity"),
      x => s.serialize_f32(*x),
    }
```

src/wkt/src/float.rs
```diff
@@ -0,0 +1,137 @@
...
+}
+
+#[allow(dead_code)]
+struct FloatVisitor;
+
+/// A helper to deserialize `f32` from ProtoJSON format.
+// TODO(#1767): use float_serialize from generated code.
+#[allow(dead_code)]
+fn float_deserialize<'de, D>(deserializer: D) -> std::result::Result<f32, D::Error>
```
coryan, line 41: Likewise, I think this should work:

```suggestion
struct FloatVisitor;

/// A helper to deserialize `f32` from ProtoJSON format.
// TODO(#1767): use float_serialize from generated code.
pub fn float_deserialize<'de, D>(deserializer: D) -> std::result::Result<f32, D::Error>
```

src/wkt/src/float.rs
```diff
@@ -0,0 +1,137 @@
...
+                &"a valid ProtoJSON string for f32 (NaN, Infinity, -Infinity)",
+            )),
+        }
+    }
+
+    fn visit_f32<E>(self, value: f32) -> std::result::Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        // TODO: Find a way to test this code path, serde_json floats
```
coryan, line 71: We always use `TODO(# <GitHubIssueNumber>)` sometimes `TODO(#...)` while writing the PR if we are not certain about the issue number.

src/wkt/src/float.rs
```diff
@@ -0,0 +1,137 @@
...
+        // stored as f64.
+        Ok(value)
+    }
+
+    // Floats in serde_json are f64.
+    fn visit_f64<E>(self, value: f64) -> std::result::Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        Ok(value as f32)
```
coryan, line 81: This is correct, but I had to look it up:

https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric.float-narrowing

Maybe a comment?

src/wkt/src/lib.rs
```diff
@@ -31,6 +31,7 @@ mod empty;
 pub use crate::empty::*;
 mod field_mask;
 pub use crate::field_mask::*;
+mod float;
```
coryan, line 34: This will need to be a `pub mod float` sooner or later.

coryan: I wonder if it should be `internal` since we do not expect application developers to call this API.

coryan:
```suggestion
#[doc(hidden)]
pub mod internal;
```

coryan (approved):
A few nits, but otherwise good. Please fix them and then merge.

src/wkt/src/float.rs
suzmue: This + making the module public worked.

src/wkt/src/float.rs
suzmue: This + making the module public worked.

src/wkt/src/float.rs
suzmue: Nice! Thanks for the suggestion :)

src/wkt/src/lib.rs
suzmue: Done.

src/wkt/src/float.rs
suzmue: Done.

src/wkt/src/float.rs
suzmue: Done.

coryan (approved):

PR #1906, opened 2025-04-24
refactor(wkt): DRY for `impl Message`

dbolduc:
Tangential to #1609, #1767 

There are only 2 cases for encoding, either we put the serialization of something into a `value` field, or we flatten it.

So use helper functions for the two cases and remove instances of logic.

While we're here, write up my understanding of why the `Any` business is the way it is.

### DIFF
--- a/src/wkt/src/message.rs
+++ b/src/wkt/src/message.rs
@@ -45,11 +45,10 @@ pub trait Message {
 
 /// Write the serialization of `T` flatly into a map.
 ///
-/// This is used for types where the JSON serialization of `T` is an object.
+/// We use this for types that do not have special encodings, as defined in:
+/// https://protobuf.dev/programming-guides/json/
 ///
-/// `T` is a message with fields that cannot be named `@type`, so flattening the
-/// object saves some bytes over the wire (as compared with encoding the
-/// serialization into a `value` field).
+/// That typically means that `T` is an object.
 pub(crate) fn to_json_object<T>(message: &T) -> Result<Map, Error>
 where
     T: Message + serde::ser::Serialize,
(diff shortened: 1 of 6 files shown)


src/wkt/src/message.rs
```diff
@@ -39,19 +39,17 @@ pub trait Message {
...
     }
 }
 
+/// Write the serialization of `T` flatly into a map.
+///
+/// This is used for types where the JSON serialization of `T` is an object.
+///
+/// `T` is a message with fields that cannot be named `@type`, so flattening the
+/// object saves some bytes over the wire (as compared with encoding the
+/// serialization into a `value` field).
```
coryan, line 52: I am not sure that is the motivation. We are just trying to comply with the ProtoJSON spec:

https://protobuf.dev/programming-guides/json/

src/wkt/src/message.rs
```diff
@@ -71,28 +69,44 @@ where
...
 
-pub(crate) fn to_json_string<T>(message: &T) -> Result<Map, Error>
+/// Write the serialization of `T` into the `value` field of a map.
+///
+/// This is used for types where the JSON serialization of `T` is not an
+/// object (e.g. bool, number, string, etc.). The encoding is a map, and these
+/// values need a key.
+///
+/// It is also used for `Any`. Flatly serializing an `Any` would have
+/// conflicting `@type` fields.
```
coryan, line 79: I think the motivation is to match the ProtoJSON serialization which defines special encodings for Duration, Timestamp, Any, and the `*Value` types (which look like `bool`, `number`, etc. in Rust).

coryan (approved):
I think you should reference the ProtoJSON spec in the comments more than performance motivations.

src/wkt/src/message.rs
dbolduc: I think I was trying to explain why the proto spec is the way it is. I will just defer to the spec.

Issue #1846, opened 2025-04-18
Use routing annotation is Rust generated code


coryan:

coryan:
This is now done.

PR #1908, opened 2025-04-24
impl(gax-internal): improve routing info matchers

coryan:
Use a Rust enum to avoid `if x == "**"`, the generator can convert the
wildcards into branches of the enum ahead of time. Also, the generator
can use a special branch for "trailing" multi-segment wildcards. Those
have slightly different matching rules.

Part of the work for #1846


dbolduc (approved):
just nits

Issue #1843, opened 2025-04-18
Consider using idiomatic Rust naming for CredentialsTrait in credentials


chrischiedo:
I'm not 100% sure about this, but I don't think it's _idiomatic_ Rust for a trait name to have a `Trait` suffix (like it's the case with [CredentialsTrait](https://github.com/googleapis/google-cloud-rust/blob/2646b6c9bfb0aa7d9acb34c2f03e15cd88da28c2/src/auth/src/credentials.rs#L143) in [credentials.rs](https://github.com/googleapis/google-cloud-rust/blob/main/src/auth/src/credentials.rs)). While I see the reason why the trait has it's current name (in this particular case), I still think it would be better to follow Rust's convention.

**Potential solutions**:

1. Maintain the current name and be consistent with the naming convention throughout, i.e. adding the `Trait` suffix to all trait names -- this would also require that the decision is well documented _somewhere_.

1. Change the name of the trait (see suggested alternatives below).

**Alternative naming suggestions**:

**Option 1**:

```rust
pub trait AuthCredentials: std::fmt::Debug {
    //...
    fn token(&self) -> impl Future<Output = Result<crate::token::Token>> + Send;

    //...
    fn headers(&self) -> impl Future<Output = Result<Vec<(HeaderName, HeaderValue)>>> + Send;

    //...
    fn universe_domain(&self) -> impl Future<Output = Option<String>> + Send;
}
```
Usage:

```rust
pub struct Credentials {
    //...
    inner: Arc<dyn dynamic::AuthCredentials>,
}

impl AuthCredentials for TestCredentials {
   // ...
}
```

**Option 2**:

```rust
pub trait CredentialsProvider: std::fmt::Debug {
    //...
    fn token(&self) -> impl Future<Output = Result<crate::token::Token>> + Send;

    //...
    fn headers(&self) -> impl Future<Output = Result<Vec<(HeaderName, HeaderValue)>>> + Send;

    //...
    fn universe_domain(&self) -> impl Future<Output = Option<String>> + Send;
}
```

Usage:

```rust
pub struct Credentials {
    //...
    inner: Arc<dyn dynamic::CredentialsProvider>,
}

impl CredentialsProvider for TestCredentials {
   // ...
}
```
cc/ @sai-sunder-s

sai-sunder-s:
Thank you for raising this issue @chrischiedo . Will update the trait name to be more idiomatic in Rust.

PR #1909, opened 2025-04-24
refactor(auth)!: rename CredentialsTrait to CredentialsProvider

harkamaljot:
As mentioned in https://github.com/googleapis/google-cloud-rust/issues/1843 updating the name of CredentialsTrait to CredentialsProvider to make it more rusty.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -288,7 +288,7 @@ tonic       = { version = "0.13", default-features = false, features = ["prost",
 tonic-build = { version = "0.13", default-features = false }
 tracing     = { version = "0.1" }
 # Local packages used as dependencies
-auth                          = { version = "0.20", path = "src/auth", package = "google-cloud-auth" }
+auth                          = { version = "0.21", path = "src/auth", package = "google-cloud-auth" }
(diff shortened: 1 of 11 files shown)


sai-sunder-s (approved):

Cargo.toml
```diff
@@ -288,7 +288,7 @@ tonic       = { version = "0.13", default-features = false, features = ["prost",
 tonic-build = { version = "0.13", default-features = false }
 tracing     = { version = "0.1" }
 # Local packages used as dependencies
-auth                          = { version = "0.20", path = "src/auth", package = "google-cloud-auth" }
+auth                          = { version = "0.21", path = "src/auth", package = "google-cloud-auth" }
```
dbolduc, line 291: FYI: we only need to bump the auth version once per release. Bumping it multiple times is harmless though.

(The version bump process is very manual. We just keep track of things in our heads. We would like to refine the process, but it hasn't been a priority yet.)

dbolduc: Also, thanks for using `!` in the commit title

PR #1911, opened 2025-04-25
feat(generator/rust): gRPC clients use routinginfo

coryan:
With this change, the generator produces code to populate the
`x-goog-request-params` header using the routing info, if available.
A separate PR will change the gaxi::grpc::Client to ignore empty
headers.

Part of the work for #1846

### DIFF
--- a/src/storage/src/generated/gapic/transport.rs
+++ b/src/storage/src/generated/gapic/transport.rs
@@ -73,7 +73,16 @@ impl super::stub::Storage for Storage {
             e
         };
         let path = http::uri::PathAndQuery::from_static("/google.storage.v2.Storage/DeleteBucket");
-        let x_goog_request_params = [""; 0].into_iter().fold(String::new(), |b, p| b + "&" + &p);
+        let x_goog_request_params = {
+            use gaxi::routing_parameter::Segment;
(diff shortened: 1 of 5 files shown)


src/storage/src/generated/gapic/transport.rs
```diff
@@ -381,7 +476,24 @@ impl super::stub::Storage for Storage {
...
+            gaxi::routing_parameter::format(&[gaxi::routing_parameter::value(
+                Some(&req).map(|v| v.upload_id.as_str()),
+                &[],
+                &[
+                    Segment::Literal("projects"),
+                    Segment::Literal("/"),
+                    Segment::SingleWildcard,
+                    Segment::Literal("/"),
+                    Segment::Literal("buckets"),
+                    Segment::Literal("/"),
```
dbolduc, line 490: super nit: the `routing_parameter::value` tests had this as a single segment: `Literal("/buckets/")`. If you care enough, you could make the tests look more like the generated code.

dbolduc (approved):

src/storage/src/generated/gapic/transport.rs
coryan: You are right, I meant to add a feature request to optimize this and I forgot: #1919.

PR #1913, opened 2025-04-25
feat(gax): return default value for struct for 204 NoContent responses

alvarowolfx:
In a typical AIP compliant library, successful delete operations return google.protobuf.Empty as their response [AIP-135](https://google.aip.dev/135). When in REST Fallback mode  this means a response would be `{}` - empty JSON. However, certain services return `null` with a 204 `No Content` response, instead of an empty JSON response. This behavior happens with BigQuery and Compute for example. This means that with certain RPCs (mainly some delete operations), users using BigQuery or Compute in REST mode will get an error returned when the call actually succeeded on the call side.

`serde_json` throws an error when trying to deserialize with an empty byte array:
```
Error("EOF while parsing a value", line: 1, column: 0)
```

This PR refactor the `reqwest::Response` parsing to a separated method and handles the case for `204` responses with no body, returning the `Default` value for the target struct.

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -373,32 +373,46 @@ mod test {
     }
 
     #[tokio::test]
-    #[test_case(reqwest::StatusCode::OK, "{}", false; "200 with empty object")]
-    #[test_case(reqwest::StatusCode::NO_CONTENT, "{}", false; "204 with empty object")]
-    #[test_case(reqwest::StatusCode::OK, "", true; "200 with empty content")]
-    #[test_case(reqwest::StatusCode::NO_CONTENT, "", false; "204 with empty content")]
-    async fn client_empty_content(
-        code: reqwest::StatusCode,
-        content: &str,
-        should_fail: bool,
-    ) -> TestResult {
-        let http_resp = http::Response::builder()
-            .header("Content-Type", "application/json")
-            .status(code)
-            .body(content.to_owned())?;
-        let response: reqwest::Response = http_resp.into();
+    #[test_case(reqwest::StatusCode::OK, "{}"; "200 with empty object")]
+    #[test_case(reqwest::StatusCode::NO_CONTENT, "{}"; "204 with empty object")]
+    #[test_case(reqwest::StatusCode::NO_CONTENT, ""; "204 with empty content")]
+    async fn client_empty_content(code: reqwest::StatusCode, content: &str) -> TestResult {
+        let response = resp_from_code_content(code, content)?;
         assert!(response.status().is_success());
 
         let response = ReqwestClient::to_http_response::<wkt::Empty>(response).await;
-        if should_fail {
-            assert!(response.is_err());
-            return Ok(());
-        }
         assert!(response.is_ok());
 
         let response = response.unwrap();
         let body = response.into_body();
         assert_eq!(body, wkt::Empty::default());
         Ok(())
+    }
+
+    #[tokio::test]
+    #[test_case(reqwest::StatusCode::OK, ""; "200 with empty content")]
+    async fn client_error_with_empty_content(
+        code: reqwest::StatusCode,
+        content: &str,
+    ) -> TestResult {
+        let response = resp_from_code_content(code, content)?;
+        assert!(response.status().is_success());
+
+        let response = ReqwestClient::to_http_response::<wkt::Empty>(response).await;
+        assert!(response.is_err());
+        Ok(())
+    }
+
+    fn resp_from_code_content(
+        code: reqwest::StatusCode,
+        content: &str,
+    ) -> http::Result<reqwest::Response> {
+        let http_resp = http::Response::builder()
+            .header("Content-Type", "application/json")
+            .status(code)
+            .body(content.to_string())?;
+
+        let response: reqwest::Response = http_resp.into();
+        Ok(response)
     }
 }
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -137,7 +137,7 @@ impl ReqwestClient {
             return Self::to_http_error(response).await;
         }
 
-        return Self::to_http_response(response).await;
+        Self::to_http_response(response).await
     }
 
     async fn to_http_error<O>(response: reqwest::Response) -> Result<O> {
@@ -168,7 +168,7 @@ impl ReqwestClient {
             .map_err(Error::io)?;
 
         let response = match body.to_bytes() {
-            content if (content.len() == 0 && is204ok) => O::default(), // 204 No Content has no body, throws EOF error if we try to parse
+            content if (content.is_empty() && is204ok) => O::default(), // 204 No Content has no body, throws EOF error if we try to parse
             content => serde_json::from_slice::<O>(&content).map_err(Error::serde)?,
         };
 
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -257,6 +257,7 @@ const SENSITIVE_HEADER: &str = "[sensitive]";
 mod test {
     use super::*;
     use std::collections::HashMap;
+    use test_case::test_case;
     type TestResult = std::result::Result<(), Box<dyn std::error::Error>>;
 
     #[test]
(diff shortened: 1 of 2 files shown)


coryan (commented):
I wonder if the best way to handle this is to teach the generator about the clients with weird behavior, rather than tweaking the deserialization for `Empty`.

It would be nice if we could assert: "`wkt::Empty` performs normal serialization and deserialization for  [ProtoJSON](https://protobuf.dev/programming-guides/json/), including validation (it only accepts valid serialization)".

alvarowolfx:
> teach the generator about the clients with weird behavior, rather than tweaking the deserialization for `Empty`.

I was having the same questions here, in Node for example we actually changed the gax layer to have a flag set for a given service and check for this behavior just for particular services that enabled that. I couldn't find a way to do some similar to that in the Rust SDK, but it's probably because I need to get more familiar with the codebase. 

In Go this is handled by the generator, basically RPC that returns Empty, it just skips deserialization.

alvarowolfx:
Thanks for all the pointers @coryan and also as you mentioned things here, I just found out about this `ReturnsEmpty` property that is already doing some logic around `Empty` types. So maybe we just need some improvements on that to make it work with BigQuery/Compute.

This is what is being done in the Go generator, which just skips decode the response when it's the Empty type:
* https://github.com/googleapis/gapic-generator-go/blob/b3fc5aa68a7da1f0fc11cddc1b64e75bea22f525/internal/gengapic/genrest.go#L570
* https://github.com/googleapis/gapic-generator-go/blob/b3fc5aa68a7da1f0fc11cddc1b64e75bea22f525/internal/gengapic/genrest.go#L1048

coryan:
> This is what is being done in the Go generator, which just skips decode the response when it's the Empty type:

I read your comment.  it feels wrong to ignore the return body.  If we know the response should be `{}` and it is not, then somebody bad is going on. We should report that to the application.

alvarowolfx:
> > This is what is being done in the Go generator, which just skips decode the response when it's the Empty type:
> 
> I read your comment. it feels wrong to ignore the return body. If we know the response should be `{}` and it is not, then somebody bad is going on. We should report that to the application.

You have a point. But we still check the status code, so if the response is ok and body is _suppose_ to be empty, then I think it's fine. When there is an error, we parse the body as expected.

coryan:
> 
> You have a point. But we still check the status code, so if the response is ok and body is _suppose_ to be empty, then I think it's fine. When there is an error, we parse the body as expected.

Yes. The case I was considering is getting a ok response code, expecting a `{}` in the response body, and yet receiving an empty response body.  That is a bug somewhere. We should not be hiding it.

alvarowolfx:
> Yes. The case I was considering is getting a ok response code, expecting a `{}` in the response body, and yet receiving an empty response body. That is a bug somewhere. We should not be hiding it.

That's exactly what BigQuery and Compute (and some other APIs that we don't generate clients) are doing 😬 It is a problem on the service side + Apiary shenanigans, more details can be seen on the internal b/411675301. Sadly, if we don't do something about that behavior, users will get a deserialization error, even when the call succeeded.

coryan:
> > Yes. The case I was considering is getting a ok response code, expecting a `{}` in the response body, and yet receiving an empty response body. That is a bug somewhere. We should not be hiding it.
> 
> That's exactly what BigQuery and Compute (and some other APIs that we don't generate clients) are doing 😬 It is a problem on the service side + Apiary shenanigans, more details can be seen on the internal b/411675301. 

Yes, I do realize that. Are you suggesting that we ignore the payload because other clients are doing that?  I frankly do not find such arguments compelling *at all*.

> Sadly, if we don't do something about that behavior, users will get a deserialization error, even when the call succeeded.

The "success" is a lie.  If it was successful it would have returned the correct body as well as having the correct return code.  This is no different than calling `GetObjectMetadata` getting a success and an empty body, or a JSON object for a bucket. 

The service is broken (or something), ignoring the problem is not the "Right Thing"[tm] to do.

alvarowolfx:
> > > Yes. The case I was considering is getting a ok response code, expecting a `{}` in the response body, and yet receiving an empty response body. That is a bug somewhere. We should not be hiding it.
> > 
> > 
> > That's exactly what BigQuery and Compute (and some other APIs that we don't generate clients) are doing 😬 It is a problem on the service side + Apiary shenanigans, more details can be seen on the internal b/411675301.
> 
> Yes, I do realize that. Are you suggesting that we ignore the payload because other clients are doing that? I frankly do not find such arguments compelling _at all_.
> 
> > Sadly, if we don't do something about that behavior, users will get a deserialization error, even when the call succeeded.
> 
> The "success" is a lie. If it was successful it would have returned the correct body as well as having the correct return code. This is no different than calling `GetObjectMetadata` getting a success and an empty body, or a JSON object for a bucket.
> 
> The service is broken (or something), ignoring the problem is not the "Right Thing"[tm] to do.

and what do you suggest doing ?

coryan:
> and what do you suggest doing ? Changing that on the service side is a breaking change

For services that we know they should return `{}`, then we should expect that, and return a deserialization error when they return anything else.

For services that we know they should return `null`, then we should expect that, and return a deserialization error when they return anything else.

For services that we know they should return an empty body, then we should expect that, and return a deserialization error when they return anything.

If there are services that may return any of those things (hopefully not at random!), the we can discuss alternatives, but I would be inclined to accept the minimum set of possible outputs, e.g, only `{}`, `null`, or an empty payload and reject anything else.

alvarowolfx:
Changed approach on PR to handle handles the case for 204 responses with no body, returning the Default value for the target struct.

src/gax-internal/src/http.rs
```diff
@@ -165,6 +156,28 @@ impl ReqwestClient {
         Err(error)
     }
 
+    async fn to_http_response<O: serde::de::DeserializeOwned + Default>(
+        response: reqwest::Response,
+    ) -> Result<Response<O>> {
+        let is204ok = response.status() == 204;
```
coryan, line 162: Consider https://docs.rs/reqwest/latest/reqwest/struct.StatusCode.html#associatedconstant.NO_CONTENT, also the naming convention in Rust is `snake_case`, this variable should be `is_204_ok`.  I would recommend `no_content_status`.

src/gax-internal/src/http.rs
```diff
@@ -165,6 +156,28 @@ impl ReqwestClient {
...
+        let response = http::Response::from(response);
+        let (parts, body) = response.into_parts();
+
+        let body = http_body_util::BodyExt::collect(body)
+            .await
+            .map_err(Error::io)?;
+
+        let response = match body.to_bytes() {
+            content if (content.is_empty() && is204ok) => O::default(), // 204 No Content has no body, throws EOF error if we try to parse
```
coryan, line 171: Consider moving the comment to the variable definition.

src/gax-internal/src/http.rs
```diff
@@ -356,4 +369,22 @@ mod test {
...
+        assert!(response.status().is_success());
+
+        let response = ReqwestClient::to_http_response::<wkt::Empty>(response).await;
+        assert!(response.is_ok());
+
+        let response = response.unwrap();
+        let body = response.into_body();
+        assert_eq!(body, wkt::Empty::default());
+        Ok(())
+    }
```
coryan, line 389: Do we have a test for 200 with body == `{}`?  What about a 204 with a non-empty body?

coryan (approved):
I think we are missing a couple of tests and I left some nits.

We should improve this to make some of these decisions at compile time.  I believe that can happen in a follow up PR though (just open a tracking bug).

The basic idea would be to define a new special type `NoContent` that is received in the `O` parameter instead of `wkt::Empty`.  As you note the serde deserialization would not work, but we could do something like:

```rust
#[derive(Clone, Debug)] 
pub struct NoContent;

trait DeserializeResponse {
  fn deserialize(response: reqwest::Response) -> Result<Self>;
}

impl DeserializeResponse for NoContent {
  fn deserialize(response: reqwest::Response) -> Result<Self> {
    if 204 && content.is_empty() {
        return Ok(NoContent);
    }
    Err(blah blah
  }
}

impl<T> DeserializeResponse for T where T: serde::de::DeserializeOwned {
  fn deserialize(...)  ... {
    // Do the stuff we were doing before
  }
}

```

Then the generated code can replace `wkt::Empty` with `gaxi::NoContent` in the right places.

src/gax-internal/src/http.rs
```diff
@@ -370,21 +371,33 @@ mod test {
...
 
-    #[tokio::test]
-    async fn client_empty_content() -> TestResult {
-        let http_resp = http::Response::builder()
-            .header("Content-Type", "application/json")
-            .status(204)
-            .body(r#""#)?;
-        let response: reqwest::Response = http_resp.into();
-        assert!(response.status().is_success());
+    macro_rules! test_client_empty_content {
```
coryan, line 374: Cool use of macros, but probably not the right tool here.  I think you can do this with [test_case](https://crates.io/crates/test_case) you should be able to find multiple examples in the code.

coryan (changes requested):

src/gax-internal/src/http.rs
alvarowolfx: ah interesting, I was trying to find ways to do table-driven tests and found this method with macros
* https://users.rust-lang.org/t/table-driven-aka-data-driven-testing/3848/2
* https://github.com/BurntSushi/fst/blob/master/src/raw/tests.rs#L111-L158

But I'll check `test_case` crate, thanks

src/gax-internal/src/http.rs
```diff
@@ -371,33 +372,33 @@ mod test {
...
-                assert!(response.is_ok());
+    #[tokio::test]
+    #[test_case(reqwest::StatusCode::OK, "{}", false; "200 with empty object")]
+    #[test_case(reqwest::StatusCode::NO_CONTENT, "{}", false; "204 with empty object")]
+    #[test_case(reqwest::StatusCode::OK, "", true; "200 with empty content")]
+    #[test_case(reqwest::StatusCode::NO_CONTENT, "", false; "204 with empty content")]
+    async fn client_empty_content(
+        code: reqwest::StatusCode,
+        content: &str,
+        should_fail: bool,
```
coryan, line 383: optional: go/tott/661

Consider splitting this into a test function for the `should_fail == true` base and  one for the `should_fail == false` case.  I know it is also bad to refactor test code, but a function to create the `response` would make those two functions pretty small.

coryan (approved):

src/gax-internal/src/http.rs
alvarowolfx: done

coryan (approved):

coryan:
Do you have enough permissions to merge this?  If not, I can do it for you.  I am going to ignore the multi_approvers, that is a bug (you are a Googler, it should pass).

alvarowolfx:
> Do you have enough permissions to merge this?  If not, I can do it for you.  I am going to ignore the multi_approvers, that is a bug (you are a Googler, it should pass).

yeah, I have the necessary permissions. I was just double-checking that I covered most of the points you raised before merging.

coryan:
Separately, we prefer to use forks for development in this repo.

src/gax-internal/src/http.rs
```diff
@@ -356,4 +371,48 @@ mod test {
...
+    #[test_case(reqwest::StatusCode::NO_CONTENT, "{}"; "204 with empty object")]
+    #[test_case(reqwest::StatusCode::NO_CONTENT, ""; "204 with empty content")]
+    async fn client_empty_content(code: reqwest::StatusCode, content: &str) -> TestResult {
+        let response = resp_from_code_content(code, content)?;
+        assert!(response.status().is_success());
+
+        let response = ReqwestClient::to_http_response::<wkt::Empty>(response).await;
+        assert!(response.is_ok());
+
+        let response = response.unwrap();
```
dbolduc, line 386: nit: in general we prefer `?` over `unwrap()`. Not a big deal at all in tests, but we use `?` above, so why not use it here too.

dbolduc (approved):
Looks good

PR #1914, opened 2025-04-25
test(bigquery): add integration tests for dataset admin operations 

alvarowolfx:
Add basic admin operation on BigQuery `Datasets`. Later the method `cleanup_stale_datasets` can be reused to clean up resources on future BigQuery integration tests, as a `Dataset` is the highest level in the hierarchy and delete operations can delete all child resources.

Towards #1773


src/integration-tests/src/bigquery.rs
```diff
@@ -0,0 +1,145 @@
...
+    let client = builder.build().await?;
+    cleanup_stale_datasets(&client, &project_id).await?;
+
+    let rand_suffix: String = rand::rng()
+        .sample_iter(&Alphanumeric)
+        .take(8)
+        .map(char::from)
+        .collect();
+
+    let ds_name = format!("rust_bq_test_dataset_{rand_suffix}");
```
coryan, line 46: In general we prefer using labels to identify test resources.  Can we do that here?

src/integration-tests/src/bigquery.rs
```diff
@@ -0,0 +1,145 @@
...
+    println!("LIST DATASET = {} entries", list.datasets.len());
+
+    assert!(!list.datasets.is_empty());
+    assert!(list.datasets.len() > 1);
+    assert!(
+        list.datasets
+            .iter()
+            .find(|v| v.id.contains(&ds_name))
+            .is_some()
+    );
```
coryan, line 71: All of this can be changed to:

```suggestion
    assert!(
        list.datasets
            .iter()
            .find(|v| v.id.contains(&ds_name))
            .is_some(),
            "{:?}", list.datasets
    );
```

Separately, why "contains"? Can we make that a more specific predicate?

src/integration-tests/src/bigquery.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    let list = client.list_datasets(project_id).send().await?;
+    let pending_all_datasets = list
+        .datasets
+        .iter()
+        .map(|v| {
+            return client
+                .get_dataset(
+                    project_id,
+                    v.dataset_reference.as_ref().map_or("", |v| &v.dataset_id),
```
coryan, line 104: This deserves some comment explaining what is going on.  If the `dataset_reference` is not set it seems we cannot make a successful call to `.get_dataset()`?  Maybe returning `None` is more appropriate? or maybe using `v.id` and stripping the project_id?

Maybe we should have something like:

```rust
fn extract_dataset_id(project_id: &str, v: bigquery::model::ListFormatDataset) -> Option<String> {
    match v.dataset_reference {
        Some(r) => r.dataset_id.clone,
        None => v.id.strip_prefix("projects/").strip_prefix(project_id).owned(),
    }
}
```

and then we issue the request only if we have `Some()`...

src/integration-tests/src/bigquery.rs
```diff
@@ -0,0 +1,145 @@
...
+        .into_iter()
+        .filter_map(|r| {
+            if r.as_ref()
+                .is_ok_and(|ds| ds.creation_time < stale_deadline && ds.id.contains("bq_rust"))
+            {
+                return r.ok();
+            }
+            return None;
+        })
+        .collect::<Vec<_>>();
```
coryan, line 121: This is a case where using `into_stream()` and [StreamExt](https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html) should help.

To start with, we should be able to write:

```rust
  use futures::StreamExt;
  ... ... ..

  let list = client.list_datasets(project_id).send().await?;
  futures::stream::iter(list.datasets.iter())
    .flat_map_unordered(16, // limit the number of concurrent requests 
        async |item| {
            if let Some(id) = extract_dataset_id(item) {
                return client.get_dataset(project_id, id).send().await;
            }
            None
        })
    .flat_map_unordered(16, |result| -> Option<Result<()>> {
      let result = result?;
      match result {
        Err(e) => Some(Err(e)),
        Ok(dataset) => {
            let id = extract_dataset_id(project_id, dataset.)?; // Blegh, this needs some refactoring
            if dataset.creation_time < stale_deadline && ds.labels.find("integration-test") == Some("true") {             
              return Some(client.delete_dataset(project_id, id).await);
            }
            None
        }
      }
    });
```

Finally you may need to collect all the errors.

coryan (commented):
Drive-by ...

src/integration-tests/src/bigquery.rs
alvarowolfx: the Dataset `id` contains the fully qualified id of the dataset, so it would be something like `projects/{projectId}/datasets/{datasetId}`, so that's why I used `contains`.

coryan: Yes, I was suggesting we do something more specific, such as:

```rust
  v.id.strip_suffix(&ds_name).and(Some(true)).unwrap_or_default()
```

or
```rust
  v.id == format!("projects/{project_id}/datasets/{ds_name}")
```

or (gross):

```rust
  v.id.strip_prefix("projects/").and_then(|s| s.strip_prefix(&project_id)).and_then(|s|  s.strip_prefix("/datasets/")).and_then(|s| s == &ds_name).unwrap_or_default()
```

src/integration-tests/tests/driver.rs
```diff
@@ -42,6 +42,18 @@ mod driver {
             .map_err(report)
     }
 
+    #[test_case(bigquery::client::DatasetService::builder(); "default")]
+    #[test_case(bigquery::client::DatasetService::builder().with_tracing(); "with tracing enabled")]
+    #[test_case(bigquery::client::DatasetService::builder().with_retry_policy(retry_policy()); "with retry enabled")]
```
coryan, line 47: I think we do not need all three cases for each library. One is enough, maybe this?

```suggestion
    #[test_case(bigquery::client::DatasetService::builder().with_retry_policy(retry_policy()).with_tracing(); "with retry and tracing enabled")]
```

PR #1923, opened 2025-04-25
feat(lustre): generate library

### LINKED ISSUE #1892: Generate library for `lustre.googleapis.com`
https://github.com/googleapis/googleapis/tree/master/google/cloud/lustre/v1

Instructions:

https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-generated-code-maintenance.md

dbolduc:
Fixes #1892


src/generated/cloud/lustre/v1/Cargo.toml
```diff
@@ -0,0 +1,47 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+# Code generated by sidekick. DO NOT EDIT.
+
+[package]
+name                   = "google-cloud-lustre-v1"
+version                = "0.3.0"
```
coryan, line 19: note for the future: when we are at 1.0, would we want to create these as 1.0?

src/generated/cloud/lustre/v1/.repo-metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+    "api_id": "lustre.googleapis.com",
+    "api_shortname": "lustre",
+    "client_documentation": "https://docs.rs/google-cloud-lustre-v1/latest",
+    "distribution_name": "google-cloud-rust",
+    "language": "rust",
+    "library_type": "GAPIC_AUTO",
+    "name_pretty": "Google Cloud Managed Lustre API",
+    "release_level": "preview",
```
coryan, line 9: In the future: would we want to generate clients for services that are not GA? Would we want to mark some clients as GA and not others?

coryan (approved):
Some food for thought, but it is independent of this change. Good thing that no changes to .typos.toml was required.

src/generated/cloud/lustre/v1/Cargo.toml
dbolduc: I think so. We expect them to be stable and just work.

dbolduc: Sorry, I missed your comment about this not being `"release_level": "GA"`. We probably aren't allowed to make it `1.0`.

src/generated/cloud/lustre/v1/.repo-metadata.json
dbolduc: If we want to programmatically treat `release_level != GA` differently, that is fine with me.

I veto doing anything manual.

> would we want to generate clients for services that are not GA?

I think so, if they are `v[0-9]+`. Customers may be previewing these services using the Rust libraries.

> Would we want to mark some clients as GA and not others?

I think so.

PR #1925, opened 2025-04-25
cleanup(storage)!: trim methods and messages

coryan:
We only want the admin RPCs and their messages in this client.

Part of the work for #1813


dbolduc (approved):
Just curious: names for the admin vs. data plane crates?

I assumed `google-cloud-storage` would be the data plane.

Should we rename this one to `google-cloud-storage-admin` (or `storage-control` or something)?

coryan:
> Just curious: names for the admin vs. data plane crates?
> 
> I assumed `google-cloud-storage` would be the data plane.
> 
> Should we rename this one to `google-cloud-storage-admin` (or `storage-control` or something)?

Yup. This crate should be renamed.  I updated the tracking bug.

PR #1927, opened 2025-04-25
feat(storage): initial client and integration test

coryan:
Part of the work for #1813

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -1,0 +1,301 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Contains the Storage client and related types.
+
+// TODO(#1813) - remove this once we documentation links work correctly.
+#[doc(hidden)]
+pub use super::generated::gapic::client::Storage as StorageV2Client;
+
+/// Implements a client for the Cloud Storage API.
+///
+/// # Example
+/// ```
+/// # tokio_test::block_on(async {
+/// # use google_cloud_storage::client::Storage;
+/// let client = Storage::builder().build().await?;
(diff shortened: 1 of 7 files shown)


src/storage/src/client.rs
```diff
@@ -0,0 +1,297 @@
...
+    /// Creates a new bucket.
+    ///
+    /// # Parameters
+    /// * `name` - the bucket name. In `projects/_/buckets/{bucket_id}` format.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_storage::client::Storage;
+    /// async fn example(client: &Storage) -> gax::Result<()> {
+    ///     let bucket = client.create_bucket("projects/my-project", "my-bucket").send().await?;
```
dbolduc, line 145: Are we concerned about handwritten sample code that doesn't run?

We could maybe use `#[doc(include = example.rs)` for the sample body, and run the examples in our integration tests.

Also, this is a thing: https://doc.rust-lang.org/stable/rustdoc/scraped-examples.html

(Non-blocking for this PR, but potentially something we want in the long term)

src/storage/src/client.rs
```diff
@@ -0,0 +1,297 @@
...
+    ///     }
+    ///     Ok(())
+    /// }
+    /// ```
+    pub fn list_buckets<T: Into<String>>(&self, parent: T) -> super::builder::storage::ListBuckets {
+        self.inner.list_buckets().set_parent(parent)
+    }
+
+    /// Updates a bucket. Equivalent to JSON API's storage.buckets.patch method.
+    pub fn update_bucket(&self) -> super::builder::storage::UpdateBucket {
```
dbolduc, line 193: This is a strange API. Should there at least be a required bucket field?

dbolduc: And an example?

dbolduc (commented):
Mainly curious about `update_bucket`

dbolduc (approved):

src/storage/src/client.rs
```diff
@@ -0,0 +1,296 @@
...
+///   ```
+///   Note that object names can contain `/` characters, which are treated as
+///   any other character (no special directory semantics).
+///
+/// [with_endpoint()]: super::builder::storage::ClientBuilder::with_endpoint
+/// [with_credentials()]: super::builder::storage::ClientBuilder::with_credentials
+/// [Private Google Access with VPC Service Controls]: https://cloud.google.com/vpc-service-controls/docs/private-connectivity
+/// [Application Default Credentials]: https://cloud.google.com/docs/authentication#adc
+#[derive(Clone, Debug)]
+pub struct Storage {
```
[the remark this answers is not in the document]
coryan: You should ask, I have no answer for that yet.

PR #1946, opened 2025-04-28
refactor(wkt): refactor `wkt::message::Message` trait to use new design

chrischiedo:
This PR is part of #1609.

What the PR does:

- Split up the old `Message` trait into two new traits (`Message` and `MessageSerializer<T>`)
- The `MessageSerializer<T>` is hidden, as it's an internal API that's not intended for direct use by consumers of the `wkt` crate.

**Note**:

- Even though in the initial design, we had proposed the `serializer()` function to have the following signature (using _static dispatch_):
```rust
fn serializer() -> impl MessageSerializer<Self> {
    DefaultSerializer::<Self>::new()
}
```
In this PR, I've decided to go with an approach that would allow dynamic dispatch through trait objects:

```rust
fn serializer() -> Box<dyn MessageSerializer<Self>> {
    Box::new(DefaultSerializer::<Self>::new())
}
```

The dynamic dispatch approach generally allows for more cleaner code (but I'm open to feedback on this during review).

- In order to make the code simpler (and solve for some compiler warnings), I chose to make `MessageSerializer` public and hide it using `#[doc(hidden)]`, instead of giving it the `pub(crate)` visibility.
- Clippy complains about the naming of `from_map()` function: `methods called `from_* ` usually take no `self``. I'm not sure if that should warrant us changing the name of the function.

This is my first take on this refactoring task. Sharing early to get feedback. I'll be waiting to make any recommended changes.

cc: @dbolduc @coryan

### DIFF
--- a/src/lro/src/details.rs
+++ b/src/lro/src/details.rs
@@ -24,8 +24,8 @@ pub(crate) fn handle_start<R, M>(
     result: Result<Operation<R, M>>,
 ) -> (Option<String>, PollingResult<R, M>)
 where
-    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
-    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
+    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned,
+    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned,
 {
     match result {
         Err(e) => (None, PollingResult::Completed(Err(e))),
@@ -41,8 +41,8 @@ pub(crate) fn handle_poll<R, M>(
     result: Result<Operation<R, M>>,
 ) -> (Option<String>, PollingResult<R, M>)
 where
-    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
-    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
+    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned,
+    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned,
 {
     match result {
         Err(e) => {
@@ -85,8 +85,8 @@ where
 
 fn handle_common<R, M>(op: Operation<R, M>) -> (Option<String>, PollingResult<R, M>)
 where
-    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
-    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
+    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned,
+    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned,
 {
     if op.done() {
         let result = as_result(op);
@@ -99,7 +99,7 @@ where
 
 fn as_result<R, M>(op: Operation<R, M>) -> Result<R>
 where
-    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
+    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned,
     M: wkt::message::Message + serde::de::DeserializeOwned,
 {
     if let Some(any) = op.response() {
@@ -114,7 +114,7 @@ where
 fn as_metadata<R, M>(op: Operation<R, M>) -> Option<M>
 where
     R: wkt::message::Message + serde::de::DeserializeOwned,
-    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
+    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned,
 {
     op.metadata().and_then(|a| a.try_into_message::<M>().ok())
 }
--- a/src/lro/src/details.rs
+++ b/src/lro/src/details.rs
@@ -24,8 +24,8 @@ pub(crate) fn handle_start<R, M>(
     result: Result<Operation<R, M>>,
 ) -> (Option<String>, PollingResult<R, M>)
 where
-    R: wkt::message::Message + serde::de::DeserializeOwned,
-    M: wkt::message::Message + serde::de::DeserializeOwned,
+    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
+    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
 {
     match result {
         Err(e) => (None, PollingResult::Completed(Err(e))),
@@ -41,8 +41,8 @@ pub(crate) fn handle_poll<R, M>(
     result: Result<Operation<R, M>>,
 ) -> (Option<String>, PollingResult<R, M>)
 where
-    R: wkt::message::Message + serde::de::DeserializeOwned,
-    M: wkt::message::Message + serde::de::DeserializeOwned,
+    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
+    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
 {
     match result {
         Err(e) => {
@@ -85,8 +85,8 @@ where
 
 fn handle_common<R, M>(op: Operation<R, M>) -> (Option<String>, PollingResult<R, M>)
 where
-    R: wkt::message::Message + serde::de::DeserializeOwned,
-    M: wkt::message::Message + serde::de::DeserializeOwned,
+    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
+    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
 {
     if op.done() {
         let result = as_result(op);
@@ -99,7 +99,7 @@ where
 
 fn as_result<R, M>(op: Operation<R, M>) -> Result<R>
 where
-    R: wkt::message::Message + serde::de::DeserializeOwned,
+    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
     M: wkt::message::Message + serde::de::DeserializeOwned,
 {
     if let Some(any) = op.response() {
@@ -114,7 +114,7 @@ where
 fn as_metadata<R, M>(op: Operation<R, M>) -> Option<M>
 where
     R: wkt::message::Message + serde::de::DeserializeOwned,
-    M: wkt::message::Message + serde::de::DeserializeOwned,
+    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
 {
     op.metadata().and_then(|a| a.try_into_message::<M>().ok())
 }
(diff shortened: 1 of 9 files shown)


src/lro/src/details.rs
```diff
@@ -24,8 +24,8 @@ pub(crate) fn handle_start<R, M>(
     result: Result<Operation<R, M>>,
 ) -> (Option<String>, PollingResult<R, M>)
 where
-    R: wkt::message::Message + serde::de::DeserializeOwned,
-    M: wkt::message::Message + serde::de::DeserializeOwned,
+    R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
+    M: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned + 'static,
```
dbolduc, line 28: optional: we could avoid these other changes if we separated the serialization and deserialization into two separate traits: `MessageSerializer` and `MessageDeserializer`.

don't feel obligated to do this.

src/wkt/src/message.rs
```diff
@@ -20,27 +20,52 @@ use crate::AnyError as Error;
...
 /// name and then validates extraction and insertion against this type.
 pub trait Message {
     /// The typename of this message.
     fn typename() -> &'static str;
 
+    /// Returns the serializer for this message type
     #[doc(hidden)]
-    /// Store the value into a JSON object.
-    fn to_map(&self) -> Result<Map, Error>
+    fn serializer() -> Box<dyn MessageSerializer<Self>>
```
dbolduc, line 31: We should prefer static dispatch. It is more performant. We only use dynamic dispatch when needed.

For any given `T`, there is only ever one `MessageSerializer` returned, so static dispatch will work just fine.

src/wkt/src/message.rs
```diff
@@ -20,27 +20,52 @@ use crate::AnyError as Error;
...
-    #[doc(hidden)]
-    /// Extract the value from a JSON object.
-    fn from_map(map: &Map) -> Result<Self, Error>
-    where
-        Self: serde::de::DeserializeOwned,
-    {
+/// Internal API for message serialization.
+/// This is not intended for direct use by consumers of this crate.
+#[doc(hidden)]
+pub trait MessageSerializer<T> {
```
dbolduc, line 42: We do not want downstream crates to implement this, so it should be: 

```rs
pub(crate) trait MessageSerializer<T> { ... }
```

and then we will have to deal with all of the warnings from running:
```
cargo clippy -p google-cloud-wkt -- --deny warnings
```

src/wkt/src/message.rs
```diff
@@ -20,27 +20,52 @@ use crate::AnyError as Error;
...
 /// Messages sent to and received from Google Cloud services may be wrapped in
-/// [Any][crate::any::Any]. `Any` uses a `@type` field to encoding the type
+/// [Any][crate::any::Any]. `Any` uses a `@type` field to encode the type
 /// name and then validates extraction and insertion against this type.
 pub trait Message {
     /// The typename of this message.
     fn typename() -> &'static str;
 
+    /// Returns the serializer for this message type
     #[doc(hidden)]
```
dbolduc, line 30: Heads up: we will also need to add a `#[allow(private_interfaces)]` annotation to this and every other `fn serializer()` after we make `MessageSerializer` private.

src/wkt/src/message.rs
```diff
@@ -20,27 +20,52 @@ use crate::AnyError as Error;
...
+
+impl<T> DefaultSerializer<T> {
+    pub fn new() -> Self {
+        Self {
+            _phantom: std::marker::PhantomData,
+        }
+    }
+}
+
+impl<T> MessageSerializer<T> for DefaultSerializer<T>
```
dbolduc, line 60: This is good.

I think we want to define another serializer in here:

```rs
// Serializes the type `T` into the `value` field.
pub(crate) struct ValueSerializer<T> {
    _phantom: std::marker::PhantomData<T>,
}

impl<T> ValueSerializer<T> {
    pub fn new() -> Self {
        Self {
            _phantom: std::marker::PhantomData,
        }
    }
}

impl<T> MessageSerializer<T> for ValueSerializer<T>
where
    T: serde::ser::Serialize + serde::de::DeserializeOwned + Sized + Message,
{
    fn to_map(&self, message: &T) -> Result<Map, Error> {
        to_json_other(message)
    }

    fn from_map(&self, map: &Map) -> Result<T, Error> {
        from_other(map)
    }
}
```

src/wkt/src/any.rs
```diff
@@ -136,10 +149,24 @@ impl crate::message::Message for Any {
...
+    fn serializer() -> Box<dyn crate::message::MessageSerializer<Self>>
+    where
+        Self: Sized,
+    {
+        Box::new(AnySerializer)
     }
-    fn from_map(map: &crate::message::Map) -> Result<Self, AnyError> {
+}
+
+struct AnySerializer;
```
dbolduc, line 162: ... and then we can use `crate::message::ValueSerializer<Any>` here, instead of repeating ourselves.

src/wkt/src/any.rs
```diff
@@ -136,10 +149,24 @@ impl crate::message::Message for Any {
...
-    fn to_map(&self) -> Result<crate::message::Map, AnyError> {
-        crate::message::to_json_other(self)
+
+    // Override the default serializer to use custom serialization
+    fn serializer() -> Box<dyn crate::message::MessageSerializer<Self>>
+    where
+        Self: Sized,
+    {
+        Box::new(AnySerializer)
     }
```
dbolduc, line 159: ... so this looks more like:

```rs
    fn serializer() -> impl crate::message::MessageSerializer<Self>
    {
        crate::message::ValueSerializer::<Self>::new()
    }
```

src/wkt/src/duration.rs
```diff
@@ -164,10 +164,24 @@ impl crate::message::Message for Duration {
...
+    // Override the default serializer to use custom serialization
+    fn serializer() -> Box<dyn crate::message::MessageSerializer<Self>>
+    where
+        Self: Sized,
+    {
+        Box::new(DurationSerializer)
+    }
+}
+
+struct DurationSerializer;
```
dbolduc, line 177: ... then same goes for `Duration`, `Timestamp`, all the types in `rstruct.rs`, etc.

src/wkt/src/wrappers.rs
```diff
@@ -113,17 +113,43 @@ pub type StringValue = String;
...
+
+impl<T> ProtobufTypeSerializer<T> {
+    pub fn new() -> Self {
+        Self {
+            _phantom: std::marker::PhantomData,
+        }
+    }
+}
+
+impl<T> crate::message::MessageSerializer<T> for ProtobufTypeSerializer<T>
```
dbolduc, line 128: This one can also be a `crate::message::ValueSerializer<T>`

src/wkt/src/wrappers.rs
```diff
@@ -157,10 +183,23 @@ impl crate::message::Message for UInt64Value {
...
+
+    fn serializer() -> Box<dyn crate::message::MessageSerializer<Self>>
+    where
+        Self: Sized,
+    {
+        Box::new(UInt64ValueSerializer)
+    }
+}
+
+struct UInt64ValueSerializer;
```
dbolduc, line 195: comment: This needs to stay custom, because it has custom `to_map`, `from_map` implementations.

Same goes for `Int64Value`, `BytesValue`. They also stay custom.

src/wkt/src/message.rs
```diff
@@ -20,27 +20,52 @@ use crate::AnyError as Error;
...
-    fn from_map(map: &Map) -> Result<Self, Error>
-    where
-        Self: serde::de::DeserializeOwned,
-    {
+/// Internal API for message serialization.
+/// This is not intended for direct use by consumers of this crate.
+#[doc(hidden)]
+pub trait MessageSerializer<T> {
+    fn to_map(&self, message: &T) -> Result<Map, Error>;
+    fn from_map(&self, map: &Map) -> Result<T, Error>;
```
dbolduc, line 44: `clippy` does not like `from_*` when something accepts a `&self`. That is valid, but we are using `&self` and concrete serializer objects to workaround the fact that we can't default types in a trait.

So maybe we change these function names to `serialize(...)`, `deserialize(...)` ?

dbolduc (changes requested):
Thanks! We are getting closer...

chrischiedo:
> Thanks! We are getting closer...

@dbolduc Thanks for the quick review! Much appreciated. Let me work on effecting the suggested changes.

src/wkt/src/message.rs
chrischiedo: > So maybe we change these function names to `serialize(...)`, `deserialize(...)` ?

Okay. I think that makes sense.

src/wkt/src/message.rs
chrischiedo: Noted.

src/wkt/src/message.rs
chrischiedo: Reverted back to static dispatch.

src/wkt/src/message.rs
chrischiedo: Done.

src/wkt/src/message.rs
chrischiedo: Done.

src/wkt/src/any.rs
chrischiedo: Done.

src/wkt/src/duration.rs
chrischiedo: Done.

src/wkt/src/wrappers.rs
chrischiedo: Indeed. Done.

src/wkt/src/wrappers.rs
chrischiedo: Yes. No changes have been made to these.

src/wkt/src/any.rs
chrischiedo: Done.

src/wkt/src/message.rs
chrischiedo: I have gone with `serialize_to_map` and `deserialize_from_map` respectively.

src/lro/src/details.rs
chrischiedo: For now: The only change I've made here is to remove the static lifetime bound, now that we've reverted back to static dispatch for our `fn serializer()` in the `Message` trait. We can always raise another PR if we think separating `MessageSerializer` would be _more_ beneficial.

chrischiedo:
Here's the update PR with the review changes.

cc: @dbolduc

coryan (changes requested):

dbolduc (approved):
This came together really nicely. I appreciate the collaboration on design + all the implementation work you did, @chrischiedo. Thanks a bunch.

Approved, in anticipation of fixing @coryan's documentation nits.

src/wkt/src/duration.rs
```diff
@@ -164,11 +164,11 @@ impl crate::message::Message for Duration {
     fn typename() -> &'static str {
         "type.googleapis.com/google.protobuf.Duration"
     }
-    fn to_map(&self) -> Result<crate::message::Map, crate::AnyError> {
-        crate::message::to_json_other(self)
-    }
-    fn from_map(map: &crate::message::Map) -> Result<Self, crate::AnyError> {
-        crate::message::from_other(map)
+
+    // Override the default serializer to use custom serialization
```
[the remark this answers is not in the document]
chrischiedo: Decided to remove the internal-only comments.

chrischiedo:
> This came together really nicely. I appreciate the collaboration on design + all the implementation work you did, @chrischiedo. Thanks a bunch.

Thank YOU! This was a GREAT learning opportunity for me. Thanks a lot for allowing me to work on it. Really appreciated, @dbolduc.

coryan (approved):

PR #1952, opened 2025-04-28
test(storage): add CRD tests

dbolduc:
Part of the work for #1813, will be useful for testing #1951 

Add bucket CRUD tests minus the U.

### DIFF
--- a/src/integration-tests/src/storage.rs
+++ b/src/integration-tests/src/storage.rs
@@ -15,8 +15,10 @@
 use crate::Error;
 use crate::Result;
 use gax::paginator::{ItemPaginator, Paginator};
+use rand::Rng;
+use storage::model::Bucket;
 
-pub const BUCKET_ID_LENGTH: usize = 32;
+pub const BUCKET_ID_LENGTH: usize = 63;
 
 pub async fn buckets(builder: storage::client::ClientBuilder) -> Result<()> {
     // Enable a basic subscriber. Useful to troubleshoot problems and visually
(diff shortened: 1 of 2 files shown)


src/integration-tests/src/storage.rs
```diff
@@ -38,6 +40,48 @@ pub async fn buckets(builder: storage::client::ClientBuilder) -> Result<()> {
 
     cleanup_stale_buckets(&client, &project_id).await?;
 
+    let bucket_id = random_bucket_id();
+    let bucket_name = format!("projects/_/buckets/{bucket_id}");
+
+    println!("\nTesting create_bucket()");
+    let create = client
+        .create_bucket("projects/_", bucket_id)
```
coryan, line 48: optional: As you may recall, the rate limit for this call is 0.5Hz.  We should to adjust the backoff to start a 2s, at least for this call.

src/integration-tests/src/storage.rs
```diff
@@ -38,6 +40,48 @@ pub async fn buckets(builder: storage::client::ClientBuilder) -> Result<()> {
...
+    println!("SUCCESS on create_bucket: {create:?}");
+    assert_eq!(create.name, bucket_name);
+
+    println!("\nTesting get_bucket()");
+    let get = client.get_bucket(&bucket_name).send().await?;
+    println!("SUCCESS on get_bucket: {get:?}");
+    assert_eq!(get.name, bucket_name);
+
+    println!("\nTesting list_buckets()");
+    let mut paginator = client
```
coryan, line 65: nit: I have taken to calling these "items"... they certainly are not a paginator after the `items()` call?

src/integration-tests/src/storage.rs
```diff
@@ -38,6 +40,48 @@ pub async fn buckets(builder: storage::client::ClientBuilder) -> Result<()> {
...
+        .items();
+    let mut bucket_names = Vec::new();
+    while let Some(bucket) = paginator.next().await {
+        bucket_names.push(bucket?.name);
+    }
+    println!("SUCCESS on list_buckets");
+    assert!(
+        bucket_names.iter().any(|name| name == &bucket_name),
+        "missing bucket name {} in {bucket_names:?}",
+        &bucket_name
```
coryan, line 78: nit:

```suggestion
        "missing bucket name {bucket_name} in {bucket_names:?}"
```

src/integration-tests/src/storage.rs
```diff
@@ -102,3 +146,14 @@ async fn cleanup_bucket(client: storage::client::Storage, name: String) -> Resul
     let _ = futures::future::join_all(pending).await;
     client.delete_bucket(&name).send().await
 }
+
+pub(crate) fn random_bucket_id() -> String {
+    const CHARSET: &[u8] = b"abcdefghijklmnopqrstuvwxyz0123456789";
+    const PREFIX: &str = "rust-sdk-testing-";
```
coryan, line 152: If we use the `integration-tests` label we do not need a long prefix like this.

src/integration-tests/src/storage.rs
```diff
@@ -102,3 +146,14 @@ async fn cleanup_bucket(client: storage::client::Storage, name: String) -> Resul
...
+    let mut bucket_id = String::new();
+    for _ in 0..(BUCKET_ID_LENGTH - PREFIX.len()) {
+        let idx = rand::rng().random_range(0..CHARSET.len());
+        bucket_id.push(CHARSET[idx] as char);
+    }
```
coryan, line 157: Consider:

```suggestion
    let id = std::iter::from_fn(|| move { rand::rng().random_range(0..CHARSET.len()) })
        .take(BUCKET_ID_LENGTH - PREFIX.len())
        .map(|idx| char::from(CHARSET[idx])
        .collect();
```

Or if you want to be nice, add an analog to `Alphanumeric` for lowercase characters to `lib.rs`.

coryan (approved):
Lots of nits, nothing blocking I think.

src/integration-tests/src/storage.rs
dbolduc: I like the prefix though. It makes it more obvious what these things are if they leak (especially in a testing project). I don't mind sacrificing some bits of randomness.

src/integration-tests/src/storage.rs
dbolduc: > the rate limit for this call is 0.5Hz.

I did read that. I saw we ran CRUD tests for buckets in Cloud C++ so figured it couldn't be that bad to run them here.

> adjust the backoff to start a 2s

Good idea. Done.

src/integration-tests/src/storage.rs
dbolduc: Ack, I am going to call them `bucket(s)`. Seems more specific.

src/integration-tests/src/storage.rs
dbolduc: Fixed.

src/integration-tests/src/storage.rs
dbolduc: You are now the proud owner of `RandomChars` which impls `rand::distr::Distribution<u8>` for any set of characters we may care about.

coryan (approved):

PR #1953, opened 2025-04-29
feat(generator): use Rust enumerations

### LINKED ISSUE #1904: Consider switching enums from a struct with associated consts to a non-exhaustive enum
**Is your feature request related to a problem? Please describe.**
Today, enums are generated as a `struct $FOO` with an associated `impl $FOO` block, where each variant is a `const` inside the `impl` block. Here's an example from [Spanner](https://github.com/googleapis/google-cloud-rust/blob/876163da976167bee6de1804531bef50fcfb3515/src/generated/spanner/admin/database/v1/src/model.rs#L397-L409):

```rust
 /// Indicates the current state of the backup.
 #[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
 pub struct State(i32);

 impl State {
     /// Not specified.
     pub const STATE_UNSPECIFIED: State = State::new(0);

     /// The pending backup is still being created. Operations on the
     /// backup may fail with `FAILED_PRECONDITION` in this state.
     pub const CREATING: State = State::new(1);

     /// The backup is complete and ready for use.
     pub const READY: State = State::new(2);
}
```

This pattern appears to have been introduced in [this PR](https://github.com/googleapis/google-cloud-rust/pull/20)[^1]. While I think that this pattern is useful from a forwards-compatibility standpoint—adding an enum variant is not a breaking change!—it can be somewhat frustrating to work with (for example: rust-analyzer's [`add_missing_match_arms`](https://rust-analyzer.github.io/book/assists.html#add_missing_match_arms) doesn't work with this, and while I *am* in a position to fix this in rust-analyzer, I had to squint at the generated enum to understand what was happening...) and could be better represented by other patterns.

**Describe the solution you'd like**
I'd prefer that this enum be represented using a Rust `enum`, a `#[non_exhaustive]` annotation[^2] and a `From<i32>` to handle integer-based enums https://github.com/googleapis/google-cloud-rust/pull/1448. Here's an example:

```rust
/// Indicates the current state of the backup.
#[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
#[non_exhaustive]
pub enum State {
    Unspecified,
    Creating,
    Ready,
    Unknown
}

impl From<i32> for State {
    fn from(value: i32) -> Self {
        match value {
            0 => State::Unspecified,
            1 => State::Creating,
            2 => State::Ready,
            _ => State::Unknown
        }    
    }
}
```

With this approach, adding a new enum variant isn't a semver-breaking change, as a caller matching against this enum will be forced to add a catch-all branch in their `match` expression.

**Describe alternatives you've considered**
The main one is "leave it as is". I *think* this pattern might be useful from in some limited FFI circumstances, but this SDK appears to largely eschew FFI.

**Additional context**
N/A.

[^1]: Cody: turns out that cbro is a mutual friend of ours! small world.
[^2]:  for what's worth, it's probably a good idea to slap `#[non_exhaustive]` on pretty much every struct and enum with public fields. that'll make adding a new enum variant and/or field a non-semver breaking change.

coryan:
With this change, the generator emits Rust enumerations for Protobuf
enumerations. Because Google Cloud uses open enumerations, the generated
enumerations can accept unknown values, as strings or integers and are
marked as `#[non_exhaustive]`.

Both unknown integers and unknown strings can roundtrip via JSON. Only
unknown integers can roundtrip the Protobuf wire format (currently
implemented with Prost-generated types).

The user guide gets a new section showing how these work and with some
recommendations to use them in `match` expressions.

Fixes #1904

### DIFF
--- a/generator/internal/rust/codec.go
+++ b/generator/internal/rust/codec.go
@@ -650,9 +650,9 @@ func addQueryParameter(f *api.Field) string {
 	switch f.Typez {
 	case api.ENUM_TYPE:
 		if f.Optional || f.Repeated {
-			return fmt.Sprintf(`let builder = req.%s.iter().fold(builder, |builder, p| builder.query(&[("%s", p.value())]));`, fieldName, f.JSONName)
+			return fmt.Sprintf(`let builder = req.%s.iter().fold(builder, |builder, p| builder.query(&[("%s", p)]));`, fieldName, f.JSONName)
 		}
-		return fmt.Sprintf(`let builder = builder.query(&[("%s", &req.%s.value())]);`, f.JSONName, fieldName)
+		return fmt.Sprintf(`let builder = builder.query(&[("%s", &req.%s)]);`, f.JSONName, fieldName)
 	case api.MESSAGE_TYPE:
 		if f.TypezID == ".google.protobuf.FieldMask" {
 			// `FieldMask` (and other well-known types) are special. Their JSON
@@ -685,7 +685,7 @@ func addQueryParameterOneOf(f *api.Field) string {
 	fieldName := toSnake(f.Name)
 	switch f.Typez {
 	case api.ENUM_TYPE:
-		return fmt.Sprintf(`let builder = req.%s().iter().fold(builder, |builder, p| builder.query(&[("%s", p.value())]));`, fieldName, f.JSONName)
+		return fmt.Sprintf(`let builder = req.%s().iter().fold(builder, |builder, p| builder.query(&[("%s", p)]));`, fieldName, f.JSONName)
 	case api.MESSAGE_TYPE:
 		// Query parameters in nested messages are first converted to a
 		// `serde_json::Value`` and then recursively merged into the request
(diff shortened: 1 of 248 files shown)


coryan:
PTAL

generator/internal/rust/codec.go
```diff
@@ -766,8 +766,67 @@ func enumValueName(e *api.EnumValue) string {
...
+//	    BlackAndBlue,
+//	    _123,
+//	    UnknownVariant(/* implementation detail */),
+//	}
+//
+// ```
+// sometimes it is not followed.
+func enumValueVariantName(e *api.EnumValue) string {
+	// The most common case is trying to strip the prefix for `FOO_BAR_UNSPECIFIED`.
+	// The naming conventions being what they are, we need to test with with a
```
dbolduc, line 805: s/with with/with/

src/generated/cloud/essentialcontacts/v1/src/model.rs
```diff
@@ -638,160 +638,289 @@ impl wkt::message::Message for SendTestMessageRequest {
...
+    }
+}
+
+impl<'de> serde::de::Deserialize<'de> for ValidationState {
+    fn deserialize<D>(deserializer: D) -> std::result::Result<Self, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        deserializer.deserialize_any(wkt::internal::EnumVisitor::<ValidationState>::new(
+            ".google.cloud.essentialcontacts.v1.ValidationState",
```
dbolduc, line 923: hmm so we only use this name when printing error messages... arguably not worth it, but fine.

src/generated/cloud/essentialcontacts/v1/src/model.rs
```diff
@@ -638,160 +638,289 @@ impl wkt::message::Message for SendTestMessageRequest {
...
+}
 
-    /// Creates a new NotificationCategory instance.
-    pub(crate) const fn new(value: i32) -> Self {
-        Self(value)
-    }
+#[doc(hidden)]
+pub mod notification_category {
+    #[allow(unused_imports)]
+    use super::*;
```
dbolduc, line 679: when are these ever used?

generator/internal/rust/templates/common/enum.mustache
```diff
@@ -18,60 +18,128 @@ limitations under the License.
...
+    The `UnknownValue` variant must be public, but we do not want people to
+    build that variant with a known value. We use a struct, with a `pub(crate)`
+    constructor, and a `pub(crate)` enum.
+}}
+#[doc(hidden)]
+{{> /templates/common/feature_gate}}
+pub mod {{Codec.ModuleName}} {
+    {{! Very rarely, this is unused. It is easier to always disable the warning. }}
+    #[allow(unused_imports)]
+    use super::*;
```
dbolduc, line 47: I don't see us using any types from `super::`

dbolduc (approved):

generator/internal/rust/templates/common/enum.mustache
coryan: Depending on how the code is generated then `wkt::` may be a top-level external crate, may be a module in super, or may be an external crate imported via `extern crate` and carried to each module via these `use super::*`.

So, using `wkt::` below works most times (but not when it is a module), `::wkt` works other times, and `super::wkt` works in yet other times.  The only approach that works always in what we have here.

We may want to clean this up, but I think the biggest obstable are the gRPC clients.

src/generated/cloud/essentialcontacts/v1/src/model.rs
coryan: These are needed at least in `src/wkt/src/generated/mod.rs`:

```
git commit -m"Address review comments" generator/
[feat-generator-rust-enums-take-3 a9e00b470] Address review comments
 1 file changed, 1 insertion(+), 1 deletion(-)
(.venv) coryan-macbookpro:google-cloud-rust coryan$ git diff
diff --git a/src/wkt/src/generated/mod.rs b/src/wkt/src/generated/mod.rs
index 0233a9961..8840b6657 100755
--- a/src/wkt/src/generated/mod.rs
+++ b/src/wkt/src/generated/mod.rs
@@ -1063,8 +1063,6 @@ pub mod extension_range_options {
 
     #[doc(hidden)]
     pub mod verification_state {
-        #[allow(unused_imports)]
-        use super::*;
         #[derive(Clone, Debug, PartialEq)]
         pub struct UnknownValue(pub(crate) wkt::internal::UnknownEnumValue);
     }
(.venv) coryan-macbookpro:google-cloud-rust coryan$ cargo build -p google-cloud-wkt
   Compiling google-cloud-wkt v0.4.0 (/Users/coryan/google-cloud-rust/src/wkt)
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `wkt`
    --> src/wkt/src/generated/mod.rs:1067:44
     |
1067 |         pub struct UnknownValue(pub(crate) wkt::internal::UnknownEnumValue);
     |                                            ^^^ use of unresolved module or unlinked crate `wkt`
     |
     = help: if you wanted to use a crate named `wkt`, use `cargo add wkt` to add it to your `Cargo.toml`
help: consider importing this module
     |
1066 +         use crate::internal;
     |
help: if you import `internal`, refer to it directly
     |
1067 -         pub struct UnknownValue(pub(crate) wkt::internal::UnknownEnumValue);
1067 +         pub struct UnknownValue(pub(crate) internal::UnknownEnumValue);
     |

For more information about this error, try `rustc --explain E0433`.
error: could not compile `google-cloud-wkt` (lib) due to 1 previous error
```

coryan: I created #1981 to fix this.

generator/internal/rust/codec.go
coryan: Done.

generator/internal/rust/templates/common/enum.mustache
dbolduc: Got it, thanks. I forgot about the generated types in `wkt` (despite looking at them last night)

PR #1954, opened 2025-04-29
cleanup: use clippy in more targets

coryan:


.github/workflows/sdk.yaml
```diff
@@ -94,7 +94,7 @@ jobs:
             cargo "${sub}" --package google-cloud-wkt --no-default-features --features time
             cargo "${sub}" --package google-cloud-wkt --all-features
           done
-          cargo clippy --no-deps --package google-cloud-wkt --all-features --profile=test -- --deny warnings
+          cargo clippy --no-deps --package google-cloud-wkt --all-features --all-targets --profile=test -- --deny warnings
 
           cargo clean
           echo "==== google-cloud-gax ===="
```
dbolduc, line 100: Unrelated aside: should we split the different packages into different steps of the build?

dbolduc (approved):
Interesting. I do not grok why `--profile=test` missed some of these integration tests

.github/workflows/sdk.yaml
coryan: We could, but then we need to add requirements to GitHub pull requests when packages that have features change.

coryan:
> Interesting. I do not grok why `--profile=test` missed some of these integration tests

I *think* that there are two controls: the targets built by `cargo check` (which is what clippy uses under the hood) and the features / profile enabled while building those targets.  Just `--profile=test` (without `--all-targets`) compiles the code protected by `#[cfg(test)]` but does not compile the integration tests.  `cargo check --help` may clarify things better than I do.

PR #1955, opened 2025-04-29
chore(auth): Add test cases to cover uncaught mutants

harkamaljot:
This PR improves mutation test coverage for the authentication components by adding missing test cases and refactoring code for better testability.

**Key Changes:**

* **`service_account.rs`:**
    * Replaced `std::time::Instant` with `tokio::time::Instant` to enhance testability.
    * Extracted current time (`now`) and expiry time calculations into dedicated functions to enable isolated testing.
* **`token_cache.rs`:**
    * Updated test cases to specifically verify the handling of long expiry timeouts.
* **`user.rs`:**
    * Added a test case to confirm that the default `universe_domain` is correctly set to "gdu".
* **`src/credentials.rs`:**
    * Introduced `skip` tags for the testing module.
* **`tests/credentials.rs`:**
    * Added test cases to validate that non-default headers and non-default `universe_domain` values are accurately retrieved and utilized.

Other mutants
The cargo-mutants tool currently does not support conditional skipping of functions using cfg_attr in a way that would allow code to be mutation-tested on one OS (e.g., Windows) but skipped on others (e.g., Linux) directly via source code attributes (see [mutants.rs attrs documentation](https://mutants.rs/attrs.html)).

To ensure that we do not hide any mutants, we have chosen not to add unconditional #[mutants::skip] attributes to this code.


src/auth/src/credentials.rs
```diff
@@ -548,6 +548,7 @@ fn adc_well_known_path() -> Option<String> {
 /// This module is mainly relevant to other `google-cloud-*` crates, but some
 /// external developers (i.e. consumers, not developers of `google-cloud-rust`)
 /// may find it useful.
+#[cfg_attr(test, mutants::skip)]
```
coryan, line 551: Please include a comment explaining why we need to skip this for mutant testing.  Remember to use just `// ` so it does not go into the customer-facing documentation.

src/auth/tests/credentials.rs
```diff
@@ -191,6 +191,7 @@ mod test {
             .unwrap();
         let fmt = format!("{:?}", creds);
         assert!(fmt.contains("ApiKeyCredentials"), "{fmt:?}");
+        assert!(fmt.contains("api_key: \"[censored]\""), "{fmt:?}");
```
coryan, line 194: I think a property you want to test is  `!fmt.contains("test-api-key")`

src/auth/Cargo.toml
```diff
@@ -54,3 +54,4 @@ test-case   = "3"
 tokio       = { version = "1", features = ["macros", "rt-multi-thread", "test-util"] }
 tokio-test  = "0.4"
 url         = "2"
+mutants     = "0.0.3"
```
coryan, line 57: I think we should start moving all such dependencies to the top-level `Cargo.toml` file and just use `mutants.workspace = true` here.  I am not asking you change the existing deps, just the new one.

coryan (approved):
A few minor improvements, please fix and then merge.

src/auth/tests/credentials.rs
harkamaljot: In this test basically I wanted to verify that api key credentials is not printing the key. i.e. the debug implementation is appropriately censoring the key (https://github.com/googleapis/google-cloud-rust/blob/main/src/auth/src/credentials/api_key_credentials.rs#L83)

coryan: Yes, but you are testing something slightly different.  Maybe the API key is printed in another field, or maybe somebody changes the code to print `api_key: [censored], uncensored: $full_api_key`...

harkamaljot: ohh I see, thanks for the explanation.

sai-sunder-s (approved):

PR #1956, opened 2025-04-29
impl(generator/rust): response protos from other packages

dbolduc:
Part of the work for #1951 

Use the response message's package for gRPC. Do not assume it is the same package as the service. We want: `google::iam::v1::Policy`, not `google::storage::v2::Policy`.

I considered using `SourceFQN` instead of introducing a new field. But the `toSnake(...)` conversion for the module name gave me pause. In practice, all of the package names are already snake case though. Whatever.

### DIFF
--- a/generator/internal/rust/annotate.go
+++ b/generator/internal/rust/annotate.go
@@ -120,6 +120,9 @@ type messageAnnotation struct {
 	// The fully qualified name, relative to `codec.modulePath`. Typically this
 	// is the `QualifiedName` with the `crate::model::` prefix removed.
 	RelativeName string
+	// The package name mapped to Rust modules. That is, `google.service.v1`
+	// becomes `google::service::v1`.
+	PackageModuleName string
 	// The FQN is the source specification
 	SourceFQN         string
 	MessageAttributes []string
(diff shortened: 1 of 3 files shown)


generator/internal/rust/annotate.go
```diff
@@ -428,6 +431,15 @@ func (c *codec) addFeatureAnnotations(model *api.API, ann *modelAnnotations) {
...
 
+// Maps "google.foo.v1" to "google::foo::v1"
+func packageToModuleName(p string) string {
+	components := strings.Split(p, ".")
+	for i, c := range components {
+		components[i] = toSnake(c)
+	}
+	return strings.Join(components, "::")
+}
+
```
coryan, line 442: I am almost certain there is a function doing something similar in `codec.go`. Consider moving this function there.

generator/internal/rust/templates/grpc-client/transport.rs.mustache
```diff
@@ -149,7 +149,7 @@ impl super::stub::{{Codec.Name}} for {{Codec.Name}} {
           (),
           {{/ReturnsEmpty}}
           {{^ReturnsEmpty}}
-          crate::{{Service.Codec.PackageModuleName}}::{{OutputType.Codec.RelativeName}},
+          crate::{{OutputType.Codec.PackageModuleName}}::{{OutputType.Codec.RelativeName}},
```
coryan, line 152: I was expecting a change in some generated file, but there is none because the cases where there would be are suppressed. And this shows nothing else is affected. Neat.

PR #1960, opened 2025-04-29
feat(storage): IAM operations

### LINKED ISSUE #1951: Add IAM methods to `storage-control`

dbolduc:
Fixes #1951 

Add the client surface for IAM operations from `google/storage/v2`, with examples (that build, but don't run), and integration tests for the new methods.

The method signatures (which are handwritten in this veneer) are consistent with the Get/Set/Test IAM methods from other crates.

### DIFF
--- a/src/storage-control/src/client.rs
+++ b/src/storage-control/src/client.rs
@@ -301,30 +301,15 @@ impl Storage {
     ///
     /// ```
     /// # use google_cloud_storage_control::client::Storage;
-    /// async fn example(client: &Storage) -> gax::Result<()> {
-    ///     // Get the current IAM policy for this bucket
+    /// # use iam_v1::model::Policy;
+    /// async fn example(client: &Storage, updated_policy: Policy) -> gax::Result<()> {
     ///     let policy = client
-    ///         .get_iam_policy("projects/_/buckets/my-bucket")
-    ///         .send()
-    ///         .await?;
-    ///
-    ///     // Add a binding to the policy
-    ///     let mut new_policy = policy.clone();
-    ///     new_policy.bindings.push(
-    ///         iam_v1::model::Binding::new()
-    ///             .set_role("roles/storage.legacyBucketReader")
-    ///             .set_members(["user:me@example.com"]),
-    ///     );
-    ///
-    ///     // Update the IAM policy for this bucket
-    ///     let response = client
     ///         .set_iam_policy("projects/_/buckets/my-bucket")
     ///         .set_update_mask(wkt::FieldMask::default().set_paths(["bindings"]))
-    ///         .set_policy(new_policy)
+    ///         .set_policy(updated_policy)
     ///         .send()
     ///         .await?;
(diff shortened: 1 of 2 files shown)


src/storage-control/src/client.rs
```diff
@@ -261,6 +261,112 @@ impl Storage {
...
+    ///     let response = client
+    ///         .set_iam_policy("projects/_/buckets/my-bucket")
+    ///         .set_update_mask(wkt::FieldMask::default().set_paths(["bindings"]))
+    ///         .set_policy(new_policy)
+    ///         .send()
+    ///         .await?;
+    ///     println!("response details={response:?}");
+    ///
+    ///     Ok(())
+    /// }
```
coryan, line 329: This example is too big for the IDE tooltip.  Consider something like:

```rust
    /// async fn example(client: &Storage) -> gax::Result<()> {
    ///     let policy = compute_updated_policy(client, "projects/_/buckets/my-bucket").await?;
    ///     let response = client
    ///         .set_iam_policy("projects/_/buckets/my-bucket")
    ///         .set_policy(policy)
    ///         .set_update_mask(wkt::FieldMask::default().set_paths(["bindings"]))
    ///         .send().await?;
    ///     Ok(())
    /// }
    ///
    /// async fn compute_updated_policy(client: &Storage, bucket: &str) -> gax::Result<Policy> {
    /// # panic!();
    /// }
    ///
```

src/storage-control/src/client.rs
```diff
@@ -261,6 +261,112 @@ impl Storage {
...
+
+    /// Tests a set of permissions on the given bucket, object, or managed folder
+    /// to see which, if any, are held by the caller.
+    ///
+    /// # Parameters
+    /// * `resource` should be
+    ///   * `projects/_/buckets/{bucket}` for a bucket,
+    ///   * `projects/_/buckets/{bucket}/objects/{object}` for an object, or
+    ///   * `projects/_/buckets/{bucket}/managedFolders/{managedFolder}` for a
+    ///     managed folder.
```
coryan, line 346: It is odd that the available resource names do not match between these functions.  @benwhitehead any thoughts?

coryan (approved):

src/storage-control/src/client.rs
dbolduc: +1

BenWhitehead: Logged an internal bug (b/414625980) for determination of whether these can also be used for {Get,Set}IamPolicy.

src/storage-control/src/client.rs
dbolduc: Done.

src/storage-control/src/client.rs
coryan: Looks like these resource names apply to all the IAM functions . We should add them to the hand-crafted comments.

dbolduc: Oops, missed this comment before merging. I will send a follow up with the changes.

PR #1963, opened 2025-04-29
impl(wkt): add tests for serializing/deserializing messages with f32

suzmue:
Add a new type to use in the serde_as attribute for supporting ProtoJSON serialization and deserialization of floats. This just adds the tests for using the new type, this will then need to be added to the generated libraries to make sure that infinite and nan values are handled correctly.

For #1767

### DIFF
--- a/src/wkt/src/internal.rs
+++ b/src/wkt/src/internal.rs
@@ -12,6 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+pub struct F32;
+
+impl serde_with::SerializeAs<f32> for F32 {
+    fn serialize_as<S>(value: &f32, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::ser::Serializer,
+    {
+        float_serialize(value, serializer)
+    }
+}
+
+impl<'de> serde_with::DeserializeAs<'de, f32> for F32 {
+    fn deserialize_as<D>(deserializer: D) -> Result<f32, D::Error>
+    where
+        D: serde::de::Deserializer<'de>,
+    {
+        float_deserialize(deserializer)
+    }
+}
+
 /// A helper to serialize `f32` to ProtoJSON format.
 pub fn float_serialize<S>(x: &f32, s: S) -> std::result::Result<S::Ok, S::Error>
 where
--- a/src/wkt/tests/message_with_f32.rs
+++ b/src/wkt/tests/message_with_f32.rs
@@ -1,0 +1,108 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#[cfg(test)]
+mod test {
+    use serde_json::json;
+    type Result = std::result::Result<(), Box<dyn std::error::Error>>;
+    use test_case::test_case;
+
+    #[serde_with::serde_as]
+    #[serde_with::skip_serializing_none]
+    #[derive(Clone, Debug, Default, PartialEq, serde::Deserialize, serde::Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct MessageWithF32 {
+        #[serde_as(as = "google_cloud_wkt::internal::F32")]
+        pub singular: f32,
+        #[serde_as(as = "Option<google_cloud_wkt::internal::F32>")]
+        pub optional: Option<f32>,
+        #[serde_as(as = "Vec<google_cloud_wkt::internal::F32>")]
+        pub repeated: Vec<f32>,
+    }
+
+    #[test_case(9876.5, 9876.5)]
+    #[test_case(f32::INFINITY, "Infinity")]
+    #[test_case(f32::NEG_INFINITY, "-Infinity")]
+    #[test_case(f32::NAN, "NaN")]
+    fn test_singular<T>(input: f32, want: T) -> Result
+    where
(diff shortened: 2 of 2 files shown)


src/wkt/src/internal.rs
```diff
@@ -12,6 +12,26 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+pub struct F32;
```
coryan, line 15: Oh, this is clever.... 

Optional: you may be able to remove the `float_serialize()` and `float_deserialize()` functions.

src/wkt/tests/message_with_f32.rs
```diff
@@ -0,0 +1,108 @@
...
+    #[serde_with::skip_serializing_none]
+    #[derive(Clone, Debug, Default, PartialEq, serde::Deserialize, serde::Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub struct MessageWithF32 {
+        #[serde_as(as = "google_cloud_wkt::internal::F32")]
+        pub singular: f32,
+        #[serde_as(as = "Option<google_cloud_wkt::internal::F32>")]
+        pub optional: Option<f32>,
+        #[serde_as(as = "Vec<google_cloud_wkt::internal::F32>")]
+        pub repeated: Vec<f32>,
```
coryan, line 31: Please include a `HashMap<String, f32>`, I think those appear in the generated code too.

coryan (approved):

src/wkt/tests/message_with_f32.rs
```diff
@@ -0,0 +1,106 @@
...
+    fn test_singular<T>(input: f32, want: T) -> Result
+    where
+        T: serde::ser::Serialize,
+    {
+        let msg = MessageWithF32 {
+            singular: input,
+            ..Default::default()
+        };
+        let got = serde_json::to_value(&msg)?;
+        let want = json!({"singular": want, "repeated": []});
```
dbolduc, line 47: Ideally, we skip serializing if `repeated` is empty

suzmue: I tried skipping for both repeated and hashmap, and while I was able to serialize fine, I then ran into issues trying to deserialize back:

```
Error: Error("missing field `repeated`", line: 0, column: 0)
```

from this line:
```rust
        let roundtrip = serde_json::from_value::<MessageWithF32>(got)?;
```

It looks like we have these skip annotations in our code as well.

Based on https://protobuf.dev/programming-guides/json/ it looks like null should be allowed to be accepted for repeated, but it seems like maybe not for maps.

src/wkt/src/internal.rs
suzmue: Inlined!

src/wkt/tests/message_with_f32.rs
suzmue: Done.

src/wkt/tests/message_with_f32.rs
coryan: FWIW, the generated code does this:

https://github.com/googleapis/google-cloud-rust/blob/3a75999806b98cf6bc72adbf156a8bba1ae19c35/src/generated/api/servicecontrol/v1/src/model.rs#L325-L331

I do not think skipping this field is important for these tests Though it would make some test cases cleaner, I am happy with the tests as they are.  @suzmue your call, we can always chat more and change the tests if @dbolduc is thinking of something that I missed.

dbolduc: I just think the tests should look like the code we are going to generate.

I think all we need to do is add the `serde(default)` annotation to the message:

https://github.com/googleapis/google-cloud-rust/blob/3a75999806b98cf6bc72adbf156a8bba1ae19c35/src/generated/api/servicecontrol/v1/src/model.rs#L42

And add the `skip_serialize_if = ...` to the repeated fields:

https://github.com/googleapis/google-cloud-rust/blob/3a75999806b98cf6bc72adbf156a8bba1ae19c35/src/generated/api/servicecontrol/v1/src/model.rs#L325-L326

src/wkt/tests/message_with_f32.rs
```diff
@@ -0,0 +1,133 @@
...
+        hashmap.insert("-inf".to_string(), f32::NEG_INFINITY);
+        hashmap.insert("nan".to_string(), f32::NAN);
+
+        let msg = MessageWithF32 {
+            hashmap,
+            ..Default::default()
+        };
+
+        let got = serde_json::to_value(&msg)?;
+        let want = json!({"singular": 0.0, "repeated": [],"hashmap": {"number": 9876.5, "inf": "Infinity", "-inf": "-Infinity", "nan": "NaN"}});
```
dbolduc, line 109: nit: consider addine newlines for readability:

```rs
        let want = json!({
            "singular": 0.0,
            "hashmap": {
                "number": 9876.5,
                "inf": "Infinity",
                "-inf": "-Infinity",
                "nan": "NaN"
            }
        });
```

(I am assuming we will remove the `repeated` field)

src/wkt/tests/message_with_f32.rs
```diff
@@ -0,0 +1,133 @@
...
+        Ok(())
+    }
+
+    fn assert_float_eq(left: f32, right: f32) {
+        // Consider all NaN as equal.
+        if left.is_nan() && right.is_nan() {
+            return;
+        }
+        // Consider all infinites floats of the same sign as equal.
+        if (left.is_infinite() && right.is_infinite())
```
dbolduc, line 126: nit: extra parentheses

dbolduc (approved):
code all looks good, just some nits in the test.

src/wkt/tests/message_with_f32.rs
suzmue: I agree that it should look like in the code (message_with_i64 and message_with_bytes need to be updated too which these tests were based on, will do in a follow up). 

TY for the default annotation tip :)

src/wkt/tests/message_with_f32.rs
suzmue: Done.

src/wkt/tests/message_with_f32.rs
suzmue: Filed #1973 for updating the other tests.

dbolduc (approved):

PR #1964, opened 2025-04-30
impl: prepare for errors converting to protos

coryan:
To support deserialization of enums from unknown strings values we need
to support errors during conversion to proto (currently Prost) types.

This gave me an opportunity to reorganize the generated code that
converts to and from protos. I also covered more cases for enums, even
though they do not appear in the current code.

Part of the work for #1904


src/gax-internal/src/prost.rs
```diff
@@ -15,14 +15,46 @@
...
+    #[error("enum {0} does not contain an integer value")]
+    EnumNoIntegerValue(&'static str),
+}
+
+type Result<T> = std::result::Result<T, ConvertError>;
+
+/// Converts from `Self` into `T`, where `T` is expected to be a Protobuf-generated type.
+pub trait ToProto<T>: Sized {
+    type Output;
+    fn to_proto(self) -> Result<Self::Output>;
```
dbolduc, line 31: Good call. It is easier to keep track of what is a proto vs. what is our type with the new names.

dbolduc (approved):

Issue #1815, opened 2025-04-16
Consider making the required fields mandatory for request or document why this does not work


coryan:
At the moment creating a request only requires the fields necessary to create the request path.

However, other fields are required for some requests.  Maybe the field behavior annotations (see #27) can be used to decide which fields are mandatory in requests.  We tried this in #462, but did not work. It would be useful to have a clear document describing why this is not helpful.

Note that there will always be cases where the field is optional for some requests, and required for other requests, maybe depending on the presence or even the values of other fields.  I think it is fine if we cannot handle all cases, just document the rationale.

dbolduc:
We should check if the API folks consider it a breaking change to drop the `REQUIRED` annotation on a field.

As far as the proto is concerned, it is not a breaking change. The same messages that were written before continue to work.

But if we have required fields as parameters in our client APIs, their removal is definitely a breaking change.

coryan:
> We should check if the API folks consider it a breaking change to drop the `REQUIRED` annotation on a field.

They do not: https://google.aip.dev/203#backwards-compatibility

> As far as the proto is concerned, it is not a breaking change. The same messages that were written before continue to work.
> 
> But if we have required fields as parameters in our client APIs, their removal is definitely a breaking change.

Yes, and we can increase the major version for that client.  The alternative is that every request that requires some fields fails the first 3 times you make it. And that is if you are lucky and the error message tells you what field is missing.

dbolduc:
We can fail the message on the client side with an informative error saying why we did it

coryan:
I just realized that all of this is a violation of https://google.aip.dev/client-libraries/4223, very unfortunate, I really liked the resulting code and functions.

dbolduc:
FWIW, I don't think AIP-4223 makes an argument against client-side validation. Their argument is more like: "robust request validation must be done on the server side".

The only argument I see against client-side validation it is that the annotation/documentation could be incorrect. That's not much of an argument.

amanda-tarafa:
PHP and/or Ruby used to have required fields as parameters and they "recently" removed that because dropping the `required` annotation was always a breaking change client side. In those languages at least, I believe that was found to be more disruptive than the request failing for some users that did not include required fields.

Also, going through the "required fields are parameters route" does not take into account nested required fields in messages referecend through the request (you'd end up with many parameters potentially), so it's only solving the user problem partially.

coryan:
[...]

PR #1965, opened 2025-04-30
feat(generator/rust): document required setters

coryan:
Motivated by #1815

### DIFF
--- a/generator/internal/rust/templates/crate/src/builder.rs.mustache
+++ b/generator/internal/rust/templates/crate/src/builder.rs.mustache
@@ -176,6 +176,10 @@ pub mod {{Codec.ModuleName}} {
         {{#InputType.Codec.SingularFields}}
 
         /// Sets the value of [{{Codec.FieldName}}][{{Codec.FQMessageName}}::{{Codec.SetterName}}].
+        {{#DocumentAsRequired}}
+        ///
+        /// This is documented as a **required** field for requests.
+        {{/DocumentAsRequired}}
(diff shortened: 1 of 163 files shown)


generator/internal/rust/templates/crate/src/builder.rs.mustache
```diff
@@ -176,6 +176,10 @@ pub mod {{Codec.ModuleName}} {
         {{#InputType.Codec.SingularFields}}
 
         /// Sets the value of [{{Codec.FieldName}}][{{Codec.FQMessageName}}::{{Codec.SetterName}}].
+        {{#DocumentAsRequired}}
+        ///
+        /// This is documented as a **required** field for requests.
```
dbolduc, line 181: s/documented as //

too meta for me.

dbolduc: here and below.

dbolduc (approved):

generator/internal/rust/templates/crate/src/builder.rs.mustache
coryan: Technically the field may not be required anymore.  The service may have dropped this requirement. You are right though, it is too "Ackchyually".

Fixed.

dbolduc: > Technically the field may not be required anymore. The service may have dropped this requirement.

Doesn't matter. The documentation is wrong whether we say "This is ..." vs. "This is documented as ...".

My point was more that these comments are the documentation. I think we were distinguishing between service documentation vs. client library documentation, but customers will not recognize that distinction.

PR #1970, opened 2025-05-01
fix(generator): initialize ID field for enums

coryan:
I am not sure why, but this field was not initialized. And I also do not
understand why this changed the generated code (for one enum).


dbolduc (approved):
:shrug:

PR #1972, opened 2025-05-01
cleanup(lro)!: seal Poller

dbolduc:
Part of the work for #1609, analog of #1971 

Seal `lro::Poller`. This lets us add functions to the trait without breaking customers (because customers are not able to implement this trait).

### DIFF
--- a/src/lro/src/lib.rs
+++ b/src/lro/src/lib.rs
@@ -100,7 +100,6 @@ impl<R, M> Operation<R, M> {
     }
 }
 
-// This fools clippy into thinking `sealed::Poller` is public.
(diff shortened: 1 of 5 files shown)


src/lro/src/lib.rs
```diff
@@ -100,14 +100,19 @@ impl<R, M> Operation<R, M> {
     }
 }
 
+// This fools clippy into thinking `sealed::Poller` is public.
```
coryan, line 103: nit: does not fool it.  It is a public type, the name is not public.

coryan (approved):

PR #1975, opened 2025-05-01
impl(wkt): support types for "enums take 3"

coryan:
Part of the work for #1904

### DIFF
--- a/src/wkt/src/internal/enums.rs
+++ b/src/wkt/src/internal/enums.rs
@@ -111,8 +111,7 @@ pub fn display_enum(
     value: Option<i32>,
 ) -> Result<(), std::fmt::Error> {
     match (name, value) {
-        (Some(n), Some(_)) => f.write_str(n),
-        (Some(n), None) => f.write_str(n),
+        (Some(n), _) => f.write_str(n),
         (None, Some(v)) => write!(f, "{}", v),
         (None, None) => unreachable!("enums must have a numeric or string value"),
     }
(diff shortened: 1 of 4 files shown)


src/wkt/src/internal/enums.rs
```diff
@@ -0,0 +1,232 @@
...
+}
+
+pub fn display_enum(
+    f: &mut std::fmt::Formatter<'_>,
+    name: Option<&str>,
+    value: Option<i32>,
+) -> Result<(), std::fmt::Error> {
+    match (name, value) {
+        (Some(n), Some(_)) => f.write_str(n),
+        (Some(n), None) => f.write_str(n),
```
dbolduc, line 115: optional nit: can this be: `(Some(n), _) =>`

src/wkt/src/internal/enums.rs
```diff
@@ -0,0 +1,232 @@
...
+    use serde_json::json;
+    use test_case::test_case;
+
+    #[test]
+    fn unknown_enum_value_accessors() {
+        let u = UnknownEnumValue::Integer(123);
+        assert_eq!(u.value(), Some(123));
+        assert_eq!(u.name(), None);
+
+        let u = UnknownEnumValue::String("123".into());
```
dbolduc, line 133: nit: can we call this "foo" or something? Just so the reader doesn't think it is related to `123`, the value.

Here and in the next test.

dbolduc (approved):

src/wkt/src/internal/enums.rs
coryan: Done.

src/wkt/src/internal/enums.rs
coryan: Done.

Issue #1958, opened 2025-04-29
Use `#[deprecated]` in the Rust mapping


coryan:
Protobuf elements such as RPCs, messages, fields, and even enum values may be marked as deprecated.

We should mark the corresponding entities as deprecated too, so customers get the right warnings.  We need to figure out how this works with warnings while compiling the crate.

PR #1977, opened 2025-05-01
feat(generator): capture deprecated markers

coryan:
Both Protobuf and OpenAPIv3 supporting marking some fields, messages,
requests and (in the case of Protobuf) even complete services as
deprecated. In this change we capture the information, but no generated
code changes.

Part of the work for #1958

### DIFF
--- a/generator/internal/parser/openapi.go
+++ b/generator/internal/parser/openapi.go
@@ -217,6 +217,7 @@ func makeMethods(a *api.API, model *libopenapi.DocumentModel[v3.Document], packa
 			m := &api.Method{
 				Name:          op.Operation.OperationId,
 				ID:            mID,
+				Deprecated:    op.Operation.Deprecated != nil && *op.Operation.Deprecated,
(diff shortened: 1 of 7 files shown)


generator/internal/parser/openapi.go
```diff
@@ -217,6 +217,7 @@ func makeMethods(a *api.API, model *libopenapi.DocumentModel[v3.Document], packa
 			m := &api.Method{
 				Name:          op.Operation.OperationId,
 				ID:            mID,
+				Deprecated:    op.Operation.Deprecated != nil && *op.Operation.Deprecated,
```
dbolduc, line 220: unit tests for OpenAPI?

dbolduc (approved):

generator/internal/parser/openapi.go
coryan: Good idea. Done.

dbolduc (approved):

PR #1986, opened 2025-05-02
feat(generator): better classify fields

coryan:
Add attributes to directly determine if fields are maps, repeated, or
singular.

I will use this to refactor some of the templates.

### DIFF
--- a/generator/internal/api/model.go
+++ b/generator/internal/api/model.go
@@ -482,8 +482,16 @@ type Field struct {
 	JSONName string
 	// Optional indicates that the field is marked as optional in proto3.
 	Optional bool
+
+	// For a given field, at most one of `Repeated` or `Map` is true.
+	//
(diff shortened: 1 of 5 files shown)


generator/internal/api/model.go
```diff
@@ -524,6 +532,10 @@ func (field *Field) DocumentAsRequired() bool {
 	return slices.Contains(field.Behavior, FIELD_BEHAVIOR_REQUIRED)
 }
 
+func (f *Field) Singular() bool {
```
dbolduc, line 535: Call this in a test? Or save for a follow up?

dbolduc (approved):

generator/internal/api/model.go
coryan: Saving for a followup, with that we can write `{{#Singular}}` in the mustache templates.

PR #1991, opened 2025-05-02
impl: add F64 type for ProtoJSON serde

suzmue:
Rewrite the internal float visitor to also support F64. Made the code generic instead of duplicating to make it more likely that if we need to update the code, we will remember to update both.

The main changes are in internal.rs

For #1767


src/wkt/src/internal.rs
```diff
@@ -17,73 +17,114 @@
 //! These types are intended for developers of the Google Cloud client libraries
 //! for Rust. They are undocumented and may change at any time.
 
-pub struct F32;
+pub type F32 = Float<f32>;
+pub type F64 = Float<f64>;
```
coryan, line 21: I would like to avoid exporting a public type like `Float<T>`.  I wonder if we can keep F32 and F64 as unit types, something like:

```suggestion
pub type F32;
pub type F64;

trait FloatExt: num_traits::Float + std::fmt::Debug {
    type Type;
    fn serialize<S: serde::ser::Serializer>(self, serializer: S) -> Result<S::Ok, S::Error>;
}

impl FloatExt for f32 {
    type Type = f32;
    fn serialize<S: serde::ser::Serializer>(self, serializer: S) -> Result<S::Ok, S::Error> {
        serializer.serialize_f32(self)
    }
}

impl<T> serde_with::SerializeAs<T> for T where T: FloatExt + 'de {
  // Can reference T::Type here to get back the f32
```

As a last resort: using macros to generate the code below?

suzmue: Great suggestion, I'll give that a try.

I had left it as exported since it had to be in order to export the type aliases, but would much prefer to not.

suzmue: I fear I may have veered overly far in the macro  direction, but I think its better to avoid adding the num_traits dependency if necessary.

Might try another refactor to share the code without the macros.

suzmue: Seems like I'd need to bring back in generics, which I'm not yet convinced is better, so I think I'll take a review as is

src/wkt/src/internal.rs
```diff
@@ -18,19 +18,78 @@
...
+                // Handle special strings, see https://protobuf.dev/programming-guides/json/.
+                match value {
+                    "NaN" => Ok(<$t>::NAN),
+                    "Infinity" => Ok(<$t>::INFINITY),
+                    "-Infinity" => Ok(<$t>::NEG_INFINITY),
+                    _ => Err(serde::de::Error::invalid_value(
+                        serde::de::Unexpected::Other(value),
+                        &format!(
+                            "a valid ProtoJSON string for {} (NaN, Infinity, -Infinity)",
+                            std::any::type_name::<$t>()
```
coryan, line 61: This is the only thing that gives me pause.  I wonder if this would work without the caveats that `std::any::type_name` gives:

https://doc.rust-lang.org/std/macro.stringify.html

src/wkt/src/internal.rs
```diff
@@ -18,19 +18,78 @@
...
+                    )),
+                }
+            }
+
+            // Floats and doubles in serde_json are f64.
+            fn visit_f64<E>(self, value: f64) -> std::result::Result<Self::Value, E>
+            where
+                E: serde::de::Error,
+            {
+                // Casting f64 to f32 to produce the closest possible float value.
```
coryan, line 73: nit: when `$t == f64` the comment is not quite right, maybe:

```suggestion
                // This is trivial for `f64`. For `f32`, casting f64 to f32 is guaranteed to produce the closest possible float value:
```

coryan (approved):
No worries about the "too many macros" vs. "too many generics" thing.  This works, and (AFAICT) whether we use macros or generics is an implementation detail we can change later.

A couple of minor nits, and then  :shipit:

src/wkt/src/internal.rs
suzmue: Done.

src/wkt/src/internal.rs
suzmue: Done.

PR #2003, opened 2025-05-05
cleanup: update local format commands

dbolduc:
This is what we run in our CI.

I am considering adding scripts for each of these commands that we would reference here and call from our CI. Then this stuff will not fall out of sync.

### DIFF
--- a/doc/contributor/howto-guide-set-up-development-environment.md
+++ b/doc/contributor/howto-guide-set-up-development-environment.md
@@ -272,7 +272,7 @@ use with:
 ```bash
 git ls-files -z -- \
     '*.md' ':!:**/testdata/**' ':!:**/generated/**' | \
-    xargs -0 -r -P "$(nproc)" -n 50 mdformat
+    xargs -0 mdformat
 ```
 
 ### Format YAML files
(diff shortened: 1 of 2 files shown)


doc/contributor/howto-guide-set-up-development-environment.md
```diff
@@ -268,7 +270,9 @@ pip install -r ci/requirements.txt
 use with:
 
 ```bash
-git ls-files -z -- '*.md' ':!:**/testdata/**' | xargs -0 -r -P "$(nproc)" -n 50 mdformat
+git ls-files -z -- \
+    '*.md' ':!:**/testdata/**' ':!:**/generated/**' | \
+    xargs -0 -r -P "$(nproc)" -n 50 mdformat
```
coryan, line 275: nit: I do not think we have enough `.md` files to justify the complex `xargs` arguments, maybe just do the same as with the other commands:

```suggestion
    xargs -0 mdformat
```

coryan (approved):

doc/contributor/howto-guide-set-up-development-environment.md
dbolduc: Updated here and in `lint.yaml`

PR #2005, opened 2025-05-05
feat(auth): Introduce GCE_METADATA_HOST env var to set the metadata server host

sai-sunder-s:
By default MDS uses `metadata.google.internal`  host for the metadata server. If a user wants to use a different host name, currently the only was is to use builder and set the `endpoint`.

This change introduces another way: set the `GCE_METADATA_HOST` environment variable.

Note: If user sets the env var and also provides an endpoint override through builder, we will use the env var.

### DIFF
--- a/src/auth/src/credentials/mds.rs
+++ b/src/auth/src/credentials/mds.rs
@@ -19,7 +19,9 @@
 //! This is a local service to the VM (or pod) which (as the name implies) provides
 //! metadata information about the VM. The service also provides access
 //! tokens associated with the [default service account] for the corresponding
-//! VM. The default host name of the metadata service is `metadata.google.internal`.
+//! VM.
+//!
+//! The default host name of the metadata service is `metadata.google.internal`.
 //! If you would like to use a different hostname, you can set it using the
 //! `GCE_METADATA_HOST` environment variable.
 //!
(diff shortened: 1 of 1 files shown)


coryan:
Please mark it as "Ready for review" once the tests pass.

src/auth/src/credentials/mds.rs
```diff
@@ -455,6 +463,44 @@ mod test {
...
+                    scopes: Some(scopes.join(",")),
+                    recursive: None,
+                },
+                Arc::new(Mutex::new(0)),
+            ),
+        )]))
+        .await;
+
+        // Trim out 'http://' from the endpoint provided by the fake server
+        let _e = ScopedEnv::set(super::GCE_METADATA_HOST_ENV_VAR, &endpoint[7..]);
```
coryan, line 491: Hmmm.... I think I would prefer:

```suggestion
        let _e = ScopedEnv::set(super::GCE_METADATA_HOST_ENV_VAR, endpoint.strip_prefix("http://").unwrap_or(endpoint));
```

src/auth/src/credentials/mds.rs
```diff
@@ -19,7 +19,9 @@
 //! This is a local service to the VM (or pod) which (as the name implies) provides
 //! metadata information about the VM. The service also provides access
 //! tokens associated with the [default service account] for the corresponding
-//! VM.
+//! VM. The default host name of the metadata service is `metadata.google.internal`.
+//! If you would like to use a different hostname, you can set it using the
+//! `GCE_METADATA_HOST` environment variable.
```
coryan, line 24: I think this deserves a new paragraph: it introduces a new concept or idea:

```suggestion
//! VM.
//!
//! The default host name of the metadata service is `metadata.google.internal`.
//! If you would like to use a different hostname, you can set it using the
//! `GCE_METADATA_HOST` environment variable.
```

coryan (approved):

PR #2011, opened 2025-05-06
test: add integration tests for custom serialization

suzmue:
Was not able to find examples to use in the existing dependencies, so pulled in the aiplatform package as well.

For #1767

### DIFF
--- a/src/integration-tests/tests/serialization.rs
+++ b/src/integration-tests/tests/serialization.rs
@@ -42,6 +42,54 @@ mod serialization {
     }
 
     #[test]
+    fn serde_with_float() -> Result<()> {
+        let generation_config =
+            aiplatform::model::GenerationConfig::default().set_temperature(9876.5);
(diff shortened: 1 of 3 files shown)


src/integration-tests/tests/serialization.rs
```diff
@@ -41,6 +41,54 @@ mod serialization {
         Ok(())
     }
 
+    #[test]
+    fn serde_with_float() -> Result<()> {
```
dbolduc, line 45: optional: these are only `ser`, consider validating the `de`:

```rs
let roundtrip = serde_json::from_value::<T>(got)?;
assert_eq!(input, roundtrip); // or whatever we call the original message
```

dbolduc (approved):

src/integration-tests/tests/serialization.rs
suzmue: done!

PR #2014, opened 2025-05-06
test: add integration test for sql admin service

### LINKED ISSUE #1773: Tests: Additional tests for SQL Admin and BigQuery v2 clients

PhongChuong:
fixes #1773


src/integration-tests/src/sql.rs
```diff
@@ -0,0 +1,136 @@
...
+
+async fn cleanup_stale_sql_instances(
+    client: &sql::client::SqlInstancesService,
+    project_id: &str,
+) -> Result<()> {
+    use std::time::{Duration, SystemTime, UNIX_EPOCH};
+    let stale_deadline = SystemTime::now()
+        .duration_since(UNIX_EPOCH)
+        .map_err(Error::other)?;
+    // let stale_deadline = stale_deadline - Duration::from_secs(48 * 60 * 60);
```
coryan, line 101: That seems like something you want to restore...

src/integration-tests/tests/driver.rs
```diff
@@ -88,6 +90,18 @@ mod driver {
             .map_err(report)
     }
 
+    #[test_case(sql::client::SqlInstancesService::builder(); "default")]
+    #[test_case(sql::client::SqlInstancesService::builder().with_tracing(); "with tracing enabled")]
+    #[test_case(sql::client::SqlInstancesService::builder().with_retry_policy(retry_policy()); "with retry enabled")]
```
coryan, line 95: I think you probably want a combination of `with_tracing().with_retry_policy(...)` the motivation for having multiple tests is not very strong for this library.

src/integration-tests/src/sql.rs
```diff
@@ -0,0 +1,136 @@
...
+    };
+
+    let project_id = crate::project_id()?;
+    let name = random_sql_instance_name(&project_id);
+    let client: sql::client::SqlInstancesService = builder.build().await?;
+
+    cleanup_stale_sql_instances(&client, &project_id).await?;
+
+    println!("\nTesting insert sql instance");
+    let insert = client
```
coryan, line 41: It would be nice if the tests included any attribute that uses the `json_name` attribute. For example:

https://github.com/googleapis/googleapis/blob/351a2dc65443e3fd50280a372830ba963699cfa6/google/cloud/sql/v1/cloud_sql_tiers.proto#L68

Maybe in the `get` request you can check the value of `RAM`?

coryan (changes requested):
Did you have to enable the service for the test project? I would have expected some changes in the Terraform scripts to deal with that?

src/integration-tests/src/sql.rs
```diff
@@ -134,3 +139,37 @@ async fn cleanup_stale_sql_instances(
...
+
+    let list = client.list(&project_id).send().await?;
+
+    assert_eq!(
+        list.items
+            .into_iter()
+            .find(|v| v.tier.eq("db-f1-micro"))
+            .ok_or_else(|| Error::other("tiers list should contain db-f1-micro".to_string()))?
+            .ram,
+        644245094
```
coryan, line 171: Maybe just verify that it is not zero, the precise number is of less interest to us?

PhongChuong: Done.

PhongChuong (commented):
Added the terraform services for sqladmin and bigquery.

dbolduc (approved):

Issue #638, opened 2025-01-07
Serialization of `oneofs` for `bytes` and maps


coryan:
I think our serialization code for `oneof` where one of the branches is a `bytes` or a `map` may be incorrect.  I do not see any code to add the right attributes into generated code.

We do not need such fields for the moment, so I am just adding this note and punting.

PR #2030, opened 2025-05-07
fix: add attributes for oneof variants in generated code

### LINKED ISSUE #2010: Update code generator to allow for annotations to oneof enum variants
We use enums to represent oneof fields. Example:

http://cloud/python/docs/reference/aiplatform/latest/google.cloud.aiplatform_v1.types.FeatureValue
Becomes:
```rust
    #[derive(Clone, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
    #[serde(rename_all = "camelCase")]
    #[non_exhaustive]
    pub enum Value {
        /// Bool type feature value.
        BoolValue(bool),
        /// Double type feature value.
        DoubleValue(f64),
        /// Int64 feature value.
        Int64Value(i64),
        /// String feature value.
        StringValue(std::string::String),
        /// A list of bool type feature value.
        BoolArrayValue(std::boxed::Box<crate::model::BoolArray>),
        /// A list of double type feature value.
        DoubleArrayValue(std::boxed::Box<crate::model::DoubleArray>),
        /// A list of int64 type feature value.
        Int64ArrayValue(std::boxed::Box<crate::model::Int64Array>),
        /// A list of string type feature value.
        StringArrayValue(std::boxed::Box<crate::model::StringArray>),
        /// Bytes feature value.
        BytesValue(::bytes::Bytes),
        /// A struct type feature value.
        StructValue(std::boxed::Box<crate::model::StructValue>),
    }
```

AFAICT we do not yet support adding annotations to these types, which are needed for doing custom serde.

The serde_as annotations need to be added to the value, not the variant:
```rust
        DoubleValue([#serde_as(as="wkt::internal::F64")]f64),
```

For context, discovered when adding additional integration testing serialization for #1973.

suzmue:
Add the necessary serde attributes to the generated enum values for oneof types. These need to be added to the values for the variant.

Update tests to reflect that oneofs have explicit presence, and default values should be serialized. See https://protobuf.dev/programming-guides/field_presence/. Also add serialization integration tests for a oneof type.

Fixes #2010


generator/internal/rust/codec.go
```diff
@@ -408,6 +408,17 @@ func fieldFormatter(typez api.Typez) string {
 }
 
 func fieldSkipAttributes(f *api.Field) []string {
+	// oneofs have explicit presence, and default values should be serialized:
+	// https://protobuf.dev/programming-guides/field_presence/.
```
coryan, line 412: Aside: the comment is fine for now.  Eventually we would want this to work for OpenAPI and discovery docs too.  I am not sure how those would represent `oneof`, something for all of us to keep in mind.

src/generated/cloud/redis/cluster/v1/src/model.rs
```diff
@@ -3711,6 +3715,7 @@ pub mod connection_detail {
     /// Connectivity Automation (auto-registered connection) during the cluster
     /// creation, or it could be created by customer themselves (user-registered
     /// connection).
+    #[serde_with::serde_as]
```
coryan, line 3718: It is a bit sad that we have all these annotations that mostly go unused. They are also harmless, so LGTM

coryan (approved):
Thanks for fixing i64 and bytes along the way. Do you think this fixes #638 ? If not, please update that bug with what you think remains.

generator/internal/rust/codec.go
suzmue: Ack/

suzmue:
> Thanks for fixing i64 and bytes along the way. Do you think this fixes #638 ? If not, please update that bug with what you think remains.

I think it should but none of the tests cover any attributes being added for maps, so having that might be worth it.

Issue #1934, opened 2025-04-26
Convert LROs returning `wkt::Empty` to return `()`


coryan:
We do this for normal RPCs, it seems odd to return something different for LROs.

PR #2031, opened 2025-05-07
feat(lro): convert `wkt::Empty` to unit type

coryan:
Part of the work for #1934

### DIFF
--- a/src/lro/src/internal.rs
+++ b/src/lro/src/internal.rs
@@ -69,7 +69,7 @@ where
     Q: Fn(String) -> QF + Send + Sync + Clone,
     QF: std::future::Future<Output = Result<Operation<wkt::Empty, MetadataType>>> + Send + 'static,
 {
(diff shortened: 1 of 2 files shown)


src/lro/src/internal.rs
```diff
@@ -51,6 +51,144 @@ where
     PollerImpl::new(polling_error_policy, polling_backoff_policy, start, query)
 }
 
+/// Creates a new `impl Poller<(), M>` from the closures created by the generator.
+///
+/// This is intended as an implementation detail of the generated clients.
+/// Applications should have no need to create or use this struct.
```
dbolduc, line 57: nit: s/struct/function/

dbolduc (approved):

src/lro/src/internal.rs
coryan: Fixed.

PR #2032, opened 2025-05-08
impl(storage): bootstrap media client

coryan:


src/integration-tests/src/storage.rs
```diff
@@ -16,10 +16,92 @@ use crate::{Error, Result};
...
+
+        tracing::subscriber::set_default(subscriber)
+    };
+
+    // Create a temporary bucket for the test.
+    let (control, bucket) = create_test_bucket().await?;
+
+    let client = builder.build().await?;
+
+    tracing::info!("testing insert_object()");
```
dbolduc, line 43: Q: should we prefer this to `println!()` ?

src/integration-tests/src/storage.rs
```diff
@@ -16,10 +16,92 @@ use crate::{Error, Result};
...
+        .insert_object(
+            &bucket.name,
+            "quick.text",
+            "the quick brown fox jumps over the lazy dog",
+        )
+        .await?;
+    tracing::info!("success with insert={insert:?}");
+
+    tracing::info!("testing read_object()");
+    let contents = client.read_object(&bucket.name, &insert.name).await?;
```
dbolduc, line 54: verify `contents` matches the input?

src/storage/src/client.rs
```diff
@@ -0,0 +1,325 @@
...
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_storage::client::Storage;
+    /// async fn example(client: &Storage) -> gax::Result<()> {
+    ///     let response = client
+    ///         .insert_object(
+    ///             "projects/_/buckets/my-bucket",
+    ///             "my-object",
+    ///             bytes::Bytes::from("the quick brown fox jumped over the lazy dog"),
```
dbolduc, line 114: nit: `bytes::Bytes::from()` is unnecessary, but maybe you are just showing that it is not necessarily a string

src/storage/src/client.rs
```diff
@@ -0,0 +1,325 @@
...
+    ///         )
+    ///         .await?;
+    ///     println!("response details={response:?}");
+    ///     Ok(())
+    /// }
+    /// ```
+    pub async fn insert_object<B, O, P>(
```
dbolduc, line 121: should these return a builder?

should they have `RequestOptions` ?

src/storage/src/client.rs
```diff
@@ -0,0 +1,325 @@
...
+///   Note that object names can contain `/` characters, which are treated as
+///   any other character (no special directory semantics).
+///
+/// [with_endpoint()]: ClientBuilder::with_endpoint
+/// [with_credentials()]: ClientBuilder::with_credentials
+/// [Private Google Access with VPC Service Controls]: https://cloud.google.com/vpc-service-controls/docs/private-connectivity
+/// [Application Default Credentials]: https://cloud.google.com/docs/authentication#adc
+#[derive(Clone, Debug)]
+pub struct Storage {
+    inner: reqwest::Client,
```
dbolduc, line 86: why do we want to use a `reqwest::Client` directly, instead of our fancy schmancy wrapper?

dbolduc: I imagine `read_object` is retry-able.

src/storage/src/client.rs
```diff
@@ -0,0 +1,325 @@
...
+/// # Configuration
+///
+/// To configure `Storage` use the `with_*` methods in the type returned
+/// by [builder()][Storage::builder]. The default configuration should
+/// work for most applications. Common configuration changes include
+///
+/// * [with_endpoint()]: by default this client uses the global default endpoint
+///   (`https://storage.googleapis.com`). Applications using regional
+///   endpoints or running in restricted networks (e.g. a network configured
+//    with [Private Google Access with VPC Service Controls]) may want to
```
BenWhitehead, line 39:
```suggestion
///   with [Private Google Access with VPC Service Controls]) may want to
```

src/storage/src/client.rs
```diff
@@ -0,0 +1,325 @@
...
+///   override this default.
+/// * [with_credentials()]: by default this client uses
+///   [Application Default Credentials]. Applications using custom
+///   authentication may need to override this default.
+///
+/// # Pooling and Cloning
+///
+/// `Storage` holds a connection pool internally, it is advised to
+/// create one and the reuse it.  You do not need to wrap `Storage` in
```
BenWhitehead, line 48:
```suggestion
/// create one and then reuse it.  You do not need to wrap `Storage` in
```

src/storage/src/client.rs
```diff
@@ -0,0 +1,325 @@
...
+///
+/// # Service Description
+///
+/// The Cloud Storage API allows applications to read and write data through
+/// the abstractions of buckets and objects. For a description of these
+/// abstractions please see <https://cloud.google.com/storage/docs>.
+///
+/// This client is used to perform metadata operations, such as creating
+/// buckets, deleting objects, listing objects, etc. It does not expose any
+/// functions to write or read data in objects.
```
BenWhitehead, line 60: copy-pasted from storage control?

src/storage/src/client.rs
coryan: > why do we want to use a reqwest::Client directly, instead of our fancy schmancy wrapper?

Because we will need to do all kinds of ad-hoc things:

- We will need to pool multiple connections.
- We will need to remove connections from the pool when a request fails with 429 or 503 (or anything that hints at an overloaded server or front-end).
- We will need to stream data, which our client cannot do and probably should not do either.

> I imagine read_object is retry-able.

Resumable.  You would need to start from the last byte received.  So I am not sure we would even use the same retry policies.

src/storage/src/client.rs
coryan: Fixed.

src/storage/src/client.rs
coryan: Yes and yes, but this is the "bootstrap" part.  We will need also stubs (for tracing and mocks).

src/integration-tests/src/storage.rs
coryan: I think so, they show up in order with other tracing logs.

src/integration-tests/src/storage.rs
coryan: Done.

src/storage/src/client.rs
coryan: Thanks

coryan:
PTAL

src/storage/src/client.rs
dbolduc: Ack, thanks.

src/storage/src/client.rs
dbolduc: Hmm, okay, I will defer to you on bootstrapping / trailblazing.

coryan: Can I get an approval then?

dbolduc (approved):

src/storage/src/client.rs
dbolduc: Sorry, I was about to approve then got distracted.

PR #2034, opened 2025-05-08
feat(auth)!: update token and header interface

harkamaljot:
As part of https://github.com/googleapis/google-cloud-rust/issues/2036, this PR introduces the following changes:

1.  **Interface Update for `token` and `headers`:**
    * Methods now accept an optional `http::Extension` argument, allowing to pass in any additional options.
2.  **Return Type Change for `headers`:**
    * The `headers` method now returns `Result<HeaderMap>` instead of `Result<Vec<(HeaderName, HeaderValue)>>`.
3.  **Test Code Refactor:**
    * Removed duplicated testing code in `service_account.rs` for better maintainability.

In the upcoming PR, I am planning to introduce the entity tag implementation and update the return type of `headers` method.


src/auth/src/credentials/api_key_credentials.rs
```diff
@@ -157,16 +157,12 @@ mod test {
...
 
-        assert_eq!(
-            headers,
-            vec![HV {
-                header: API_KEY_HEADER_KEY.to_string(),
-                value: "test-api-key".to_string(),
-                is_sensitive: true,
-            }]
-        );
+        assert_eq!(headers.len(), 1);
```
coryan, line 163: Here and below. When this fails you will want to know more:

```suggestion
        assert_eq!(headers.len(), 1, "{headers:?}");
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=85830927b4b6cb1289ef810e8c32bcca

src/auth/src/credentials/api_key_credentials.rs
```diff
@@ -157,16 +157,12 @@ mod test {
...
+        assert_eq!(value, HeaderValue::from_str("test-api-key").unwrap());
```
coryan, line 164: Why not:

```suggestion
        assert_eq!(value, HeaderValue::from_static("test-api-key"));
```

src/storage/src/client.rs
```diff
@@ -148,9 +148,13 @@ impl Storage {
...
             );
-        let auth_headers = self.cred.headers().await.map_err(Error::authentication)?;
+        let auth_headers = self
+            .cred
+            .headers(None)
+            .await
+            .map_err(Error::authentication)?;
         let builder = auth_headers
-            .into_iter()
+            .iter()
```
coryan, line 157: Why this change? It forces a copy that we could avoid?

src/gax-internal/src/http.rs
```diff
@@ -127,8 +127,12 @@ impl ReqwestClient {
...
             .into_iter()
             .fold(builder, |b, t| b.timeout(t));
-        let auth_headers = self.cred.headers().await.map_err(Error::authentication)?;
-        for (key, value) in auth_headers.into_iter() {
+        let auth_headers = self
+            .cred
+            .headers(None)
+            .await
+            .map_err(Error::authentication)?;
+        for (key, value) in auth_headers.iter() {
```
coryan, line 135: Why `iter()` instead of `into_iter()`?

src/storage/src/client.rs
harkamaljot: The behavior of HeaderMap::iter(), where it returns None for the HeaderName in cases of duplicate headers, felt like it would make the code unnecessarily complex(https://docs.rs/http/latest/http/header/struct.HeaderMap.html#method.iter). I found HeaderMap::iter() to be a more straightforward alternative because it guarantees the HeaderName is always present with its corresponding HeaderValue. But looks like we prefer to avoid the copy instead, I can update the code to reflect that.

coryan: Ah, I completely forgot about that. Let's use `.iter()` or (where appropriate) `.extend()`.

src/auth/src/credentials.rs
```diff
@@ -88,12 +88,12 @@ where
 }
 
 impl Credentials {
-    pub async fn token(&self) -> Result<crate::token::Token> {
-        self.inner.token().await
+    pub async fn token(&self, extensions: Option<Extensions>) -> Result<crate::token::Token> {
```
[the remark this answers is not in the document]
harkamaljot: You're right that the observable behavior if we just used an empty Extension will be similar. However, I opted for Option<Extension> as it makes explicit at the type level that an extension is not required, especially during the first call to the token method, when user will not have an entity tag. Please let me know if you think passing empty Extension is more clear, I can make that change.

src/auth/src/credentials/service_account.rs
```diff
@@ -317,7 +317,7 @@ fn token_expiry_time(current_time: OffsetDateTime) -> OffsetDateTime {
 
 #[async_trait]
 impl TokenProvider for ServiceAccountTokenProvider {
-    async fn token(&self) -> Result<Token> {
+    async fn token(&self, _extensions: Option<Extensions>) -> Result<Token> {
```
sai-sunder-s, line 320: Is the `token` method in TokenProvider of any of the cred types going to use the `extensions`?

I believe only TokenCache needs `extensions` to provide appropriate CachedResult and control duration we wait before returning back to the caller?

If you were forced to add the parameter everywhere because the trait is updated to fit for token cache, we can consider a different trait for cache.

harkamaljot: I'm not sure how that would work, as the client only has access to the trait through which they make calls, and they are unaware of the caching logic. If you have any specific idea, please can you share?

src/auth/src/credentials.rs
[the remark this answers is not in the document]
harkamaljot: Reading more about it, looks it it is preferred in general to have an empty map in API map, let me change it.

src/auth/src/credentials/service_account.rs
sai-sunder-s: https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1da20f0508af152e6df851d0f25eed66

This is what I have in mind

coryan (approved):

src/auth/src/credentials.rs
```diff
@@ -667,7 +654,7 @@ mod test {
             .to_string()
     }
 
-    fn b64_decode_to_json(s: String) -> serde_json::Value {
+    pub fn b64_decode_to_json(s: String) -> serde_json::Value {
```
sai-sunder-s, line 657: can this be `pub(crate)`? Same for `generate_pkcs8_private_key`

PR #2059, opened 2025-05-08
impl(storage-control): hide unused builders

dbolduc:
Part of the work for #1813 

We are going to add another client (for StorageControl), which also has a `ClientBuilder`. We do not use either, and the types will clash, so stop generating them for veneers.

We also skip most of the client documentation (which references the `builder()`), if we will not be exposing the client.

### DIFF
--- a/generator/internal/rust/templates/crate/src/builder.rs.mustache
+++ b/generator/internal/rust/templates/crate/src/builder.rs.mustache
@@ -26,7 +26,7 @@ limitations under the License.
 pub mod {{Codec.ModuleName}} {
     use crate::Result;
 
-    {{^Codec.SkipBuilderDocs}}
+    {{^Codec.HasVeneer}}
     /// A builder for [{{Codec.Name}}][super::super::client::{{Codec.Name}}].
     ///
     /// ```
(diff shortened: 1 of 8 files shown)


generator/internal/rust/templates/crate/src/builder.rs.mustache
```diff
@@ -40,9 +40,9 @@ pub mod {{Codec.ModuleName}} {
     ///     .build().await?;
     /// # gax::Result::<()>::Ok(()) });
     /// ```
-    {{/Codec.SkipBuilderDocs}}
     pub type ClientBuilder =
         gax::client_builder::ClientBuilder<client::Factory, gaxi::options::Credentials>;
+    {{/Codec.HasVeneer}}
 
     pub(crate) mod client {
```
coryan, line 47: Do we still use the code inside `pub(crate) mod client`?

coryan (approved):

generator/internal/rust/templates/crate/src/builder.rs.mustache
dbolduc: No, we don't. We use `Client::new(...)` directly. Removed.

PR #2063, opened 2025-05-09
impl(storage-control): add control internals

dbolduc:
Part of the work for #1813 

Generate (internal) library from `storage/control/v2`. We defer the one LRO for later.

I am calling this an `impl`, but we are adding the builders and messages to the public API. They are kind of useless on their own though. :shrug: 

There are duplicated protos, but we can clean that up later.


src/storage-control/src/generated/gapic_control/transport.rs
```diff
@@ -0,0 +1,512 @@
...
+
+mod info {
+    const NAME: &str = env!("CARGO_PKG_NAME");
+    const VERSION: &str = env!("CARGO_PKG_VERSION");
+    lazy_static::lazy_static! {
+        pub(crate) static ref X_GOOG_API_CLIENT_HEADER: String = {
+            let ac = gaxi::api_header::XGoogApiClient{
+                name:          NAME,
+                version:       VERSION,
+                library_type:  gaxi::api_header::GAPIC,
```
coryan, line 31: I just realized: if we are saying `has-veneer = true` this should be `gaxi::api_header:GCCL`. Please fix in a separate PR.

coryan (approved):

src/storage-control/src/generated/gapic_control/transport.rs
dbolduc: hmm... ok. It is kind of a GAPIC disguised as a veneer.

dbolduc: But yeah, we would want to do that for any other veneer.

PR #2065, opened 2025-05-09
feat(storage-control): add some of the Folders API

dbolduc:
Part of the work for #1813 

Add `{create, get, delete, list}_folder` APIs to the storage control client.

I picked out the client signature, and handwrote the examples in the method documentation. We will lock down the surface later. For now, we just want to get a decent first iteration out.

These are validated in an integration test.


coryan (approved):

dbolduc:
cc: @benwhitehead

PR #2083, opened 2025-05-10
fix(auth): Hide testing mod from public documentation

### LINKED ISSUE #2077: Hide or Remove Testing Module from Documentation
Public documentation currently contains testing module.

Either:
1. Make the testing module private. https://github.com/googleapis/google-cloud-rust/blob/882e734279608ebb097bdd594c0cdd18d948eb1a/src/auth/src/credentials.rs#L554
2. Mark the documentation as `#[doc(hidden)]`

sai-sunder-s:
Fixed #2077 

The testing module is used in gax tests, so hiding it instead of making it `pub(crate)`

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -551,6 +551,7 @@ fn adc_well_known_path() -> Option<String> {
 // Skipping mutation testing for this module. As it exclusively provides
 // hardcoded credential stubs for testing purposes.
 #[cfg_attr(test, mutants::skip)]
+#[doc(hidden)]
 pub mod testing {
     use crate::Result;
     use crate::credentials::Credentials;


src/auth/src/credentials.rs
```diff
@@ -551,6 +551,7 @@ fn adc_well_known_path() -> Option<String> {
 // Skipping mutation testing for this module. As it exclusively provides
 // hardcoded credential stubs for testing purposes.
 #[cfg_attr(test, mutants::skip)]
+#[doc(hidden)]
```
dbolduc, line 554: Auth probably wants to offer some form of insecure credentials. They are useful for customers when testing a real client against an emulator, for example.

Thoughts?

dbolduc (approved):
I am okay with this, but I suspect we will want something public that serves the same purpose eventually

src/auth/src/credentials.rs
sai-sunder-s: A cred that just returns back an empty headermap?

dbolduc: > A cred that just returns back an empty headermap?

Yeah. This is not a necessity, though.

Applications that need it could trivially implement `CredentialsProvider`. We would just be saving them some time by writing that code for them.

PR #2089, opened 2025-05-12
fix(auth)!: Single method in Service Account Builder for scopes and audience

### LINKED ISSUE #2078: Improve audience and scope setting in Service Account Builder
The service account credentials builder currently has methods to set audience and scopes. Only one of them can be set at a time and this is mentioned in the description and enforced in code. Investigate if we can simplify the interface by combining the set methods to clearly communicate to the user that only one can be set.

sai-sunder-s:
Fixes #2078

### DIFF
--- a/src/auth/src/credentials/service_account.rs
+++ b/src/auth/src/credentials/service_account.rs
@@ -92,98 +92,28 @@ use tokio::time::Instant;
 
 const DEFAULT_SCOPE: &str = "https://www.googleapis.com/auth/cloud-platform";
 
-#[derive(Debug)]
-enum ServiceAccountRestrictions {
-    Audience(String),
-    Scopes(Vec<String>),
-}
-
-impl ServiceAccountRestrictions {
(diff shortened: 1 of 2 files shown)


src/auth/src/credentials/service_account.rs
```diff
@@ -92,33 +92,104 @@ use tokio::time::Instant;
...
+/// specifying either OAuth 2.0 [scopes] or a [JWT] audience.
+///
+/// It ensures that only one of these access specifiers can be applied
+/// for a given credential setup.
+///
+/// [JWT]: https://google.aip.dev/auth/4111
+/// [scopes]: https://developers.google.com/identity/protocols/oauth2/scopes
 #[derive(Debug)]
-enum ServiceAccountRestrictions {
+pub enum AccessSpecifier {
```
coryan, line 104: I think you want this to be `#[derive(Clone, Debug, PartialEq)]`

coryan (approved):

PR #2114, opened 2025-05-12
impl(generator/rust)!: prepare to support additional bindings

### LINKED ISSUE #2104: Change code generation in Rust to take no parameters for RPC methods
See #2068, we will need to change all the generated code to (1) only take `&self` as a parameter in the methods starting an RPC, and (2) perform some runtime validation to verify (at least) the first binding parameter is set to something.

A future change will support more bindings, and will validate the format of the path parameters.

coryan:
Request may have multiple "bindings" for HTTP+JSON. The same RPC may be sent to
different URL paths based on what fields are set in the request, and even on
the format of these fields.

We believe that supporting multiple bindings is valuable, and the only way to
do so is to break the current client library surface. The request methods no
longer require the "path parameters" because, depending on the binding,
different sets of parameters may be chose for the path.

If these is any consolation, the new API will allow us to validate the format
of common request fields such as `name` and `parent`. That will be the subject
of future PRs though.

Fixes #2104

### DIFF
--- a/generator/internal/rust/codec.go
+++ b/generator/internal/rust/codec.go
@@ -925,9 +925,33 @@ func derefFieldPath(fieldPath string, message *api.Message, state *api.APIState)
 	return expression.String()
 }
 
+func typezFieldPath(fieldPath string, message *api.Message, state *api.APIState) api.Typez {
+	typez := api.UNDEFINED_TYPE
+	components := strings.Split(fieldPath, ".")
+	msg := message
+	for _, name := range components {
+		if msg == nil {
+			slog.Error("cannot build full expression", "fieldPath", fieldPath, "message", msg)
+			return typez
+		}
+		for _, field := range message.Fields {
+			if name != field.Name {
+				continue
+			}
+			typez = field.Typez
+			if field.Typez == api.MESSAGE_TYPE {
+				msg = state.MessageByID[field.TypezID]
+			}
+			break
+		}
+	}
+	return typez
+}
+
 type pathArg struct {
-	Name     string
-	Accessor string
+	Name          string
+	Accessor      string
+	CheckForEmpty bool
 }
 
 func httpPathArgs(h *api.PathInfo, method *api.Method, state *api.APIState) []pathArg {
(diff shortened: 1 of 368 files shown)


src/generated/api/apikeys/v2/src/transport.rs
```diff
@@ -124,18 +152,15 @@ impl super::stub::ApiKeys for ApiKeys {
         options: gax::options::RequestOptions,
     ) -> Result<gax::response::Response<longrunning::model::Operation>> {
         let options = gax::options::internal::set_default_idempotency(options, false);
+        let path = format!("/v2/{}", {
+            &req.key
+                .as_ref()
+                .ok_or_else(|| gaxi::path_parameter::missing("key"))?
+                .name
```
dbolduc, line 159: wouldn't we check if this is empty?

Also funny aside: the `name` field is annotated as `OUTPUT_ONLY` despite being part of the path. :upside_down_face: 

https://github.com/googleapis/googleapis/blob/6942862f40d7a65b8e65ec491605d90d48458eaa/google/api/apikeys/v2/resources.proto#L49

generator/internal/rust/codec.go
```diff
@@ -925,9 +925,33 @@ func derefFieldPath(fieldPath string, message *api.Message, state *api.APIState)
...
+			slog.Error("cannot build full expression", "fieldPath", fieldPath, "message", msg)
+			return typez
+		}
+		for _, field := range message.Fields {
+			if name != field.Name {
+				continue
+			}
+			typez = field.Typez
+			if field.Typez == api.MESSAGE_TYPE {
+				msg = state.MessageByID[field.TypezID]
```
dbolduc, line 943: I am confused by what we are doing with `msg` here. Is it supposed to be recursive?

dbolduc (approved):
I think we are intending to report when string fields are empty (and thus our path leads to horrible errors).

I am confused by how (or if) the implementation handles the nested message case.

But feel free to merge this PR because everything else looks good and it touches like a bajillion files.

generator/internal/rust/codec.go
coryan: Yes, it was.  Fixed.

My tests were completely ineffective because they reused the field names. Fixed those too.

coryan (commented):
PTAL

dbolduc (approved):
Thanks for fixing

generator/internal/rust/codec.go
dbolduc: >  they reused the field names. 

Ah, I could not explain why some of the test cases worked.

Issue #2067, opened 2025-05-09
Flake in sql admin integration test


coryan:
The test failed:

https://console.cloud.google.com/cloud-build/builds;region=us-central1/fcb019d6-f163-4a63-b006-649cf6f9b4f6?project=rust-sdk-testing

```
---- driver::run_sql::with_tracing_retry_enabled stdout ----
--
07:53:37 |  
07:53:37 | Testing insert sql instance
07:53:37 | SUCCESS on insert sql instance: Operation { kind: "sql#operation", target_link: "https://www.googleapis.com/v1/projects/rust-sdk-testing/instances/rust-sdk-testing-uhexzkrth5nueiu6i8qw4huy3d84fhdz7um5ujnljoh765rdquwcnjckt8zhvbe7", status: Pending, user: "integration-test-runner@rust-sdk-testing.iam.gserviceaccount.com", insert_time: Some(Timestamp { seconds: 1746791599, nanos: 498000000 }), start_time: None, end_time: None, error: None, api_warning: None, operation_type: Create, import_context: None, export_context: None, backup_context: None, name: "7875ad6a-602b-4b64-a900-780a00000032", target_id: "rust-sdk-testing-uhexzkrth5nueiu6i8qw4huy3d84fhdz7um5ujnljoh765rdquwcnjckt8zhvbe7", self_link: "https://www.googleapis.com/v1/projects/rust-sdk-testing/operations/7875ad6a-602b-4b64-a900-780a00000032", target_project: "rust-sdk-testing", acquire_ssrs_lease_context: None, _unknown_fields: {} }
07:53:37 | Testing get sql instance
[...]

PR #2117, opened 2025-05-13
test(integration-tests): disable flaky assertions

coryan:
Part of the work for #2067


dbolduc (approved):
Strange that we can `get()` the instance but not `list()` it right after. 🤷

PR #2120, opened 2025-05-13
ci(auth): deflake UC test

### LINKED ISSUE #2119: Flake in `credentials::user_account::test::token_provider_full`
```
---- credentials::user_account::test::token_provider_full stdout ----
endpoint = http://127.0.0.1:37885/token

thread 'credentials::user_account::test::token_provider_full' panicked at src/auth/src/credentials/user_account.rs:724:9:
now: Instant { tv_sec: 417, tv_nsec: 480651973 }, expires_at: Some(Instant { tv_sec: 4017, tv_nsec: 480446569 })


failures:
    credentials::user_account::test::token_provider_full
```

dbolduc:
Fixes #2119 

When we switched to the builder pattern, we started including the token cache, which fetches a token upon credential creation, not just on calls for the `token()`.

We were comparing the expiration time to a time from between when the credentials are created and from when we ask for a token.

We want to compare the expiration time to a time from before the credentials are created.

### DIFF
--- a/src/auth/src/credentials/user_account.rs
+++ b/src/auth/src/credentials/user_account.rs
@@ -694,6 +694,7 @@ mod test {
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn token_provider_full() -> TestResult {
+        let now = std::time::Instant::now();
         let response = Oauth2RefreshResponse {
             access_token: "test-access-token".to_string(),
             expires_in: Some(3600),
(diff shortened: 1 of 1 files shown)


coryan:
> which fetches a token upon credential creation, not just on calls for the `token()`.
> 
> We were comparing the expiration time to a time from between when the credentials are created and from when we ask for a token.
> 
> We want to compare the expiration time to a time from before the credentials are created.

Maybe we should be controlling how time advances in these tests. We have control of both the service and the client.

In any case, this looks less flakey.

coryan (approved):

dbolduc:
> we should be controlling how time advances in these tests.

Probably

Issue #2037, opened 2025-05-08
Support LROs over gRPC


dbolduc:
As written, our transport code cannot convert messages with `Any` fields. LROs use `Any` in their `metadata` and `result` fields.

We cannot (in general) convert from a `prost_types::Any` -> `wkt::Any`. If we know the encoded type beforehand, or set of possible types, we can convert (after some copying).

We need to write some extra code that:
- extracts the expected `prost_types::M` from the `prost_types::Any`
- converts that `prost_types::M` -> our `M`
- packs the result into our `wkt::Any`

For a given LRO, we know the expected metadata and result types from proto annotations. We will use these.

For the `GetOperation`, we will want to match on the typename (considering all possible metadata types and all possible result types in a service) to perform the conversion.

---

- [x] `FromProto` can error
- [x] Cleanup `to_gax_response`
- [x] Generate `longrunning`, with stubbed out conversions for `Operation`
- [x] Implement `any_to_prost` conversion for StorageControl
- [x] Implement `any_from_prost` conversion for StorageControl
- [x] Implement `Operation` conversions
- [x] Add `rename_folder` to client with an integration test
- [x] Generator support for `Operation` conversions

PR #2125, opened 2025-05-13
impl(storage-control): unimplemented LRO internals

dbolduc:
Part of the work for #2037 

Generate LROs in `storage-control`. Note that the RPCs do not make it to the client surface. (The messages do, and that is fine).

We omit conversion functions for `Operation` and provide our own (which are initially unimplemented.

The plan is to have the generator own these conversion functions and generate them as part of `transport.rs`. (The conversion functions are particular to a service).


coryan (approved):

src/gax-internal/src/prost.rs
```diff
@@ -21,6 +21,8 @@ use std::collections::BTreeMap;
 pub enum ConvertError {
     #[error("enum {0} does not contain an integer value")]
     EnumNoIntegerValue(&'static str),
+    #[error("Conversion unimplemented")]
+    Unimplemented,
```
coryan, line 25: I think  we should mark this `enum` as `#[non_exhaustive]`.

coryan (approved):

Issue #2039, opened 2025-05-08
Complete implementation for `v1::Object`


coryan:
We need to complete the code here:

https://github.com/googleapis/google-cloud-rust/blob/32ef2f3dc7419710b8fe9f9c2193d89cfacbdd55/src/storage/src/client.rs#L294

All the fields need to be present, and we need to convert them to and from a `control::model::Object`.

suzmue:
For reference, JSON API definition is: https://cloud.google.com/storage/docs/json_api/v1/objects#resource

Examples of the conversion to and from proto type:
- Go: https://github.com/googleapis/google-cloud-go/blob/037b55cf453e23451b59ee04077ca599e3ffe031/storage/storage.go#L1775
- C++: https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/storage/internal/grpc/object_request_parser.cc

PR #2127, opened 2025-05-13
impl(storage): add simple fields to the v1::Object type

suzmue:
Also adds a small test for some of the different types that we need to deserialize and convert to control::model::Object.

For #2039

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -372,14 +372,11 @@ mod v1 {
 
     #[cfg(test)]
     mod tests {
-        use serde_json::json;
-
         use super::*;
-        // use control::model; // For control::model::Object and control::model::ObjectChecksums
 
         #[test]
         fn test_deserialize_object() {
-            let json = json!({
+            let json = serde_json::json!({
                 // string fields:
                 "id": "obj1",
                 "name": "test-object.txt",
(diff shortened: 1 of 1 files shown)


src/storage/src/client.rs
```diff
@@ -325,6 +348,111 @@ mod v1 {
...
+                generation: 123,
+                metageneration: 456,
+                size: 789,
+                // boolean fields:
+                temporary_hold: true,
+                // number fields:
+                component_count: 5,
+                // datetime fields:
+                time_created: wkt::Timestamp::new(1747132200, 0)
+                    .expect("hardcoded timestamp value in object test should be in range"),
```
coryan, line 420: You may prefer:

```suggestion
                time_created: wkt::Timestamp::clamp(1747132200, 0)
```

src/storage/src/client.rs
```diff
@@ -312,6 +313,28 @@ mod v1 {
...
         #[serde_as(as = "serde_with::DisplayFromStr")]
         size: u64,
+        component_count: i32,
+        kms_key_name: String,
+        etag: String,
+        restore_token: String,
+        content_encoding: String,
+        content_disposition: String,
+        content_language: String,
+        cache_control: String,
```
coryan, line 323: no change needed: it has always been unclear to me if these are `Option<String>` or `String`.  Maybe the v2 proto specification is clearer?

coryan (approved):

src/storage/src/client.rs
suzmue: Done, thanks!

Issue #2072, opened 2025-05-09
Remove token() from CredentialsProvider trait


sai-sunder-s:
The `CredentialsProvider` trait has 2 ways to get a token currently:
- `token()`
- `headers()`

Gax is currently using only `headers()`. There is no currently known purpose for getting a token directly, so remove it from the public interface.

Issue #2073, opened 2025-05-09
Make Token struct pub(crate)


sai-sunder-s:
Since `fn token(&self) -> Result<crate::token::Token>` is going to be removed from the public interface, which is the only place where `Token` is exposed in our public interface, we can make `Token` into `pub(crate)`.

PR #2131, opened 2025-05-13
refactor(auth)!: Remove token method credential provider

harkamaljot:
Clients only need access to headers, therefore removing token method from the credential provider. Additionally, as the token is now an internal implementation detail, `std::time::Instant` has been replaced with `tokio::time::Instant` to eliminate test flakiness.

The PR resolves https://github.com/googleapis/google-cloud-rust/issues/2072 and unblocks https://github.com/googleapis/google-cloud-rust/issues/2073


src/auth/src/credentials.rs
```diff
@@ -657,6 +616,28 @@ mod test {
...
+    pub(crate) fn get_token_from_headers(headers: &HeaderMap) -> String {
+        let token = headers.get(AUTHORIZATION).unwrap();
+        token
+            .to_str()
+            .unwrap()
+            .split_whitespace()
+            .nth(1)
+            .unwrap()
+            .to_string()
+    }
```
coryan, line 628: I know this is test code, and crashes / assertions are fine, but if you return `Option<String>` you can simplify this code a lot:

```suggestion
    pub(crate) fn get_token_from_headers(headers: &HeaderMap) -> Option<String> {
        let token = headers.get(AUTHORIZATION)?;
        token
            .to_str()?
            .split_whitespace()
            .nth(1)
            .map(|v| v.to_string())
    }
```

And then the call sites can do an `.unwrap()` if that is appropriate.

src/auth/src/credentials.rs
```diff
@@ -657,6 +616,28 @@ mod test {
...
+
+    pub(crate) fn get_token_type_from_headers(headers: &HeaderMap) -> String {
+        let token = headers.get(AUTHORIZATION).unwrap();
```
coryan, line 631: Same here.

src/auth/src/credentials/mds.rs
```diff
@@ -592,11 +550,13 @@ mod test {
...
-            .with_scopes(scopes)
-            .with_endpoint(endpoint)
-            .build()?;
-        let token = mdsc.token(Extensions::new()).await?;
+        let token = MDSAccessTokenProvider {
+            scopes: Some(scopes.clone()),
+            endpoint,
+        }
+        .token()
+        .await?;
```
coryan, line 558: Maybe your `Builder` should have a `pub(crate) fn build_token_provider()` that would make the tests more readable and easier to convince me that "oh, this is just doing the normal thing but getting a provider instead".  What do you think?

src/auth/src/credentials/user_account.rs
```diff
@@ -692,9 +652,8 @@ mod test {
         }
     }
 
-    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[tokio::test(start_paused = true)]
```
coryan, line 655: Neat.

src/auth/src/token.rs
```diff
@@ -15,7 +15,7 @@
 use crate::Result;
 use http::Extensions;
 use std::collections::HashMap;
-use std::time::Instant;
+use tokio::time::Instant;
```
coryan, line 18: Should we make the `struct Token` private (or `pub(crate)` as part of this PR)? A follow up PR soon seems fine.

If we ever resurrect `Token`: we may use private fields and an accessor to return `std::time::Instant`.

coryan (approved):
A few nits and suggestions, nothing blocking I think.

src/auth/src/token.rs
harkamaljot: Thanks for bringing this up. We've already got a task open for it at https://github.com/googleapis/google-cloud-rust/issues/2073. I'll be sure to include the changes in a follow-up PR.

src/auth/src/credentials/mds.rs
harkamaljot: Thanks for the suggestion! You're right, the current approach relies on understanding how the code work, so updating the code will make it easier to follow. I'll update the code to add pub(crate) fn build_token_provider() for the credentials.

src/auth/src/credentials.rs
harkamaljot: I think to_str() return back a result, so this won't work, but thanks for the suggestion, simplifying it using something like this

```rust
        headers.get(AUTHORIZATION)
            .and_then(|token_value| token_value.to_str().ok())
            .and_then(|s| s.split_whitespace().nth(1))
            .map(|s| s.to_string())
```

src/auth/src/credentials/mds.rs
harkamaljot: I tried this approach and it works well(just need to clone quota_project_id, universe_domain) for all of the credentials except user_account.rs(https://github.com/googleapis/google-cloud-rust/blob/main/src/auth/src/credentials/user_account.rs#L202). As quota project is embedded inside AuthorizedUser Struct, so creating a function there gets tricky as now it involves cloning a whole struct, which seems like a anti-pattern to me. I tried a few other approaches, however I wasn't able to get this approach working for user_account.rs, so I would suggest we avoid creating build_token_provider method for user_account.rs. what do you think @coryan?

coryan: Sad, but sounds like something we can fix later. Go ahead and merge.

PR #2135, opened 2025-05-14
feat: telemetry includes REST vs. gRPC fields

### LINKED ISSUE #2087: Include `rest/vvv` value in the `x-goog-api-client` header
go/actools-regapic-tracking-metrics for details

coryan:
The `rest/{vvv}` field is required for showcase and part of the client
library specification.

Fixes #2087

### DIFF
--- a/src/gax-internal/src/api_header.rs
+++ b/src/gax-internal/src/api_header.rs
@@ -36,7 +36,7 @@ mod build_info {
 
 impl XGoogApiClient {
     /// Format the struct as needed for the `x-goog-api-client` header.
-    pub fn header_value(&self) -> String {
+    pub fn rest_header_value(&self) -> String {
         // Strip out the initial "rustc " string from `RUSTC_VERSION`. If not
         // found, leave RUSTC_VERSION unchanged.
         let rustc_version = build_info::RUSTC_VERSION;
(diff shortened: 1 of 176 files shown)


src/gax-internal/src/api_header.rs
```diff
@@ -48,7 +48,25 @@ impl XGoogApiClient {
         let gax_version = build_info::PKG_VERSION;
```
dbolduc, line 48: Observation: gax version is really gaxi version. gaxi is more interesting to us

src/gax-internal/src/api_header.rs
```diff
@@ -48,7 +48,25 @@ impl XGoogApiClient {
...
 
         format!(
-            "gl-rust/{rustc_version} gax/{gax_version} {}/{}",
+            "gl-rust/{rustc_version} gax/{gax_version} rest/{gax_version}-reqwest {}/{}",
```
dbolduc, line 51: Q: why duplicate `{gax_version}`?

Ideally would we capture the version of `reqwest`?

src/gax-internal/src/api_header.rs
```diff
@@ -48,7 +48,25 @@ impl XGoogApiClient {
...
+        let rustc_version = build_info::RUSTC_VERSION;
+        let rustc_version = rustc_version
+            .strip_prefix("rustc ")
+            .unwrap_or(build_info::RUSTC_VERSION);
+
+        // Capture the gax version too.
+        let gax_version = build_info::PKG_VERSION;
+
+        format!(
+            "gl-rust/{rustc_version} gax/{gax_version} grpc/{gax_version}-tonic {}/{}",
```
dbolduc, line 69: Same Q here about `tonic`

dbolduc (approved):

src/gax-internal/src/api_header.rs
coryan: ack. go/cloud-api-headers says it should be called `gax/`.  We could try to capture both.

src/gax-internal/src/api_header.rs
coryan: The spec (go/actools-regapic-tracking-metrics) says we can use either the gax version or the version of the http client.  And I cannot find a way to capture the version of `reqwest` or any other dependency.

src/gax-internal/src/api_header.rs
coryan: Same answer.  We can use either, and I cannot find a way to capture the tonic version.

Issue #1358, opened 2025-02-20
Provide an Etag-like value as part of `Credentials`


coryan:
`Credentials` is (soon) going to cache the tokens and headers as these are long lived. Yay!

Sometimes the users of `Credentials` will want to cache data derived from these headers. For example, when using Tonic we may want to convert the headers to a [MetadataMap](https://docs.rs/tonic/latest/tonic/metadata/struct.MetadataMap.html) and we may want to reuse this map as long as it is still valid.

If we changed this:

https://github.com/googleapis/google-cloud-rust/blob/6ced131d63a7395b5307a2fbbc55add46900e47e/src/auth/src/credentials.rs#L89-L91

To something like:

```rust
    pub async fn get_headers(&self, etag: Option<&str>) -> Result<CacheableHeaders> { ... }
```

Where `CacheableHeaders` is something like:

```rust
enum CacheableHeaders {
  NotModified, // Indicates that the values associated with `etag` remain valid.
  New { etag: String, headers: Vec<(HeaderKey, HeaderValue)> }, // The new values. 
}
```

The user then can do something like:

```rust
struct Client {
  credentials:  auth::Credentials,
  current_etag: Option<String>,
  current_headers: HeaderMap,
}


impl Client {
   async fn refresh_headers(&mut self) -> Result<HeaderMap> {
       match credentials.get_headers(self.current_headers.as_ref()).await? {
           CacheableHeaders::NotModified => {
            },
            CacheableHeaders::New { etag: e, headers: h } => {
               // Update current_headers with the values from New.
              self.etag = Some(e);
              self.current_headers = h.into_iter().map( blah blah).collect();
               // and then ...
              return Ok(self.current_headers.clone());
            }
       };
       return Ok(self.current_headers.clone());
   }
}
```

coryan:
ETag is a terrible name, but it tells you where I stole the idea from.

sai-sunder-s:
Should `match credentials.get_headers(self.current_headers.as_ref()).await? {` instead be something like
`match credentials.get_headers(Some(self.current_etag.as_ref())).await? {`

coryan:
> Should `match credentials.get_headers(self.current_headers.as_ref()).await? {` instead be something like `match credentials.get_headers(Some(self.current_etag.as_ref())).await? {`

Yes, you are probably right.

sai-sunder-s:
@coryan do you need the same Cacheable concept for `token()` as well, or only for `headers()`?

Issue #2036, opened 2025-05-08
Auth: Allow user to pass in request options for headers and token method


harkamaljot:
- [x]  Update the interface of header and token methods to accept optional Extensions and header method to return HeaderMap. This would be a breaking change.
- [x] Update return type of header and token methods to return back a cachable resource, this would also be a breaking change.

- [x] Add implementation for entity_tag in the token_cache so that callers can use it.

~~- [ ] Add implementation for request_deadline in the corresponding token methods.~~

PR #2145, opened 2025-05-14
feat(auth)!: Implement ETag-based caching for authentication headers

harkamaljot:
As part of https://github.com/googleapis/google-cloud-rust/issues/2036, this PR introduces the following changes:

Key changes:
- Introduces `EntityTag` and `CacheableResource<T>` to represent and manage cached authentication data.
- Modifies `CredentialsProvider::headers()` and related methods to return `Result<CacheableResource<HeaderMap>>`, allowing callers to reuse cached headers if unchanged.
- `TokenCache` now generates and manages ETags for tokens, which are then used by `CredentialsProvider` implementations to determine header freshness.
- Updates downstream consumers in `gax-internal` and `storage` to correctly handle the new `CacheableResource` type, expecting new headers if no ETag was provided.

This PR fixes https://github.com/googleapis/google-cloud-rust/issues/1358 and https://github.com/googleapis/google-cloud-rust/issues/2036


src/storage/src/client.rs
```diff
@@ -211,11 +224,21 @@ impl Storage {
...
             .map_err(Error::authentication)?;
+        let auth_headers = match cached_auth_headers {
+            CacheableResource::New { data, .. } => Ok(data),
+            CacheableResource::NotModified => {
+                Err(Error::authentication(CredentialsError::from_str(
+                    false,
+                    "Auth headers not refreshed; client requires new headers to proceed with the request.",
+                )))
+            }
+        }?;
```
coryan, line 240: Neat trick.  A bit easy to miss. Consider:

```suggestion
        };
   let auth_headers = auth_headers?;
```

src/auth/src/credentials.rs
```diff
@@ -29,6 +29,45 @@ use std::sync::Arc;
...
+/// to prevent accidental logging of potentially sensitive information.
+#[derive(Clone, PartialEq, Default)]
+pub struct EntityTag(pub(crate) String);
+
+impl EntityTag {
+    /// Creates a new `EntityTag` from a string value.
+    pub fn new(value: String) -> Self {
+        Self(value)
+    }
+}
```
coryan, line 49: Consider:

```suggestion
#[derive(Clone, PartialEq, Default)]
pub struct EntityTag(u64);

use std::sync::atomic::{AtomicU64, Ordering};
static ENTITY_TAG_GENERATOR: AtomicU64 = AtomicU64::new(0);

impl EntityTag {
    pub fn new() -> Self {
        let value = ENTITY_TAG_GENERATOR.fetch_add(1, Ordering::SeqCst);
        Self(value)
    }
}
```

Basically, you don't need the tags to be related in any way to the input string, you just need them to be unique for the lifetime of the program.  An atomic gives you that and they are faster to generate. And there are no security implications for them either.

src/gax-internal/src/grpc.rs
```diff
@@ -140,11 +141,21 @@ impl Client {
...
             .map_err(Error::authentication)?;
+
+        let auth_headers = match cached_auth_headers {
+            CacheableResource::New { data, .. } => Ok(data),
+            CacheableResource::NotModified => {
+                Err(Error::authentication(CredentialsError::from_str(
+                    false,
+                    "Auth headers not refreshed; client requires new headers to proceed with the request.",
+                )))
+            }
```
coryan, line 157: I wonder if this should be `unreachable!()` ... at least until we provide something useful in the `http::Extensions` value.

src/gax-internal/src/http.rs
```diff
@@ -128,11 +127,22 @@ impl ReqwestClient {
...
             .map_err(Error::authentication)?;
+
+        let auth_headers = match cached_auth_headers {
+            CacheableResource::New { data, .. } => Ok(data),
+            CacheableResource::NotModified => {
+                Err(Error::authentication(CredentialsError::from_str(
+                    false,
+                    "Auth headers not refreshed; client requires new headers to proceed with the request.",
+                )))
+            }
```
coryan, line 143: Ditto, I think this can be `unreachable!()` until we provide some value.

src/storage/src/client.rs
coryan: Separately, we need to refactor this code in the storage crate. Filed #2209

src/auth/src/token_cache.rs
```diff
@@ -74,23 +71,53 @@ impl CachedTokenProvider for TokenCache {
...
 }
 
-async fn refresh_task<T>(token_provider: T, tx_token: watch::Sender<Option<Result<Token>>>)
-where
+fn get_etag_from_token(token: &str) -> String {
+    let mut hasher = blake3::Hasher::new();
+    hasher.update(token.as_bytes());
+    let hash_bytes = hasher.finalize();
+    format!("\"{}\"", hash_bytes.to_hex())
+}
```
coryan, line 105: If you use my idea with the atomic, this code goes away and you can remove the dependencies.

coryan (changes requested):
I think we can simplify this, let me know if my idea makes sense.

src/auth/src/credentials.rs
harkamaljot: that's really smart! thanks for the suggestion.

src/gax-internal/src/grpc.rs
harkamaljot: yeah that make sense because ideally we don't expect not modified to not to be called here for now.

src/gax-internal/src/http.rs
harkamaljot: yeah that make sense because ideally we don't expect not modified to not to be called here for now.

src/auth/src/token_cache.rs
```diff
@@ -74,23 +71,46 @@ impl CachedTokenProvider for TokenCache {
...
+        let (token, tag) = self.get_latest_token_and_entity_tag().await?;
+        if given_tag.is_some_and(|given_tag| tag.eq(given_tag)) {
+            return Ok(CacheableResource::NotModified);
+        } else {
+            return Ok(CacheableResource::New {
+                entity_tag: tag,
+                data: token,
+            });
+        }
+    }
```
coryan, line 88: Consider:

```suggestion
        let (data, entity_tag) = self.get_latest_token_and_entity_tag().await?;
        match extensions.get::<EntityTag>() {
            Some(tag) if tag == entity_tag => Ok(CacheableResource::NotModified),
            _ => Ok(CacheableResource::New { entity_tag, data }),
        }
    }
```

src/auth/src/headers_util.rs
```diff
@@ -13,14 +13,32 @@
...
 use crate::errors;
+use crate::token::Token;
 
 use http::HeaderMap;
 use http::header::{AUTHORIZATION, HeaderName, HeaderValue};
 
 const API_KEY_HEADER_KEY: &str = "x-goog-api-key";
 
+/// A utility function to create cacheable headers.
+pub(crate) fn build_cacheable_headers(
```
coryan, line 26: It seems to me that we are computing the headers even if the token has not changed. Is that the case?  If so, then we should optimize that away: we can build the headers only once.  Even if I am right (1) we should not make that optimization in this PR, we should file a feature request to implement later, (2) it probably belongs in the `http.rs` and `grpc.rs` files.

src/auth/src/credentials.rs
```diff
@@ -29,6 +29,44 @@ use std::sync::Arc;
...
+
+impl std::fmt::Debug for EntityTag {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("EntityTag")
+            .field("tag", &"[censored]")
+            .finish()
+    }
+}
```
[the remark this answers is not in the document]
harkamaljot: yes that is correct.

src/auth/src/token_cache.rs
```diff
@@ -27,41 +28,37 @@ const SHORT_REFRESH_SLACK: Duration = Duration::from_secs(10);
...
         });
 
         Self { rx_token }
     }
-}
 
-#[async_trait::async_trait]
-impl CachedTokenProvider for TokenCache {
-    async fn token(&self, _extensions: Extensions) -> Result<Token> {
+    async fn get_latest_token_and_entity_tag(&self) -> Result<(Token, EntityTag)> {
```
sai-sunder-s, line 48: I believe in Rust we do not prefix methods with `get`?

src/auth/src/headers_util.rs
harkamaljot: Yes, you are correct, we need to implement that in `http.rs` and `grpc.rs` files. I created a issue (https://github.com/googleapis/google-cloud-rust/issues/2220) to track this work.

src/auth/src/credentials.rs
```diff
@@ -557,6 +613,8 @@ pub mod testing {
     use http::{Extensions, HeaderMap};
     use std::sync::Arc;
 
+    use super::{CacheableResource, EntityTag};
```
coryan, line 616: nit: you should keep all the `use` declarations together so `cargo fmt` can sort them. Feel free to fix these nits in a separate PR.

coryan (dismissed):

src/auth/src/credentials.rs
harkamaljot: cargo fmt, doesn't seem to do that for me, let me do that here only as this is the only place.

harkamaljot: nvm I see what you meant here.

coryan (approved):

PR #2153, opened 2025-05-15
impl(storage): add acl field to v1::Object

suzmue:
For #2039


src/storage/src/client.rs
```diff
@@ -332,10 +332,75 @@ mod v1 {
...
+                .set_role(value.role)
+                .set_email(value.email)
+                .set_domain(value.domain)
+                .set_entity_id(value.entity_id)
+                .set_etag(value.etag)
+                .set_project_team(to_control_project_team(value.project_team))
+        }
+    }
+
+    fn to_control_project_team(p: Option<ProjectTeam>) -> Option<control::model::ProjectTeam> {
```
suzmue, line 395: Was unable to do fromFrom<Option> for Option<> since this crate does not own the Option type I believe, but if there is something I was missing in order to make that work please let me know.

coryan (changes requested):
Some ideas for improvement, but I am not sure, you need to try them.

Testing could use some improvement.  I think if we run `cargo mutants` on this test it will find a lot of testing holes.

src/storage/src/client.rs
suzmue: Done. thank you!

coryan (dismissed):

coryan:
I just merged the changes for `set_or_clear_`, you may want to rebase to make sure that did not break any of your code (I think it does).

coryan (approved):

src/storage/src/client.rs
```diff
@@ -448,6 +535,7 @@ mod v1 {
 
             assert_eq!(got.generation, 123);
             assert_eq!(got.bucket, "projects/_/buckets/my-bucket");
+            assert_eq!(got.acl.len(), 2)
```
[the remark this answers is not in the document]
suzmue: Re-did the tests. Done.

PR #2156, opened 2025-05-15
cleanup!: remove error-prone setters for `oneof`

### LINKED ISSUE #2150: Remove helpers to set `oneof` branches
## Problem

For each `oneof` group of fields we generate an enum and a number of helpers to set the enum variants. These helpers (despite the documentation warnings) may lead consumers of the library to believe that more than one value can be set at a time. That confusion has led to data loss, we should avoid it.

## Background

Something like:

```proto
message FooBar {
  oneof widget {
    int32 quux = 1;
    string bar = 2;
    Baz baz = 3;
  }
}
```

Generates:

```rust
// ... ... ... Generated Rust ... ... ...
#[serde_with::skip_serializing_none]
#[derive(Debug, Default, PartialEq, serde::Deserialize, serde::Serialize)]
#[serde(rename_all = "camelCase")]
#[non_exhaustive]
struct FooBar {
    widget: Option<foo_bar::Widget>,
}
```

And an enum in the right namespace:

```rust
pub mod foo_bar { 
    #[non_exhaustive]
    #[derive(Debug, PartialEq, serde::Deserialize, serde::Serialize)]
    #[serde(rename_all = "camelCase", untagged)]
    pub enum Widget {
      Quux { quux: i32 },
      Bar { bar: String },
      Baz(Baz),
    }
}
```

With some helpers:

```rust
impl FooBar {
  pub fn set_widget<V: Into<foo_bar::Widget>>(self, v: V) -> Self { self.widget = Some(v.into()); self }
  pub fn set_quux<V: Into<i32>>(self, v: V) -> Self { self.widget = Some(foo_bar::Widget::Quux(v.into())); self }
  // ... more set_*() helpers
```

## Proposal

Keep `set_widget()` remove `set_quux()` and the other setters for the Widget variants.

coryan:
Setters for `oneof` are error-prone: they give the impression that the
message may contain more than one of these values at a time. Providing
just `set_{oneof_name}()` is less error prone: it consumes a Rust `enum`
and the behavior of those is well understood.

Admittedly the types are harder to use, and some services have a lot of
complex, deeply nested `oneof` fields.

Fixes #2150

### DIFF
--- a/generator/internal/rust/templates/common/message.mustache
+++ b/generator/internal/rust/templates/common/message.mustache
@@ -120,56 +120,6 @@ impl {{Codec.Name}} {
             _ => std::option::Option::None,
         })
     }
-
-    /// Sets the value of [{{Group.Codec.FieldName}}][{{Codec.FQMessageName}}::{{Group.Codec.FieldName}}]
-    /// to hold a `{{Codec.BranchName}}`.
-    ///
-    /// Note that all the setters affecting `{{Group.Codec.FieldName}}` are
-    /// mutually exclusive.
-    {{#Deprecated}}
-    #[deprecated]
-    {{/Deprecated}}
-    {{#Singular}}
-    pub fn set_{{Codec.SetterName}}<T: std::convert::Into<{{{Codec.FieldType}}}>>(mut self, v: T) -> Self {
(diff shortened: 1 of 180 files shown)


coryan:
We decided not to move forward with this.

Issue #944, opened 2025-02-02
Generate library for `google/storage/control/v2/storage_v2.yaml`


coryan:
Blocked on:

- [x] StorageControl service lacks HTTP annotations
- [ ] #439 -- or at least we won't have a fully-featured generated library until that is implemented too

----

coryan:
We could generate this as a gRPC-based library. We would need to fix LROs to support gRPC-based libraries.

PR #2166, opened 2025-05-16
feat(storage-control): add `RenameFolder` API

dbolduc:
Part of the work for #2037 and #944 

Add `RenameFolder` (an LRO) to the client surface, and use it from the integration test.

This is not the greatest integration test of LROs over gRPC, as the initial RPC returns a `done` Operation. It would be nice to exercise the polling loop (i.e. the `GetOperation` calls).

This RPC fails if a quota project is provided, let's see if one is set in our CI. :crossed_fingers:

### DIFF
--- a/src/integration-tests/src/storage.rs
+++ b/src/integration-tests/src/storage.rs
@@ -17,6 +17,7 @@ use gax::exponential_backoff::{ExponentialBackoff, ExponentialBackoffBuilder};
 use gax::options::RequestOptionsBuilder;
 use gax::paginator::ItemPaginator as _;
 use gax::retry_policy::RetryPolicyExt;
+use lro::Poller;
 use std::time::Duration;
 use storage_control::model::Bucket;
 use storage_control::model::bucket::iam_config::UniformBucketLevelAccess;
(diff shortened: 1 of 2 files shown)


src/storage-control/src/client.rs
```diff
@@ -460,6 +460,42 @@ impl Storage {
...
+    /// # use google_cloud_storage_control::client::Storage;
+    /// async fn example(client: &Storage) -> gax::Result<()> {
+    ///     use lro::Poller as _;
+    ///     let folder = client.rename_folder()
+    ///         .set_name("projects/_/buckets/my-bucket/folders/my-folder/my-subfolder/")
+    ///         .set_destination_folder_id("my-folder/my-renamed-subfolder/")
+    ///         .poller()
+    ///         .until_done()
+    ///         .await?;
+    ///     Ok(())
```
coryan, line 482: nit: maybe `println!("successfully created folder {folder:?}");`

src/integration-tests/src/storage.rs
```diff
@@ -235,8 +237,19 @@ async fn folders(client: &storage_control::client::Storage, bucket_name: &str) -
         "missing folder name {folder_name} in {folder_names:?}"
     );
 
+    println!("\nTesting rename_folder()");
+    let rename = client
```
coryan, line 241: I did not think about this before: can you delete a bucket that has folders in it? Even if the folders are empty?  If we cannot, then the `cleanup_stale_buckets()` function needs some fixes.

coryan (dismissed):

src/storage-control/src/client.rs
dbolduc: Done

src/integration-tests/src/storage.rs
dbolduc: > can you delete a bucket that has folders in it?

Yes, I commented out the `delete_folder()` call in the integration test, and the subsequent `delete_bucket()` call still succeeded.

coryan (approved):

PR #2186, opened 2025-05-16
chore: bump versions for all libraries

coryan:
We already made breaking changes across the board.


Cargo.toml
```diff
@@ -327,32 +327,32 @@ num-bigint-dig = { version = "0.8" }
 auth                          = { version = "0.20", path = "src/auth", package = "google-cloud-auth" }
 gax                           = { version = "0.22", path = "src/gax", package = "google-cloud-gax" }
 gaxi                          = { version = "0.3", path = "src/gax-internal", package = "google-cloud-gax-internal" }
```
dbolduc, line 329: these are upstream of `wkt`. They need a bump, right?

Cargo.toml
```diff
@@ -327,32 +327,32 @@ num-bigint-dig = { version = "0.8" }
...
-iam_v1                        = { version = "0.3", path = "src/generated/iam/v1", package = "google-cloud-iam-v1" }
-location                      = { version = "0.3", path = "src/generated/cloud/location", package = "google-cloud-location" }
-longrunning                   = { version = "0.24", path = "src/generated/longrunning", package = "google-cloud-longrunning" }
+iam_v1                        = { version = "0.4", path = "src/generated/iam/v1", package = "google-cloud-iam-v1" }
+location                      = { version = "0.4", path = "src/generated/cloud/location", package = "google-cloud-location" }
+longrunning                   = { version = "0.25", path = "src/generated/longrunning", package = "google-cloud-longrunning" }
 lro                           = { version = "0.2", path = "src/lro", package = "google-cloud-lro" }
```
dbolduc, line 333: same

coryan: Fixed.

dbolduc: Push?

dbolduc (approved):

Cargo.toml
coryan: Good call. AFAICT, `gax` exposes `wkt::Any` indirectly, via:

https://docs.rs/google-cloud-gax/latest/google_cloud_gax/error/rpc/enum.StatusDetails.html#variant.Other

That makes the breaking changes in `wkt` breaking in `gax`.

`gax-internal` was already bumped since the last release.

`auth` seems unaffected?  I bumped the patch version for `auth`, to reflect the dependency changes.

Cargo.toml
coryan: Sorry, yes, I made the comment before the push.  LRO is affected because `longrunning` carries the `Any` thing.

PR #2192, opened 2025-05-16
cleanup!: better mapping for enum value names

### LINKED ISSUE #2178: Sidekick should preserve the prefix enums values that start with a number
Many enum values repeat the name of the enum as a prefix, such as `enum Edition { EDITION_FOO, EDITION_7 }`. We strip the prefix to make them more readable, so the generated code is `Edition::Foo`.  Where this results in a number we use `Edition::_7`.

We should preserve the prefix in this case.

coryan:
Do not strip the prefix if the resulting variant would start with a
number.

Fixes #2178


dbolduc (approved):

generator/internal/rust/codec.go
```diff
@@ -832,26 +832,20 @@ func enumValueVariantName(e *api.EnumValue) string {
 	// The naming conventions being what they are, we need to test with a couple
 	// of different combinations. In particular, names with numbers, such as
 	// `InstancePrivateIpv6GoogleAccess` make life difficult for everyone.
```
suzmue, line 834: nit: I think one more line of elaboration for what the issues is for InstancePrivateIpv6GoogleAccess. Something like " since there is ambiguity when converting to screaming snake case with numbers where either INSTANCE_PRIVATE_IPV_6_GOOGLE_ACCESS or INSTANCE_PRIVATE_IPV6_GOOGLE_ACCESS may be valid."

I'm not sure https://google.aip.dev/140 applies since this is for the enum name, not the field name, but that specifies that each word in a field name can't begin with number, so only INSTANCE_PRIVATE_IPV6_GOOGLE_ACCESS can actually be a correct snake case version of the name. If that is true, then maybe we should just fix the prefix and do the trimming once (so drop lines 836-839).

I think we can also consider dropping our dependency on strcase and write our own since AFAICT this is actually an ambiguous conversion so we may just want our own implementation we can depend on and convert for the rules we know the APIs follow with naming.

suzmue (approved):

generator/internal/rust/codec_test.go
```diff
@@ -2039,7 +2039,6 @@ func TestEnumValueVariantName(t *testing.T) {
 		Package: "test",
 		Values: []*api.EnumValue{
 			{Number: 0, Name: "ENUM_NAME_UNSPECIFIED"},
-			{Number: 1, Name: "ENUM_NAME1"},
```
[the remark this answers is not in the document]
coryan: Agree.

generator/internal/rust/codec.go
coryan: > nit: I think one more line of elaboration for what the issues is for InstancePrivateIpv6GoogleAccess. Something like " since there is ambiguity when converting to screaming snake case with numbers where either INSTANCE_PRIVATE_IPV_6_GOOGLE_ACCESS or INSTANCE_PRIVATE_IPV6_GOOGLE_ACCESS may be valid."

See https://github.com/googleapis/google-cloud-rust/pull/2208 happy to workshop the text in that PR.

> I'm not sure https://google.aip.dev/140 applies since this is for the enum name, not the field name, but that specifies that each word in a field name can't begin with number, so only INSTANCE_PRIVATE_IPV6_GOOGLE_ACCESS can actually be a correct snake case version of the name. If that is true, then maybe we should just fix the prefix and do the trimming once (so drop lines 836-839).

Ack.

> I think we can also consider dropping our dependency on strcase and write our own since AFAICT this is actually an ambiguous conversion so we may just want our own implementation we can depend on and convert for the rules we know the APIs follow with naming.

Worth considering. This thing with the numbers vs. words separator comes up often.

PR #2193, opened 2025-05-16
impl(storage): add owner, metadata, and encryption fields to v1::Object

suzmue:
For #2039

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -333,6 +333,9 @@ mod v1 {
         updated: wkt::Timestamp,
         custom_time: Option<wkt::Timestamp>,
         acl: Vec<ObjectAccessControl>,
+        owner: Option<Owner>,
+        customer_encryption: Option<CustomerEncryption>,
+        metadata: std::collections::HashMap<std::string::String, std::string::String>,
         // The following are excluded from the protos, so we don't really need to parse them.
         media_link: String,
         self_link: String,
@@ -342,17 +345,6 @@ mod v1 {
         //     "retention": {
         //       "retainUntilTime": "datetime",
         //       "mode": string
-        //     }
-        //     "metadata": {
-        //       (key): string
-        //     },
-        //     "owner": {
-        //       "entity": string,
-        //       "entityId": string
-        //     },
-        //     "customerEncryption": {
-        //       "encryptionAlgorithm": string,
-        //       "keySha256": string
         //     }
     }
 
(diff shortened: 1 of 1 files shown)


src/storage/src/client.rs
```diff
@@ -674,15 +754,15 @@ mod v1 {
...
-                } else {
-                    panic!("expected None, got {:?}", got); // false, lets get the error.
-                }
+            if let Some(from) = &from.project_team {
+                let got = got
+                    .project_team
+                    .clone()
+                    .expect("project_team should be None");
+                assert_eq!(got.project_number, from.project_number);
+                assert_eq!(got.team, from.team);
```
coryan, line 763: I suspect this can use some simplification, maybe:

```suggestion
            match (&from.project_time, &got.project_team) {
                (None, None) => {}, // yay!
                (Some(_), None) => panic!("expected a value in the project_team, {got:?}"),
                (None, Some(_)) => panic!("unexpected value in the project team, {got:?}"),
                (Some(f), Some(g)) => {assert_eq!( ... ); assert_eq!(... ); }, 
```

src/storage/src/client.rs
```diff
@@ -620,10 +683,27 @@ mod v1 {
             assert_eq!(got.custom_time, object.custom_time);
             assert_eq!(got.soft_delete_time, object.soft_delete_time);
             assert_eq!(got.hard_delete_time, object.hard_delete_time);
+            if let Some(object) = &object.owner {
```
coryan, line 686: Like below, these Some/None cases can be more readable with a `match (object.owner, got.owner)`...

src/storage/src/client.rs
```diff
@@ -402,6 +394,39 @@ mod v1 {
...
+                .set_entity_id(value.entity_id)
+        }
+    }
+
+    #[serde_with::serde_as]
+    #[derive(Debug, Default, serde::Deserialize, PartialEq, Clone)]
+    #[serde(default, rename_all = "camelCase")]
+    struct CustomerEncryption {
+        encryption_algorithm: String,
+        #[serde_as(as = "serde_with::base64::Base64")]
```
coryan, line 418: Ah, neat.

coryan (changes requested):

src/storage/src/client.rs
suzmue: Done! thanks for the suggestion.

coryan (approved):

PR #2195, opened 2025-05-17
cleanup(wkt)!: `Timestamp` always serialize

### LINKED ISSUE #2177: Remove unreachable variants in `wkt::*Error`
Both `AnyError` and `TimestampError` have unreachable variants.  The code that uses them never executes. The error conditions they represent are unreachable because the types have invariants to prevent those errors.

We should remove them and use `.expect("")` instead.

coryan:
Given the invariants in the `Timestamp` struct, the serialization can
never fail. So we can use `.expect()` instead of returning an error.
We can remove one of the branches in the `TimestampError` enum, and
use `From<Timestamp> for String` instead of `TryFrom`.

Fixes #2177

### DIFF
--- a/src/wkt/src/timestamp.rs
+++ b/src/wkt/src/timestamp.rs
@@ -331,7 +331,7 @@ impl TryFrom<Timestamp> for time::OffsetDateTime {
 }
 
 const EXPECT_OFFSET_DATE_TIME_CONVERTS: &str = concat!(
-    "converting Timestamp to time::OffsetDataTime should always succeed. ",
+    "converting Timestamp to time::OffsetDateTime should always succeed. ",
     "The Timestamp values are always in range. ",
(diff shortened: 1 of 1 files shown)


src/wkt/src/timestamp.rs
```diff
@@ -338,24 +330,34 @@ impl TryFrom<Timestamp> for time::OffsetDateTime {
     }
 }
 
+const EXPECT_OFFSET_DATE_TIME_CONVERTS: &str = concat!(
+    "converting Timestamp to time::OffsetDataTime should always succeed. ",
```
dbolduc, line 334: nit s/Data/Date/

dbolduc (dismissed):

src/wkt/src/timestamp.rs
coryan: Fixed, thanks.

coryan:
PTAL

dbolduc (approved):

PR #2199, opened 2025-05-17
test(showcase): relevant RPCs in showcase `Compliance`

### LINKED ISSUE #2201: showcase: test with all the RPCs for `client::Compliance`

coryan:
I skipped the mixin RPCs (Locations, Operations, and IAM), as these are
tested elsewhere, and against production.

Fixes #2201

### DIFF
--- a/src/integration-tests/src/showcase.rs
+++ b/src/integration-tests/src/showcase.rs
@@ -17,6 +17,7 @@ use crate::Result;
 use gax::options::RequestOptionsBuilder;
 use gax::retry_policy::RetryPolicyExt;
 use showcase::model::compliance_data::LifeKingdom;
+use showcase::model::*;
 use std::process::Stdio;
 use std::time::Duration;
 use tokio::process::Command;
@@ -24,8 +25,6 @@ use tokio::process::Command;
 const SHOWCASE_NAME: &str = "github.com/googleapis/gapic-showcase/cmd/gapic-showcase@v0.36.2";
 
 pub async fn run() -> Result<()> {
-    use showcase::model::*;
-
     #[cfg(feature = "log-integration-tests")]
     let _guard = {
         use tracing_subscriber::fmt::format::FmtSpan;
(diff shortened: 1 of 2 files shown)


src/integration-tests/src/showcase.rs
```diff
@@ -74,7 +87,191 @@ pub async fn run() -> Result<()> {
...
+    request.info = request.info.map(|i| i.set_f_string("first/f-string-value"));
+    let response = client
+        .repeat_data_path_resource()
+        .with_request(request.clone())
+        .send()
+        .await?;
+    tracing::info!("response: {response:?}");
+    assert_eq!(
+        response.binding_uri,
+        "/v1beta1/repeat/{info.f_string=first/*}/{info.f_child.f_string=second/*}/bool/{info.f_bool}:pathresource"
```
dbolduc, line 168: (for my understanding): we are sending a request to some path where all the substitutions were made, and then gapic-showcase is telling us what the generic path template looked like?

So if, let's say, the library picked the wrong path for this request, the assertion would fail.

src/integration-tests/src/showcase.rs
```diff
@@ -74,7 +87,191 @@ pub async fn run() -> Result<()> {
...
+        .set_f_bool(true)
+        .set_f_continent(Continent::Europe)
+        .set_f_child(
+            grandchild
+                .clone()
+                .set_f_string("grandchild-in-f-child-field"),
+        )
+        .set_p_string(concat!(
+            "Benjamín pidió una bebida de kiwi y fresa. ",
+            "Noé, sin vergüenza, la más exquisita champaña del menú"
```
dbolduc, line 262: :champagne:

dbolduc (approved):

src/integration-tests/src/showcase.rs
coryan: Yes. Presumably at some point we would test different bindings using calls with different variables set the same variables matching different templates.

Issue #2038, opened 2025-05-08
Run an endurance test for a generated client


coryan:
We would like to run a test for multiple days, performing millions (ideally at least 1 billion) requests.  The idea is to shake out any memory leaks, or race conditions, or terrible performance, or bad caching for auth, or any other number of things that are only found when running an application for a long time.

My suggestion would be to use secret manager for this service.  The [quota](https://cloud.google.com/secret-manager/quotas) for access requests is sufficiently high.  To make the test interesting we should:

- Use the same client (suitably cloned) from multiple threads.
- Each thread should create its own secret.
- Each thread should mostly perform access requests to this secret trying to hit about 90,000 / N requests per minute.
- From time to time the thread should update the secret.

We should deploy this program to GKE where we get monitoring metrics more easily, and then monitor the process CPU and memory usage.

We want at least 1B successful requests.

coryan:
Completed 1B requests without errors, without crashes, without memory leaks, and with flat CPU usage:

![Image](https://github.com/user-attachments/assets/7c24fbb4-0450-4848-a179-824f9e43034c)

PR #2210, opened 2025-05-18
test(endurate-test): new long-running test

coryan:
Introduce a long running test. We want to minimize the chances of race
conditions, resource leaks, bad caching of access tokens, or other
problems that may go undetected in short-lived unit or integration
tests.

Part of the work for #2038


src/endurance-test/README.md
```diff
@@ -0,0 +1,96 @@
...
+cd google-cloud-cpp
+cargo run --release -p endurance-test
+```
+
+That should print some progress metrics every 10 seconds or so. If it fails to
+start or cannot successfully update and access the secrets, then check the
+permissions.
+
+Once it is working we want to run it in the background, with the logs going to
+Cloud Logging. First, create a service unit file:
```
dbolduc, line 78: TIL

src/endurance-test/src/main.rs
```diff
@@ -0,0 +1,232 @@
...
+                        success_count,
+                        error_count,
+                        total_workers
+                    ),
+                    "task[0]",
+                );
+            }
+            last_report = Instant::now();
+            error_count = 0;
+            success_count = 0;
```
dbolduc, line 81: Need to reset `update_count` too

src/endurance-test/src/main.rs
```diff
@@ -0,0 +1,230 @@
...
+        }
+        // We want to average about 30,000 requests per minute. That will keep the
+        // test well below the quota:
+        //   https://cloud.google.com/secret-manager/quotas
+        let wait = tokio::time::Instant::now()
+            + Duration::from_secs(60)
+                .mul_f32(total_workers as f32)
+                .div_f32(90_000_f32);
```
dbolduc, line 89: Why is this 3x the rate cited in the comment

src/endurance-test/src/main.rs
```diff
@@ -0,0 +1,230 @@
...
+            .send()
+            .await;
+        match access {
+            Ok(_) => {
+                success_count += 1;
+            }
+            Err(e) => {
+                error_count += 1;
+                if error_count > TOO_MANY_ERRORS && success_count == 0 {
+                    return Err(e.into());
```
dbolduc, line 103: These are double counting

dbolduc (changes requested):
Looks good, but I dont trust some of the numbers

src/endurance-test/src/main.rs
dbolduc: Looks like you also caught this, and my comment is obsolete

src/endurance-test/README.md
```diff
@@ -0,0 +1,96 @@
...
+```shell
+sudo apt install gcc rustup git vim
+rustup update stable
+```
+
+Clone the code and run a test:
+
+```shell
+git clone https://github.com/googleapis/google-cloud-cpp.git
+cd google-cloud-cpp
```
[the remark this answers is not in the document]
coryan: I guess my fingers are still used to typing `-cpp`.

src/endurance-test/src/main.rs
coryan: Thanks.

src/endurance-test/src/main.rs
coryan: Did not update the comments on numbers after testing.

dbolduc (approved):

Issue #2079, opened 2025-05-09
URL Sourced Credentials


sai-sunder-s:
Support Workload Identity Federation through URL Sourced Credentials.

PR #2218, opened 2025-05-19
fix(auth): allow token exchange without auth info

alvarowolfx:
Token exchange calls doesn't require authentication. From [RFC8693 Request section](https://datatracker.ietf.org/doc/html/rfc8693#Request):

```
The supported methods of client authentication and whether or not to allow unauthenticated or unidentified clients are deployment decisions that are at the discretion of the authorization server. [...] Thus, client authentication allows for additional authorization checks by the STS as to which entities are permitted to impersonate or receive delegations from other entities.
```

-  Example on GCP auth documentation page shows example without any auth header: https://cloud.google.com/iam/docs/workforce-obtaining-short-lived-credentials#use_the_rest_api

### DIFF
--- a/src/auth/src/credentials/internal/sts_exchange.rs
+++ b/src/auth/src/credentials/internal/sts_exchange.rs
@@ -138,12 +138,8 @@ impl ClientAuthentication {
             if let Ok(value) = header {
                 headers.insert("Authorization", value);
             }
-            return Ok(());
         }
-        Err(crate::errors::CredentialsError::from_str(
-            false,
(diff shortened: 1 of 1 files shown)


sai-sunder-s:
Lets test this with the byoid setup and then merge.

alvarowolfx:
> Lets test this with the byoid setup and then merge.

this is actually kind of blocking that test on #2079 , as the default BYOID setup doesn't require client_id/client_secret

sai-sunder-s (approved):

PR #2228, opened 2025-05-20
impl(generator/rust): support for LRO over gRPC

dbolduc:
Most of the remaining work for #2037.

Support LRO conversion in the generator. This will stay up to date if new LROs are added to the storage control service.

Reminder that we have unit tests in `convert.rs` and an integration test for the rename folder LRO.

I haven't decided if I want to move the `ToProto` / `FromProto` impls into transport. We definitely could.

Note that the mustache template for each `LROType` could be used for each `ErrorInfo` in a `rpc::Status`.

### DIFF
--- a/src/storage-control/src/convert.rs
+++ b/src/storage-control/src/convert.rs
@@ -27,8 +27,10 @@
 
 use crate::google;
 use gaxi::prost::{ConvertError, FromProto, ToProto};
-// This would fail if storage admin introduces LROs. We are ok with it, because
-// we know the service will not grow.
+// This would fail if storage admin introduces LROs. This is unlikely as 
+// one of the motivations for `StorageControl` was to separate the LROs.
+// Worst case, we can skip the offending functions until we have fixed
+// this code to support them.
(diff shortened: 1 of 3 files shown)


src/storage-control/src/convert.rs
```diff
@@ -27,6 +27,9 @@
 
 use crate::google;
 use gaxi::prost::{ConvertError, FromProto, ToProto};
+// This would fail if storage admin introduces LROs. We are ok with it, because
+// we know the service will not grow.
```
coryan, line 31:
```suggestion
// This would fail if storage admin introduces LROs. This is unlikely as 
// one of the motivations for `StorageControl` was to separate the LROs.
// Worst case, we can skip the offending functions until we have fixed
// this code to support them.
```

coryan (dismissed):

src/storage-control/src/convert.rs
dbolduc: Done.

coryan (approved):

PR #2230, opened 2025-05-20
impl(storage): add checksum fields to v1::Object

suzmue:
For #2039

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -315,6 +315,8 @@ pub(crate) mod info {
 }
 
 mod v1 {
+    use base64::Engine as _;
+
     #[serde_with::serde_as]
     #[derive(Debug, Default, serde::Deserialize, PartialEq, Clone)]
     #[serde(default, rename_all = "camelCase")]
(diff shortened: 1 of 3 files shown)


src/storage/src/client.rs
```diff
@@ -552,6 +618,9 @@ mod v1 {
                     encryption_algorithm: "algorithm".to_string(),
                     key_sha256: bytes::Bytes::from("the quick brown fox jumps over the lazy dog"),
                 }),
+                md5_hash: bytes::Bytes::from("the quick brown fox jumps over the lazy dog"),
+                // base64 encoded uint32 in BigEndian order field:
+                crc32c: Some(1234567890),
```
coryan, line 623: I think the test is fine as-is, but if you are interested in realistic values try:

```
echo 'The quick brown fox jumps over the lazy dog' > quick.txt

gcloud storage hash quick.txt 
---
crc32c_hash: /ieOcg==
digest_format: base64
md5_hash: N8S4ft/8XRmP9aGFzufuCQ==
url: quick.txt
```

coryan (approved):

src/storage/src/client.rs
suzmue: Nice. updated the deserialization test to use these values.

coryan (approved):

PR #2233, opened 2025-05-20
refactor: absolute builder links

dbolduc:
Part of the work for #1813, #2232 

Use absolute links to a client instead of relative links from the request builder.

This is enough to unblock the builder docs for storage control, without too much special handling.

I suspect more complicated veneers will have handwritten builders that are not exposed in the client surface. We will cross that bridge when we get to it.

### DIFF
--- a/src/generated/api/apikeys/v2/src/builder.rs
+++ b/src/generated/api/apikeys/v2/src/builder.rs
@@ -17,7 +17,7 @@
 pub mod api_keys {
     use crate::Result;
 
-    /// A builder for [ApiKeys][super::super::client::ApiKeys].
+    /// A builder for [ApiKeys][crate::client::ApiKeys].
     ///
     /// ```
     /// # tokio_test::block_on(async {
@@ -45,7 +45,7 @@ pub mod api_keys {
         }
     }
 
-    /// Common implementation for [super::super::client::ApiKeys] request builders.
+    /// Common implementation for [crate::client::ApiKeys] request builders.
     #[derive(Clone, Debug)]
     pub(crate) struct RequestBuilder<R: std::default::Default> {
         stub: std::sync::Arc<dyn super::super::stub::dynamic::ApiKeys>,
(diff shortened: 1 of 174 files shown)


src/storage-control/src/generated/gapic_control/builder.rs
```diff
@@ -17,7 +17,7 @@
 pub mod storage_control {
     use crate::Result;
 
-    /// Common implementation for [super::super::client::StorageControl] request builders.
+    /// Common implementation for [crate::client::StorageControl] request builders.
```
coryan, line 20: This is not the name of the client. It is called `crate::client::Storage`:

https://github.com/googleapis/google-cloud-rust/blob/b753d14da427d86c9603982c9cd21792381e8e82/src/storage-control/src/client.rs#L83

And if we rename the client, then the changes in src/storage-control/src/generated/gapic/builder.rs would be wrong.

You need a more sophisticated change. Maybe an option setting the client path?

coryan (changes requested):

src/storage-control/src/generated/gapic_control/builder.rs
dbolduc: This is where we're going:

https://github.com/dbolduc/google-cloud-rust/tree/fix-veneer-builder-docs

My plan is to add an option that lets you rename a service. Then we just need to rename `StorageControl` -> `Storage`, and we are good.

dbolduc: Opened #2238 from that branch (because I am sure I will delete the branch at some point). I will leave it as a draft and rebase if we go with this approach.

WDYT?

src/generated/api/apikeys/v2/src/builder.rs
```diff
@@ -17,7 +17,7 @@
 pub mod api_keys {
     use crate::Result;
 
-    /// A builder for [ApiKeys][super::super::client::ApiKeys].
+    /// A builder for [ApiKeys][crate::client::ApiKeys].
```
coryan, line 20: I wish we could avoid all this churn, but I don't see a good way.

Issue #2237, opened 2025-05-20
Seal the `*Policy` traits


coryan:
We may need to change these in the near future. If we can seal them and remove the methods from the documentation we can change them safely and then open them up later.

PR #2241, opened 2025-05-20
docs(gax)!: hide policy functions

coryan:
Temporarily we are excluding the retry policy, backoff policy, polling
error policy, and polling backoff policy functions from the public
APIs. That will make it easier to evolve these APIs as we finalize the
design for the client libraries.

Part of the work for #2237

### DIFF
--- a/src/gax/src/polling_error_policy.rs
+++ b/src/gax/src/polling_error_policy.rs
@@ -59,6 +59,7 @@ pub trait PollingErrorPolicy: Send + Sync + std::fmt::Debug {
     ///   attempt. This method called after LRO successfully starts, it is
     ///   always non-zero.
     /// * `error` - the last error when attempting the request.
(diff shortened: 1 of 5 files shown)


src/gax/src/polling_error_policy.rs
```diff
@@ -59,6 +59,7 @@ pub trait PollingErrorPolicy: Send + Sync + std::fmt::Debug {
     ///   attempt. This method called after LRO successfully starts, it is
     ///   always non-zero.
     /// * `error` - the last error when attempting the request.
+    #[cfg_attr(not(feature = "_internal-semver"), doc(hidden))]
```
dbolduc, line 62: comment: Hm, I was thinking that if we don't use these downstream, semver should be unaffected.

But I know the polling policy, at least, is exposed in stubs with LROs.

dbolduc (approved):

Issue #439, opened 2024-12-16
Automatically populate fields in the request message (google.aip.dev/4235)


coryan:
Some Google APIs use request id fields (maybe under a different name). These fields should be automatically populated by the client library, and when set, they change the retry loops: the RPCs become idempotent.

PR #2248, opened 2025-05-20
feat(storage-control): complete the API surface

### LINKED ISSUE #944: Generate library for `google/storage/control/v2/storage_v2.yaml`

dbolduc:
Part of the work for #1813 

Fill out the storage control v2 API surface. The client signatures are copied exactly from `gapic_control/client.rs`

(The storage v2 API surface is not complete. I know we are still missing at least `update_bucket`. I plan to address that in a different PR).

We had been writing handwritten examples involving the RPCs. We do not plan to handwrite examples in the long run, so I am not going to do it now.

We have also been using the APIs in handwritten integration tests. I think we have a critical mass of integration tests. If we are missing anything, it is an integration test for an LRO that does not return immediately. (Our `RenameFolder` call returns `done` without needing to poll.)

---

Fixes #944 

The client cannot do auto-population of request IDs, but that is a nice to have not something necessary. Let's track that work in #439, and not block #944 on it.

### DIFF
--- a/src/storage-control/src/client.rs
+++ b/src/storage-control/src/client.rs
@@ -321,6 +321,26 @@ impl StorageControl {
         self.control.create_folder()
     }
 
+    /// Permanently deletes an empty folder.
+    ///
+    /// This operation is only applicable to a hierarchical namespace enabled
+    /// bucket.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_storage_control::client::StorageControl;
+    /// async fn example(client: &StorageControl) -> gax::Result<()> {
+    ///     client.delete_folder()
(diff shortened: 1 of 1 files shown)


coryan (approved):
The lack of examples makes me sad. I know why you are going this way, but I think the GAPICs should be gaining examples too. This was a good way to learn what it took to create them.

dbolduc:
> the GAPICs should be gaining examples

when this is a thing, we can copy the generated examples over to the handwritten surface.

Issue #2256, opened 2025-05-22
Conflicting message vs. oneof names in generated code


coryan:
With the next update of the googleapis SHA we will run into errors in the generated code.

Both Prost (see https://github.com/tokio-rs/prost/issues/505) and sidekick generate conflicting types for this:

```proto
  message Filter {
    message CloudStorageBuckets {
       // ...
    }
    oneof cloud_storage_buckets {
        // ...
    }
  }
```

https://github.com/googleapis/googleapis/blob/488c8cc0d41aca4b4f33748ab6166129edcfd836/google/storage/control/v2/storage_control.proto#L1112-L1126

PR #2259, opened 2025-05-23
feat(generator): support post-processing for Prost

coryan:
Part of the work for #2256


src/storage-control/src/generated/protos/control/build.rs
```diff
@@ -32,5 +32,32 @@ fn main() {
...
+        let content = content.replace(
+            "prost(oneof = \"filter::CloudStorageBuckets\"",
+            "prost(oneof = \"filter::CloudStorageBucketsOneOf\"",
+        );
+        let content = content.replace(
+            "pub cloud_storage_buckets: ::core::option::Option<filter::CloudStorageBuckets>",
+            "pub cloud_storage_buckets: ::core::option::Option<filter::CloudStorageBucketsOneOf>",
+        );
+        std::fs::write(name, content.as_bytes())
+            .expect(format!("should be able to write '{name}'").as_str());
```
dbolduc, line 61: shouldn't we see changes to `google.storage.control.v2.rs` in this PR?

dbolduc: D'oh, this is before we have updated googleapis.

src/storage-control/src/generated/protos/control/build.rs
```diff
@@ -32,5 +32,32 @@ fn main() {
         config.type_name_domain(&["."], "type.googleapis.com");
         config.out_dir(".");
         config.compile_protos(files, includes).expect("error compiling protos");
+        let name = "google.storage.control.v2.rs";
```
dbolduc, line 35: Comment: Maybe we should try to patch the proto, not the prost code

dbolduc (approved):
Ship it. I just left some thoughts

Issue #2103, opened 2025-05-12
Use builder for `Storage::read_object`


coryan:
The function should return a request builder  that saves all the parameters from:

https://github.com/googleapis/google-cloud-rust/blob/12789de56a016d3a6b6e93e69dcb7b5e5b073880/src/storage-control/src/generated/gapic/model.rs#L1140

Then we need to map each parameter that is set to the corresponding query parameter in the JSON request or reject the request as "unimplemented", if no mapping is possible.

PR #2263, opened 2025-05-23
fix(storage)!: update read_object to return a builder

suzmue:
Create a new ReadObject builder in Storage. The send() function is the same as read_object was previously, but instead pulls the bucket and object values from the ReadObjectRequest.

This is the first step to allow for additional parameters to be set for the read object request. No new functionality is really being added in this change.

For #2103


src/storage/src/client.rs
```diff
@@ -313,6 +277,200 @@ pub(crate) mod info {
     }
 }
 
+pub struct ReadObject<'a> {
+    client: &'a Storage,
+    request: control::model::ReadObjectRequest,
```
coryan, line 282: In a future PR: we may want a way to make `Storage` mockable, and doing so may require introducing some kind of stub at this layer.

src/storage/src/client.rs
```diff
@@ -313,6 +277,200 @@ pub(crate) mod info {
...
+    pub fn set_object<T: std::convert::Into<std::string::String>>(mut self, v: T) -> Self {
+        self.request.object = v.into();
+        self
+    }
+
+    /// Sets the value of [generation][control::model::ReadObjectRequest::generation].
+    pub fn set_generation<T: std::convert::Into<i64>>(mut self, v: T) -> Self {
+        self.request.generation = v.into();
+        self
+    }
```
coryan, line 309: We need to start fleshing out the tests.  In generated code we do not test these functions. If am unsure because these functions are trivial (which recommends "no tests") and at the same time they are hand-crafted (which recommends "write a unit test").

Thoughts?

src/storage/src/client.rs
```diff
@@ -313,6 +277,200 @@ pub(crate) mod info {
...
+
+    /// Sets the value of [read_offset][control::model::ReadObjectRequest::read_offset].
```
coryan, line 311: Maybe we should say something more useful in the documentation of these?

src/storage/src/client.rs
```diff
@@ -313,6 +277,200 @@ pub(crate) mod info {
...
+    pub fn set_if_metageneration_match<T>(mut self, v: T) -> Self
+    where
+        T: std::convert::Into<i64>,
+    {
+        self.request.if_metageneration_match = std::option::Option::Some(v.into());
+        self
+    }
+
+    /// Sets or clears the value of [if_metageneration_match][control::model::ReadObjectRequest::if_metageneration_match].
+    pub fn set_or_clear_if_metageneration_match<T>(mut self, v: std::option::Option<T>) -> Self
```
coryan, line 369: In generated code we use `std::option::Option` and `std::convert::Into` to avoid conflicts, but that is not idiomatic. We should use `Option<>` and `Into<>` in hand-written code.

coryan (approved):
I would be happy to merge this as-is and improve things over time. I left you some questions, if you want to take action on some of them that is fine too.

src/storage/src/client.rs
suzmue: Fixed in #2272

src/storage/src/client.rs
suzmue: Wrote unit tests in #2279 and after writing them, they feel mostly like clutter, especially since these are copy pasted over with 0 changes except in the doc comments.

More useful tests would make sure we set each of these and then see if they have the desired effect.

coryan: > Wrote unit tests in #2279 and after writing them, they feel mostly like clutter, especially since these are copy pasted over with 0 changes except in the doc comments.

Feel free to close #2279 if that is the case.

> More useful tests would make sure we set each of these and then see if they have the desired effect.

By "desired effect" I think you mean "the right request message + query parameters + headers gets built / sent to GCS"? You could refactor the code to prepare the `reqwest::RequestBuilder` and then (I think?) you can examine the values.  That does seem more useful.

Issue #2221, opened 2025-05-19
Reconsider design of `gax::Error`


coryan:
We need to meet two (somewhat conflicting) requirements:

- We need to expose full error details, particularly when they arrive from the service
- Some services return incomplete or poorly formatted errors, we need to return whatever information we get from HTTP in this case.
- Some errors happen before any HTTP request, we need to know where.
- Some errors happen after a *successful* HTTP request (deserialization)
- Some errors are unrelated to HTTP requests (loading credentials).

We would like to capture all these into a single error type. We can use a `pub enum` with variant for each case. We can use the `reqwest::Error` style, with a `is_foo(&self)` predicate for each kind, we can use something more akin to `std::io::Error`.

We need to support errors created from outside the crate (constructors in `internal`?)

Applications and retry policies may be interested in the details of the error condition.

It seems like we should flatten `CredentialsError` into `Error` with two new "kinds" representing:

- "failed to create the credentials" (with details)
- "failed to fetch the token" (with details)

PR #2268, opened 2025-05-25
fix(gax): correct `source()` in `CredentialsError`

coryan:
This was returning a reference to `Arc<T>`, but we really wanted a
reference to `T`, the original error source.

Found while fixing #2221, but otherwise unrelated.


dbolduc (approved):
Thanks

PR #2293, opened 2025-05-28
cleanup(auth): scope is optional on TokenResponse

alvarowolfx:
Also removed `RefreshAccessTokenRequest` as it's unused and leftover from previous PR. 

Towards https://github.com/googleapis/google-cloud-rust/issues/2079

### DIFF
--- a/src/auth/src/credentials/internal/sts_exchange.rs
+++ b/src/auth/src/credentials/internal/sts_exchange.rs
@@ -112,7 +112,7 @@ pub struct TokenResponse {
     pub issued_token_type: String,
     pub token_type: String,
     pub expires_in: u64,
-    pub scope: String,
+    pub scope: Option<String>,
     pub refresh_token: Option<String>,
 }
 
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/internal/sts_exchange.rs
```diff
@@ -257,7 +248,7 @@ mod test {
                 issued_token_type: ACCESS_TOKEN_TYPE.to_string(),
                 token_type: "Bearer".to_string(),
                 expires_in: 3600,
-                scope: "https://www.googleapis.com/auth/cloud-platform".to_string(),
+                scope: Some("https://www.googleapis.com/auth/cloud-platform".to_string()),
```
coryan, line 251: Should we use the constant here?

alvarowolfx: oh, just found that is declared in `src/auth/src/credentials/service_account.rs`

sai-sunder-s: Maybe we can define this in a common place. Not sure what is the usual way to do in Rust.

How about creating a `Constants.rs` mod under `src/auth`?

alvarowolfx: will merge this and send a PR creating the constants file to not mix with the main goal of this PR

PR #2296, opened 2025-05-28
cleanup(auth): move default scope to constants module

alvarowolfx:


src/auth/src/lib.rs
```diff
@@ -45,6 +45,9 @@ pub(crate) type BuildResult<T> = std::result::Result<T, build_errors::Error>;
 /// [Credentials]: https://cloud.google.com/docs/authentication#credentials
 pub mod credentials;
 
+/// Auth constants
```
dbolduc, line 48: optional: remove the comment, as the module name is sufficient

dbolduc (dismissed):

src/auth/src/lib.rs
alvarowolfx: good point, done

dbolduc (dismissed):
approved but grr... you have to fix a merge conflict.

sai-sunder-s (dismissed):

coryan (approved):

PR #2300, opened 2025-05-28
feat(storage): support additional query parameters in read_object

suzmue:
For #2103

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -496,36 +496,6 @@ impl ReadObject {
             })?;
         let object: String = self.request.object;
 
-        // Collect the optional query parameters
-        let mut query: Vec<(String, String)> = vec![("alt".into(), "media".into())];
-        if self.request.generation != 0 {
-            query.push(("generation".into(), format!("{}", self.request.generation)))
-        }
-        if let Some(if_generation_match) = self.request.if_generation_match {
-            query.push((
-                "ifGenerationMatch".into(),
-                format!("{}", if_generation_match),
-            ));
-        }
-        if let Some(if_generation_not_match) = self.request.if_generation_not_match {
-            query.push((
-                "ifGenerationNotMatch".into(),
-                format!("{}", if_generation_not_match),
-            ));
-        }
-        if let Some(if_metageneration_match) = self.request.if_metageneration_match {
-            query.push((
-                "ifMetagenerationMatch".into(),
-                format!("{}", if_metageneration_match),
-            ));
-        }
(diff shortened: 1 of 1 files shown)


src/storage/src/client.rs
```diff
@@ -489,6 +493,35 @@ impl ReadObject {
...
         let object: String = self.request.object;
+
+        // Collect the optional query parameters
+        let mut query: Vec<(String, String)> = vec![("alt".into(), "media".into())];
+        if let Some(if_generation_match) = self.request.if_generation_match {
+            query.push((
+                "ifGenerationMatch".into(),
+                format!("{}", if_generation_match),
+            ));
+        }
```
coryan, line 504: Consider moving these  after the initial `let builder = ...` call and do:

```suggestion
        let builder = self.request.if_generation_match
            .iter().fold(builder, |b, v| b.query([("ifGenerationMatch", v)]);
```

Otherwise we do a bunch of allocations for the vector and the strings.  You may need to do a `format!("{}")`, but in principle anything that can be serialized can be used as an argument to `build.query(...)`.

coryan (changes requested):

src/storage/src/client.rs
suzmue: Done. Thanks for the suggestion!

src/storage/src/client.rs
```diff
@@ -554,6 +590,34 @@ mod tests {
...
+            .set_if_metageneration_not_match(40)
+            .http_request_builder()
+            .await?
+            .build()?;
+
+        assert_eq!(read_object_builder.method(), reqwest::Method::GET);
+        assert_eq!(
+            read_object_builder.url().as_str(),
+            "https://storage.googleapis.com/storage/v1/b/bucket/o/object?alt=media&generation=5&ifGenerationMatch=10&ifGenerationNotMatch=20&ifMetagenerationMatch=30&ifMetagenerationNotMatch=40"
+        );
```
coryan, line 618: nit: I do not love that this depends on the order of the arguments and that it is a long string.  I think you can rewrite this as:

```suggestion
            concat!("https://storage.googleapis.com/storage/v1/b/bucket/o/object?",
          "alt=media&",
          "generation=5&",
          "ifGenerationMatch=10&",
          "ifGenerationNotMatch=20&",
          "ifMetagenerationMatch=30&",
          "ifMetagenerationNotMatch=40")
        );
```

In the future, consider something like: https://docs.rs/reqwest/latest/reqwest/struct.Url.html#method.query_pairs to compare the query parameters as (sorted) vectors or something.

coryan (approved):

Issue #652, opened 2025-01-09
Support GOOGLE_CLOUD_QUOTA_PROJECT


sai-sunder-s:
As described in the **Post Processing** step of https://google.aip.dev/auth/4110, GOOGLE_CLOUD_QUOTA_PROJECT env var should be checked to get the quota project to use for the credential. This is applicable to all credential types.

PR #2304, opened 2025-05-29
feat(auth): Enable quota project setting via env variable

harkamaljot:
This change allows the users to specify quota project via env variable `GOOGLE_CLOUD_QUOTA_PROJECT`.

Currently, we prioritize the value provided in env variable rather than value provided via builder methods. Verified we only want to support env variable setting for access token credentials.

Fixes: https://github.com/googleapis/google-cloud-rust/issues/652


src/auth/src/credentials.rs
```diff
@@ -450,7 +452,10 @@ impl Builder {
                 AdcContents::FallbackToMds => None,
             },
         };
-        build_credentials(json_data, self.quota_project_id, self.scopes)
+        let quota_project_id = self
+            .quota_project_id
+            .or_else(|| std::env::var(GOOGLE_CLOUD_QUOTA_PROJECT_VAR).ok());
```
sai-sunder-s, line 457: In Rust we prioritize env var over value provided through code. This is a deviation from the design doc and AIP but it is idiomatic with the Env Var usage in rest of the library.

harkamaljot: I feel we should do a log in `with_quota_project` method if env variable is set to let the user know as otherwise it will be confusing the end user. Similarly for other places as well like in mds.rs where we prioritize env var over value provided through code. What do you think?

coryan (dismissed):

sai-sunder-s (approved):

PR #2308, opened 2025-05-29
test(storage): add test for read_object with error credentials

suzmue:
For #2103

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -575,6 +575,24 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_read_object_error_credentials() -> Result {
+        let client = Storage::builder()
+            .with_endpoint("http://private.googleapis.com")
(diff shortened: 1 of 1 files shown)


src/storage/src/client.rs
```diff
@@ -574,6 +574,24 @@ mod tests {
...
+            .build()
+            .await?;
+
+        client
+            .read_object()
+            .set_bucket("projects/_/buckets/bucket")
+            .set_object("object")
+            .http_request_builder()
+            .await
+            .expect_err("invalid credentials should err");
```
dbolduc, line 591: I think we should confirm the type of error

https://github.com/googleapis/google-cloud-rust/blob/827a0b3f336edeb1cb5bf2adf94a9abeacf0c484/src/gax-internal/tests/http_auth.rs#L149-L154

suzmue: Done. Used is_authentication() and inspect_err

PR #2311, opened 2025-05-29
refactor(gax): simplify `Error` implementation

coryan:
Most of our error kinds have a source, so we can refactor that to an
`Option<BoxError>`. Also refactor the details for transport and service
errors to (private) structs, so they are easier to work with.

These changes will make it easier to add more specific errors for
deserialization, unimplemented functions, and other branches we need to
handle.

Motivated by #2312

### DIFF
--- a/src/gax/src/error/core_error.rs
+++ b/src/gax/src/error/core_error.rs
@@ -55,6 +55,7 @@ type BoxError = Box<dyn std::error::Error + Send + Sync>;
 #[derive(Debug)]
 pub struct Error {
     kind: ErrorKind,
+    source: Option<BoxError>,
 }
 
 impl Error {
@@ -69,12 +70,15 @@ impl Error {
     /// assert_eq!(error.status(), Some(&status));
     /// ```
     pub fn service(status: Status) -> Self {
-        let kind = ErrorKind::Service {
+        let details = ServiceDetails {
             status,
             status_code: None,
             headers: None,
         };
-        Self { kind }
+        Self {
+            kind: ErrorKind::Service(Box::new(details)),
+            source: None,
+        }
     }
 
     /// Creates an error representing a timeout.
@@ -89,7 +93,8 @@ impl Error {
     /// ```
     pub fn timeout<T: Into<BoxError>>(source: T) -> Self {
         Self {
-            kind: ErrorKind::Timeout(source.into()),
+            kind: ErrorKind::Timeout,
+            source: Some(source.into()),
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/gax/src/error/core_error.rs
```diff
@@ -291,6 +289,7 @@ impl Error {
     pub fn authentication(source: CredentialsError) -> Self {
         Self {
             kind: ErrorKind::Authentication(source),
+            source: None,
```
dbolduc, line 292: This looks strange. Seems like `ErrorKind::Authentication` should be a unit variant and we put the credential error into `source`.

src/gax/src/error/core_error.rs
```diff
@@ -409,81 +412,98 @@ impl Error {
...
+        match (&self.kind, &self.source) {
+            (ErrorKind::Binding, Some(e)) => {
+                write!(f, "cannot find a matching binding to send the request {e}")
             }
-            ErrorKind::Serialization(e) => write!(f, "cannot serialize the request: {e}"),
-            ErrorKind::Authentication(e) => {
-                write!(f, "cannot create the authentication headers: {e}")
+            (ErrorKind::Serialization, Some(e)) => write!(f, "cannot serialize the request {e}"),
+            (ErrorKind::Authentication(e), _) => {
+                write!(f, "cannot create the authentication headers {e}")
```
dbolduc, line 421: and here

src/gax/src/error/core_error.rs
```diff
@@ -409,81 +412,98 @@ impl Error {
...
-            ErrorKind::Serialization(e) => Some(e.as_ref()),
-            ErrorKind::Authentication(e) => Some(e),
-            ErrorKind::Timeout(e) => Some(e.as_ref()),
-            ErrorKind::Transport { source, .. } => source
-                .as_ref()
-                .map(|e| e.as_ref() as &(dyn std::error::Error)),
-            ErrorKind::Service { .. } => None,
-            ErrorKind::Other(e) => Some(e.as_ref()),
+        if let ErrorKind::Authentication(e) = &self.kind {
+            return Some(e);
```
dbolduc, line 445: and here.

src/gax/src/error/core_error.rs
coryan: Eventually we may want to expose the `is_transient()` attribute of `CredentialsError`, but we can change things back when that happens.

dbolduc: Ah, I see. We could do like:

```
enum AuthErrorKind {
  Transient,
  Permanent,
}
```

```
Authentication(AuthRetryLoopStatus)
```

to simplify our internals a tad.

PR #2314, opened 2025-05-29
ci(auth): deflake token cache thundering herd tests

### LINKED ISSUE #2309: Flake in auth token cache test
https://github.com/googleapis/google-cloud-rust/actions/runs/15314959492/job/43086974209#step:8:1

```
thread 'token_cache::test::no_initial_token_thundering_herd_success' panicked at src\auth\src\token_cache.rs:658:9:
assertion `left == right` failed
  left: 2
 right: 1


failures:
    token_cache::test::no_initial_token_thundering_herd_success
```

---

I think that less than 100 of the tasks started...

https://github.com/googleapis/google-cloud-rust/blob/827a0b3f336edeb1cb5bf2adf94a9abeacf0c484/src/auth/src/token_cache.rs#L642

Within the 50ms we give for them to pile up...

https://github.com/googleapis/google-cloud-rust/blob/827a0b3f336edeb1cb5bf2adf94a9abeacf0c484/src/auth/src/token_cache.rs#L618

And the token we give back is expired...

https://github.com/googleapis/google-cloud-rust/blob/827a0b3f336edeb1cb5bf2adf94a9abeacf0c484/src/auth/src/token_cache.rs#L633

So the cache makes another call to the fake provider.

---

To deflake, I would assert that we make less than 100 calls to the inner token provider. 🤷

https://github.com/googleapis/google-cloud-rust/pull/1554#discussion_r2006006442

dbolduc:
Fixes #2309 

Relax expectations around thundering herds.

We would get a false positive if 100 tasks could not launch within 50ms. Now we will get a false positive if 100 tasks cannot launch within 500ms.

### DIFF
--- a/src/auth/src/token_cache.rs
+++ b/src/auth/src/token_cache.rs
@@ -614,7 +614,8 @@ mod test {
     #[async_trait::async_trait]
     impl TokenProvider for FakeTokenProvider {
         async fn token(&self) -> Result<Token> {
-            // We give enough time for the a thundering herd to pile up waiting for a change notification from watch channel
+            // We give enough time for a thundering herd to pile up, while
+            // waiting for a change notification from the watch channel.
             sleep(Duration::from_millis(50)).await;
 
             // Track how many calls were made to the inner token provider.
@@ -648,19 +649,20 @@ mod test {
 
         // Wait for the N token requests to complete, verifying the returned token.
         for task in tasks {
-            let actual = task.await.unwrap();
-            assert!(actual.is_ok(), "{}", actual.err().unwrap());
-            assert_eq!(get_cached_token(actual.unwrap())?, token);
+            let actual = task.await?;
+            assert!(actual.is_ok(), "{}", actual.unwrap_err());
+            assert_eq!(get_cached_token(actual?)?, token);
         }
 
         let calls = tp.calls();
-        // Only one call to token provider should have been made
-        assert_eq!(calls, 1);
+        // We expect one call to be made to the inner token provider. But if the
+        // 100 tasks take longer than 50ms to launch, we may see multiple.
+        assert!(calls < 10, "calls to inner token provider: {calls}");
         Ok(())
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 8)]
-    async fn no_initial_token_thundering_herd_failure_shares_error() {
+    async fn no_initial_token_thundering_herd_failure_shares_error() -> TestResult {
         let err = Err(errors::non_retryable_from_str("epic fail"));
 
         let tp = FakeTokenProvider::new(err);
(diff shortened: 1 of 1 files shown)


src/auth/src/token_cache.rs
```diff
@@ -614,7 +614,8 @@ mod test {
     #[async_trait::async_trait]
     impl TokenProvider for FakeTokenProvider {
         async fn token(&self) -> Result<Token> {
-            // We give enough time for the a thundering herd to pile up waiting for a change notification from watch channel
+            // We give enough time for a thundering herd to pile up, while
+            // waiting for a change notification from the watch channel.
             sleep(Duration::from_millis(50)).await;
```
sai-sunder-s, line 619: description said increasing to 500ms?

src/auth/src/token_cache.rs
```diff
@@ -648,19 +649,20 @@ mod test {
...
+            assert!(actual.is_ok(), "{}", actual.unwrap_err());
+            assert_eq!(get_cached_token(actual?)?, token);
         }
 
         let calls = tp.calls();
-        // Only one call to token provider should have been made
-        assert_eq!(calls, 1);
+        // We expect one call to be made to the inner token provider. But if the
+        // 100 tasks take longer than 50ms to launch, we may see multiple.
+        assert!(calls < 10, "calls to inner token provider: {calls}");
```
sai-sunder-s, line 660: Why will there be more calls to inner token provider if tasks are delayed to start? Number of tasks and time of call shouldnt matter because inner token provider is called by the transmitter task?

dbolduc: > Why will there be more calls to inner token provider if tasks are delayed to start?

The delay is not enough to launch all of the tasks. The fake token provider sleeps whenever it gets the first request. I could not figure out how to make it wait until tokens were requested from all N tasks.

See the issue write up, and let me know what you disagree with: #2309 

> is called by the transmitter task?

I don't know what this is.

dbolduc: > Hmmm... You could request the token (without the `.await`), then signal a barrier, then call `.await`.

Argh, I never considered delaying the `.await`. That might work.

src/auth/src/token_cache.rs
```diff
@@ -677,14 +679,16 @@ mod test {
...
+            let e = format!("{}", actual.unwrap_err());
             assert!(e.contains("epic fail"), "{e}");
         }
 
         let calls = tp.calls();
-        // Only one call to token provider should have been made
-        assert_eq!(calls, 1);
+        // We expect one call to be made to the inner token provider. But if the
+        // 100 tasks take longer than 50ms to launch, we may see multiple.
+        assert!(calls < 10, "calls to inner token provider: {calls}");
```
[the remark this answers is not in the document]
coryan: It is just that at most 8 tasks can do something "at the same time", but I guess 8 tasks could be active and waiting for Tokio synchronization primitives.  Ignore me.

src/auth/src/token_cache.rs
dbolduc: I am going to merge this PR, because it will decrease flakes. I will spend an hour trying the `Barrier` approach.

dbolduc: I don't think it works.

If we have the N tasks wait at a barrier, they could not have asked for a token. We have the same problem. We don't know when all N tasks are waiting on a token/error from the channel. (and thus we should release a token from our fake provider).

I remember using a plain old condition variable (effectively as a barrier) when writing the test and deciding it was not possible to know when all N tasks were awaiting the channel. I decided that trying to almost synchronize the tasks just made the tests harder to read. And I settled on the current form of the test.

(we need to sleep in the fake provider, no matter what. so let's do that and nothing else complicated).

PR #2322, opened 2025-05-30
feat(generator): support overlapping proto roots

coryan:
For the ProtoJSON conformance tests we will need to configure two
`*-root` sources with overlapping contents. The current approach where
we find the code in "any" of the `*-root` sources breaks down in that
case.

With this change a `.sidekick.toml` file can specify which `*-root`
sources to use. I think that is less suprising, and (marginally) faster.

The default configuration is to only search in `googleapis-root`. I
changed a couple of directories to use different sources, which
demonstrates how this works.

Part of the work for #1910


.sidekick.toml
```diff
@@ -17,15 +17,17 @@ language             = 'rust'
...
+roots                       = 'googleapis-root'
+showcase-extracted-name     = 'gapic-showcase-0.36.2'
+showcase-root               = 'https://github.com/googleapis/gapic-showcase/archive/refs/tags/v0.36.2.tar.gz'
+showcase-sha256             = '0914bdbb088713aa087a53b355ff6631ad95e4769afd8bbd97d5d9e78d4cdf09'
 googleapis-root             = 'https://github.com/googleapis/googleapis/archive/488c8cc0d41aca4b4f33748ab6166129edcfd836.tar.gz'
 googleapis-sha256           = '23d2c98f6c09282a804cba07cd1628192b09540ef3fc3f970454cf0fa0022c83'
-protobuf-extracted-name     = 'protobuf-29.3'
-protobuf-root               = 'https://github.com/protocolbuffers/protobuf/releases/download/v29.3/protobuf-29.3.tar.gz'
-protobuf-sha256             = '008a11cc56f9b96679b4c285fd05f46d317d685be3ab524b2a310be0fbad987e'
-protobuf-subdir             = "src"
```
dbolduc, line 28: Q: so are these just markers? Looks like only the `*-root` is used?

coryan (commented):
PTAL

.sidekick.toml
dbolduc: > All the fields are used.

Gotcha, my mistake. I was `grep`ping on `"protobuf-extracted-name"` instead of just the suffix, `"-extracted-name"`

dbolduc (approved):

Issue #2312, opened 2025-05-29
Add functionality to `gax::error::Error`


coryan:
Now that `gax::error::Error` can be grown without breaking customer code we can make a number of improvements:

- [ ] Split the serialization and deserialization error predicates.
- [ ] Add an specific predicate for "unimplemented function" (mostly used in default Stub traits).
- [ ] Remove the `other()` constructor because it leads to bad outcomes.
- [ ] Add a troubleshooting guide with suggestions on what to do for each error type.

PR #2331, opened 2025-05-30
feat: new error kind for deserialization

coryan:
I think we want to distinguish between serialization problems (as these
are most likely caused by the application), and deserialization problems
(as these are most likely problems with the client libraries). They also
have different semantics: the former cannot result in outgoing RPCs,
the latter only happen after an RPC completed.

part of the work for #2312

### DIFF
--- a/src/gax/src/error/core_error.rs
+++ b/src/gax/src/error/core_error.rs
@@ -171,7 +171,7 @@ impl Error {
     /// prefer not to use, GitHub to discuss this problem, then contact
     /// [Google Cloud support].
     ///
-    /// [open an issue]: https://github.com/googleapis/google-cloud-rust/issues
+    /// [open an issue]: https://github.com/googleapis/google-cloud-rust/issues/new/choose
     /// [Google Cloud support]: https://cloud.google.com/support
     pub fn is_deserialization(&self) -> bool {
         matches!(self.kind, ErrorKind::Deserialization)
--- a/src/gax/src/error/core_error.rs
+++ b/src/gax/src/error/core_error.rs
@@ -119,6 +119,65 @@ impl Error {
         matches!(self.kind, ErrorKind::Timeout)
     }
 
+    /// Not part of the public API, subject to change without notice.
+    ///
+    /// Creates an error representing a deserialization problem.
+    ///
+    /// Applications should have no need to use this function. The exception
+    /// could be mocks, but this error is too rare to merit mocks. If you are
+    /// writing a mock that extracts values from [wkt::Any], consider using a
+    /// `.expect()` calls instead.
+    ///
+    ///
+    /// # Example
+    /// ```
+    /// use std::error::Error as _;
+    /// use google_cloud_gax::error::Error;
+    /// let error = Error::deser("simulated problem");
+    /// assert!(error.is_deserialization());
+    /// assert!(error.source().is_some());
+    /// ```
+    #[cfg_attr(not(feature = "_internal-semver"), doc(hidden))]
+    pub fn deser<T: Into<BoxError>>(source: T) -> Self {
+        Self {
+            kind: ErrorKind::Deserialization,
+            source: Some(source.into()),
+        }
+    }
+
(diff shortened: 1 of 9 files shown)


src/gax/src/error/core_error.rs
```diff
@@ -119,6 +119,65 @@ impl Error {
         matches!(self.kind, ErrorKind::Timeout)
     }
 
+    /// Not part of the public API, subject to change without notice.
+    ///
+    /// Creates an error representing a deserialization problem.
+    ///
+    /// Applications should have no need to use this function. The exception
+    /// could be mocks, but this error is too rare to merit mocks. If you are
+    /// writing a mock that extracts values from [wkt::Any], consider using a
```
dbolduc, line 128: s/using a/using/

src/gax/src/error/core_error.rs
```diff
@@ -119,6 +119,65 @@ impl Error {
...
+    /// `.expect()` calls instead.
+    ///
```
dbolduc, line 130: nit: extra newline

src/gax/src/error/core_error.rs
```diff
@@ -119,6 +119,65 @@ impl Error {
...
+    /// ```
+    #[cfg_attr(not(feature = "_internal-semver"), doc(hidden))]
+    pub fn deser<T: Into<BoxError>>(source: T) -> Self {
+        Self {
+            kind: ErrorKind::Deserialization,
+            source: Some(source.into()),
+        }
+    }
+
+    /// The request could not be completed before its deadline.
```
dbolduc, line 148: "The response could not be deserialized."

src/gax/src/error/core_error.rs
```diff
@@ -119,6 +119,65 @@ impl Error {
...
+    ///
+    /// This is always a client-side generated error. Note that the request may
+    /// or may not have started, and it may or may not complete in the service.
+    /// If the request mutates any state in the service, it may or may not be
+    /// safe to attempt the request again.
+    ///
```
dbolduc, line 154: Remove? The request was definitely started.

src/gax/src/error/core_error.rs
```diff
@@ -636,6 +699,24 @@ mod test {
         assert!(error.status().is_none(), "{error:?}");
     }
 
+    #[test]
+    fn serialization() {
+        let source = wkt::TimestampError::OutOfRange;
+        let error = Error::deser(source);
+        assert!(error.is_deserialization(), "{error:?}");
+        assert!(error.source().is_some(), "{error:?}");
```
dbolduc, line 707: nit: redundant, given the `matches!` below

src/lro/src/details.rs
```diff
@@ -149,12 +149,12 @@ where
...
 {
     if let Some(any) = op.response() {
-        return any.to_msg::<R>().map_err(Error::serde);
+        return any.to_msg::<R>().map_err(Error::deser);
     }
     if let Some(e) = op.error() {
         return Err(Error::service(gax::error::rpc::Status::from(e.clone())));
     }
-    Err(Error::serde("missing result in completed operation"))
+    Err(Error::deser("missing result in completed operation"))
```
dbolduc, line 157: Hmm... this is some sort of service bug. I don't like `serde` or `deser`.

I think it would get its own error type. Maybe switch it to `other` temporarily to unblock this PR?

src/gax/src/error/core_error.rs

src/gax/src/error/core_error.rs
coryan: LOL, now you know where I cut&pasted this from.

src/gax/src/error/core_error.rs
coryan: Removed.

src/lro/src/details.rs
coryan: Hmmm... grumble... Okay, but you need to come up with a name for that error kind.

dbolduc (approved):

PR #2334, opened 2025-05-30
impl(wkt): helper type for `i32` deserialization

### LINKED ISSUE #2323: Accept both strings and integers for `int32`, `fixed32` and `uint32` fields
>  int32, fixed32, uint32: JSON value will be a decimal number. **Either numbers or strings are accepted**. Empty strings are invalid.
> https://protobuf.dev/programming-guides/json/

coryan:
ProtoJSON requires accepting both strings and integers for integers. This PR
only adds support for signed 32-bit integers. Future PRs will deal with
unsigned integers and with 64-bit integers.

The changes to the generator and the generated code will also happen in
separate PRs.

Fixes #2323

### DIFF
--- a/src/wkt/src/internal/int32.rs
+++ b/src/wkt/src/internal/int32.rs
@@ -40,7 +40,11 @@ impl serde::de::Visitor<'_> for I32Visitor {
     where
         E: serde::de::Error,
     {
-        value.parse::<i32>().map_err(E::custom)
+        // ProtoJSON says that both strings and numbers are accepted. Parse the
+        // string as a f64 number (all JSON numbers are f64) and then try to
+        // parse that as a
+        let number = value.parse::<f64>().map_err(E::custom)?;
+        self.visit_f64(number)
     }
 
     fn visit_i64<E>(self, value: i64) -> std::result::Result<Self::Value, E>
@@ -48,12 +52,8 @@ impl serde::de::Visitor<'_> for I32Visitor {
         E: serde::de::Error,
     {
         match value {
-            _ if value < i32::MIN as i64 => {
-                Err(E::invalid_value(Other(&format!("{value}")), &ERRMSG))
-            }
-            _ if value > i32::MAX as i64 => {
-                Err(E::invalid_value(Other(&format!("{value}")), &ERRMSG))
-            }
+            _ if value < i32::MIN as i64 => Err(self::value_error(value)),
+            _ if value > i32::MAX as i64 => Err(self::value_error(value)),
             _ => Ok(value as i32),
         }
     }
@@ -63,9 +63,22 @@ impl serde::de::Visitor<'_> for I32Visitor {
         E: serde::de::Error,
     {
         match value {
-            _ if value > i32::MAX as u64 => {
-                Err(E::invalid_value(Other(&format!("{value}")), &ERRMSG))
-            }
+            _ if value > i32::MAX as u64 => Err(self::value_error(value)),
+            _ => Ok(value as i32),
+        }
+    }
+
+    fn visit_f64<E>(self, value: f64) -> std::result::Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        match value {
+            _ if value < i32::MIN as f64 => Err(self::value_error(value)),
+            _ if value > i32::MAX as f64 => Err(self::value_error(value)),
+            _ if value.fract().abs() > 0.0 => Err(self::value_error(value)),
+            // The number is "rounded towards zero". Because we are in range,
+            // and the fractional part is 0, this conversion should be safe.
+            // See https://doc.rust-lang.org/reference/expressions/operator-expr.html#r-expr.as.numeric.float-as-int
             _ => Ok(value as i32),
         }
     }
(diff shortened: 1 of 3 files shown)


src/wkt/src/internal/int32.rs
```diff
@@ -0,0 +1,160 @@
...
+impl serde::de::Visitor<'_> for I32Visitor {
+    type Value = i32;
+
+    fn visit_str<E>(self, value: &str) -> std::result::Result<Self::Value, E>
+    where
+        E: serde::de::Error,
+    {
+        // ProtoJSON says that both strings and numbers are accepted. Parse the
+        // string as a f64 number (all JSON numbers are f64) and then try to
+        // parse that as a
```
amanda-tarafa, line 45: nit: incomplete comment.

src/wkt/src/internal/int32.rs
```diff
@@ -0,0 +1,160 @@
...
+mod test {
+    use super::*;
+    use anyhow::Result;
+    use serde_json::{Value, json};
+    use serde_with::{DeserializeAs, SerializeAs};
+    use test_case::test_case;
+
+    #[test_case(0, 0)]
+    #[test_case("0", 0; "zero string")]
+    #[test_case("2.0", 2)]
```
amanda-tarafa, line 118: Is culture relevant? i.e. could this sometimes be "2,0"?

amanda-tarafa (approved):

src/wkt/src/internal/int32.rs
coryan: Fixed.

src/wkt/src/internal/int32.rs
coryan: I do not see how localization could possibly work for ProtoJSON, since (1) the sender and receiver may have different locales, and (2) there is nothing in the format to specify the source locale. I can only conclude that it always uses the "C" locale.  The same must be true of plain JSON.

Your question raises a followup: does Rust use the current locale to parse floating point numbers? The documentation is silent, which I read as "no localization", and the Interwebs recommend using external crates for localization (including, gulp!, `icu`).

I think we are safe.

coryan: PS: I also grew up writing "2,0", and MM-DD-YY still irks me.

src/wkt/tests/message_with_i32.rs
```diff
@@ -0,0 +1,272 @@
...
+    fn test_optional<T>(input: T, want: i32) -> Result
+    where
+        T: serde::ser::Serialize,
+    {
+        let value = json!({"optional": input});
+        let got = serde_json::from_value::<MessageWithI32>(value)?;
+        let output = json!({"optional": want});
+        if want == 0 {
+            return Ok(());
+        }
```
dbolduc, line 105: why? our annotations should respect presence, right?

src/wkt/tests/message_with_i32.rs
```diff
@@ -0,0 +1,272 @@
...
+    fn test_repeated<T>(input: T, want: i32) -> Result
+    where
+        T: serde::ser::Serialize,
+    {
+        let value = json!({"repeated": [input]});
+        let got = serde_json::from_value::<MessageWithI32>(value)?;
+        let output = json!({"repeated": [want]});
+        if want == 0 {
+            return Ok(());
+        }
```
dbolduc, line 141: `want` is never 0, and if it were, it'd be fine.

src/wkt/tests/message_with_i32.rs
```diff
@@ -0,0 +1,272 @@
...
+    #[test_case("0", 0)]
+    #[test_case("123", 123)]
+    #[test_case("-789", -789)]
+    fn test_map_key<T>(input: T, want: i32) -> Result
+    where
+        T: Into<String>,
+    {
+        let value = json!({"mapKey": {input: "test"}});
+        let got = serde_json::from_value::<MessageWithI32>(value)?;
+        let output = json!({"mapKey": {want.to_string(): "test"}});
```
dbolduc, line 212: `to_string()`? shouldn't we reserialize these without quotes?

(I might be lost here)

src/wkt/tests/message_with_i32.rs
```diff
@@ -0,0 +1,272 @@
...
+    #[test_case("-789", 456, -789, 456)]
+    #[test_case("-789", "567", -789, 567)]
+    fn test_map_key_value<K, V>(key: K, value: V, want_key: i32, want_value: i32) -> Result
+    where
+        K: Into<String>,
+        V: serde::Serialize,
+    {
+        let value = json!({"mapKeyValue": {key: value}});
+        let got = serde_json::from_value::<MessageWithI32>(value)?;
+        let output = json!({"mapKeyValue": {want_key.to_string(): want_value}});
```
dbolduc, line 249: here it is again. is this a macro thing I am not getting?

dbolduc (approved):

coryan (commented):
PTAL

dbolduc (approved):

src/wkt/tests/message_with_i32.rs
dbolduc: Ack, thanks

PR #2336, opened 2025-05-30
feat(gax): new error kind for serialization

coryan:
With this we have `is_serialization()` for serialization and
`is_deserialization()` for, well, deserialization.

In a future PR I will change the generator to use `ser()` directly.

Part of the work for #2312


src/gax/src/error/core_error.rs
```diff
@@ -177,6 +177,64 @@ impl Error {
...
+            kind: ErrorKind::Serialization,
+            source: Some(source.into()),
+        }
+    }
+
+    /// The request could not be serialized.
+    ///
+    /// This is always a client-side generated error, generated before the
+    /// request is made. This error is never transient: the serialization is
+    /// deterministic (module out of memory conditions), and will fail on future
```
dbolduc: nit: s/module/modulo/

src/gax/src/error/core_error.rs
```diff
@@ -177,6 +177,64 @@ impl Error {
...
+    /// prevent them. If you cannot use a public issue tracker, contact
+    /// [Google Cloud support].
+    ///
+    /// A less common cause for serialization problems may be an out of memory
+    /// condition, or any other runtime error. Use `format!("{:?}", ...)` to
+    /// examine the error as it should include the original problem.
+    ///
+    /// Finally, using a [wkt::Any] with a gRPC-based client is unsupported. As
+    /// of this writing, no client libraries send `Any` via gRPC, but this
+    /// could be a problem in the future.
```
dbolduc: `StorageControl::get_operation` sends an `Any` over gRPC.

It is more like "sending arbitrary messages in an `wkt::Any` over gRPC is unsupported".

Also, customers do not know which clients are HTTP based vs. gRPC based.

src/gax/src/error/core_error.rs

src/gax/src/error/core_error.rs
coryan: > `StorageControl::get_operation` sends an `Any` over gRPC.

I believe it *receives* `Any` over gRPC, but does not send them.  This error is about serialization *only*. We have a different error kind / predicate for deserialization.

> Also, customers do not know which clients are HTTP based vs. gRPC based.

Yes. For now I am going to ignore that, but we will need to document it. It is not possible to hide the transport.

dbolduc: Oops, you're right.

dbolduc (approved):

PR #2338, opened 2025-05-30
fix(gax): no sleeps past overall timeout

### LINKED ISSUE #2307: Client library should not sleep past overall timeout
https://github.com/googleapis/google-cloud-rust/blob/827a0b3f336edeb1cb5bf2adf94a9abeacf0c484/src/gax/src/retry_loop_internal.rs#L73-L79

if `delay` is more than `remaining_time`, we should kill the loop instead of sleeping past our deadline.

dbolduc:
Fixes #2307 

Compare how long we are going to sleep against the remaining time. If there is not enough time left, terminate the retry loop with a `timeout` error. The source of this timeout error holds the error from the previous attempt.

We refactor the implementation (with `RetryLoopAttempt`) to make it clear via code that either (1) we are on the initial attempt or (2) we have a previous error we can return.

For the tests, I preferred to use a single `mockall::Sequence` for all operations, instead of using one per mock (and only where necessary). I think one `Sequence` makes the flow easier to follow, and makes it more obvious where the expectations come from.

### DIFF
--- a/src/gax/src/retry_loop_internal.rs
+++ b/src/gax/src/retry_loop_internal.rs
@@ -14,10 +14,27 @@
 
 use super::Result;
 use super::backoff_policy::BackoffPolicy;
+use super::error::Error;
 use super::loop_state::LoopState;
 use super::retry_policy::RetryPolicy;
 use super::retry_throttler::RetryThrottler;
 use std::sync::{Arc, Mutex};
+
+enum RetryLoopAttempt {
+    // The first attempt
+    Initial,
+    // (Attempt count, previous error)
+    Retry(u32, Error),
+}
+
+impl RetryLoopAttempt {
+    fn count(&self) -> u32 {
+        match self {
+            RetryLoopAttempt::Initial => 0,
+            RetryLoopAttempt::Retry(count, _) => *count,
+        }
+    }
+}
 
 /// Runs the retry loop for a given function.
 ///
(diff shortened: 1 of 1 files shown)


src/gax/src/retry_loop_internal.rs
```diff
@@ -76,29 +97,26 @@ where
...
                     .expect("retry throttler lock is poisoned")
                     .on_retry_failure(&flow);
-                on_error(&sleep, flow, delay).await?;
+                match flow {
+                    LoopState::Permanent(e) | LoopState::Exhausted(e) => return Err(e),
+                    LoopState::Continue(e) => {
+                        // TODO(#2307) - For a stricter limit, we should query
+                        //               the retry policy to determine the
+                        //               remaining time.
+                        // let remaining_time = retry_policy.remaining_time(loop_start, attempt_count);
```
coryan, line 106: or you could could move the sleep to the beginning of the loop, and just not do it for the first request? Your new `Attempt` enum should make that much easier.

dbolduc: Good suggestion. https://github.com/googleapis/google-cloud-rust/pull/2351

dbolduc:
Thanks for the review. Closing. It will be easier to send new PRs than to salvage this one.

src/gax/src/retry_loop_internal.rs
```diff
@@ -637,15 +656,219 @@ mod test {
...
+        )
+        .await;
+        let err = response.expect_err("retry loop should terminate");
+        assert!(err.is_timeout(), "{err:?}");
+        // Confirm that we expose the last seen status from the operation
+        let got = err
+            .source()
+            .and_then(|e| e.downcast_ref::<Error>())
+            .and_then(|e| e.status());
+        assert_eq!(got, Some(&transient_status()), "{err}");
```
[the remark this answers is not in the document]
dbolduc: Done

Issue #2328, opened 2025-05-30
Breakdown the fixes for the ProtoJSON conformance test


coryan:
We need to break down this work:

- [ ] `Required.Proto3.JsonInput.DoubleFieldMaxNegativeValue.JsonOutput` # Output was not equivalent to reference message: modified: optional_double: -2.22507e-308 -> -2.2250700000000003e-308
- [ ] `Required.Proto3.JsonInput.DoubleFieldMinPositiveValue.JsonOutput` # Output was not equivalent to reference message: modified: optional_double: 2.22507e-308 -> 2.2250700000000003e-308
- [ ] `Required.Proto3.JsonInput.TimestampJsonInputLowercaseT` # Should have failed to parse, but didn't.
- [ ] `Required.Proto3.JsonInput.TimestampJsonInputLowercaseZ` #` Should have failed to parse, but didn't.
- [ ] `Required.Proto3.JsonInput.TimestampJsonInputMissingT` # Should have failed to parse, but didn't.

----

```
ERROR, test=Required.Proto3.JsonInput.DoubleFieldMinPositiveValue.JsonOutput: Output was not equivalent to reference message: modified: optional_double: 2.22507e-308 -> 2.2250700000000003e-308
, request=goo.gle/debugonly   json_payload: "{\"optionalDouble\": 2.22507e-308}" requested_output_format: JSON message_type: "protobuf_test_messages.proto3.TestAllTypesProto3" test_category: JSON_TEST, response=goo.gle/debugonly   json_payload: "{\"optionalDouble\":2.2250700000000003e-308}"
ERROR, test=Required.Proto3.JsonInput.DoubleFieldMaxNegativeValue.JsonOutput: Output was not equivalent to reference message: modified: optional_double: -2.22507e-308 -> -2.2250700000000003e-308
, request=goo.gle/debugonly   json_payload: "{\"optionalDouble\": -2.22507e-308}" requested_output_format: JSON message_type: "protobuf_test_messages.proto3.TestAllTypesProto3" test_category: JSON_TEST, response=goo.gle/debugonly   json_payload: "{\"optionalDouble\":-2.2250700000000003e-308}"
[...]

PR #2345, opened 2025-05-31
impl(generator/rust): support `bool` keys in maps

coryan:
Evidently they are legal in Protobuf, though no Google Cloud service
uses them. ProtoJSON requires converting the `bool` to strings when
used as map keys.

Related to #2328


src/protojson-conformance/src/generated/test_protos/mod.rs
```diff
@@ -339,6 +339,7 @@ pub struct TestAllTypesProto3 {
     pub map_int32_double: std::collections::HashMap<i32, f64>,
 
     #[serde(skip_serializing_if = "std::collections::HashMap::is_empty")]
+    #[serde_as(as = "std::collections::HashMap<serde_with::DisplayFromStr, _>")]
     pub map_bool_bool: std::collections::HashMap<bool, bool>,
```
dbolduc, line 343: An incredibly important data structure

PR #2346, opened 2025-06-01
impl(generator/rust): skip enums in ProtoJSON output

coryan:
Motivated by #2328


generator/internal/rust/codec.go
```diff
@@ -453,6 +453,8 @@ func fieldSkipAttributes(f *api.Field) []string {
 		api.SINT32_TYPE,
 		api.SINT64_TYPE:
 		return []string{`#[serde(skip_serializing_if = "wkt::internal::is_default")]`}
+	case api.ENUM_TYPE:
```
dbolduc: nit: combine with previous case

dbolduc (approved):
Thanks

generator/internal/rust/codec.go
coryan: Done.

PR #2351, opened 2025-06-02
refactor(gax): capture prev error in retry loop

dbolduc:
Towards #2307 

Refactor the retry loop to preserve previous errors. This makes it easier to chain the error when reporting a timeout from the retry loop being exhausted.

### DIFF
--- a/src/gax/src/retry_loop_internal.rs
+++ b/src/gax/src/retry_loop_internal.rs
@@ -14,10 +14,28 @@
 
 use super::Result;
 use super::backoff_policy::BackoffPolicy;
+use super::error::Error;
 use super::loop_state::LoopState;
 use super::retry_policy::RetryPolicy;
 use super::retry_throttler::RetryThrottler;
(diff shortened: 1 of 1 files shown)


src/gax/src/retry_loop_internal.rs
```diff
@@ -14,10 +14,28 @@
...
 use super::loop_state::LoopState;
 use super::retry_policy::RetryPolicy;
 use super::retry_throttler::RetryThrottler;
 use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+enum RetryLoopAttempt {
+    // The first attempt
+    Initial,
+    // (Attempt count, time to sleep, previous error)
```
coryan: I think this is more descriptive:

```suggestion
    // (Attempt count, backoff delay, previous error)
```

dbolduc: Done.

coryan (approved):

Issue #2040, opened 2025-05-08
Use builder for `Storage::insert_object`


coryan:
These functions should return a `RequestBuilder`. We will need to re-enable some things:

https://github.com/googleapis/google-cloud-rust/blob/32ef2f3dc7419710b8fe9f9c2193d89cfacbdd55/src/storage-control/src/generated/gapic/.sidekick.toml#L40

https://github.com/googleapis/google-cloud-rust/blob/32ef2f3dc7419710b8fe9f9c2193d89cfacbdd55/src/storage-control/src/generated/gapic/.sidekick.toml#L44-L49

coryan:
I think we can mark this closed, we have a builder and it can initialize all the request parameters.  We still need to add options for retries, but those belong in the corresponding bugs.

PR #2352, opened 2025-06-02
fix(storage)!: use builder for Storage::insert_object

suzmue:
Simple rewrite for using the same builder pattern as read_object uses. For now, just use fields for each of the required values.

Refactor for testing like for read_object coming in follow up CL.

For #2040


src/storage/src/client.rs
```diff
@@ -111,64 +111,18 @@ impl Storage {
...
-        tracing::info!("builder={builder:?}");
-
-        let response = builder.send().await.map_err(Error::io)?;
-        if !response.status().is_success() {
-            return gaxi::http::to_http_error(response).await;
-        }
-        let response = response.json::<v1::Object>().await.map_err(Error::io)?;
-
-        Ok(Object::from(response))
+    pub fn insert_object(&self) -> InsertObject {
```
coryan, line 124: For a future PR: I think we should restore the required arguments for bucket and object, and maybe payload too.

coryan (approved):

PR #2354, opened 2025-06-02
fix(storage): support customer supplied encryption key

suzmue:
Support a customer supplied encryption key for both insert_object and read_object. Instead of piping through the exact ReadObjectRequest interface, only require the user to set the key value. The algorithm is always AES256 and the sha256 can be computed from the key. Also add an integration test to make sure upload and download with the key are working.

For #2263


src/storage/src/client.rs
```diff
@@ -706,6 +728,54 @@ mod tests {
...
+            .set_object("object")
+            // TODO: this is a bad interface, algorithm is always AES256 and the sha256 is calculated from the key bytes.
+            // Other client libraries allow user to simply specify "key", we should do that.
+            .set_common_object_request_params(
+                control::model::CommonObjectRequestParams::new()
+                    .set_encryption_algorithm("AES256")
+                    // Random key generated for this test.
+                    .set_encryption_key_bytes(key)
+                    .set_encryption_key_sha256_bytes(sha256),
+            )
```
coryan: Here and elsewhere, convert `TODO:` into `TODO(#....)` with a GitHub issue backing them.

For a future PR:

```rust
struct KeyAes256 {
  key: bytes::Bytes
};

impl KeyAes256 {
  pub fn from(key: &[u8]) -> Result<Self> { /* validate length */
}

impl std::convert::From<KeyAes256> for control::model::CommonObjectRequestParams {
  fn from(value: KeyAes256) -> Self {
    // Set the algorithm, and SHA256 and key computed from `value.key`.
  }
}

// Now `KeyAes256` implements `Into<CommonObjectRequestParams>` and one could write:
//
client.read_object().set_common_object_request_params(KeyAes256::from(blah blah )?).send().await?
```

src/storage/src/client.rs
```diff
@@ -426,12 +428,20 @@ impl ReadObject {
...
+    /// Sets the value of key in [common_object_request_params][control::model::ReadObjectRequest::common_object_request_params].
+    pub fn set_key<T>(mut self, v: T) -> Self
     where
-        T: Into<control::model::CommonObjectRequestParams>,
+        T: Into<bytes::Bytes>,
     {
-        self.request.common_object_request_params = Option::Some(v.into());
+        use sha2::{Digest, Sha256};
+        let v: bytes::Bytes = v.into();
+        // let sha256 = Sha256::digest(v.clone()).to_owned();
```
coryan: Remove?

src/storage/src/client.rs
```diff
@@ -426,12 +428,20 @@ impl ReadObject {
         self
     }
 
-    /// Sets the value of [common_object_request_params][control::model::ReadObjectRequest::common_object_request_params].
-    pub fn set_common_object_request_params<T>(mut self, v: T) -> Self
+    /// Sets the value of key in [common_object_request_params][control::model::ReadObjectRequest::common_object_request_params].
+    pub fn set_key<T>(mut self, v: T) -> Self
     where
-        T: Into<control::model::CommonObjectRequestParams>,
+        T: Into<bytes::Bytes>,
```
coryan, line 487: I think you want some specific type because there are length restrictions on that key, no?

See my other comment and let me know if you would rather do that in a future PR.

src/storage/src/client.rs
```diff
@@ -556,6 +564,26 @@ impl ReadObject {
             .iter()
             .fold(builder, |b, v| b.query(&[("ifMetagenerationNotMatch", v)]));
 
+        // TODO: Add some client side verication that all fields are filled and algorithm is AES256.
```
coryan: Use a github issue to track this bug.

src/storage/src/client.rs
```diff
@@ -706,6 +734,48 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_read_object_headers() -> Result {
+        let client = Storage::builder()
+            .with_credentials(auth::credentials::testing::test_credentials())
+            .build()
+            .await?;
+        // A precomputed random key encoded in base64 and its Sha256.
+        let key_base64 = "V8knzq04ASfjVn7eNM/UGsXnHEvpRByfCRE/gkNJfK0=";
+        let key_sha256_base64 = "0hA2ubfDbe2m0t8tZIDKYGd53WYJkoWSXGCPvYESW58=";
```
coryan: Consider using a plain text "key" and computing the expected base64 and SHA256 values.  These will look like secrets to the secret scanners, and I would rather avoid the false positives.

src/storage/src/client.rs
suzmue: These TODOs were intended to be resolved in this PR, will remove or update.

src/storage/src/client.rs

src/storage/src/client.rs

src/storage/src/client.rs
suzmue: I made some other significant changes to this PR, so may leave this out of this PR. Want to think about how users will use this if they get this far in the code and have a vector of unknown length.

src/storage/src/client.rs
```diff
@@ -650,6 +817,49 @@ mod tests {
...
+
+    #[tokio::test]
+    async fn test_read_object_headers() -> Result {
+        let client = Storage::builder()
+            .with_credentials(auth::credentials::testing::test_credentials())
+            .build()
+            .await?;
+        // Make a 32-byte key.
+        let key: Vec<u8> = "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa".into();
+        assert_eq!(key.len(), 32);
```
coryan: optional, match the other key?

```suggestion
        let key = vec![b'a'; 32];
```

src/storage/src/client.rs
```diff
@@ -546,17 +575,155 @@ impl ReadObject {
...
+            .insert_object("projects/_/buckets/bucket", "object", "hello")
+            .http_request_builder()
+            .await
+            .inspect_err(|e| assert!(e.is_authentication()))
+            .expect_err("invalid credentials should err");
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_insert_object_bad_bucket() -> Result {
```
coryan: Seems unrelated to this PR?

src/storage/src/client.rs
```diff
@@ -546,17 +575,155 @@ impl ReadObject {
...
+        );
+        assert_eq!(
+            b"hello",
+            insert_object_builder.body().unwrap().as_bytes().unwrap()
+        );
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_insert_object_error_credentials() -> Result {
```
coryan: Ditto

src/storage/src/client.rs
```diff
@@ -546,17 +575,155 @@ impl ReadObject {
...
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
     use std::collections::HashMap;
     type Result = std::result::Result<(), Box<dyn std::error::Error>>;
 
+    #[tokio::test]
+    async fn test_insert_object() -> Result {
```
coryan: And this too?

src/integration-tests/src/storage.rs
```diff
@@ -76,6 +76,63 @@ pub async fn objects(builder: storage::client::ClientBuilder) -> Result<()> {
     Ok(())
 }
 
+pub async fn objects_customer_supplied_encryption(
```
coryan, line 79: Oh, I see, you need `insert_object()` with support of encryption so you can test the reading function... Hmm... I think it would be better if that was a separate PR, but the code looks great.

coryan (approved):
The PR title does not match the contents anymore. Consider splitting the code for `insert_object()` and then rebase.

src/storage/src/client.rs
suzmue: Moved these tests to #2369

src/storage/src/client.rs
suzmue: Moved these tests to #2369

src/integration-tests/src/storage.rs
suzmue: Exactly 😄  Moved the basic tests and initial refactor to #2369

src/storage/src/client.rs
suzmue: Oh yes, missed this spot! Thanks!

src/storage/src/client.rs
suzmue: Moved to other PR

coryan (approved):

Issue #2080, opened 2025-05-09
Executable Sourced Credentials


sai-sunder-s:
Support Workload Identity Federation through Executable Sourced Credentials.

sai-sunder-s:
Implemented in #2356

PR #2356, opened 2025-06-02
feat(auth): add support for executable sourced credentials

alvarowolfx:
Towards #2080

### DIFF
--- a/src/auth/src/credentials/external_account.rs
+++ b/src/auth/src/credentials/external_account.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use super::dynamic::CredentialsProvider;
+use super::external_account_sources::executable_sourced::ExecutableSourcedCredentials;
 use super::external_account_sources::url_sourced::UrlSourcedCredentials;
 use super::internal::sts_exchange::{ClientAuthentication, ExchangeTokenRequest, STSHandler};
 use super::{CacheableResource, Credentials};
@@ -47,7 +48,7 @@ enum CredentialSource {
     Url(UrlSourcedCredentials),
     File {},
     Aws {},
-    Executable {},
+    Executable(ExecutableSourcedCredentials),
 }
 
 impl CredentialSource {
@@ -58,20 +59,10 @@ impl CredentialSource {
     ) -> Credentials {
         match self {
             Self::Url(source) => {
-                let token_provider = ExternalAccountTokenProvider {
-                    subject_token_provider: source,
-                    config,
-                };
-                let cache = TokenCache::new(token_provider);
-                Credentials {
-                    inner: Arc::new(ExternalAccountCredentials {
-                        token_provider: cache,
-                        quota_project_id,
-                    }),
-                }
+                Self::make_credentials_from_source(source, config, quota_project_id)
             }
-            Self::Executable { .. } => {
-                unimplemented!("executable sourced credential not supported yet")
+            Self::Executable(source) => {
+                Self::make_credentials_from_source(source, config, quota_project_id)
             }
             Self::File { .. } => {
                 unimplemented!("file sourced credential not supported yet")
@@ -79,6 +70,27 @@ impl CredentialSource {
             Self::Aws { .. } => {
                 unimplemented!("AWS sourced credential not supported yet")
             }
+        }
+    }
+
+    fn make_credentials_from_source<T>(
+        subject_token_provider: T,
+        config: ExternalAccountConfig,
+        quota_project_id: Option<String>,
+    ) -> Credentials
+    where
+        T: SubjectTokenProvider + 'static,
+    {
+        let token_provider = ExternalAccountTokenProvider {
+            subject_token_provider,
+            config,
+        };
+        let cache = TokenCache::new(token_provider);
+        Credentials {
+            inner: Arc::new(ExternalAccountCredentials {
+                token_provider: cache,
+                quota_project_id,
+            }),
         }
     }
 }
--- a/src/auth/src/credentials/external_account_sources/executable_sourced.rs
+++ b/src/auth/src/credentials/external_account_sources/executable_sourced.rs
@@ -1,0 +1,231 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use gax::error::CredentialsError;
+use serde::{Deserialize, Serialize};
+use tokio::process::Command;
+
+use crate::{Result, credentials::external_account::SubjectTokenProvider};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub(crate) struct ExecutableSourcedCredentials {
+    pub executable: ExecutableConfig,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, Default)]
+pub(crate) struct ExecutableConfig {
+    pub command: Option<String>,
+    pub timeout_millis: Option<u32>,
+    pub output_file: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+struct ExecutableResponse {
+    version: i32,
+    success: bool,
+    token_type: Option<String>,
+    expiration_time: Option<i64>,
+    id_token: Option<String>,
+    saml_response: Option<String>,
+    code: Option<String>,
+    message: Option<String>,
+}
+
+impl ExecutableResponse {
+    fn to_cred_error(&self) -> CredentialsError {
+        return match &self {
+            ExecutableResponse {
+                message: Some(message),
+                code: Some(code),
+                ..
+            } => {
+                let msg =
+                    format!("{MSG}, response contains unsuccessful response: ({code}) {message}");
+                CredentialsError::from_msg(false, msg)
+            }
+            _ => {
+                let msg = format!(
+                    "{MSG}, response must include `code` and `message` fields when unsuccessful"
+                );
+                CredentialsError::from_msg(false, msg)
+            }
+        };
+    }
+}
+
+const MSG: &str = "failed to read subject token";
+// const DEFAULT_TIMEOUT: Duration = Duration::from_secs(30); TODO: enforce timeout
+
+#[async_trait::async_trait]
+impl SubjectTokenProvider for ExecutableSourcedCredentials {
+    async fn subject_token(&self) -> Result<String> {
+        let output = match self.executable.clone() {
+            ExecutableConfig {
+                output_file: Some(output_file),
+                ..
+            } => Self::from_output_file(output_file).await,
+            ExecutableConfig {
+                command: Some(command),
+                ..
+            } => Self::from_command(command).await,
+            _ => Err(CredentialsError::from_msg(
+                false,
+                format!("{MSG}, either `output_file` or `command` needs to be informed"),
+            )),
+        }?;
+
+        let output = output.trim().to_string();
+        let subject_token = Self::parse_token(output)?;
+
+        if subject_token.is_empty() {
+            let msg = format!("{MSG}, subject token is empty");
+            return Err(CredentialsError::from_msg(false, msg));
+        }
+
+        Ok(subject_token)
+    }
+}
+
+impl ExecutableSourcedCredentials {
+    async fn from_output_file(output_file: String) -> Result<String> {
+        let content = std::fs::read_to_string(output_file)
+            .map_err(|e| CredentialsError::from_source(false, e))?;
+        return Ok(content);
+    }
+
+    async fn from_command(command: String) -> Result<String> {
+        let (command, args) = Self::split_command(command);
+        let output = Command::new(command.clone())
+            .args(&args)
+            .output()
+            .await
+            .map_err(|e| CredentialsError::from_source(false, e))?;
+        // let output = timeout(DEFAULT_TIMEOUT, output); // TODO: enforce timeout
+
+        if !output.status.success() {
+            let stderr = String::from_utf8(output.stderr)
+                .map_err(|e| CredentialsError::from_source(false, e))?;
+            let msg = format!("{MSG}, command execution failed, stderr=<{stderr}>");
+            return Err(CredentialsError::from_msg(false, msg));
+        }
+
+        let subject_token = String::from_utf8(output.stdout)
+            .map_err(|e| CredentialsError::from_source(false, e))? // TODO: add more details and better handling
+            .to_string();
+
+        Ok(subject_token)
+    }
+
+    fn split_command(command: String) -> (String, Vec<String>) {
+        let mut parts = command.split_whitespace();
+
+        let command = parts.next().unwrap(); // TODO: remove unwrap
+        let args: Vec<String> = parts.map(String::from).collect();
+
+        (command.to_string(), args)
+    }
+
+    fn parse_token(output: String) -> Result<String> {
+        let res = serde_json::from_str::<ExecutableResponse>(output.as_str())
+            .map_err(|e| CredentialsError::from_source(false, e))?; // TODO: Add details on expected format
+
+        if !res.success {
+            return Err(res.to_cred_error());
+        }
+
+        if let Some(id_token) = res.id_token {
+            return Ok(id_token);
+        }
+
+        if let Some(saml_response) = res.saml_response {
+            return Ok(saml_response);
+        }
+
+        Err(CredentialsError::from_msg(
+            false,
+            "contains unsupported token type",
+        ))
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::credentials::internal::sts_exchange::JWT_TOKEN_TYPE;
+    use serde_json::json;
+    use tokio::time::{Duration, Instant};
+
+    type TestResult = std::result::Result<(), Box<dyn std::error::Error>>;
+
+    #[tokio::test]
+    async fn read_token_from_command() -> TestResult {
+        let expiration = (Instant::now() + Duration::from_secs(3600))
+            .elapsed()
+            .as_millis();
+        let json_response = json!({
+            "success": true,
+            "version": 1,
+            "expiration_time": expiration,
+            "token_type": JWT_TOKEN_TYPE,
+            "id_token":"an_example_token",
+        })
+        .to_string();
+        let file = tempfile::NamedTempFile::new().unwrap();
+        let path = file.into_temp_path();
+        std::fs::write(&path, json_response).expect("Unable to write to temp file with command");
+
+        let token_provider = ExecutableSourcedCredentials {
+            executable: ExecutableConfig {
+                command: Some(format!("cat {}", path.to_str().unwrap()).into()),
+                ..ExecutableConfig::default()
+            },
+        };
+        println!("{}", token_provider.executable.command.clone().unwrap());
+        let resp = token_provider.subject_token().await?;
+
+        assert_eq!(resp, "an_example_token".to_string());
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn read_token_from_output_file() -> TestResult {
+        let expiration = (Instant::now() + Duration::from_secs(3600))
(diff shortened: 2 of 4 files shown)


src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,317 @@
...
+    token_type: String,
+    expiration_time: i64,
+    id_token: Option<String>,
+    saml_response: Option<String>,
+    code: Option<String>,
+    message: Option<String>,
+}
+
+impl ExecutableResponse {
+    fn to_cred_error(&self) -> CredentialsError {
```
coryan: This is a lot of code to leave untested, please fix.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,317 @@
...
+        match &self {
+            ExecutableResponse {
+                message: Some(message),
+                code: Some(code),
+                ..
+            } => {
+                let msg =
+                    format!("{MSG}, response contains unsuccessful response: ({code}) {message}");
+                CredentialsError::from_msg(false, msg)
+            }
```
coryan: It seems like you should define a specific error type for this.  The caller should not be forced to parse the code and message from a string.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,317 @@
...
+                    "{MSG}, response must include `code` and `message` fields when unsuccessful"
+                );
+                CredentialsError::from_msg(false, msg)
+            }
+        }
+    }
+}
+
+const MSG: &str = "failed to read subject token";
+const DEFAULT_TIMEOUT_SECS: u32 = 30;
```
coryan: You should be able to create a `Duration::from_secs()`, they are `const`:

https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_secs

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,317 @@
...
+        // For security reasons, we need our consumers to set this environment variable to allow executables to be run.
+        let allow_executable = std::env::var(ALLOW_EXECUTABLE_ENV)
+            .ok()
+            .unwrap_or("0".to_string());
+        if allow_executable != "1" {
+            return Err(CredentialsError::from_msg(
+                false,
+                "executables need to be explicitly allowed (set GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES to '1') to run",
+            ));
+        }
```
coryan, line 152: This is very strange.  So they set an explicit configuration to use an executable and *then* they also need to set an environment variable?

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    fn split_command(command: String) -> (String, Vec<String>) {
+        let mut parts = command.split_whitespace();
+
+        if let Some(command) = parts.next() {
+            let args: Vec<String> = parts.map(String::from).collect();
+            return (command.to_string(), args);
+        }
+
+        (command, vec![])
```
coryan: I think you can simplify this to:


```suggestion
    let mut parts = command.split_whitespace();
    let Some(cmd) = parts.next() else {
        return (command, vec![]);
    };
    (cmd.to_string(), parts.map(String::from).collect())
}
```

coryan: Aside: this is a really terrible design in the input file format, they should have used a struct with the command and the arguments.  It is impossible to use arguments (or executable paths) that have spaces in them with the current design.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,317 @@
...
+            "id_token":"an_example_token",
+        })
+        .to_string();
+        let file = tempfile::NamedTempFile::new().unwrap();
+        let path = file.into_temp_path();
+        std::fs::write(&path, json_response).expect("Unable to write to temp file with command");
+
+        let token_provider = ExecutableSourcedCredentials {
+            executable: ExecutableConfig {
+                command: Some(format!("cat {}", path.to_str().unwrap())),
```
coryan: How do you know that `cat` is in the path?  That may be true in the CI build for Windows, but won't be true in general.

Also, how do you know that `file` does not have a space?

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,317 @@
...
+#[cfg(test)]
+mod test {
+    use super::*;
+    use crate::constants::JWT_TOKEN_TYPE;
+    use scoped_env::ScopedEnv;
+    use serde_json::json;
+    use serial_test::serial;
+    use tokio::time::{Duration, Instant};
+
+    type TestResult = std::result::Result<(), Box<dyn std::error::Error>>;
```
coryan: Consider: `type TestResult = anyhow::Result<()>` we started using that everywhere else.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,317 @@
...
+    pub timeout_millis: Option<u32>,
+    pub output_file: Option<String>,
+}
+
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+struct ExecutableResponse {
+    version: i32,
+    success: bool,
+    token_type: String,
+    expiration_time: i64,
```
coryan: This is only used in tests, and it is unclear what the units are. Are these seconds from the epoch? Milliseconds from the epoch? Something else?

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,317 @@
...
+
+    type TestResult = std::result::Result<(), Box<dyn std::error::Error>>;
+
+    #[tokio::test]
+    #[serial]
+    async fn read_token_from_command() -> TestResult {
+        let _e = ScopedEnv::set(ALLOW_EXECUTABLE_ENV, "1");
+        let expiration = (Instant::now() + Duration::from_secs(3600))
+            .elapsed()
+            .as_millis();
```
coryan: I do not think this is a meaningful number:

https://doc.rust-lang.org/std/time/struct.Instant.html#method.elapsed

If this field is of any use it should be "time since the epoch", or some kind of timestamp. Not the time since the timestamp was created (which in this test will likely be something close to -3600s).

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,317 @@
...
+        use std::error::Error;
+        use std::os::unix::fs::PermissionsExt;
+
+        let _e = ScopedEnv::set(ALLOW_EXECUTABLE_ENV, "1");
+
+        let file_contents = "#!/bin/bash
+while true;
+do
+    echo \"working\"
+done";
```
coryan, line 473: If you need something like this, create a binary in the crate and then you can invoke it...

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,324 @@
...
+        let resp = token_provider.subject_token().await?;
+
+        assert_eq!(resp, "an_example_token".to_string());
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    #[serial]
+    async fn read_token_from_output_file() -> TestResult {
```
coryan: This would succeed even if no executable is invoked to generate the file. Not a really good test, please fix.

coryan (changes requested):

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: I'll add more documentation about that field, but the AIP docs says this:

```
expiration_time | number | The optional 3rd party subject token expiration time in seconds (unix epoch time). Only required in the response when an output file is specified in the credential configuration
```

So yes, it's seconds from epoch

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: good point, I'll create a new error type and use it as source to create a `CredentialsError`.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: changed.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: yeah, I agree that it's weird, but out of my control as I'm just following the AIP guide.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: TIL that inline `else`, thanks for the tip.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: done

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: duh, I don't know why I used that `elapsed` method. Indeed right now the expiration time is not being validated to be a time in the future and the AIP is not explicit about that. But in any case, I changed it to grab system time and add the extra 1h

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: that's a good question, I was having a lot of trouble running these tests, specially the one with timeout, as I was trying to use the `sleep` command and it was not available on `macos` (tho it is available on my mac machine here). Do you have any recommendation on a better way to test this ? 

On the `file` name, AFAIK `temp` files don't have spaces on it.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: that's the flow defined on the AIP, if the output file is informed, we need to check it first, check if is valid and just return it without executing any command. One flow that is missing is that when the file is expired or not there, then we need to execute the command.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: hmm, will search how to do that. Do we have any example on the repo already ?

alvarowolfx (commented):
answered some question and pushed some minor fixes. Will add more coverage and need to fix the flow where the output file is not valid. Also the `command` is always required per AIP

src/auth/src/credentials/external_account_sources/executable_sourced.rs
coryan: > that's a good question, I was having a lot of trouble running these tests, specially the one with timeout, as I was trying to use the `sleep` command and it was not available on `macos` (tho it is available on my mac machine here). Do you have any recommendation on a better way to test this ?

I believe you can create a binary target in `src/auth/${new target}`, take a dependency on this crate, and then run that binary from your program. You will need to compute the path, but there are variables to discover it.

> 
> On the `file` name, AFAIK `temp` files don't have spaces on it.

Until you run into somebody who sets `export TMPDIR="/tmp/some people like spaces in their  paths/foo"`.  And on Windows they are even more common.

coryan: BTW, an assertion that says "this test does not work if temporary files have spaces" would be enough.

src/auth/src/credentials/external_account.rs
```diff
@@ -81,6 +72,27 @@ impl CredentialSource {
...
         }
     }
+
+    fn make_credentials_from_source<T>(
+        subject_token_provider: T,
+        config: ExternalAccountConfig,
+        quota_project_id: Option<String>,
+    ) -> Credentials
+    where
+        T: SubjectTokenProvider + 'static,
```
sai-sunder-s, line 169: curious why we need static here?

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,449 @@
...
+    async fn subject_token(&self) -> Result<String> {
+        if let Some(output_file) = self.executable.output_file.clone() {
+            let token = Self::from_output_file(output_file).await;
+            if token.is_ok() {
+                return token;
+            }
+        }
+        let timeout = match self.executable.timeout_millis {
+            Some(timeout) => Duration::from_millis(timeout.into()),
+            None => DEFAULT_TIMEOUT_SECS,
```
sai-sunder-s: this can be done once instead of each time?

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,449 @@
...
+        }
+
+        let output = String::from_utf8(output.stdout)
+            .map_err(|e| CredentialsError::from_source(true, e))?
+            .to_string();
+
+        Self::parse_token(output)
+    }
+
+    fn split_command(command: String) -> (String, Vec<String>) {
```
sai-sunder-s, line 179: would be helpful to understand with a comment showing input and output for this method

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,449 @@
...
+                "executables need to be explicitly allowed (set GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES to '1') to run",
+            ));
+        }
+
+        let (command, args) = Self::split_command(command);
+        let output = Command::new(command.clone()).args(&args).output();
+        let output = tokio_timeout(timeout, output.into_future())
+            .await
+            .map_err(|e| CredentialsError::from_source(true, e))?
+            .map_err(|e| CredentialsError::from_source(true, e))?;
```
sai-sunder-s, line 158: Should this be twice?

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,449 @@
...
+            .ok()
+            .unwrap_or("0".to_string());
+        if allow_executable != "1" {
+            return Err(CredentialsError::from_msg(
+                false,
+                "executables need to be explicitly allowed (set GOOGLE_EXTERNAL_ACCOUNT_ALLOW_EXECUTABLES to '1') to run",
+            ));
+        }
+
+        let (command, args) = Self::split_command(command);
```
sai-sunder-s: can we split the command once and store it?

src/auth/src/credentials/external_account_sources/executable_sourced.rs
sai-sunder-s: @lsirac can you answer this

sai-sunder-s (changes requested):

src/auth/src/credentials/external_account.rs
alvarowolfx: to instantiate the `TokenCache` with the `new` method, it requires the generic argument type to be `'static`

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: yeah, one is to map the `tokio_timeout` error and the other is to map the `Commmand::output` error

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: good point, gonna add

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: this is an issue because we are reusing the ExternalAccountConfig struct as both the struct representing the external account config file and internal config, which makes hard to have computed values. This PR will improve this and then I'll update it here.

https://github.com/googleapis/google-cloud-rust/pull/2406

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: Same issue as on the timeout, we are reusing the ExternalAccountConfig struct as both the struct representing the external account config file and internal config, which makes hard to have computed values. Will merge the PR and improve things here.

Ref: https://github.com/googleapis/google-cloud-rust/pull/2406

alvarowolfx (commented):
answering some points on computing values. Will be improved by PR  https://github.com/googleapis/google-cloud-rust/pull/2406

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: done

src/auth/src/credentials/external_account_sources/executable_sourced.rs
alvarowolfx: done

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,500 @@
...
+use tokio::{process::Command, time::timeout as tokio_timeout};
+
+#[derive(Serialize, Deserialize, Debug, Clone)]
+pub(crate) struct ExecutableSourcedCredentials {
+    command: String,
+    args: Vec<String>,
+    timeout: Duration,
+    output_file: Option<String>,
+}
+/// Executable command should adere to this format.
```
coryan, line 36: nit: missing blank line.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,500 @@
...
+/// Format is documented on [executable source credentials].
+///
+/// [executable sourced credentials]: https://google.aip.dev/auth/4117#determining-the-subject-token-in-executable-sourced-credentials
+#[derive(Serialize, Deserialize, Debug, Clone, PartialEq)]
+struct ExecutableResponse {
+    version: i32,
+    success: bool,
+    token_type: String,
+    /// 3rd party subject token expiration time in seconds (unix epoch time).
+    expiration_time: Option<i64>,
```
coryan, line 46: I would prefer usng `std::time::Duration` instead of "i64 which is interpreted as time in seconds".  

If it is not a public API, something like [UtcDateTime](https://docs.rs/time/latest/time/struct.UtcDateTime.html) is even more appropriate. Unfortunately that is not a stable API, so cannot go in public APIs.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,500 @@
...
+    id_token: Option<String>,
+    saml_response: Option<String>,
+    code: Option<String>,
+    message: Option<String>,
+}
+/// Represents an error when executing a command line tool to fetch subject tokens.
```
coryan, line 53: nit: blank line before this comment.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
```diff
@@ -0,0 +1,502 @@
...
+    }
+
+    /// Executable and output file should follow the format defined in [ExecutableResponse] struct.
+    /// This tries to parse the output in the given format and extract just the OAuth token.
+    fn parse_token(output: String) -> Result<String> {
+        let res = serde_json::from_str::<ExecutableResponse>(output.as_str())
+            .map_err(|e| CredentialsError::from_source(false, e))?;
+
+        if !res.success {
+            return Err(ExecutionError::from_executable_response(res));
```
sai-sunder-s, line 196: why do we throw ExecutionError here but not when execution was not successful?

alvarowolfx: the `code` and `message` on the `ExecutableResponse` are only required for situations where the execution returns `success = false`.

src/auth/src/credentials/external_account_sources/executable_sourced.rs
lsirac: This is for security reasons e.g. considering SaaS vendors and shared machines. Some SaaS vendors allow uploading JSON key files. 

Malicious actors can exploit the file format of 3PI credentials to access resources on the shared machines running the SaaS software that should otherwise be inaccessible.

PR #2357, opened 2025-06-02
feat(storage-control): add new PRPCs

dbolduc:
This is an easy command to see what is missing from the handwritten surface:

```
git grep -ho "super::builder::storage_control::[A-Za-z]*[^ :]" | sort | uniq -c | grep 2
```
```
      2 super::builder::storage_control::GetFolderIntelligenceConfig
      2 super::builder::storage_control::GetOrganizationIntelligenceConfig
      2 super::builder::storage_control::GetProjectIntelligenceConfig
      2 super::builder::storage_control::UpdateFolderIntelligenceConfig
      2 super::builder::storage_control::UpdateOrganizationIntelligenceConfig
      2 super::builder::storage_control::UpdateProjectIntelligenceConfig
```

We expect the builders to be referenced three times: twice in generated code, and once from the veneer client.


src/storage-control/tests/mocking.rs
```diff
@@ -304,6 +364,12 @@ mod test {
             resume_anywhere_cache,
             get_anywhere_cache,
             list_anywhere_caches,
+            get_folder_intelligence_config,
```
dbolduc, line 367: Observation: we could also write the mocking test inside this macro

coryan: Or a similar macro with the same input.

PR #2361, opened 2025-06-03
fix(storage)!: add required parameters to initial builder

suzmue:
For #2103, #2040

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -111,18 +111,20 @@ impl Storage {
     /// # use google_cloud_storage::client::Storage;
     /// async fn example(client: &Storage) -> gax::Result<()> {
     ///     let response = client
-    ///         .insert_object()
-    ///         .set_bucket("projects/_/buckets/my-bucket")
-    ///         .set_object("my-object")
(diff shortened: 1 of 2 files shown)


src/storage/src/client.rs
```diff
@@ -132,17 +134,19 @@ impl Storage {
...
+    ///         .read_object("projects/_/buckets/my-bucket", "my-object")
     ///         .send()
     ///         .await?;
     ///     println!("object contents={contents:?}");
     ///     Ok(())
     /// }
     /// ```
-    pub fn read_object(&self) -> ReadObject {
-        ReadObject::new(self.inner.clone())
+    pub fn read_object<B, O>(&self, bucket: B, object: O) -> ReadObject
```
dbolduc, line 155: non-blocking: Might be nice to document the required parameters like we used to do in storage-control:

https://github.com/googleapis/google-cloud-rust/blob/dbe7a0e152fe0cbc2b4f0b3b9bed14f375670a42/src/storage-control/src/client.rs#L197-L200

dbolduc (dismissed):

src/storage/src/client.rs
suzmue: Done!

PR #2383, opened 2025-06-04
feat(auth): allow customers to provide a SubjectTokenProvider impl

alvarowolfx:
Draft to showcase one idea on how to allow customers to provide their own implementation of SubjectTokenProvider.


src/auth/src/credentials/external_account.rs
```diff
@@ -26,13 +26,31 @@ use crate::{BuildResult, Result};
...
+use std::future::Future;
 use std::sync::Arc;
 use tokio::time::{Duration, Instant};
 
-#[async_trait::async_trait]
 pub trait SubjectTokenProvider: std::fmt::Debug + Send + Sync {
-    /// Generate subject token that will be used on STS exchange.
-    async fn subject_token(&self) -> Result<String>;
+    fn subject_token(&self) -> impl Future<Output = Result<String>> + Send;
+}
```
coryan, line 37: What is the advantage of using a trait with a single function vs. a thing that implements the `AsyncFn` trait?

alvarowolfx: TIL `AsyncFn`. Gonna check it out how to use that. But to your point, I was just trying to follow the same pattern from the repo with things like `CredentialsProvider`, but this one in particular has two methods, so makes sense to be a trait.

coryan: AsyncFn is a Rust 2024 feature.... I feat it might not work if the caller is using Rust Edition 2021, so make sure it does, otherwise the trait is a better idea.

alvarowolfx: trying to use `AsyncFn` and as it's optional on the `external_account::Builder` , I'm not finding a way to use without `dyn` and `AsyncFn` is not `dyn` compatible. Any tips ?

src/auth/src/credentials/external_account.rs
```diff
@@ -241,6 +267,16 @@ impl Builder {
...
 
+    /// bring your own custom implementation of
+    /// SubjectTokenProvider for OIDC/SAML credentials.
+    pub fn with_subject_token_provider<T: SubjectTokenProvider + 'static>(
+        mut self,
+        subject_token_provider: T,
+    ) -> Self {
+        self.subject_token_provider = Some(Box::new(subject_token_provider));
+        self
+    }
```
sai-sunder-s, line 363: If we are taking subject token provider as an input in the builder, we need to make some more changes as well. It becomes tricky now.  There is also a credential source in the config they initialized the builder with.

src/auth/src/credentials/external_account.rs
sai-sunder-s: Says here that asyncfn is nightly-only experimental API https://doc.rust-lang.org/std/ops/trait.AsyncFn.html

src/auth/src/credentials/external_account.rs
alvarowolfx: when setting the subject token provider via the `Bulder`, it overrides the credential source from the config

src/auth/src/credentials/external_account.rs
coryan: > Says here that asyncfn is nightly-only experimental API https://doc.rust-lang.org/std/ops/trait.AsyncFn.html

That is not quite right.  `AsyncFn` is stable:

https://github.com/rust-lang/rust/pull/132706

Manually implementing the `AsyncFn` trait is not stable.  That is,  you cannot say `impl AsyncFn for MyType ...` because the types and methods of the trait are unstable. But you can assume that `AsyncFn` will be around, and that you can call such functions.

As to the `dyn-compatible` problems: you could solve that with a private trait:

```rust
struct Foo<T> {
  function: T
}
#[async_trait]
trait MyAsyncFn {
  async fn call() -> Blah;
}
#[async_trait]
impl MyAsyncFn for Foo<T> where T: AsyncFn<....> {
  async fn call() -> Blah { function().await }
}
```

But overall it seems that `AsyncFn` is more trouble than it is worth.

src/auth/src/credentials/external_account.rs
```diff
@@ -25,13 +26,31 @@ use crate::{BuildResult, Result};
...
 use std::sync::Arc;
 use tokio::time::{Duration, Instant};
 
-#[async_trait::async_trait]
-pub(crate) trait SubjectTokenProvider: std::fmt::Debug + Send + Sync {
-    /// Generate subject token that will be used on STS exchange.
-    async fn subject_token(&self) -> Result<String>;
+pub trait SubjectTokenProvider: std::fmt::Debug + Send + Sync {
+    fn subject_token(&self) -> impl Future<Output = Result<String>> + Send;
+}
```
coryan, line 37: So to override this I need to implement my own client, error detection, parsing, etc.?  Bummer.

src/auth/src/credentials/external_account.rs
sai-sunder-s: Hmm that maybe ok. When implementing this change lets make sure to document that clearly.

sai-sunder-s: Another way could be to introduce `new(subject_token_provider)`. But that means we add builder methods to provide all other fields like impersonation url, provider ppol, token url etc.

src/auth/src/credentials/external_account.rs
sai-sunder-s: I see. Thank you for the context!.

src/auth/src/credentials/external_account.rs
alvarowolfx: I think the main point of this custom subject token provider is indeed for the customer own all the stack around fetching the subject token, specially in complex scenarios like the ones we discussed where the customers needs to use a custom openssl version and other details. I think we are also evaluating how to allow injecting custom headers and/or provide a custom http client cc @sai-sunder-s

sai-sunder-s: Yeah this is for people with advanced use case and know what they are doing. It is just complete freedom to do whatever they want. Yes that means they have to handle error detection and parsing as well but that should be ok. For example, they know best how to handle a kerberos error.

coryan:
Can we close this?

alvarowolfx:
superseded by #2541

Issue #2371, opened 2025-06-03
Use custom KeyAes256 type for Storage


suzmue:
Follow up from PR #2354, we currently allow any bytes::Bytes, but we would like to make a custom type to allow verification that the key is the right size.

```rust
struct KeyAes256 {
  key: bytes::Bytes
};

impl KeyAes256 {
  pub fn from(key: &[u8]) -> Result<Self> { /* validate length */
}

impl std::convert::From<KeyAes256> for control::model::CommonObjectRequestParams {
  fn from(value: KeyAes256) -> Self {
    // Set the algorithm, and SHA256 and key computed from `value.key`.
  }
}

// Now `KeyAes256` implements `Into<CommonObjectRequestParams>` and one could write:
//
client.read_object().set_common_object_request_params(KeyAes256::from(blah blah )?).send().await?
```

PR #2388, opened 2025-06-05
fix(storage): use new type KeyAes256 for customer supplied encryption key

suzmue:
For #2371

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -563,50 +563,67 @@ impl ReadObject {
 }
 
 #[derive(Debug)]
-/// KeyAes256 is the encryption key used with the Customer-Supplied Encryption Keys
-/// feature. In raw bytes format (not base64-encoded).
+/// KeyAes256 represents an AES-256 encryption key used with the
+/// Customer-Supplied Encryption Keys (CSEK) feature.
 ///
-/// Example:
+/// This key must be exactly 32 bytes in length and should be provided in its
+/// raw (unencoded) byte format.
+///
+/// # Examples
+///
+/// Creating a `KeyAes256` instance from a valid byte slice:
 /// ```
-/// # use google_cloud_storage::client::KeyAes256;
-/// let key: &[u8] = &[b'a'; 32];
-/// let key = KeyAes256::try_from(key);
+/// # use google_cloud_storage::client::{KeyAes256, KeyAes256Error};
+/// let raw_key_bytes: [u8; 32] = [0x42; 32]; // Example 32-byte key
+/// let key_aes_256 = KeyAes256::try_from(&raw_key_bytes)?;
+/// # Ok::<(), KeyAes256Error>(())
+/// ```
+///
+/// Handling an error for an invalid key length:
+/// ```
+/// # use google_cloud_storage::client::{KeyAes256, KeyAes256Error};
+/// let invalid_key_bytes: &[u8] = b"too_short_key"; // Less than 32 bytes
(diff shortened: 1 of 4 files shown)


src/storage/src/client.rs
```diff
@@ -539,6 +563,35 @@ impl ReadObject {
     }
 }
 
+#[derive(Debug)]
+pub struct KeyAes256 {
```
coryan: This is a public API, it needs a comment block.

src/storage/src/client.rs
```diff
@@ -539,6 +563,35 @@ impl ReadObject {
...
+    key: bytes::Bytes,
```
coryan: We may be able to use `[u8; 32]` which seems simpler?

src/storage/src/client.rs
```diff
@@ -539,6 +563,35 @@ impl ReadObject {
...
+}
+
+impl KeyAes256 {
+    pub fn from(mut key: &[u8]) -> Result<Self> {
```
coryan: I think we should use a specific type, like we do for `AnyError` or `DurationError`.  It is also a public API, it needs comments.

src/storage/src/client.rs
suzmue: Updated to use this. LMK what you think.

src/storage/src/client.rs

src/storage/src/client.rs

src/storage/src/client.rs
```diff
@@ -539,6 +562,86 @@ impl ReadObject {
...
+    /// This conversion will succeed only if the input slice is exactly 32 bytes long.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_storage::client::{KeyAes256, KeyAes256Error};
+    /// let raw_key_bytes: [u8; 32] = [0x42; 32]; // Example 32-byte key
+    /// let key_aes_256 = KeyAes256::try_from(&raw_key_bytes)?;
+    /// # Ok::<(), KeyAes256Error>(())
+    /// ```
+    pub fn try_from(key: &[u8]) -> std::result::Result<Self, KeyAes256Error> {
```
suzmue: @coryan I renamed this to `try_from`, since it seemed more accurate. WDYT?

Will actually look into implementing the TryFrom trait as well.

suzmue: Not having a lot of success with TryFrom trait to accept all the different types, especially for [u8; 32].

src/storage/src/client.rs
```diff
@@ -539,6 +562,86 @@ impl ReadObject {
     }
 }
 
+#[derive(Debug)]
```
coryan, line 567: We will probably want, `Clone`, and `PartialEq` here too.

src/storage/src/client.rs
coryan: You would need to implement `Tryfrom` in addition to writing the function, no? Maybe:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ed2f686f386c4fedbfe3b9f499231c73

coryan: FWIW, it is not weird to have both `new()` and `try_from()`.

suzmue: > You would need to implement Tryfrom in addition to writing the function, no? 

I was inlining the new function effectively.

I dropped try_from for this PR, since I did not see a scenario that it would be easier to use try_from or try_into instead of just doing KeyAes256::new().

If there is a scenario where that would be better, we can definitely still add it (much easier than to take it away :) )

coryan (approved):

Issue #2376, opened 2025-06-03
Detect if two variants of a `oneof` field are set in JSON inputs


coryan:
When deserializing a `oneof` from JSON we need to detect if there are multiple variants set. Ignoring this could lead to silent data loss, or at least confusing behavior.

- [ ] `Required.Proto3.JsonInput.OneofFieldDuplicate` # Should have failed to parse, but didn't.
- [ ] `Required.Proto3.JsonInput.OneofFieldNullFirst.JsonOutput` # JSON output we received from test was unparseable.

----

```
ERROR, test=Required.Proto3.JsonInput.OneofFieldDuplicate: Should have failed to parse, but didn't., request=goo.gle/debugonly   json_payload: "{\"oneofUint32\": 1, \"oneofString\": \"test\"}" requested_output_format: JSON message_type: "protobuf_test_messages.proto3.TestAllTypesProto3" test_category: JSON_TEST, response=goo.gle/debugonly   json_payload: "{\"oneofUint32\":1,\"oneofString\":\"test\"}"
ERROR, test=Required.Proto3.JsonInput.OneofFieldNullFirst.JsonOutput: JSON output we received from test was unparseable., request=goo.gle/debugonly   json_payload: "{\"oneofUint32\": null, \"oneofString\": \"test\"}" requested_output_format: JSON message_type: "protobuf_test_messages.proto3.TestAllTypesProto3" test_category: JSON_TEST, response=goo.gle/debugonly   json_payload: "{\"oneofUint32\":0,\"oneofString\":\"test\"}"
```

coryan:
The fixes for this are going to take multiple steps, which I think deserve at least an outline.

1. I have improved testing to include messages with all the primitive types as required, optional, repeated, and map fields.
2. I have improved testing to include messages with enum types as required, optional, repeated, and map fields.
3. I have improved testing to include message types as optional (by default and explicitly), repeated, and map fields.
4. I have improved testing to include a well-known message type as optional (by default and explicitly), repeated, and map fields that serializes to strings.
5. I have improved testing to include the `google.protobuf.Value` type as optional (by default and explicitly), repeated, and map fields.
6. For all these messages, I will change the generator to emit a helper type with hand-crafted serialization
7. Initially only the body (no fields) will be serialized.
8. I will slowly add different field types to this serialization, including unit tests.
9. I will start with fields that do not require `serde_with` and `serde_as()` mapping.
[...]

PR #2389, opened 2025-06-06
test(wkt): integration test for `enum` fieds

coryan:
Motivated by #2376


src/wkt/tests/message_with_enum.rs
```diff
@@ -0,0 +1,116 @@
...
+        Ok(())
+    }
+
+    #[test]
+    fn test_repeated() -> Result {
+        let msg = MessageWithEnum::new().set_repeated([TestEnum::Red, TestEnum::Unspecified]);
+        let got = serde_json::to_value(&msg)?;
+        let want = json!({"repeated": [1, 0]});
+        assert_eq!(want, got);
+        let roundtrip = serde_json::from_value(serde_json::to_value(&msg)?)?;
```
dbolduc, line 76: nit: can we use `got`?

src/wkt/tests/message_with_enum.rs
```diff
@@ -0,0 +1,116 @@
...
+    fn test_map() -> Result {
+        let input = json!({"map": { "favorite": "RED", "not-so-much": "GREEN"}});
+        let got = serde_json::from_value::<MessageWithEnum>(input.clone())?;
+        let want = MessageWithEnum::new().set_map([
+            ("favorite", TestEnum::Red),
+            ("not-so-much", TestEnum::Green),
+        ]);
+        assert_eq!(want, got);
+        let roundtrip = serde_json::from_value(serde_json::to_value(&got)?)?;
+        assert_eq!(got, roundtrip);
```
dbolduc, line 102: nit: Just do `to_value(&got)` and compare against `input`? I assume that is why we cloned `input` above.

dbolduc (approved):

src/wkt/tests/message_with_enum.rs
coryan: Cannot do that here.  to `to_value(&got) == json!({"map: { "favorite": 1, "not-so-much": 2 })`.

I will  remove the `clone()`.

The tests are written as they are because in general JSON -> `type` -> JSON does not get you the same thing (that is true from most primitive types too), while `type` -> JSON -> `type` does give you the same thing.

src/wkt/tests/message_with_enum.rs
coryan: I could, but I would rather not think about which cases work and which do not.

src/wkt/tests/message_with_enum.rs
coryan: Done.

dbolduc (approved):

PR #2393, opened 2025-06-07
test: verify synthetic fields are skipped

coryan:
Synthetic fields need to be skipped during serialization. We had no
explicit tests for this functionality.

Motivated by #2376

### DIFF
--- a/src/integration-tests/tests/serialization.rs
+++ b/src/integration-tests/tests/serialization.rs
@@ -17,6 +17,23 @@
 #[cfg(test)]
 mod serialization {
     use anyhow::Result;
+    use serde_json::json;
+
+    // The generator introduces synthetic fields for some OpenAPI-based
+    // services. Those are skipped during serialization, we need a test to
+    // verify this is the case.
+    #[test]
+    fn skip_synthetic_fields() -> Result<()> {
+        use smo::model::{CreateSecretRequest, Secret};
+        let input = CreateSecretRequest::new()
+            .set_project("a-synthetic-field")
+            .set_secret_id("another-synthetic-field")
+            .set_request_body(Secret::new().set_labels([("test-name", "test-value")]));
+        let got = serde_json::to_value(input)?;
+        let want = json!({"requestBody": {"labels": {"test-name": "test-value"}}});
+        assert_eq!(got, want);
+        Ok(())
+    }
 
     #[test]
     fn multiple_serde_attributes() -> Result<()> {


src/integration-tests/tests/serialization.rs
```diff
@@ -17,6 +17,23 @@
...
+    use serde_json::json;
+
+    // The generator introduces synthetic fields for some OpenAPI-based
+    // services. Those are skipped during serialization, we need a test to
+    // verify this is the case.
+    #[test]
+    fn skip_synthetic_fields() -> Result<()> {
+        use smo::model::{CreateSecretRequest, Secret};
+        let input = CreateSecretRequest::new()
+            .set_project("a-synthetic-field")
```
dbolduc, line 29: I am late to the party and confused. I read: https://github.com/googleapis/google-cloud-rust/blob/56d472fb0048dc33dd0e3fcf7cd219b03ddf08e7/generator/internal/api/model.go#L505-L508

So these synthetic fields are part of the path we would use to make the RPC, not part the request? That makes sense. But don't things work the same way with gRPC transcoding?

Hmm, I thought fields in the path need not be repeated in the body (even when the body is `*`). But I don't see that stated in https://google.aip.dev/127.

dbolduc: Ah, 2.4.2 in go/actools-regapic-grpc-transcoding says fields from the path should not be included in the body, but this is optional. So what we are doing is fine.

Still the way this is unified feels weird to me. Seems like we could serialize the fields, even if they are synthetic. But then in the `transport.rs` layer, we could `std::mem::take` the fields out of the request for use in the path.

Thoughts?

dbolduc (approved):
Food for thought, but nothing blocking this PR.

It sort of affects my current work which involves building the paths.

FTR, my plan is to keep taking references to fields used in the path and copying them into the path. I am not gonna block on this.

src/integration-tests/tests/serialization.rs
coryan: > Thoughts?

OpenAPI needs a lot of work in this area.  

> Seems like we could serialize the fields, even if they are synthetic.

I do not think we can, because (sometimes) they do not match any existing field in the original request.

Consider this:

https://github.com/googleapis/google-cloud-rust/blob/56d472fb0048dc33dd0e3fcf7cd219b03ddf08e7/src/generated/openapi-validation/src/model.rs#L3563-L3590

The synthetic fields are `project`, `location` and `secret_id`. That is all the info we have in the source specification:

https://github.com/googleapis/google-cloud-rust/blob/56d472fb0048dc33dd0e3fcf7cd219b03ddf08e7/testdata/secretmanager_openapi_v1.json#L462-L488

If we wanted to include them in the body we should be sending `secret_id` and `parent`:

https://github.com/googleapis/googleapis/blob/1804a603281707a1f0e6fff27851cae115ac3c8b/google/cloud/secretmanager/v1/service.proto#L65-L68

https://github.com/googleapis/googleapis/blob/1804a603281707a1f0e6fff27851cae115ac3c8b/google/cloud/secretmanager/v1/service.proto#L322-L343

But we do not even know the original name of the (decomposed) `project` and `location` fields.

PR #2404, opened 2025-06-09
feat(storage): support read ranges in Storage::read_object

### LINKED ISSUE #2399: Handle ranged reads in Storage::read_object
API supports ranged reads via headers: http://cloud/storage/docs/json_api/v1/parameters#range. Support this from read_object

suzmue:
Support ranged reads by exposing the `read_offset` and `read_limit` fields. These are then converted to a [Range](http://cloud.google.com/storage/docs/json_api/v1/parameters#range) header.

Missed the Range header being available on first pass for #2103, this is a follow up to support with the simple reads.

Fixes #2399


src/storage/src/client.rs
```diff
@@ -459,6 +459,115 @@ impl ReadObject {
         self
     }
 
+    /// The offset for the first byte to return in the read, relative to
+    /// the start of the object.
+    ///
+    /// A negative `read_offset` value will be interpreted as the number of bytes
+    /// back from the end of the object to be returned. For example, if an object's
+    /// length is 15 bytes, a ReadObjectRequest with `read_offset` = -5 and
+    /// `read_limit` = 3 would return bytes 10 through 12 of the object. Requesting
```
coryan: This works for gRPC, but not for JSON... Maybe edit the comment?

src/storage/src/client.rs
```diff
@@ -559,11 +668,40 @@ impl ReadObject {
             self.request.common_object_request_params,
         );
 
+        // Apply "range" header for read limits and offsets.
+        let builder = match (self.request.read_offset, self.request.read_limit) {
```
coryan, line 665: This is so much cleaner than the equivalent code in C++, very nice.

src/storage/src/client.rs
```diff
@@ -559,11 +668,40 @@ impl ReadObject {
...
         Ok(builder)
     }
 }
 
+/// Represents an error that can occur with invalid range is specified.
+#[derive(thiserror::Error, Debug, PartialEq)]
+#[non_exhaustive]
+enum RangeError {
+    /// The provided read limit was negative.
+    #[error("Read limit was negative, expected non-negative value.")]
```
coryan: nit: I read somewhere that error messages should be lowercase ...

suzmue: Done!

src/storage/src/client.rs
suzmue: Ah nice catch. Edited.

PR #2406, opened 2025-06-10
cleanup(auth): separate ExternalAccountFile from internal configuration

alvarowolfx:
Previously, `ExternalAccountConfig` and `CredentialSource` were being used to mirror the external account file format and also for internal configuration usage. But this introduces challenges to have pre-computed and/or cache values, so it's better to treat them as separated structs. 

This PR introduces `ExternalAccountFile` and `CredentialSourceFile` to represent the [AIP-4117 file](https://google.aip.dev/auth/4117#configuration-file-generation-and-usage), with `ExternalAccountConfig` and `CredentialSource` being the internal configuration used across implementations.

### DIFF
--- a/src/auth/src/credentials/external_account_sources/url_sourced.rs
+++ b/src/auth/src/credentials/external_account_sources/url_sourced.rs
@@ -38,12 +38,9 @@ impl UrlSourcedCredentials {
         headers: Option<HashMap<String, String>>,
         format_source: Option<CredentialSourceFormat>,
     ) -> Self {
-        let mut format = "text".to_string();
-        let mut subject_token_field_name = "".to_string();
-        if let Some(f) = format_source {
-            format = f.format_type;
-            subject_token_field_name = f.subject_token_field_name;
-        }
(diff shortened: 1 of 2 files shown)


src/auth/src/credentials/external_account_sources/url_sourced.rs
```diff
@@ -27,11 +27,34 @@ use crate::{
...
+        url: String,
+        headers: Option<HashMap<String, String>>,
+        format_source: Option<CredentialSourceFormat>,
+    ) -> Self {
+        let mut format = "text".to_string();
+        let mut subject_token_field_name = "".to_string();
+        if let Some(f) = format_source {
+            format = f.format_type;
+            subject_token_field_name = f.subject_token_field_name;
+        }
```
coryan: I am sure that are other ways to simplify this:

```suggestion
        let (format, subject_token_field_name) = format_source
            .map(|f| (f.format_source, f.subject_token_field_name))
            .unwrap_or(("text".to_string(), String::new()));
```

PR #2412, opened 2025-06-11
impl(generator): handle messages and simple maps

coryan:
Messages and some maps do not require serde transformations.

Part of the work for #2376

### DIFF
--- a/src/wkt/tests/message_with_bool.rs
+++ b/src/wkt/tests/message_with_bool.rs
@@ -24,13 +24,9 @@ mod test {
     #[test_case(MessageWithBool::new().set_singular(true), json!({"singular": true}))]
     #[test_case(MessageWithBool::new().set_optional(false), json!({"optional": false}))]
     #[test_case(MessageWithBool::new().set_or_clear_optional(None::<bool>), json!({}))]
-    #[test_case(MessageWithBool::new().set_repeated([false; 0]), json!({}))]
     #[test_case(MessageWithBool::new().set_repeated([true, true, false]), json!({"repeated": [true, true, false]}))]
-    #[test_case(MessageWithBool::new().set_map_key([(true, ""); 0]), json!({}))]
     #[test_case(MessageWithBool::new().set_map_key([(true, "trueValue"), (false, "falseValue")]), json!({"mapKey": {"true": "trueValue", "false": "falseValue"}}))]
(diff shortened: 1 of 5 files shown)


src/wkt/tests/message_with_bool.rs
```diff
@@ -26,6 +26,12 @@ mod test {
     #[test_case(MessageWithBool::new().set_or_clear_optional(None::<bool>), json!({}))]
     #[test_case(MessageWithBool::new().set_repeated([false; 0]), json!({}))]
     #[test_case(MessageWithBool::new().set_repeated([true, true, false]), json!({"repeated": [true, true, false]}))]
+    #[test_case(MessageWithBool::new().set_map_key([(true, ""); 0]), json!({}))]
```
dbolduc: I will plead my case a final time, then drop it.

This is not a different test case. The inputs are identical to other test cases.

If you want to check that providing empty input to a default message results in a default message, write that test elsewhere.

generator/internal/rust/annotate.go
```diff
@@ -690,6 +690,26 @@ func (c *codec) annotateOneOf(oneof *api.OneOf, message *api.Message, state *api
...
+		if msg, ok := state.MessageByID[field.TypezID]; ok && msg.IsMap {
+			if len(msg.Fields) != 2 {
+				slog.Error("expected exactly two fields for map message", "field ID", field.ID, "map ID", field.TypezID)
+				return false
+			}
+			key := c.requiresSerdeAs(msg.Fields[0], state)
+			value := c.requiresSerdeAs(msg.Fields[1], state)
+			return key || value
+		}
+		return field.TypezID == ".google.protobuf.Value"
```
dbolduc, line 707: spoiler? what's up with this?

dbolduc (dismissed):

src/wkt/tests/message_with_bool.rs
coryan: Removed here and elsewhere.

generator/internal/rust/annotate.go
coryan: `google.protobuf.Value` is special because of the `null` stuff. I think it has no effect at the moment, and I will revert this change over the next few PRs. Apologies for the churn.

dbolduc: > `google.protobuf.Value` is special because of the `null` stuff.

Ah right.

> Apologies for the churn.

No worries at all. So far so good.

PR #2414, opened 2025-06-11
impl(generator): serde for singular fields

coryan:
This implements serialization and deserialization for singular fields
that require mapping. It handles skipping the fields and deserializing
`null` to the default value.

Part of the work for #2376


generator/internal/rust/annotate.go
```diff
@@ -710,7 +711,49 @@ func (c *codec) requiresSerdeAs(field *api.Field, state *api.APIState) bool {
...
+				return ""
+			}
+			key := c.serdeAs(msg.Fields[0], state)
+			value := c.serdeAs(msg.Fields[1], state)
+			return fmt.Sprintf("std::collections::HashMap<%s, %s>", key, value)
+		}
+		//		if field.TypezID == ".google.protobuf.Value" {
+		//			return "wkt::internal::OptionalValue"
+		//		}
+		formatter = "serde_with::Same"
```
dbolduc, line 744: TIL.

Q: do we need `RequiresSerdeAs` vs. always using `As` + `serde_with::Same` as the formatter?

dbolduc: eh, nevermind, I see the extra code needed for `As`. I understand why we do not want that everywhere.

src/wkt/tests/message_with_f32.rs
```diff
@@ -20,18 +20,63 @@ mod test {
     type Result = anyhow::Result<()>;
 
     #[test_case(MessageWithF32::new(), json!({}))]
+    #[test_case(MessageWithF32::new().set_singular(0.0), json!({}))]
+    #[test_case(MessageWithF32::new().set_singular(1.5), json!({"singular": 1.5}))]
+    #[test_case(MessageWithF32::new().set_singular(f32::INFINITY), json!({"singular": "Infinity"}))]
+    #[test_case(MessageWithF32::new().set_singular(-f32::INFINITY), json!({"singular": "-Infinity"}); "singular minus inf")]
+    #[test_case(MessageWithF32::new().set_singular(f32::NAN), json!({"singular": "NaN"}))]
+    #[test_case(MessageWithF32::new().set_optional(0.0), json!({"optional": 0.0}))]
+    #[test_case(MessageWithF32::new().set_or_clear_optional(None::<f32>), json!({}))]
```
dbolduc, line 29: I'll admit that it is irrational that this bothers me...

src/wkt/tests/message_with_f32.rs
```diff
@@ -20,18 +20,63 @@ mod test {
     type Result = anyhow::Result<()>;
 
     #[test_case(MessageWithF32::new(), json!({}))]
+    #[test_case(MessageWithF32::new().set_singular(0.0), json!({}))]
```
dbolduc, line 23: ... but this doesn't.

dbolduc: I think I know what it is. I think it is the difference in C++ between:

```cc
int x;
```
and
```cc
std::optional<int> y;
```

I cannot have an uninitialized `optional<>`, but I could have an uninitialized `int`.

dbolduc: I think this is a trauma response.

generator/internal/rust/annotate.go
coryan: FWIW, this evolves in the next few PRs.

PR #2415, opened 2025-06-11
impl(storage): use the WriteObjectRequest for insert_object

suzmue:
Generates the WriteObjectRequest type to use for storing data for the insert_object request.

For #2040


src/storage/src/client.rs
```diff
@@ -318,16 +320,19 @@ impl InsertObject {
...
-        let bucket_id = bucket
-            .as_str()
-            .strip_prefix("projects/_/buckets/")
-            .ok_or_else(|| {
-                Error::binding(format!(
-                    "malformed bucket name, it must start with `projects/_/buckets/`: {bucket}"
-                ))
-            })?;
-        let object: String = self.object;
+        let resource = match self.request.first_message {
```
suzmue, line 325: Using the helper functions instead doesn't allow us to take ownership of the data and requires an additional clone (`let resource = self.request.write_object_spec().unwrap().resource.clone().unwrap();`)

src/storage/src/client.rs
```diff
@@ -318,16 +320,19 @@ impl InsertObject {
...
+            Some(control::model::write_object_request::FirstMessage::WriteObjectSpec(spec)) => {
```
coryan: Consider a `use control::model::write_object_request::*;` or similar to make this more readable?

coryan (approved):

PR #2417, opened 2025-06-12
chore: update googleapis SHA circa 2025-06-12

coryan:


dbolduc (approved):
Was concerned to see changes to storage/v2, but it was not a new RPC. So all good.

PR #2420, opened 2025-06-12
fix(auth): exec source cred not being detected properly

alvarowolfx:
Executable sourced credential was not being detected because of the enum order. Changed and added test to check that.

### DIFF
--- a/src/auth/src/credentials/external_account.rs
+++ b/src/auth/src/credentials/external_account.rs
@@ -459,7 +459,7 @@ mod test {
             CredentialSource::Executable(source) => {
                 assert_eq!(source.command, "cat");
                 assert_eq!(source.args, vec!["/some/file"]);
-                assert_eq!(source.output_file, Some("/some/file".to_string()));
+                assert_eq!(source.output_file.as_deref(), Some("/some/file"));
                 assert_eq!(source.timeout, Duration::from_secs(5));
             }
             _ => {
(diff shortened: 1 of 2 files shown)


src/auth/src/credentials/external_account.rs
```diff
@@ -433,4 +433,38 @@ mod test {
...
+        let file: ExternalAccountFile =
+            serde_json::from_value(contents).expect("failed to parse external account config");
+        let config: ExternalAccountConfig = file.into();
+        let source = config.credential_source;
+
+        match source {
+            CredentialSource::Executable(source) => {
+                assert_eq!(source.command, "cat");
+                assert_eq!(source.args, vec!["/some/file"]);
+                assert_eq!(source.output_file, Some("/some/file".to_string()));
```
coryan: nit: you can typically write:

```suggestion
                assert_eq!(source.output_file.as_deref(), Some("/some/file"));
```
https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=ef239cbcdc26ce1b6be6a0c9cc17d2b2

Should be microscopically  faster as it performs fewer allocations.

coryan (approved):

src/auth/src/credentials/external_account.rs
alvarowolfx: nice, now I want to find all places where I use to_string for that

PR #2421, opened 2025-06-12
test(auth): integration tests for exec sourced external account

alvarowolfx:
Depends on https://github.com/googleapis/google-cloud-rust/pull/2420

### DIFF
--- a/src/auth/integration-tests/src/lib.rs
+++ b/src/auth/integration-tests/src/lib.rs
@@ -123,13 +123,7 @@ pub async fn api_key() -> anyhow::Result<()> {
 pub async fn workload_identity_provider_url_sourced() -> anyhow::Result<()> {
     let project = std::env::var("GOOGLE_CLOUD_PROJECT").expect("GOOGLE_CLOUD_PROJECT not set");
     let audience = get_oidc_audience();
-    let service_account = get_byoid_service_account();
-    let client_email = match service_account.get("client_email") {
-        Some(serde_json::Value::String(v)) => v.clone(),
-        None | Some(_) => {
-            panic!("missing `client_email` string in service account: {service_account:?}")
-        }
-    };
+    let (service_account, client_email) = get_byoid_service_account_and_email();
 
     let id_token = generate_id_token(audience.clone(), client_email, service_account).await?;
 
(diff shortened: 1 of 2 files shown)


src/auth/integration-tests/src/lib.rs
```diff
@@ -183,6 +177,60 @@ pub async fn workload_identity_provider_url_sourced() -> anyhow::Result<()> {
...
+      "type": "external_account",
+      "audience": audience,
+      "subject_token_type": "urn:ietf:params:oauth:token-type:jwt",
+      "token_url": "https://sts.googleapis.com/v1/token",
+      "credential_source": {
+        "executable": {
+            "command": format!("cat {path}"),
+            "output_file": path,
+        },
+      }
```
coryan: I think you want the program to generate the file, otherwise I am not sure the test matches its title?

alvarowolfx: the [AIP for Executable Sourced creds](https://google.aip.dev/auth/4117#determining-the-subject-token-in-executable-sourced-credentials) doesn't requires that the target executable caches the output into a given path defined on `output_file`, that's more of a optional optimization that customer can implement.  The executable needs primarily to output the response via `stdout`.

With that said, I can change the test to not inform the `output_file`, which will indeed executes the `cat` program, but that can be more environment dependent (like `cat` not being available on Windows) and that's why I ended up choosing to inform the output file, since it involves just reading the file.  Other combinations of the executable credential source are exercised anyway in other tests.

coryan: You need to decide what scenario is tested here.  If the scenario is "we call an executable and it prints the credentials to stdout" then configure only that.  And make sure the program only performs that test (it is fine if the program sources a file, as long as the file is not also configured in the test as an input to the credentials).

If the scenario is "we call an executable and it both prints the credentials to stdout and it caches them to a file" then configure and test that.

It is also fine to only test one scenario in the integration tests, but be clear about which one you are testing.

alvarowolfx: make sense, just pushed an update setting just the `command` and enabling execution with env var.

coryan (approved):

PR #2422, opened 2025-06-13
feat(generator): maps with custom serde

coryan:
Implement serialization and deserialization for maps where at least one
of the fields requires a custom transform.

Part of the work for #2376

### DIFF
--- a/generator/internal/rust/annotate.go
+++ b/generator/internal/rust/annotate.go
@@ -264,16 +264,17 @@ type fieldAnnotations struct {
 	// The templates need to generate different code for boxed fields.
 	IsBoxed bool
 	// If true, it requires a serde_with::serde_as() transformation.
-	RequiresSerdeAs bool
-	SerdeAs         string
+	SerdeAs string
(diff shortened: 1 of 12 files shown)


generator/internal/rust/annotate.go
```diff
@@ -691,69 +692,28 @@ func (c *codec) annotateOneOf(oneof *api.OneOf, message *api.Message, state *api
...
-			return fmt.Sprintf("std::collections::HashMap<%s, %s>", key, value)
-		}
-		//		if field.TypezID == ".google.protobuf.Value" {
-		//			return "wkt::internal::OptionalValue"
-		//		}
-		formatter = "serde_with::Same"
+		return "serde_with::base64::Base64"
 	default:
-		formatter = "serde_with::Same"
+		return ""
```
dbolduc, line 712: comment: Hm if we kept returning `serde_with::Same` instead of a sentinel, we wouldn't need to reassign `key` and `value` later.

It's fine with me though.

PR #2424, opened 2025-06-13
feat(gax): error kind for exhausted policies

coryan:
Part of the work for #2312

### DIFF
--- a/src/gax/src/error/core_error.rs
+++ b/src/gax/src/error/core_error.rs
@@ -136,20 +136,23 @@ impl Error {
         }
     }
 
-    /// The request could not be before the retry policy expired.
+    /// The request could not complete be before the retry policy expired.
     ///
     /// This is always a client-side generated error, but it may be the result
-    /// of multiple errors received from the service. Sometimes the error
+    /// of multiple errors received from the service.
     ///
     /// # Troubleshooting
     ///
-    /// The most common cause of this problem is setting a timeout value that is
-    /// based on the observed latency when the service is not under load.
-    /// Consider increasing the timeout value to handle temporary latency
-    /// increases too.
+    /// The most common cause of this problem is a transient problem that lasts
+    /// longer than your retry policy. For example, your retry policy may
+    /// effectively be exhausted after a few seconds, but some services may take
+    /// minutes to recover.
     ///
-    /// It could also indicate a congestion in the network, a service outage, or
-    /// a service that is under load and will take time to scale up.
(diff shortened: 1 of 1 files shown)


src/gax/src/error/core_error.rs
```diff
@@ -119,6 +119,41 @@ impl Error {
...
+    /// assert!(error.source().is_some());
+    /// ```
+    pub fn exhausted<T: Into<BoxError>>(source: T) -> Self {
+        Self {
+            kind: ErrorKind::Exhausted,
+            source: Some(source.into()),
+        }
+    }
+
+    /// The request could not be before the retry policy expired.
```
dbolduc: "could not complete before ..."

src/gax/src/error/core_error.rs
```diff
@@ -119,6 +119,41 @@ impl Error {
...
+    ///
+    /// This is always a client-side generated error, but it may be the result
+    /// of multiple errors received from the service. Sometimes the error
```
dbolduc: Sometimes the error... ?

dbolduc (approved):
I think we discussed this case being a `LimitedElapsedTimeError`:

https://github.com/googleapis/google-cloud-rust/blob/88c24a251eab8e688674bebd1ddfaf1166dbc59f/src/gax/src/retry_loop_internal.rs#L68

maybe for a follow up?

src/gax/src/retry_policy.rs
```diff
@@ -395,9 +421,8 @@ where
         if now < deadline {
             None
         } else {
-            Some(Error::other(format!(
-                "limited time retry policy exhausted {:?} ago",
-                now.saturating_duration_since(deadline)
+            Some(Error::exhausted(LimitedElapsedTimeError::new(
+                self.maximum_duration,
             )))
```
dbolduc, line 426: Now that I think about it, doesn't this drop any service errors from the retry loop? Can we surface the last failure?

src/gax/src/error/core_error.rs
coryan: Fixed, thanks.

src/gax/src/error/core_error.rs

src/gax/src/retry_policy.rs
coryan: I think yes, we want to surface the last failure, but we need to reorganize the code to propagate the error from the retry loop, and (maybe) to return the previous error if it should continue.

coryan: #2432

PR #2433, opened 2025-06-16
impl(generator): transform bool map keys

coryan:
Part of the work for #2376

### DIFF
--- a/src/wkt/tests/message_with_bool.rs
+++ b/src/wkt/tests/message_with_bool.rs
@@ -41,6 +41,7 @@ mod test {
     #[test_case(MessageWithBool::new().set_or_clear_optional(None::<bool>), json!({}))]
     #[test_case(MessageWithBool::new().set_repeated([true, true, false]), json!({"repeated": [true, true, false]}))]
     #[test_case(MessageWithBool::new().set_map_key([(true, "trueValue"), (false, "falseValue")]), json!({"mapKey": {"true": "trueValue", "false": "falseValue"}}))]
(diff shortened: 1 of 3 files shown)


src/wkt/tests/message_with_bool.rs
```diff
@@ -41,6 +41,7 @@ mod test {
     #[test_case(MessageWithBool::new().set_or_clear_optional(None::<bool>), json!({}))]
     #[test_case(MessageWithBool::new().set_repeated([true, true, false]), json!({"repeated": [true, true, false]}))]
     #[test_case(MessageWithBool::new().set_map_key([(true, "trueValue"), (false, "falseValue")]), json!({"mapKey": {"true": "trueValue", "false": "falseValue"}}))]
+    #[test_case(MessageWithBool::new().set_map_key([(true, "trueValue")]), json!({"mapKey": {"tr\u{0075}e": "trueValue"}}))]
```
dbolduc: let's use 'u' instead of the unicode for 'u'. :slightly_smiling_face:

coryan: That is the point... we are trying to verify the unicode for `u` deserializes correctly.

dbolduc: Can you name the test case then? `; "handles unicode in key"`

I was obviously confused by `tr\u{0075}e`, the next person may be too.

dbolduc (dismissed):

dbolduc (approved):
Thanks

PR #2436, opened 2025-06-16
cleanup(lro): prefer `Error::deser`

coryan:
Part of the work for #2312

### DIFF
--- a/src/lro/src/details.rs
+++ b/src/lro/src/details.rs
@@ -148,13 +148,15 @@ fn as_result<R, M>(op: Operation<R, M>) -> Result<R>
 where
     R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned,
 {
-    if let Some(any) = op.response() {
-        return any.to_msg::<R>().map_err(Error::deser);
+    // The result must set either the response *or* the error. Setting neither
+    // is a deserialization error, as the incoming data does not satisfy the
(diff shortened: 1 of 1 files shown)


src/lro/src/details.rs
```diff
@@ -148,13 +148,15 @@ fn as_result<R, M>(op: Operation<R, M>) -> Result<R>
...
     R: wkt::message::Message + serde::ser::Serialize + serde::de::DeserializeOwned,
 {
-    if let Some(any) = op.response() {
-        return any.to_msg::<R>().map_err(Error::deser);
+    // The result must set either the response *or* the error. Setting neither
+    // is a deserialization error, as the incoming data does not satisfy the
+    // invariants required by the receiving type.
+    match (op.response(), op.error()) {
+        (Some(any), None) => any.to_msg::<R>().map_err(Error::deser),
+        (None, Some(e)) => Err(Error::service(gax::error::rpc::Status::from(e.clone()))),
```
dbolduc, line 156: nit: the clone seems unnecessary

dbolduc (approved):

src/lro/src/details.rs
coryan: It is a consequence of returning a `Option<&Status>`.  We may want to fix that, though I think a `from(&Status)` would be a better solution: #2440

PR #2451, opened 2025-06-17
impl(generator): reject duplicate fields

coryan:
Part of the work for #2376


src/wkt/tests/message_with_f32.rs
```diff
@@ -88,6 +88,16 @@ mod test {
         Ok(())
     }
 
+    #[test_case(r#"{"singular": 0,  "singular": 0}"#)]
```
dbolduc, line 91: > Multiple values for singular fields (using duplicate or equivalent JSON keys) are accepted and the last value is retained, as with binary format parsing. Note that not all protobuf JSON parser implementations are conformant, and some nonconformant implementations may reject duplicate keys instead.

I thought we were going to go for conformance?

dbolduc: Nvm, we are not going to conform to something that doesn't make sense.

PR #2452, opened 2025-06-18
feat(auth): Impersonated Service Account Credentials

sai-sunder-s:
Introduces the impersonated creds.

Refer to https://cloud.google.com/iam/docs/reference/credentials/rest/v1/projects.serviceAccounts/generateAccessToken for the API call we are making for getting the impersonated token.

### DIFF
--- a/src/auth/src/credentials/impersonated.rs
+++ b/src/auth/src/credentials/impersonated.rs
@@ -142,7 +142,7 @@ impl Builder {
     /// # use serde_json::json;
     /// #
     /// # tokio_test::block_on(async {
-    /// let source_credentials = user_account::Builder::new(json!("{ /* add details here */ }")).build()?;
+    /// let source_credentials = user_account::Builder::new(json!({ /* add details here */ })).build()?;
     ///
     /// let creds = impersonated::Builder::from_source_credentials(source_credentials)
     ///     .with_target_principal("test-principal")
@@ -170,7 +170,7 @@ impl Builder {
     /// # use serde_json::json;
     /// #
     /// # tokio_test::block_on(async {
-    /// let impersonated_credential = json!("{ /* add details here */ }");
+    /// let impersonated_credential = json!({ /* add details here */ });
     ///
     /// let creds = impersonated::Builder::new(impersonated_credential.into())
     ///     .with_target_principal("test-principal")
@@ -193,7 +193,7 @@ impl Builder {
     /// # use serde_json::json;
     /// #
     /// # tokio_test::block_on(async {
-    /// let impersonated_credential = json!("{ /* add details here */ }");
+    /// let impersonated_credential = json!({ /* add details here */ });
     ///
     /// let creds = impersonated::Builder::new(impersonated_credential.into())
     ///     .with_delegates(["delegate1", "delegate2"])
@@ -218,7 +218,7 @@ impl Builder {
     /// # use serde_json::json;
     /// #
     /// # tokio_test::block_on(async {
-    /// let impersonated_credential = json!("{ /* add details here */ }");
+    /// let impersonated_credential = json!({ /* add details here */ });
     ///
     /// let creds = impersonated::Builder::new(impersonated_credential.into())
     ///     .with_scopes(["https://www.googleapis.com/auth/pubsub"])
@@ -252,7 +252,7 @@ impl Builder {
     /// # use serde_json::json;
     /// #
     /// # tokio_test::block_on(async {
-    /// let impersonated_credential = json!("{ /* add details here */ }");
+    /// let impersonated_credential = json!({ /* add details here */ });
     ///
     /// let creds = impersonated::Builder::new(impersonated_credential.into())
     ///     .with_quota_project_id("my-project")
@@ -274,7 +274,7 @@ impl Builder {
     /// # use std::time::Duration;
     /// #
     /// # tokio_test::block_on(async {
-    /// let impersonated_credential = json!("{ /* add details here */ }");
+    /// let impersonated_credential = json!({ /* add details here */ });
     ///
     /// let creds = impersonated::Builder::new(impersonated_credential.into())
     ///     .with_lifetime(Duration::from_secs(500))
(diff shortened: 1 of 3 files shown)


src/auth/src/credentials/impersonated.rs
```diff
@@ -0,0 +1,1263 @@
...
+// limitations under the License.
+
+//! [Impersonated service account] credentials.
+//!
+//! When the principal you are using doesn't have the permissions you need to
+//! accomplish your task, or you want to use a service account in a development
+//! environment, you can use service account impersonation. The typical principals
+//! used to impersonate a service account are [User Account] or another [Service Account].
+//!
+//! The princpial that is trying to impersonate a target service account should have
```
coryan:
```suggestion
//! The principal that is trying to impersonate a target service account should have
```

src/auth/src/credentials/impersonated.rs
```diff
@@ -0,0 +1,1263 @@
...
+    /// Creates a new builder with a source credentials object.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_auth::credentials::impersonated;
+    /// # use google_cloud_auth::credentials::user_account;
+    /// # use serde_json::json;
+    /// #
+    /// # tokio_test::block_on(async {
+    /// let source_credentials = user_account::Builder::new(json!("{ /* add details here */ }")).build()?;
```
coryan: nit: here and in other examples:

```suggestion
    /// let source_credentials = user_account::Builder::new(json!({ /* add details here */ })).build()?;
```

The quotes make that a JSON string, you want a JSON object, both are JSON values.

src/auth/src/credentials/impersonated.rs
```diff
@@ -0,0 +1,1263 @@
...
+    service_account_impersonation_url: String,
+    delegates: Option<Vec<String>>,
+    scopes: Vec<String>,
+    lifetime: Duration,
+}
+
+impl Debug for ImpersonatedTokenProvider {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("ImpersonatedTokenProvider")
+            .field("source_credentials", &"...")
```
coryan: Is there are reason to hide the source credentials? Presumably they are safe to print for debugging?  If there is a good reason, please add some comments.

src/auth/src/credentials/impersonated.rs
```diff
@@ -0,0 +1,1263 @@
...
+#[async_trait]
+impl TokenProvider for ImpersonatedTokenProvider {
+    async fn token(&self) -> Result<Token> {
+        let source_headers = self.source_credentials.headers(Extensions::new()).await?;
+        let source_headers = match source_headers {
+            CacheableResource::New { data, .. } => data,
+            CacheableResource::NotModified => {
+                return Err(errors::non_retryable_from_str(
+                    "unexpected not modified response for source credentials",
+                ));
```
coryan: I think you can use `unreachable!()` (or `.expect()`) here, anything else is a bug....
```suggestion
            CacheableResource::NotModified => unreachable!("requested source credentials without a caching etag"),
```

src/auth/src/credentials/impersonated.rs
```diff
@@ -0,0 +1,1263 @@
...
+                let retryable = !e.is_decode();
+                CredentialsError::from_source(retryable, e)
+            })?;
+
+        let parsed_dt = OffsetDateTime::parse(
+            &token_response.expire_time,
+            &time::format_description::well_known::Rfc3339,
+        )
+        .map_err(|e| {
+            errors::non_retryable_from_str(format!("failed to parse expireTime: {}", e))
```
coryan: can we preserve the original error instead of turning that to a string?  `non_retryable()` seems like a better choice.

src/auth/src/credentials/impersonated.rs
```diff
@@ -0,0 +1,1263 @@
...
+    expire_time: String,
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use httptest::{Expectation, Server, matchers::*, responders::*};
+    use serde_json::json;
+
+    type TestResult = std::result::Result<(), Box<dyn std::error::Error>>;
```
coryan: We started using `anyhow::Result` for these.

coryan (changes requested):
Very cool, thanks.  A few minor changes.  I have not looked at the tests in any detail, maybe work with @harkamaljot or @alvarowolfx to review those?

src/auth/tests/credentials.rs
```diff
@@ -132,6 +132,55 @@ mod test {
         assert!(fmt.contains(quota_project));
     }
 
+    #[tokio::test]
+    #[serial_test::serial]
+    async fn create_access_token_credentials_adc_impersonated_service_account() {
+        let contents = r#"{
```
alvarowolfx: prefer using `json!(...).to_string()`

src/auth/tests/credentials.rs
```diff
@@ -132,6 +132,55 @@ mod test {
...
+        let _e = ScopedEnv::set("GOOGLE_APPLICATION_CREDENTIALS", path.to_str().unwrap());
+
+        let ic = AccessTokenCredentialBuilder::default().build().unwrap();
+        let fmt = format!("{:?}", ic);
+        assert!(fmt.contains("ImpersonatedServiceAccount"));
+    }
+
+    #[tokio::test]
+    async fn create_access_token_credentials_json_impersonated_service_account() {
+        let contents = r#"{
```
alvarowolfx: same as above, `json!(...).to_string()`

alvarowolfx (approved):
some nits on `#{""` vs `json!(..).to_string`, but LGTM

src/auth/src/credentials/impersonated.rs
sai-sunder-s: They are safe to print. But problem is we have the full Credentials here for the source, so printing it doesn't provide the inner details of the token provider of the source. I have updated to print it anyway, we could atleast check the type of the source cred.

coryan (approved):

dbolduc:
add an integration test to `auth-integration-tests`?

Issue #2335, opened 2025-05-30
Prefer to return `Result<(), Error>` over `Option<Error>`


dbolduc:
Returning a `Result` is more idiomatic. `Result`s are easier to chain with `?`.

We should change
- `PollingErrorPolicy::on_in_progress()`
- `RetryPolicy::on_throttle()`

Note that these fns are hidden from our docs, and thus not part of our public API.

PR #2456, opened 2025-06-18
feat(gax)!: preserve retry loop errors

### LINKED ISSUE #2432: Propagate previous error in `on_throttle()` callback
The retry loop may stop in the `on_throttle()` callback. In that case, we want to propagate the previous error back to the application, along with the exhausted policy error.

coryan:
Fixes #2432

### DIFF
--- a/src/gax/src/retry_policy.rs
+++ b/src/gax/src/retry_policy.rs
@@ -493,38 +493,6 @@ where
     }
 }
 
-/// An error type created when the [LimitedAttemptCount] policy is exhausted.
-#[derive(thiserror::Error, Debug)]
-pub struct LimitedAttemptCountError {
-    maximum_attempts: u32,
-    #[source]
-    source: Error,
-}
-
-impl LimitedAttemptCountError {
-    pub(crate) fn new(maximum_attempts: u32, source: Error) -> Self {
-        Self {
-            maximum_attempts,
-            source,
-        }
-    }
-
-    /// Returns the maximum number of attempts in the exhausted policy.
-    pub fn maximum_attempts(&self) -> u32 {
-        self.maximum_attempts
-    }
-}
-
-impl std::fmt::Display for LimitedAttemptCountError {
-    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
-        write!(
-            f,
-            "retry policy exhausted after {} attempts, the last retry attempt was throttled",
-            self.maximum_attempts
-        )
-    }
-}
-
 /// A retry policy decorator that limits the number of attempts.
 ///
 /// This policy decorates an inner policy and limits the total number of
@@ -587,17 +555,6 @@ where
             maximum_attempts,
         }
     }
-
-    fn stop_if_exhausted(&self, attempt_count: u32, error: Error) -> ThrottleResult {
-        if attempt_count < self.maximum_attempts {
-            ThrottleResult::Continue(error)
-        } else {
-            ThrottleResult::Exhausted(Error::exhausted(LimitedAttemptCountError::new(
-                self.maximum_attempts,
-                error,
-            )))
-        }
-    }
 }
 
 impl<P> RetryPolicy for LimitedAttemptCount<P>
@@ -630,10 +587,10 @@ where
         attempt_count: u32,
         error: Error,
     ) -> ThrottleResult {
-        match self.inner.on_throttle(loop_start, attempt_count, error) {
-            ThrottleResult::Continue(e) => self.stop_if_exhausted(attempt_count, e),
-            ThrottleResult::Exhausted(e) => ThrottleResult::Exhausted(e),
-        }
+        // The retry loop only calls `on_throttle()` if the policy has not
+        // been exhausted.
+        assert!(attempt_count < self.maximum_attempts);
+        self.inner.on_throttle(loop_start, attempt_count, error)
     }
 
     fn remaining_time(
@@ -988,16 +945,6 @@ mod tests {
     }
 
     #[test]
-    fn limited_attempt_count_error() {
-        let limit = 1234_u32;
-        let err = LimitedAttemptCountError::new(limit, unavailable());
-        assert_eq!(err.maximum_attempts(), limit);
-        let fmt = err.to_string();
-        assert!(fmt.contains("1234"), "display={fmt}, debug={err:?}");
-        assert!(err.source().is_some(), "{err:?}");
-    }
-
-    #[test]
     fn test_limited_attempt_count_on_error() {
         let mut mock = MockPolicy::new();
         mock.expect_on_error()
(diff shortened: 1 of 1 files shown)


src/gax/src/retry_policy.rs
```diff
@@ -798,7 +834,9 @@ mod tests {
         mock.expect_on_error()
             .times(1..)
             .returning(|_, _, _, e| LoopState::Continue(e));
-        mock.expect_on_throttle().times(1..).returning(|_, _| None);
+        mock.expect_on_throttle()
+            .times(1..)
```
dbolduc, line 809: oh neat, these things accept ranges. TIL.

src/gax/src/retry_policy.rs
```diff
@@ -93,8 +93,13 @@ pub trait RetryPolicy: Send + Sync + std::fmt::Debug {
     /// * `attempt_count` - the number of attempts. This method is never called
     ///   before the first attempt.
     #[cfg_attr(not(feature = "_internal-semver"), doc(hidden))]
-    fn on_throttle(&self, _loop_start: std::time::Instant, _attempt_count: u32) -> Option<Error> {
```
dbolduc, line 96: FYI: this fixes half of #2335

src/gax/src/retry_policy.rs
```diff
@@ -93,8 +93,13 @@ pub trait RetryPolicy: Send + Sync + std::fmt::Debug {
     /// * `attempt_count` - the number of attempts. This method is never called
     ///   before the first attempt.
```
dbolduc, line 95: should add the `error` parameter

src/gax/src/retry_loop_internal.rs
```diff
@@ -75,11 +75,14 @@ where
                 .throttle_retry_attempt()
             {
                 // This counts as an error for the purposes of the retry policy.
-                if let Some(error) = retry_policy.on_throttle(loop_start, attempt_count) {
-                    return Err(error);
-                }
+                let error = match retry_policy.on_throttle(loop_start, attempt_count, prev_error) {
```
dbolduc, line 80: It is strange that `on_throttle` can yield a `LoopState::Permanent(e)`.

Permanent errors terminate the retry loop immediately without involving the throttler.

If we get here `prev_error` was already deemed to not be permanent. Why would throttling change that?

dbolduc (commented):
Returning `LoopState`

Pros:
- applications do not need to learn a new type

Cons:
- permanent should never happen, and could be disallowed

I am not opposed to returning this type but wanted to call this out. Thoughts?

src/gax/src/retry_loop_internal.rs
coryan: Fixed.

src/gax/src/retry_policy.rs
coryan: Fixed.

src/gax/src/retry_policy.rs
```diff
@@ -974,42 +1025,40 @@ mod tests {
 
     #[test]
     fn test_limited_attempt_count_on_throttle_none() {
```
dbolduc: optional nit: s/none/continue/

src/gax/src/retry_policy.rs
```diff
@@ -587,16 +624,15 @@ where
...
+    fn on_throttle(
+        &self,
+        loop_start: std::time::Instant,
+        attempt_count: u32,
+        error: Error,
+    ) -> ThrottleResult {
+        match self.inner.on_throttle(loop_start, attempt_count, error) {
+            ThrottleResult::Continue(e) => self.stop_if_exhausted(attempt_count, e),
+            ThrottleResult::Exhausted(e) => ThrottleResult::Exhausted(e),
         }
```
dbolduc: optional nit: I think we can just return `self.inner.on_throttle(...)`

If we have hit the maximum attempt count, the retry loop was terminated before the throttler gets involved.

(Attempts do not increase on throttle).

src/gax/src/retry_policy.rs
```diff
@@ -974,42 +1025,40 @@ mod tests {
...
+        assert!(matches!(
+            policy.on_throttle(now, 1, unavailable()),
+            ThrottleResult::Continue(_)
+        ));
+        assert!(matches!(
+            policy.on_throttle(now, 2, unavailable()),
+            ThrottleResult::Continue(_)
+        ));
+        assert!(matches!(
+            policy.on_throttle(now, 3, unavailable()),
```
dbolduc: this doesn't happen in reality, so maybe just test a single `on_throttle` returning `Continue`?

src/gax/src/throttle_result.rs
```diff
@@ -0,0 +1,67 @@
...
+///        # panic!();
+///     }
+/// }
+/// ```
+#[derive(Debug)]
+pub enum ThrottleResult {
+    /// The error is retryable, but the policy is stopping the loop.
+    ///
+    /// Loop control policies may stop the loop on retryable errors, for
+    /// example, because the policy only allows a limited number of attempts.
```
dbolduc, line 62: super nit: do we need a word other than "attempts" that means rpc attempts + throttles?

Maybe "attempts + throttles"?

dbolduc: Oh wait, nothing is tracking the # of throttles per client call. Hmm... Should that be an argument to `on_throttle`?

dbolduc (approved):
I feel like `attempt_count` does not tell us enough, when throttles don't increase attempts.

But I also remember that the interface is hidden, and not public. This PR is about preserving the last error, which it does. So approved and we can nit about the interface later.

src/gax/src/throttle_result.rs
coryan: Maybe, but that seems like an unrelated issue, #2495

src/gax/src/retry_policy.rs
coryan: The exact sequence of calls does not happen, yes.  I am testing that `on_throttle()` returns `Continue` as long as the attempt count is small enough.  I think the test is fine.

src/gax/src/retry_policy.rs
coryan: Done.

src/gax/src/retry_policy.rs
coryan: True. That means the error type is unused too.

src/gax/src/retry_policy.rs
```diff
@@ -587,17 +581,16 @@ where
...
+    fn on_throttle(
+        &self,
+        loop_start: std::time::Instant,
+        attempt_count: u32,
+        error: Error,
+    ) -> ThrottleResult {
+        // The retry loop only calls `on_throttle()` if the policy has not
+        // been exhausted.
+        assert!(attempt_count < self.maximum_attempts);
+        self.inner.on_throttle(loop_start, attempt_count, error)
```
dbolduc, line 593: 👍

dbolduc (approved):

PR #2468, opened 2025-06-19
fix(generator): preserve numbered lists

coryan:
Fixes most of  the clippy warnings from:

https://github.com/googleapis/google-cloud-rust/actions/runs/15762699268/job/44432800432?pr=2464

I do not know why these clippy warnings only appear after #2464 The warnings are completely valid, the documentation output is terrible:

https://docs.rs/google-cloud-dialogflow-v2/0.3.0/google_cloud_dialogflow_v2/model/struct.StreamingDetectIntentRequest.html

Note that the source protos really are a numbered list:

https://github.com/googleapis/googleapis/blob/f01a17a560b4fbc888fd552c978f4e1f8614100b/google/cloud/dialogflow/v2/session.proto#L393-L418

With these changes the documentation output looks much better:

image


dbolduc (approved):
Nice

PR #2473, opened 2025-06-20
impl(generator): support documentation overrides

coryan:
From time to time we may need to patch the documentation for a service,
method, message, field, enum, or enum value. This new configuration
option provides the necessary functionality.

Part of the work for #2376


dbolduc (approved):
Thanks, I hate it.

coryan:
> Thanks, I hate it.

LOL.  One good thing is that we will get errors once the upstream fixes roll in.

PR #2478, opened 2025-06-20
docs(wkt): hide `serde` trait implementations

coryan:
We do not want to advertise these too much. We need them to serialize
the messages over HTTP+JSON. We do not want to commit to supporting
these nor do we want to guarantee they work over any serde serialization
format.

### DIFF
--- a/src/wkt/src/any.rs
+++ b/src/wkt/src/any.rs
@@ -222,6 +222,7 @@ impl crate::message::Message for Any {
 }
 
 /// Implement [`serde`](::serde) serialization for [Any].
+#[cfg_attr(not(feature = "_internal-semver"), doc(hidden))]
 impl serde::ser::Serialize for Any {
     fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error>
     where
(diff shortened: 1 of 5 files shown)


src/wkt/src/any.rs
```diff
@@ -222,6 +222,7 @@ impl crate::message::Message for Any {
 }
 
 /// Implement [`serde`](::serde) serialization for [Any].
+#[cfg_attr(not(feature = "_internal-semver"), doc(hidden))]
```
dbolduc, line 225: It gave me pause to hide the `serde` traits from our messages, given that `serde` is a prereq for use with `wkt::Any`:

https://github.com/googleapis/google-cloud-rust/blob/74de4242705e73092a371371829f55eef2b1134a/src/wkt/src/any.rs#L172

... but I think the right fix is to enforce that our `Message`s are `serde`. I sent #2480

dbolduc (approved):

src/wkt/src/any.rs
coryan: Ack.

PR #2481, opened 2025-06-20
impl(generator): generated `serde` by default [< b]

coryan:
Part of the work for #2376


generator/internal/rust/annotate.go
```diff
@@ -316,11 +316,14 @@ type enumValueAnnotation struct {
 // tags. For example, the Mustache tag {{#Services}} uses the
 // [Template.Services] field.
 func annotateModel(model *api.API, codec *codec) *modelAnnotations {
+	packageName := PackageName(model, codec.packageNameOverride)
+
 	codec.hasServices = len(model.State.ServiceByID) > 0
+	// TODO(#2376) - slowly introduce generated serialization by default
+	codec.withGeneratedSerde = codec.withGeneratedSerde || packageName < "google-cloud-b"
```
dbolduc, line 323: Last time I comment:

I think 10 PRs x 200k lines is not any better than 1 PR x 2M lines.

It is not like we have any coverage for most of these libraries. A "roll out" doesn't give us any feedback.

I vote to send it.

coryan: Okay, let me get that ready.

Issue #2317, opened 2025-05-29
Support multiple additional bindings for generated clients


coryan:
Currently we support only one binding, which limits what some clients can do.

dbolduc:
Checkpointing progress... I think I am finally over the hump with this feature.

https://github.com/dbolduc/google-cloud-rust/commits/additional-bindings-2-dev

This branch implements additional bindings / path validation for some of the services in our integration tests, e.g. `secretmanager`.

The generated code, and the golang to get there needs a lot of cleaning up. (`transport.rs` is 5x as long). Here is the generated code:

https://github.com/dbolduc/google-cloud-rust/blob/2f2420df38f2a9fa08d8032ed9941ffc65f38c1d/src/generated/cloud/secretmanager/v1/src/transport.rs#L481-L607

I am able to run the integration tests, although now this test fails (because the binding is invalid. Yay! But also, boo, now I have to update the test):

https://github.com/googleapis/google-cloud-rust/blob/aa1a7abf72d3a8a6ea6b414a2fd9093958c404ec/src/integration-tests/tests/idempotency.rs#L80

Time to start cleaning up and breaking this down into PRs....

dbolduc:
Another day of progress. I factored out the errors and supported non-string fields. The generated code looks like:

https://github.com/dbolduc/google-cloud-rust/blob/a4c36bc7358a3cbe4aad472f2b9cfd7a3ff31973/src/generated/cloud/secretmanager/v1/src/transport.rs#L456-L544

I can run the integration tests (including gapic-showcase) for real this time.

Some weirdness - additional bindings for mixins will be non-trivial. The LRO methods aggregate all of the additional bindings across all of the services. There are ~90 of these. Not super helpful. Hm.

dbolduc:
> additional bindings for mixins will be non-trivial. The LRO methods aggregate all of the additional bindings across all of the services.

I was wrong.

> There are ~90 of these

That is just `aiplatform`. Gross.

https://github.com/googleapis/googleapis/blob/8ee3aad1fd2f2376373a030cafc61b3e9788e48a/google/cloud/aiplatform/v1/aiplatform_v1.yaml#L253-L340

dbolduc:
Pushed another commit that simplified the path construction. No more local variables. They do not save us anything.

And another commit to support query parameters *per binding*. Note that "supporting query parameters" just means omitting fields that end up in the path. I have a lot of controversial thoughts here. I will write them up somewhere else...

[...]

Issue #557, opened 2024-12-30
Reconcile between `api.PathInfo` and `httprule.PathTemplate`


dangazineu:
In #545 we introduced an HTTP Rule parser, to represent the recursive [Path Template](https://github.com/googleapis/googleapis/blob/9c586e8f14bcd6dfe47c35a5dcbeac446e634f73/google/api/http.proto#L220-L227) syntax. But that left us with two different data structures to represent the same concept, and a function to parse between both (`httprule.ParseSegments`). 

We should refactor the code to either reuse the `httprule.PathTemplate` where appropriate, or reduce its visibility within its package.

PR #2485, opened 2025-06-20
refactor(generator): make way for a new `PathSegment` representation

dbolduc:
Part of the work for #2317, #557

I want the name `api.PathSegment` for myself. So make room for it, by giving the current representation a disparaging `Legacy*` prefix.

There are too many changes to swap out the representation for path segments in place. Both will have to exist while we build out the new implementation.


dbolduc:
If my code is lucky, it will get deleted instead of having to bear the "legacy" label

PR #2489, opened 2025-06-21
feat(auth): Add ProgrammaticSourcedCredentials

harkamaljot:
As part of allowing the user to provide their own subject token provider, this PR makes following changes:
1. Make SubjectTokenProvider trait public.
2. A new builder ProgrammaticSourcedCredentials is created to utilize this trait and a corresponding test case has been added.

In future PR, more fields will be added to the builder for ProgrammaticSourcedCredentials


src/auth/src/credentials/external_account_sources/url_sourced.rs
```diff
@@ -53,9 +56,9 @@ impl UrlSourcedCredentials {
 const MSG: &str = "failed to request subject token";
 const JSON_FORMAT_TYPE: &str = "json";
 
-#[async_trait::async_trait]
```
sai-sunder-s, line 56: What are the implications of removing this async trait?

harkamaljot: The only implication is SubjectTokenProvider is not dyn compatible, so we can create dynamic module, similar to one we discussed in the design doc, if we want this trait to be dyn compatible.

src/auth/src/credentials/external_account.rs
```diff
@@ -335,6 +336,136 @@ impl Builder {
...
+impl<T> ProgrammaticBuilder<T>
+where
+    T: SubjectTokenProvider + 'static,
+{
+    /// Creates a new builder with a custom [`SubjectTokenProvider`] and [external_account_credentials] JSON value.
+    ///
+    /// [external_account_credentials]: https://google.aip.dev/auth/4117#configuration-file-generation-and-usage
+    pub fn new_with_external_account_config<S: Into<Value>>(
+        subject_token_provider: T,
+        external_account_config: S,
```
sai-sunder-s, line 420: Since they are already providing part of the config through code, we should provide them a way to build the full credential through code.

src/auth/src/credentials/external_account.rs
```diff
@@ -76,7 +72,7 @@ struct ExternalAccountFile {
     client_id: Option<String>,
     client_secret: Option<String>,
     scopes: Option<Vec<String>>,
-    credential_source: CredentialSourceFile,
+    credential_source: Option<CredentialSourceFile>,
```
sai-sunder-s, line 75: Instead of making optional here, it would be better to create another struct just for programmatic and do not include credential_source in it.

sai-sunder-s: If you have a `ProgrammaticExternalAccountFile` struct which doesn't have credential_source, then you can combine that with the custom subject token provider and build the ExternalAccountConfig.

I believe that will make the code nicer, you do not have to pass source separately to make_credentials and pull it out of ExternalAccountConfig.

src/auth/src/credentials/external_account.rs
harkamaljot: Yes, I plan to do that in the next PR, as mentioned in the PR description.

src/auth/src/credentials/external_account_sources/url_sourced.rs
sai-sunder-s: Thanks, sorry missed that.

src/auth/src/credentials/external_account.rs
harkamaljot: Sorry I don't follow this comment do you mean to say do the following
1. create a new CredentialSource i.e. CredentialSource::Programmatic
2. Now we do not remove CredentialSource from the ExternalAccountConfig but rather, when ProgrammaticExternalAccount.into() is called it automatically converts to ExternalAccountConfig with source set as CredentialSource::Programmatic
3. ProgrammaticBuilder::Build method now calls the make_credentials rather than calling the make_credentials_from_source method?

src/auth/src/credentials/external_account.rs
sai-sunder-s: Lets add all the required builder methods as part of this PR. Lets do the method for loading from a config later. We can add that from config one if we get customer feedback that it is needed.

PR #2491, opened 2025-06-22
impl(gaxi): path mismatch error builder

dbolduc:
Part of the work for #2317 

Introduce a type to help us build errors. It is nicer if we factor this logic out into a common place to keep the generated code down.

If you are wondering "why not `let mut builder = ...` in the examples, that is because some paths do not have any substitutions. We could avoid this code altogether in the mustache template (as these paths can never error), but I find it easier to generate code that works for them anyways.

### DIFF
--- a/src/gax-internal/src/path_parameter.rs
+++ b/src/gax-internal/src/path_parameter.rs
@@ -20,6 +20,7 @@
 //! a small helper function makes the generated code easier to read.
 
 use crate::routing_parameter::Segment;
+use gax::error::binding::{PathMismatch, SubstitutionFail, SubstitutionMismatch};
 
 /// Checks if a string field matches a given path template
 ///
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/path_parameter.rs
```diff
@@ -43,6 +44,87 @@ pub fn try_match<'a>(value: &'a str, template: &[Segment]) -> Option<&'a str> {
...
+            Some(actual) if try_match(actual, template).is_none() => {
+                self.0.subs.push(SubstitutionMismatch {
+                    field_name,
+                    problem: SubstitutionFail::MismatchExpecting(actual.to_string(), expecting),
+                });
+            }
+            _ => {}
+        };
+        self
+    }
```
coryan, line 107: Can you enumerate the list of `T` types?  If so, you could refactor this code to use (sealed) trait and make the generated code event simpler:

```suggestion
    pub fn maybe_add_error(
        mut self,
        value: Option<&T>,
        field_name: &'static str,
        template: &[Segment],
    ) -> Self 
    where T: SomeTrait   
    {
    }
    
 trait SomeTrait { fn try_match(&self, template) ... }
 impl SomeTrait for Option<&str> { ... }
 impl SomeTrait for Option<&i32> { ... }
```

src/gax-internal/src/path_parameter.rs
```diff
@@ -65,17 +147,75 @@ mod tests {
     #[test_case("projects/", None)]
     #[test_case("projects/my-project/", None)]
     #[test_case("projects/my-project/locations/my-location", None)]
-    fn try_match(input: &str, expected: Option<&str>) {
-        let p = super::try_match(
+    fn test_try_match(input: &str, expected: Option<&str>) {
```
coryan: nit: I do not like the redundancy in these `tests::test_*` function names 🤷

coryan (dismissed):

src/gax-internal/src/path_parameter.rs
dbolduc: > Can you enumerate the list of `T` types?

Probably. I know gapic-showcase uses an enum. We can always call `.value()`/`.name()` on the enum to get to the `T` we need.

I haven't actually tested against gapic-showcase to see if they expect a string / int / or either.

---

I planned to use whether the leaf type is a string in the path construction, too:

https://github.com/dbolduc/google-cloud-rust/blob/ab1db63660c8e2c22df67b701ffa1266202b7fd5/generator/internal/rust/templates/crate/src/transport.rs.mustache#L156-L161

That could similarly be built into the trait though, and then I would not need `IsString` anymore. Hm.

Let me defer this clean up until I am closer to generating and testing the code.

src/gax-internal/src/path_parameter.rs
dbolduc: Fair. I know I prefer `use super::*` to listing all the types I want.

Renamed the offending tests.

src/gax-internal/src/path_parameter.rs
coryan: > Probably. I know gapic-showcase uses an enum. We can always call `.value()`/`.name()` on the enum to get to the `T` we need.

s/always call/almost always call/ When the enum is holding unknown integers `.name()` returns `None` and when it is holding unknown strings `.name()` returns `None`.

> Let me defer this clean up until I am closer to generating and testing the code.

SGTM.

coryan (approved):

src/gax-internal/src/path_parameter.rs
dbolduc: > Let me defer this clean up

Opened #2497 

> s/always call/almost always call/

I am hoping that `std::fmt::Display for` our enum types will do the job for me.

FWIW, enums only appear in gapic-showcase bindings. They do not appear in real GCP service bindings... So I am tempted not to care. (I have a discussion item to call this out in our weekly meeting).

PR #2503, opened 2025-06-23
docs: more prerequisites to set up development environment

pcoet:
Add a call out in the dev setup guide noting that you have to enable the Secret Manager API.

### DIFF
--- a/guide/src/setting_up_your_development_environment.md
+++ b/guide/src/setting_up_your_development_environment.md
@@ -76,8 +76,9 @@ Manager API, do the following:
    cargo add google-cloud-secretmanager-v1
    ```
 
-   You'll need to enable the Secret Manager API in [APIs and services](https://console.cloud.google.com/apis),
-   if you haven't already done so. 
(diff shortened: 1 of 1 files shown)


guide/src/setting_up_your_development_environment.md
```diff
@@ -76,6 +76,9 @@ Manager API, do the following:
    cargo add google-cloud-secretmanager-v1
    ```
 
+   You'll need to enable the Secret Manager API in [APIs and services](https://console.cloud.google.com/apis),
+   if you haven't already done so. 
+
```
coryan: Thanks!

The robots would like you to wrap this differently (let me know if you cannot see the build logs). Once that goes through I would gladly approve.

dbolduc, line 86: Can we also give the command-line way to do this:
````suggestion

   ```shell
   gcloud services enable secretmanager.googleapis.com
   ```

````

dbolduc (approved):
Thanks!

PR #2510, opened 2025-06-24
refactor(generator): httprule parser uses `api.PathTemplate`

dbolduc:
Part of the work for #557 

We are back down to 2 representations for a path template.

I kept the `Literal` and `Identifier` types for the `parseLiteral`, `parseIdentifier` functions. I converted them back to strings in the callers.

Note that my last PR duplicated the tests. So we are not losing coverage.


generator/internal/parser/httprule/http_rule_parser_test.go
```diff
@@ -22,75 +22,6 @@ import (
 	"github.com/googleapis/google-cloud-rust/generator/internal/api"
 )
 
-func TestProtobuf_Parse(t *testing.T) {
```
coryan, line 25: Why remove the unit test?

coryan: Ah, because they were duplicated in a previous pass. Got it.

coryan (approved):

PR #2511, opened 2025-06-24
fix(auth): Implement Debug Trait for TokenCache

sai-sunder-s:
Implement std::fmt::Debug for Token Cache. This improves the respectability of TokenCache. This is done mainly for improving the respectability of source credentials of impersonated credentials.

### DIFF
--- a/src/auth/src/token_cache.rs
+++ b/src/auth/src/token_cache.rs
@@ -33,12 +33,7 @@ pub(crate) struct TokenCache<T: TokenProvider> {
     token_provider: Arc<T>,
 }
 
-// We are implementing Clone instead of using the Default implementation
-// because the default implementation requires all the members of the struct
-// to be Clone as well. TokenProvider is not Clone currently. Adding Clone to
-// TokenProvider will create a cascade of changes and will also need the
-// MockTokenProvider in token.rs to be Clone and it is not straightforward
-// to implement Clone for a mockall.
+// The default implementation requires `T` to implement `Clone`, which is not always the case.
 impl<T: TokenProvider> Clone for TokenCache<T> {
     fn clone(&self) -> Self {
         Self {


src/auth/src/token_cache.rs
```diff
@@ -26,23 +27,43 @@ use tokio::time::{Duration, Instant, sleep};
...
-        T: TokenProvider + Send + Sync + 'static,
-    {
+impl<T: TokenProvider> Clone for TokenCache<T> {
+    fn clone(&self) -> Self {
+        Self {
+            rx_token: self.rx_token.clone(),
+            token_provider: self.token_provider.clone(),
+        }
+    }
+}
```
coryan, line 44: Can you use `#[derive(Clone)]` for this? If not, do you mind adding a comment explaining why?

src/auth/src/token_cache.rs
```diff
@@ -26,23 +27,43 @@ use tokio::time::{Duration, Instant, sleep};
...
+    T: TokenProvider + Send + Sync + 'static,
+{
+    pub(crate) fn new(inner: T) -> Self {
         let (tx_token, rx_token) = watch::channel::<Option<Result<(Token, EntityTag)>>>(None);
+        let token_provider = Arc::new(inner);
 
-        tokio::spawn(async move {
-            refresh_task(inner, tx_token).await;
-        });
+        tokio::spawn(refresh_task(token_provider.clone(), tx_token));
```
coryan, line 54: nice

src/auth/src/token_cache.rs
```diff
@@ -371,7 +395,7 @@ mod test {
         let (tx, mut rx) = watch::channel::<Option<Result<(Token, EntityTag)>>>(None);
 
         tokio::spawn(async move {
-            refresh_task(mock, tx).await;
+            refresh_task(Arc::new(mock), tx).await;
         });
```
coryan, line 392: Here and below: maybe you could simplify this to:

```rust
tokio::spawn(refresh_task(mock, tx));
```

src/auth/src/token_cache.rs
```diff
@@ -26,23 +27,43 @@ use tokio::time::{Duration, Instant, sleep};
...
+}
+
+impl<T: TokenProvider> std::fmt::Debug for TokenCache<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("TokenCache")
+            .field("rx_token", &self.rx_token)
+            .field("token_provider", &self.token_provider)
+            .finish()
+    }
+}
```
coryan: Ditto, any reason why a `derive` does not work? Maybe add a comment explaining why so we don't forget?

src/auth/src/token_cache.rs
sai-sunder-s: refresh_task needs an Arc now

src/auth/src/token_cache.rs
sai-sunder-s: Good catch. Yeah I get same with Default implementation as well. Will remove.

src/auth/src/token_cache.rs
sai-sunder-s: Added explanation.

src/auth/src/token_cache.rs
```diff
@@ -27,11 +27,18 @@ use tokio::time::{Duration, Instant, sleep};
...
 
+#[derive(Debug)]
 pub(crate) struct TokenCache<T: TokenProvider> {
     rx_token: watch::Receiver<Option<Result<(Token, EntityTag)>>>,
     token_provider: Arc<T>,
 }
 
+// We are implementing Clone instead of using the Default implementation
+// because the default implementation requires all the members of the struct
+// to be Clone as well. TokenProvider is not Clone currently. Adding Clone to
```
coryan: And yet you call `.clone()` in all members ... While `TokenProvider` is not `Clone`, you have `Arc<TokenProvider>`.  Maybe the default implementation does work?

sai-sunder-s: Ah I see what you are saying. My comment is not accurate. Default Clone wants T to be Clone.

Currently .clone() is called on Arc, not the inner type. So effectively we are not cloning the token provider itself, we are just getting another reference right now. So we do not need TokenProvider to be Clone.

But for Default Clone, we need the TokenProvider to be Clone as well.

This is the error I get when I use Default Clone:
```
error[E0599]: the method `clone` exists for struct `TokenCache<MockTokenProvider>`, but its trait bounds were not satisfied
   --> src/auth/src/token_cache.rs:350:41
    |
31  |   pub(crate) struct TokenCache<T: TokenProvider> {
    |   ---------------------------------------------- method `clone` not found for this struct because it doesn't satisfy `token_cache::TokenCache<MockTokenProvider>: Clone`
...
350 |                   let cache_clone = cache.clone();
    |                                           ^^^^^ method cannot be called on `TokenCache<MockTokenProvider>` due to unsatisfied trait bounds
    |
   ::: src/auth/src/token.rs:88:5
    |
88  | /     mockall::mock! {
89  | |         #[derive(Debug)]
90  | |         pub TokenProvider { }
    | |_________________________- doesn't satisfy `MockTokenProvider: Clone`
    |
note: trait bound `MockTokenProvider: Clone` was not satisfied
   --> src/auth/src/token_cache.rs:30:17
    |
30  | #[derive(Debug, Clone)]
    |                 ^^^^^ unsatisfied trait bound introduced in this `derive` macro
help: consider annotating `MockTokenProvider` with `#[derive(Clone)]`
   --> src/auth/src/token.rs:88:5
    |
88  +     #[derive(Clone)]
89  |     mockall::mock! {
```

Not sure why the error is only for the mock token provider and not other Token Providers though. SA, User token provider etc. do not implement Clone as well.

src/auth/src/token_cache.rs
```diff
@@ -26,23 +27,41 @@ use tokio::time::{Duration, Instant, sleep};
...
-    pub(crate) fn new<T>(inner: T) -> Self
-    where
-        T: TokenProvider + Send + Sync + 'static,
-    {
+// We are implementing Clone instead of using the Default implementation
+// because the default implementation requires all the members of the struct
+// to be Clone as well. TokenProvider is not Clone currently. Adding Clone to
+// TokenProvider will create a cascade of changes and will also need the
+// MockTokenProvider in token.rs to be Clone and it is not straightforward
+// to implement Clone for a mockall.
```
coryan: Maybe just say:

```suggestion
// The default implementation requires `T` to implement `Clone`, which is not always the case.
```

coryan (dismissed):
Please adjust the comment and then I can approve again.

coryan (approved):

PR #2522, opened 2025-06-25
cleanup(gax): idiomatic return type for `PollingErrorPolicy::on_in_progress()`

chrischiedo:
Resolves the remaining half of #2335.

The other half was fixed in #2456 (`RetryPolicy::on_throttle()`)

cc: @dbolduc

### DIFF
--- a/src/gax/src/polling_error_policy.rs
+++ b/src/gax/src/polling_error_policy.rs
@@ -75,8 +75,8 @@ pub trait PollingErrorPolicy: Send + Sync + std::fmt::Debug {
         _loop_start: std::time::Instant,
         _attempt_count: u32,
         _operation_name: &str,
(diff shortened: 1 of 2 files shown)


dbolduc:
Thanks. Code in this file looks good. We also need corresponding changes in `google-cloud-lro` (which will hopefully simplify that code)

chrischiedo:
> which will hopefully simplify that code

Not sure that the change makes the code any simpler, but I think it's more idiomatic.

dbolduc:
> Not sure that the change makes the code any simpler, but I think it's more idiomatic.

You're right. I should have actually read the `lro` code before making such a statement. :grinning:

dbolduc (approved):
Awesome, thanks!

coryan (approved):

PR #2525, opened 2025-06-25
impl(generator/rust): introduce `bindingSubstitution`

dbolduc:
Part of the work for #2317

Defer adding the field accessor, half to keep the PRs manageable, half because it was blocked when I started on this PR.

I might resurrect #557 and remove `api.PathMatch`, and just store it as a `*` in a string. It would let me skip some processing here. :shrug:

### DIFF
--- a/generator/internal/rust/annotate_test.go
+++ b/generator/internal/rust/annotate_test.go
@@ -1481,13 +1481,13 @@ func TestBindingSubstitutionTemplates(t *testing.T) {
 	want := "projects/*/locations/*/**"
 
 	if want != got {
-		t.Errorf("TemplateAsString() failed. want=%s, got=%s", want, got)
+		t.Errorf("TemplateAsString() failed. want=%q, got=%q", want, got)
 	}
 
 	got = b.TemplateAsArray()
 	want = `&[Segment::Literal("projects/"), Segment::SingleWildcard, Segment::Literal("/locations/"), Segment::SingleWildcard, Segment::TrailingMultiWildcard]`
 
 	if want != got {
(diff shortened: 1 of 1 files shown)


generator/internal/rust/annotate_test.go
```diff
@@ -1437,3 +1471,23 @@ func TestPathBindingAnnotations(t *testing.T) {
...
+
+	if want != got {
+		t.Errorf("TemplateAsString() failed. want=%s, got=%s", want, got)
+	}
+
+	got = b.TemplateAsArray()
+	want = `&[Segment::Literal("projects/"), Segment::SingleWildcard, Segment::Literal("/locations/"), Segment::SingleWildcard, Segment::TrailingMultiWildcard]`
+
+	if want != got {
+		t.Errorf("TemplateAsArray() failed. want=%s, got=%s", want, got)
```
suzmue: nit: consider using %q to quote the result so its easier to verify which parts are actually in the string.

Since this does have quotes in the string they do end up being escaped, making it slightly less readable so also fine to leave as it is.

suzmue (dismissed):

generator/internal/rust/annotate_test.go
dbolduc: > `%q` to quote the result

Ah, TIL. Thanks.

> they do end up being escaped, making it slightly less readable

I see. Yeah, that is not easy on the eyes. If I manually quote them with a backtick I think we get the best of both worlds.

dbolduc:
PTAL

coryan (approved):

Issue #2523, opened 2025-06-25
Future-proof testing for HTTP path creation and validation


dbolduc:
A follow up to #2317 

I wrote tests against `gapic-showcase`, but the coverage is underwhelming. It cannot verify that "first match wins" for example. Also, we (Rust folk) don't own the thing. Its RPCs could gain additional bindings at any time. We either have to loosen our test expectations or possibly deal with breaks when updating showcase.

For better coverage, we are adding tests in `gax-internal` against "once-generated-but-now-maintained-by-hand" code. Ideally this code should be owned by the generator.

To do this, we should factor out the `let builder = ...` code in `transport.rs` into a common template. Then we generate just that bit for our unit tests.

`src/wkt/tests/common/src/generated/.sidekick.toml` is an example of how this can work.

dbolduc:
The model would be a single service with a single RPC. We would define it in a proto. The service here would be perfect:

https://github.com/googleapis/google-cloud-rust/blob/64e6538b3caa9f6b5cff05c9d98921328f6c4bc5/generator/internal/rust/annotate_test.go#L1268-L1420

Stretch: we could consider loading from that proto in the generator's rust annotation unit tests. It might simplify the tests a bit. (If you read protobuf easier than you read golang).

PR #2527, opened 2025-06-26
test: additional bindings

dbolduc:
Part of the work for #2317

Add unit tests for the additional binding logic. This is what our generated code will look like.

`gapic-showcase` does not give us enough coverage, so these tests are necessary. Note also that we can set strict expectations if we own the service definition. (We are not at the mercy of a service adding more bindings).

---

Related to #2523 - If you are curious how the binding code was generated, see:

https://github.com/dbolduc/google-cloud-rust/commit/007bc88771772eddb3b2b41b77f5f05feeaae571

### DIFF
--- a/src/integration-tests/tests/bindings.rs
+++ b/src/integration-tests/tests/bindings.rs
@@ -1,0 +1,437 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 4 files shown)


src/integration-tests/tests/bindings.rs
```diff
@@ -0,0 +1,437 @@
...
+
+        let reqwest = builder.build()?;
+        assert_eq!(reqwest.method(), reqwest::Method::POST);
+
+        let url = reqwest.url();
+        assert_eq!(url.path(), "/v1/projects/p/locations/l:first");
+
+        let actual_qps: HashSet<String> =
+            url.query_pairs().map(|(key, _)| key.into_owned()).collect();
+        let want_qps: HashSet<String> = ["project", "location", "id"]
```
coryan, line 312: You may be able to use a `HashSet<&str>` here.

dbolduc: I tried that originally, but it seemed like I still had to take ownership of the keys in `actual_qps`. So I would be converting them to some `Collection<String>`, then converting that to a `HashSet<&str>`. Basically it was 2 conversions either way.

I could definitely be missing something though.

coryan: SGTM as-is then.

PR #2529, opened 2025-06-26
impl(generator/rust): additional bindings in showcase

dbolduc:
Part of the work for #2317 

Add the mustache template needed to support additional bindings for Rust.

Enable this for gapic-showcase, with tests that it works.

I prefer to hardcode the `AdditionalBindings` flag, instead of plumbing a sidekick feature. The flag should have a lifetime of <1 day.

### DIFF
--- a/src/generated/showcase/src/transport.rs
+++ b/src/generated/showcase/src/transport.rs
@@ -17,12 +17,6 @@
 use crate::Result;
 #[allow(unused_imports)]
 use gax::error::Error;
-#[allow(unused_imports)]
-use gax::error::binding::BindingError;
-#[allow(unused_imports)]
-use gaxi::path_parameter::{PathMismatchBuilder, try_match};
-#[allow(unused_imports)]
-use gaxi::routing_parameter::Segment;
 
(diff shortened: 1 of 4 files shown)


src/generated/showcase/src/transport.rs
coryan, line 1: The amount of generated code is quite big, but I see that for methods with one binding it is not too bad.  Do you have a guess / estimate for the amount of new generated code?

dbolduc: My memory is that it was like +1M lines. Or if you don't trust my memory wait for the next PR.

A big chunk of the length comes from the fact that query params are per binding.

If our algorithm was to:
- take path fields out of the request
- take body out of the request
- add the rest as query parameters

... then we would not need to enumerate query parameters per binding.

---

I think the other big opportunity to reduce the code is to combine the path and error building. We should avoid building the errors in place while considering paths though.

Maybe there is a way to either yield a path, or a function that would produce the `PathMismatch` error. Then if no paths worked, we call all of the error functions.

dbolduc: I'll try this after I get the current impl checked in and write the user guide

Issue #2042, opened 2025-05-08
Add streaming support for `storage::Client::insert_object`


coryan:
We would want to enable large single-shot uploads from a streaming source.

PR #2531, opened 2025-06-26
impl(storage): implement source for uploads

coryan:
We want to upload arbitrarily long (up to the GCS limits, 5TiB) objects.
This introduces a trait and a number of implementations to represent
sources of upload data. These sources represent a stream of data, and
support rewinding the stream. That will make the uploads more efficient
for things like files or objects returned from other storage systems.

Part of the work for #2042

### DIFF
--- a/src/storage/src/upload_source.rs
+++ b/src/storage/src/upload_source.rs
@@ -210,7 +210,7 @@ mod test {
 
     #[tokio::test]
     async fn simple_bytes() -> Result {
-        let buffer = InsertPayload::from(bytes::Bytes::from_static(&CONTENTS));
+        let buffer = InsertPayload::from(bytes::Bytes::from_static(CONTENTS));
         let got = collect(buffer).await?;
         assert_eq!(got[..], CONTENTS[..], "{got:?}");
         Ok(())
@@ -252,7 +252,7 @@ mod test {
     #[tokio::test]
     async fn small_file() -> Result {
         let mut file = NamedTempFile::new()?;
-        file.write(&CONTENTS)?;
+        assert_eq!(file.write(CONTENTS)?, CONTENTS.len());
         file.flush()?;
         let read = file.reopen()?;
         let got = collect(tokio::fs::File::from(read)).await?;
@@ -263,7 +263,7 @@ mod test {
     #[tokio::test]
     async fn small_file_seek() -> Result {
         let mut file = NamedTempFile::new()?;
-        file.write(&CONTENTS)?;
+        assert_eq!(file.write(CONTENTS)?, CONTENTS.len());
         file.flush()?;
         let mut read = tokio::fs::File::from(file.reopen()?);
         read.seek(8).await?;
@@ -275,10 +275,10 @@ mod test {
     #[tokio::test]
     async fn larger_file() -> Result {
         let mut file = NamedTempFile::new()?;
-        file.write(&[0_u8; READ_SIZE])?;
-        file.write(&[1_u8; READ_SIZE])?;
-        file.write(&[2_u8; READ_SIZE])?;
-        file.write(&[3_u8; READ_SIZE])?;
+        assert_eq!(file.write(&[0_u8; READ_SIZE])?, READ_SIZE);
+        assert_eq!(file.write(&[1_u8; READ_SIZE])?, READ_SIZE);
+        assert_eq!(file.write(&[2_u8; READ_SIZE])?, READ_SIZE);
+        assert_eq!(file.write(&[3_u8; READ_SIZE])?, READ_SIZE);
         file.flush()?;
         assert_eq!(READ_SIZE % 2, 0);
         let mut read = tokio::fs::File::from(file.reopen()?);
--- a/src/storage/src/upload_source.rs
+++ b/src/storage/src/upload_source.rs
@@ -89,7 +89,7 @@ where
 /// Provides bytes for upload.
 ///
 /// Implementations of this trait provide data for Google Cloud Storage uploads.
-/// The data may be received asynchronously, such downloads from Google Cloud
+/// The data may be received asynchronously, such as downloads from Google Cloud
 /// Storage, other remote storage systems, or the result of repeatable
 /// computations.
 pub trait StreamingSource {
--- a/src/storage/src/upload_source.rs
+++ b/src/storage/src/upload_source.rs
@@ -255,6 +255,8 @@ mod test {
     async fn empty_stream() -> Result {
         let source = VecStream::new(vec![]);
         let payload = InsertPayload::from(source);
+        let range = payload.size_hint();
+        assert_eq!(range, (0, Some(0)));
         let got = collect(payload).await?;
         assert!(got.is_empty(), "{got:?}");
 
--- a/src/storage/src/upload_source.rs
+++ b/src/storage/src/upload_source.rs
@@ -222,6 +222,27 @@ mod test {
     }
 
     #[tokio::test]
+    async fn simple_u8() -> Result {
+        let buffer = InsertPayload::from(CONTENTS);
+        let range = buffer.size_hint();
+        assert_eq!(range, (CONTENTS.len() as u64, Some(CONTENTS.len() as u64)));
+        let got = collect(buffer).await?;
+        assert_eq!(got[..], CONTENTS[..], "{got:?}");
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn simple_str() -> Result {
+        const LAZY: &str = "the quick brown fox jumps over the lazy dog";
+        let buffer = InsertPayload::from(LAZY);
+        let range = buffer.size_hint();
+        assert_eq!(range, (LAZY.len() as u64, Some(LAZY.len() as u64)));
+        let got = collect(buffer).await?;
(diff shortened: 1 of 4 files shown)


src/storage/src/upload_source.rs
```diff
@@ -0,0 +1,361 @@
...
+where
+    T: StreamingSource,
+{
+    type Error = T::Error;
+
+    fn seek(&mut self, offset: u64) -> impl Future<Output = Result<(), Self::Error>> + Send {
+        self.payload.seek(offset)
+    }
+
+    fn next(&mut self) -> impl Future<Output = Option<Result<bytes::Bytes, Self::Error>>> + Send {
```
coryan, line 50: @suzmue should I make this `Result<Option<bytes::Bytes `?  I thought the former was more obviously a mapping for `futures::Stream`, but I see how the approach you took in the read response is easier to use.

suzmue: For the read library I saw others implemented both options (with `next()` and `try_next()` respectively).

If we are expecting users to implement this trait on their own types, I would go with whichever is more easy to implement from a type that implements `futures::Stream`.

Converting from one to the other is easy:
```rust
    pub async fn next(&mut self) -> Option<Result<bytes::Bytes>> {
        match self.inner.try_next().await {
            Ok(Some(item)) => Some(Ok(item)),
            Ok(None) => None,
            Err(e) => Some(Err(e)),
        }
    }
```

Note: I just realized I thought next and try_next were flipped in the aws BytesStream.

I went with the slightly easier to use API for read_object (matches reqwest::Response::chunk), but if we might want both I think `try_next` makes more sense for the Result response, so we would update to match here and (optionally) add another method try_next:
-  `next -> Option<Result<>>`
- `try_next -> Result<Option<>>

suzmue: Ah TIL about transpose: (thank you @dbolduc), even easier to convert:

https://doc.rust-lang.org/std/option/enum.Option.html#method.transpose

https://doc.rust-lang.org/std/result/enum.Result.html#method.transpose

coryan: I think using `Option<Result<T>>` makes it behave more like  `futures::Stream<Item = Result<T>>`, and (I speculate) this should become more familiar to developers over time.

That compiled with `transpose()` is pretty easy to use:

```rust
let mut source : impl StreamingSource = ...;
while let Some(bytes) = stream.next().await.transpose()? {
  // use bytes
}
```

Thoughts?

src/storage/src/upload_source.rs
```diff
@@ -0,0 +1,359 @@
...
+{
+    fn from(value: S) -> Self {
+        Self { payload: value }
+    }
+}
+
+/// Provides bytes for upload.
+///
+/// Implementations of this trait provide data for Google Cloud Storage uploads.
+/// The data may be received asynchronously, such as
```
dbolduc: ...

src/storage/src/upload_source.rs
```diff
@@ -0,0 +1,359 @@
...
+
+    /// Gets the next set of data tp upload.
+    fn next(&mut self) -> impl Future<Output = Option<Result<bytes::Bytes, Self::Error>>> + Send;
+
+    /// A estimate of the upload size.
+    ///
+    /// Returns the expected size as a [min, max) range. Where `None` represents
+    /// an unknown limit for the upload.
+    ///
+    /// When the size
```
dbolduc: ...

src/storage/src/upload_source.rs
```diff
@@ -0,0 +1,359 @@
...
+    /// may require resetting the stream to an arbitrary point.
+    ///
+    /// The client library assumes that `seek(N)` followed by `next()` always
+    /// returns the same data.
+    fn seek(&mut self, offset: u64) -> impl Future<Output = Result<(), Self::Error>> + Send;
+
+    /// Gets the next set of data tp upload.
+    fn next(&mut self) -> impl Future<Output = Option<Result<bytes::Bytes, Self::Error>>> + Send;
+
+    /// A estimate of the upload size.
```
dbolduc: s/A/An/

src/storage/src/upload_source.rs
```diff
@@ -0,0 +1,359 @@
...
+    ///
+    /// The client library automatically restarts uploads when the connection
+    /// is reset or there is some kind of partial failure. Resuming an upload
+    /// may require resetting the stream to an arbitrary point.
+    ///
+    /// The client library assumes that `seek(N)` followed by `next()` always
+    /// returns the same data.
+    fn seek(&mut self, offset: u64) -> impl Future<Output = Result<(), Self::Error>> + Send;
+
+    /// Gets the next set of data tp upload.
```
dbolduc: s/tp/to/

src/storage/src/upload_source.rs
```diff
@@ -0,0 +1,359 @@
...
+    fn size_hint(&self) -> (u64, Option<u64>) {
+        (0_u64, None)
+    }
+}
+
+const READ_SIZE: usize = 256 * 1024;
+
+impl<S> StreamingSource for S
+where
+    S: tokio::io::AsyncRead + tokio::io::AsyncSeek + Unpin + Send,
```
dbolduc, line 127: TIL

src/storage/src/upload_source.rs
```diff
@@ -0,0 +1,359 @@
...
+    async fn seek(&mut self, offset: u64) -> Result<(), Self::Error> {
+        let _ = tokio::io::AsyncSeekExt::seek(self, std::io::SeekFrom::Start(offset)).await?;
+        Ok(())
+    }
+
+    async fn next(&mut self) -> Option<Result<bytes::Bytes, Self::Error>> {
+        let mut buffer = vec![0_u8; READ_SIZE];
+        match tokio::io::AsyncReadExt::read(self, &mut buffer).await {
+            Err(e) => Some(Err(e)),
+            Ok(0) => None,
```
dbolduc, line 140: Q: is there a difference between "the stream is done" and "hold on, let me get you more bytes"

(If what I am asking makes sense)

src/storage/src/upload_source.rs
```diff
@@ -0,0 +1,359 @@
...
+    fn next(&mut self) -> impl Future<Output = Option<Result<bytes::Bytes, Self::Error>>> + Send;
+
+    /// A estimate of the upload size.
+    ///
+    /// Returns the expected size as a [min, max) range. Where `None` represents
+    /// an unknown limit for the upload.
+    ///
+    /// When the size
+    fn size_hint(&self) -> (u64, Option<u64>) {
+        (0_u64, None)
```
dbolduc, line 119: comment just to convey understanding: the first type is `u64` because 0 is an appropriate lower bound. We don't need `None`.

src/storage/src/upload_source.rs
```diff
@@ -0,0 +1,359 @@
...
+    fn size_hint(&self) -> (u64, Option<u64>) {
+        let s = self.contents.len() as u64;
+        (s, Some(s))
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use super::*;
+    use std::{collections::VecDeque, io::Write};
```
dbolduc, line 184: fun to say

src/storage/src/upload_source.rs
coryan: Fixed, thanks

src/storage/src/upload_source.rs
coryan: Fixed.

src/storage/src/upload_source.rs
coryan: Fixed.

src/storage/src/upload_source.rs
coryan: Fixed.

src/storage/src/upload_source.rs
coryan: The `AsyncRead` API says that "0" implies (normal, expected) EOF *or* an empty destination buffer, but we know that is not the case here.

src/storage/src/upload_source.rs
coryan: Interesting... we may want to consider just returning `Option<u64>`, I cannot think of a case where we would need to know the lower bound.

dbolduc: Good point

src/storage/src/upload_source.rs
```diff
@@ -0,0 +1,362 @@
...
+{
+    fn from(value: S) -> Self {
+        Self { payload: value }
+    }
+}
+
+/// Provides bytes for upload.
+///
+/// Implementations of this trait provide data for Google Cloud Storage uploads.
+/// The data may be received asynchronously, such downloads from Google Cloud
```
dbolduc: nit s/such downloads/such as downloads/

dbolduc (approved):

src/storage/src/upload_source.rs
coryan: Fixed.

src/storage/src/upload_source.rs
suzmue: Agreed, will update read_object to match.

Issue #2047, opened 2025-05-08
Add streaming support for `Storage::read_object()`


coryan:
We need to change the API to return a streaming thing, instead of a buffer. The download may be as large as 5TiB, we do not want to allocate that much RAM.

PR #2533, opened 2025-06-26
feat(storage)!: support streaming response for Storage::read_object

suzmue:
Update the read object request to return a new type, `ReadObjectResponse`. The ReadObjectResponse has two methods: `all_bytes()` and `next()` to access the data. This type will also ultimately have methods for accessing metadata additional helpers for accessing data.

This PR updates the integration tests to test the new `next` method as well, but this code could use additional testing, but will plan to do that in a follow up PR, since it is no less tested than the existing code and will likely require choosing a mocking library, etc.

For #2047

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -401,7 +401,7 @@ impl InsertObject {
 /// #   .with_endpoint("https://storage.googleapis.com")
 /// #    .build().await?;
 /// let builder: ReadObject = client.read_object("projects/_/buckets/my-bucket", "my-object");
-/// let contents = builder.send().await?;
+/// let contents = builder.send().await?.all_bytes().await?;
 /// println!("object contents={contents:?}");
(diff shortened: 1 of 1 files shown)


src/integration-tests/src/storage.rs
```diff
@@ -56,10 +56,24 @@ pub async fn objects(builder: storage::client::ClientBuilder) -> Result<()> {
...
 
+    tracing::info!("testing read_object() streaming");
+    let mut contents = bytes::BytesMut::new();
+    let mut resp = client
+        .read_object(&bucket.name, &insert.name)
+        .send()
+        .await?;
+    while let Some(chunk) = resp.next().await? {
+        contents.extend_from_slice(&chunk);
+    }
```
coryan, line 73: For a future PR: it would be nice to have a test where this loop went through more than once or twice.

suzmue: Agreed.

PR #2536, opened 2025-06-26
impl: idempotency per binding

dbolduc:
Part of the work for #2317 

Additional bindings can have different HTTP methods (e.g. "GET" vs. "POST"). This means the default idempotency needs to be set per binding.

---

I wanted to remove the concept of `default_idempotency` from `RequestOptions` altogether, but ...

- `reqwest::ReqwestBuilder` is given the HTTP method, but it doesn't expose it until it becomes a `reqwest::Request`.
- We need to know the idempotency when we start the retry loop.
- We create the `reqwest::Reqwest` per attempt of the retry loop (where we calculate the timeout).

So let's just keep doing what we're doing.

### DIFF
--- a/generator/internal/rust/annotate.go
+++ b/generator/internal/rust/annotate.go
@@ -276,6 +276,9 @@ type pathBindingAnnotation struct {
 
 	// The variables to be substituted into the path
 	Substitutions []*bindingSubstitution
+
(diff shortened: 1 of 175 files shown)


dbolduc:
There is also the content length business, which depends on the method. I need to deal with that too.

I will likely close this PR and start over.

PR #2537, opened 2025-06-26
chore: update to rust 1.88

dbolduc:


src/lro/src/internal.rs
```diff
@@ -454,7 +454,7 @@ mod test {
         };
 
         use futures::StreamExt;
-        let mut stream = new_poller(
```
sai-sunder-s, line 457: seems unrelated?

dbolduc: `clippy` started flagging this in 1.88.

`clippy` is right. We are calling it `mut`, but never mutating it.

sai-sunder-s: I guess it is unused mut warning

sai-sunder-s (approved):

PR #2538, opened 2025-06-26
feat: Support Service Account Impersonation for External Accounts

sai-sunder-s:
Support impersonation in external accounts.

Users can add a `service_account_impersonation_url` in the byoid credential configuration to require impersonation.

The approach taken here is to use the code in impersonated credential to perform the impersonation call in external_account.


src/auth/integration-tests/src/lib.rs
```diff
@@ -157,6 +159,17 @@ pub async fn workload_identity_provider_url_sourced() -> anyhow::Result<()> {
       }
     });
 
+    if with_impersonation {
+        let impersonated_email =
+            "impersonation-target@cloud-sdk-auth-test-project.iam.gserviceaccount.com";
```
dbolduc, line 161: how do I run this in my test project?

src/auth/integration-tests/tests/driver.rs
```diff
@@ -28,12 +28,24 @@ mod driver {
     #[cfg(all(test, feature = "run-byoid-integration-tests"))]
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn run_workload_identity_provider_url_sourced() -> anyhow::Result<()> {
-        auth_integration_tests::workload_identity_provider_url_sourced().await
+        auth_integration_tests::workload_identity_provider_url_sourced(true)
+            .await
+            .unwrap();
```
dbolduc: Here and below, use the try operator `?`.

`anyhow::Result` will handle any input, so we don't need to `map_err()` or anything.

src/auth/integration-tests/src/lib.rs
```diff
@@ -157,6 +159,17 @@ pub async fn workload_identity_provider_url_sourced() -> anyhow::Result<()> {
...
+        let impersonation_url = format!(
+            "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{}:generateAccessToken",
```
dbolduc: `clippy` on 1.88 will make you say `{impersonated_email}`

src/auth/integration-tests/src/lib.rs
```diff
@@ -212,6 +227,17 @@ pub async fn workload_identity_provider_executable_sourced() -> anyhow::Result<(
...
```
dbolduc: same

src/auth/src/credentials/external_account.rs
```diff
@@ -191,6 +197,8 @@ where
     config: ExternalAccountConfig,
 }
 
+use crate::errors::non_retryable;
+
```
dbolduc: nit: move to top of file

src/auth/src/credentials/external_account.rs
```diff
@@ -218,6 +238,24 @@ where
 
         let token_res = STSHandler::exchange_token(req).await?;
 
+        if let Some(impersonation_url) = self.config.service_account_impersonation_url.clone() {
```
dbolduc: nit: we should be able to borrow this without cloning

src/auth/src/credentials/external_account.rs
```diff
@@ -467,4 +514,242 @@ mod test {
             }
         }
     }
+
+    #[tokio::test]
+    async fn test_external_account_with_impersonation_success() {
+        let subject_token_server = Server::run();
+        let sts_server = Server::run();
+        let impersonation_server = Server::run();
```
dbolduc, line 520: oh wow, I have not been paying attention to @alvarowolfx's work. This is cool.

src/auth/src/credentials/external_account.rs
```diff
@@ -201,14 +209,26 @@ where
...
         let subject_token_type = self.config.subject_token_type.clone();
-        let scope = self.config.scopes.clone();
+        let user_scopes = self.config.scopes.clone();
         let url = self.config.token_url.clone();
+
+        // User provides the scopes to be set on the final token they receive. The scopes they
+        // provide does not necessarily have to include the iam scope or cloud platform scope
+        // which are needed to make the call to iamcredentials endpoint. So when minting the
+        // sts token, we use the user provided or iam scope depending on whether the sts provided
+        // token is to be used with impersonation or directly.
```
dbolduc: nit: you should get in the habit of capitalizing acronyms like STS or IAM. And stylizing things like `iamcredentials` with backticks. It makes things easier to read.

src/auth/src/credentials/external_account.rs
```diff
@@ -467,4 +514,242 @@ mod test {
...
+        );
+
+        sts_server.expect(
+            Expectation::matching(request::method_path("POST", "/token"))
+                .respond_with(status_code(500)),
+        );
+
+        let creds = Builder::new(contents).build().unwrap();
+        let err = creds.headers(Extensions::new()).await.unwrap_err();
+        assert!(err.is_transient());
```
dbolduc, line 701: Q: does the error tell us which leg of the impersonated flow failed?

IMO it should, and we should be able to verify that programmatically.

dbolduc: Ok, I see that one has a message: "failed to fetch token" and the other has a message "failed to exchange token".

And the underlying `reqwest::Error` might have a URL.

https://docs.rs/reqwest/latest/reqwest/struct.Error.html

Hm, this may be good enough.

src/auth/src/credentials/impersonated.rs
```diff
@@ -428,6 +428,60 @@ struct GenerateAccessTokenRequest {
...
+        })?;
+
+    let parsed_dt = OffsetDateTime::parse(
+        &token_response.expire_time,
+        &time::format_description::well_known::Rfc3339,
+    )
+    .map_err(errors::non_retryable)?;
+
+    let remaining_duration = parsed_dt - OffsetDateTime::now_utc();
+    let expires_at = Instant::now() + remaining_duration.try_into().unwrap();
```
dbolduc, line 474: nit: (I see this code is just moved, but) I think we prefer `.expect("reason why this can't fail")` over `.unwrap()`

src/auth/integration-tests/src/lib.rs
sai-sunder-s: https://github.com/googleapis/google-cloud-rust/blob/main/src/auth/integration-tests/README.md

src/auth/src/credentials/external_account.rs
sai-sunder-s: Oh yeah borrow is working, I was moving before and ended up doing clone

src/auth/src/credentials/external_account.rs
sai-sunder-s: Oh yeah httptest is cool! Cleaner than axum

src/auth/src/credentials/impersonated.rs
sai-sunder-s: I would end up writing something like expect("Remaining Duration should have been convertible into Instant but it failed" which is inferrable.

src/auth/src/credentials/external_account.rs
sai-sunder-s: Added str contains check

src/auth/integration-tests/src/lib.rs
sai-sunder-s: Oh running on your test project is going to take lot of setup, You could use the test project I have setup.

dbolduc: Sorry, I should be clear. I know how to run it in the auth test project.

Part of the spec for the auth integration tests are that they can be run in any project, and there are instructions to do so. That should stay working.

At least open an issue to make sure the auth integration docs are working

src/auth/src/credentials/impersonated.rs
dbolduc: Ack, and fair point. I think I have just trained myself to recoil every time I see `.unwrap()`

dbolduc (approved):

src/auth/integration-tests/tests/driver.rs
```diff
@@ -28,12 +28,16 @@ mod driver {
     #[cfg(all(test, feature = "run-byoid-integration-tests"))]
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn run_workload_identity_provider_url_sourced() -> anyhow::Result<()> {
-        auth_integration_tests::workload_identity_provider_url_sourced().await
```
alvarowolfx, line 31: use the `test_case` crate to pass the `with_impersonation` param. Would be something like:

```
#[test_case(false; "default")]
#[test_case(true; "with impersonation")]
async fn run_workload_identity_provider_url_sourced(with_impersonation: bool) -> anyhow::Result<()> {
  auth_integration_tests::workload_identity_provider_url_sourced(with_impersonation).await?;
}
```

src/auth/integration-tests/tests/driver.rs
```diff
@@ -28,12 +28,16 @@ mod driver {
...
+        auth_integration_tests::workload_identity_provider_url_sourced(true).await?;
+        auth_integration_tests::workload_identity_provider_url_sourced(false).await?;
+        Ok(())
     }
 
     #[cfg(all(test, feature = "run-byoid-integration-tests"))]
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     async fn run_workload_identity_provider_executable_sourced() -> anyhow::Result<()> {
```
alvarowolfx, line 38: same as above, use the `test_case` crate.

alvarowolfx (commented):
tests can be improved

PR #2540, opened 2025-06-27
test(auth): Integration test for Impersonated Credentials

sai-sunder-s:


src/auth/integration-tests/src/lib.rs
```diff
@@ -78,6 +80,57 @@ pub async fn service_account() -> anyhow::Result<()> {
...
+        .expect("missing payload in test-sa-creds-json response")
+        .data;
+
+    let source_sa_json: serde_json::Value = serde_json::from_slice(&source_sa_json)?;
+
+    let source_sa_creds = ServiceAccountCredentialsBuilder::new(source_sa_json).build()?;
+
+    let impersonated_creds =
+        ImpersonatedCredentialsBuilder::from_source_credentials(source_sa_creds)
+            .with_target_principal("impersonation-target@rust-auth-testing.iam.gserviceaccount.com")
```
dbolduc: This should use `project_id`

src/auth/integration-tests/src/lib.rs
```diff
@@ -78,6 +80,57 @@ pub async fn service_account() -> anyhow::Result<()> {
...
+    let client = SecretManagerService::builder()
+        .with_credentials(impersonated_creds)
+        .build()
+        .await?;
+
+    // Access a secret, which only this principal has permissions to do.
+    let response = client
+        .access_secret_version()
+        .set_name(format!(
+            "projects/{project}/secrets/impersonation-target-secret/versions/latest"
```
dbolduc, line 124: I assume terraform changes to create this secret are coming in a follow up PR...

src/auth/integration-tests/src/lib.rs
```diff
@@ -78,6 +80,57 @@ pub async fn service_account() -> anyhow::Result<()> {
...
+        ))
+        .send()
+        .await?;
+    let secret = response
+        .payload
+        .expect("missing payload in impersonation-target-secret response")
+        .data;
+    assert_eq!(secret, "impersonated_secret_value");
+
```
dbolduc, line 133: optional: consider also trying to use the impersonating principal directly to access the secret and verify that that fails.

On my last team, we claimed we had implemented impersonation, when actually we were using the impersonating SA principal directly.

The impersonating SA used in our tests had the same permissions as the impersonated SA. 🤦

dbolduc: This is less important if the test SA configuration is checked into the repo

src/auth/integration-tests/src/lib.rs
sai-sunder-s: Yes I will create a follow up PR for terraform

PR #2541, opened 2025-06-27
feat(auth): Add ProgrammaticSourcedCredentials 

harkamaljot:
As part of allowing the user to provide their own subject token provider, this PR makes following changes:

Make SubjectTokenProvider trait public.
A new builder ProgrammaticSourcedCredentials is created to utilize this trait and a corresponding test case has been added.


src/auth/src/credentials/external_account.rs
```diff
@@ -349,6 +350,159 @@ impl Builder {
...
+    /// Sets the required subject token type.
+    ///
+    /// This is the STS subject token type based on the OAuth 2.0 token exchange spec.
+    pub fn with_subject_token_type<S: Into<String>>(mut self, subject_token_type: S) -> Self {
+        self.config.subject_token_type(subject_token_type);
+        self
+    }
+
+    /// Sets the required token URL for the STS token exchange.
+    pub fn with_token_url<S: Into<String>>(mut self, token_url: S) -> Self {
```
sai-sunder-s: I believe we can default to `sts.googleapis.com`. We do not have to force user into providing this. @leosiracusa thoughts?

src/auth/src/credentials/external_account.rs
```diff
@@ -349,6 +350,159 @@ impl Builder {
...
+
+    /// Sets the optional [quota project] for this credentials.
+    ///
+    /// In some services, you can use a service account in
+    /// one project for authentication and authorization, and charge
+    /// the usage to a different project. This requires that the
+    /// service account has `serviceusage.services.use` permissions on the quota project.
+    ///
+    /// [quota project]: https://cloud.google.com/docs/quotas/quota-project
+    pub fn with_quota_project_id<S: Into<String>>(mut self, quota_project_id: S) -> Self {
```
sai-sunder-s, line 589: missing examples on the builder methods

harkamaljot: The reason for not adding examples here was that since builder methods are justing passing in strings, similar to other places in the repo(eg: another builder in this moudule) I think we have only added examples in builder method, if user was passing in something complex. But if you think it increases clarity here, please let me know I can add it here.

src/auth/src/credentials/external_account.rs
harkamaljot: yeah having a default here is better, I added a default here for now.

src/auth/Cargo.toml
```diff
@@ -38,6 +38,8 @@ thiserror.workspace   = true
 time                  = { workspace = true, features = ["serde"] }
 tokio                 = { workspace = true, features = ["fs", "process"] }
 bon.workspace         = true
+derive_builder        = { workspace = true, features = ["alloc"] }
```
alvarowolfx: why is this change needed ?

src/auth/src/credentials/external_account.rs
```diff
@@ -127,13 +121,17 @@ impl From<CredentialSourceFile> for CredentialSource {
     }
 }
 
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, Builder)]
+#[builder(setter(into))]
```
coryan: We already use `bon` in this crate, why a second  crate for builders?  Candidly, I think external crates for this are a liability and we should just write the code ourselves.  It is not that hard.

coryan: Maybe this is the same question that @alvarowolfx is asking.

harkamaljot: I tried using bon in the beginning however `bon` does a compile time checks therefore wasn't suitable here(https://bon-rs.com/guide/alternatives) as build method's can't panic. But yeah if we do not prefer to include any other dependency I can write a builder here.

src/auth/Cargo.toml
harkamaljot: I was using it to auto-generate the builder, but let me remove it and add a manual builder for the config.

src/auth/src/credentials/external_account.rs
```diff
@@ -139,13 +132,83 @@ struct ExternalAccountConfig {
     credential_source: CredentialSource,
 }
 
+#[derive(Debug, Default)]
+struct ExternalAccountConfigBuilder {
```
sai-sunder-s, line 136: You could do in a follow up PR. You need to add support for impersonation url as well. It was introduced in #2538

src/auth/src/credentials/external_account.rs
sai-sunder-s: We are adding for all public builder methods, irrespective of the level of difficulty. See 

https://github.com/googleapis/google-cloud-rust/blob/b5c298dd927ff5a20775e5b898f59c394612c51d/src/auth/src/credentials/user_account.rs#L176-L186

src/auth/src/credentials/external_account.rs
```diff
@@ -752,4 +1001,45 @@ mod test {
...
+    #[tokio::test]
+    async fn create_programmatic_builder() {
+        let provider = Arc::new(TestSubjectTokenProvider);
+        let creds = ProgrammaticBuilder::new(provider)
+            .with_audience("test-audience")
+            .with_subject_token_type("test-token-type")
+            .with_client_id("test-client-id")
+            .with_client_secret("test-client-secret")
+            .with_quota_project_id("test-quota-project")
+            .with_scopes(vec!["scope1", "scope2"])
```
sai-sunder-s: with_token_url is not called here.

src/auth/tests/credentials.rs
```diff
@@ -320,6 +330,57 @@ mod test {
...
+        })
+        .to_string();
+
+        let server = Server::run();
+        server.expect(
+            Expectation::matching(all_of![
+                request::method_path("GET", "/source_token"),
+                request::headers(contains(("metadata", "True",))),
+            ])
+            .respond_with(status_code(429).body(source_token_response_body)),
```
sai-sunder-s: any reason to pick `429` as the error code? Typically a `429` response is for throttling and the error would not be `invalid_token`.

src/auth/src/credentials/external_account.rs
harkamaljot: sounds good, I will do the next PR.

src/auth/tests/credentials.rs
harkamaljot: changed it 400 to reflect bad request.

alvarowolfx (approved):

sai-sunder-s (approved):

src/auth/src/credentials/external_account.rs
```diff
@@ -752,4 +1241,55 @@ mod test {
...
+        if let Some(scopes) = scopes {
+            builder = builder.with_scopes(scopes);
+        }
+        if let Some(token_url) = token_url {
+            builder = builder.with_token_url(token_url);
+        }
+
+        let creds = builder.build().unwrap();
+
+        let fmt = format!("{creds:?}");
```
dbolduc, line 1268: We should also use `creds` to get a token and verify it is `"test-subject-token"`

src/auth/src/credentials/external_account.rs
```diff
@@ -139,13 +132,83 @@ struct ExternalAccountConfig {
...
+    }
+}
+
 #[derive(Debug, Clone)]
 #[allow(dead_code)]
 enum CredentialSource {
     Url(UrlSourcedCredentials),
     Executable(ExecutableSourcedCredentials),
     File {},
     Aws {},
```
dbolduc, line 210: nit, predating this PR: braces here are unnecessary

harkamaljot:
> GOOGLE_CLOUD_QUOTA_PROJECT

According to the guidance in [AIP-4110](https://google.aip.dev/auth/4110), this variable is intended to specify a quota project for Application Default Credentials (ADC). I've also reviewed the [Java implementation](https://github.com/googleapis/google-auth-library-java/blob/f793c20be7f2f0bd4c112dda45544015d878a083/oauth2_http/java/com/google/auth/oauth2/DefaultCredentialsProvider.java#L58) of this, and it confirms that `GOOGLE_CLOUD_QUOTA_PROJECT` is only used in the ADC workflow.

Currently, our `api_key_credentials` builder also reads this environment variable. So it shouldn't be reading this env variable value.

dbolduc:
Ack, thanks for checking

src/auth/src/credentials/external_account.rs
harkamaljot: thanks! let me send a PR to clean this up.

src/auth/src/credentials/external_account.rs
harkamaljot: thanks! let me send a PR to clean this up.

PR #2542, opened 2025-06-27
impl: idempotency, empty body per binding

dbolduc:
Part of the work for #2317 

Additional bindings can have different HTTP methods (e.g. "GET" vs. "POST"). This means the default idempotency needs to be set per binding. This also affects whether we should include a payload with the request.

Move that logic into `gaxi`. Moving logic from generation time to runtime, is not typically what you want, but this stuff is lightweight, so I am doing it.

---

I wanted to remove the concept of `default_idempotency` from `RequestOptions` altogether, but ...

- `reqwest::ReqwestBuilder` is given the HTTP method, but it doesn't expose it until it becomes a `reqwest::Request`.
- We need to know the idempotency when we start the retry loop.
- We create the `reqwest::Reqwest` per attempt of the retry loop (where we calculate the timeout).

So let's just keep doing what we're doing.

### DIFF
--- a/generator/internal/rust/templates/crate/src/transport.rs.mustache
+++ b/generator/internal/rust/templates/crate/src/transport.rs.mustache
@@ -68,17 +68,13 @@ impl super::stub::{{Codec.Name}} for {{Codec.Name}} {
         req: {{InputType.Codec.QualifiedName}},
         options: gax::options::RequestOptions,
     ) -> Result<gax::response::Response<{{Codec.ReturnType}}>> {
-        let options = gax::options::internal::set_default_idempotency(
-            options,
-            {{PathInfo.Codec.IsIdempotent}},
-        );
         use gax::error::binding::BindingError;
         use gaxi::path_parameter::PathMismatchBuilder;
         {{#Codec.HasBindingSubstitutions}}
         use gaxi::path_parameter::try_match;
         use gaxi::routing_parameter::Segment;
         {{/Codec.HasBindingSubstitutions}}
-        let builder = None
+        let (builder, method) = None
         {{#PathInfo.Bindings}}
         .or_else(|| {
             {{#Codec.HasVariablePath}}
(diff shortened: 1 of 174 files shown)


generator/internal/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -103,14 +99,14 @@ impl super::stub::{{Codec.Name}} for {{Codec.Name}} {
...
                 Ok(builder)
             })();
-            Some(builder)
             {{/Codec.QueryParamsCanFail}}
             {{^Codec.QueryParamsCanFail}}
             {{#Codec.QueryParams}}
             {{{Codec.AddQueryParameter}}}
             {{/Codec.QueryParams}}
-            Some(Ok(builder))
+            let builder = Ok(builder);
```
coryan, line 107: The output from this is quite odd. It may be better to duplicate some code in the mustache template to produce more natural code. 🤷

generator/internal/rust/templates/crate/src/transport.rs.mustache
```diff
@@ -68,17 +68,13 @@ impl super::stub::{{Codec.Name}} for {{Codec.Name}} {
         req: {{InputType.Codec.QualifiedName}},
         options: gax::options::RequestOptions,
     ) -> Result<gax::response::Response<{{Codec.ReturnType}}>> {
-        let options = gax::options::internal::set_default_idempotency(
-            options,
-            {{PathInfo.Codec.IsIdempotent}},
```
coryan, line 73: 😢

coryan (approved):

generator/internal/rust/templates/crate/src/transport.rs.mustache
dbolduc: why cry?

coryan: The generation-time vs. runtime thing.

dbolduc: Option A: return the precomputed `(builder, default_idempotency, body)`

Option B: return `(builder, http_method)` then compute the default idempotency and the NoBody in some cases

Which do you prefer? I am happy to do either

coryan: Option C: leave as-is and we can fix this later.

dbolduc: Sounds like B -> A.

Ack, thanks for the opinion.

PR #2545, opened 2025-06-27
docs: explain binding errors

### LINKED ISSUE #2317: Support multiple additional bindings for generated clients

dbolduc:
Fixes #2317  (there are some cleanups we should do, but this closes out everything that is public-facing).

Add a user guide section on handling binding errors. My main goal is to explain what is a `*`.

It renders like this: [ug_binding_errors.pdf](https://github.com/user-attachments/files/20953752/ug_binding_errors.pdf)

### DIFF
--- a/guide/src/binding_errors.md
+++ b/guide/src/binding_errors.md
@@ -102,7 +102,6 @@ Also, when a template ends in `/**`, that initial slash is optionally included.
 If you need to inspect the error programmatically, you can do so by checking
 that it is a binding error, then downcasting it to a `BindingError`.
 
-
 ```rust,ignore
 {{#include ../samples/src/binding_errors.rs:inspect}}
 ```
--- a/guide/src/binding_errors.md
+++ b/guide/src/binding_errors.md
@@ -20,19 +20,28 @@ You might have tried to make a request and run into an error that looks like:
 
 ```norust
 Error: cannot find a matching binding to send the request: at least one of the
-conditions must be met: (1) field `name` needs to be set and match:
-'projects/*/secrets/*' OR (2) field `name` needs to be set and match:
-'projects/*/locations/*/secrets/*'
+conditions must be met: (1) field `name` needs to be set and match the template:
+'projects/*/secrets/*' OR (2) field `name` needs to be set and match the
+template: 'projects/*/locations/*/secrets/*'
 ```
 
-This is a binding error. Let's break this down.
+This guide will explain how to troubleshoot this type of error.
 
 ## When this happens
 
-When a client cannot match the request to a [URI] for the service, it will fail
-the request locally with a binding error.
+The Google Cloud Client Libraries for Rust primarily use HTTP to send requests
+to Google Cloud services.
 
-Typically this happens when a field is either missing, or in an invalid format.
+Some RPCs correspond to multiple [URI]s. The contents of the request determine
+which URI is used.
+
+The client library considers all possible URIs, and only returns a binding error
+if no URIs work. Typically this happens when a field is either missing, or in an
+invalid format.
+
+The example error above was produced by trying to get a resource, without naming
+the resource. Specifically, the `name` field on a [`GetSecretRequest`] was
+required, but not set.
 
 ```rust,ignore
 {{#include ../samples/src/binding_errors.rs:request}}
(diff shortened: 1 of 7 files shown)


guide/src/binding_errors.md
```diff
@@ -0,0 +1,109 @@
...
+You might have tried to make a request and run into an error that looks like:
+
+```norust
+Error: cannot find a matching binding to send the request: at least one of the
+conditions must be met: (1) field `name` needs to be set and match:
+'projects/*/secrets/*' OR (2) field `name` needs to be set and match:
+'projects/*/locations/*/secrets/*'
+```
+
+This is a binding error. Let's break this down.
```
coryan: Maybe:

```suggestion
This guide will explain how to troubleshoot this type of error.
```

guide/src/binding_errors.md
```diff
@@ -0,0 +1,109 @@
...
+
+## When this happens
+
+When a client cannot match the request to a [URI] for the service, it will fail
```
coryan: This assumes the customer knows the client is using HTTP and JSON. Maybe:

```suggestion
The Google Cloud Client Libraries for Rust primarily use HTTP to send requests to Google Cloud services.
The client libraries inspect the request parameters to find a matching [URI] for the service, it will fail
```

I feel we have not explained what the request parameters or request are in this context, so think if we can introduce those or we have an existing intro.

guide/src/binding_errors.md
```diff
@@ -0,0 +1,109 @@
...
+
+OR
+
+```rust,ignore
+{{#include ../samples/src/binding_errors.rs:request-success-2}}
+```
+
+## Interpreting the templates
+
+In the template strings, there are two special matchers.
```
coryan: Maybe introduce what the "template strings" are:

```suggestion
The error message includes a number of "template strings" showing possible values for the request fields.
Most template strings, use `*` and `**` as wildcards to match the field values.
```

guide/src/binding_errors.md
dbolduc: Done.

guide/src/binding_errors.md
dbolduc: You're right. I forgot that the audience might not have context. Fixed.

guide/src/binding_errors.md
dbolduc: Good call, done.

I also started calling them "template" in the error message.

guide/src/binding_errors.md
```diff
@@ -0,0 +1,121 @@
...
+OR
+
+```rust,ignore
+{{#include ../samples/src/binding_errors.rs:request-success-2}}
+```
+
+## Interpreting the templates
+
+The error message includes a number of "template strings" showing possible
+values for the request fields. Most template strings, use `*` and `**` as
```
coryan: nit:

```suggestion
values for the request fields. Most template strings include `*` and `**` as
```

guide/src/binding_errors.md
```diff
@@ -0,0 +1,121 @@
...
+wildcards to match the field values.
+
+### Single wildcard
+
+The `*` alone means: a non-empty string without a `/`. It can be thought of as
```
coryan:
```suggestion
The `*` wildcard alone means: a non-empty string without a `/`. It can be thought of as
```

guide/src/binding_errors.md
```diff
@@ -0,0 +1,121 @@
...
+| `"*"`                      | `"string/with/slashes"`       | `false` |
+| `"projects/*"`             | `"projects/"` (empty)         | `false` |
+| `"projects/*"`             | `"projects/p/"` (extra slash) | `false` |
+| `"projects/*"`             | `"projects/p/locations/l"`    | `false` |
+| `"projects/*/locations"`   | `"projects/p"`                | `false` |
+| `"projects/*/locations"`   | `"projects/p/locations/l"`    | `false` |
+
+### Double wildcard
+
+Less common is the `**`, which means: any string. The string can be empty or
```
coryan:
```suggestion
Less common is the `**` wildcard, which means: any string. The string can be empty or
```

coryan (approved):

guide/src/binding_errors.md
dbolduc: Done

guide/src/binding_errors.md
dbolduc: Done

guide/src/binding_errors.md
dbolduc: Done

PR #2546, opened 2025-06-27
test(storage): use a larger file for chunking read

suzmue:
Make the test more realistic and use a larger object. Chose a size that returned 4 chunks when running the integration tests locally.

While it would be easier to create a `const CONTENTS: &[u8] = &[5_u8; TOTAL_BYTES]`, it would be harder to verify that the data actually matched, since downloading the first chunk repeatedly would yield the same result.

### DIFF
--- a/src/integration-tests/src/storage.rs
+++ b/src/integration-tests/src/storage.rs
@@ -62,18 +62,6 @@ pub async fn objects(builder: storage::client::ClientBuilder) -> Result<()> {
     assert_eq!(contents, CONTENTS.as_bytes());
     tracing::info!("success with contents={contents:?}");
 
-    tracing::info!("testing read_object() streaming");
-    let mut contents = bytes::BytesMut::new();
-    let mut resp = client
-        .read_object(&bucket.name, &insert.name)
-        .send()
-        .await?;
-    while let Some(chunk) = resp.next().await? {
-        contents.extend_from_slice(&chunk);
-    }
-    assert_eq!(contents, CONTENTS.as_bytes());
-    tracing::info!("success with contents={contents:?}");
-
     control
         .delete_object()
         .set_bucket(&insert.bucket)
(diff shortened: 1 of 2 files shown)


src/integration-tests/src/storage.rs
```diff
@@ -149,6 +137,71 @@ pub async fn objects_customer_supplied_encryption(
...
+        .read_object(&bucket.name, &insert.name)
+        .send()
+        .await?;
+
+    // This should take multiple chunks to download.
+    let mut got = bytes::BytesMut::new();
+    while let Some(chunk) = resp.next().await? {
+        got.extend_from_slice(&chunk);
+    }
+    assert_eq!(got, contents);
```
coryan: So maybe count the chunks?

```suggestion
    // This should take multiple chunks to download.
    let mut count = 0;
    let mut got = bytes::BytesMut::new();
    while let Some(chunk) = resp.next().await? {
        got.extend_from_slice(&chunk);
        count += 1;
    }
    assert_eq!(got, contents);
    assert!(count > 1, "{count:?}");
```

src/integration-tests/src/storage.rs
```diff
@@ -149,6 +137,71 @@ pub async fn objects_customer_supplied_encryption(
...
+
+    let client = builder.build().await?;
+
+    // Create a large enough file that will require multiple chunks to download.
+    const TOTAL_BYTES: usize = 4000;
+    const BLOCK_SIZE: usize = 500;
+    let mut contents: Vec<u8> = vec![0; TOTAL_BYTES];
+    for i in 0..TOTAL_BYTES {
+        contents[i] = (i / BLOCK_SIZE) as u8;
+    }
```
coryan: optional: I probably would have written something like:

```suggestion
    const BLOCK_SIZE: usize = 500;
    let mut contents = Vec::new();
    for i in 0..8 {
        contents.extend_from_slice(&[i as u8; BLOCK_SIZE]);
    }
```

If I was feeling ~clever~ too clever I would write

```
    let contents = (0..8).fold(Vec::new(), |mut v, i| { v.extend_from_slice(&[i as u8; BLOCK_SIZE]); v });
```

src/integration-tests/src/storage.rs
suzmue: Sounds good. I was torn on whether checking this explicitly was worth it, since it failing wouldn't mean a problem with the code, but with the test. I agree that its worth having a signal that the test isn't testing what we expect it to.

src/integration-tests/src/storage.rs
suzmue: Done! Thanks for the suggestion

> If I was feeling clever too clever I would write

😆

src/integration-tests/src/storage.rs
```diff
@@ -149,6 +137,73 @@ pub async fn objects_customer_supplied_encryption(
...
+    assert!(count > 1, "{count:?}");
+    tracing::info!("success with large contents");
+
+    control
+        .delete_object()
+        .set_bucket(&insert.bucket)
+        .set_object(&insert.name)
+        .set_generation(insert.generation)
+        .send()
+        .await?;
```
dbolduc, line 197: nit: can we skip this and just delete the bucket?

src/integration-tests/src/storage.rs
```diff
@@ -149,6 +137,73 @@ pub async fn objects_customer_supplied_encryption(
...
+            .with_level(true)
+            .with_thread_ids(true)
+            .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+            .finish();
+
+        tracing::subscriber::set_default(subscriber)
+    };
+
+    // Create a temporary bucket for the test.
+    let (control, bucket) = create_test_bucket().await?;
```
dbolduc, line 156: For the future, and predates this PR: I don't like how if any of the manipulations fail, we exit this scope before cleaning up the resources. Ideally we `defer` the cleanup, but this is Rust, not go.

We `cleanup_stale_buckets`, so it's not the biggest deal. But we could be stricter about our resource management.

I don't think we can return a handle with an `impl Drop for BucketHandle`, because our delete function is async, and `drop` is not.

A simple way around this is to just put everything in a context:

```rs
    let (control, bucket) = create_test_bucket().await?;

    let result = async move {
      // do all the things
      let insert = client.insert_object(...)
          .send()
          .await?; // Use `?`, we can chain in this context.
      let mut resp = client.read_object(...)
          .send()
          .await?; // Use `?`, we can chain in this context.
      // etc.
      Ok(())
    }.await; // No `?`, we have more work to do even if this fails.

    let _ = control.delete_bucket()
        .set_name(bucket_name)
        .send()
        .await; // No `?`, we want to return `result`.

    result
```

and maybe if this is common enough, we abstract this into a class that can run an `AsyncFn` in a bucket context. That might make the above code less ugly.

dbolduc (approved):
Just something to think about for the future, tangential to this PR.

src/integration-tests/src/storage.rs
coryan: You must delete all the objects in a bucket (manually or through some lifecycle policy) before you can delete the bucket.

src/integration-tests/src/storage.rs
suzmue: I was noticing the same. Created #2561

src/integration-tests/src/storage.rs
dbolduc: Lame.

PR #2558, opened 2025-06-30
cleanup(generator): weaker typed variable subsegments

dbolduc:
Related to #557 

Just represent a variable subsegment as a string, instead of a strongly typed variant. e.g. `{"projects", "*"}`

This is what we do with explicit routing, might as well be consistent. The code is a littler simpler too. There is just less structure in the model.


coryan (approved):

coryan:
Approved, even though it makes the type-system gods angry.

PR #2560, opened 2025-06-30
feat(storage)!: change return type for next()

suzmue:
Make the method signature more similar to futures::StreamExt. It is simple to convert between the two with `transpose()` so this will not greatly affect the usability of the API. We can also consider adding a try_next helper to return a Result<Option<Bytes>> as well.

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -737,13 +737,13 @@ impl ReadObjectResponse {
     ///     .send()
     ///     .await?;
     ///
-    /// while let Some(next) = resp.next().await? {
-    ///     println!("next={next:?}");
+    /// while let Some(next) = resp.next().await {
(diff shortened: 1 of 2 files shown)


src/storage/src/client.rs
```diff
@@ -737,13 +737,13 @@ impl ReadObjectResponse {
     ///     .send()
     ///     .await?;
     ///
-    /// while let Some(next) = resp.next().await? {
-    ///     println!("next={next:?}");
+    /// while let Some(next) = resp.next().await {
+    ///     println!("next={:?}", next?);
```
coryan, line 741: Consider writing this as:

```suggestion
    /// while let Some(next) = resp.next().await.transpose()? {
    ///     println!("next={next:?}");
```

Issue #2557, opened 2025-06-30
Object names should get URL Encoded


suzmue:
The following code is not able to download the object:
```rust
    let contents = client
        .read_object(
            "projects/_/buckets/gcp-public-data-arco-era5",
            "my/object.path",
        )
```
But this does work:
```rust
    let contents = client
        .read_object(
            "projects/_/buckets/gcp-public-data-arco-era5",
            "my%2Fobject.path",
        )
```

We should URL encode the object names.

dbolduc:
You probably already know this, but in case you haven't seen it, there is an example of percent encoding here:

https://github.com/googleapis/google-cloud-rust/blob/dcbb9481857e314e9c6501ed81b5bc6e351006bb/src/gax-internal/src/routing_parameter.rs#L171

PR #2562, opened 2025-07-01
fix(storage): percent encode object names in read and insert requests

suzmue:
For #2557

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -345,7 +345,7 @@ impl InsertObject {
                 format!("{}/upload/storage/v1/b/{bucket_id}/o", &self.inner.endpoint),
             )
             .query(&[("uploadType", "media")])
-            .query(&[("name", object)])
+            .query(&[("name", enc(object))])
             .header("content-type", "application/octet-stream")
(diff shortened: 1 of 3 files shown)


src/storage/src/client.rs
```diff
@@ -1096,6 +1133,53 @@ mod tests {
         Ok(())
     }
 
+    #[test_case("projects/p", "projects%2Fp")]
+    #[test_case("kebab-case", "kebab-case")]
+    #[test_case("dot.name", "dot.name")]
+    #[test_case("under_score", "under_score")]
+    #[test_case("tilde~123", "tilde~123")]
+    #[test_case("exclamation%21point%21", "exclamation%21point%21")]
```
coryan: This is also interesting, but maybe if it was called "preserve%21percent%21" would be easier to grok?

Should we have something like `#[test_case("test each case !#$&'()*+,/:;=?@[] ", "test each case %......")]` ?

suzmue: Done!

coryan (dismissed):

coryan (approved):

PR #2565, opened 2025-07-01
feat(auth): Enable impersonation in ProgrammaticBuilder

harkamaljot:
This change introduces the ability to use service account impersonation with external account credentials that are created using the ProgrammaticBuilder.

Previously, the ProgrammaticBuilder was missing a method to specify the service_account_impersonation_url, which prevented its use in scenarios requiring service account impersonation.

 This commit addresses the limitation by:
   - Adding the with_service_account_impersonation_url method to the ProgrammaticBuilder.
   - Adding a comprehensive unit test to verify that the end-to-end impersonation flow is working correctly when configured through the ProgrammaticBuilder.

### DIFF
--- a/src/auth/src/credentials/external_account.rs
+++ b/src/auth/src/credentials/external_account.rs
@@ -187,11 +187,19 @@ impl ExternalAccountConfigBuilder {
 
     fn build(self) -> BuildResult<ExternalAccountConfig> {
         Ok(ExternalAccountConfig {
-            audience: self.audience.ok_or(BuilderError::missing_field("audience"))?,
-            subject_token_type: self.subject_token_type.ok_or(BuilderError::missing_field("subject_token_type"))?,
-            token_url: self.token_url.ok_or(BuilderError::missing_field("token_url"))?,
+            audience: self
+                .audience
+                .ok_or(BuilderError::missing_field("audience"))?,
+            subject_token_type: self
+                .subject_token_type
+                .ok_or(BuilderError::missing_field("subject_token_type"))?,
+            token_url: self
+                .token_url
+                .ok_or(BuilderError::missing_field("token_url"))?,
             scopes: self.scopes.ok_or(BuilderError::missing_field("scopes"))?,
-            credential_source: self.credential_source.ok_or(BuilderError::missing_field("credential_source"))?,
+            credential_source: self
+                .credential_source
+                .ok_or(BuilderError::missing_field("credential_source"))?,
             service_account_impersonation_url: self.service_account_impersonation_url,
             client_id: self.client_id,
             client_secret: self.client_secret,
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/external_account.rs
```diff
@@ -825,6 +831,42 @@ impl ProgrammaticBuilder {
...
+    /// #     type Error = MyProviderError;
+    /// #     async fn subject_token(&self) -> Result<SubjectToken, Self::Error> {
+    /// #         Ok(SubjectTokenBuilder::new("my-programmatic-token".to_string()).build())
+    /// #     }
+    /// # }
+    /// # let provider = Arc::new(MyTokenProvider);
+    /// let builder = ProgrammaticBuilder::new(provider)
+    ///     .with_service_account_impersonation_url("my-service-account-impersonation-url");
+    /// ```
+    pub fn with_service_account_impersonation_url<S: Into<String>>(mut self, url: S) -> Self {
```
sai-sunder-s: We should get `target_principal` instead of the whole url. 
1. It is unnecessary for the user to figure out the whole url
2. when we introduce universe domain, we can incorporate it into the url template easily

Check out impersonated builder for reference: https://github.com/googleapis/google-cloud-rust/blob/f8208dd3d008c3a33b06fb3011603868b9f94141/src/auth/src/credentials/impersonated.rs#L180

src/auth/src/credentials/external_account.rs
```diff
@@ -860,9 +862,13 @@ impl ProgrammaticBuilder {
     /// # }
     /// # let provider = Arc::new(MyTokenProvider);
     /// let builder = ProgrammaticBuilder::new(provider)
-    ///     .with_service_account_impersonation_url("my-service-account-impersonation-url");
+    ///     .with_target_principal("test-principal");
     /// ```
-    pub fn with_service_account_impersonation_url<S: Into<String>>(mut self, url: S) -> Self {
+    pub fn with_target_principal<S: Into<String>>(mut self, target_principal: S) -> Self {
+        let url = format!(
+            "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{}:generateAccessToken",
```
sai-sunder-s, line 869: move this to a constant and use the same constant in impersoanted cred as well

sai-sunder-s (approved):

src/auth/src/credentials/external_account.rs
harkamaljot: format! macro doesn't work with const variables, so we would have to use replace function. Here now replace would be a runtime operation hence I avoided that. We can create a macro in the const file to create this url, however for just 2 usages it doesn't seem worth it, we can do that if any other usage of the same url pops up.

PR #2566, opened 2025-07-01
feat(internal/gcloud): add structs for gcloud config YAML

julieqiu:
Introduce Go structs for parsing a sample gcloud config YAML file. This is a first step to evaluate if we can use sidekick for automating gcloud command generation.

### DIFF
--- a/generator/internal/gcloud/gcloud.go
+++ b/generator/internal/gcloud/gcloud.go
@@ -1,0 +1,222 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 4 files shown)


generator/internal/gcloud/gcloud.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// ResourcePatterns describes resource patterns not included in
+	// descriptors, providing additional patterns that might be used for
+	// resource identification or command generation.
+	ResourcePatterns []ResourcePattern `yaml:"resource_patterns,omitempty"`
+}
+
+// API describes an API to generate a surface for. This structure holds
+// configurations specific to a single API within the gcloud surface.
+type API struct {
```
coryan, line 34: We already have:

https://github.com/googleapis/google-cloud-rust/blob/615eaeb45cae68bc53fb278f09c33af451d86eaf/generator/internal/api/model.go#L128

Which may prove confusing.  If you can think of a different name consider changing it.

coryan (approved):

Issue #2559, opened 2025-06-30
Generate `managedkafka.schemaregistry.v1`


dbolduc:
https://github.com/googleapis/googleapis/tree/master/google/cloud/managedkafka/schemaregistry/v1

We need to update googleapis SHA, then can generate with:

```rs
library="managedkafka/schemaregistry"
yaml="google/cloud/${library}/v1/managedkafka_v1.yaml"
```

https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-generated-code-maintenance.md

whuffman36:
Merged #2576 to generate managedkafka.schemaregistry.v1

PR #2569, opened 2025-07-01
chore: update googleapis circa 2025-06-01

dbolduc:
Motivated by #2559 

Use the documentation overrides from #2473 to handle typos. The benefit of this over `.typos.toml` is that
- we are notified when the typos are fixed
- the typos are fixed in the public docs


dbolduc:
> I am beginning to think that excluding `src/generated/**` from typos may be a better solution.

I support this.

Less work for us, but worse documentation overall. We do Cloud Rust, not Google Cloud documentation though.

Excluding would make it easier to automate this stuff.

PR #2579, opened 2025-07-02
cleanup(auth): Parameterize project-id in integration tests

harkamaljot:


dbolduc:
nit: I only use `fix:` for fixes that are customer facing that we want to surface in the changelog.

Even if I am fixing something internal, I call that a `cleanup` or an `impl`.

alvarowolfx (approved):

Issue #2043, opened 2025-05-08
Support resumable uploads


coryan:
We need a way to perform resumable uploads. Some things to keep in mind:

- We should design an API that works with [BidiWriteObject], that RPC supports pipelining, which may be better than HTTP POST + PUT
- We need to be able to resume uploads, so applications should be able to save the upload id.
- We need to be able to cancel uploads.
- Streaming uploads need tracing for the full upload.

[BidiWriteObject]: https://github.com/googleapis/googleapis/blob/b4cafb0a8e0f5a72baffe3b6e29840182da313fd/google/storage/v2/storage.proto#L386-L387

coryan:
I think this is done. We are going to post-pone some work: #2633.

PR #2585, opened 2025-07-03
impl(storage): serialize `Object` for uploads

coryan:
When inserting new objects the metadata can be included in the intial
POST request (for resumable uploads), or in the first MIME-multipart
section (for multipart uploads).

Only a subset of the fields may be included in this initial upload, and
some fields require transformations from the gRPC to the JSON data
models.

This change introduces the function to serialize via a
`serde_json::Value`. We may optimize this further in a future set of
changes.

Part of the work for #2043

### DIFF
--- a/src/storage/src/client/v1.rs
+++ b/src/storage/src/client/v1.rs
@@ -244,9 +244,114 @@ impl From<Object> for control::model::Object {
     }
 }
 
+/// Create a JSON object for the [control::model::Object] fields used in uploads.
+///
+/// When uploading (aka inserting) an object number of metadata fields can be
+/// sent via the POST body.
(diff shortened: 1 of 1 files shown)


src/storage/src/client/v1.rs
```diff
@@ -244,9 +244,114 @@ impl From<Object> for control::model::Object {
     }
 }
 
+/// Create a JSON object for the [control::model::Object] fields used in uploads.
+///
+/// When uploading (aka inserting) an object number of metadata fields can be
```
dbolduc: any number of?

dbolduc (dismissed):
Ack on the PR description for why we cannot just serialize the message with serde

PR #2589, opened 2025-07-06
cleanup: helpful errors for mustache substitutions

dbolduc:
If mustache cannot look something up, it will surface an error instead of printing an empty string.

This doesn't seem to apply to `#`/`^` blocks. So you can still gate code blocks:
```
{{#SkipThisBlock}}
{{#Foos}}
  {{#Bars}}
    // Baz = {{Baz}}
  {{/Bars}}
{{/Foos}}
{{/SkipThisBlock}}
```


dbolduc:
Allow me to demonstrate how strong my convictions are about this change.

PR #2590, opened 2025-07-06
refactor(storage): rename upload source traits

coryan:
Separate the `Seek` trait so we can create functions that only require
`StreamingSource`.

Part of the work for #2043

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -14,7 +14,7 @@
 
 pub use crate::Error;
 pub use crate::Result;
-use crate::upload_source::{InsertPayload, StreamingSource};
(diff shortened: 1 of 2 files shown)


src/storage/src/client.rs
```diff
@@ -136,7 +136,7 @@ impl Storage {
         B: Into<String>,
         O: Into<String>,
         T: Into<InsertPayload<P>>,
-        InsertPayload<P>: StreamingSource,
+        InsertPayload<P>: StreamingSource + Seek,
```
dbolduc, line 139: Ah, clever

dbolduc (approved):

PR #2591, opened 2025-07-06
impl(storage-control): auto-populate request IDs

### LINKED ISSUE #439: Automatically populate fields in the request message (google.aip.dev/4235)

dbolduc:
Fixes #439 

Adds support for auto-populating request ID fields over both HTTP and gRPC.

At the moment, only storage-control is impacted. We use that in our tests.

I noticed that paginated requests should each have a different request ID. While not explicitly stated in [AIP-4235](https://google.aip.dev/client-libraries/4235), I think it can be inferred from:
> Client libraries must reuse automatically populated values for retries of the same request. In other words, the automatically populated fields must not be regenerated for each RPC attempt with the same request message.

## Notes

- The builder template could be simplified to always say `let req = self.0.request;`. I didn't do it just to keep the diff down.
- I am taking some TODOs on cleaning up idempotency handling.
- Testing one request ID per client operation is hard
  - Mocks do not have retry loops, so they are insufficient.
  - gapic-showcase only returns the request ID if a request is successful. That is insufficient to check that the same request ID is sent with each attempt.
  - We'd have to spin up a server.
  - I settled for an argument: If the request ID is set outside of the retry loop, we can assume it is the same for the whole retry loop. :shrug: 
- While gapic-showcase is again underwhelming, if we ever moved the auto-population from the builder -> transport, we would not have to rewrite the test. We would have to rewrite the tests involving mock stubs.

### DIFF
--- a/generator/internal/rust/templates/crate/src/builder.rs.mustache
+++ b/generator/internal/rust/templates/crate/src/builder.rs.mustache
@@ -149,7 +149,13 @@ pub mod {{Codec.ModuleName}} {
                 traits (e.g. `std::clone::Clone`) making the call ambiguous.
                 Using `*self.0.stub` makes the call not-ambiguous.
             }}
+            {{#HasAutoPopulatedFields}}
+            let req = Self::auto_populate(self.0.request, false);
+            (*self.0.stub).{{Codec.Name}}(req, self.0.options).await.map(gax::response::Response::into_body)
+            {{/HasAutoPopulatedFields}}
+            {{^HasAutoPopulatedFields}}
             (*self.0.stub).{{Codec.Name}}(self.0.request, self.0.options).await.map(gax::response::Response::into_body)
+            {{/HasAutoPopulatedFields}}
         }
         {{#Pagination}}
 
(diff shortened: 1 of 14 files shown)


generator/internal/rust/templates/crate/src/builder.rs.mustache
```diff
@@ -164,6 +170,19 @@ pub mod {{Codec.ModuleName}} {
...
+                    If we have a new page token, this is a new request. New
+                    requests need new request IDs.
+                }}
+                {{^Optional}}
+                let initial = builder.0.request.{{Codec.FieldName}} == token;
+                {{/Optional}}
+                {{#Optional}}
+                let initial = builder.0.request.{{Codec.FieldName}}.as_deref().is_none_or(|s| s == token);
+                {{/Optional}}
+                builder.0.request = Self::auto_populate(builder.0.request, !initial);
```
dbolduc, line 184: Instead of passing in `force`, we could clear the request ID field in the pagination case. Then we let `send()` do the work. That might be cleaner.

PR #2592, opened 2025-07-07
refactor(storage): split request builders

coryan:
The `client.rs` file was getting too long, and it is bound to get longer
as we add more tests to each request builder.

Part of the work for #2043

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -426,8 +426,6 @@ fn apply_customer_supplied_encryption_headers(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use futures::TryStreamExt;
-    use httptest::{Expectation, Server, matchers::*, responders::status_code};
     use std::sync::Arc;
     use test_case::test_case;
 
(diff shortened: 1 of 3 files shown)


src/storage/src/client.rs
```diff
@@ -125,20 +129,25 @@ impl Storage {
...
     /// let response = client
-    ///     .insert_object("projects/_/buckets/my-bucket", "my-object", "the quick brown fox jumped over the lazy dog")
+    ///     .insert_object("projects/_/buckets/my-bucket", "my-object", "hello world")
     ///     .send()
     ///     .await?;
     /// println!("response details={response:?}");
     /// # Ok::<(), anyhow::Error>(()) });
     /// ```
-    pub fn insert_object<B, O, T, P>(&self, bucket: B, object: O, payload: T) -> InsertObject<P>
+    pub fn insert_object<B, O, T, P>(
```
dbolduc, line 138: Q: should this method also get renamed? (in a follow up)

coryan: Yes, it will.

PR #2597, opened 2025-07-07
cleanup(auth): add test for programmatic credential

harkamaljot:
As a follow up on https://github.com/googleapis/google-cloud-rust/pull/2541 I added another unit test in programmatic sourced credentials and remove the extra braces in the credential sources


dbolduc (dismissed):

dbolduc (approved):
Thanks

PR #2605, opened 2025-07-08
feat(auth): Add RetryTokenProvider wrapper

harkamaljot:
Add a RetryTokenProvider wrapper module. Currently this module is not used anywhere. In upcoming change we will modify the credentials to wrap their token providers in the given mod.


src/auth/src/retry.rs
```diff
@@ -0,0 +1,481 @@
...
+        retry_loop(
+            move |_| {
+                let inner = inner.clone();
+                async move {
+                    inner
+                        .token()
+                        .await
+                        .map_err(gax::error::Error::authentication)
+                }
+            },
```
sai-sunder-s: Can we put this into a callback method to make it easier to understand

src/auth/src/retry.rs
```diff
@@ -0,0 +1,481 @@
...
+
+    pub(crate) fn with_backoff_policy(mut self, backoff_policy: Arc<dyn BackoffPolicy>) -> Self {
+        self.backoff_policy = Some(backoff_policy);
+        self
+    }
+
+    pub(crate) fn with_retry_throttler(mut self, retry_throttler: SharedRetryThrottler) -> Self {
+        self.retry_throttler = Some(retry_throttler);
+        self
+    }
```
sai-sunder-s, line 66: Do we need a builder for this internal token provider. We could collect all these details in an internal struct and pass that along?

src/auth/src/retry.rs
```diff
@@ -0,0 +1,481 @@
...
+        assert_eq!(token.token, "test_token");
+    }
+
+    #[tokio::test]
+    async fn test_no_retry_policy_failure() {
+        let mut mock_provider = MockTokenProvider::new();
+        mock_provider
+            .expect_token()
+            .times(1)
+            .returning(|| Err(CredentialsError::from_msg(false, "non transient error")));
```
sai-sunder-s: Add a no_retry_policy test with transient error?

src/auth/src/retry.rs
```diff
@@ -0,0 +1,481 @@
...
+    }
+
+    #[tokio::test]
+    async fn test_full_retry_mechanism() {
+        // 1. Setup Mocks
+        let mut mock_provider = MockTokenProvider::new();
+        let mut mock_throttler = MockRetryThrottler::new();
+
+        // Token provider fails once, then succeeds.
+        let mut seq = Sequence::new();
```
sai-sunder-s, line 407: you seem to be achieving same effect as sequence in other tests like `test_success_after_retry`, what do we gain with sequence here

src/auth/src/retry.rs
harkamaljot: You're right, we could use a struct. However, the builder pattern offers two main advantages here:

Cleaner Defaults: It allows us to manage default values (like the upcoming retry policy) and current default values, cleanly within the builder itself, rather than complicating the token method.

Improved Readability: A builder is more readable and self-documenting when an object requires multiple configuration parameters, which is the case here.

src/auth/src/retry.rs
harkamaljot: You've made a good point. To clarify, the Sequence serves the same crucial purpose in both tests: to strictly enforce that the failing call happens before the successful one.

The reason for having this separate test is to verify that this exact same ordered flow works correctly under a different configuration. Specifically:

test_success_after_retry verifies the retry logic with pre-defined defaults.

This test verifies the logic works correctly when a user provides custom settings for the backoff and throttler.

So, while the role of Sequence is the same, this test is necessary to validate a distinct user-facing scenario.

src/auth/src/retry.rs
harkamaljot: I followed the same pattern as other places in the code base eg: https://github.com/googleapis/google-cloud-rust/blob/main/src/gax-internal/src/http.rs#L112 to make it more cleaner.

sai-sunder-s (approved):

Issue #2608, opened 2025-07-08
File Sourced Credentials


alvarowolfx:
Support Workload Identity Federation through File Sourced Credentials.

Details on AIP-4117 - https://google.aip.dev/auth/4117#determining-the-subject-token-in-file-sourced-credentials

PR #2609, opened 2025-07-08
feat(auth): add support for file sourced credentials

alvarowolfx:
Fixes [#2608](https://github.com/googleapis/google-cloud-rust/issues/2608)


src/auth/src/credentials/external_account_sources/file_sourced.rs
```diff
@@ -0,0 +1,62 @@
...
+            file,
+            format,
+            subject_token_field_name,
+        }
+    }
+}
+
+const JSON_FORMAT_TYPE: &str = "json";
+
+impl SubjectTokenProvider for FileSourcedCredentials {
```
harkamaljot, line 49: Is it possible we can add unit tests here same as other files eg: https://github.com/googleapis/google-cloud-rust/blob/main/src/auth/src/credentials/external_account_sources/url_sourced.rs As right now test coverage tool shows file is missing coverage.

sai-sunder-s: +1. Please add tests for json type and text type file sources and some negative cases as well.

alvarowolfx: pushed tests, I was gonna send in another PR, but makes sense to send on this.

src/auth/src/credentials/external_account_sources/file_sourced.rs
```diff
@@ -60,3 +74,92 @@ impl SubjectTokenProvider for FileSourcedCredentials {
         }
     }
 }
+
+#[cfg(test)]
+mod test {
```
sai-sunder-s, line 79: A test with empty file?

sai-sunder-s (dismissed):

src/auth/src/credentials/external_account_sources/file_sourced.rs
```diff
@@ -0,0 +1,165 @@
...
+            .unwrap_or(("text".to_string(), String::new()));
+        Self {
+            file,
+            format,
+            subject_token_field_name,
+        }
+    }
+}
+
+const JSON_FORMAT_TYPE: &str = "json";
```
harkamaljot, line 47: should we move this consts file in this module?

alvarowolfx: will do that in a follow up PR

src/auth/src/credentials/external_account_sources/file_sourced.rs
alvarowolfx: added extra test cases for that

sai-sunder-s (approved):

PR #2611, opened 2025-07-08
impl(storage): checkpoint buffered uploads

coryan:
Split "large" uploads in multiple PUT calls, checkpointing each PUT will allow
us to resume the upload with a limited buffer.

Part of the work for #2043

### DIFF
--- a/src/storage/src/client/upload_object_buffered.rs
+++ b/src/storage/src/client/upload_object_buffered.rs
@@ -1581,17 +1581,29 @@ mod tests {
             .map(|i| new_line_string(i, LEN))
             .collect::<Vec<_>>()
             .join("");
-        let stream = VecStream::new(vec![bytes::Bytes::from_owner(buffer.clone()), new_line(3, LEN)]);
+        let stream = VecStream::new(vec![
+            bytes::Bytes::from_owner(buffer.clone()),
+            new_line(3, LEN),
+        ]);
         let mut payload = InsertPayload::from(stream);
 
         let remainder = None;
         let (vec, remainder) = super::next_chunk(&mut payload, remainder, 2 * LEN).await?;
         assert!(remainder.is_some());
-        assert_eq!(vec, vec![bytes::Bytes::from_owner(buffer.clone()).slice(0..(2*LEN))]);
+        assert_eq!(
+            vec,
+            vec![bytes::Bytes::from_owner(buffer.clone()).slice(0..(2 * LEN))]
+        );
 
         let (vec, remainder) = super::next_chunk(&mut payload, remainder, 2 * LEN).await?;
         assert!(remainder.is_none());
-        assert_eq!(vec, vec![bytes::Bytes::from_owner(buffer.clone()).slice((2*LEN)..), new_line(3, LEN)]);
+        assert_eq!(
+            vec,
+            vec![
+                bytes::Bytes::from_owner(buffer.clone()).slice((2 * LEN)..),
+                new_line(3, LEN)
+            ]
+        );
 
         Ok(())
     }
--- a/src/storage/src/client/upload_object_buffered.rs
+++ b/src/storage/src/client/upload_object_buffered.rs
@@ -670,9 +670,14 @@ where
         let mut remainder = None;
         let mut offset = 0_usize;
         loop {
-            let (chunk, r) = self::next_chunk(&mut self.payload, remainder, target_size).await?;
+            let (chunk, chunk_size, r) = self::next_chunk(&mut self.payload, remainder, target_size).await?;
+            let full_size = if chunk_size < target_size {
+                Some(offset + chunk_size)
+            } else {
+                None
+            };
             let (builder, chunk_size) = self
-                .partial_upload_request(upload_url, offset, chunk, target_size)
+                .partial_upload_request(upload_url, offset, chunk, chunk_size, full_size)
                 .await?;
             let response = builder.send().await.map_err(Error::io)?;
             match self::partial_upload_handle_response(response, offset + chunk_size).await? {
@@ -697,13 +702,14 @@ where
         upload_url: &str,
         offset: usize,
         chunk: VecDeque<bytes::Bytes>,
-        target_size: usize,
+        chunk_size: usize,
+        full_size: Option<usize>,
     ) -> Result<(reqwest::RequestBuilder, usize)> {
-        let chunk_size = chunk.iter().fold(0, |s, b| s + b.len());
-        let range = match chunk_size {
-            0 => format!("bytes */{offset}"),
-            n if n == target_size => format!("bytes {}-{}/*", offset, offset + n - 1),
-            n => format!("bytes {}-{}/{}", offset, offset + n - 1, offset + n),
+        let range = match (chunk_size, full_size) {
+            (0, Some(s)) => format!("bytes */{s}"),
+            (0, None) => format!("bytes */{offset}"),
+            (n, Some(s)) => format!("bytes {offset}-{}/{s}", offset + n - 1),
+            (n, None) => format!("bytes {offset}-{}/*", offset + n - 1),
         };
         let builder = self
             .inner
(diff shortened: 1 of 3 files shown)


src/storage/src/client/upload_object_buffered.rs
```diff
@@ -730,6 +790,107 @@ async fn handle_start_resumable_upload_response(response: reqwest::Response) ->
...
+                return Ok((partial, None));
+            }
+            _ => {
+                size += b.len();
+                partial.push_back(b);
+            }
+        }
+    }
+
+    while let Some(mut b) = payload.next().await.transpose().map_err(Error::io)? {
```
dbolduc: optional: there is probably an opportunity to unify the loop with the above... but the `async` makes it challenging.

I wanted to say something like `while let Some(mut b) = remainder.or_else(|| payload.next()...)` and reset remainder if we use it. but that won't work.

src/storage/src/client/upload_object_buffered.rs
```diff
@@ -730,6 +790,107 @@ async fn handle_start_resumable_upload_response(response: reqwest::Response) ->
...
+        chunk_remainder,
+    })
+}
+
+fn parse_range_end(headers: &reqwest::header::HeaderMap) -> Option<usize> {
+    let Some(range) = headers.get("range") else {
+        // A missing `Range:` header indicates that no bytes are persisted.
+        return Some(0_usize);
+    };
+    let end = std::str::from_utf8(range.as_bytes().strip_prefix(b"bytes=0-")?).ok()?;
```
dbolduc, line 900: does this always start at 0?

src/storage/src/client/upload_object_buffered.rs
```diff
@@ -658,30 +665,83 @@ where
...
+        let (chunk, target_size) = {
+            let mut chunk = VecDeque::new();
+            let mut size = 0_usize;
+            while let Some(b) = self.payload.next().await.transpose().map_err(Error::io)? {
+                size += b.len();
+                chunk.push_back(b);
+            }
+            (chunk, size)
+        };
+        let target_size = target_size.div_ceil(RESUMABLE_UPLOAD_QUANTUM);
```
dbolduc: wat.

dbolduc: I am totally confused by what `target_size` is supposed to be. I thought `target_size == RESUMABLE_UPLOAD_QUANTUM`, or `target_size = min(target_size, RESUMABLE_UPLOAD_QUANTUM)`

src/storage/src/client/upload_object_buffered.rs
```diff
@@ -658,30 +665,83 @@ where
...
             }
             None
-        }));
-        let builder = builder.body(reqwest::Body::wrap_stream(stream));
+        });
+        Ok((builder.body(reqwest::Body::wrap_stream(stream)), chunk_size))
+    }
+
+    async fn upload_request(mut self, upload_url: String) -> Result<reqwest::RequestBuilder> {
+        let (chunk, target_size) = {
```
dbolduc: nit: call this `chunk_size`?

src/storage/src/client/upload_object_buffered.rs
```diff
@@ -1106,4 +1277,508 @@ mod tests {
         }
         Ok(())
     }
+
+    fn new_line_string(i: i32, len: usize) -> String {
+        format!("{i:022} {:width$}\n", "", width = len - 22 - 2)
```
dbolduc, line 1317: TIL `$` for named variable formatting

src/storage/src/client/upload_object_buffered.rs
```diff
@@ -1106,4 +1277,508 @@ mod tests {
...
+            .respond_with(status_code(308).append_header("range", "bytes=0-63")),
+        );
+
+        server.expect(
+            Expectation::matching(all_of![
+                request::method_path("PUT", path.clone()),
+                request::headers(contains(("content-range", "bytes 64-127/*"))),
+                request::body(chunk1.clone()),
+            ])
+            .respond_with(status_code(308).append_header("range", "bytes=0-127")),
```
dbolduc, line 1359: same Q about starting from 0

src/storage/src/client/upload_object_buffered.rs
coryan: I chose "full_size" because it is the size of the full upload.

src/storage/src/client/upload_object_buffered.rs
coryan: Add a comment to explain the terrible hack.  In the future `target_size` will be a configurable request option.

src/storage/src/client/upload_object_buffered.rs
coryan: I was struggling to figure that out earlier too. A capture does the trick.

src/storage/src/client/upload_object_buffered.rs
coryan: yes. Uploads must be sequential, and consequently the persisted range always starts at 0. I added some comments to explain this.

src/storage/src/client/upload_object_buffered.rs
coryan: Pfft, old news: I learned this like 24hours ago.

src/storage/src/client/upload_object_buffered.rs
coryan: Same answer.

coryan:
PTAL

src/storage/src/client/upload_object_buffered.rs
dbolduc: Nice. I spent way too long trying to figure it out myself, then gave up.

src/storage/src/client/upload_object_buffered.rs
dbolduc: Still confused....

This calculation is telling us the expected number of chunks.... but why is that `target_size`??

https://github.com/googleapis/google-cloud-rust/blob/632b2d4639882478685d3bab156d2c2153c66f4f/src/storage/src/client/upload_object_buffered.rs#L705

My understanding is that `target_size` is supposed to be the number of bytes we want per chunk.

---

https://doc.rust-lang.org/std/primitive.u32.html#method.div_ceil says:

```rs
assert_eq!(7_u32.div_ceil(4), 2);
```

as in: 7/4 rounds up to 2

coryan: Doh!  Okay, I removed the ugly hack.

src/storage/src/client/upload_object_buffered.rs
```diff
@@ -806,32 +809,34 @@ where
     let mut process_buffer = |mut b: bytes::Bytes| match b.len() {
         n if size + n > target_size => {
             let remainder = b.split_off(target_size - size);
+            size += b.len();
```
dbolduc: consider: `size = target_size;`

might be clearer?

src/storage/src/client/upload_object_buffered.rs
```diff
@@ -806,32 +809,34 @@ where
...
             partial.push_back(b);
             Some(Some(remainder))
         }
         n if size + n == target_size => {
+            size += n;
```
dbolduc: same

src/storage/src/client/upload_object_buffered.rs
```diff
@@ -797,7 +800,7 @@ async fn next_chunk<T>(
     payload: &mut InsertPayload<T>,
     remainder: Option<bytes::Bytes>,
     target_size: usize,
-) -> Result<(VecDeque<bytes::Bytes>, Option<bytes::Bytes>)>
+) -> Result<(VecDeque<bytes::Bytes>, usize, Option<bytes::Bytes>)>
```
dbolduc: optional: A comment describing the return value could be helpful when I forget the interface, which should happen by next week.

dbolduc (approved):
Non-blocking, but comments might be helpful for:
- parameters of `partial_upload_request`
- parameters and return type of `next_chunk`

I was able to understand them (eventually) by context, but it could have been easier.

src/storage/src/client/upload_object_buffered.rs
coryan: Fixed by using a struct

src/storage/src/client/upload_object_buffered.rs
coryan: Done.

src/storage/src/client/upload_object_buffered.rs
coryan: Done.

coryan (commented):
PTAL

Issue #2397, opened 2025-06-09
Rust generator support for include list


dbolduc:
We have an exclude list: `skipped-ids`. We would like an include list, maybe: `included-ids`.

We want to use the include list from the `storage/v2` sidekick config that lives in `storage-control`. The set of RPCs/messages that should go into `storage-control` is fixed. It will not grow.

https://github.com/googleapis/google-cloud-rust/blob/499eec5a9985d68d1d8d3186b54fcf04189630cb/src/storage-control/src/generated/gapic/.sidekick.toml#L21

However, the set of media RPCs/messages can grow. Whenever a new message is added to that service, it will get surfaced in our API. We want to prevent these sorts of leaks.

https://github.com/googleapis/google-cloud-rust/blob/499eec5a9985d68d1d8d3186b54fcf04189630cb/src/storage-control/src/lib.rs#L43

PR #2613, opened 2025-07-09
impl(sidekick): find any dependencies

dbolduc:
Part of the work for #2397 

Add support for finding dependencies of given IDs.

Conceptually, it is more complicated than finding the transitive closure of a graph. (I think it is finding the transitive closure of two different graphs). Hopefully the write up makes sense.

I know about `FindServiceDependencies`. In a follow up, I will [refactor it to use this](https://github.com/dbolduc/google-cloud-rust/commit/4b9700508aef98901e86aa711db16ad310bfc97e).

### DIFF
--- a/generator/internal/api/dependencies.go
+++ b/generator/internal/api/dependencies.go
@@ -41,8 +41,6 @@ import (
 // - a method to its request/response messages
 // - an LRO method to its metadata/response type messages
 // - a message to each of its fields' types (if they are a message or enum)
-// - a parent message to each of its nested messages
-// - a parent message to each of its nested enums
 //
 // In the second pass, we compute the reachable set of the nodes found from the
 // first pass. This time, the graph has edges from...
@@ -98,12 +96,6 @@ func FindDependencies(model *API, ids []string) (map[string]bool, error) {
 
 		message, ok := model.State.MessageByID[id]
 		if ok {
-			for _, msg := range message.Messages {
-				add(msg.ID)
-			}
-			for _, enum := range message.Enums {
-				add(enum.ID)
-			}
 			for _, field := range message.Fields {
 				if field.Typez == ENUM_TYPE || field.Typez == MESSAGE_TYPE {
 					add(field.TypezID)
--- a/generator/internal/api/dependencies.go
+++ b/generator/internal/api/dependencies.go
@@ -1,0 +1,164 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package api
+
+import (
+	"fmt"
+)
+
+// Returns the IDs of model elements required by the given set of `ids`
+//
+// We can think of the `model` as a directed graph, with nodes for each of the
+// model elements. Finding required elements is a graph traversal problem.
+//
+// First observe that:
+//   - if we are given a service, we want to include all of its methods.
+//   - if we are given a method, we need to include its service, but not
+//     necessarily any of its sibling methods.
+//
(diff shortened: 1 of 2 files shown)


generator/internal/api/dependencies.go
```diff
@@ -0,0 +1,164 @@
...
+		}
+
+		message, ok := model.State.MessageByID[id]
+		if ok {
+			for _, msg := range message.Messages {
+				add(msg.ID)
+			}
+			for _, enum := range message.Enums {
+				add(enum.ID)
+			}
```
coryan: Do you need this?  I mean, typically  child messages and enums are used for fields of the parent message... but the generator code does *not* need a child message or enum if the parent message does not use it as a field?

generator/internal/api/dependencies_test.go
```diff
@@ -0,0 +1,481 @@
...
+	if err != nil {
+		t.Fatal(err)
+	}
+	want := []string{grandparent, parent, child} // No `sibling`
+	if diff := cmp.Diff(want, flatten(got), cmpopts.SortSlices(less)); diff != "" {
+		t.Errorf("dependencies mismatch (-want, +got):\n%s", diff)
+	}
+}
+
+func TestFindDependenciesMessage(t *testing.T) {
```
coryan, line 178: I think you want a test for these cases:

```proto
message Recursive {
  Recursive field = 1;
}

message A {
  B left = 1;
  B right = 2;
}

message B {
  A value = 1;
}

```

generator/internal/api/dependencies.go
dbolduc: The first pass is about "wants" not "needs".

A service does not need all of its methods... but if we give this thing a service, we want it to include all of the service's methods.

I think the same argument goes for messages and their nested messages/enums.

generator/internal/api/dependencies.go
```diff
@@ -0,0 +1,164 @@
...
+			for _, field := range message.Fields {
+				if field.Typez == ENUM_TYPE || field.Typez == MESSAGE_TYPE {
+					add(field.TypezID)
+				}
+			}
```
dbolduc, line 103: Hmm, maybe this should only be in the second pass.

dbolduc: Considering: 

```proto
message A {
  B b = 1;
}

message B {
  enum C {}
}
```

Given `A`, should we generate `B.C` ?

Ummm... I think so. Which means we want this code.

coryan: I think you do not *need* to generate `B.C`, we are trying to prune to the **minimum** set that satisfies "I can use message `A`".  If we want a larger set we have an easier way to do it.

dbolduc: Thanks, I think I agree.

Did I convince you that `B` should imply `B.C` ?

coryan: I see... a bit subtle ... We can start with that approach, yes.  I suspect we will largely use this methods and services anyway.

dbolduc: > I suspect we will largely use this for methods and services anyway.

Yep.

> a bit subtle

Yeah... I like the formulation of: `the minimum set that satisfies "I can use the inputs"`.

But when I think about what is the minimum set that satisfies I can use a message, it is the message with no fields. The minimum set that satisfies I can use a service is the service with no methods.

So I think my formulation is more like `the minimum set that satisfies "I can use the inputs completely"`. :shrug: 

And, even in that formulation, we do not need `B` to imply `B.C`... So I think I will drop that edge.

> Did I convince you that `B` should imply `B.C` ?

I am the one who has been convinced.

generator/internal/api/dependencies_test.go
dbolduc: Yeah, I missed testing cycles. Thanks.

Added these + a cycle involving enums.

generator/internal/api/dependencies.go
dbolduc: I ended up removing these.

And I agree that in practice it won't matter, because nested messages are almost always used as a field.

Thanks for the review.

generator/internal/api/dependencies.go
```diff
@@ -0,0 +1,156 @@
...
+		method, ok := model.State.MethodByID[id]
+		if ok {
+			add(method.Service.ID)
+			continue
+		}
+
+		message, ok := model.State.MessageByID[id]
+		if ok {
+			if message.Parent != nil {
+				add(message.Parent.ID)
```
coryan, line 134: I had not noticed this before, good thinking.

PR #2618, opened 2025-07-10
refactor(auth)!: remove quota project env var from api_key

harkamaljot:
According to AIP-4110 (https://google.aip.dev/auth/4110), the GOOGLE_CLOUD_QUOTA_PROJECT environment variable is intended to be used exclusively by the Application Default Credentials (ADC) flow. 

Therefore, this change removes support for the GOOGLE_CLOUD_QUOTA_PROJECT environment variable from the api_key_credentials builder. This is a breaking change intended to align our library with official Google Cloud standards and improve behavioral consistency.

### DIFF
--- a/src/auth/src/credentials/api_key_credentials.rs
+++ b/src/auth/src/credentials/api_key_credentials.rs
@@ -126,9 +126,7 @@ impl Builder {
 
     /// Returns a [Credentials] instance with the configured settings.
     pub fn build(self) -> Credentials {
-        let quota_project_id = std::env::var("GOOGLE_CLOUD_QUOTA_PROJECT")
-            .ok()
-            .or(self.quota_project_id.clone());
+        let quota_project_id = self.quota_project_id.clone();
 
         Credentials {
             inner: Arc::new(ApiKeyCredentials {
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/api_key_credentials.rs
```diff
@@ -126,9 +126,7 @@ impl Builder {
 
     /// Returns a [Credentials] instance with the configured settings.
     pub fn build(self) -> Credentials {
-        let quota_project_id = std::env::var("GOOGLE_CLOUD_QUOTA_PROJECT")
-            .ok()
-            .or(self.quota_project_id.clone());
+        let quota_project_id = self.quota_project_id.clone();
```
dbolduc, line 129: nit, predating this PR: can we just say: `quota_project_id: self.quota_project_id,` below?

dbolduc (approved):

src/auth/src/credentials/api_key_credentials.rs
harkamaljot: When we call TokenCache::new(self.build_token_provider()) below we move the ownership of self to build_token_provider method. So I don't think we can do that directly here.

dbolduc: Ack, thanks

PR #2619, opened 2025-07-10
impl(sidekick): introduce `included-ids`

dbolduc:
Part of the work for #2397 

Introduce `included-ids` as a counterpart to `skipped-ids` in sidekick.

I didn't want to think about the behavior if both options are set, so I disallowed it.

A follow up will switch `storage-control` to use `included-ids` instead of `skipped-ids`.

### DIFF
--- a/generator/internal/api/skip.go
+++ b/generator/internal/api/skip.go
@@ -15,6 +15,7 @@
 package api
 
 import (
+	"fmt"
 	"slices"
 	"strings"
 )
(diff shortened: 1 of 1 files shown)


generator/internal/api/skip.go
```diff
@@ -20,29 +20,47 @@ import (
...
 
 func SkipModelElements(model *API, options map[string]string) {
-	ids, ok := options["skipped-ids"]
-	if !ok {
-		return
+	included_ids, included_ok := options["included-ids"]
+	skipped_ids, skipped_ok := options["skipped-ids"]
+	if included_ok && skipped_ok {
+		panic("Both `included-ids` and `skipped-ids` set. Only set one.")
 	}
```
coryan: Seems like we should change the return type from this function to be `error` instead of panicking.

coryan: Or use `slog.Fatal()` which I believe we use elsewhere.  But I would prefer returning an error.

dbolduc: You're right. Done.

Issue #2049, opened 2025-05-08
Verify checksums during downloads


coryan:
We need to verify the CRC32C checksum during downloads, and signal an error if the checksum does not match the reported object checksum.

We can compute the CRC32C checksum on the fly, but sometimes applications already know what to expect. We should allow applications to provide the expected CRC32C checksum and skip all the computation.

We need to think about MD5 checksums too.  These are relatively expensive, and it may be better to only enable them if the application asks.

coryan:
The remaining work can happen in the next milestone, moving.

coryan:
AFAICT #2845 fixed this bug.

PR #2620, opened 2025-07-10
feat(storage): validate crc32c checksum on read

suzmue:
Validate the crc32c checksum when downloading objects. For `all_bytes`, the data is not returned to the user since all of the data is read at once. For streams, the error is not returned until all data has been read.

We skip this check when we:
1. did not request all data
2. did not receive a value for crc32c from the server
3. detect the server decompressed the data from gzip
4. received StatusCode::PartialContent

(See [reader.go](https://github.com/googleapis/google-cloud-go/blob/a44a5a4ee1a05be705bc21a0c1ab81d71e494eec/storage/reader.go#L196) for reference)

This PR also disables automatic decompression from reqwest::Client, which can happen if a user has enabled the relevant features. We may want to re-expose this functionality in the future, with a warning that we cannot check the crc32c matches the expected value. 

This PR does not provide a way to turn off this check, but we can consider adding it. Support for `into_stream()` will be done in a future PR.

For #2049


src/storage/src/client/read_object.rs
```diff
@@ -473,6 +576,62 @@ mod tests {
...
+    async fn read_object_incorrect_crc32c_check() -> Result {
+        // Calculate and serialize the crc32c checksum
+        let u = crc32c::crc32c("goodbye world".as_bytes());
+        let bytes = [
+            (u >> 24 & 0xFF) as u8,
+            (u >> 16 & 0xFF) as u8,
+            (u >> 8 & 0xFF) as u8,
+            (u & 0xFF) as u8,
+        ];
+        let value = base64::prelude::BASE64_STANDARD.encode(bytes);
```
coryan, line 589: I have similar code in the upload side, I think we should refactor it.

src/storage/src/client/v1.rs
```diff
@@ -73,7 +73,7 @@ struct Retention {
 }
 
 // CRC32c checksum is a unsigned 32-bit int encoded using base64 in big-endian byte order.
-struct Crc32c;
+pub(super) struct Crc32c;
```
coryan, line 76: TIL.

src/storage/src/client/read_object.rs
coryan: Probably in a separate PR.

src/storage/src/client/read_object.rs
```diff
@@ -367,16 +394,92 @@ impl ReadObjectResponse {
...
+            self.inner.status(),
+            self.inner.headers(),
+        )
+    }
+}
+
+/// Represents an error that can occur when reading response data.
+#[derive(thiserror::Error, Debug, PartialEq)]
+#[non_exhaustive]
+enum ReadError {
```
coryan, line 444: This is private, should we make the error public eventually? I am not sure what is the right answer yet.

coryan (approved):

src/storage/src/client/read_object.rs
suzmue: Filed https://github.com/googleapis/google-cloud-rust/issues/2630

PR #2622, opened 2025-07-10
impl(storage): introduce retry policy

coryan:
A retry policy following the storage service recommendations.

Motivated by #2043, but useful for other things too.

### DIFF
--- a/src/storage/src/retry_policy.rs
+++ b/src/storage/src/retry_policy.rs
@@ -27,8 +27,6 @@
 //! In addition, resumable uploads return [308 - Resume Incomplete]. This is
 //! not handled by the retry policy.
 //!
-//! The types in this
-//!
 //! [recommends]: https://cloud.google.com/storage/docs/retry-strategy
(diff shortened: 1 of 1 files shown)


src/storage/src/retry_policy.rs
```diff
@@ -0,0 +1,119 @@
...
+//! - [429 - Too Many Requests][429]
+//! - [500 - Internal Server Error][500]
+//! - [502 - Bad Gateway][502]
+//! - [503 - Service Unavailable][503]
+//! - [504 - Gateway Timeout][504]
+//!
+//! In addition, resumable uploads return [308 - Resume Incomplete]. This is
+//! not handled by the retry policy.
+//!
+//! The types in this
```
dbolduc: ...

dbolduc (approved):

src/storage/src/retry_policy.rs
coryan: Removed.

dbolduc (approved):

PR #2624, opened 2025-07-10
docs(storage): samples were reaching production

coryan:
As written the samples were trying to contact production, and failing.

### DIFF
--- a/src/storage/src/client.rs
+++ b/src/storage/src/client.rs
@@ -130,28 +130,26 @@ impl Storage {
     ///
     /// # Example
     /// ```
-    /// # tokio_test::block_on(async {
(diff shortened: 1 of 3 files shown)


coryan:
Maybe `docs` is not the right tag for this commit, but I cannot think of a better one. `test` maybe?

dbolduc:
`cleanup` could work.

`docs` we tend to use for public facing docs and internal (e.g. contributor) docs. That does not help us in releases.

Spitballing, we have gotten used to `impl`, but we could have an `i_` prefix for commits that are not customer facing.

e.g. `feat` vs. `i_feat`, `docs` vs. `i_docs`, `fix` vs. `i_fix`

PR #2627, opened 2025-07-10
cleanup(storage-control)!: cap storage v2 methods, messages

### LINKED ISSUE #2397: Rust generator support for include list

dbolduc:
Fixes #2397 

Include storage/v2 by RPCs.

We also include the messages `WriteObjectSpec`, `ReadObjectRequest` which are not used by the RPCs, but are used by `google-cloud-storage`.

We lose the following messages:
- `CancelResumableWriteRequest`
- `WriteObjectRequest`
- `QueryWriteStatusRequest`
- `StartResumableWriteRequest`
- `ServiceConstants`
- `ChecksummedData`
- `ContentRange`


src/storage-control/src/generated/convert/storage/convert.rs
```diff
@@ -592,29 +592,6 @@ impl gaxi::prost::FromProto<crate::generated::gapic::model::CommonObjectRequestP
     }
 }
 
-impl gaxi::prost::ToProto<service_constants::Values> for crate::generated::gapic::model::service_constants::Values {
```
dbolduc, line 595: https://github.com/googleapis/googleapis/blob/08226f7ca552505dcea983c28d90d13076170b8d/google/storage/v2/storage.proto#L1890

This thing defines service constants in an enum. Yuck.

coryan: Yeah, known thing.

coryan (approved):
I am a bit surprised about the size of the delta, but we can fix things later as needed.

PR #2629, opened 2025-07-10
feat(auth): Allow custom retry policy for mds

harkamaljot:
This PR introduces the ability for users to configure custom retry, backoff, and throttling policies for the Metadata Service (MDS) credentials.


src/auth/src/credentials.rs
```diff
@@ -35,6 +38,18 @@ pub mod user_account;
 pub(crate) const QUOTA_PROJECT_KEY: &str = "x-goog-user-project";
 pub(crate) const DEFAULT_UNIVERSE_DOMAIN: &str = "googleapis.com";
 
+/// A helper type to use [RetryPolicy] in auth builders.
+#[derive(Clone)]
+pub struct RetryPolicyArg(pub(crate) Arc<dyn RetryPolicy>);
```
harkamaljot: Using the same pattern as other place in the repo eg: https://github.com/googleapis/google-cloud-rust/blob/main/src/gax/src/retry_policy.rs#L131

src/auth/src/credentials/mds.rs
```diff
@@ -175,6 +180,77 @@ impl Builder {
         self.scopes = Some(scopes.into_iter().map(|s| s.into()).collect());
         self
     }
+    /// Configure the retry policy for fetching tokens.
+    ///
+    /// The authentication library can automatically retry operations that fail. The
```
sai-sunder-s: `The authentication library can automatically retry operations that fail.` seems to indicate that w have a default retry policy, which we do not.

sai-sunder-s: Same for backoff. You could just remove that sentence.

src/auth/src/credentials/mds.rs
```diff
@@ -175,6 +180,77 @@ impl Builder {
...
+        self.retry_builder = self.retry_builder.with_backoff_policy(v.into().0);
+        self
+    }
+
+    /// Configure the retry throttler.
+    ///
+    /// Advanced applications may want to configure a retry throttler to
+    /// [Address Cascading Failures] and when [Handling Overload] conditions.
+    /// The authentication library throttles its retry loop, using a policy to
+    /// control the throttling algorithm. Use this method to fine tune or
```
sai-sunder-s, line 233: Same here, we do not have a default throttler. Customer has to provide one if they need throttling control.

src/auth/src/retry.rs
```diff
@@ -31,23 +32,18 @@ pub(crate) struct TokenProviderWithRetry<T: TokenProvider> {
...
-#[allow(dead_code)]
 impl<T: TokenProvider> Builder<T> {
-    pub(crate) fn new(inner: T) -> Self {
-        Self {
-            inner: Arc::new(inner),
-            retry_policy: None,
-            backoff_policy: None,
-            retry_throttler: None,
-        }
+    pub(crate) fn with_token_provider(mut self, inner: T) -> Self {
```
sai-sunder-s: token_provider is required. Why not have `fn new(token_provider: T)` to initialize the builder?

sai-sunder-s: Ah I guess because the token provider is only available in the end. You need this builder to set the policies from cred builder.

src/auth/src/retry.rs
```diff
@@ -65,19 +61,23 @@ impl<T: TokenProvider> Builder<T> {
...
             .backoff_policy
             .unwrap_or_else(|| Arc::new(ExponentialBackoff::default()));
         let retry_throttler = self
             .retry_throttler
             .unwrap_or_else(|| Arc::new(Mutex::new(AdaptiveThrottler::default())));
-        TokenProviderWithRetry {
-            inner: self.inner,
+        let inner = self
+            .inner
+            .ok_or(BuilderError::missing_field("token_provider"))?;
```
sai-sunder-s: missing_field indicates that some field in the credential json is missing. This will be misleading to the user. They will think their json input needs `token_provider` field.

sai-sunder-s (changes requested):

src/auth/src/retry.rs
harkamaljot: yes that is correct, another solution could be to duplicate the fields, however given it's an internal implementation so I felt this would be cleaner. But we can do the other way as well.

src/auth/src/retry.rs
harkamaljot: this error should not happen, do you think adding unreachable here makes more sense? or should we just store the retry policies directly and avoid throwing any errors here.

src/auth/src/credentials/mds.rs
harkamaljot: we have default throttler here: https://github.com/googleapis/google-cloud-rust/blob/main/src/auth/src/retry.rs#L74 however it won't do anything if you do not have retry policy.

src/auth/src/credentials/mds.rs
harkamaljot: Thanks removed.

src/auth/src/credentials.rs
```diff
@@ -689,14 +689,76 @@ pub mod testing {
...
+            _loop_start: std::time::Instant,
+            attempt_count: u32,
+            _idempotent: bool,
+            error: gax::error::Error,
+        ) -> RetryResult {
+            if attempt_count >= self.max_attempts {
+                return RetryResult::Exhausted(error);
+            }
+
+            if error.is_authentication() {
```
sai-sunder-s: why cant we retry for non-auth errors? Like 503?

src/auth/src/retry.rs
sai-sunder-s: How about we make token_provider a required param to `build` method? That way we will always have it and no need of the `with_` method.

src/auth/src/retry.rs
```diff
@@ -31,53 +32,54 @@ pub(crate) struct TokenProviderWithRetry<T: TokenProvider> {
...
 }
 
-#[derive(Debug)]
+#[derive(Debug, Default)]
 pub(crate) struct Builder<T: TokenProvider> {
-    inner: Arc<T>,
-    retry_policy: Option<Arc<dyn RetryPolicy>>,
-    backoff_policy: Option<Arc<dyn BackoffPolicy>>,
-    retry_throttler: Option<SharedRetryThrottler>,
+    inner: Option<Arc<T>>,
```
sai-sunder-s: why is inner token provider optional?

src/auth/src/retry.rs
sai-sunder-s: That would also skip need for returning Result from build and avoid a bunch of `?` you have added.

src/auth/src/retry.rs
sai-sunder-s: See other comment. We can avoid this altogether by passing in tp to `build()`

src/gax/src/backoff_policy.rs
```diff
@@ -85,6 +85,13 @@ impl From<Arc<dyn BackoffPolicy>> for BackoffPolicyArg {
     }
 }
 
+impl BackoffPolicyArg {
```
sai-sunder-s: @coryan could you please review this change, since it touches gax.

src/auth/src/credentials.rs
harkamaljot: we can however in this test policy it doesn't make sense to add it here, as Token method always returns back a CredentialsError, so that code will never execute.

src/gax/src/backoff_policy.rs
coryan: I understand the need for this change. We should start by removing the `pub(crate)` and fixing `gax` to use the new accessor (or new `From` if my next suggestion works).

src/gax/src/backoff_policy.rs
```diff
@@ -85,6 +85,13 @@ impl From<Arc<dyn BackoffPolicy>> for BackoffPolicyArg {
...
+    /// Consumes the argument and returns the wrapped policy object.
+    pub fn into_inner(self) -> Arc<dyn BackoffPolicy> {
+        self.0
+    }
+}
```
coryan: I believe (please check) we can replace these with:

```suggestion
impl From<BackoffPolicyArg> for Arc<dyn BackoffPolicy> {
    fn from(self) -> Arc<dyn BackoffPolicy> {
        self.0
    }
}
```

src/auth/src/credentials/mds.rs
```diff
@@ -359,6 +434,51 @@ mod tests {
...
+            "backoff_policy: ExponentialBackoff", 
+            "retry_throttler: Mutex { data: AdaptiveThrottler", 
+            "factor: 2.0"
+        ];
+        "with_default_values"
+    )]
+    fn test_mds_retry_policies(use_custom_config: bool, expected_substrings: &[&str]) {
+        let mut builder = Builder::default();
+
+        if use_custom_config {
```
coryan: Better use two tests for this rather than a true/false argument.  Or receive the builder as an argument.

src/auth/src/credentials/mds.rs
```diff
@@ -359,6 +434,51 @@ mod tests {
...
+        }
+
+        let provider = builder.build_token_provider();
+        let debug_str = format!("{provider:?}");
+
+        for sub in expected_substrings {
+            assert!(
+                debug_str.contains(sub),
+                "Expected to find '{sub}' in '{debug_str:?}'"
+            );
```
coryan: A better test would check the behavior of the builder: does it create credentials with the right behavior.

src/gax/src/retry_policy.rs
```diff
@@ -145,6 +145,13 @@ impl std::convert::From<Arc<dyn RetryPolicy>> for RetryPolicyArg {
     }
 }
 
+impl RetryPolicyArg {
```
coryan: Same comments as in backoff_policy.rs

src/gax/src/retry_throttler.rs
```diff
@@ -129,6 +129,13 @@ impl From<SharedRetryThrottler> for RetryThrottlerArg {
     }
 }
 
+impl RetryThrottlerArg {
```
coryan: Ditto.

src/auth/src/credentials.rs
```diff
@@ -689,14 +689,76 @@ pub mod testing {
...
+        fn on_error(
+            &self,
+            _loop_start: std::time::Instant,
+            attempt_count: u32,
+            _idempotent: bool,
+            error: gax::error::Error,
+        ) -> RetryResult {
+            if attempt_count >= self.max_attempts {
+                return RetryResult::Exhausted(error);
+            }
```
coryan: You should use composition for this.  Define a base policy (without attempt limits) and then use `RetryPolicyExt` to layer the retry limit.

src/auth/src/credentials.rs
```diff
@@ -689,14 +689,76 @@ pub mod testing {
...
+    impl BackoffPolicy for TestBackoffPolicy {
+        fn on_failure(
+            &self,
+            _loop_start: std::time::Instant,
+            _attempt_count: u32,
+        ) -> std::time::Duration {
+            self.was_called.store(true, Ordering::SeqCst);
+            std::time::Duration::from_millis(1)
+        }
+    }
```
coryan: This looks like something `mockall` could do.

src/auth/src/credentials.rs
```diff
@@ -689,14 +689,76 @@ pub mod testing {
...
     use std::sync::LazyLock;
+    use std::sync::atomic::{AtomicBool, Ordering};
     use test_case::test_case;
 
+    #[derive(Debug)]
+    pub struct AuthRetryPolicy {
+        pub max_attempts: u32,
+    }
+
+    impl RetryPolicy for AuthRetryPolicy {
```
coryan: Could you do this with a mock policy?

src/auth/src/credentials/mds.rs
harkamaljot: I think I still need a test to check if values are correctly set in the builder, however you are correct a tests are missing to verify the behavior, so I added those as well.

src/auth/src/credentials/mds.rs
```diff
@@ -175,6 +180,75 @@ impl Builder {
         self.scopes = Some(scopes.into_iter().map(|s| s.into()).collect());
         self
     }
+    /// Configure the retry policy for fetching tokens.
```
coryan, line 184: nit: missing blank line.

src/auth/src/credentials/mds.rs
```diff
@@ -354,11 +431,158 @@ mod tests {
...
+                            .body("".to_string())
+                            .unwrap()
+                    } else {
+                        // Succeed on the 3rd attempt
+                        http::Response::builder()
+                            .status(200)
+                            .header("Content-Type", "application/json")
+                            .body(serde_json::to_string(&response_clone).unwrap())
+                            .unwrap()
+                    }
```
coryan: You could write this as in:

https://github.com/googleapis/google-cloud-rust/pull/2645/files#diff-2b4aeb5ea86ee66b88cf7fb29d4047f6f039e57149f184e6bc3a791348cbc54aR997-R1001

🤷

src/auth/src/credentials/mds.rs
coryan: Hmmm... since you know the mocks are called, you know the values are correctly set by the builder.  No?

src/gax/src/backoff_policy.rs
coryan: I see you added the new `From` but you did not 

- Remove the `pub(crate)` 
- fix gax to use the new `From` accessor / convert function.


When I said "start by" I was hinting that an initial PR just affecting `gax` would be a good idea.

harkamaljot: sorry I missed that. I will send a PR after this one to remove the `pub(crate)` from gax.

sai-sunder-s (approved):

src/gax/src/backoff_policy.rs
coryan: SGTM.

PR #2631, opened 2025-07-11
chore: improve deserde of errors from Status

### LINKED ISSUE #2440: Improve deserialization of errors from `rpc::model::Status`
We have a function to convert a `rpc::model::Status` to a `gax::error::Status`:

https://github.com/googleapis/google-cloud-rust/blob/e5f0162e04baa3ffb6059b3310604076b4efe75c/src/gax/src/error/rpc.rs#L378-L386

Sometimes we only have a reference of the `rpc::model::Status` and cloning that reference before converting is overkill, as we are going to copy the data anyway.

varun-doshi:
Fixes #2440 

Implements `impl From<&rpc::model::Status> for Status` to avoid cloning before converting

### DIFF
--- a/src/gax/src/error/rpc.rs
+++ b/src/gax/src/error/rpc.rs
@@ -375,12 +375,12 @@ impl TryFrom<&bytes::Bytes> for Status {
     }
 }
 
-impl From<rpc::model::Status> for Status {
-    fn from(value: rpc::model::Status) -> Self {
+impl From<&rpc::model::Status> for Status {
+    fn from(value: &rpc::model::Status) -> Self {
         Self {
             code: value.code.into(),
-            message: value.message,
-            details: value.details.into_iter().map(StatusDetails::from).collect(),
+            message: value.message.clone(),
+            details: value.details.iter().map(|d| StatusDetails::from(d)).collect(),
         }
     }
 }
@@ -418,8 +418,8 @@ pub enum StatusDetails {
     Other(wkt::Any),
 }
 
-impl From<wkt::Any> for StatusDetails {
-    fn from(value: wkt::Any) -> Self {
+impl From<&wkt::Any> for StatusDetails {
+    fn from(value: &wkt::Any) -> Self {
         macro_rules! try_convert {
             ($($variant:ident),*) => {
                 $(
@@ -443,7 +443,7 @@ impl From<wkt::Any> for StatusDetails {
             RetryInfo
         );
 
-        StatusDetails::Other(value)
+        StatusDetails::Other(value.clone())
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/gax/src/error/rpc.rs
```diff
@@ -375,12 +375,12 @@ impl TryFrom<&bytes::Bytes> for Status {
     }
 }
 
-impl From<rpc::model::Status> for Status {
-    fn from(value: rpc::model::Status) -> Self {
+impl From<&rpc::model::Status> for Status {
```
coryan, line 388: Thanks for the PR!

We should not remove the existing `From<>`. This is now part of the public API, and we are trying to avoid breaking changes (we may have to make them, but we want to avoid them).

coryan (changes requested):

src/gax/src/error/rpc.rs
varun-doshi: Understood
Will add them back

src/gax/src/error/rpc.rs
```diff
@@ -418,6 +428,7 @@ pub enum StatusDetails {
     Other(wkt::Any),
 }
 
+
```
coryan: nit: remove extra line.

src/integration-tests/src/workflows.rs
```diff
@@ -253,7 +253,7 @@ pub async fn manual(
         match result {
             LR::Error(status) => {
                 println!("LRO completed with error {status:?}");
-                let status = gax::error::rpc::Status::from(*status);
+                let status = gax::error::rpc::Status::from(&*status);
```
coryan: I would probably leave these alone, moving the value *may* save some computation (not that it matters in tests).

src/gax/src/error/rpc.rs
```diff
@@ -769,7 +809,7 @@ mod tests {
             .set_message("try-again")
             .set_details(vec![wkt::Any::from_msg(&detail).unwrap()]);
 
-        let status = Status::from(input);
+        let status = Status::from(&input);
```
coryan: Maybe test both versions?

```suggestion
        let from_ref = Status::from(&input);
        let status = Status::Status::from(input);
        assert_eq!(from_ref, status);
```

src/gax/src/error/rpc.rs
```diff
@@ -784,7 +824,7 @@ mod tests {
             .set_code(Code::Unavailable as i32)
             .set_message("try-again")
             .set_details(vec![any.clone()]);
-        let got = Status::from(input);
+        let got = Status::from(&input);
```
coryan: Same

coryan:
Looking good, just need to fix the CI problems.

coryan (approved):

dbolduc (approved):

coryan:
Thanks for the PR!

Issue #2057, opened 2025-05-08
Automatically resume interrupted resumable uploads


coryan:
That is the point of using resumable uploads.

The client should resume the upload if it is interrupted. The client must query the status of the upload, and start from the last byte received.

Note that when the client is reading from a source that cannot be rewound, then the client must buffer enough data to resume the upload. For HTTP, this implies some limited buffer, and PUTs, and stalling until the previous PUT returns successfully.  For BidiStreamingWrite, this can be better, with periodic commits and periodic flushing of the buffer.

Issue #2056, opened 2025-05-08
Automatically retry failed single-shot uploads


coryan:
Subject to the conditional idempotency rules, we should retry any single-shot upload that fails, with all the right policies, of course.

Issue #2045, opened 2025-05-08
Support progress-based timeouts for downloads


coryan:
It is hard to define an end-to-end timeout for downloads as (1) the size varies from 0 bytes to 5TiB, (2) part of the delay is caused by the application pulling data fast enough.

We need to define some kind of "progress based" (or rather "lack of progress based") timeouts.  The request should timeout if we were not able to receive at least X bytes per second.  We should use a different timeout value for "the operation should start downloading within X milliseconds".

Issue #2044, opened 2025-05-08
Support progress-based timeouts for uploads


coryan:
It is hard to define an end-to-end timeout for uploads as (1) the size varies from 0 bytes to 5TiB, (2) part of the delay is caused by the application providing data fast enough.

We need to define some kind of "progress based" (or rather "lack of progress based") timeouts.  The request should timeout if we were not able to send at least X bytes per second.  We should use a different timeout value for "the operation should start uploading within X milliseconds".

PR #2632, opened 2025-07-11
impl(storage): capture retry policies

coryan:
Capture the retry policies into the `StorageInner` object. Provide
defaults for the backoff and retry policy.

Part of the work for #2056, #2057, and I suspect other bugs too.


src/storage/src/client.rs
```diff
@@ -219,17 +230,39 @@ impl Storage {
...
+        config.endpoint = Some(endpoint);
+        let inner = Arc::new(StorageInner::new(client, config));
+        Ok(Self { inner })
     }
 }
 
 impl StorageInner {
+    /// Builds a client assuming `config.cred` and `config.endpoint` are initialized.
+    ///
+    /// panics if
```
dbolduc: "panics otherwise", or just remove

src/storage/src/retry_policy.rs
```diff
@@ -36,7 +36,19 @@
...
+use gax::{
+    retry_policy::{RetryPolicy, RetryPolicyExt},
+    retry_result::RetryResult,
+};
+use std::time::Duration;
+
+/// The default retry policy for the Storage client.
+///
+/// The client will retry all the errors shown as retryable in the service
+/// documentation, and stop retrying after 10 seconds.
```
dbolduc, line 48: comment: I am curious how this will work for big downloads. I can wait and see.

src/storage/src/client.rs
```diff
@@ -219,17 +230,39 @@ impl Storage {
...
     }
 }
 
 impl StorageInner {
+    /// Builds a client assuming `config.cred` and `config.endpoint` are initialized.
+    ///
+    /// panics if  
+    pub(self) fn new(client: reqwest::Client, config: gaxi::options::ClientConfig) -> Self {
+        let retry_policy = config
+            .retry_policy
```
dbolduc, line 244: comment: I guess we can't test these until we use them. It would be nice to verify there is a default policy in the future. (e.g. a mock server returning a 503 expects more calls)

dbolduc (dismissed):

src/storage/src/client.rs
coryan: Fixed, thanks

src/storage/src/client.rs
coryan: Yes.

src/storage/src/retry_policy.rs
coryan: They won't.  We need these to create the resumable upload, and (I think) for small single-shot uploads. We will need completely new policies to deal with "the upload is not making enough progress, need to restart". See #2044 and #2045

dbolduc (approved):

PR #2636, opened 2025-07-11
impl(storage): use JSON multipart uploads

### LINKED ISSUE #2635: Include object metadata in single-shot uploads
When uploading objects via single-shot uploads we need to include the object metadata attributes too.

coryan:
We need to send the object metadata, that requires using JSON multipart
uploads.

Fixes #2635

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -321,7 +321,7 @@ flate2         = { default-features = false, version = "1" }
 httptest       = { default-features = false, version = "0.16.3" }
 md5            = { default-features = false, version = "0.8" }
 mockall        = { default-features = false, version = "0.13" }
-multer         = { default-features = false, version = "3.1.0" }
+multer         = { default-features = false, version = "3" }
 mutants        = { default-features = false, version = "0.0.3" }
 num-bigint-dig = { default-features = false, version = "0.8" }
 paste          = { default-features = false, version = "1" }
--- a/src/storage/src/client/upload_object/unbuffered.rs
+++ b/src/storage/src/client/upload_object/unbuffered.rs
@@ -33,7 +33,7 @@ where
     /// # Ok(()) }
     /// ```
     pub async fn send_unbuffered(self) -> crate::Result<Object> {
-        // TODO(#2634) - use to resumable uploads for large payloads.
+        // TODO(#2634) - use resumable uploads for large payloads.
         let builder = self.single_shot_builder().await?;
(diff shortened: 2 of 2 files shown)


Cargo.toml
```diff
@@ -321,7 +321,9 @@ flate2         = { default-features = false, version = "1" }
 httptest       = { default-features = false, version = "0.16.3" }
 md5            = { default-features = false, version = "0.8" }
 mockall        = { default-features = false, version = "0.13" }
+multer         = { default-features = false, version = "3.1.0" }
```
dbolduc: Can we use `version = "3"`?

src/storage/src/client/upload_object/unbuffered.rs
```diff
@@ -33,7 +33,8 @@ where
     /// # Ok(()) }
     /// ```
     pub async fn send_unbuffered(self) -> crate::Result<Object> {
-        let builder = self.http_request_builder().await?;
+        // TODO(#2634) - use to resumable uploads for large payloads.
```
dbolduc: nit: s/use to/use/

src/storage/src/client/upload_object/unbuffered.rs
```diff
@@ -77,8 +77,18 @@ where
...
+        let metadata = reqwest::multipart::Part::text(v1::insert_body(&self.resource).to_string())
+            .mime_str("application/json; charset=UTF-8")
+            .map_err(Error::ser)?;
+        let media = reqwest::multipart::Part::stream(reqwest::Body::wrap_stream(stream));
+        let form = reqwest::multipart::Form::new()
+            .part("metadata", metadata)
+            .part("media", media);
+        let builder = builder.header(
+            "content-type",
+            format!("multipart/related; boundary={}", form.boundary()),
```
dbolduc, line 89: comment: I was curious about how they generate a MIME string. They do 256 random bytes.

https://github.com/seanmonstar/reqwest/blob/036f9c7cbed83eb6fec8fb60095195d264b4374c/src/async_impl/multipart.rs#L591-L599

Cargo.toml
coryan: Forgot about that, fixed.

src/storage/src/client/upload_object/unbuffered.rs
coryan: Thanks, fixed.

coryan (commented):
PTAL

PR #2642, opened 2025-07-11
feat(storage): validate streaming response data matches server crc32c

suzmue:
We also want to validate the response when it is converted into a futures::Stream. This PR wraps the bytes_stream to track the crc32c of the bytes that are read, and then returns an error at the end of the stream if the value doesn't match the expected state.

For #2049


src/storage/src/client/read_object.rs
```diff
@@ -438,6 +444,60 @@ impl ReadObjectResponse {
...
+impl<S> Stream for Crc32CStream<S>
+where
+    S: Stream<Item = Result<bytes::Bytes>>,
+{
+    type Item = Result<bytes::Bytes>;
+
+    fn poll_next(
+        self: Pin<&mut Self>,
+        cx: &mut futures::task::Context<'_>,
+    ) -> futures::task::Poll<Option<Self::Item>> {
```
dbolduc: nit: `use futures::task::Poll` could help with readability

src/storage/src/client/read_object.rs
```diff
@@ -438,6 +444,60 @@ impl ReadObjectResponse {
...
+        // Poll the inner stream.
+        let this = self.project();
+        let inner_stream_pin = this.inner_stream;
+        match inner_stream_pin.poll_next(cx) {
+            futures::task::Poll::Ready(Some(Ok(chunk))) => {
+                *this.crc32c = crc32c::crc32c_append(*this.crc32c, &chunk);
```
coryan: This computation is wasted if `this.response.crc32c.is_none()` (same thing in the `next()` function, I missed it there).  Maybe a separate bug and PR to skip it?

src/storage/src/client/read_object.rs
```diff
@@ -413,11 +416,14 @@ impl ReadObjectResponse {
...
     #[cfg_attr(docsrs, doc(cfg(feature = "unstable-stream")))]
-    /// Convert the response to a [futures::Stream].
-    pub fn into_stream(self) -> impl futures::Stream<Item = Result<bytes::Bytes>> {
-        // TODO(#2049): implement checksum checks for streams.
+    /// Convert the response to a [Stream].
+    pub fn into_stream(self) -> impl Stream<Item = Result<bytes::Bytes>> {
         use futures::TryStreamExt;
-        self.inner.bytes_stream().map_err(Error::io)
+        let response_crc32c = self.check_crc32c();
+        Crc32CStream::new(
```
coryan: I am not sure if this idea is any better:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a520b10536b79869c6b70e7737a2923b

src/storage/src/client/read_object.rs
```diff
@@ -413,11 +416,14 @@ impl ReadObjectResponse {
...
+            self.inner.bytes_stream().map_err(Error::io),
+            response_crc32c,
+        )
```
coryan: Or maybe something based on this:

```suggestion
        use futures::StreamExt;
        let stream = self.inner.bytes_stream().map_err(Error::io);
        futures::stream::unfold(Some((stream, 0_u32)), async move |s| {
            if let Some((mut stream, mut crc32c)) = s {
                return match stream.next().await {
                    None =>  if crc32c == 42 { None } else { Some((Err(Error::deser("bad checksum")), None)) }
                    Some(Err(e)) => Some((Err(e), None)),
                    Some(Ok(b)) => { crc32c += 1; Some((Ok(b), Some((stream, crc32c)))) },
                };
            }
            None
        })
```

src/storage/src/client/read_object.rs
dbolduc: +1

and in `all_bytes`: https://github.com/googleapis/google-cloud-rust/blob/cd2830e995b68579ccf14bbf1cfcf945e60e79ec/src/storage/src/storage/read_object.rs#L373

Also question: can an application call `read_object.next()`, then call `read_object.all_bytes()`? The crc32c logic in all_bytes is assuming no. I think we want that to also use `crc32c_append`.

dbolduc (commented):
FYI: I moved some of the files around while this PR was in flight. You might have to do some ugly merging/rebasing. Sorry for the inconvenience :grimacing:

src/storage/src/client/read_object.rs
suzmue: Since I'm rebasing anyway I'll make the changes in this PR.

Yeah good point they can consume the stream into an iter or do all_bytes after next since that just has a reference. Both need to be updated then.

coryan (approved):

src/storage/src/client/read_object.rs
suzmue: I'm looking at your two suggestions, but would likely swap to these in a follow up PR.

src/storage/src/client/read_object.rs
suzmue: Added tests that cover these test cases and skipped calculations.

src/storage/src/storage/read_object.rs
```diff
@@ -550,6 +606,81 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn read_object_next_then_consume_response() -> Result {
```
dbolduc, line 610: Nice

src/storage/src/storage/read_object.rs
```diff
@@ -414,31 +425,68 @@ impl ReadObjectResponse {
 
     #[cfg(feature = "unstable-stream")]
     #[cfg_attr(docsrs, doc(cfg(feature = "unstable-stream")))]
-    /// Convert the response to a [futures::Stream].
-    pub fn into_stream(self) -> impl futures::Stream<Item = Result<bytes::Bytes>> {
-        // TODO(#2049): implement checksum checks for streams.
+    /// Convert the response to a [Stream].
+    pub fn into_stream(self) -> impl Stream<Item = Result<bytes::Bytes>> {
```
dbolduc, line 429: non-blocking comment / food for thought: we should be able to either

A: write this in terms of `next()`

https://github.com/googleapis/google-cloud-rust/blob/7c71010b19efd23b72fdad0ed087c140413ea5e9/src/lro/src/internal.rs#L326-L341

or B: hold a stream, and write `next()` in terms of the stream:

https://github.com/googleapis/google-cloud-rust/blob/7c71010b19efd23b72fdad0ed087c140413ea5e9/src/gax/src/paginator.rs#L217-L235

I am not sure how B will work with `all_bytes` though. We'd probably have to extend the bytes manually.

dbolduc (approved):

Issue #2403, opened 2025-06-09
Consider moving `StorageControl` client into the `google-cloud-storage` crate


dbolduc:
It might make more sense for the storage control client to live alongside the media client in the `google-cloud-storage` crate.

Observations:
- There are many types like `Object` which we want in the interface for both clients.
- It is probably easier to split crates than it is to consolidate crates.
- The precedence is to have one crate per `BUILD` in googleapis. This would go counter to that.
- Leaving the crates separate means we might be able to have a fully generated `storage-control` crate.
  - Note that this is nowhere near as important as having a fully generated client.
- The media client will have handwritten samples. We know we want at least one for the admin client.
  - The samples for both can live in `google-cloud-storage`. That is not sooo confusing.

dbolduc is leaning towards keeping the crates separate.

dbolduc:
If we keep the crates separate, we should probably rename this to `google-cloud-storage-control-v2`.

dbolduc:
We decided that we do want to move the `StorageControl` client into `google-cloud-storage`. Googlers can see go/cloud-rust:one-gcs-crate for more details.

My rough plan is to:
- re-export all public `storage-control` types from `storage` and remove any downstream uses of `storage-control`
- move the internals all at once, deleting `google-cloud-storage-control`
- clean up any resulting messes

PR #2643, opened 2025-07-11
refactor(storage): export storage-control models

dbolduc:
Part of the work for #2403 

Move the models from storage-control into storage.

### DIFF
--- a/src/storage/src/lib.rs
+++ b/src/storage/src/lib.rs
@@ -34,3 +34,8 @@ pub mod backoff_policy;
 pub mod client;
 pub mod retry_policy;
 pub mod upload_source;
+
+// TODO(#2403) - This includes implementation details like `ReadObjectRequest`.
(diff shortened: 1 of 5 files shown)


src/storage/src/lib.rs
```diff
@@ -34,3 +34,8 @@ pub mod backoff_policy;
 pub mod client;
 pub mod retry_policy;
 pub mod upload_source;
+
+// TODO(#2403) - This includes implementation details like `ReadObjectRequest`.
+// We do not want to expose those in the long run.
```
coryan, line 39: I assume you have a plan to fix this before GA?

dbolduc: Yeah, I would not leak this in the release. I just think this will be a more efficient implementation path while we are still `0.0`.

We need to get the models generated in `google-cloud-storage` crate before we can decide which ones to expose and which ones to hide.

PR #2644, opened 2025-07-11
fix(auth): Implement framework for Auth Request metric headers and use in impersonated creds

sai-sunder-s:
Metrics headers from auth and client libraries are sent through `x-goog-api-client` header. These headers are used for auth requests to get token as well as while using the auth tokens and making an API call.

This PR focuses on the headers for the auth requests only.

There will be future PRs for auth request headers for other cred types and for token usage through grpc and http requests.

### DIFF
--- a/src/auth/src/headers_util.rs
+++ b/src/auth/src/headers_util.rs
@@ -20,6 +20,34 @@ use crate::token::Token;
 use http::HeaderMap;
 use http::header::{AUTHORIZATION, HeaderName, HeaderValue};
 
+mod build_info {
+    // The file has been placed there by the build script.
+    include!(concat!(env!("OUT_DIR"), "/build_env.rs"));
+}
+
+/// The name of the telemetry header.
+pub(crate) const X_GOOG_API_CLIENT: &str = "x-goog-api-client";
+
+/// Access token request type.
+pub(crate) const ACCESS_TOKEN_REQUEST_TYPE: &str = "at";
+
+/// Format the struct as needed for the `x-goog-api-client` header.
+pub(crate) fn metrics_header_value(request_type: &str, cred_type: &str) -> String {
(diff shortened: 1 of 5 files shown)


src/auth/src/headers_util.rs
coryan, line 1: Naming things `*util` or `*utils` is a recipe for attracting unrelated things into giant convoluted modules.  Consider renaming this to `metric_headers` or `telemetry_headers`.

src/auth/src/headers_util.rs
```diff
@@ -20,6 +20,34 @@ use crate::token::Token;
...
+pub(crate) const ACCESS_TOKEN_REQUEST_TYPE: &str = "at";
+
+/// Format the struct as needed for the `x-goog-api-client` header.
+pub(crate) fn metrics_header_value(request_type: &str, cred_type: &str) -> String {
+    // Strip out the initial "rustc " string from `RUSTC_VERSION`. If not
+    // found, leave RUSTC_VERSION unchanged.
+    let rustc_version = build_info::RUSTC_VERSION;
+    let rustc_version = rustc_version
+        .strip_prefix("rustc ")
+        .unwrap_or(build_info::RUSTC_VERSION);
```
coryan: Computing this on every request seems wasteful, you could move the computation to the `build.rs` script. Though I guess in the auth library it happens rarely ?

Your call.

coryan (approved):

src/auth/src/headers_util.rs
sai-sunder-s: There are other headers related code in this file. We could rename from headers_util to headers.rs or if more metrics related code comes, we can create a module called headers and sub mods for metrics and auth headers.

coryan: Ack.  This can wait, but please considering splitting the module before it becomes unmanageable.

coryan (approved):

PR #2646, opened 2025-07-11
refactor(storage): export storage-control client,builder,stub

dbolduc:
Part of the work for #2403 

Move the `StorageControl` client, builders, stub into `google-cloud-storage`.

Move all uses of `storage_control::client::StorageControl` outside the `google-cloud-storage-control` crate to using `storage::client::StorageControl`.

While currently the `iam_v1`, `longrunning` deps are only used in tests, they will be needed in the library soon enough. So add them as regular-old dependencies.


src/integration-tests/tests/auto_populated.rs
```diff
@@ -22,9 +22,9 @@ mod tests {
     mockall::mock! {
         #[derive(Debug)]
         StorageControl {}
-        impl storage_control::stub::StorageControl for StorageControl {
-            async fn create_folder(&self, req: storage_control::model::CreateFolderRequest, _options: gax::options::RequestOptions) -> gax::Result<gax::response::Response<storage_control::model::Folder>>;
-            async fn list_anywhere_caches(&self, req: storage_control::model::ListAnywhereCachesRequest, _options: gax::options::RequestOptions) -> gax::Result<gax::response::Response<storage_control::model::ListAnywhereCachesResponse>>;
+        impl storage::stub::StorageControl for StorageControl {
```
coryan, line 25: This code probably would be more readable with some `use` declarations.

coryan (approved):

src/integration-tests/tests/auto_populated.rs
dbolduc: Probably, but it is below my line of things that I care about. :grimacing:

PR #2647, opened 2025-07-11
fix(auth)!: Remove option for loading any credential type from json

sai-sunder-s:
Removing the option for loading arbitrary credential json in order to prevent misuse.

Users can still use ADC and can still load json through credential specific builders.

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -254,19 +254,13 @@ pub(crate) mod dynamic {
     }
 }
 
-#[derive(Debug, Clone)]
-enum CredentialsSource {
-    CredentialsJson(Value),
-    DefaultCredentials,
-}
-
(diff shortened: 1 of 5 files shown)


src/auth/src/credentials.rs
```diff
@@ -254,19 +254,13 @@ pub(crate) mod dynamic {
...
-#[derive(Debug, Clone)]
-enum CredentialsSource {
-    CredentialsJson(Value),
-    DefaultCredentials,
-}
-
 /// A builder for constructing [`Credentials`] instances.
 ///
-/// By default (using [`Builder::default`]), the builder is configured to load
+/// This builder is configured to load
```
coryan: Maybe just 

```
/// This builder loads credentials according to ...
```

and reflow the text.

coryan (dismissed):

PR #2650, opened 2025-07-12
refactor(storage)!: move storage-control impl

### LINKED ISSUE #2403: Consider moving `StorageControl` client into the `google-cloud-storage` crate

dbolduc:
Most of the work for #2403 

Moves the storage control implementation into `google-cloud-storage`, and removes the `google-cloud-storage-control` crate.


src/storage/Cargo.toml
```diff
@@ -27,36 +27,42 @@ categories.workspace   = true
...
+lro.workspace         = true
+rpc.workspace         = true
+wkt.workspace         = true
 
 [features]
 unstable-stream = ["reqwest/stream"]
 
 [dev-dependencies]
 anyhow.workspace         = true
-crc32c                   = { workspace = true }
```
coryan, line 59: 👍

coryan (approved):

PR #2653, opened 2025-07-13
impl(storage): retry single shot uploads

coryan:
Part of the work for #2056

### DIFF
--- a/src/storage/src/storage/upload_object/unbuffered.rs
+++ b/src/storage/src/storage/upload_object/unbuffered.rs
@@ -12,12 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::upload_source::Seek;
 use super::*;
 
 impl<T> UploadObject<T>
 where
-    T: StreamingSource + Send + Sync + 'static,
-    T::Error: std::error::Error + Send + Sync + 'static,
+    T: StreamingSource + Seek + Send + Sync + 'static,
+    <T as StreamingSource>::Error: std::error::Error + Send + Sync + 'static,
+    <T as Seek>::Error: std::error::Error + Send + Sync + 'static,
 {
     /// A simple upload from a buffer.
     ///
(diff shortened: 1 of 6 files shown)


src/storage/src/storage/upload_object/unbuffered.rs
```diff
@@ -32,7 +34,24 @@ where
...
     /// # Ok(()) }
     /// ```
-    pub async fn send_unbuffered(self) -> crate::Result<Object> {
+    pub async fn send_unbuffered(self) -> Result<Object> {
+        // TODO(#2056) - make idempotency configurable.
+        // Single shot uploads are idempotent only if they have pre-conditions.
+        let idempotent =
+            self.spec.if_generation_match.is_some() || self.spec.if_metageneration_match.is_some();
+        gax::retry_loop_internal::retry_loop(
+            // TODO(#2044) - we need to apply any timeouts here.
```
dbolduc, line 43: Q: doesn't the retry loop handle this?

src/storage/src/storage/upload_object/unbuffered.rs
```diff
@@ -328,4 +359,141 @@ mod tests {
...
+    async fn single_shot_retry_transient_failures_then_success() -> Result {
+        let server = Server::run();
+        let matching = || {
+            Expectation::matching(all_of![
+                request::method_path("POST", "/upload/storage/v1/b/test-bucket/o"),
+                request::query(url_decoded(contains(("name", "test-object")))),
+                request::query(url_decoded(contains(("uploadType", "multipart")))),
+            ])
+        };
+        server.expect(matching().times(3).respond_with(cycle![
```
dbolduc, line 402: TIL: `std::iter::Cycle`

dbolduc: This doesn't actually cycle though, right?

dbolduc (approved):
No changes, just questions.

src/storage/src/storage/upload_object/unbuffered.rs
coryan: The loop provides the timeout value, but the value needs to be used.

src/storage/src/storage/upload_object/unbuffered.rs
coryan: No, but that is the only way I found to specify varying responses in `httptest`.  I think we could use a mutable lambda with a counter, but this seems more readable, maybe?

src/storage/src/storage/upload_object/unbuffered.rs
dbolduc: https://github.com/googleapis/google-cloud-rust/blob/main/src%2Fgax%2Fsrc%2Fretry_loop_internal.rs#L72

dbolduc: And attempt/overall timeouts come from the retry policy, and are used in the loop, no?

Is this about setting timeout on the reqwest?

coryan: That is the backoff, no?  What I mean is something like:

https://github.com/googleapis/google-cloud-rust/blob/2f46e4e6798906c72eb8c45294cd63253e3c9556/src/gax-internal/src/http.rs#L112

And then:

https://github.com/googleapis/google-cloud-rust/blob/2f46e4e6798906c72eb8c45294cd63253e3c9556/src/gax-internal/src/http.rs#L136

We are not setting a timeout on the attempt... Maybe we shouldn't, maybe we should rely on minimum throughput or maybe a combination of the two. We are doing neither.

coryan: > And attempt/overall timeouts come from the retry policy,

No.  The attempt timeout should be a combination of the retry policy and any per-request and client configurations. The configurations do not even exist at the moment.

> and are used in the loop, no?

The function called by the retry loop is supposed to apply this combination of retry policy and per-attempt timeouts. We currently are not.

> Is this about setting timeout on the reqwest?

Yes. As the comment says " **apply** any timeouts"

PR #2659, opened 2025-07-14
feat(auth): Allow custom retry policy for user

harkamaljot:
This PR introduces the ability for users to configure custom retry, backoff, and throttling policies for the user account credentials.

### DIFF
--- a/src/auth/src/credentials/user_account.rs
+++ b/src/auth/src/credentials/user_account.rs
@@ -201,8 +201,7 @@ impl Builder {
     ///
     /// ```
     /// # use google_cloud_auth::credentials::user_account::Builder;
-    /// # use gax::retry_policy;
-    /// # use gax::retry_policy::RetryPolicyExt;
+    /// use gax::retry_policy::{self, RetryPolicyExt};
     /// # tokio_test::block_on(async {
     /// let authorized_user = serde_json::json!({
     ///     "client_id": "YOUR_CLIENT_ID.apps.googleusercontent.com",
@@ -235,10 +234,7 @@ impl Builder {
     ///     "refresh_token": "YOUR_REFRESH_TOKEN",
     ///     "type": "authorized_user",
     /// });
-    /// let policy = ExponentialBackoffBuilder::new()
-    ///     .with_initial_delay(Duration::from_millis(100))
-    ///     .with_maximum_delay(Duration::from_secs(5))
-    ///     .with_scaling(4.0)
+    /// let policy = ExponentialBackoffBuilder::default()
     ///     .build().expect("well-known policy values should succeed");
     /// let credentials = Builder::new(authorized_user)
     ///     .with_backoff_policy(policy)
@@ -263,7 +259,7 @@ impl Builder {
     ///
     /// ```
     /// # use google_cloud_auth::credentials::user_account::Builder;
-    /// # use gax::retry_throttler::AdaptiveThrottler;
+    /// use gax::retry_throttler::AdaptiveThrottler;
     /// # tokio_test::block_on(async {
     /// let authorized_user = serde_json::json!({
     ///     "client_id": "YOUR_CLIENT_ID.apps.googleusercontent.com",
@@ -272,8 +268,7 @@ impl Builder {
     ///     "type": "authorized_user",
     /// });
     /// let credentials = Builder::new(authorized_user)
-    ///     .with_retry_throttler(AdaptiveThrottler::new(2.0)
-    ///         .expect("well-known policy values should succeed"))
+    ///     .with_retry_throttler(AdaptiveThrottler::default())
     ///     .build();
     /// # });
     /// ```
--- a/src/auth/src/credentials/user_account.rs
+++ b/src/auth/src/credentials/user_account.rs
@@ -225,7 +225,7 @@ impl Builder {
     ///
     /// ```
     /// # use google_cloud_auth::credentials::user_account::Builder;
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/user_account.rs
```diff
@@ -188,26 +194,101 @@ impl Builder {
         self
     }
 
-    /// Returns a [Credentials] instance with the configured settings.
+    /// Configure the retry policy for fetching tokens.
```
coryan, line 197: I think we need to update the user guide (and maybe the `Builder` comments) to explain how these policies affect the credentials behavior. It would be fine to do that in a future PR.

src/auth/src/credentials/user_account.rs
```diff
@@ -188,26 +194,101 @@ impl Builder {
...
-    /// Returns a [CredentialsError] if the `authorized_user`
-    /// provided to [`Builder::new`] cannot be successfully deserialized into the
-    /// expected format. This typically happens if the JSON value is malformed or
-    /// missing required fields. For more information, on how to generate
-    /// `authorized_user` json, consult the relevant section in the
-    /// [application-default credentials] guide.
+    /// ```
+    /// # use google_cloud_auth::credentials::user_account::Builder;
+    /// # use gax::retry_policy;
+    /// # use gax::retry_policy::RetryPolicyExt;
```
coryan: It may be hard to read the example if you don't see these:

```suggestion
    /// use gax::retry_policy::{self, RetryPolicyExt};
```

src/auth/src/credentials/user_account.rs
```diff
@@ -188,26 +194,101 @@ impl Builder {
...
+
+    /// Configure the retry backoff policy.
     ///
-    /// [application-default credentials]: https://cloud.google.com/docs/authentication/application-default-credentials
-    pub fn build(self) -> BuildResult<Credentials> {
+    /// The backoff policy controls how long to wait in between retry attempts.
+    ///
+    /// ```
+    /// # use google_cloud_auth::credentials::user_account::Builder;
+    /// # use gax::exponential_backoff::ExponentialBackoffBuilder;
```
coryan: Ditto, consider making this visible.

src/auth/src/credentials/user_account.rs
```diff
@@ -188,26 +194,101 @@ impl Builder {
...
+    /// The authentication library throttles its retry loop, using a policy to
+    /// control the throttling algorithm. Use this method to fine tune or
+    /// customize the default retry throttler.
+    ///
+    /// [Handling Overload]: https://sre.google/sre-book/handling-overload/
+    /// [Address Cascading Failures]: https://sre.google/sre-book/addressing-cascading-failures/
+    ///
+    /// ```
+    /// # use google_cloud_auth::credentials::user_account::Builder;
+    /// # use gax::retry_throttler::AdaptiveThrottler;
```
coryan: Ditto.

src/auth/src/credentials/user_account.rs
```diff
@@ -188,26 +194,101 @@ impl Builder {
...
+    ///     "client_id": "YOUR_CLIENT_ID.apps.googleusercontent.com",
+    ///     "client_secret": "YOUR_CLIENT_SECRET",
+    ///     "refresh_token": "YOUR_REFRESH_TOKEN",
+    ///     "type": "authorized_user",
+    /// });
+    /// let policy = ExponentialBackoffBuilder::new()
+    ///     .with_initial_delay(Duration::from_millis(100))
+    ///     .with_maximum_delay(Duration::from_secs(5))
+    ///     .with_scaling(4.0)
+    ///     .build().expect("well-known policy values should succeed");
```
coryan: maybe use `ExponentialBackoffBuilder::default()` so you can skip the `.expect()` and the code is shorter too.

src/auth/src/credentials/user_account.rs
```diff
@@ -188,26 +194,101 @@ impl Builder {
...
+    /// # use gax::retry_throttler::AdaptiveThrottler;
+    /// # tokio_test::block_on(async {
+    /// let authorized_user = serde_json::json!({
+    ///     "client_id": "YOUR_CLIENT_ID.apps.googleusercontent.com",
+    ///     "client_secret": "YOUR_CLIENT_SECRET",
+    ///     "refresh_token": "YOUR_REFRESH_TOKEN",
+    ///     "type": "authorized_user",
+    /// });
+    /// let credentials = Builder::new(authorized_user)
+    ///     .with_retry_throttler(AdaptiveThrottler::new(2.0)
```
coryan: Also consider the `AdaptiveThrottler::default()`.

coryan (approved):
I think you should cleanup the examples a bit...

src/auth/src/credentials/user_account.rs
coryan: I realized the samples are based on the existing client builder and request builder examples.  Maybe we should fix those too

src/auth/src/credentials/user_account.rs
harkamaljot: Sounds good, As this would need to be done for other creds as well. I created a issue to track this: https://github.com/googleapis/google-cloud-rust/issues/2669. I will update the doc after adding default retries for all the builders.

src/auth/src/credentials/user_account.rs
harkamaljot: sounds good, I will send a PR out after this one to fix the existing examples.

alvarowolfx (approved):

Issue #1699, opened 2025-04-03
Full error details from gRPC requests


coryan:
We want to get all the error details with gRPC requests.  The returned error should be a `gax::error::ServiceError` and the `status().details()` should be fully populated.

PR #2660, opened 2025-07-14
impl(storage): full error details in `rpc::Status` messages

dbolduc:
Related to #1699

This makes sure full error details are captured in fields that have `rpc::Status`.

It has no bearing on converting the `tonic::Status` of an RPC with full error details. :frowning_face: 

All `.sidekick.toml` and `status.rs` are copied from `src/firestore`. The only edit was `use crate::google` instead of `use super::google` in `status.rs`.


src/storage/src/control/status.rs
```diff
@@ -0,0 +1,321 @@
...
+impl gaxi::prost::FromProto<rpc::model::Status> for google::rpc::Status {
+    fn cnv(self) -> Result<rpc::model::Status, gaxi::prost::ConvertError> {
+        Ok(rpc::model::Status::new()
+            .set_code(self.code)
+            .set_message(self.message)
+            .set_details(self.details.into_iter().filter_map(any_from_prost)))
+    }
+}
+
+fn any_to_prost(value: wkt::Any) -> Option<prost_types::Any> {
```
coryan, line 37: You probably noticed this: it /should/ be easy to refactor this function (and its inverse) to `gaxi`, it just requires exporting a function that depends on `prost_types::Any` and `wkt::Any`.

src/storage/src/control/generated/protos/rpc/build.rs
```diff
@@ -0,0 +1,39 @@
+// Copyright 2025 Google LLC
```
coryan, line 1: Separately, we should figure out a way to generate these files to $TMP. We only use them when running `sidekick`.

src/storage/src/control/status.rs
dbolduc: I am thinking about it currently. I think it also requires generating `google/rpc` protos in `gaxi`.

src/storage/src/control/generated/protos/rpc/build.rs
dbolduc: Now that you remind me, I remember that we want to do this.

Opened #2662 so GitHub can remember for me.

src/storage/src/control/status.rs
dbolduc: But yeah, you are right that we wouldn't have to expose the generated protos in `gaxi` (or some other common crate) to make this work. That is nice.

PR #2665, opened 2025-07-14
impl(storage): use custom client builder

coryan:
We are going to need ad-hoc options that only apply to storage. And we
do not need other options, such as LROs.

Part of the work for #2056

### DIFF
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -103,10 +103,10 @@ impl Storage {
     ///
     /// # Example
     /// ```
-    /// # tokio_test::block_on(async {
     /// # use google_cloud_storage::client::Storage;
+    /// # async fn sample() -> anyhow::Result<()> {
     /// let client = Storage::builder().build().await?;
-    /// # gax::client_builder::Result::<()>::Ok(()) });
+    /// # Ok(()) }
     /// ```
     pub fn builder() -> ClientBuilder {
         ClientBuilder::new()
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/client.rs
```diff
@@ -101,14 +101,15 @@ pub(crate) struct StorageInner {
 impl Storage {
     /// Returns a builder for [Storage].
     ///
-    /// ```no_run
+    /// # Example
+    /// ```
     /// # tokio_test::block_on(async {
     /// # use google_cloud_storage::client::Storage;
     /// let client = Storage::builder().build().await?;
```
dbolduc, line 108: don't we want to avoid creating credentials? should we wrap this in a fn?

src/storage/src/storage/client.rs
```diff
@@ -266,31 +264,169 @@ impl StorageInner {
 ///
 /// ```
 /// # tokio_test::block_on(async {
-/// # use google_cloud_storage::*;
-/// # use builder::storage::ClientBuilder;
-/// # use client::Storage;
-/// let builder : ClientBuilder = Storage::builder();
+/// # use google_cloud_storage::client::Storage;
```
dbolduc, line 266: same, and I think all of the samples do this.

src/storage/src/storage/client.rs
coryan: Done.

src/storage/src/storage/client.rs
coryan: Done, here and below.

coryan (commented):
PTAL

dbolduc (approved):

PR #2666, opened 2025-07-14
docs: add hint in how to guide for lint errors

suzmue:
With older versions of rust, there end up being lint errors that are no longer relevant. Add a hint for devs if they encounter this problem.

### DIFF
--- a/doc/contributor/howto-guide-set-up-development-environment.md
+++ b/doc/contributor/howto-guide-set-up-development-environment.md
@@ -9,7 +9,7 @@ compile the code, run the unit tests, and formatting miscellaneous files.
 We recommend that you follow the [Getting Started][getting-started-rust] guide.
 Once you have `cargo` and `rustup` installed the rest is relatively easy.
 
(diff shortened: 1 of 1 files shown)


doc/contributor/howto-guide-set-up-development-environment.md
```diff
@@ -74,6 +74,9 @@ cargo fmt && cargo clippy --profile=test -- --deny warnings && cargo test
 git status # Shows any diffs created by `cargo fmt`
 ```
 
+If you are seeing errors when running locally that are not present in the CI,
+you may need to update your local rust version.
```
dbolduc, line 78: Should we bump the minimum version here to one that doesn't yield errors?

https://github.com/googleapis/google-cloud-rust/blob/dd9e4da76c89d9cdf05d18271dfcc2573f847bed/doc/contributor/howto-guide-set-up-development-environment.md?plain=1#L12

suzmue: updated to 1.87 which appears to be the minimum with no errors.

PR #2678, opened 2025-07-15
Style fixes to Vertex guide

pcoet:
Made a few minor style fixes to the Vertex guide:

* Avoid using "we": https://developers.google.com/style/word-list#we
* Use plain verb for heading ("Add" not "Adding"): https://developers.google.com/style/headings
* Move "If you want to run this function in your own code..." after the example. It makes more sense after you've seen the code.
* Several other nitty fixes.

Also, I pointed to a better Vertex setup example. Their quickstart (or the current version, anyway) doesn't actually show how to set up a project and enable the API.


coryan (approved):
Thanks!

PR #2679, opened 2025-07-15
chore: update googleapis SHA circa 2025-07-15

coryan:


src/storage/src/generated/gapic/model.rs
```diff
@@ -10886,12 +10912,13 @@ pub mod bucket {
         #[derive(Clone, Debug, Default, PartialEq)]
         #[non_exhaustive]
         pub struct GoogleManagedEncryptionEnforcementConfig {
-            /// Whether Google Managed Encryption (GMEK) is restricted for new
-            /// objects within the bucket.
-            /// If true, new objects can't be created using GMEK encryption.
-            /// If false or unset, creation of new objects with GMEK encryption is
-            /// allowed.
-            pub restricted: std::option::Option<bool>,
```
dbolduc, line 10894: Looks like a breaking change from the service 😒

src/storage/src/generated/gapic/model.rs
```diff
@@ -15830,6 +15863,350 @@ impl serde::ser::Serialize for ObjectChecksums {
     }
 }
 
+/// The payload of a single user-defined object context.
+#[derive(Clone, Debug, Default, PartialEq)]
+#[non_exhaustive]
+pub struct ObjectCustomContextPayload {
```
dbolduc, line 15869: (assuming the builds pass), nice, looks like the include list handled a new storage/v2 type.

Issue #2626, opened 2025-07-10
Add method for getting object metadata from Storage::ReadObjectResponse


suzmue:
Object metadata is returned from requests, either [via HTTP/1.1 response headers](https://github.com/googleapis/java-storage/blob/140f6c68251b30cfc0598373dce56463a7acfeac/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableWrite.java#L135-L150), or as a [sibling field in gRPC](https://github.com/googleapis/googleapis/blob/09ba8d83538db44efe3b81eb46bbdaf4828771b7/google/storage/v2/storage.proto#L978). The API should expose this metadata.

suzmue:
Other fields to add: content-language, content-type, x-goog-storage-class)

coryan:
The remaining work can happen in the next milesone, moving.

suzmue:
List of all headers used by the JSON API: http://cloud/storage/docs/json_api/v1/parameters

PR #2685, opened 2025-07-15
feat(storage): add method to retrieve object metadata

suzmue:
There is some object metadata returned [via HTTP/1.1 response headers](https://github.com/googleapis/java-storage/blob/140f6c68251b30cfc0598373dce56463a7acfeac/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableWrite.java#L135-L150), or as a [sibling field in gRPC](https://github.com/googleapis/googleapis/blob/09ba8d83538db44efe3b81eb46bbdaf4828771b7/google/storage/v2/storage.proto#L978) that we want to return to users (see [java implementation](https://github.com/googleapis/java-storage/blob/140f6c68251b30cfc0598373dce56463a7acfeac/google-cloud-storage/src/main/java/com/google/cloud/storage/JsonResumableWrite.java#L135-L150) for another example). To make it more clear that not all fields will be filled out, we return a ObjectHighlights that includes a subset of the fields.

For #2626


src/storage/src/storage/read_object.rs
```diff
@@ -541,6 +607,28 @@ impl ReadObjectResponse {
     }
 }
 
+/// ObjectHighlights contains select metadata from a [crate::model::Object].
+pub struct ObjectHighlights {
+    /// The content generation of this object. Used for object versioning.
+    pub generation: i64,
+    /// The version of the metadata for this generation of this
```
coryan, line 701: nit: blank line ?

src/storage/src/storage/read_object.rs
```diff
@@ -541,6 +607,28 @@ impl ReadObjectResponse {
...
+    pub size: i64,
+    /// Content-Encoding of the object data, matching
+    /// [<https://tools.ietf.org/html/rfc7231#section-3.1.2.2>][RFC 7231 §3.1.2.2]
+    pub content_encoding: String,
+    /// Hashes for the data part of this object. The checksums of the complete
+    /// object regardless of data range. If the object is downloaded in full,
+    /// the client should compute one of these checksums over the downloaded
+    /// object and compare it against the value provided here.
+    pub checksums: std::option::Option<crate::model::ObjectChecksums>,
+}
```
coryan, line 720: For a future PR: I there are more things we can get (`content-language`, `content-type`, `x-goog-storage-class`), but this is a excellent set of things to capture first. And we can always add more.

src/storage/src/storage/read_object.rs
```diff
@@ -541,6 +607,28 @@ impl ReadObjectResponse {
     }
 }
 
+/// ObjectHighlights contains select metadata from a [crate::model::Object].
+pub struct ObjectHighlights {
```
coryan:
```suggestion
/// ObjectHighlights contains select metadata from a [crate::model::Object].
#[derive(Clone, Debug, PartialEq)]
pub struct ObjectHighlights {
```

src/storage/src/storage/read_object.rs
suzmue: Done.

src/storage/src/storage/read_object.rs

src/storage/src/storage/read_object.rs
suzmue: Ack! Added comment to issue

PR #2687, opened 2025-07-15
ci(storage): fail build on new RPCs

### LINKED ISSUE #2400: Block googleapis updates on new storage-control RPCs.
A good first step to #2398.

We want to block updates to googleapis on new storage-control RPCs.

We can use a custom `build.rs` file to maintain the `stub_tests` macro.

https://github.com/googleapis/google-cloud-rust/blob/499eec5a9985d68d1d8d3186b54fcf04189630cb/src/storage-control/tests/mocking.rs#L97

Then when a new RPC is added to storage-control, we will have failing tests until the associated manual changes are made.

dbolduc:
Fixes #2400 

Fails a build of `google-cloud-storage` if there are RPCs missing in the client surface. This can happen on an update of googleapis.

I am not sure if this is feasible via a test. Test environments are normally pretty hermetic.

We could gate this behind a feature, and enable the feature specially in one of our builds.

### DIFF
--- a/src/storage/Cargo.toml
+++ b/src/storage/Cargo.toml
@@ -16,6 +16,7 @@
 name        = "google-cloud-storage"
 version     = "0.25.0-preview1"
 description = "Google Cloud Client Libraries for Rust - Storage"
+build       = "build.rs"
 # Inherit other attributes from the workspace.
 edition.workspace      = true
 authors.workspace      = true
@@ -75,3 +76,7 @@ tempfile.workspace       = true
 tokio-test.workspace     = true
 tokio.workspace          = true
 google-cloud-storage     = { path = ".", features = ["unstable-stream"] }
+
+[build-dependencies]
+anyhow.workspace = true
+regex            = { workspace = true, features = ["std"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4262,6 +4262,7 @@ dependencies = [
  "pin-project",
  "prost",
  "prost-types",
+ "regex",
  "reqwest",
  "serde",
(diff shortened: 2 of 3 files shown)


src/storage/Cargo.toml
```diff
@@ -16,6 +16,7 @@
 name        = "google-cloud-storage"
 version     = "0.25.0-preview1"
 description = "Google Cloud Client Libraries for Rust - Storage"
+build       = "build.rs"
```
coryan, line 19: Eeeek. This gets shipped to customers, and it becomes part of what they run on `cargo build ${thing that uses cloud storage}`...

coryan:
> 
> We could gate this behind a feature, and enable the feature specially in one of our builds.

We already have a feature:

https://github.com/googleapis/google-cloud-rust/blob/6bb4cfaa1d900db7d4ea3bc673fe0c92e2e33959/src/storage/src/generated/protos/control/Cargo.toml#L28-L29

Could we make this run in the `build.rs` for `control` and `gapic_control`?  There is already a way to customize those build files:

https://github.com/googleapis/google-cloud-rust/blob/6bb4cfaa1d900db7d4ea3bc673fe0c92e2e33959/src/storage/src/generated/protos/control/.sidekick.toml#L22

and that runs only when the code is regenerated.

src/storage/Cargo.toml
dbolduc: Yeah, I am not stoked about that.

dbolduc:
We are not going to do this via a `build.rs`, so this PR is dead.

PR #2704, opened 2025-07-16
ci: test storage features

coryan:


dbolduc (approved):
Did you consider `--all-targets` for the runs with default features?

coryan:
> Did you consider `--all-targets` for the runs with default features?

No, but seems like a good idea. Done.

Issue #2634, opened 2025-07-11
Use resumable uploads for large `send_unbuffered()` calls


coryan:
For very large payloads, even if they support `Seek`, we may want to use resumable uploads in `send_unbuffered()`. If the upload is interrupted (which is likely when the object is really large) you want to avoid resending a lot of the data. Keep in mind that the maximum object size is 5TiB, which may take hours to upload.  Restarting from the beginning, while possible, is not ideal.

I am not sure if we need this for GA: the API does not break if we add this feature, but the behavior changes for the better.

coryan:
Fixed in #2730, I tagged the wrong thing.

PR #2705, opened 2025-07-16
impl(storage): resumable upload size options

coryan:
I will use this for #2634

### DIFF
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -423,6 +423,80 @@ impl ClientBuilder {
         self.default_options.retry_throttler = v.into().into();
         self
     }
+
+    /// Changes the threshold for resumable uploads.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_storage::client::Storage;
+    /// # async fn sample() -> anyhow::Result<()> {
+    /// let client = Storage::builder()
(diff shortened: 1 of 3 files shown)


src/storage/src/storage/client.rs
```diff
@@ -423,6 +423,80 @@ impl ClientBuilder {
         self.default_options.retry_throttler = v.into().into();
         self
     }
+
+    /// Changes the threshold for resumable uploads.
+    ///
```
dbolduc, line 428: Can we explain threshold?

"Objects whose size is lower than the threshold will be uploaded in a single shot. Objects whose size is greater than the threshold will be uploaded using a resumable upload."

src/storage/src/storage/upload_object.rs
```diff
@@ -636,6 +636,74 @@ impl<T> UploadObject<T> {
         self
     }
 
+    /// Changes the threshold for resumable uploads.
+    ///
```
dbolduc, line 640: Same

dbolduc (dismissed):

src/storage/src/storage/client.rs
coryan: I prefer to keep the summary to a single sentence. And there is a longer description after the example.

Let me know if that works.

dbolduc: Much clearer summary, thanks.

dbolduc (approved):

PR #2706, opened 2025-07-16
cleanup(auth)!: update retry loop

harkamaljot:
### What this PR does

* Modifies the authentication retry loop to consistently wrap and return a non-transient error.

### Why this change is needed

* The previous implementation returned either transient or non-transient errors, forcing the GAX retry policy to have complex, special-case logic to handle them. As auth errors should never be retried from Gax layer.

### Important Caveat

* This PR does **not** yet implement a default retry policy for the auth layer.
* As a result, clients without a custom retry policy we add a default retry policy with attempts equal to 1.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1253,7 +1253,7 @@ dependencies = [
 
 [[package]]
 name = "google-cloud-auth"
-version = "0.22.1"
+version = "0.22.2"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -334,7 +334,7 @@ tokio-stream   = { default-features = false, version = "0.1" }
 tokio-test     = { default-features = false, version = "0.4" }
 
 # Local packages used as dependencies
-auth                          = { version = "0.22.1", path = "src/auth", package = "google-cloud-auth" }
+auth                          = { version = "0.22.2", path = "src/auth", package = "google-cloud-auth" }
 gax                           = { version = "0.23.2", path = "src/gax", package = "google-cloud-gax" }
 gaxi                          = { version = "0.4.1", path = "src/gax-internal", package = "google-cloud-gax-internal" }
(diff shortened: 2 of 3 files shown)


coryan:
We already have this:

https://github.com/googleapis/google-cloud-rust/blob/aef399f2ce13a5ec21898a995e910d5b374078f8/src/gax/src/retry_policy.rs#L325

harkamaljot:
> We already have this:
> 
> https://github.com/googleapis/google-cloud-rust/blob/aef399f2ce13a5ec21898a995e910d5b374078f8/src/gax/src/retry_policy.rs#L325

Got it, thanks @coryan. In that case, I'm still unclear on the use case for NoRetryPolicy, since my goal is to retry on most errors while avoiding retries on authentication failure.

Is the expectation that our default policy handles this correctly (by not retrying auth errors), and that we should honor any custom policy a customer provides to Gax, even if it retries on everything?

coryan:
> > We already have this:
> > https://github.com/googleapis/google-cloud-rust/blob/aef399f2ce13a5ec21898a995e910d5b374078f8/src/gax/src/retry_policy.rs#L325
> 
> Got it, thanks @coryan. In that case, I'm still unclear on the use case for NoRetryPolicy, since my goal is to retry on most errors while avoiding retries on authentication failure.

Then you need to change `Aip194Strict`.

> Is the expectation that our default policy handles this correctly (by not retrying auth errors), and that we should honor any custom policy a customer provides to Gax, even if it retries on everything?

Yes.

src/auth/src/retry.rs
```diff
@@ -64,9 +64,15 @@ impl Builder {
             None => Arc::new(Mutex::new(AdaptiveThrottler::default())),
         };
 
+        let retry_policy = match self.retry_policy {
+            Some(p) => p,
+            None => AlwaysRetry.with_attempt_limit(1).into(),
```
sai-sunder-s, line 69: In a future PR, we can update to use the same default retry policy was gax.

harkamaljot: perfect sounds good.

PR #2710, opened 2025-07-17
impl(storage): configurable buffered upload types

### LINKED ISSUE #2056: Automatically retry failed single-shot uploads

### LINKED ISSUE #2709: Use resumable uploads for large `send_unbuffered() / send()` requests

coryan:
Switch the upload type based on the size of the payload. Small uploads
perform better when using single-shot uploads. Large uploads require
resumable uploads to keep the buffer size under control, and the extra
latency is not that bad in relative terms.

Fixes #2709, fixes #2056

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4231,7 +4231,7 @@ dependencies = [
 
 [[package]]
 name = "google-cloud-storage"
-version = "0.25.0-preview3"
+version = "0.25.0-preview4"
 dependencies = [
  "anyhow",
  "async-trait",
--- a/src/storage/Cargo.toml
+++ b/src/storage/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name        = "google-cloud-storage"
-version     = "0.25.0-preview3"
+version     = "0.25.0-preview4"
 description = "Google Cloud Client Libraries for Rust - Storage"
 # Inherit other attributes from the workspace.
 edition.workspace      = true


src/storage/src/storage/upload_object/buffered.rs
```diff
@@ -33,22 +33,16 @@ where
     /// # Ok(()) }
     /// ```
     pub async fn send(self) -> crate::Result<Object> {
-        let upload_url = self.start_resumable_upload().await?;
         // TODO(#2043) - make the threshold to use resumable uploads and the
         //    target size for each chunk configurable.
```
dbolduc: nit: stale comment

src/storage/src/storage/upload_object/buffered.rs
```diff
@@ -33,22 +33,16 @@ where
...
-        if self.payload.lock().await.size_hint().0 > RESUMABLE_UPLOAD_QUANTUM as u64 {
+        let hint = self.payload.lock().await.size_hint().1;
+        if hint.is_some_and(|max| max >= self.options.resumable_upload_threshold as u64) {
```
dbolduc: Q: should the default for unknown-sized objects be resumable uploads? i.e. `s/is_some_and/is_none_or/`

src/storage/src/storage/upload_source.rs
```diff
@@ -205,23 +214,84 @@ impl Seek for BytesSource {
     }
 }
 
+pub struct IterSource {
```
dbolduc, line 217: is this intentionally public?

src/storage/src/storage/upload_source.rs
```diff
@@ -205,23 +214,84 @@ impl Seek for BytesSource {
...
+    fn size_hint(&self) -> (u64, Option<u64>) {
+        let s = self.contents.iter().fold(0_u64, |a, i| a + i.len() as u64);
+        (s, Some(s))
+    }
+}
+
+impl Seek for IterSource {
+    type Error = std::io::Error;
+    async fn seek(&mut self, offset: u64) -> std::result::Result<(), Self::Error> {
+        let mut current: VecDeque<_> = self.contents.iter().cloned().collect();
```
dbolduc: if this is a public type, this implementation should be more efficient. We only need to clone the chunks from `contents` which have already been consumed from `current`.

dbolduc (commented):
Move `IterSource` back into `mod tests`?

src/storage/src/storage/upload_object/buffered.rs
coryan: Fixed.

src/storage/src/storage/upload_object/buffered.rs
coryan: D'oh! yes.

src/storage/src/storage/upload_source.rs
coryan: Hmm... clone for `bytes::Bytes` is just incrementing a reference count, but sure.

src/storage/src/storage/upload_source.rs
coryan: I think eventually it will need to be (to return `InsertPayload<IterSource>`, but let's leave it `pub(crate)` for now.

src/storage/src/storage/upload_source.rs
dbolduc: My mistake, I forgot the clone was super cheap

dbolduc (approved):

PR #2364, opened 2025-06-03
test(wkt): more primitive types integration tests

coryan:
Add more integration tests where primitive types appear as fields of
messages.

This made the `message_with_map` test redundant, we test the map fields
with each primitive type.

Motivated by #2349, I will need these tests in place when the "null as default" thing is implemented.

PR #2730, opened 2025-07-20
impl(storage): unbuffered resumable uploads

coryan:
Use resumable uploads for `send_unbuffered()` if the upload is large or
of unknown size. Resumable uploads can recover more efficiently from
interrupted uploads.

Fixes #2364


src/integration-tests/src/storage.rs
```diff
@@ -24,6 +24,72 @@ use storage::model::Bucket;
...
+        Self {
+            size,
+            hint: (size, Some(size)),
+            offset: 0,
+        }
+    }
+    fn without_size_hint(size: u64) -> Self {
+        Self {
+            size,
+            hint: (size, None),
```
dbolduc: nit: `(0, None)`

src/storage/src/storage/upload_object/unbuffered.rs
```diff
@@ -35,6 +36,114 @@ where
...
+        &self,
+        url: &mut Option<String>,
+        hint: (u64, Option<u64>),
+    ) -> Result<Object> {
+        let offset = if let Some(upload_url) = url {
+            match self.query_resumable_upload(upload_url).await? {
+                ResumableUploadStatus::Finalized(o) => {
+                    return Ok(*o);
+                }
+                ResumableUploadStatus::Partial(o) => o,
```
dbolduc: nit: spell out `object` and/or `offset`, because these `o`s are totally different ?

src/storage/src/storage/upload_object/unbuffered.rs
```diff
@@ -35,6 +36,114 @@ where
...
+            }
+        } else {
+            let upload_url = self.start_resumable_upload_attempt().await?;
+            *url = Some(upload_url);
+            0_u64
+        };
+        assert!(url.is_some());
```
dbolduc: remove?

src/storage/src/storage/upload_object/unbuffered.rs
```diff
@@ -35,6 +36,114 @@ where
...
+            .lock()
+            .await
+            .seek(offset)
+            .await
+            .map_err(Error::ser)?;
+
+        let range = match (offset, hint.0, hint.1) {
+            (o, _, None) => format!("bytes {o}-*/*"),
+            (_, 0, Some(0)) => "bytes */0".to_string(),
+            (o, s, Some(u)) if s == u => format!("bytes {o}-{}/{s}", s - 1),
```
dbolduc, line 98: I see that this is the case where we know the size of the object.

Q: So we do not have to break the payload into partial uploads and there is only one `Content-Range` header per stream?

https://cloud.google.com/storage/docs/performing-resumable-uploads#chunked-upload
https://cloud.google.com/storage/docs/streaming-uploads#json-api

src/storage/src/storage/upload_object/unbuffered.rs
```diff
@@ -35,6 +36,114 @@ where
...
+                if let Some(next) = guard.next().await {
+                    drop(guard);
+                    return Some((next, Some(payload)));
+                }
+            }
+            None
+        }));
+
+        let builder = builder.body(reqwest::Body::wrap_stream(stream));
+        let response = builder.send().await.map_err(Error::io)?;
```
dbolduc, line 126: does reqwest send the stream in chunks?

src/storage/src/storage/upload_object/unbuffered/resumable_tests.rs
```diff
@@ -0,0 +1,795 @@
...
+//!    - If the request fails with a non-retryable error return immediately.
+//!    - If the request fails with a retryable error and the retry policy is
+//!      **not** exhausted, go to step 3.
+//!    - If the request fails with a retryable error and the retry policy
+//!      **is** exhausted, return immediately.
+//! 3. Query the resumable upload session to find the persisted size.
+//!    - If that succeeds, go to 4.
+//!    - If that fails with a non-retryable error, return immediately.
+//!    - If that fails with a retryable error, try step 3 again until the
+//!      retry policy.
```
dbolduc: "retry policy is exhausted."

dbolduc (approved):
I just have questions about the `Content-Range` protocol

src/storage/src/storage/upload_object/unbuffered/resumable_tests.rs
coryan: Done.

src/storage/src/storage/upload_object/unbuffered.rs
coryan: "chunk" is the most overloaded term.  I think the short answer is "no". This will send a single PUT request with all the data.

However, because the size is not known, it will use chunk-transfer encoding, but those chunks have nothing to do with the PUT transfers, which we also call chunks in places.

src/storage/src/storage/upload_object/unbuffered.rs
coryan: > Q: So we do not have to break the payload into partial uploads and there is only one Content-Range header per stream?

No. That is the point of using unbuffered uploads with `Seek`: send all the data in a single PUT request. If it succeeds you get really good performance (all the TCP windows get large, the server can optimize writes, etc.).  If the PUT request fails, you query the status of the upload and send another PUT request from the last persisted byte: that is where having a `Seek` implementation helps.

src/storage/src/storage/upload_object/unbuffered.rs
coryan: I was leaving it as documentation... Add a comment to explain

src/storage/src/storage/upload_object/unbuffered.rs
coryan: Done.

src/integration-tests/src/storage.rs
coryan: Done. aside: I think the minimum size hint is not proving useful and I should remove it.

coryan (commented):
PTAL

src/storage/src/storage/upload_object/unbuffered.rs
dbolduc: Thanks for the explainer.

src/storage/src/storage/upload_object/unbuffered.rs
dbolduc: meh. Consider something like:

```rs
let (offset, url_ref) = if let Some(upload_url) = url {
            match self.query_resumable_upload(upload_url).await? {
                ResumableUploadStatus::Finalized(o) => {
                    return Ok(*o);
                }
                ResumableUploadStatus::Partial(o) => (o, upload_url),
            }
        } else {
            let upload_url = self.start_resumable_upload_attempt().await?;
            (0, url.insert(upload_url).as_str())
    };
// use `url_ref` below instead of `url`
```

dbolduc (approved):

src/storage/src/storage/upload_object/unbuffered.rs
dbolduc: https://github.com/googleapis/google-cloud-rust/pull/2739

PR #2736, opened 2025-07-21
chore: update googleapis SHA circa 2025-07-21

coryan:
`google-cloud-policysimulator-v1` gained a new ad-hoc dependency which
required manual changes to the top-level `Cargo.toml`, and
`.sidekick.toml` files.

Manually updated the version for the affected libraries.


src/gax/src/exponential_backoff.rs
```diff
@@ -244,6 +244,7 @@ impl crate::backoff_policy::BackoffPolicy for ExponentialBackoff {
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::mock_rng::MockRng;
```
dbolduc: could have rebased, but merge should be fine.

dbolduc (approved):
nit: manual edits to `.sidekick.toml` don't belong in an "Update generated files" commit.

src/gax/src/exponential_backoff.rs
coryan: Fixed.  And we don't even need a second review.

PR #2740, opened 2025-07-21
feat(types): Hide empty _unknown_fields for debugging

whuffman36:


generator/internal/rust/templates/common/message.mustache
```diff
@@ -338,6 +338,27 @@ impl serde::ser::Serialize for {{Codec.Name}} {
         state.end()
     }
 }
+
+#[doc(hidden)]
```
coryan: I think we want to leave these in the documentation, so people know that they can use `{message:?}`, e.g.

https://docs.rs/google-cloud-storage/0.25.0-preview3/google_cloud_storage/model/struct.Bucket.html#impl-Default-for-Bucket

coryan (changes requested):
It would be good to implement one integration test in `wkt` using one of the (generated) test types.  Make sure you test the case with and without values in the `_unknown_fields`.

src/wkt/tests/message_with_string.rs
```diff
@@ -228,4 +228,12 @@ mod tests {
         assert_eq!(output, json!({}));
         Ok(())
     }
+
+    #[test_case(json!({}), "MessageWithString { singular: \"\", optional: None, repeated: [], map_value: {}, map_key: {}, map_key_value: {} }")]
+    #[test_case(json!({"unknown0": 23, "unknown1": "test"}), "MessageWithString { singular: \"\", optional: None, repeated: [], map_value: {}, map_key: {}, map_key_value: {}, _unknown_fields: {\"unknown0\": Number(23), \"unknown1\": String(\"test\")} }")]
+    fn test_debug_trait(input: Value, debug_str: &str) -> Result {
+        let got = serde_json::from_value::<MessageWithString>(input)?;
+        assert_eq!(format!("{got:?}"), debug_str);
```
coryan: I think this test depends on the order for unknown fields, and since they are in a hashmap, that seems brittle.

Consider testing for "the `_unknown_fields` thing appears in the formatted debug test" and "the `unknown0` field and value also appear", etc.

PR #2741, opened 2025-07-22
impl(storage): resume large buffered uploads

### LINKED ISSUE #2057: Automatically resume interrupted resumable uploads

coryan:
Implement resume for large buffered uploads. Previously we just used
resumable uploads, but did not handle any errors or partial uploads.

The basic idea is to keep an in-memory buffer with that data that has
not been persisted in the service. When the buffer is full we send the
`PUT` request. If that request is successful (or partially successful)
we discard the data persisted on the service and continue from that
point.

Fixes #2057

### DIFF
--- a/src/storage/src/storage/upload_object/buffered.rs
+++ b/src/storage/src/storage/upload_object/buffered.rs
@@ -54,8 +54,12 @@ where
 
     async fn send_buffered_resumable(self, hint: (u64, Option<u64>)) -> Result<Object> {
         let mut progress = InProgressUpload::new(self.options.resumable_upload_buffer_size, hint);
+        let mut url = None;
         gax::retry_loop_internal::retry_loop(
-            async |_| self.buffered_resumable_attempt(&mut progress).await,
+            async |_| {
+                self.buffered_resumable_attempt(&mut progress, &mut url)
+                    .await
+            },
             async |duration| tokio::time::sleep(duration).await,
             true,
             self.options.retry_throttler.clone(),
@@ -65,18 +69,22 @@ where
         .await
     }
 
-    async fn buffered_resumable_attempt(&self, progress: &mut InProgressUpload) -> Result<Object> {
-        // Cannot borrow `progress.url` because we plan to borrow `progress` as
-        // mutable below.
-        let upload_url = if let Some(url) = progress.upload_session() {
-            url
+    // Use separate arguments for `progress` and `url` so we can borrow them
+    // separately.
+    async fn buffered_resumable_attempt(
+        &self,
+        progress: &mut InProgressUpload,
+        url: &mut Option<String>,
+    ) -> Result<Object> {
+        let upload_url = if let Some(u) = url.as_deref() {
+            u
         } else {
-            let url = self.start_resumable_upload_attempt().await?;
-            progress.set_upload_session(url)
+            let u = self.start_resumable_upload_attempt().await?;
+            url.insert(u).as_str()
         };
 
         if progress.needs_query() {
-            match self.query_resumable_upload_attempt(&upload_url).await? {
+            match self.query_resumable_upload_attempt(upload_url).await? {
                 ResumableUploadStatus::Finalized(object) => return Ok(*object),
                 ResumableUploadStatus::Partial(persisted_size) => {
                     progress.handle_partial(persisted_size)?;
@@ -88,7 +96,7 @@ where
             progress
                 .next_buffer(&mut *self.payload.lock().await)
                 .await?;
-            let builder = self.partial_upload_request(&upload_url, progress).await?;
+            let builder = self.partial_upload_request(upload_url, progress).await?;
             let response = builder.send().await.map_err(Error::io)?;
             match super::query_resumable_upload_handle_response(response).await {
                 Err(e) => {


src/storage/src/storage/upload_object/buffered/resumable_tests.rs
```diff
@@ -0,0 +1,741 @@
...
+//!
+//! When the upload source does **not** implement [Seek] single-shot uploads are
+//! only suitable for relatively small sources. To handle failures in the upload
+//! all the data needs to be buffered in memory, as the source offers no
+//! "rewind" feature.
+//!
+//! Resumable uploads can be used for artitrarily large uploads (up to the
+//! service limit), but we need to perform some amount of buffering. Recall that
+//! resumable uploads consist of a POST request to create the resumable upload
+//! session, followed by a number of PUT requests with the data. If the POST
```
BenWhitehead: Dangling `If the POST`?

src/storage/src/storage/upload_object/buffered.rs
```diff
@@ -147,122 +150,218 @@ where
...
-    remainder: Option<bytes::Bytes>,
+#[derive(Error, Debug)]
+enum ProgressError {
+    #[error(
+        "the service previously persisted {offset} bytes, but now reports only {persisted} as persisted"
+    )]
+    UnexpectedRewind { offset: u64, persisted: u64 },
+    #[error("the service reports {persisted} bytes as persisted, but we only sent {sent} bytes")]
+    TooMuchProgress { sent: u64, persisted: u64 },
 }
```
BenWhitehead: I've got an enum similar to this in java with some more scenarios https://github.com/googleapis/java-storage/blob/6ba56e5a4b86a75a9f48beccf79ff6d5fdd3e19f/google-cloud-storage/src/main/java/com/google/cloud/storage/ResumableSessionFailureScenario.java#L51-L82

src/storage/src/storage/upload_object/buffered.rs
```diff
@@ -147,122 +150,218 @@ where
...
-            size = target_size;
-            partial.push_back(b);
-            Some(Some(remainder))
-        }
-        n if size + n == target_size => {
-            size = target_size;
-            partial.push_back(b);
-            Some(None)
+    /// The expected size for the full object.
+    hint: (u64, Option<u64>),
```
BenWhitehead: I'd expect a size to be a single value, not a tuple. Maybe this comment needs to be updated to explain what the tuple values mean?

src/storage/src/storage/upload_object/buffered/resumable_tests.rs
coryan: Fixed.

src/storage/src/storage/upload_object/buffered.rs
coryan: It is kind of idiomatic: all streams return a `(minimum, Option<maximum>)` pair for the size hint.  Aside: I think I will remove the minimum anyway.

src/storage/src/storage/upload_object/buffered.rs
coryan: I am marking this `#[non_exhaustive]` so we can add more variants in the future. Seems like a good idea to include final size and final CRC mismatches.

src/storage/src/storage/upload_object/buffered/resumable_tests.rs
```diff
@@ -0,0 +1,744 @@
...
+//!
+//! A separate module eases navigation and provides some structure to explain
+//! the testing strategy.
+//!
+//! When the upload source does **not** implement [Seek] single-shot uploads are
+//! only suitable for relatively small sources. To handle failures in the upload
+//! all the data needs to be buffered in memory, as the source offers no
+//! "rewind" feature.
+//!
+//! Resumable uploads can be used for artitrarily large uploads (up to the
```
dbolduc: s/artitrarily/arbitrarily/

src/storage/src/storage/upload_object/buffered/resumable_tests.rs
```diff
@@ -0,0 +1,744 @@
...
+//! service limit), but we need to perform some amount of buffering. Recall that
+//! resumable uploads consist of a POST request to create the resumable upload
+//! session, followed by a number of PUT requests with the data.
+//! The POST request may fail. We use the normal retry loop to handle the
```
dbolduc, line 30: nit: line break before this?

src/storage/src/storage/upload_object/buffered/resumable_tests.rs
```diff
@@ -0,0 +1,744 @@
...
+//! failure.
+//!
+//! The PUT requests may fail, in which case the client library need to resend
```
dbolduc: s/need/needs/

src/storage/src/storage/upload_object/buffered/resumable_tests.rs
```diff
@@ -0,0 +1,744 @@
...
+//!    - If the request fails with a non-retryable error return immediately.
+//!    - If the request fails with a retryable error and the retry policy is
+//!      **not** exhausted, go to step 5.
+//!    - If the request fails with a retryable error and the retry policy
+//!      **is** exhausted, return immediately.
+//! 4. A 308 response indicates that all or at least part of the `PUT` request
+//!    was successful.
+//!    - Discard the portion of the buffer successfully persisted.
+//!    - Add the remaining portion of the buffer to any data saved in step 2.
+//!    - Got back to step 2.
```
dbolduc: s/Got/Go/

src/storage/src/storage/upload_object/buffered.rs
```diff
@@ -39,73 +41,74 @@ where
...
+            self.options.retry_throttler.clone(),
+            Arc::new(ContinueOn308::new(self.options.retry_policy.clone())),
+            self.options.backoff_policy.clone(),
+        )
+        .await
+    }
+
+    async fn buffered_resumable_attempt(&self, progress: &mut InProgressUpload) -> Result<Object> {
+        // Cannot borrow `progress.url` because we plan to borrow `progress` as
+        // mutable below.
```
dbolduc: could split out a `url: &mut Option<String>` from `InProgressUpload`. It is already split out in the `partial_upload_request` signature.

dbolduc (dismissed):
PR too big to look closely

src/storage/src/storage/upload_object/buffered.rs
coryan: I do not believe that is possible:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a7be6fe66df2636372c7538d0b527cbc

A fundamental rule in Rust is that there can be only one mutable reference to a thing at a time *and* if there is a mutable reference there are no other references:

https://doc.rust-lang.org/book/ch04-02-references-and-borrowing.html#the-rules-of-references

src/storage/src/storage/upload_object/buffered/resumable_tests.rs
coryan: Thanks.  It feels like `typos` should have caught that.

src/storage/src/storage/upload_object/buffered/resumable_tests.rs
coryan: Fixed.

src/storage/src/storage/upload_object/buffered/resumable_tests.rs
coryan: Fixed.

src/storage/src/storage/upload_object/buffered/resumable_tests.rs
coryan: Fixed.

coryan (commented):
> PR too big to look closely

Fair.  I think I can break out the `InProgressUpload` class and its tests first and see how it looks after that.

coryan:
> > PR too big to look closely
> 
> Fair. I think I can break out the `InProgressUpload` class and its tests first and see how it looks after that.

I broke that code to #2749, once that gets merged this PR will get smaller. Still big, but mostly with tests.

coryan:
PTAL. I hope it is easier to grok what is going on.

dbolduc (approved):
Thanks

PR #2749, opened 2025-07-22
impl(storage): a helper for in-progress uploads

coryan:
To resume a buffered, resumable upload we will need to keep significant
state, with some invariants. A helper type to keep this state seems
useful. In this PR it duplicates existing functionality. A follow-up PR
will clean that up.

Part of the work for #2057

### DIFF
--- a/src/storage/src/storage/upload_object/buffered/progress.rs
+++ b/src/storage/src/storage/upload_object/buffered/progress.rs
@@ -1,0 +1,656 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::RESUMABLE_UPLOAD_QUANTUM;
+use super::StreamingSource;
+use crate::Error;
+use crate::Result;
+use crate::storage::UploadError;
+use futures::stream::unfold;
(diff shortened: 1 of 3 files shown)


src/storage/src/storage/upload_object/buffered/progress.rs
```diff
@@ -0,0 +1,675 @@
...
+        let mut payload = InsertPayload::from("");
+        let mut upload = InProgressUpload::new(0, (0, None));
+        upload.next_buffer(&mut payload).await?;
+        assert_eq!(upload.range_header(), "bytes */0");
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn range_header_unknown_size() -> Result {
+        const LINES: i32 = 257;
```
dbolduc: optional: `RESUMABLE_UPLOAD_QUANTUM / 1024 + 1`

src/storage/src/storage/upload_object/buffered/progress.rs
```diff
@@ -0,0 +1,675 @@
...
+    /// and greater than 0.
+    target_size: usize,
+    /// The expected size [minimum, maximum) for the full object.
+    ///
+    /// If the maximum is `None` the maximum size is not known.
+    hint: (u64, Option<u64>),
+    /// The upload session URL.
+    ///
+    /// Starts as `None` and is initialized before the first `PUT` request.
+    url: Option<String>,
```
dbolduc: FTR, I was suggesting moving `url` out of this class. Then we can take a mutable reference to this class without needing to take one for `url`.

The `clone()`s are no big deal though.

dbolduc (dismissed):

src/storage/src/storage/upload_object/buffered/progress.rs
coryan: Done.

src/storage/src/storage/upload_object/buffered/progress.rs
coryan: Ah, will do.

coryan:
PTAL

dbolduc (approved):

Issue #2398, opened 2025-06-09
Consider writing storage-control surface via custom `build.rs`


dbolduc:
Copying RPCs into the `storage-control` client surface is rote. We do it once when an RPC is added.

This is a manual process. It would preferably be automated as part of updating googleapis.

Also, there is nothing updating the client documentation. It is only copied once.

We could fix these problems with a custom `build.rs` that writes the `client.rs` surface for us, as well as the `stub.rs` surface.

Note that we expect the need for this feature to be temporary. That is why we think customizing `build.rs` is appropriate, vs. adding specific support in sidekick for this.

dbolduc:
I am planning to

- add a `rust-refreshall` sidekick command that writes the client after generation occurs
- use strings / regex operations to rewrite the code from the individual clients to the consolidated client
- use a mustache template to inject the variable code. This is not controlled by a `.sidekick.toml`

Short of updating the CI and beautifying the code, I have this working:

https://github.com/dbolduc/google-cloud-rust/commit/62e60b84154a56f44d22e7f37fe2a485d5c01511

We lose the handwritten examples. If examples are auto-generated for the GAPICs, we should get them back. 🤷

coryan:
> * add a `rust-refreshall` sidekick command that writes the client after generation occurs

So far I am with you.

> * use strings / regex operations to rewrite the code from the individual clients to the consolidated client

It seems more robust to load the `api.API` models for both services and give both of them to an ad-hoc generator.  You would need to refactor this function:

https://github.com/googleapis/google-cloud-rust/blob/ecf9ffde02ae6b02546de6f80317369c10f80dea/generator/internal/sidekick/refresh.go#L54

Into a function that loads the model and config for a directory:

```go
func refreshDir(rootConfig *config.Config, cmdLine *CommandLine, output string) error {
    model, config, err := loadDir(rootConfig, cmdLine, output);
    if err != nil { return err }
	switch config.General.Language {
	case "rust":    
... ...
```

  Then we could do:

```go
      controlModel, controlConfig, err := loadDir(rootConfig, cmdLine, "src/storage/src/generated/gapic_control/gapic");
      storageModel, storageConfig, err := loadDir(rootConfig, cmdLine, "src/storage/src/generated/gapic");

        // Build a custom codec that generates the code we want:
	codec := newCodec(cfg)
	codec.annotateModels(controlModel, storageModel, cfg)
	provider := templatesProvider()
[...]

PR #2751, opened 2025-07-23
impl: service name overrides in doc links

dbolduc:
Part of the work for #2398 

When we start using the generated client docs, the broken link to `Storage::restore_object` becomes public. We need to apply the service name overrides for doc links.

Aside: the name overrides could apply to more link types, (e.g. messages / fields / etc.) but we will add the support as needed.

### DIFF
--- a/generator/internal/rust/codec_test.go
+++ b/generator/internal/rust/codec_test.go
@@ -1237,8 +1237,7 @@ func TestFormatDocCommentsCrossLinks(t *testing.T) {
 [Message][test.v1.SomeMessage]
 [Enum][test.v1.SomeMessage.SomeEnum]
 [Message][test.v1.SomeMessage] repeated
-[Service][test.v1.SomeService]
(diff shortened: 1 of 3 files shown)


generator/internal/rust/codec_test.go
```diff
@@ -1237,40 +1237,48 @@ func TestFormatDocCommentsCrossLinks(t *testing.T) {
 [Message][test.v1.SomeMessage]
 [Enum][test.v1.SomeMessage.SomeEnum]
 [Message][test.v1.SomeMessage] repeated
-[Service][test.v1.SomeService] [field][test.v1.SomeMessage.field]
```
coryan: I think the intent was to verify that more than one replacement per line worked. Maybe restore that and add a comment?

dbolduc: Done.

coryan (approved):

PR #2755, opened 2025-07-23
refactor(storage): avoid `include!` macro

dbolduc:
Avoid using the `include!` macro which gives us formatting, and makes the source code lookup-able in the docs. Thanks @coryan for the suggestion.

image


src/storage/src/control.rs
```diff
@@ -25,3 +25,5 @@ pub mod stub;
 
 mod convert;
 mod status;
+
```
coryan, line 28: nit: extra blank line?

coryan (approved):

src/storage/src/control.rs
dbolduc: without the blank line it bugged me that they were not in alphabetical order.

and if I put them in alphabetical order, it bugged me that the generated mod was sandwiched between handwritten mods. :shrug:

PR #2760, opened 2025-07-24
impl(storage): capture headers for resuming reads

coryan:
To resume a download we need to know:

- What was the object generation, as we do not want to download a new
  object.
- What range of the object are we reading, so we can resume from the
  last received byte.

In this PR we capture this information. Later we will use and update it
as the download makes progress.

Part of the work for #2057

### DIFF
--- a/src/storage/src/storage/read_object.rs
+++ b/src/storage/src/storage/read_object.rs
@@ -551,7 +551,7 @@ enum ReadError {
     BadCrc { got: u32, want: u32 },
 
     /// Only 200 and 206 status codes are expected in successful responses.
-    #[error("unexpected success code {0} in read request, only 200 and 206")]
+    #[error("unexpected success code {0} in read request, only 200 and 206 are expected")]
     UnexpectedSuccessCode(u16),
 
     /// Successful HTTP response must include some headers.
@@ -627,7 +627,7 @@ fn response_range(response: &reqwest::Response) -> std::result::Result<ReadRange
                 ReadError::BadHeaderFormat("content-range", "missing / separator".into())
             })?;
             let (start, end) = range.split_once('-').ok_or_else(|| {
-                ReadError::BadHeaderFormat("content-range", "missing / separator".into())
+                ReadError::BadHeaderFormat("content-range", "missing - separator".into())
             })?;
             let start = start
                 .parse::<u64>()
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/read_object.rs
```diff
@@ -534,12 +543,27 @@ impl ReadObjectResponse {
...
-#[derive(thiserror::Error, Debug, PartialEq)]
+#[derive(thiserror::Error, Debug)]
 #[non_exhaustive]
 enum ReadError {
     /// The calculated crc32c did not match server provided crc32c.
     #[error("bad CRC on read: got {got}, want {want}")]
     BadCrc { got: u32, want: u32 },
+
+    /// Only 200 and 206 status codes are expected in successful responses.
+    #[error("unexpected success code {0} in read request, only 200 and 206")]
```
dbolduc: "only 200 and 206 are expected" ?

src/storage/src/storage/read_object.rs
```diff
@@ -585,6 +609,68 @@ fn check_crc32c_match(crc32c: u32, response: Option<u32>) -> Result<()> {
...
+        reqwest::StatusCode::PARTIAL_CONTENT => {
+            let header = required_header(response, "content-range")?;
+            let header = header.strip_prefix("bytes ").ok_or_else(|| {
+                ReadError::BadHeaderFormat("content-range", "missing bytes prefix".into())
+            })?;
+            let (range, _) = header.split_once('/').ok_or_else(|| {
+                ReadError::BadHeaderFormat("content-range", "missing / separator".into())
+            })?;
+            let (start, end) = range.split_once('-').ok_or_else(|| {
+                ReadError::BadHeaderFormat("content-range", "missing / separator".into())
```
dbolduc: s;missing / separator;missing - separator;

src/storage/src/storage/read_object.rs
```diff
@@ -585,6 +609,68 @@ fn check_crc32c_match(crc32c: u32, response: Option<u32>) -> Result<()> {
...
+                .parse::<u64>()
+                .map_err(|e| ReadError::BadHeaderFormat("content-range", e.into()))?;
+            let end = end
+                .parse::<u64>()
+                .map_err(|e| ReadError::BadHeaderFormat("content-range", e.into()))?;
+            let limit = end
+                .checked_sub(start)
+                .ok_or_else(|| ReadError::BadHeaderFormat("content-range", format!("range start ({start}) should be less than or equal to the range end ({end})").into()))?;
+            // HTTP ranges are inclusive, but it is more idiomatic to use a non-inclusive end point.
+            Ok(ReadRange { start, limit })
```
dbolduc: This comment makes me think you are going to add 1 to `limit`

dbolduc (approved):

src/storage/src/storage/read_object.rs
coryan: Fixed. thanks

src/storage/src/storage/read_object.rs

src/storage/src/storage/read_object.rs
coryan: Yes, fixed.

coryan (commented):
PTAL

dbolduc (approved):

PR #2764, opened 2025-07-25
impl(storage): use `next()` for `all_bytes()`

coryan:
This will simplify the implementation of resuming downloads. And since
we have a good idea of the total size, we can improve allocations.

Part of the work for #2057

### DIFF
--- a/src/storage/src/storage/read_object.rs
+++ b/src/storage/src/storage/read_object.rs
@@ -478,13 +478,12 @@ impl ReadObjectResponse {
     /// println!("object contents={contents:?}");
     /// # Ok::<(), anyhow::Error>(()) });
     /// ```
-    pub async fn all_bytes(self) -> Result<bytes::Bytes> {
-        let bytes = self.inner.bytes().await.map_err(Error::io)?;
-        if self.response_crc32c.is_some() {
-            let crc32c = crc32c::crc32c_append(self.crc32c, &bytes); // bytes may have already been read by `next`.
-            check_crc32c_match(crc32c, self.response_crc32c)?;
+    pub async fn all_bytes(mut self) -> Result<bytes::Bytes> {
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/read_object.rs
```diff
@@ -478,13 +478,12 @@ impl ReadObjectResponse {
...
     /// ```
-    pub async fn all_bytes(self) -> Result<bytes::Bytes> {
-        let bytes = self.inner.bytes().await.map_err(Error::io)?;
-        if self.response_crc32c.is_some() {
-            let crc32c = crc32c::crc32c_append(self.crc32c, &bytes); // bytes may have already been read by `next`.
-            check_crc32c_match(crc32c, self.response_crc32c)?;
+    pub async fn all_bytes(mut self) -> Result<bytes::Bytes> {
+        let mut contents = Vec::with_capacity(self.range.limit as usize);
+        while let Some(b) = self.next().await.transpose()? {
+            contents.extend_from_slice(&b);
```
dbolduc, line 484: This looks like an extra copy, but it may be happening internally on `reqwest::Response::bytes()`.

It seems like there is no good way to move bytes out of a `Bytes` without copying, because the `Bytes` may not be the only owner of the underlying data?

dbolduc (approved):
I have an incomplete understanding of how to use `bytes::Bytes`. Left a comment, which may be wrong.

src/storage/src/storage/read_object.rs
coryan: Yes.  And `bytes::Bytes` requires a contiguous slice of bytes, you cannot use a linked list of buffers.  We *may* want to use something like `https://docs.rs/bytes/latest/bytes/buf/trait.Buf.html`.

Issue #2048, opened 2025-05-08
Automatically resume interrupted downloads


coryan:
Downloads can fail to start, we need a retry policy to restart the download if the connection is not established.

Downloads can fail after they start: we need a different policy to resume such connections.  The resume policy may resume on more errors, and  maybe does not backoff.

To safely resume a download we need to capture the generation of the object, this is available as a header when the object download starts.  Then we should resume using the same generation, to avoid RAW hazards.

PR #2765, opened 2025-07-25
impl(storage): resume interrupted downloads

coryan:
If a download is interrupted or gets too little data, resume the
download from the last received byte. To do so, we need to update the
remaining range, and that creates the possibility of mismatches between
the actual download size and the size implied by `content-length` or
`content-range`.

Part of the work for #2048

### DIFF
--- a/src/storage/src/storage/read_object.rs
+++ b/src/storage/src/storage/read_object.rs
@@ -606,7 +606,7 @@ enum ReadError {
     #[error("missing {0} bytes at the end of the stream")]
     ShortRead(u64),
 
-    #[error("too many bytes received: expected {expected} (at most), got {got}")]
+    #[error("too many bytes received: expected {expected}, stopped download at {got}")]
     LongRead { got: u64, expected: u64 },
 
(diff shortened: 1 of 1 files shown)


amanda-tarafa (approved):

src/storage/src/storage/read_object.rs
```diff
@@ -549,6 +603,12 @@ enum ReadError {
     #[error("bad CRC on read: got {got}, want {want}")]
     BadCrc { got: u32, want: u32 },
 
+    #[error("missing {0} bytes at the end of the stream")]
+    ShortRead(u64),
+
+    #[error("too many bytes received: expected {expected} (at most), got {got}")]
```
suzmue: nit: I would drop "at most" in this error message. We give a ShortRead error if its < expected, so it seems to be vague when it doesn't need to be since we are expecting *exactly* expected AFAICT.

Another nit with this error message is that got may not be all the bytes that would be returned. Since we send this as soon as we go over, not at the end.

suzmue (approved):

src/storage/src/storage/read_object.rs

PR #2766, opened 2025-07-25
impl(storage): introduce resume policies

coryan:
We will need a way to configure what type of error we tolerate for resumes, and
how many errors we tolerate too.

Part of the work for #2048

### DIFF
--- a/src/storage/src/download_resume_policy.rs
+++ b/src/storage/src/download_resume_policy.rs
@@ -1,0 +1,306 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Defines the download resume policies for Google Cloud Storage.
+//!
+//! Even if a download request starts successfully, it may be fail after it
+//! starts. For example, the download may be interrupted or become too slow and
+//! "stall". The client library can automatically recover from such errors. The
+//! application may want to control what errors are treated as recoverable, and
+//! how many failures are tolerated before abandoning the download.
+//!
+//! The traits and types defined in this module allow for such customization.
+//!
+//! # Example
+//! ```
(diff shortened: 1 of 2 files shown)


src/storage/src/download_resume_policy.rs
```diff
@@ -0,0 +1,306 @@
...
+impl ResumeQuery {
+    /// Create a new instance.
+    pub fn new(attempt_count: u32) -> Self {
+        Self { attempt_count }
+    }
+}
+
+/// The recommended policy for storage downloads.
+///
+/// This policy resumes any download that fails due to I/O errors, but
```
amanda-tarafa: nit: Incomplete

src/storage/src/download_resume_policy.rs
```diff
@@ -0,0 +1,306 @@
...
+//! }
+//! ```
+
+use crate::Error;
+
+pub use gax::retry_result::RetryResult as ResumeResult;
+
+/// Defines the interface to resume policies.
+pub trait DownloadResumePolicy: Send + Sync + std::fmt::Debug {
+    fn on_error(&self, status: &ResumeQuery, error: Error) -> ResumeResult;
```
amanda-tarafa, line 47: Wondering why the error is not part of the ResumeQuery and this is called `on_failure`? If this has been discussed already or is idiomatic that's fine.

amanda-tarafa: Also, could the download and upload resume policy types could be the same? Predefined policies could be suffixed with the operation they are meant for, like `DownloadRecommended`, but even some of those may be shared, like `NeverResume`?

amanda-tarafa (approved):

src/storage/src/download_resume_policy.rs
coryan: > Wondering why the error is not part of the ResumeQuery and this is called `on_failure`? If this has been discussed already or is idiomatic that's fine.

I am calling this "idiomatic".  And it follows an idea from the existing retry policies. We want to transfer ownership of `error` to the policy, so it can do with it whatever it needs without copying.  Meanwhile the `status` parameter should be unchanged.

> Also, could the download and upload resume policy types could be the same? Predefined policies could be suffixed with the operation they are meant for, like `DownloadRecommended`, but even some of those may be shared, like `NeverResume`?

We do not have upload resume policies: the retry policies do the job just fine.  If we find we need them, we can introduce aliases to keep backwards compatibility (e.g. sprinkle some `use ResumePolicy as DownloadPolicy;` and so forth).

src/storage/src/download_resume_policy.rs
coryan: Fixed, thanks!

coryan (commented):
PTAL

amanda-tarafa (approved):
Thanks!

PR #2767, opened 2025-07-25
docs(storage): add rewrite object guide

### LINKED ISSUE #2396: Sample for StorageControl RewriteObject
This is a complicated RPC. It needs a sample.

Where this sample lives is an open question.
- Client documentation seems wrong. (1) It will probably be too large to put in the client documentation. (2) We want to have this be fully generated in the longrun
- User guide seems wrong. It is too specific for a user guide. We could have a user guide per crate. But meh. Where do we host them? meh.
- `integration-tests` is good for running the sample, but bad for discoverability. That is our test code.

I think I would make a `src/storage-control/samples` directory and throw it in there. We would have to run the samples as part of the CI. (This is #1948). I think I would move the existing client samples (which have dummy values) in there too.

dbolduc:
Fixes #2396 

[rewriting_objects.pdf](https://github.com/user-attachments/files/21437631/rewriting_objects.pdf)

### DIFF
--- a/guide/src/storage/rewrite_object.md
+++ b/guide/src/storage/rewrite_object.md
@@ -70,7 +70,7 @@ larger value.
 
 Rewriting an object allows you to copy its data to a different bucket, copy its
 data to a different object in the same bucket, change its encryption key, and/or
-change its storage class. The rewrite loop is identical for all these
+change its [storage class]. The rewrite loop is identical for all these
 transformations. We will change the storage class to illustrate the code.
 
 ```rust,ignore,noplayground


guide/src/storage/rewrite_object.md
```diff
@@ -0,0 +1,141 @@
...
+larger value.
+
+```rust,ignore,noplayground
+{{#rustdoc_include ../../samples/tests/storage/rewrite_object.rs:limit-bytes-per-call}}
+```
+
+In this example, we change the object's [storage class]. This is done for
+illustrative purposes, as it forces the service to physically copy all of the
+bytes in the object. Changing the storage class is not necessary to perform a
+successful rewrite object request.
```
coryan: Maybe say something like: "Rewriting an object allows you to make copy its data to a different bucket, a different object in the same bucket, change the object's encryption key, and change the storage class. The rewrite loop is identical for all these transformations. We will change the storage class to illustrate the code".

guide/src/SUMMARY.md
```diff
@@ -23,6 +23,7 @@ limitations under the License.
 - [Generate text using the Vertex AI Gemini API](generate_text_using_the_vertex_ai_gemini_api.md)
 - [Using Google Cloud Storage](storage.md)
   - [Push data on object uploads](storage/queue.md)
+  - [Rewriting objects](storage/rewrite_object.md)
```
coryan, line 26: You may also want to add the new content to the quickstart guide?

guide/samples/tests/storage/rewrite_object.rs
```diff
@@ -0,0 +1,132 @@
...
+use gcs::model::Object;
+use google_cloud_gax::retry_policy::AlwaysRetry;
+use google_cloud_storage as gcs;
+
+pub async fn rewrite_object(bucket_name: &str) -> anyhow::Result<()> {
+    let source_object = upload(bucket_name).await?;
+
+    // ANCHOR: client
+    let control = StorageControl::builder()
+        .with_retry_policy(AlwaysRetry)
```
coryan: I would rather use this: 

https://github.com/googleapis/google-cloud-rust/blob/0ef44316736b14067decea05d7272bf71c37a25c/src/storage/src/retry_policy.rs#L76

Maybe with a limited count to avoid infinite loops?

guide/samples/tests/storage/rewrite_object.rs
```diff
@@ -0,0 +1,132 @@
...
+// ANCHOR_END: make-one-request
+
+// Upload an object to rewrite
+async fn upload(bucket_name: &str) -> anyhow::Result<Object> {
+    let storage = gcs::client::Storage::builder().build().await?;
+    // We need the size to exceed 1MB to exercise the rewrite token logic.
+    let payload = bytes::Bytes::from(vec![65_u8; 3 * 1024 * 1024]);
+    let object = storage
+        .upload_object(bucket_name, "rewrite-object-source", payload)
+        .send()
```
coryan: nit: `.send_unbuffered()` would be more efficient in this case.

guide/samples/tests/storage/rewrite_object.rs
```diff
@@ -0,0 +1,132 @@
...
+    // ANCHOR_END: builder
+
+    // Optionally limit the max bytes written per request.
+    // ANCHOR: limit-bytes-per-call
+    builder = builder.set_max_bytes_rewritten_per_call(1024 * 1024);
+    // ANCHOR_END: limit-bytes-per-call
+
+    // Optionally change the storage class to force GCS to copy bytes
+    // ANCHOR: change-storage-class
+    builder = builder.set_destination(Object::new().set_storage_class("NEARLINE"));
```
coryan, line 50: This example will cost 30 days of storage:

https://cloud.google.com/storage/docs/storage-classes#classes
https://cloud.google.com/storage/pricing-examples#early-delete

It is way below $0.01[^1], but we should warn people "If you run this example with larger data consider the costs [link]".

[^1]: https://cloud.google.com/products/calculator?hl=en&dl=CjhDaVJtWW1NNU5XTTRZaTFrTVRnNExUUXdObVl0WVRNME1DMDROelUzWmpBelpqRXlOemNRQVE9PRAJGiRFNDc2MTMyMy04NzBBLTRGRUYtOTkyOS05OUFDNTE2QjlEMEQ

guide/samples/tests/storage/rewrite_object.rs
```diff
@@ -0,0 +1,132 @@
...
+        let progress = make_one_request(builder.clone()).await?;
+        match progress {
+            // ANCHOR: set-rewrite-token
+            RewriteProgress::Incomplete(rewrite_token) => {
+                builder = builder.set_rewrite_token(rewrite_token);
+            }
+            // ANCHOR_END: set-rewrite-token
+            RewriteProgress::Done(o) => return Ok(o),
+        };
+    }
```
coryan: I would unroll this loop into the main function.

coryan (commented):
drive-by

guide/src/SUMMARY.md
coryan: As a "next step"

guide/samples/tests/storage/rewrite_object.rs
dbolduc: Done.

guide/src/storage/rewrite_object.md
dbolduc: Done

guide/src/SUMMARY.md
dbolduc: Done.

Aside: I see that I will race with some of your outstanding PRs. We will figure it out.

guide/samples/tests/storage/rewrite_object.rs
dbolduc: Fixed, thanks.

guide/samples/tests/storage/rewrite_object.rs
dbolduc: Added a warning.

> This example will cost 30 days of storage:

Yeah, I saw...

> It is way below $0.01

and this is why I wasn't deterred.

TIL: shareable pricing calculator. neat.

guide/samples/tests/storage/rewrite_object.rs
dbolduc: Done, thanks. Forgot that Rust is cool and lets me return values via `break`.

coryan (approved):

PR #2768, opened 2025-07-26
docs(user-guide): improve enum examples

coryan:
Avoid `unreachable!()` in code that may be reached, it is bad guidance,
even if it was not the main point of the example. Prefer returning an
error. Document cases where `unreachable!()` may be appropriate.

I also made some cosmetic changes to keep the examples from rendering
with a horizontal scroll bar.

Motivated by https://www.reddit.com/r/rust/comments/1luqn9l/comment/n25a19z

### DIFF
--- a/guide/samples/tests/enums.rs
+++ b/guide/samples/tests/enums.rs
@@ -51,8 +51,8 @@ mod tests {
         assert_eq!(state.value(), None);
         assert_eq!(state.name(), Some("STATE_NAME_FROM_THE_FUTURE"));
         println!("json = {}", serde_json::to_value(&state)?);
-        let s2 = serde_json::from_value::<State>(json!("STATE_NAME_FROM_THE_FUTURE"))?;
-        assert_eq!(state, s2);
+        let u = serde_json::from_value::<State>(json!("STATE_NAME_FROM_THE_FUTURE"))?;
+        assert_eq!(state, u);
         // ANCHOR_END: unknown_string
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


guide/samples/tests/enums.rs
```diff
@@ -79,41 +79,53 @@ mod tests {
...
-            State::Unspecified => unreachable!("the service documents this value as unused"),
+            State::Unspecified => {
+                return Err(Error::msg("the documentation says this is never used"));
+            }
             State::Enabled => println!("the secret is enabled and can be accessed"),
-            State::Disabled => println!(
-                "the secret version may not be accessed, but can be enabled and then accessed"
-            ),
+            State::Disabled => {
+                println!("the secret version it not accessible until it is enabled")
```
dbolduc: nit: s/version it not/version is not/

guide/samples/tests/enums.rs
```diff
@@ -79,41 +79,53 @@ mod tests {
...
```
dbolduc: same

guide/samples/tests/enums.rs
```diff
@@ -122,7 +134,7 @@ mod tests {
     #[test_case(State::Destroyed)]
     #[test_case(State::from("UNKNOWN"))]
     fn drive_match_expression(state: State) {
-        match_with_warnings(state.clone());
-        match_with_wildcard(state);
+        match_with_warnings(state.clone()).expect("example includes all branches");
+        match_with_wildcard(state).expect("example include all branches");
```
dbolduc: s/include/includes/

dbolduc (dismissed):

guide/samples/tests/enums.rs
coryan: Fixed.

guide/samples/tests/enums.rs
coryan: Fixed

coryan: Fixed.

coryan (commented):
PTAL

dbolduc (approved):

PR #2770, opened 2025-07-26
docs(storage): user guide for striped downloads

coryan:


guide/src/storage/striped_downloads.md
```diff
@@ -0,0 +1,168 @@
...
+## Prerequisites
+
+The guide assumes you have an existing [Google Cloud project] with
+[billing enabled], and a Cloud Storage bucket in that project.
+
+You will create some large objects during this tutorial, remember to clean up
+any resources to avoid excessive billing.
+
+The tutorial assumes you are familiar with the basics of using the client
+library, read the [quickstart guide]
```
dbolduc: s/, read/. If not, read/

s/$/.$/

dbolduc: Can also add this to quickstart next steps section. :)

guide/src/storage/striped_downloads.md
```diff
@@ -0,0 +1,168 @@
...
+## Add the client library as a dependency
+
+```toml
+{{#include ../../samples/Cargo.toml:storage}}
+```
+
+## Create source data
+
+To run this tutorial you will need some large objects in Cloud Storage. You can
+create such objects by seeding a smaller object and then repeatedly compose it
```
dbolduc: s/compose/composing/

guide/src/storage/striped_downloads.md
```diff
@@ -0,0 +1,168 @@
...
+
+Use the storage control client to query the object metadata. This metadata
+includes the object size and the current generation:
+
+```rust,ignore,noplayground
+{{#rustdoc_include ../../samples/tests/storage/striped.rs:get-metadata}}
+```
+
+We will split the download of each stripe to a separate function. You will see
+the details of this function in a moment, for now just note that is is `async`,
```
dbolduc: s/is is/is/

guide/samples/tests/storage/striped.rs
```diff
@@ -0,0 +1,284 @@
...
+            .set_destination(Object::new().set_bucket(bucket_name).set_name(&name))
+            .set_source_objects((0..s).map(|_| {
+                SourceObject::new()
+                    .set_name(&seed_1024.name)
+                    .set_generation(seed_1024.generation)
+            }))
+            .send()
+            .await?;
+        println!(
+            "Created object {} size={}",
```
dbolduc: size units?

guide/samples/tests/storage/striped.rs
```diff
@@ -0,0 +1,284 @@
...
+
+    // ANCHOR: create-destination
+    let file = tokio::fs::File::create(destination).await?;
+    // ANCHOR_END: create-destination
+    let start = std::time::Instant::now();
+
+    // ANCHOR: compute-stripes
+    let limit = stripe_size as i64;
+    let count = metadata.size / limit;
+    let stripes = (0..count)
```
dbolduc: do we want to handle the case where `metadata.size % limit != 0` ? We would drop the end of the object, right?

guide/src/storage/striped_downloads.md
```diff
@@ -0,0 +1,168 @@
...
+```rust,ignore,noplayground
+{{#rustdoc_include ../../samples/tests/storage/striped.rs:write-stripe-loop}}
+```
+
+## Full program
+
+```rust,ignore,noplayground
+{{#rustdoc_include ../../samples/tests/storage/striped.rs:all}}
+```
+
```
dbolduc, line 185: optional: add example output

guide/src/storage/striped_downloads.md
```diff
@@ -0,0 +1,168 @@
...
+We will split the download of each stripe to a separate function. You will see
+the details of this function in a moment, for now just note that is is `async`,
+so it returns a `Future`:
+
+```rust,ignore,noplayground
+{{#rustdoc_include ../../samples/tests/storage/striped.rs:write-stripe-function}}
+    // ... details omitted ...
+{{#rustdoc_include ../../samples/tests/storage/striped.rs:write-stripe-function-end}}
```
dbolduc, line 106: need to add this anchor to the sample.

Also, thanks for adding the PDF. It is 138 pages, though. Maybe just print the relevant section next time?

guide/samples/tests/storage/striped.rs
coryan: Fixed.

guide/samples/tests/storage/striped.rs
coryan: Good catch, fixed.

guide/src/storage/striped_downloads.md
coryan: Fixed.

guide/src/storage/striped_downloads.md
coryan: Fixed.

guide/src/storage/striped_downloads.md
coryan: Fixed.

guide/src/storage/striped_downloads.md
coryan: Fixed.

guide/src/storage/striped_downloads.md
coryan: I include a section with all the caveats, or at least all the caveats I can think of.  It is only too easy to make this program run really slowly.

guide/src/storage/striped_downloads.md
```diff
@@ -160,6 +160,23 @@ Then you read the data and write it to the local file:
 
 - Consider optimizing the case where the last stripe only has a few bytes
 
+## Expected performance
+
+The performance of these downloads depends on:
+
+- The I/O subsystem: if your local storage is not fast enough the downloads will
+  be throttled by the writes to disk.
+- The configuration of your VM: if you do not have enough CPUs the downloads
```
dbolduc: nit: s/enough CPUs/enough CPUs,/

guide/src/storage/striped_downloads.md
```diff
@@ -160,6 +160,23 @@ Then you read the data and write it to the local file:
...
+  will throttled trying to decrypt the on data, as Cloud Storage and the client
```
dbolduc: nit: "will be throttled on trying to decrypt the data"

guide/src/storage/striped_downloads.md
```diff
@@ -97,12 +97,12 @@ includes the object size and the current generation:
 ```
 
 We will split the download of each stripe to a separate function. You will see
-the details of this function in a moment, for now just note that is is `async`,
-so it returns a `Future`:
+the details of this function in a moment, for now just note that is `async`, so
```
dbolduc: ugh, mislead you, sorry.

We want: "that it is `async`"

guide/src/storage/striped_downloads.md
```diff
@@ -156,6 +156,10 @@ Then you read the data and write it to the local file:
 {{#rustdoc_include ../../samples/tests/storage/striped.rs:write-stripe-loop}}
 ```
 
+## Next steps
+
+- Consider optimizing the case where the last stripe only has a few bytes
```
dbolduc, line 161: comment: if we think in terms of number of stripes, the special case goes away.

We could change the input to be `max_stripe_size`, and then balance the stripes with:

```rs
let stripe_count = metadata.size.div_ceil(max_stripe_size);
let stripe_size = metadata.size / stripe_count;
```

---

I don't know if there are advantages to having "nicely sized stripes", though. (e.g. stripe sizes that are 2^n for some n.)

dbolduc (dismissed):

guide/src/storage/striped_downloads.md
coryan: Fixed.

guide/src/storage/striped_downloads.md
coryan: Fixed.

guide/src/storage/striped_downloads.md
coryan: Depending on what you mean, that approach may miss bytes if `metadata.size % stripe_count != 0`:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=aa6349af995d3acac86181a1d5b66f9f

dbolduc: You are right that I didn't handle the remainder. The fix is to round up the stripe size:

```rs
    let stripe_count = size.div_ceil(max_stripe_size);
    let stripe_size = size.div_ceil(stripe_count);
```

> https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=aa6349af995d3acac86181a1d5b66f9f

I disagree that what we want to do is increase the stripe size to 35, when given a max size of 32.

The above snippet yields: `[(0, 26), (26, 26), (52, 26), (78, 26)]`

coryan: That fails with a slightly larger size:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=92516c08e8944df7daa79908176bcc18

Maybe I have convinced you that a simple rule is not easy?

I think we want to define a minimum stripe size `m`, and find a split that minimizes the number of stripes, while keeping the stripe size smaller than `2 * m`.

dbolduc: > Maybe I have convinced you

No you have not convinced me, because you did not implement the change.

image

---

I think the math checks out.

(1) `stripe_size = size.div_ceil(stripe_count)` => `stripe_size * stripe_count >= size`, so we are covering the full object.

(2) And we know that `max_stripe_size >= stripe_size`, so we are respecting the max size.

---

I nerdsniped myself proving (2)...

Let's say `size + q = max_stripe_size * stripe_count` for some `0 <= q < max_stripe_size`.

Then if `size + r = stripe_count * stripe_size` for some `0 <= r < stripe_count`, we know that `r <= q` (because `q` would work, setting `stripe_size == max_stripe_size`).

=> `q-r = stripe_count * (max_stripe_size - stripe_size)`

=> `max_stripe_size >= stripe_size`.

Issue #2050, opened 2025-05-08
Verify checksums during uploads


coryan:
When possible, we should send the checksum when the upload starts, so the service can verify it at the end.

There is no way to compute and send the checksum at the end of an HTTP+JSON upload, but there is a way to do this over gRPC.

PR #2775, opened 2025-07-28
impl(storage): checksum abstractions

coryan:
Create some abstractions to compute CRC32C checksums and MD5 hashes.

Part of the work for #2050

### DIFF
--- a/src/storage/src/storage/checksum.rs
+++ b/src/storage/src/storage/checksum.rs
@@ -1,0 +1,536 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::model::ObjectChecksums;
+
+/// The error type for checksum comparisons.
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+pub enum ChecksumMismatch {
+    #[error("mismatched CRC32C values {0}")]
+    Crc32c(String),
+    #[error("mismatched MD5 values: {0}")]
+    MD5(String),
+    #[error("mismatched CRC32C and MD5 values {0}")]
+    Both(String),
+}
+
+/// Compare the received object checksums vs. the computed checksums.
+///
+/// If the `crc32c` field is `None`, or the `md5_mash` field is empty, they do
+/// not participate in the comparison. That accounts for disabled checksums in
+/// the client (where only CRC32C is enabled by default) and for missing MD5
+/// hashes on the server (common with composed objects).
+pub fn validate(
+    expected: ObjectChecksums,
+    received: &Option<ObjectChecksums>,
+) -> Result<(), ChecksumMismatch> {
+    let Some(recv) = received else {
+        return Ok(());
+    };
+    let crc32c = match (&expected.crc32c, &recv.crc32c) {
+        (None, _) => None,
+        (Some(_), None) => None,
+        (Some(e), Some(r)) if e == r => None,
+        (Some(e), Some(r)) => Some(format!("{e} != {r}")),
+    };
+    let md5 = match (&expected.md5_hash, &recv.md5_hash) {
+        (h, _) if h.is_empty() => None,
+        (_, r) if r.is_empty() => None,
+        (h, r) if h == r => None,
+        (h, r) => Some(format!("{h:?} !=- {r:?}")),
+    };
+    match (crc32c, md5) {
+        (None, None) => Ok(()),
+        (Some(m), None) => Err(ChecksumMismatch::Crc32c(m)),
+        (None, Some(m)) => Err(ChecksumMismatch::MD5(m)),
+        (Some(m), Some(n)) => Err(ChecksumMismatch::Both(format!("{m} AND {n}"))),
+    }
+}
+
+/// Computes a checksum or hash for [Cloud Storage] transfers.
+///
+/// We want to minimize code complexity in our implementation of data integrity
+/// checks for uploads and downloads. This trait defines a composable interface
+/// to support:
+/// - No checksums (`Null`): the client library does not compute any checksums,
+///   and therefore does not validate checksums either.
+/// - Precomputed checksums (`Precompted`): the client library assumes the
+///   application provided checksums in the object metadata.
+/// - Only crc32c (`Crc32c` or `Crc32c<Null>`)`: compute (and validate) only
+///   crc32c checksums.
+/// - Only MD5 (`Md5` or `Md5<Null>`): compute (and validate) only MD5 hashes.
+/// - Both: (`Crc32c<Md5>` or `Md5<Crc32>`): compute (and validate) both crc32
+///   checksums and MD5 hashes.
+///
+/// The application should have no need to interact with these types directly,
+/// or even name them. They are used only as implementation details. They may
+/// be visible in debug messages.
+pub trait ChecksumEngine: std::fmt::Debug + sealed::Sealed {
+    fn update(&mut self, offset: u64, data: &bytes::Bytes);
+    fn finalize(&self) -> ObjectChecksums;
+}
+
+mod sealed {
+    pub trait Sealed {}
+}
+
+/// YOLO checksum engine.
+#[derive(Clone, Debug)]
+pub(crate) struct Null;
+impl sealed::Sealed for Null {}
+impl ChecksumEngine for Null {
+    fn update(&mut self, _offset: u64, _data: &bytes::Bytes) {}
+    fn finalize(&self) -> ObjectChecksums {
+        ObjectChecksums::new()
+    }
+}
+
+/// Assumes the checksums are provided as part of the object metadata.
+#[derive(Clone, Debug)]
+pub(crate) struct Precomputed;
+impl sealed::Sealed for Precomputed {}
+impl ChecksumEngine for Precomputed {
+    fn update(&mut self, _offset: u64, _data: &bytes::Bytes) {}
+    fn finalize(&self) -> ObjectChecksums {
+        ObjectChecksums::new()
+    }
+}
+
+/// Automatically computes the CRC32C checksum.
+#[derive(Clone, Debug)]
+pub(crate) struct Crc32c<C = Null> {
+    checksum: u32,
+    offset: u64,
+    inner: C,
+}
+impl<C> sealed::Sealed for Crc32c<C> {}
+impl<C> Crc32c<C> {
+    pub fn from_inner(inner: C) -> Self {
+        Self {
+            checksum: 0,
+            offset: 0,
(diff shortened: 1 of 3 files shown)


src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,536 @@
...
+}
+
+/// Computes a checksum or hash for [Cloud Storage] transfers.
+///
+/// We want to minimize code complexity in our implementation of data integrity
+/// checks for uploads and downloads. This trait defines a composable interface
+/// to support:
+/// - No checksums (`Null`): the client library does not compute any checksums,
+///   and therefore does not validate checksums either.
+/// - Precomputed checksums (`Precompted`): the client library assumes the
```
dbolduc: s/Precompted/Precomputed/

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::model::ObjectChecksums;
+
+/// The error type for checksum comparisons.
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+pub enum ChecksumMismatch {
```
dbolduc, line 20: comment: I assume the plan is to export the error type when the client library starts returning it.

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+    received: &Option<ObjectChecksums>,
+) -> Result<(), ChecksumMismatch> {
+    let Some(recv) = received else {
+        return Ok(());
+    };
+    let crc32c = match (&expected.crc32c, &recv.crc32c) {
+        (None, _) => None,
+        (Some(_), None) => None,
+        (Some(e), Some(r)) if e == r => None,
+        (Some(e), Some(r)) => Some(format!("{e} != {r}")),
```
dbolduc: consider:

```suggestion
        (Some(e), Some(r)) if e != r => Some(format!("{e} != {r}")),
        _ => None,
```

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+    };
+    let md5 = match (&expected.md5_hash, &recv.md5_hash) {
+        (h, _) if h.is_empty() => None,
+        (_, r) if r.is_empty() => None,
+        (h, r) if h == r => None,
+        (h, r) => Some(format!("{h:?} !=- {r:?}")),
```
dbolduc: same, and check the extra '-' in the message

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+/// The application should have no need to interact with these types directly,
+/// or even name them. They are used only as implementation details. They may
+/// be visible in debug messages.
+pub trait ChecksumEngine: std::fmt::Debug + sealed::Sealed {
+    fn update(&mut self, offset: u64, data: &bytes::Bytes);
+    fn finalize(&self) -> ObjectChecksums;
+}
+
+mod sealed {
+    pub trait Sealed {}
```
dbolduc: nit: I'd call this `sealed::CheckSumEngine`. Gives a little bit more context.

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+/// YOLO checksum engine.
```
dbolduc, line 85: live fast, die young

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+            .finish()
+    }
+}
+
+fn checked_update<F>(current: u64, offset: u64, data: &bytes::Bytes, updater: F) -> u64
+where
+    F: FnOnce(&bytes::Bytes),
+{
+    match (current, offset, offset + data.len() as u64) {
+        (o, s, end) if o == s => {
```
dbolduc: what do these letters mean?

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+            updater(data);
+            end
+        }
+        (o, s, end) if o > s && o < end => {
+            let data = data.clone().split_off((o - s) as usize);
+            updater(&data);
+            end
+        }
+        (o, s, e) if o > s && o >= e => current,
```
dbolduc: nit: `o >= e` implies that `o >= s`

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+        (_, _, _) => current,
```
dbolduc: can `current < offset` ? That would be like a gap in the checksum.

src/storage/src/storage/checksum.rs
dbolduc: It is like `(known_offset, update_start, update_end)`. I had to reverse engineer the context.

Hm, consider like `(c, start, end)`. The `c` we will know is `current` from context.

src/storage/src/storage/checksum.rs
dbolduc: Like what if we `seek()` forward on a `ChecksummedSource`?

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+}
+
+impl<C, S> StreamingSource for ChecksummedSource<C, S>
+where
+    C: ChecksumEngine + Send + Sync,
+    S: StreamingSource + Send + Sync,
+{
+    type Error = S::Error;
+    async fn next(&mut self) -> Option<Result<bytes::Bytes, Self::Error>> {
+        match self.source.next().await {
```
dbolduc, line 251: nice composition.

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+    }
+
+    fn finalize(&self) -> ObjectChecksums {
+        self.inner.finalize().set_crc32c(self.checksum)
+    }
+}
+
+/// Automatically computes the MD5 checksum.
+#[derive(Clone)]
+pub(crate) struct Md5<C = Null> {
```
dbolduc, line 149: TIL: default type parameters.

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+    #[test_case(md5_only(), md5_only().set_md5_hash(bytes::Bytes::from_static(b"cde")))]
+    #[test_case(both(), md5_only().set_md5_hash(bytes::Bytes::from_static(b"cde")))]
+    fn validate_bad_md5(expected: ObjectChecksums, received: ObjectChecksums) {
+        let err = super::validate(expected.clone(), &Some(received.clone()))
+            .expect_err("values should not match");
+        assert!(matches!(&err, &ChecksumMismatch::MD5 { .. }), "{err:?}");
+    }
+
+    #[test_case(both(), both().set_crc32c(0_u32).set_md5_hash(bytes::Bytes::from_static(b"cde")))]
+    fn validate_both_bad_md5(expected: ObjectChecksums, received: ObjectChecksums) {
```
dbolduc: nit: s/_md5//

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
+        let md5_want = bytes::Bytes::from_owner(Vec::from_iter(digest.0));
+        let crc32_want = crc32c::crc32c(&input);
+
+        engine.update(0, &input.slice(0..4));
+        engine.update(0, &input.slice(0..4));
+        engine.update(4, &input.slice(4..8));
+        engine.update(6, &input.slice(6..12));
+        engine.update(0, &input.slice(0..4));
+        engine.update(8, &input.slice(8..));
+        engine.update(100, &input.slice(0..));
```
dbolduc, line 437: comment: this is kinda wacky. I think you are testing an update past the size of the data. Maybe remove or explain what this is for?

src/storage/src/storage/checksum.rs
```diff
@@ -0,0 +1,539 @@
...
```
dbolduc, line 461: same.

src/storage/src/storage/checksum.rs
coryan: Yes, though the library will never return one of these directly, only as a `source()` for a `crate::Error`.

src/storage/src/storage/checksum.rs
coryan: Thanks.

src/storage/src/storage/checksum.rs
coryan: Hmm, done, but I am not sure that works as well for this branch.

src/storage/src/storage/checksum.rs
coryan: Fixed.

src/storage/src/storage/checksum.rs
coryan: Hmmm, done.

src/storage/src/storage/checksum.rs
coryan: Changed all the code to use an interval, I think it is clearer now.

src/storage/src/storage/checksum.rs
coryan: Fixed.

src/storage/src/storage/checksum.rs
coryan: Done.

src/storage/src/storage/checksum.rs
coryan: Yeah. Added a comment, the data is ignored. The alternative is to `panic!()` because the caller is doing something weird.

coryan (commented):
Maybe we should reconsider the case of `seek()` past the current offset and just `panic!()` or return no checksums.

src/storage/src/storage/checksum.rs
```diff
@@ -40,16 +40,12 @@ pub fn validate(
...
-        (Some(e), Some(r)) => Some(format!("{e} != {r}")),
+        (Some(e), Some(r)) if e != r => Some(format!("{e} != {r}")),
+        _ => None,
     };
     let md5 = match (&expected.md5_hash, &recv.md5_hash) {
-        (h, _) if h.is_empty() => None,
-        (_, r) if r.is_empty() => None,
-        (h, r) if h == r => None,
-        (h, r) => Some(format!("{h:?} !=- {r:?}")),
+        (e, r) if e != r && !e.is_empty() && !r.is_empty() => Some(format!("{e:?} != {r:?}")),
```
dbolduc: Ah I see. Maybe this for shorter lines:

```rs
    (e, r) if e.is_empty() || r.is_empty() || e == r => None,
    (e, r) => Some(format!("{e:?} != {r:?}")),
```

dbolduc (approved):

src/storage/src/storage/checksum.rs
dbolduc: Much nicer, thanks.

src/storage/src/storage/checksum.rs
dbolduc: (optional)

coryan: Also a little easier to read without all the `!` thingies.

coryan: I like it better. Fewer `!` thingies.

dbolduc (approved):

PR #2778, opened 2025-07-28
feat(storage): return bytes::Buf from all_bytes function

suzmue:
This avoids copying the Bytes out of the buffer.

WIP: want to add unit tests for AllBytesBuf.


suzmue:
We have decided to remove all_bytes for now instead.

PR #2780, opened 2025-07-28
test(generator): simple routing combos

dbolduc:
I wrote these tests while debugging something. Might as well add them to the codebase.

### DIFF
--- a/generator/internal/api/model_test.go
+++ b/generator/internal/api/model_test.go
@@ -15,11 +15,88 @@
 package api
 
 import (
-	"github.com/google/go-cmp/cmp"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
 )
 
-func Test_RoutingCombos(t *testing.T) {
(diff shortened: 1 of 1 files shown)


generator/internal/api/model_test.go
```diff
@@ -15,11 +15,88 @@
...
 
 import (
-	"github.com/google/go-cmp/cmp"
 	"testing"
+
+	"github.com/google/go-cmp/cmp"
 )
 
-func Test_RoutingCombos(t *testing.T) {
+func Test_RoutingCombosSimpleOr(t *testing.T) {
```
suzmue: nit: we (mostly) use the TestXXX naming convention throughout this repo, so I think these should match.

(gotemplate_test.go and http_rule_parser_test.go have the other two tests in our repo named this way)

dbolduc: Fixed, thanks, and sent #2783 for the other ones.

coryan (approved):

PR #2784, opened 2025-07-29
docs: user guide for updating a resource

pcoet:
I think this makes sense, but ... still learning Rust. :-) Just let me know what needs fixing!

BTW it looks like the user guide has a new storage guide, but I don't see it in prod. Does the site need to be deployed manually?

### DIFF
--- a/guide/samples/src/update_resource.rs
+++ b/guide/samples/src/update_resource.rs
@@ -19,10 +19,10 @@ use google_cloud_wkt;
 pub async fn update_field(project_id: &str) -> anyhow::Result<()> {
     use google_cloud_secretmanager_v1::client::SecretManagerService;
     use std::collections::HashMap;
-    
+
     // ANCHOR: create
     let client = SecretManagerService::builder().build().await?;
-    
+
     let secret = client
         .create_secret()
         .set_parent(format!("projects/{project_id}"))
@@ -34,13 +34,13 @@ pub async fn update_field(project_id: &str) -> anyhow::Result<()> {
         .await?;
     println!("CREATE = {secret:?}");
     // ANCHOR_END: create
-    
+
     // ANCHOR: update
     let tag = |mut labels: HashMap<_, _>, msg: &str| {
         labels.insert("updated".to_string(), msg.to_string());
         labels
     };
-    
+
     let update = client
         .update_secret()
         .set_secret(
@@ -51,13 +51,15 @@ pub async fn update_field(project_id: &str) -> anyhow::Result<()> {
                 .set_annotations(tag(secret.annotations, "your-annotations")),
         )
         // ANCHOR: set-update-mask
-        .set_update_mask(google_cloud_wkt::FieldMask::default().set_paths(["annotations", "labels"]))
+        .set_update_mask(
+            google_cloud_wkt::FieldMask::default().set_paths(["annotations", "labels"]),
+        )
         // ANCHOR_END: set-update-mask
         .send()
         .await?;
     println!("UPDATE = {update:?}");
     // ANCHOR_END: update
-    
+
     Ok(())
 }
 // ANCHOR_END: update-field


dbolduc:
> BTW it looks like the user guide has a new storage guide, but I don't see it in prod. Does the site need to be deployed manually?

We only deploy the user guide when we do a release of the repo. We do that so the user guide stays in sync with our released crates (vs. staying in sync with HEAD).

guide/samples/src/update_resource.rs
```diff
@@ -0,0 +1,64 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use google_cloud_secretmanager_v1::model;
+use google_cloud_wkt;
+
+// ANCHOR: update-field
+pub type Result = std::result::Result<(), Box<dyn std::error::Error>>;
```
coryan: I would use `anyhow::Result<()>` here.  It is an extra dependency, but a well-known one, and it reduces the syntactic noise.

guide/samples/src/update_resource.rs
```diff
@@ -0,0 +1,64 @@
...
+        .send()
+        .await?;
+    println!("CREATE = {secret:?}");
+    // ANCHOR_END: create
+    
+    // ANCHOR: update
+    let tag = |mut map: std::collections::HashMap<String, String>, msg: &str| {
+        map.insert("updated".to_string(), msg.to_string());
+        map
+    };
```
coryan: I would let Rust do more of the type deduction, and add a use declaration for `HashMap`:

```suggestion
    let tag = |mut labels: HashMap<_, _>, msg: &str| {
        labels.insert("updated".to_string(), msg.to_string());
        labels
    };
```

guide/samples/src/update_resource.rs
```diff
@@ -0,0 +1,64 @@
...
+    
+    let update = client
+        .update_secret()
+        .set_secret(
+            model::Secret::new()
+                .set_name(&secret.name)
+                .set_etag(secret.etag)
+                .set_labels(tag(secret.labels.clone(), "your-label"))
+                .set_annotations(tag(secret.annotations.clone(), "your-annotations")),
```
coryan: We do not need to clone these if we are not going to use them again?

```suggestion
                .set_labels(tag(secret.labels, "your-label"))
                .set_annotations(tag(secret.annotations, "your-annotations")),
```

guide/src/update_resource.md
```diff
@@ -0,0 +1,123 @@
...
+```
+
+If you examine the output from the create operation, you'll see that both the
+`labels` and `annotations` fields are empty.
+
+The following code updates the `labels` and `annotations` fields:
+
+```rust,ignore
+{{#include ../samples/src/update_resource.rs:update}}
+```
```
coryan, line 80: Maybe say something about how setting `etag` prevents overwriting any concurrent updates?

guide/src/update_resource.md
```diff
@@ -0,0 +1,123 @@
...
+
+To complete this tutorial, you need a Rust development environment with the
+following dependencies installed:
+
+* The Secret Manager client library
+* [Tokio]
+
+To get set up, follow the steps in [Setting up your development environment].
+
+## Install well known types
```
coryan, line 34: Nice background sections!  We should write more of those.

coryan:
> > BTW it looks like the user guide has a new storage guide, but I don't see it in prod. Does the site need to be deployed manually?
> 
> We only deploy the user guide when we do a release of the repo. We do that so the user guide stays in sync with our released crates (vs. staying in sync with HEAD).

We should consider a second deployment to `github.io/preview` or something...

guide/src/update_resource.md
pcoet: ACK!

guide/src/update_resource.md
pcoet: Done.

guide/samples/src/update_resource.rs
pcoet: Done.

guide/samples/src/update_resource.rs
pcoet: Fixed.

guide/samples/src/update_resource.rs
pcoet: Fixed.

coryan:
Looks like `cargo fmt -p user-guide-samples` would fix the code formatting .

PR #2785, opened 2025-07-29
impl(storage): add checksums to upload builder

coryan:
Change `UploadObject` to hold the checksum engine. By default the client
enables CRC32C. Applications can add MD5, or remove all checksums. The
application can also provide pre-computed checksums, in which case the
engines are disabled (and cannot be enabled again).

Finally, if the payload implements `Seek`, the application can ask the
client library to make a pass over the data and pre-compute the
checksums, so they can be included at the start of the upload.

The code to compare the computed checksums against the values returned
by the service will happen in a future PR.

Part of the work for #2050

### DIFF
--- a/src/storage/src/storage/upload_object.rs
+++ b/src/storage/src/storage/upload_object.rs
@@ -536,7 +536,7 @@ impl<T, C> UploadObject<T, C> {
     /// use crc32c::crc32c;
     /// let response = client
     ///     .upload_object("projects/_/buckets/my-bucket", "my-object", "hello world")
-    ///     .with_crc32c(crc32c(b"hello world"))
+    ///     .with_known_crc32c(crc32c(b"hello world"))
     ///     .send()
     ///     .await?;
     /// println!("response details={response:?}");
@@ -576,7 +576,7 @@ impl<T, C> UploadObject<T, C> {
     /// let hash = md5::compute(b"hello world");
     /// let response = client
     ///     .upload_object("projects/_/buckets/my-bucket", "my-object", "hello world")
-    ///     .with_md5_hash(bytes::Bytes::from_owner(hash.0))
+    ///     .with_known_md5_hash(bytes::Bytes::from_owner(hash.0))
     ///     .send()
     ///     .await?;
     /// println!("response details={response:?}");
@@ -597,7 +597,7 @@ impl<T, C> UploadObject<T, C> {
     ///
     /// [compute_crc32c()]: UploadObject::compute_crc32c
     /// [compute_md5()]: UploadObject::compute_md5
-    pub fn with_md5_hash<I, V>(mut self, i: I) -> UploadObject<T, Precomputed>
+    pub fn with_known_md5_hash<I, V>(mut self, i: I) -> UploadObject<T, Precomputed>
     where
         I: IntoIterator<Item = V>,
         V: Into<u8>,
@@ -801,7 +801,7 @@ impl<T> UploadObject<T, Null> {
     /// let response = client
     ///     .upload_object("projects/_/buckets/my-bucket", "my-object", payload)
     ///     .disable_computed_checksums() // override any defaults
-    ///     .with_computed_crc32c()
+    ///     .compute_crc32c()
     ///     .send()
     ///     .await?;
     /// println!("response details={response:?}");
@@ -811,7 +811,7 @@ impl<T> UploadObject<T, Null> {
     /// See [precompute_checksums][UploadObject::precompute_checksums] for more
     /// details on how checksums are used by the client library and their
     /// limitations.
-    pub fn with_computed_crc32c(self) -> UploadObject<T, Crc32c> {
+    pub fn compute_crc32c(self) -> UploadObject<T, Crc32c> {
         self.switch_checksum(Crc32c::from_inner)
     }
 
@@ -824,7 +824,7 @@ impl<T> UploadObject<T, Null> {
     /// let payload = tokio::fs::File::open("my-data").await?;
     /// let response = client
     ///     .upload_object("projects/_/buckets/my-bucket", "my-object", payload)
-    ///     .with_computed_md5()
+    ///     .compute_md5()
     ///     .send()
     ///     .await?;
     /// println!("response details={response:?}");
@@ -849,7 +849,7 @@ impl<T, C> UploadObject<T, Crc32c<C>> {
     /// let payload = tokio::fs::File::open("my-data").await?;
     /// let response = client
     ///     .upload_object("projects/_/buckets/my-bucket", "my-object", payload)
-    ///     .with_computed_md5()
+    ///     .compute_md5()
     ///     .send()
     ///     .await?;
     /// println!("response details={response:?}");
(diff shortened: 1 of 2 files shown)


src/storage/src/storage/upload_object.rs
```diff
@@ -525,30 +527,76 @@ impl<T> UploadObject<T> {
...
         I: IntoIterator<Item = V>,
         V: Into<u8>,
     {
-        let mut checksum = self.mut_resource().checksums.take().unwrap_or_default();
+        let checksum = self.mut_resource().checksums.get_or_insert_default();
         checksum.md5_hash = i.into_iter().map(|v| v.into()).collect();
-        // TODO(#2050) - should we return an error (or panic?) if the size is wrong?
-        self.mut_resource().checksums = Some(checksum);
-        self
+        self.switch_checksum(|_| Precomputed)
```
dbolduc, line 607: maybe I am confused, but won't this also set `crc32c` to be precomputed?

how do I ask the client library to use the precomputed `md5` hash I give it and compute its own `crc32c` ?

coryan: > how do I ask the client library to use the precomputed `md5` hash I give it and compute its own `crc32c` ?

You don't.  If you have pre-computed hashes, that is all what the library will use.  It may be a terrible idea, but it seems wasteful to do something like `client.upload_object(...).with_crc32c(...).with_computed_crc32c(...)`. And I cannot figure how to disable that without also disabling `.with_computed_md5()`.

src/storage/src/storage/upload_object.rs
```diff
@@ -525,30 +527,76 @@ impl<T> UploadObject<T> {
...
+    ///
+    /// In some applications, the payload's MD5 hash is already known. For
+    /// example, the application may be downloading the data from another blob
+    /// storage system.
+    ///
+    /// In such cases, it is safer to pass the known MD5 of the payload to
+    /// [Cloud Storage], and more efficient to skip the computation in the
+    /// client library.
+    ///
+    /// Note that once you provide a CRC32C value to this builder you cannot
```
suzmue: s/CRC23C/MD5

src/storage/src/storage/upload_object.rs
```diff
@@ -525,30 +527,76 @@ impl<T> UploadObject<T> {
...
+    /// In some applications, the payload's CRC32C checksum is already known.
+    /// For example, the application may be downloading the data from another
+    /// blob storage system.
+    ///
+    /// In such cases, it is safer to pass the known CRC32C of the payload to
+    /// [Cloud Storage], and more efficient to skip the computation in the
+    /// client library.
+    ///
+    /// Note that once you provide a CRC32C value to this builder you cannot
+    /// enable additional checksums.
```
suzmue: What does this mean? I had assumed it meant this is not allowed but it is:

```rust
let response = client
    .upload_object("projects/_/buckets/my-bucket", "my-object", "hello world")
    .with_crc32c(crc32c)
    .with_md5(md5)
    .send()
    .await?;
```

src/storage/src/storage/upload_object.rs
```diff
@@ -766,9 +978,59 @@ where
...
+    /// Precomputing the checksums can be expensive if the data source is slow
+    /// to read. Therefore, the client library does not precompute the checksums
+    /// by default. The client library compares the checksums computed by the
+    /// service against its own checksums. If they do not match, the client
+    /// library returns an error. However, the service has already created the
+    /// object with the (likely incorrect) data.
+    ///
+    /// If the checksums are known when the upload starts, the client library
+    /// can include the checksums with the upload request, and the service can
+    /// reject the upload if the payload and the checksums do not match.
```
suzmue: This seems like the most important part of this doc comment, since this is the reason to use this function. Maybe we can move that earlier in the description?

I also wonder if the name could reflect this better, though if we switch to grpc we may be able to upload this at the end. Don't have great ideas for that naming tho

src/storage/src/storage/upload_object.rs
```diff
@@ -709,6 +757,179 @@ impl<T> UploadObject<T> {
...
+    ///     .send()
+    ///     .await?;
+    /// println!("response details={response:?}");
+    /// # Ok(()) }
+    /// ```
+    ///
+    /// See [precompute_checksums][UploadObject::precompute_checksums] for more
+    /// details on how checksums are used by the client library and their
+    /// limitations.
+    pub fn with_computed_md5(self) -> UploadObject<T, Md5> {
```
suzmue: nit: I think the naming for these is a little confusing with "computed" and "precompute" and figuring out how those interact. Specifically I think "computed" actually makes me think this value is already known, and it took me a minute to realize this is not going to do what with_md5() does.

Maybe a rename to compute_md5? Though we are also not computing at that time so don't know if thats better.
```rust
let response = client
     .upload_object("projects/_/buckets/my-bucket", "my-object", payload)
     .compute_md5()  // .with_computed_md5()
     .precompute_checksums()
     .await?
     .send()
     .await?;
```

src/storage/src/storage/upload_object.rs
coryan: Clarified.

src/storage/src/storage/upload_object.rs
coryan: Fixed. and clarified the text

src/storage/src/storage/upload_object.rs
coryan: Done.

src/storage/src/storage/upload_object.rs
coryan: Done.

src/storage/src/storage/upload_object.rs
coryan: I think we could relax this in the future by introducing `PrecomputedCrc32c` and `PrecomputedMd5`, where each of those would still allow you to still compute the other value.

coryan (commented):
PTAL

src/storage/src/storage/upload_object.rs
```diff
@@ -766,9 +986,59 @@ where
...
+    /// [JSON API]: https://cloud.google.com/storage/docs/json_api
+    pub async fn precompute_checksums(mut self) -> Result<UploadObject<T, Precomputed>>
+    where
+        C: ChecksumEngine + Send + Sync + 'static,
+    {
+        if self.mut_resource().checksums.is_some() {
+            return Ok(self.switch_checksum(|_| Precomputed));
+        }
+        let mut offset = 0_u64;
+        self.payload.seek(offset).await.map_err(Error::ser)?;
```
dbolduc, line 1030: Q: do we need to rewind at the *end* of the computation?

dbolduc: Nvm, I see that we rewind before starting uploads:

https://github.com/googleapis/google-cloud-rust/blob/59edd349a94a3c3ad1b26d3aaa13dcac336d66c3/src/storage/src/storage/perform_upload/unbuffered.rs#L75-L82

and

https://github.com/googleapis/google-cloud-rust/blob/59edd349a94a3c3ad1b26d3aaa13dcac336d66c3/src/storage/src/storage/perform_upload/unbuffered.rs#L149

coryan:
I merged this to unblock other work. We will need to make a pass through all the function names and documentation again anyway.

src/storage/src/storage/upload_object.rs
coryan: Seems like we both missed this:

https://github.com/googleapis/google-cloud-rust/blob/a2940f1d7241ea995c93346a5ad1cbebf69ca702/src/storage/src/storage/perform_upload/buffered.rs#L132-L135

Will fix in #2790

Issue #2641, opened 2025-07-11
Stabilize the `send*()` methods


coryan:
Decide if we should have `send_buffered()`, `send_unbuffered()` and make `send()` just an shortcut to the one we want to recommend (probably unbuffered).

coryan:
Closing. We may need more surface changes to implement the stubs, but we already have a bug to track this.

PR #2789, opened 2025-07-29
impl(storage): make error types public

### LINKED ISSUE #2050: Verify checksums during uploads

coryan:
Fixes #2050

### DIFF
--- a/src/storage/src/lib.rs
+++ b/src/storage/src/lib.rs
@@ -66,6 +66,9 @@ pub mod builder {
 pub use crate::control::model;
 pub use crate::control::stub;
 
+pub use storage::ChecksumMismatch;
+pub use storage::UploadError;
+
 #[allow(dead_code)]
 pub(crate) mod generated;
 
--- a/src/storage/src/storage.rs
+++ b/src/storage/src/storage.rs
@@ -77,7 +77,7 @@ pub enum UploadError {
 /// The error type for checksum comparisons.
 ///
 /// By default the client library computes a checksum of the uploaded data, and
-/// compares this checksums against the value returned by the service.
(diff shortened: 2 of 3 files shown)


src/storage/src/storage.rs
```diff
@@ -44,3 +73,46 @@ enum UploadError {
     #[error("the service reports {persisted} bytes as persisted, but we only sent {sent} bytes")]
     TooMuchProgress { sent: u64, persisted: u64 },
 }
+
+/// The error type for checksum comparisons.
+///
+/// By default the client library computes a checksum of the uploaded data, and
+/// compares this checksums against the value returned by the service.
```
dbolduc: s/this checksums/this checksum/

dbolduc (approved):

src/storage/src/lib.rs
```diff
@@ -66,6 +66,9 @@ pub mod builder {
 pub use crate::control::model;
 pub use crate::control::stub;
 
+pub use storage::ChecksumMismatch;
```
dbolduc, line 69: there might be enough of these to justify a `mod errors`.

I am fine with the current organization while the crate is preview.

coryan: Included as a list item in #2641

src/storage/src/storage.rs
coryan: Fixed.

coryan (commented):
PTAL

PR #2790, opened 2025-07-30
impl(storage): compare final checksums on upload

coryan:
Compare the computed (or pre-computed) checksums against the values
received from the service. Returning an error if the values do not
match.

Part of the work for #2050

### DIFF
--- a/src/integration-tests/src/storage.rs
+++ b/src/integration-tests/src/storage.rs
@@ -792,23 +792,127 @@ pub async fn checksums(
 
     type ObjectResult = storage::Result<storage::model::Object>;
 
-    let uploads: Vec::<(&str, std::pin::Pin<Box<dyn Future<Output = ObjectResult>>>)> = vec![
-        ("verify/default", Box::pin(client.upload_object(bucket_name, "verify/default",  VEXING).with_if_generation_match(0).send())),
-        ("verify/disabled", Box::pin(client.upload_object(bucket_name, "verify/disabled", VEXING).with_if_generation_match(0).disable_computed_checksums().send())),
-        ("verify/crc32c", Box::pin(client.upload_object(bucket_name, "verify/crc32c",   VEXING).with_if_generation_match(0).disable_computed_checksums().compute_crc32c().send())),
-        ("verify/md5", Box::pin(client.upload_object(bucket_name, "verify/md5",      VEXING).with_if_generation_match(0).disable_computed_checksums().compute_md5().send())),
-        ("verify/both", Box::pin(client.upload_object(bucket_name, "verify/both",     VEXING).with_if_generation_match(0).disable_computed_checksums().compute_md5().send())),
-        ("computed/default", Box::pin(client.upload_object(bucket_name, "computed/default",  VEXING).with_if_generation_match(0).precompute_checksums().await?.send())),
-        ("computed/disabled", Box::pin(client.upload_object(bucket_name, "computed/disabled", VEXING).with_if_generation_match(0).disable_computed_checksums().precompute_checksums().await?.send())),
-        ("computed/crc32c", Box::pin(client.upload_object(bucket_name, "computed/crc32c",   VEXING).with_if_generation_match(0).disable_computed_checksums().compute_crc32c().precompute_checksums().await?.send())),
-        ("computed/md5", Box::pin(client.upload_object(bucket_name, "computed/md5",      VEXING).with_if_generation_match(0).disable_computed_checksums().compute_md5().precompute_checksums().await?.send())),
-        ("computed/both", Box::pin(client.upload_object(bucket_name, "computed/both",     VEXING).with_if_generation_match(0).disable_computed_checksums().compute_md5().precompute_checksums().await?.send())),
+    let uploads: Vec<(&str, std::pin::Pin<Box<dyn Future<Output = ObjectResult>>>)> = vec![
+        (
+            "verify/default",
+            Box::pin(
+                client
+                    .upload_object(bucket_name, "verify/default", VEXING)
+                    .with_if_generation_match(0)
+                    .send(),
+            ),
+        ),
+        (
+            "verify/disabled",
+            Box::pin(
+                client
+                    .upload_object(bucket_name, "verify/disabled", VEXING)
+                    .with_if_generation_match(0)
+                    .disable_computed_checksums()
+                    .send(),
+            ),
+        ),
+        (
+            "verify/crc32c",
+            Box::pin(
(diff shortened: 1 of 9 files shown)


src/integration-tests/src/storage.rs
```diff
@@ -767,6 +767,162 @@ async fn abort_upload_buffered(client: storage::client::Storage, bucket_name: &s
...
+        tracing::subscriber::set_default(subscriber)
+    };
+
+    tracing::info!("checksums test, using bucket {bucket_name}");
+
+    let client = builder.build().await?;
+    const VEXING: &str = "how vexingly quick daft zebras jump";
+
+    type ObjectResult = storage::Result<storage::model::Object>;
+    type Boxed = std::pin::Pin<Box<dyn Future<Output = ObjectResult>>>;
```
dbolduc: opt: call this `BoxFuture<ObjectResult>`

 https://docs.rs/futures/latest/futures/future/type.BoxFuture.html

src/integration-tests/src/storage.rs
```diff
@@ -767,6 +767,162 @@ async fn abort_upload_buffered(client: storage::client::Storage, bucket_name: &s
...
+            ),
+        ),
+        (
+            "verify/both",
+            Box::pin(
+                client
+                    .upload_object(bucket_name, "verify/both", VEXING)
+                    .with_if_generation_match(0)
+                    .disable_computed_checksums()
+                    .compute_md5()
```
dbolduc, line 845: need crc32c also

src/integration-tests/src/storage.rs
```diff
@@ -767,6 +767,162 @@ async fn abort_upload_buffered(client: storage::client::Storage, bucket_name: &s
...
+                client
+                    .upload_object(bucket_name, "computed/default", VEXING)
+                    .with_if_generation_match(0)
+                    .precompute_checksums()
+                    .await?
+                    .send(),
+            ),
+        ),
+        (
+            "computed/disabled",
```
dbolduc, line 861: 🫨

[insert Ryan Reynolds "but why?" meme]

src/integration-tests/src/storage.rs
```diff
@@ -767,6 +767,162 @@ async fn abort_upload_buffered(client: storage::client::Storage, bucket_name: &s
...
+            ),
+        ),
+        (
+            "computed/both",
+            Box::pin(
+                client
+                    .upload_object(bucket_name, "computed/both", VEXING)
+                    .with_if_generation_match(0)
+                    .disable_computed_checksums()
+                    .compute_md5()
```
dbolduc, line 906: need crc32c

src/storage/src/storage/perform_upload/buffered.rs
```diff
@@ -127,21 +132,51 @@ where
...
+        let computed = stream.final_checksum();
+        let has = self
+            .spec
+            .resource
+            .get_or_insert_default()
+            .checksums
+            .get_or_insert_default();
+        if has.crc32c.is_none() {
+            has.crc32c = computed.crc32c;
+        }
```
dbolduc, line 153: nit: use `Option::get_or_insert`?

src/storage/src/storage/perform_upload/buffered.rs
```diff
@@ -127,21 +132,51 @@ where
...
+    async fn send_buffered_single_shot(mut self) -> Result<Object> {
         let mut stream = self.payload.lock().await;
         let mut collected = Vec::new();
         while let Some(b) = stream.next().await.transpose().map_err(Error::ser)? {
             collected.push(b);
         }
+        let source = IterSource::new(collected);
+        // Use the computed checksum, if any, and if the spec does not have a
+        // checksum already.
+        let computed = stream.final_checksum();
```
dbolduc, line 144: Can we skip this computation if the spec already has a checksum?

src/storage/src/storage/perform_upload/tests/checksums.rs
```diff
@@ -0,0 +1,353 @@
...
+
+fn vexing_md5() -> bytes::Bytes {
+    bytes::Bytes::from_owner(md5::compute(VEXING.as_bytes()).0)
+}
+
+fn bad_checksums_body() -> Value {
+    // The magic strings can be regenerated using:
+    //     rm empty.txt
+    //     touch empty.txt
+    //     gcloud hash vexing.txt
```
dbolduc: nit: s/vexing/empty/

src/integration-tests/src/storage.rs
coryan: Done.

src/integration-tests/src/storage.rs
coryan: Thanks.

src/integration-tests/src/storage.rs
coryan: /idk just to make sure nothing breaks.

src/integration-tests/src/storage.rs
coryan: Fixed.

src/storage/src/storage/perform_upload/buffered.rs
coryan: `computed.crc32c` is also an `Option<u32>`, so it needs to be `has.crc32c = has.crc32.or(computed.crc32c)` or something. Is that better?

src/storage/src/storage/perform_upload/tests/checksums.rs
coryan: Fixed.

src/storage/src/storage/perform_upload/buffered.rs
coryan: What computation? Reading the stream (which already computes the checksums) is unavoidable. Calling `final_checksum()` is (or should be) relatively cheap, and the code is already too branchy. I think I will leave as-is.

dbolduc (approved):

src/storage/src/storage/perform_upload/buffered.rs
dbolduc: > `.or()`

You're right.

> Is that better?

I think `x = x.or(y)` is more my style.

Like I would say `a = max(a, b);` instead of `if a < b { a = b; }`.

But no big deal. I would not say either way to write it is better than the other.

src/storage/src/storage/perform_upload/buffered.rs
dbolduc: Ack

PR #2791, opened 2025-07-30
feat(bigquery): query client

alvarowolfx:
This is an experimental branch, as we are doing with other languages to define the new query experience for BigQuery. This is not gonna be merged in the current state ( will be broken down into smaller PRs). 

THERE WILL BE DRAGONS, don't worry right now on deeply reviewing this. But happy to hear thoughts on overall customer facing interface. That can be seem on integration tests. 

Things to work on yet:

- [ ] Overall docs are missing 
- [ ] Overall error structure and types are confusing right now. We have `QueryError`, `ParsingError`, `ValueError` and more. Need to combine some of that and streamline error handling for customers.
- [ ] Improve overall error handling (there are `.unwrap()` calls everywhere)
- [ ] Improve overall reference passing. Too much data is being copied and can be reused and shared
   - [ ] Each `Row` has a `Schema`, but that can be a reference, since all `Rows` for a given query will have the same `Schema`.
   - [ ] Pass `QueryClient` instead of `JobService` directly. Need to figure out best way to have this as shared ref.
- [ ] Iterator should follow the same `paginator` trait and pattern from our SDK.
- [ ] Split iterator vs actual data fetching. Currently the implementation is using `jobs.getQueryResults` API, but in the future, BigQuery Storage Read API also needs to be supported.
- [ ] Client builder needs to accept `JobService` provided by customer. `TableService` and BQ Storage Read service also need to be added later.

### DIFF
--- a/src/integration-tests/src/bigquery.rs
+++ b/src/integration-tests/src/bigquery.rs
@@ -85,19 +85,30 @@ pub async fn run_query_nested_data(builder: ClientBuilder) -> Result<()> {
     assert!(nested_array.is_some());
     let nested_array = nested_array.unwrap();
     assert_eq!(nested_array.len(), 1);
-    let object = nested_array[0].as_object().expect("nested item should be an object");
-    let nested_object = object.get("object")
+    let object = nested_array[0]
+        .as_object()
+        .expect("nested item should be an object");
+    let nested_object = object
+        .get("object")
         .expect("msg object should have `object` field")
         .as_object()
         .expect("msg object should have `object` object field");
-    assert_eq!(nested_object.get("a")
-        .expect("msg object should have `a` field")
-        .as_str()
-        .expect("msg object should have `a` string field"), "1");
-    assert_eq!(nested_object.get("b")
-        .expect("msg object should have `b` field")
-        .as_str()
-        .expect("msg object should have `b` string field"), "2");
+    assert_eq!(
+        nested_object
+            .get("a")
+            .expect("msg object should have `a` field")
+            .as_str()
+            .expect("msg object should have `a` string field"),
+        "1"
+    );
+    assert_eq!(
+        nested_object
+            .get("b")
+            .expect("msg object should have `b` field")
+            .as_str()
+            .expect("msg object should have `b` string field"),
+        "2"
+    );
 
     // Parse as user defined struct
     #[derive(serde::Deserialize, Debug)]
@@ -106,15 +117,15 @@ pub async fn run_query_nested_data(builder: ClientBuilder) -> Result<()> {
     }
 
     #[derive(serde::Deserialize, Debug)]
-    struct NestedArrayObject{
-        object: NestedObject
+    struct NestedArrayObject {
+        object: NestedObject,
     }
 
     #[derive(serde::Deserialize, Debug)]
-    struct NestedObject {        
+    struct NestedObject {
         a: String,
-        b: String
-    }    
+        b: String,
+    }
 
     let my_struct: MyStruct =
         serde_json::from_value(first_row.to_value()).expect("Should parse as user defined struct");


tswast:
> Split iterator vs actual data fetching. Currently the implementation is using jobs.getQueryResults API, but in the future, BigQuery Storage Read API also needs to be supported.

So long as a table reference can be got from the query results resource, this should be possible even if not integrated into this client directly.

src/bigquery/src/client.rs
```diff
@@ -0,0 +1,195 @@
...
+            req.project_id = self.project_id.clone();
+        }
+        let res = self
+            .job_service
+            .query()
+            .set_project_id(req.project_id)
+            .set_or_clear_query_request(req.query_request)
+            .send()
+            .await?;
+        Ok(Query::new(self.job_service.clone(), res))
```
tswast, line 78: In Python, we automatically switch to jobs.insert if some parameter was passed that isn't supported by jobs query. But perhaps we'll provide a separate jobs.insert API for that instead?

src/bigquery/src/client.rs
```diff
@@ -0,0 +1,195 @@
...
+        if req.project_id.is_empty() {
+            req.project_id = self.project_id.clone();
+        }
+        let res = self
+            .job_service
+            .query()
+            .set_project_id(req.project_id)
+            .set_or_clear_query_request(req.query_request)
+            .send()
+            .await?;
```
tswast, line 77: Does this service do any exponential back off / retries? If so, is there a way to configure them? The BQ API makes it difficult to rely purely on HTTP status codes.

src/bigquery/src/iterator.rs
```diff
@@ -0,0 +1,110 @@
...
+use bigquery_v2::model::DataFormatOptions;
+use std::collections::VecDeque;
+use std::sync::Arc;
+
+use crate::Result;
+use crate::query::Query;
+use crate::row::Row;
+use crate::schema::Schema;
+
+pub struct RowIterator {
```
tswast, line 25: There's a few other properties that we've found useful to expose:

All the ones from https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#response-body / https://cloud.google.com/bigquery/docs/reference/rest/v2/jobs/getQueryResults#response-body are the main ones. totalSlotMs from jobs.query was an important recent customer ask, for example.

Also some from the Job resource, like destination table. I suppose we could make the user call jobs.get to fetch that, though, to keep this interface smaller.

src/bigquery/src/query.rs
```diff
@@ -0,0 +1,178 @@
...
+    }
+
+    pub async fn read(self) -> Result<RowIterator> {
+        if !self.completed {
+            return Err(crate::Error::ser(QueryError::NotComplete));
+        }
+        RowIterator::try_new(Arc::new(self)).await
+    }
+
+    async fn poll_job(&mut self) -> Result<bool> {
```
tswast, line 124: Would folks want a timeout for poll_job / wait?

src/bigquery/src/query.rs
```diff
@@ -0,0 +1,178 @@
...
+        if self.completed {
+            return Ok(());
+        }
+        let sleep = async |d| tokio::time::sleep(d).await;
+        loop {
+            let complete = self.poll_job().await?;
+            if complete {
+                return Ok(());
+            }
+            // TODO: exponential backoff
```
tswast, line 112: FWIW: we haven't done any exponential back off in Python for polling, just retries, AFAIK. jobs.getQueryResults hangs up to 10 seconds, so it didn't seem necessary or desirable to make folks wait any longer than they needed to.

src/bigquery/src/query.rs
```diff
@@ -0,0 +1,178 @@
...
+            return Err(crate::Error::ser(QueryError::NotComplete));
+        }
+        RowIterator::try_new(Arc::new(self)).await
+    }
+
+    async fn poll_job(&mut self) -> Result<bool> {
+        let max_results: u32 = 0;
+        let res = self
+            .job_service
+            .get_query_results()
```
tswast, line 128: FWIW: this is where a lot of the retriable query failures will surface. It can make the system feel a lot more reliable if we can re-issue a jobs.query API call here in the case of a retriable query failure.

src/bigquery/src/client.rs
alvarowolfx: but that only make sense if we provide an abstraction on top of the `PostQueryRequest` and `Job` proto definitions.  What we are proposing in the new query contract is to have different paths for users wanting to just run a query (which will build a `PostQueryRequest` and use `jobs.query`) and other with more advance use cases will use a `Job` proto with `jobs.insert`. We want to reduce the amount of abstrations and use more the proto definitions, to avoid the maintenance toil of keeping those abstractions in sync with the service side in the various supported languages.

src/bigquery/src/client.rs
alvarowolfx: right now there is no retry set up, just the base one from the `gax` package. That's what I still need to explore, but the reference implementation that I have been working is in Go for the whole query experience.

src/bigquery/src/iterator.rs
alvarowolfx: yeah, we are leaning on letting the use access the job information and query for more if needed with `jobs.get`, but good point on that, need to document that better.

src/bigquery/src/query.rs
alvarowolfx: good point, I added exponential backoff here as it's the default mechanism that we use when talking about retry. But I'll change that on the doc that we need to have a different strategy in BQ

src/bigquery/src/query.rs
alvarowolfx: yeah, in Go we added support for context cancelation and in Node with Promise/AbortController. Need to check what to do in Rust, but that is really language specific.

coryan:
Is the intention to merge this PR (after some changes) or was it just intended to capture some discussions? If the latter, please close it. We like to keep the PR list tidy.

alvarowolfx:
> Is the intention to merge this PR (after some changes) or was it just intended to capture some discussions? If the latter, please close it. We like to keep the PR list tidy.

will close for now, as stated in the PR description, this is going to be broken down into smaller PR and this is to show the bigger picture and capture thoughts on the overall design. But later when we come back to it we can reopen it.

PR #2792, opened 2025-07-30
impl(storage): use shared checksum library for download checksum

suzmue:
Migrate to use the checksum library for read_object checksum calculation, cleaning up the duplicate code.

This also is the first step to adding opt-in support for md5 calculation.

For #2049

### DIFF
--- a/src/storage/src/storage/read_object.rs
+++ b/src/storage/src/storage/read_object.rs
@@ -15,6 +15,8 @@
 use super::client::*;
 use super::*;
 use crate::download_resume_policy::DownloadResumePolicy;
+use crate::model::ObjectChecksums;
+use crate::storage::checksum::{ChecksumEngine, Crc32c, validate};
 use base64::Engine;
 #[cfg(feature = "unstable-stream")]
 use futures::Stream;
@@ -487,8 +489,10 @@ pub struct ReadObjectResponse {
     inner: Option<reqwest::Response>,
     highlights: ObjectHighlights,
     // Fields for tracking the crc checksum checks.
-    response_crc32c: Option<u32>,
-    crc32c: u32,
+    response_checksums: ObjectChecksums,
+    checksums: Crc32c,
+    offset: u64,
+    // Fields for resuming download.
     range: ReadRange,
     generation: i64,
     builder: ReadObject,
@@ -499,6 +503,7 @@ impl ReadObjectResponse {
     fn new(builder: ReadObject, inner: reqwest::Response) -> Result<Self> {
         let full = builder.request.read_offset == 0 && builder.request.read_limit == 0;
         let response_crc32c = crc32c_from_response(full, inner.status(), inner.headers());
+        let response_checksums = ObjectChecksums::new().set_or_clear_crc32c(response_crc32c);
         let range = response_range(&inner).map_err(Error::deser)?;
         let generation = response_generation(&inner).map_err(Error::deser)?;
         let headers = inner.headers();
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/read_object.rs
```diff
@@ -601,8 +609,11 @@ impl ReadObjectResponse {
         let res = inner.chunk().await.map_err(Error::io);
         match res {
             Ok(Some(chunk)) => {
-                if self.response_crc32c.is_some() {
-                    self.crc32c = crc32c::crc32c_append(self.crc32c, &chunk);
+                if self.response_checksums.crc32c.is_some()
+                    || !self.response_checksums.md5_hash.is_empty()
```
dbolduc: So if the service gives us an md5, we will update our crc32c? That seems odd.

Remove this line until we are ready for md5?

src/storage/src/storage/read_object.rs
```diff
@@ -619,9 +630,12 @@ impl ReadObjectResponse {
                 if self.range.limit != 0 {
                     return Some(Err(Error::io(ReadError::ShortRead(self.range.limit))));
                 }
-                let res = check_crc32c_match(self.crc32c, self.response_crc32c);
+                let res = validate(
+                    self.response_checksums.clone(),
```
dbolduc: Can we avoid the `clone()` by changing `validate` to take `expected: &ObjectChecksums`?

coryan (dismissed):

src/storage/src/storage/read_object.rs
```diff
@@ -533,8 +538,11 @@ impl ReadObjectResponse {
         Ok(Self {
             inner: Some(inner),
             highlights,
-            response_crc32c,
-            crc32c: 0, // no bytes read yet.
+            // Fields for computing checksums.
+            response_checksums,
+            checksums: Crc32c::default(),
+            offset: 0,
```
dbolduc: I think this is `self.range.start`

src/storage/src/storage/read_object.rs
dbolduc: I see that @coryan is making this change in #2790. Y'all will have to fight over who merges first.

Maybe it is worth refactoring `validate` in a super small PR, then rebasing this and #2790 :shrug:

suzmue: I'll have this follow #2790 since that is higher priority.

coryan: Thanks, I just merged that PR.

src/storage/src/storage/read_object.rs
suzmue: Yes, but for now we don't calculate if `self.range.start != 0`, so the only relevant case is when offset = 0. Will update this anyway.

src/storage/src/storage/read_object.rs

src/storage/src/storage/read_object.rs
dbolduc: I don't follow.

When does `offset != self.range.start` ? If they are always the same, let's use `self.range.start` and not introduce `offset`.

suzmue: My comment above is saying that since checksums are currently only supported in GCS when requesting the whole contents. Partial contents do not have checksums. So when we compute checksums, we are always starting at 0.

I thought you were just pointing out the start value, not the value at all times. Yeah it looks like they do stay synced. Updated.

dbolduc (approved):

src/storage/src/storage.rs
```diff
@@ -121,8 +121,8 @@ pub enum ChecksumMismatch {
 #[non_exhaustive]
 pub enum ReadError {
     /// The calculated crc32c did not match server provided crc32c.
-    #[error("bad CRC on read: got {got}, want {want}")]
-    BadCrc { got: u32, want: u32 },
+    #[error("checksum mismatch")]
+    ChecksumMismatch(crate::ChecksumMismatch),
```
dbolduc, line 125: @coryan - #2821 is removing this type. Do you think we should enumerate the mismatches here too?

suzmue: Decided offline @coryan will resolve merge in his PR.

PR #2794, opened 2025-07-30
feat(storage): add more fields to ObjectHighlights

### LINKED ISSUE #2626: Add method for getting object metadata from Storage::ReadObjectResponse

suzmue:
Add remaining fields for data that is returned as headers for object downloads.

Fixes #2626

### DIFF
--- a/src/storage/src/storage/read_object.rs
+++ b/src/storage/src/storage/read_object.rs
@@ -501,58 +501,37 @@ impl ReadObjectResponse {
         let response_crc32c = crc32c_from_response(full, inner.status(), inner.headers());
         let range = response_range(&inner).map_err(Error::deser)?;
         let generation = response_generation(&inner).map_err(Error::deser)?;
+
         let headers = inner.headers();
-        let highlights = ObjectHighlights {
-            generation: headers
-                .get("x-goog-generation")
-                .and_then(|g| g.to_str().ok())
-                .and_then(|g| g.parse::<i64>().ok())
-                .unwrap_or_default(),
-            metageneration: headers
-                .get("x-goog-metageneration")
-                .and_then(|m| m.to_str().ok())
-                .and_then(|m| m.parse::<i64>().ok())
-                .unwrap_or_default(),
-            size: headers
-                .get("x-goog-stored-content-length")
+        let get_as_i64 = |header_name: &str| -> i64 {
+            headers
+                .get(header_name)
                 .and_then(|s| s.to_str().ok())
                 .and_then(|s| s.parse::<i64>().ok())
-                .unwrap_or_default(),
-            content_encoding: headers
-                .get("x-goog-stored-content-encoding")
-                .and_then(|ce| ce.to_str().ok())
-                .map(|ce| ce.to_string())
-                .unwrap_or_default(),
+                .unwrap_or_default()
+        };
+        let get_as_string = |header_name: &str| -> String {
+            headers
+                .get(header_name)
+                .and_then(|sc| sc.to_str().ok())
+                .map(|sc| sc.to_string())
+                .unwrap_or_default()
+        };
+        let highlights = ObjectHighlights {
+            generation,
+            metageneration: get_as_i64("x-goog-metageneration"),
+            size: get_as_i64("x-goog-stored-content-length"),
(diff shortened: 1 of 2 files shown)


src/storage/src/storage/read_object.rs
```diff
@@ -528,6 +528,31 @@ impl ReadObjectResponse {
                     .set_or_clear_crc32c(headers_to_crc32c(headers))
                     .set_md5_hash(headers_to_md5_hash(headers))
             }),
+            storage_class: headers
+                .get("x-goog-storage-class")
+                .and_then(|sc| sc.to_str().ok())
+                .map(|sc| sc.to_string())
+                .unwrap_or_default(),
```
coryan: nit: this repeats the same expression 5 times, maybe a helper function that consumes the header name?

src/storage/src/storage/read_object.rs
```diff
@@ -697,6 +722,26 @@ pub struct ObjectHighlights {
     /// the client should compute one of these checksums over the downloaded
     /// object and compare it against the value provided here.
     pub checksums: std::option::Option<crate::model::ObjectChecksums>,
+
+    /// Storage class of the object.
+    pub storage_class: String,
+
+    /// Content-Language of the object data, matching
+    /// [<https://tools.ietf.org/html/rfc7231#section-3.1.3.2>][RFC 7231 §3.1.3.2].
+    pub content_language: std::string::String,
```
coryan: nit: why `std::string::String` here vs. `String` in the previous one? Maybe cut&paste from somewhere?

coryan (changes requested):

src/storage/src/storage/read_object.rs
suzmue: Ah thanks good catch. Was copying these from the generated crate::model::Object.

src/storage/src/storage/read_object.rs
suzmue: Added a helper.

coryan (dismissed):

src/storage/src/storage/read_object.rs
```diff
@@ -697,6 +701,26 @@ pub struct ObjectHighlights {
     /// the client should compute one of these checksums over the downloaded
     /// object and compare it against the value provided here.
     pub checksums: std::option::Option<crate::model::ObjectChecksums>,
+
+    /// Storage class of the object.
+    pub storage_class: String,
+
+    /// Content-Language of the object data, matching
+    /// [<https://tools.ietf.org/html/rfc7231#section-3.1.3.2>][RFC 7231 §3.1.3.2].
```
dbolduc: These RFC links are all backwards. We should fix them all here. It would be googley to also fix them upstream in google3.

dbolduc: observation: these comments *could* fall out of sync with the generated comments. I don't think it is worth addressing though. I doubt they ever change significantly.

src/storage/src/storage/read_object.rs
```diff
@@ -697,6 +701,26 @@ pub struct ObjectHighlights {
...
+    /// `application/octet-stream`.
+    pub content_type: String,
+
+    /// Content-Disposition of the object data, matching
+    /// [<https://tools.ietf.org/html/rfc6266>][RFC 6266].
+    pub content_disposition: String,
+
+    /// The etag of the object.
+    pub etag: String,
 }
```
dbolduc, line 725: Also, should the class be `#[non_exhaustive]`?

src/storage/src/storage/read_object.rs

src/storage/src/storage/read_object.rs
suzmue: Done.

PR #2800, opened 2025-07-31
chore: bump versions from the last release

coryan:


dbolduc:
Update `guide/samples/Cargo.toml`?

guide/samples/Cargo.toml
```diff
@@ -32,15 +32,15 @@ name = "getting_started"
 google-cloud-longrunning = { version = "0.25.2", path = "../../src/generated/longrunning" }
```
dbolduc, line 32: We could update this, and `google-cloud-lro` below

coryan: I think I am going to make a pass replacing all of these with `cargo add google-cloud-longrunning`. That is far easier to keep up to date.

PR #2806, opened 2025-08-01
refactor: delete generator/

julieqiu:
Code in generator/ has been moved to
https://github.com/googleapis/librarian/tree/main/internal/sidekick.

Code in generator/dart is moved to dart/.

For https://github.com/googleapis/librarian/issues/1510.


coryan:
We need the CI builds to start using the code from librarian. Maybe add a `librarian/` submodule pointing at the right SHA and change the scripts to use that? Or change the scripts to use a specific SHA directly?

julieqiu:
> We need the CI builds to start using the code from librarian. Maybe add a `librarian/` submodule pointing at the right SHA and change the scripts to use that? Or change the scripts to use a specific SHA directly?

Yeah I just realized they were more complicated than `go test ./...`. Closing until we figure this out.

PR #2810, opened 2025-08-01
prototype(auth): Default retry

harkamaljot:


src/auth/src/retry.rs
```diff
@@ -131,6 +150,53 @@ where
...
+        {
+            let guard = rx.borrow();
+            if let Some(result) = &*guard {
+                let is_transient = match result {
+                    Err(e) => e.is_transient(),
+                    _ => false,
+                };
+                if !is_transient {
+                    return result.clone();
+                }
```
sai-sunder-s: shouldn't we return the current result we know even if its a transient error?

As long as there is some result it should be returned immediately I believe?

harkamaljot: thanks yeah, I simplified it.

src/auth/src/token_cache.rs
```diff
@@ -102,7 +102,15 @@ where
 async fn wait_for_next_token(
     mut rx_token: watch::Receiver<Option<Result<(Token, EntityTag)>>>,
 ) -> Result<(Token, EntityTag)> {
-    rx_token.changed().await.unwrap();
+    if rx_token.changed().await.is_err() {
```
sai-sunder-s, line 105: Wont gax call wait here still?

src/auth/src/token_cache.rs
```diff
@@ -77,7 +77,7 @@ where
                     }
                 },
                 // An error in the result is still a valid result to propagate to the client library
-                Err(e) => Err(e),
+                Err(_) => wait_for_next_token(rx).await,
```
sai-sunder-s: why this change

src/auth/src/token_cache.rs
harkamaljot: this change is a no-op/cleanup here as we are directly calling unwrap here which is a bad practice.

src/auth/src/retry.rs
```diff
@@ -103,31 +118,91 @@ where
             }
         };
 
-        retry_loop(
+        // The on_retry callback sends intermediate errors and stores the last delay.
+        let last_delay = self.last_delay.clone();
+        let tx_for_retry = self.token_watch_tx.clone();
+        let on_retry = move |_, error: &gax::error::Error, delay: Duration| {
+            // Store the delay. If the loop terminates on the next attempt, this will be the relevant delay.
```
sai-sunder-s, line 125: why to delay if retry terminates?

src/auth/src/retry.rs
```diff
@@ -103,31 +118,91 @@ where
...
-            _ => constants::TOKEN_FETCH_FAILED_ERROR,
+    fn map_retry_error(e: gax::error::Error, last_delay: Option<Duration>) -> CredentialsError {
+        let mut cred_error = if !e.is_authentication() {
+            CredentialsError::from_source(false, e)
+        } else {
+            let msg = match e
+                .source()
+                .and_then(|s| s.downcast_ref::<CredentialsError>())
+            {
+                Some(cred_error) if cred_error.is_transient() => constants::RETRY_EXHAUSTED_ERROR,
```
sai-sunder-s, line 173: why exhausted here?

src/auth/src/retry.rs
```diff
@@ -103,31 +118,91 @@ where
...
+                _ => constants::TOKEN_FETCH_FAILED_ERROR,
+            };
+            CredentialsError::new(false, msg, e)
         };
-        CredentialsError::new(false, msg, e)
+        if let Some(delay) = last_delay {
+            cred_error = cred_error.with_retry_in(delay);
```
sai-sunder-s, line 179: we should make it `with_retry_at` and give it an actual instant or time to retry at. We do not know when the caller is going to get this error.

src/auth/src/token_cache.rs
sai-sunder-s: Yeah change for unwrap is ok. But still the gax call is going to wait here right?

src/auth/src/retry.rs
```diff
@@ -64,37 +66,50 @@ impl Builder {
...
             inner: Arc::new(token_provider),
             retry_policy,
             backoff_policy,
             retry_throttler,
-        }
+            token_watch_tx: tx,
+            last_delay: Arc::new(Mutex::new(None)),
+        };
+
+        tokio::spawn(refresher.run());
```
sai-sunder-s, line 86: This background task is going to stop once a token is obtained successfully I believe? What happens when we need the next token then?

src/auth/src/token_cache.rs
```diff
@@ -149,10 +157,18 @@ async fn refresh_task<T>(
                     break;
                 }
             }
-            Err(_) => {
-                // The retry policy has been used already by the inner token provider.
-                // If it ended in an error, just quit the background task.
-                break;
+            Err(e) => {
```
sai-sunder-s, line 160: Comment for ln 132. Lets assume there is a token in tx currently that is valid for 5 more mins, do we update the token in the tx with a error even though we are going to retry and may get a valid token within next 5 mins?

(Previously also we replaced valid token in tx with and error but that was ok since we do not retry and wont recover anyway)

harkamaljot:
@sai-sunder-s as discussed offline, this change was closed as it was adding unnecessary noise, and I am making the changes directly here: https://github.com/googleapis/google-cloud-rust/compare/main...harkamaljot:google-cloud-rust:default-retry

src/auth/src/retry.rs
harkamaljot: yes that is true, I just added delay as seconds for now as a proof of concept but yes for prod version we should need instant or duration here.

src/auth/src/retry.rs
harkamaljot: this is the old version before the changes were made as this was already addressed so resolving this comment.

PR #2812, opened 2025-08-04
fix(storage): percent-encode name once on uploads

coryan:
During uploads we were percent encoding the name of the objects twice.
Once manually, and another time because `.query()` automatically percent
encode things.

Consequently the names of the uploaded objects were all wrong (they were
the percent encoded names), and using the objects in `get_object()` or
`read_object()` or `list_objects()` became extremely difficult.

### DIFF
--- a/src/integration-tests/tests/driver.rs
+++ b/src/integration-tests/tests/driver.rs
@@ -210,6 +210,20 @@ mod driver {
         result
     }
 
+    #[test_case(Storage::builder(); "default")]
+    #[tokio::test]
+    async fn run_storage_object_names(
+        builder: storage::builder::storage::ClientBuilder,
+    ) -> integration_tests::Result<()> {
+        let (control, bucket) = integration_tests::storage::create_test_bucket().await?;
+        let result =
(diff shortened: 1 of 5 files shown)


src/integration-tests/tests/driver.rs
```diff
@@ -210,6 +210,20 @@ mod driver {
...
+    #[tokio::test]
+    async fn run_storage_object_names(
+        builder: storage::builder::storage::ClientBuilder,
+    ) -> integration_tests::Result<()> {
+        let (control, bucket) = integration_tests::storage::create_test_bucket().await?;
+        let result =
+            integration_tests::storage::object_names(builder, control.clone(), &bucket.name)
+                .await
+                .map_err(integration_tests::report_error);
+        // let _ = integration_tests::storage::cleanup_bucket(control, bucket.name).await;
```
dbolduc: nit: re-enable

coryan: Doh! Fixed.

dbolduc (approved):

PR #2814, opened 2025-08-04
fix(storage): gRPC in recommended retry policy

coryan:
The recommended retry policy should support gRPC as well as HTTP errors.

### DIFF
--- a/src/storage/src/retry_policy.rs
+++ b/src/storage/src/retry_policy.rs
@@ -99,6 +99,15 @@ impl RetryPolicy for RecommendedPolicy {
                 _ => RetryResult::Permanent(error),
             };
         }
+        if let Some(code) = error.status().map(|s| s.code) {
+            use gax::error::rpc::Code;
+            return match code {
+                Code::Internal | Code::ResourceExhausted | Code::Unavailable => {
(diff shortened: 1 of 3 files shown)


src/storage/src/retry_policy.rs
```diff
@@ -99,6 +99,15 @@ impl RetryPolicy for RecommendedPolicy {
                 _ => RetryResult::Permanent(error),
             };
         }
+        if let Some(code) = error.status().map(|s| s.code) {
+            use gax::error::rpc::Code;
+            return match code {
+                Code::Internal | Code::ResourceExhausted | Code::Unavailable => {
```
dbolduc, line 105: why `INTERNAL` ? That surprised me.

https://github.com/googleapis/googleapis/blob/ebcf3f532da40475bc1e7a6ad34d31c6c32004fa/google/storage/v2/storage_grpc_service_config.json#L15

Ah...

https://github.com/googleapis/googleapis/blob/ebcf3f532da40475bc1e7a6ad34d31c6c32004fa/google/storage/control/v2/storage_control_grpc_service_config.json#L99

The recommended settings are not the same for the different underlying services. :thinking:

PR #2821, opened 2025-08-06
cleanup(storage)!: merge upload errors

### LINKED ISSUE #2828: Decide if `UploadError` and `ChecksumMismatch` error should be merged into a single error type.

coryan:
Use a single error type for both I/O and checksum problems in uploads.

Fixes #2828

### DIFF
--- a/src/storage/src/storage.rs
+++ b/src/storage/src/storage.rs
@@ -21,7 +21,7 @@ pub(crate) mod upload_object;
 pub mod upload_source;
 pub(crate) mod v1;
 
-use crate::model::Object;
+use crate::model::{Object, ObjectChecksums};
 use crate::upload_source::Payload;
 use crate::{Error, Result};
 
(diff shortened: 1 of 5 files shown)


src/storage/src/storage.rs
```diff
@@ -49,71 +49,114 @@ use crate::{Error, Result};
...
+
+    /// The MD5 hash verification failed.
+    ///
+    /// # Troubleshoot
+    ///
+    /// The client library compares the MD5 hash of the uploaded data against
+    /// the hash reported by the service at the end of the upload. This error
+    /// indicates the hashes did not match.
+    ///
+    /// If you provided a known value for the the MD5 hash verify the value is
```
dbolduc: s/the the/the/

dbolduc (approved):
I am good with this, but coordinate with @suzmue on how this affects download checksum errors. I assume the error branches should be similar.

src/storage/src/storage.rs
```diff
@@ -49,71 +49,114 @@ use crate::{Error, Result};
...
-/// ```
-///  
-/// # Troubleshooting
-///
-/// Data integrity problems are notoriously difficult to root cause. If you are
-/// using pre-existing, or pre-computed checksum values, you may want to verify
-/// the source data.
-#[derive(thiserror::Error, Debug)]
-#[non_exhaustive]
-pub enum ChecksumMismatch {
```
dbolduc, line 110: #2792 wants to use this.

src/storage/src/storage.rs
coryan: Fixed.

src/storage/src/storage.rs
coryan: Refactored the code to have a common public enum for uploads and downloads.

PR #2822, opened 2025-08-06
docs: error handling full code + cleanups

pcoet:
* Added instructions for formatting code examples to the contributor guide.
* Made a few nitty style fixes, including:
  * Prefer present to future tense.
  * Prefer "you" to "we".
* Added several links to the Secret Manager docs.
* Added cross links to several other guides.
* Added complete `update_secret` function to the start of "Handling the error." (It feels a little easier to follow the explanation if you can see the entire function first.)
* Added a "What's next" section.
* Added complete code for the helper functions, and linked to them.
* Made verbs match operations (*create* a secret vs. *add* a secret version).

### DIFF
--- a/guide/src/error_handling.md
+++ b/guide/src/error_handling.md
@@ -66,11 +66,8 @@ when the container already exists.
 
 ## Handling the error
 
-Here's the code for updating a secret version:
-
-```rust,ignore
-{{#include ../samples/src/error_handling.rs:update-secret}}
-```
+This section walks you through a function to update a secret. For the full code
+sample, see [`update_secret`](#update_secret).
 
 First make an attempt to create a new secret version:
 
@@ -127,7 +124,11 @@ ______________________________________________________________________
 
 ## Code samples
 
-For the complete sample, see [error_handling.rs].
+### `update_secret`
+
+```rust,ignore
+{{#include ../samples/src/error_handling.rs:update-secret}}
+```
 
 ### `update_attempt`
 
@@ -142,7 +143,6 @@ For the complete sample, see [error_handling.rs].
 ```
 
 [configuring retry policies]: /configuring_retry_policies.md
-[error_handling.rs]: https://github.com/pcoet/google-cloud-rust/blob/main/guide/samples/src/error_handling.rs
 [examine error details]: examine_error_details.md
 [handling binding errors]: binding_errors.md
 [quickstart]: https://cloud.google.com/secret-manager/docs/quickstart
--- a/guide/src/error_handling.md
+++ b/guide/src/error_handling.md
@@ -32,9 +32,9 @@ such errors.
 The guide uses the [Secret Manager] service to demonstrate error handling, but
 the concepts apply to other services as well.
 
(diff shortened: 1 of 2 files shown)


guide/src/error_handling.md
```diff
@@ -94,28 +102,52 @@ and then only in errors that correspond to a missing secret:
...
+```
+
+### `create_secret`
+
+```rust,ignore
+{{#include ../samples/src/error_handling.rs:create-secret}}
 ```
 
 [configuring retry policies]: /configuring_retry_policies.md
+[error_handling.rs]: https://github.com/pcoet/google-cloud-rust/blob/main/guide/samples/src/error_handling.rs
```
dbolduc: We definitely should not link to `pcoet`. Linking to `main` (vs. a specific commit) can also be problematic if we rename the file.

We already dump the contents above, I don't think we need to link to the sample code too.

guide/src/error_handling.md
```diff
@@ -65,13 +66,20 @@ when the container already exists.
 
 ## Handling the error
 
+Here's the code for updating a secret version:
+
+```rust,ignore
+{{#include ../samples/src/error_handling.rs:update-secret}}
+```
+
```
dbolduc: We have been putting the full code blocks at the end. I think I would keep doing that.

Maybe at the bottom have an `update_secret` section along with the new `update_attempt`, `create_secret` sections.

Then here we would say something like:

```
We will write a function to update a secret, explaining each line of the
code. The full [`update_secret`](#update_secret) function is included below.

First make an attempt to create a new secret version:

... etc.
```

guide/src/error_handling.md
```diff
@@ -94,28 +102,51 @@ and then only in errors that correspond to a missing secret:
...
+- [Handling binding errors]
+
 ______________________________________________________________________
 
-## Error handling: complete code
+## Code samples
+
+For the complete sample, see [error_handling.rs].
+
+### `update_attempt`
```
dbolduc, line 133: Good catch that these were missing.

guide/src/error_handling.md
```diff
@@ -94,28 +102,51 @@ and then only in errors that correspond to a missing secret:
...
+
+- [Examine error details]
+- [Handling binding errors]
+
 ______________________________________________________________________
 
-## Error handling: complete code
+## Code samples
+
+For the complete sample, see [error_handling.rs].
```
dbolduc: I think this section should be `update-secret`

dbolduc (changes requested):
Thanks!

I do not like the link to `error_handling.rs`. And I suggested a different layout

guide/src/error_handling.md
pcoet: Oops, didn't mean to use my project! Removed entirely.

guide/src/error_handling.md
pcoet: Good feedback. Done. Thanks!

guide/src/error_handling.md
pcoet: Added.

PR #2825, opened 2025-08-07
prototype: Default retry

harkamaljot:


src/auth/src/retry.rs
```diff
@@ -103,31 +152,94 @@ where
...
+            if self.refresh_trigger.is_closed() {
+                return Err(CredentialsError::from_msg(
+                    false,
+                    "token provider background task has been terminated",
+                ));
+            }
+        }
+
+        // Wait for the next value to be published.
+        if rx.changed().await.is_err() {
```
sai-sunder-s, line 230: what if a token was pushed to channel between the time `refresh_trigger` send failed (i.e. there was an already existing retry loop going on) and `rx.changed` is called? The thread will wait until next token is available.

IIRC, the correct way is to first do rx.borrow so that rx knows what your code has already seen. Maybe just do borrow and see if there is new token. You could use entity tag system to know if it is a new token or existing token.

harkamaljot: thanks! yeah in an ideal world token cache shouldn't make multiple calls to get token so this edge case should never get triggered. But yeah it's better to mark it seen so that retry handles this case as well. I updated it to mark the current value as seen.

src/auth/src/retry.rs
```diff
@@ -103,31 +152,90 @@ where
             }
         };
 
-        retry_loop(
+        // The on_retry callback sends intermediate errors and stores the last delay.
+        let last_delay = self.last_delay.clone();
+        let tx_for_retry = self.token_watch_tx.clone();
+        let on_retry = move |_, error: &gax::error::Error, delay: Duration| {
```
sai-sunder-s, line 158: @coryan the API change is in `retry_loop_internal.rs`.  The updated API is used here.

coryan: Ack. What I was suggesting is to create a wrapper in line 180 here, so instead of :

```rust
    self.retry_policy.clone(),
```

You would write:

```
    Arc::new(CaptureRetry::new(self.retry_policy.clone())),
```

where `CaptureRetry` is something like:


```rust
struct CaptureRetry {
  inner: Arc<dyn RetryPolicy>
  stuff: ...
}

impl CaptureRetry {
  fn new(inner: Arc<dyn RetryPolicy>) -> Self { Self { inner, ... } }
}

impl RetryPolicy for CaptureRetry {
  fn on_error(...) {
      // the thing you do in the new callback
  }
}
```

But if you really need something ad-hoc: that is an indication that the retry loop for auth is sufficiently different that maybe we do not want to reuse it.

sai-sunder-s: Thank you @coryan . Yes creating a private wrapper like this makes sense. But I believe we would have to do the wrapper on backoff policy and not retry policy (as explained [here](https://github.com/googleapis/google-cloud-rust/pull/2833#discussion_r2264124013)).

If later we decide that we need info from retry policy as well, we could have wrapper on both retry policy and backoff policy and compose the information for use in auth lib.

If this makes sense, we can close these PRs for now and reopen when we work on them since the solution described here is not changing any public API surface.

coryan: > Thank you @coryan . Yes creating a private wrapper like this makes sense. But I believe we would have to do the wrapper on backoff policy and not retry policy

SGTM. Or even the `sleep` function.... Either way, does not look like you need a custom callback?

 
> If this makes sense, we can close these PRs for now and reopen when we work on them since the solution described here is not changing any public API surface.

What about the changes to `CredentialsError`?

sai-sunder-s: Yes I do not think a custom callback is needed.

`CredentialsError` changes are adding private field and new public methods, not updating existing public method interface, so should be ok?

coryan: Ah, yes, you are right, the field is private, we can change it without breaking things.

sai-sunder-s: Great! Closing the 2 PRs for now. Thank you @coryan !

sai-sunder-s:
Closing for now. We will work on default retries later.

Issue #2824, opened 2025-08-06
Collect data from one billion requests in the w1r3 benchmark


coryan:
We want to get about 1 billion successful requests. This is as much a test of the client library internals as of the default retry policies.

We should use an implementation of the w1r3 benchmark for this, so we can get some data about latency / throughput too.

coryan:
I got a run with 1,000,000 uploads [^1]. A number of threads got "stuck", so we do not really know the error rate.  We are going to need some kind of timeout per operation (maybe at the application / benchmark level?) and then we can estimate the number of timeouts vs. errors vs. successful requests.


[^1]: we want something like 250,000,000 uploads and 750,000,000 downloads, but the uploads had a lot of failures and I wanted to fix those first.

coryan:
Random observations:

- The benchmark requires too much configuration, I had to tweak the retry policy for deletes, the timeout policy for deletes, etc.
- Setting a timeout for each operation will unblock things, and count the errors, but won't be able to root cause the problem.
- The media client is using HTTP 1.1, good for single-stream throughput, bad for socket usage. We need an option to use HTTP/2
- When trying to use a single StorageControl client things get too slow.
- When using a StorageControl client per tasks we get too many sockets.
- When using a StorageControl client per tasks I had to share the credentials or "Bad Things" happened
- We need to retry timeouts or things get stuck. The default policy does not (should not?) retry them.
- We need to use `connect_lazy()` for gRPC clients.
- The `size_hint()` is dropped by how we wrap the stream into a `reqwest::Body` I think that is a missed optimization to avoid chunked transfer encoding.

coryan:
No errors, and good performance. We can close this. I will open a bug to track the changes to the w1r3 benchmark.

----

<img width="1280" height="960" alt="Image" src="https://github.com/user-attachments/assets/5e10022b-3005-4ac1-bbf2-230fc66f0055" />

PR #2829, opened 2025-08-07
feat(w1r3): initial implementation

coryan:
This initial implementation of the `w1r3` benchmark works, but we will
want to optimize it later, and we will also want to get more detailed
information about failures.

Part of the work for #2824

### DIFF
--- a/src/w1r3/src/main.rs
+++ b/src/w1r3/src/main.rs
@@ -1,0 +1,220 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! An implementation of the W1R3 benchmark for Rust.
+//!
+//! The W1R3 benchmark repeatedly uploads an object, then downloads the object
+//! 3 times, and then deletes the object. In each iteration of the benchmark the
+//! size and name of the object is selected at random. The benchmark runs
+//! multiple tasks concurrently.
+
+use clap::Parser;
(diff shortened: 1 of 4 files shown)


src/w1r3/src/main.rs
```diff
@@ -0,0 +1,220 @@
...
+                format!("projects/_/buckets/{}", &args.bucket_name),
+                &name,
+                buffer.slice(0..size),
+            )
+            .send_unbuffered()
+            .await;
+        let upload = match upload {
+            Ok(u) => u,
+            Err(e) => {
+                println!("# Error in upload {e}");
```
dbolduc, line 101: nit: `eprintln` for comments?

src/w1r3/src/main.rs
```diff
@@ -0,0 +1,220 @@
...
+            let mut read = client
+                .read_object(&upload.bucket, &upload.name)
+                .with_generation(upload.generation)
+                .send()
+                .await?;
+            let mut transfer_size = 0;
+            while let Some(b) = read.next().await {
+                if let Ok(b) = b {
+                    transfer_size += b.len();
+                }
```
dbolduc, line 125: Q: should we capture errors here?

I guess it is redundant given `size` vs. `transfer_size`.

src/w1r3/src/main.rs
```diff
@@ -0,0 +1,220 @@
...
+// limitations under the License.
+
+//! An implementation of the W1R3 benchmark for Rust.
+//!
+//! The W1R3 benchmark repeatedly uploads an object, then downloads the object
+//! 3 times, and then deletes the object. In each iteration of the benchmark the
+//! size and name of the object is selected at random. The benchmark runs
+//! multiple tasks concurrently.
+
+use clap::Parser;
```
dbolduc, line 22: TIL. Neat.

src/w1r3/src/main.rs
coryan: Eventually I will send all of these to `stderr` via `tracing::info!()` and `tracing::error!()`.  I will punt for now.

See https://github.com/coryan/google-cloud-rust/blob/impl-storage-introduce-w1r3-benchmark/src/w1r3/src/main.rs if you are interested.

src/w1r3/src/main.rs
coryan: Yeah, eventually this grows into a different error for partial vs. full downloads.

PR #2833, opened 2025-08-07
feat(gax): Add retry_loop_with_on_retry method.

harkamaljot:
The fundamental problem with the original retry_loop is that it completely encapsulates the retry logic, only returning the final outcome. This hides intermediate failures, preventing the end user to see why the current attempt failed and after how long next retry will happen. So adding another method to support this use case.

### DIFF
--- a/src/gax/src/retry_loop_internal.rs
+++ b/src/gax/src/retry_loop_internal.rs
@@ -758,14 +758,13 @@ mod tests {
 
         let inner = async move |d| call.call(d);
         let backoff = async move |d| sleep.sleep(d).await;
-        let response = retry_loop_with_callback(
+        let response = retry_loop(
             inner,
             backoff,
             true,
             to_retry_throttler(throttler),
             to_retry_policy(retry_policy),
             to_backoff_policy(backoff_policy),
-            |_, _, _| (),
         )
         .await;
         let err = response.expect_err("retry loop should terminate");
@@ -879,14 +878,13 @@ mod tests {
 
         let inner = async move |d| call.call(d);
         let backoff = async move |d| sleep.sleep(d).await;
-        let response = retry_loop_with_callback(
+        let response = retry_loop(
             inner,
             backoff,
             true,
             to_retry_throttler(throttler),
             to_retry_policy(retry_policy),
             to_backoff_policy(backoff_policy),
-            |_, _, _| (),
         )
         .await;
         let err = response.expect_err("retry loop should terminate");
--- a/src/gax/src/retry_loop_internal.rs
+++ b/src/gax/src/retry_loop_internal.rs
@@ -39,6 +39,45 @@ impl RetryLoopAttempt {
     }
 }
 
+/// Information about a retry attempt provided to the `on_retry` callback.
+///
+/// An `on_retry` callback can be passed to
+/// [`retry_loop_with_on_retry`][super::retry_loop_with_on_retry]. It is invoked
+/// when a retryable error occurs and the retry policy decides to continue. This
+/// struct contains details about the attempt that just failed, and the upcoming
+/// retry attempt.
+pub struct OnRetryInfo<'a> {
+    attempt: u32,
+    error: &'a Error,
+    delay: Duration,
+}
+
+impl<'a> OnRetryInfo<'a> {
+    /// Creates a new `OnRetryInfo`.
+    pub(crate) fn new(attempt: u32, error: &'a Error, delay: Duration) -> Self {
+        Self {
+            attempt,
+            error,
+            delay,
+        }
+    }
+
+    /// The number of attempts completed so far. The first attempt is 1.
+    pub fn attempt(&self) -> u32 {
+        self.attempt
+    }
+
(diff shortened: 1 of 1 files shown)


src/gax/src/retry_loop_internal.rs
```diff
@@ -39,6 +39,45 @@ impl RetryLoopAttempt {
...
+            delay,
+        }
+    }
+
+    /// The number of attempts completed so far. The first attempt is 1.
+    pub fn attempt(&self) -> u32 {
+        self.attempt
+    }
+
+    /// The error that caused this retry cycle.
```
sai-sunder-s, line 70: Maybe 
```
///The error received in the previous attempt
```

src/gax/src/retry_loop_internal.rs
```diff
@@ -39,6 +39,45 @@ impl RetryLoopAttempt {
...
+    pub fn error(&self) -> &Error {
+        self.error
+    }
+
+    /// The calculated delay before the next attempt is made.
+    pub fn delay(&self) -> Duration {
```
sai-sunder-s, line 76: Like we discussed in the prototype PR, I believe this should be an absolute time.

src/gax/src/retry_loop_internal.rs
```diff
@@ -48,27 +87,65 @@ impl RetryLoopAttempt {
...
+        retry_policy,
+        backoff_policy,
+        |_: &OnRetryInfo| {},
+    )
+    .await
+}
+
+/// Runs the retry loop for a given function with a callback for retries.
+///
+/// This functions calls an inner function as long as (1) the retry policy has
```
sai-sunder-s, line 115: nit: format this to be one point per line

src/gax/src/retry_loop_internal.rs
```diff
@@ -48,27 +87,65 @@ impl RetryLoopAttempt {
...
+    mut on_retry: OnRetry,
 ) -> Result<Response>
 where
     F: AsyncFnMut(Option<Duration>) -> Result<Response> + Send,
     S: AsyncFn(Duration) -> () + Send,
+    OnRetry: FnMut(&OnRetryInfo) + Send,
 {
     let loop_start = tokio::time::Instant::now().into_std();
-    let mut attempt = RetryLoopAttempt::Initial;
+    let mut attempt_state = RetryLoopAttempt::Initial;
```
sai-sunder-s, line 139: why is `attempt` being renamed?

src/gax/src/retry_loop_internal.rs
harkamaljot: Discussed offline the purpose of the callback here is to just share the state and any conversion should be handled by the callers.

src/gax/src/retry_loop_internal.rs
```diff
@@ -48,27 +48,65 @@ impl RetryLoopAttempt {
...
+/// The `on_retry` callback is called before sleeping, with the attempt count,
+/// the error, and the delay.
+pub async fn retry_loop_with_callback<F, S, OnRetry, Response>(
     mut inner: F,
     sleep: S,
     idempotent: bool,
     retry_throttler: Arc<Mutex<dyn RetryThrottler>>,
     retry_policy: Arc<dyn RetryPolicy>,
     backoff_policy: Arc<dyn BackoffPolicy>,
+    mut on_retry: OnRetry,
```
coryan, line 131: Why do you need a `on_retry` function when you already have a retry policy called on each retry attempt?  If you wanted to, you could decorate the existing policy with your own.

sai-sunder-s: I believe @harkamaljot  added `on_retry` as a way to communicate to [auth lib's retry logic](https://github.com/googleapis/google-cloud-rust/pull/2825/files#diff-43612f24f6413f8e5304d6af3e912553d7913ceea55c203c549db570b52d0d30R158) about how long before next attempt. 

I looked at the decorators you suggested. I believe this is the one: https://github.com/googleapis/google-cloud-rust/blob/ba085538c05c2642bd1cf7af7f3a49aa3aa34c41/src/gax/src/retry_policy.rs#L155

Looks like it is the backoff policy that determines the time to wait between attempts though, not the retry policy. We could add a decorator concept for the backoff policy as well, similar to retry policy and have a default backoff policy as well for the auth lib, and always apply the callback decorator to provide the auth lib info on time of next retry. I believe this will not be a breaking change.

Only issue with this approach is that, retry policy also has `remaining_time` logic which can stop a retry from happening and backoff policy has no knowledge about this. So backoff policy might invoke callback with 10s but the retry never happens. Its not the end of the world though, gax will come back in 10s and ask for token and we will return an error then.

Harkamal might have put the `on_retry` callback on `retry_loop` because that is the only method that has knowledge on both the backoff and retry policy's decisions.

If we are ok with the inaccurate callback due to `remaining_time` happening sometimes, adding a decorator to backoff policy will be a good solution.

WDYT?

sai-sunder-s:
Closing for now. We will work on default retries later.

PR #2834, opened 2025-08-08
chore(deps): update tonic,prost to 0.14.x

dbolduc:


src/gax-internal/grpc-server/build.rs
```diff
@@ -15,7 +15,7 @@
 fn main() {
     #[cfg(feature = "_generate-protos")]
     {
-        tonic_build::configure()
+        tonic_prost_build::configure()
```
coryan, line 18: Presumably we need similar changes in `sidekick` so it can call the right version of `Tonic` and `Prost`?

src/gax-internal/grpc-server/src/generated/protos/google.test.v1.rs
```diff
@@ -119,7 +119,7 @@ pub mod echo_service_client {
                         format!("Service was not ready: {}", e.into()),
                     )
                 })?;
-            let codec = tonic::codec::ProstCodec::default();
+            let codec = tonic_prost::ProstCodec::default();
```
coryan, line 122: And presumably the generator would produce different code for all the other files generated by Prost?

coryan (approved):
I am missing some changes in the generator and generated code, I am guessing those will happen "soon"?

src/gax-internal/grpc-server/build.rs
dbolduc: Nope, the generated code (proto messages) only uses `prost_build`, which is unaffected. Only the `grpc-server` (a gRPC server, duh) actually needs `tonic_prost_build`.

src/gax-internal/grpc-server/src/generated/protos/google.test.v1.rs
dbolduc: answered above

src/gax-internal/grpc-server/build.rs
dbolduc: The history is too difficult to recover after the repo shake-up. I think you probably remember making the change to only require `prost` here:

https://github.com/googleapis/librarian/blob/dd70986a49c846fa3a40d7bfa63f712b28a1016a/internal/sidekick/internal/rust_prost/templates/prost/build.rs.mustache#L36

PR #2843, opened 2025-08-08
feat: retries by default

### LINKED ISSUE #518: AIP-4221: Implement Client-side retry
We will need to read some more configuration to support this. I am not sure if we will have the proper information to support in in OpenAPI formats today -- this will need some investigation.

Ref: https://google.aip.dev/client-libraries/4221

- [x] #437 
- [ ] Change generator to provide default retry policies

dbolduc:
Fixes #518  well, sorta.

Every client gets a one-size-fits-all default retry policy. It is very lenient with respect to exhaustion. It has nothing to do with the settings in the service config json.

Implicitly test that we are using `Aip194Strict` as the default policy.

I didn't do any cleanups in the integration-tests.

### DIFF
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -229,7 +229,7 @@ impl Client {
                 Arc::new(
                     Aip194Strict
                         .with_attempt_limit(10)
-                        .with_time_limit(Duration::from_secs(5 * 60)),
+                        .with_time_limit(Duration::from_secs(60)),
                 )
             })
     }
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -166,7 +166,7 @@ impl ReqwestClient {
                 Arc::new(
                     RetryAip194Strict
                         .with_attempt_limit(10)
-                        .with_time_limit(Duration::from_secs(5 * 60)),
+                        .with_time_limit(Duration::from_secs(60)),
                 )
             })
     }


src/gax-internal/src/grpc.rs
```diff
@@ -232,14 +220,18 @@ impl Client {
...
             .retry_policy()
             .clone()
             .or_else(|| self.retry_policy.clone())
+            .unwrap_or_else(|| {
+                Arc::new(
+                    Aip194Strict
+                        .with_attempt_limit(10)
+                        .with_time_limit(Duration::from_secs(5 * 60)),
+                )
+            })
```
coryan: This suggests `self.retry_policy` should not be Option<> and be initialized to a value.

src/gax-internal/src/grpc.rs
```diff
@@ -232,14 +220,18 @@ impl Client {
...
+    fn get_retry_policy(&self, options: &gax::options::RequestOptions) -> Arc<dyn RetryPolicy> {
         options
             .retry_policy()
             .clone()
             .or_else(|| self.retry_policy.clone())
+            .unwrap_or_else(|| {
+                Arc::new(
+                    Aip194Strict
+                        .with_attempt_limit(10)
+                        .with_time_limit(Duration::from_secs(5 * 60)),
```
coryan: I suggest we use a lower value, maybe 60s.  I got complaints about storage being "stuck" with timeouts in that range.  But I am speculating about what will trigger fewer complaints in corner cases, most of the time 60s or 300s will behave exactly the same.

coryan (approved):

src/gax-internal/src/grpc.rs
dbolduc: Agreed. All of the policies are like this, so I am going to leave it as-is in this PR.

I will send a follow up PR to default all of the policies in `*Client::new()` instead of in the helper.

src/gax-internal/src/grpc.rs
dbolduc: Done.

coryan (approved):

src/gax-internal/src/grpc.rs
dbolduc: #2862

PR #2845, opened 2025-08-09
feat(storage): add md5 checksum support for downloads

suzmue:
This implementation mirrors the UploadObject interface. Though I have a slight preference to keep the exported types from being generic since that seems like an implementation detail, I think keeping consistent with upload is best.

For #2049

### DIFF
--- a/src/storage/src/storage/read_object.rs
+++ b/src/storage/src/storage/read_object.rs
@@ -18,7 +18,7 @@ use crate::download_resume_policy::DownloadResumePolicy;
 use crate::model::ObjectChecksums;
 use crate::storage::checksum::{
     ChecksumEngine,
-    details::{Crc32c, validate},
+    details::{Crc32c, Md5, validate},
 };
 use base64::Engine;
 #[cfg(feature = "unstable-stream")]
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/read_object.rs
```diff
@@ -748,22 +803,25 @@ pub struct ObjectHighlights {
     pub etag: String,
 }
 
-fn crc32c_from_response(
+/// checksums_from_response returns the object checksums to validate against,
+/// if it is valid to check them.
```
coryan, line 807: nit: naming the function in the comment is go-like, in Rust I think comments are more is the style:

```suggestion
/// Returns the object checksums to validate against.
///
/// For some responses, the checksums are not expected to match the data. The function returns an
/// empty `ObjectChecksums` in such a case.
```

coryan (approved):
Feel free to fix the comments in a follow up PR.

src/storage/src/storage/read_object.rs
suzmue: Ah TY, will fix in follow up :)

PR #2846, opened 2025-08-09
cleanup(storage)!: reorganize errors and builders

### LINKED ISSUE #2847: Consider adding a `errors` to group all the error types.

coryan:
Move all the errors to a module. Refactor the `KeyAes256` helper to the
builders, but maybe that is not the best place either.

Fixes #2847


src/storage/src/error.rs
```diff
@@ -0,0 +1,273 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Custom errors for the Cloud Storage client.
+//!
+//! The storage client defines additional error types. These are often returned
+//! as the `source()` of a [Error][crate::Error].
```
dbolduc: nit: an Error

src/storage/src/storage/read_object.rs
```diff
@@ -1310,21 +1312,41 @@ mod tests {
...
-    #[test_case(0, -100, RangeError::NegativeLimit; "negative limit")]
-    #[test_case(-100, 100, RangeError::NegativeOffsetWithLimit; "negative offset with positive limit")]
     #[tokio::test]
-    async fn test_range_header_error(offset: i64, limit: i64, want_err: RangeError) -> Result {
+    async fn range_header_negative_limit() -> Result {
         let inner = test_inner_client(test_builder());
         let err = ReadObject::new(inner, "projects/_/buckets/bucket", "object")
-            .with_read_offset(offset)
-            .with_read_limit(limit)
+            .with_read_limit(-100)
```
dbolduc, line 1319: Should this type be a `u64`?

Or is that too annoying bc we need the offset to accept negative values

dbolduc (approved):

src/storage/src/storage/read_object.rs
coryan: See #2849, I think we need to change the input type to something that makes these errors impossible.

dbolduc: Ah, yeah, I like your proposal

src/storage/src/error.rs
coryan: Fixed, PTAL.

PR #2860, opened 2025-08-11
cleanup(auth): unit test MDS assumptions

### LINKED ISSUE #2292: Environment-dependent Auth tests
These tests assume you do not have a MDS. If you run them on a GCE VM, they fail.
- `credentials::mds::test::adc_no_mds`
- `credentials::mds::test::builder_no_mds`

Unit tests should not depend on the environment in this way.

Note that our CI build only `cargo test -p auth-integration-tests`. We do not also `cargo test -p google-cloud-auth`. Given this bug, I think we should run the auth unit tests on GCB as part of the auth integration build.

dbolduc:
Fixes #2292 

The tests assumed there was no MDS present. They fail on a fresh GCE VM.

We fix this case by short circuiting the test if we get a valid token.

Then there is the Googler-only case of running on your Cloudtop, where the default SA doesn't work with the default scopes. In this case, fetching a token returns a `404` which is not transient.

We address this by making less assertions about the error contents.

### DIFF
--- a/src/auth/src/credentials/mds.rs
+++ b/src/auth/src/credentials/mds.rs
@@ -622,20 +622,18 @@ mod tests {
     #[tokio::test]
     #[serial]
     async fn adc_no_mds() -> TestResult {
-        let err = Builder::from_adc()
-            .build_token_provider()
-            .token()
-            .await
-            .unwrap_err();
+        let token = Builder::from_adc().build_token_provider().token().await;
+        let err = match token {
+            Err(e) => e,
+            // The environment has an MDS, skip the test.
+            Ok(_) => return Ok(()),
+        };
 
         let original_err = find_source_error::<CredentialsError>(&err).unwrap();
-        assert!(original_err.is_transient());
         assert!(
             original_err.to_string().contains("application-default"),
             "display={err}, debug={err:?}"
         );
-        let source = find_source_error::<reqwest::Error>(&err);
-        assert!(matches!(source, Some(e) if e.status().is_none()), "{err:?}");
 
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/mds.rs
```diff
@@ -622,20 +622,18 @@ mod tests {
...
-            .build_token_provider()
-            .token()
-            .await
-            .unwrap_err();
+        let token = Builder::from_adc().build_token_provider().token().await;
+        let err = match token {
+            Err(e) => e,
+            // The environment has an MDS, skip the test.
+            Ok(_) => return Ok(()),
+        };
```
coryan: Consider:

```suggestion
        let Err(err) = match Builder::from_adc().build_token_provider().token().await {
            // The environment has an MDS, skip the test.
            Ok(_) => return Ok(()),
        };
```

src/auth/src/credentials/mds.rs
```diff
@@ -668,15 +666,14 @@ mod tests {
     #[tokio::test]
     #[serial]
     async fn builder_no_mds() -> TestResult {
-        let e = Builder::default()
-            .build_token_provider()
-            .token()
-            .await
-            .err()
-            .unwrap();
+        let token = Builder::default().build_token_provider().token().await;
```
coryan: Likewise.

coryan (approved):

src/auth/src/credentials/mds.rs
dbolduc: Done, I think.

src/auth/src/credentials/mds.rs
dbolduc: Done

coryan (approved):

PR #2861, opened 2025-08-11
fix(storage): update doc comment to match rust style

suzmue:


src/storage/src/storage/read_object.rs
```diff
@@ -804,8 +804,10 @@ pub struct ObjectHighlights {
     pub etag: String,
 }
 
-/// checksums_from_response returns the object checksums to validate against,
-/// if it is valid to check them.
+/// Returns the object checksums to validate against.
+///
+/// For some responses, the checksums are not expected to match the data. The function returns an
```
dbolduc: nit: reflow this, I think it is over 80 characters.

coryan (approved):

dbolduc (approved):

PR #2862, opened 2025-08-11
cleanup(gaxi): default policies earlier

dbolduc:


src/gax-internal/src/grpc.rs
```diff
@@ -36,8 +36,8 @@ pub type InnerClient = tonic::client::Grpc<tonic::transport::Channel>;
 pub struct Client {
     inner: InnerClient,
     credentials: Credentials,
-    retry_policy: Option<Arc<dyn RetryPolicy>>,
-    backoff_policy: Option<Arc<dyn BackoffPolicy>>,
+    retry_policy: Arc<dyn RetryPolicy>,
+    backoff_policy: Arc<dyn BackoffPolicy>,
     retry_throttler: SharedRetryThrottler,
 }
```
dbolduc, line 42: observation: grpc client is not saving polling policies. I think we do not respect custom polling policies supplied to the client or supplied via `RequestOptions`

coryan: Yikes! Good catch.

coryan (approved):

PR #2868, opened 2025-08-12
Vertex API - Remove thought signature from content Part's data field

coayscue:
I realize that this is not what you guys will actually land because these changes are to a generated file, but this was what I needed to change to unblock myself. Since I don't know how to generate the file, this was the best I could do.

This is the error we were getting:
```
duplicate field `multiple values for `data`, a oneof with full ID .google.cloud.aiplatform.v1.Part.thought_signature, latest field was thoughtSignature`", line: 1, column: 217
```

I diagnosed the problem to be that the Vertex API definition for Part (https://cloud.google.com/vertex-ai/docs/reference/rest/v1/Content#Part) can contain both a "thoughtSignature" and another element of the "data" enum in it's responses' "candidates.content.parts" field.
Example response we received:
 ```
{
  "candidates": [
    {
      "content": {
        "role": "model",
        "parts": [
          {
            "text": "I need to update the Context API unit tests because the `search()` method has been refactored. This means I'll be mocking new helper",
            "thoughtSignature": "CiUBy73w8KVTix3uJpEWM2fypo7jB3h+jYHFnzk..."
          }
        ]
      }
    }
  ],
  "usageMetadata": {
    "trafficType": "ON_DEMAND"
  },
  "modelVersion": "gemini-2.5-pro",
  "createTime": "2025-08-10T19:34:56.400254Z",
  "responseId": "YPSYaP62GLWfmecPpNyYsQw"
}
```

But the Vertex Rust lib's definition of Part (https://docs.rs/google-cloud-aiplatform-v1/latest/google_cloud_aiplatform_v1/model/struct.Part.html) cannot contain both a "thoughtSignature" and a "data" field (like "text"), because "thoughtSignature" is an element in the data enum, not it's own field, like it is in the API spec.

So when we call serde_json::from_str to parse Gemini's response into the Rust lib's GenerateContentResponse, it throws the "multiple values for 'data'" exception.


dbolduc:
This is interesting. The Rust model is generated off of the protos, which has `thought_signature` as one of the `oneof` fields:

https://github.com/googleapis/googleapis/blob/79a8512f559e497ee47810341cb5d002434316b8/google/cloud/aiplatform/v1/content.proto#L142

So there is a discrepancy in the documentation. Ugh.

I will engage the Vertex AI team. FWIW, I do not think they will be allowed to move the field out of the `oneof`, but maybe they can deprecate it, and offer another top level option.

coryan:
> I realize that this is not what you guys will actually land because these changes are to a generated file, but this was what I needed to change to unblock myself. 

Thanks for the context.  We won't be able to accept this change as-is, but it is clear that there is a bug somewhere from this context.

> Since I don't know how to generate the file, this was the best I could do.

Nor should you need to know how to generate these files.  In this particular case the code is generated from:

https://github.com/googleapis/googleapis/blob/79a8512f559e497ee47810341cb5d002434316b8/google/cloud/aiplatform/v1/content.proto#L137-L142

As best I can tell, the `thought` and `thought_signature` fields should not be part of the `oneof` group.

I am sending a change for review internally, I cannot tell how successfully I will be at getting this approved and submitted.

> 
> I diagnosed the problem to be that the Vertex API definition for Part (https://cloud.google.com/vertex-ai/docs/reference/rest/v1/Content#Part) can contain both a "thoughtSignature" and another element of the "data" enum in it's responses' "candidates.content.parts" field.

Yes, that is because the REST API documentation is generated from a different `.proto` file.  Somehow, either the automation that keeps both in sync did not work, or somebody introduced a bug when updating the public `.proto` from the internal `.proto`.

> But the Vertex Rust lib's definition of Part (https://docs.rs/google-cloud-aiplatform-v1/latest/google_cloud_aiplatform_v1/model/struct.Part.html) cannot contain both a "thoughtSignature" and a "data" field (like "text"), because "thoughtSignature" is an element in the data enum, not it's own field, like it is in the API spec.

Yup. 

I am going to close this PR in the hope that the internal fix will propagate relatively quickly. Separately I will open a tracking bug here so we can continue the conversation.

PR #2871, opened 2025-08-12
feat(storage)!: return object on checksum mismatch

### LINKED ISSUE #2855: Uploads that fail with a checksum mismatch should return the object
When the object is successfully finalized and the checksums fail we should return an error that includes the object metadata.

coryan:
When uploading an object we return an error if the resulting object
checksums do not match the computed or expected checksums. We should
include the object metadata in the error, as the object was created.

Fixes #2855

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4334,7 +4334,7 @@ dependencies = [
 
 [[package]]
 name = "google-cloud-storage"
-version = "0.25.0-preview5"
+version = "0.25.0-preview6"
 dependencies = [
  "anyhow",
  "async-trait",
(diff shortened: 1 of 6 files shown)


dbolduc (dismissed):
optional: could verify `object` in the checksum unit tests: https://github.com/googleapis/google-cloud-rust/blob/f5505ac62247a15b8c28b303a460240d17fb0f2a/src/storage/src/storage/perform_upload/tests/checksums.rs#L42

coryan (commented):
> optional: could verify `object` in the checksum unit tests: 

Done. PTAL

Issue #2055, opened 2025-05-08
Create hand-crafted samples


coryan:
Find out which "snippets" are required, make a list, design how they would work in Rust (file per snippet? function per snippet?) and how do we test them.

My quick search shows that there are bout 100 of these, but could be more.

coryan:
comm -3 <(git  grep -h -o 'START storage_[a-z0-9_]*' -- *.rs  | sort -u) <(git -C ../google-cloud-cpp/ grep -h -o 'START storage_[a-z0-9_]*' -- *.cc | sort -u) | less

BenWhitehead:
### P0
* `storage_quickstart` https://cloud.google.com/storage/docs/reference/libraries#use
* `storage_control_quickstart_sample` https://cloud.google.com/storage/docs/reference/storage_control_quickstart#use_storage_control
* `storage_stream_file_download` https://cloud.google.com/storage/docs/streaming-downloads#stream_a_download
* `storage_stream_file_upload` https://cloud.google.com/storage/docs/streaming-uploads#stream_an_upload
* `storage_list_files` https://cloud.google.com/storage/docs/listing-objects#list-objects
* `storage_set_metadata` https://cloud.google.com/storage/docs/viewing-editing-metadata#edit
* `storage_delete_file` https://cloud.google.com/storage/docs/deleting-objects#delete-object

### P1
* most everything else

### P2
* alternative upload/download samples

### P3
* hamc
* notification config
* service account

coryan:
Thanks, I split the P0 stuff to #2898, I am going to remove this bug from the milestone.

suzmue:
Copy of the list of samples, each has their own sub issue (Hiding this comment immediately after posting)

- [ ] storage_activate_hmac_key
- [ ] storage_add_bucket_default_owner
- [ ] storage_add_bucket_iam_member
- [ ] storage_add_bucket_label
- [ ] storage_add_bucket_owner
- [ ] storage_add_file_owner
- [ ] storage_bucket_delete_default_kms_key
- [ ] storage_bucket_get_default_kms_key
- [ ] storage_change_default_storage_class
- [ ] storage_change_file_storage_class
- [ ] storage_compose_file
- [ ] storage_compose_file_from_many
- [ ] storage_configure_retries
- [ ] storage_copy_file
- [ ] storage_copy_file_archived_generation
- [ ] storage_cors_configuration
- [ ] storage_create_bucket_notifications
[...]

PR #2880, opened 2025-08-13
docs(storage): more examples for buckets

coryan:
Part of the work for #2055

### DIFF
--- a/src/storage/examples/src/change_default_storage_class.rs
+++ b/src/storage/examples/src/change_default_storage_class.rs
@@ -13,7 +13,6 @@
 // limitations under the License.
 
 // [START storage_change_default_storage_class]
-use google_cloud_gax::error::rpc::Code;
 use google_cloud_storage::client::StorageControl;
 use google_cloud_wkt::FieldMask;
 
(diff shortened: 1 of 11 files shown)


src/storage/examples/src/change_default_storage_class.rs
```diff
@@ -0,0 +1,57 @@
...
+    client: &StorageControl,
+    bucket_id: &str,
+) -> anyhow::Result<()> {
+    loop {
+        let mut bucket = client
+            .get_bucket()
+            .set_name(format!("projects/_/buckets/{bucket_id}"))
+            .send()
+            .await?;
+        bucket.storage_class = "NEARLINE".into();
```
dbolduc: nit: use the setter?

src/storage/examples/src/change_default_storage_class.rs
```diff
@@ -0,0 +1,57 @@
...
+            Ok(b) => {
+                println!("successfully updated bucket {b:?}");
+                break;
+            }
+            Err(e) => {
+                if e.status()
+                    .map(|s| s.code)
+                    .is_some_and(|code| code == Code::FailedPrecondition)
+                {
+                    continue;
```
dbolduc: Why the loop?

If there is a loop, shouldn't it have backoffs and a termination condition?

src/storage/examples/src/lib.rs
```diff
@@ -12,16 +12,69 @@
...
+pub mod get_bucket_metadata;
 pub mod list_buckets;
 use google_cloud_storage::client::StorageControl;
 use rand::{Rng, distr::Distribution};
 
 pub const BUCKET_ID_LENGTH: usize = 63;
 
-pub async fn run_bucket_examples(_bucket_id: &str) -> anyhow::Result<()> {
+pub async fn run_bucket_examples(buckets: &mut Vec<String>) -> anyhow::Result<()> {
+    let _guard = {
```
dbolduc, line 29: optional: typically we gate this behind a logging feature so it can be turned off

src/storage/examples/tests/integration.rs
```diff
@@ -15,16 +15,18 @@
...
-    use storage_samples::{cleanup_bucket, random_bucket_id, run_bucket_examples};
+    use storage_samples::{cleanup_bucket, run_bucket_examples};
 
     #[tokio::test]
     async fn run_all_examples() -> anyhow::Result<()> {
         let client = StorageControl::builder().build().await?;
 
-        let bucket_id = random_bucket_id();
-        let result = run_bucket_examples(&bucket_id).await;
+        let mut buckets = Vec::new();
```
dbolduc, line 24: Ah, clever. Lets us use `?` with less chance of leaking resources.

src/storage/examples/src/change_default_storage_class.rs
coryan: Done.

src/storage/examples/src/change_default_storage_class.rs
coryan: You do a loop because there may be a concurrent change, some other process or task may change the bucket at the same time and you want to try again.

But it complicates the example too much, removed.

FWIW, the expected problem is not a transient error, it is a concurrent change. A backoff won't help with those.

src/storage/examples/src/lib.rs
coryan: Yes. I think we will want to see the basic logs to troubleshoot failures, and we can change our minds later.

Issue #2896, opened 2025-08-15
Remove the `C` generic parameter for the request builders


coryan:

coryan:
Looks like this is fixed.

PR #2907, opened 2025-08-16
fix(storage)!: remove ChecksumEngine generic trait from the public API

suzmue:
Remove the generic parameters for checksums from the public API. This does open the API to misuse a little easier, since we no longer can limit which setters for the checksum are enabled based on the type.

This is still a WIP, needs a bit of cleanup. 

For #2896

### DIFF
--- a/src/storage/src/storage/checksum/details.rs
+++ b/src/storage/src/storage/checksum/details.rs
@@ -61,6 +61,113 @@ pub fn validate(
     }
 }
 
+#[derive(Clone, Debug)]
+pub enum ChecksumEnum {
+    Null(Null),
+    Md5(Md5),
+
+    KnownCrc32c(KnownCrc32c),
+    KnownMd5(KnownMd5),
+
+    Crc32cKnownMd5(Crc32c<KnownMd5>),
+    Md5KnownCrc32c(Md5<KnownCrc32c>),
+    Known(Known),
+
+    Crc32c(Crc32c),
+    Both(Crc32c<Md5>),
+}
+
+impl sealed::ChecksumEngine for ChecksumEnum {}
+
+impl std::default::Default for ChecksumEnum {
+    fn default() -> Self {
+        ChecksumEnum::Crc32c(Crc32c::default())
+    }
+}
+
+impl ChecksumEnum {
+    pub(crate) fn compute_md5(self) -> Self {
+        match self {
+            // If CRC32C is known, we want to keep it known and compute MD5.
+            ChecksumEnum::KnownCrc32c(_)
+            | ChecksumEnum::Known(_)
+            | ChecksumEnum::Md5KnownCrc32c(_) => {
+                ChecksumEnum::Md5KnownCrc32c(Md5::from_inner(KnownCrc32c))
+            }
+
+            // If we are not computing CRC32C, just compute MD5.
+            ChecksumEnum::Null(_) | ChecksumEnum::KnownMd5(_) | ChecksumEnum::Md5(_) => {
+                ChecksumEnum::Md5(Md5::default())
+            }
+
+            // If we are already computing CRC32C, compute both.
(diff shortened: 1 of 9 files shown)


src/storage/src/storage/checksum/details.rs
```diff
@@ -61,6 +61,113 @@ pub fn validate(
     }
 }
 
+#[derive(Clone, Debug)]
+pub enum ChecksumEnum {
+    Null(Null),
```
coryan: Consider just keeping the enum as a symbolic name in `UploadObject` builder and then use it to create the right engine when creating the `PerformUpload` helper.

coryan (commented):
This will probably conflict with #2906 , but the fixes will be (I hope) easy

src/storage/src/storage/checksum/details.rs
suzmue: Updated implementation to @dbolduc's suggestion from today's sync.

src/storage/src/storage/write_object.rs
```diff
@@ -852,8 +834,18 @@ impl<T> WriteObject<T, Crc32c> {
     /// use [compute_md5()] to also have the library compute the checksums.
     ///
     /// [compute_md5()]: WriteObject::compute_md5
-    pub fn with_known_crc32c<V: Into<u32>>(self, v: V) -> WriteObject<T, KnownCrc32c> {
-        let this = self.switch_checksum(|_| KnownCrc32c);
+    pub fn with_known_crc32c<V: Into<u32>>(self, v: V) -> Self {
```
dbolduc, line 837: consider spelling out `set_crc32c()` in here:

```rs
pub fn with_known_crc32c<V: Into<u32>>(self, v: V) -> Self {
  let checksum = self.mut_resource().checksums.get_or_insert_default();
  checksum.crc32c = Some(v.into());
  self.checksum.crc32c = None;
  self
}
```

src/storage/src/storage/write_object.rs
```diff
@@ -886,12 +878,22 @@ impl<T> WriteObject<T, Crc32c> {
     /// use [compute_md5()] to also have the library compute the checksums.
     ///
     /// [compute_md5()]: WriteObject::compute_md5
-    pub fn with_known_md5_hash<I, V>(self, i: I) -> WriteObject<T, Crc32c<KnownMd5>>
+    pub fn with_known_md5_hash<I, V>(self, i: I) -> Self
```
dbolduc, line 881: same

dbolduc (approved):
This is awesome! Thanks!

I just have an optional nit that could save some lines of code.

src/storage/src/storage/write_object.rs
dbolduc: Actually `self` would need to be `mut self`.

We could say:

```rs
let mut this = self;
let checksum = this.mut_resource().checksums.get_or_insert_default();
checksum.crc32c = Some(v.into());
this.checksum.crc32c = None;
this
```

src/storage/src/storage/write_object.rs
```diff
@@ -914,67 +916,23 @@ impl<T> WriteObject<T, Crc32c> {
...
-    {
-        let this = self.switch_checksum(|_| Known);
-        this.set_md5_hash(i)
-    }
-
-    /// See [WriteObject<T, Crc32c>::compute_md5()].
-    pub fn compute_md5(self) -> WriteObject<T, Md5<KnownCrc32c>> {
-        self.switch_checksum(Md5::from_inner)
+    pub fn compute_md5(self) -> Self {
+        Self {
```
dbolduc, line 920: nit: I think we could also do the thing here:

```rs
let mut this = self;
this.checksum.md5_hash = Some(Md5::default());
this
```

suzmue:
@dbolduc I'll submit as is to make your work easier, and implement the nits in a follow up.

Issue #2909, opened 2025-08-16
Missing `content-length` header reported by Google Frontend


coryan:

coryan:
I think we can close this, the problem did not repeat in 1,000,000,000 uploads.

PR #2910, opened 2025-08-16
fix(storage): missing header when resuming uploads

coryan:
The `content-length` header was missing query querying the upload
status.

Part of the work for #2909


dbolduc:
Nice

PR #2911, opened 2025-08-17
feat(w1r3): add metrics and counters

coryan:
These metrics and counters were useful in troubleshooting problems with
the storage library.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6606,6 +6606,7 @@ dependencies = [
  "parse-size",
  "rand 0.9.2",
  "tokio",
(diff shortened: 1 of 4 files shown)


.codecov.yml
```diff
@@ -25,6 +25,7 @@ ignore:
   - "src/gax-internal/echo-server"
   - "src/gax-internal/grpc-server"
   - "tools/check-copyright"
+  - "src/w1r3"
```
dbolduc: nit: this could be more alphabetical

coryan: Fixed.

PR #2912, opened 2025-08-18
refactor(storage)!: move `ObjectHighlights`

### LINKED ISSUE #2890: Move `ObjectHighlights` to `model` module

coryan:
Fixes #2890


src/storage/src/model.rs
```diff
@@ -22,3 +22,66 @@ pub use crate::control::model::*;
...
 pub mod request_helpers;
+
+/// Define types related to the `v1` JSON protocol.
+///
+/// The client library uses the JSON protocol in its implementation. For the
+/// most, the types are mapped to the v2 protocol (which is gRPC and Protobuf
+/// based). Where no accurate mapping to a type in v2 is possible, the client
+/// library returns a type in this module.
+///
+/// We anticipate no breaking changes when switching to the v2 protocol.
```
dbolduc: Hmm.. The client library is abstracting away the `v1` vs. `v2` distinction, right? So I don't like that we are talking about it.

I see that `ObjectHighlights` doesn't belong in `request_helpers` because it is response-only.

---

What do you think about:

```rs
/// The messages and enums that are part of this client library.
pub mod model;

/// More messages and enums that are part of this client library.
///
/// This module is kept separate from `model` to minimize library maintenance
/// costs. Applications should think of `model_extras` and `model` as the same,
/// conceptually.
pub mod model_extras;
```

... instead of nesting `request_helpers`, `v1`.

We would have 0 chances of conflicts, and I feel like it's pretty clear.

src/storage/src/model.rs
```diff
@@ -22,3 +22,66 @@ pub use crate::control::model::*;
 // with nested enums or messages. If we ever get a conflict, we would configure
 // sidekick to rename the generated types.
 pub mod request_helpers;
+
+/// Define types related to the `v1` JSON protocol.
+///
+/// The client library uses the JSON protocol in its implementation. For the
+/// most, the types are mapped to the v2 protocol (which is gRPC and Protobuf
```
dbolduc: nit: For the most part,

src/storage/src/model.rs
coryan: Cannot say that I was in love with `v1` either.  What about `response_helpers` ?

coryan: Inside `model` I mean.  `model_extras` seems bad, if we are going to do that, I vote we move `response_helpers` and `request_helpers` out of `model` altogether.

dbolduc: To be clear, I am suggesting:

```rs
pub mod model_extras {

  pub struct ObjectHighlights { ... }

  pub struct KeyAes256 { ... }

  // etc.
}
```

src/storage/src/model.rs
coryan: Removed

src/storage/src/model.rs
coryan: Moved to `model_ext`

PR #2915, opened 2025-08-18
fix(gax-internal): lazy connects for gRPC clients

### LINKED ISSUE #2914: Change gRPC clients to use `connect_lazy()`
Otherwise creating a lot of clients at the same time fails.

coryan:
Eagerly connecting to the client results in transient errors, and we
don't have a retry loop in the client creation. Instead of adding
another retry loop (and another set of configurations), we delay the
connection until the first request.

Fixes #2914

### DIFF
--- a/src/gax-internal/tests/grpc_timeout.rs
+++ b/src/gax-internal/tests/grpc_timeout.rs
@@ -201,19 +201,17 @@ mod tests {
             .await
     }
 
-    async fn connect_client(client: grpc::Client, server: tokio::task::JoinHandle<()>) -> anyhow::Result<std::pin::Pin<Box<tokio::task::JoinHandle<()>>>> {
+    async fn connect_client(
+        client: grpc::Client,
+        server: tokio::task::JoinHandle<()>,
+    ) -> anyhow::Result<std::pin::Pin<Box<tokio::task::JoinHandle<()>>>> {
         // Make sure the client is connected, the first request has higher
         // latency. It is hard to write a timeout test that includes the request.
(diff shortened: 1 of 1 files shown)


src/gax-internal/tests/grpc_timeout.rs
```diff
@@ -84,16 +84,38 @@ mod tests {
...
         let delay = Duration::from_millis(200);
-        let timeout = Duration::from_millis(150);
         let mut interval = tokio::time::interval(Duration::from_millis(10));
+
+        let request_options = RequestOptions::default();
+        let response = send_request(
+            client.clone(),
+            request_options,
+            "great success!",
+            Some(delay),
```
dbolduc: why have any delay?

src/gax-internal/tests/grpc_timeout.rs
```diff
@@ -84,16 +84,38 @@ mod tests {
...
+        );
+        tokio::pin!(response);
+        loop {
+            tokio::select! {
```
dbolduc: Q: I haven't tried this, but can we build the connection / initial request into `test_client()`?

so all the tests in this fixture start from the same place, and we can hide the distracting set up.

src/gax-internal/tests/grpc_timeout.rs
coryan: Cut&paste, I do not know what I am doing. Fixed.

src/gax-internal/tests/grpc_timeout.rs
coryan: Refactored to a single place.

PR #2924, opened 2025-08-19
cleanup(storage)!: ranged reads made type-safe

### LINKED ISSUE #2849: Should we use an enum for the read ranges?
Currently we validate the `offset` and `limit` values at runtime. The API allows one to set negative limits, which cannot be formatted as HTTP requests (and are also invalid in gRPC). The API also allows one to set a negative offset with a limit, which cannot be formatted as HTTP either.  In other words, these calls return in errors:

```rust
  let client: google_cloud_storage::client::Storage = ...;

  let reader = client.read_object(...).with_limit(-100).send().await.unwrap(); // panics
  let reader = client.read_object(...).with_offset(-100).with_limit(100).send().await.unwrap(); // panics
```

We should consider a different API, where it is impossible to make bad requests, e.g.:

```rust
  let client: google_cloud_storage::client::Storage = ...;

  let reader = client.read_object(...).with_range(Range::tail(100_u64)).send().await?;
  let reader = client.read_object(...).with_range(Range::head(100_u64)).send().await?;
  let reader = client.read_object(...).with_range(Range::new(1000_u64, limit: 100_u64)).send().await?;
```

That is, we would accept an enum that makes it impossible to create bad ranges:

```rust
pub struct Range(Impl);
impl Range {
  pub fn head(limit: u64) -> Self {}
  pub fn tail(limit: u64) -> Self {}
  pub fn new(offset: u64, limit: u64) -> Self {}
}

pub(crate) struct Impl {
  offset: i64,
  limit: i64,
}
```

We need to think about how to handle the case where the values are outside the `i64` range (panic? return `Result<Self, ..>` from `new()` and friends?).

We *may* want to consider limiting these ranges to the maximum size of a Cloud Storage object (5TiB), but that may break if the service was to increase the limit (something that is extremely unlikely).

coryan:
Specify ranged reads using a safe type, which does not permit invalid
ranges (such as negative length limits). No need to have a `ReadError`
if the request is well-defined by construction.

Fixes #2849


src/storage/src/model_ext.rs
```diff
@@ -131,9 +131,165 @@ impl std::convert::From<KeyAes256> for crate::model::CommonObjectRequestParams {
...
+            limit: count,
+        })
+    }
+}
+
+impl crate::model::ReadObjectRequest {
+    pub(crate) fn with_range(&mut self, range: ReadRange) {
+        // The limit for GCS objects is (currently) 5TiB, and the gRPC protocol
+        // uses i64 for the offset and limit. Clamping the values to the
+        // `[0, i64::MAX]`` range is safe, in that it does not loose any
```
dbolduc: nit: s/loose/lose/

src/storage/src/model_ext.rs
```diff
@@ -131,9 +131,165 @@ impl std::convert::From<KeyAes256> for crate::model::CommonObjectRequestParams {
...
+    /// # Ok(()) }
+    pub fn segment(offset: u64, count: u64) -> Self {
+        Self(Range::Segment {
+            offset,
+            limit: count,
+        })
+    }
+}
+
+impl crate::model::ReadObjectRequest {
```
dbolduc, line 254: interesting

src/storage/src/storage/read_object.rs
```diff
@@ -189,9 +189,11 @@ where
         self
     }
 
-    /// The offset for the first byte to return in the read, relative to
-    /// the start of the object.
+    /// The range of bytes to return in the read.
     ///
+    /// This can be all the bytes starting at a give offset (`ReadRange::offset()`),
```
dbolduc: s/give/given/

src/storage/src/storage/read_object.rs
```diff
@@ -189,9 +189,11 @@ where
...
+    /// all the bytes in a explicit range (`Range::segment`), the last N bytes
```
dbolduc: s/a explicit/an explicit/

src/storage/src/storage/read_object.rs
```diff
@@ -189,9 +189,11 @@ where
...
+    /// of the object (`ReadRange::tail`)
     /// A negative `read_offset` value will be interpreted as the number of bytes
     /// back from the end of the object to be returned.
```
dbolduc: Stale, remove

src/storage/src/storage/read_object.rs
dbolduc: I think you want an "or" and to end the sentence with a period

src/storage/src/storage/read_object.rs
```diff
@@ -521,19 +478,22 @@ where
...
             // read_limit can't be negative.
-            (_, l) if l < 0 => Err(RangeError::NegativeLimit),
+            (_, l) if l < 0 => {
+                unreachable!("ReadObject build never sets a negative read_limit value")
+            }
             // negative offset can't also have a read_limit.
-            (o, l) if o < 0 && l > 0 => Err(RangeError::NegativeOffsetWithLimit),
+            (o, l) if o < 0 && l > 0 => unreachable!(
+                "ReadObject builder never sets a positive read_offset value with a negative read_limit value"
+            ),
```
dbolduc, line 486: We can just delete these, right?

src/storage/src/storage/read_object.rs
```diff
@@ -1392,17 +1352,16 @@ mod tests {
...
     }
 
-    #[test_case(0, 0, None; "no headers needed")]
-    #[test_case(10, 0, Some(&http::HeaderValue::from_static("bytes=10-")); "offset only")]
-    #[test_case(-2000, 0, Some(&http::HeaderValue::from_static("bytes=-2000-")); "negative offset")]
-    #[test_case(0, 100, Some(&http::HeaderValue::from_static("bytes=0-99")); "limit only")]
-    #[test_case(1000, 100, Some(&http::HeaderValue::from_static("bytes=1000-1099")); "offset and limit")]
+    #[test_case(ReadRange::all(), None; "no headers needed")]
+    #[test_case(ReadRange::offset(10), Some(&http::HeaderValue::from_static("bytes=10-")); "offset only")]
+    #[test_case(ReadRange::tail(2000), Some(&http::HeaderValue::from_static("bytes=-2000-")); "negative offset")]
```
dbolduc: I do not know about this trailing hyphen

https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Range

src/storage/src/storage/read_object.rs
```diff
@@ -521,19 +478,22 @@ where
...
+            (o, l) if o < 0 && l > 0 => unreachable!(
+                "ReadObject builder never sets a positive read_offset value with a negative read_limit value"
+            ),
             // If both are zero, we use default implementation (no range header).
-            (0, 0) => Ok(builder),
+            (0, 0) => builder,
             // read_limit is zero, means no limit. Read from offset to end of file.
             // This handles cases like (5, 0) -> "bytes=5-"
-            (o, 0) => Ok(builder.header("range", format!("bytes={o}-"))),
+            (o, 0) => builder.header("range", format!("bytes={o}-")),
```
dbolduc, line 493: Should we have an `o < 0` case without the trailing hyphen?

src/storage/src/storage/read_object.rs
dbolduc: I do not see an integration test that would tell us it is ok

src/storage/src/model_ext.rs
coryan: Fixed.

src/storage/src/storage/read_object.rs
coryan: Fixed and Fixed.

src/storage/src/storage/read_object.rs

src/storage/src/storage/read_object.rs
coryan: Done.

src/storage/src/storage/read_object.rs
coryan: We want a crash if they are reached, and this documents the assumptions in the other cases.  I would just leave them.

src/storage/src/storage/read_object.rs
coryan: Yes! Good catch.

src/storage/src/storage/read_object.rs

coryan: And created new integration test.

dbolduc (approved):

PR #2928, opened 2025-08-19
Edits to the binding errors guide

pcoet:
Made a few edits to the binding errors guide:

* Added context in a couple of places.
* Made a few changes to improve readability.
* Made a couple of nitty copy edits.

I didn't add a "What's next" section because this feels like a different kind of content from the tutorials. Maybe at some point we could have a "Troubleshooting" section of the docs that includes this and other troubleshooting guides. There's a semi-consistent, implicit ordering to the site right now (getting started, using clients, handling errors, testing). If the docs get big enough, it might make sense to start making this organization explicit. Just thoughts!

### DIFF
--- a/guide/src/binding_errors.md
+++ b/guide/src/binding_errors.md
@@ -16,7 +16,8 @@ limitations under the License.
 
 # Handling binding errors
 
-You might have tried to make a request and run into an error that looks like:
+You might have tried to make a request and run into an error that looks like
+this:
 
 ```norust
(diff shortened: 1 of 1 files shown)


coryan (approved):

coryan:
Thanks! You may want to look at https://www.conventionalcommits.org/en/v1.0.0/ we require the PR title to follow that format. It makes it easier to summarize the changes into a change log later.

PR #2930, opened 2025-08-19
impl(w1r3): batch delete objects in each task

coryan:
Motivated by #2824


Cargo.toml
```diff
@@ -347,6 +347,7 @@ tokio-test     = { default-features = false, version = "0.4" }
 
 # Local packages used as dependencies
 auth                          = { version = "0.22.3", path = "src/auth", package = "google-cloud-auth" }
+google-cloud-auth             = { version = "0.22.3", path = "src/auth", package = "google-cloud-auth" }
```
dbolduc, line 350: Fine, but don't love the inconsistency. Maybe we should use the full names everywhere.

PR #2942, opened 2025-08-20
refactor(storage)!: relocate `ReadObjectResponse`

### LINKED ISSUE #2889: Move `ReadObjectResponse` to its own module

dbolduc:
Fixes #2889 

I am open to a shorter module name. Just `read_object`?


suzmue:
I see the issue says to move it to its own module, but I'm wondering if this would be appropriate to include in model_ext instead to have less places for looking for types.

> I am open to a shorter module name. Just read_object?
+1 for just read_object

coryan:
> I see the issue says to move it to its own module, but I'm wondering if this would be appropriate to include in model_ext instead to have less places for looking for types.

We have kept traits outside `model` and `model_ext`. Maybe we shouldn't....

> 
> > I am open to a shorter module name. Just read_object?
> > +1 for just read_object

Meeee toooo.  I am stuck in meetings, please make a decision and let me know.

suzmue:
> We have kept traits outside model and model_ext. Maybe we shouldn't....

Ok, sticking with our norms SGTM. Rename to read_object and it LGTM

dbolduc:
> Rename to `read_object`

Done.

I think `model_ext` also made sense as it's kind of a return type.

PR #2949, opened 2025-08-22
fix(storage): more retryable errors

coryan:
gRPC reports some I/O errors as `is_transport()` (because they may
include headers, whereas `is_io()` does not).  We need to retry them.

Over gRPC, Cloud Storage reports some internal as `DeadlineExceeded`,
need to retry those too.

And finally, timeouts are common enough that need to be retried.

Motivated by #2824


src/storage/src/retry_policy.rs
```diff
@@ -88,7 +88,13 @@ impl RetryPolicy for RetryableErrors {
         if !idempotent {
             return RetryResult::Permanent(error);
         }
-        if error.is_io() {
+        if error.is_io() || error.is_timeout() {
+            return RetryResult::Continue(error);
+        }
+        if error.is_transport() && error.http_status_code().is_none() {
+            // Sometimes gRPC returns a transport error without an HTTP status
+            // code. We treat all of these are I/O errors and therefore
```
dbolduc: s/these are/these as/

dbolduc (approved):

src/storage/src/retry_policy.rs
coryan: Fixed.

dbolduc (approved):

Issue #2495, opened 2025-06-23
Track number of throttled attempts in retry loop


coryan:
Oh wait, nothing is tracking the # of throttles per client call. Hmm... Should that be an argument to `on_throttle`?

_Originally posted by @dbolduc in https://github.com/googleapis/google-cloud-rust/pull/2456#discussion_r2161815981_

PR #3007, opened 2025-08-23
cleanup(gax)!: group retry policy arguments

coryan:
Use a struct for the retry policy arguments. This will make it possible
to add new fields to the struct without breaking any existing policies,
even those written by customers.

This could help with #2495, and it is a pre-requisite for making the retry policy APIs public.

### DIFF
--- a/src/gax/src/retry_state.rs
+++ b/src/gax/src/retry_state.rs
@@ -67,7 +67,7 @@ impl std::default::Default for RetryState {
         Self {
             start: Instant::now(),
             idempotent: false,
-            attempt_count: 0
+            attempt_count: 0,
         }
     }
 }


src/gax/src/retry_policy.rs
```diff
@@ -67,21 +68,10 @@ pub trait RetryPolicy: Send + Sync + std::fmt::Debug {
     /// Query the retry policy after an error.
     ///
     /// # Parameters
-    /// * `loop_start` - when the retry loop started.
-    /// * `attempt_count` - the number of attempts. This includes the initial
-    ///   attempt. This method called after the first attempt, so the
-    ///   value is always non-zero.
-    /// * `idempotent` - if `true` assume the operation is idempotent. Many more
-    ///   errors are retryable on idempotent operations.
+    /// * `_state` - the state of the retry loop.
```
dbolduc: nit: s/_state/state/

src/gax/src/retry_state.rs
```diff
@@ -0,0 +1,61 @@
...
+    /// The start time for this retry loop.
+    pub start: Instant,
+
+    /// The number of times the read request has been interrupted already.
+    pub attempt_count: u32,
+}
+
+impl RetryState {
+    /// Create a new instance.
+    pub fn new(idempotent: bool) -> Self {
```
dbolduc, line 42: comment: this cant grow

Maybe we should have a setter for `idempotent` and `impl Default`?

dbolduc (commented):
LGTM, I think this needs a rebase though to pick up the retry policy added to w1r3

src/gax/src/retry_state.rs
coryan: Done. I am going to leave this constructor, it is a very useful thing.

src/gax/src/retry_policy.rs
coryan: Fixed.

coryan (commented):
PTAL.

PR #3033, opened 2025-08-25
impl(w1r3): deletes can be disabled

coryan:
This can be useful to speed up the benchmark, at the cost of leaving
garbage files in the test bucket.

Motivated by #2824

### DIFF
--- a/src/w1r3/src/main.rs
+++ b/src/w1r3/src/main.rs
@@ -204,7 +204,9 @@ async fn runner(
             };
             let _ = task.tx.send(sample).await;
         }
-        deletes.push(delete(&control, &args, upload));
+        if !args.no_delete {
+            deletes.push(delete(&control, &args, upload));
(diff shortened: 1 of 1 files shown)


src/w1r3/src/main.rs
```diff
@@ -702,6 +704,10 @@ struct Args {
     #[arg(long, default_value_t = 3)]
     read_count: i32,
 
+    /// Skip the deletes.
+    #[arg(long)]
+    no_delete: bool,
```
dbolduc, line 709: consider calling this `delete` with a default value of `true`.

dbolduc (approved):

src/w1r3/src/main.rs
coryan: I could not figure out how to get `--delete=false` to work. When we do, we can change it.

PR #3124, opened 2025-08-25
feat(storage): better retry policy defaults

coryan:
These values are based on the experiment running one billion requests successfully.


dbolduc (approved):
Thanks!

PR #3132, opened 2025-08-26
fix(storage): correct mixin service name

### LINKED ISSUE #3125: Generated code for LROs uses incorrect service
When generating mixins for LROs the gRPC service name and RPC name should be the names of the original service, and not the name of the target service.

For example:

```patch
         let extensions = {
             let mut e = tonic::Extensions::new();
             e.insert(tonic::GrpcMethod::new(
-                "google.storage.control.v2.StorageControl",
+                "google.longrunning.Operations",
                 "GetOperation",
             ));
             e
         };
         let path = http::uri::PathAndQuery::from_static(
-            "/google.storage.control.v2.StorageControl/GetOperation",
+            "/google.longrunning.Operations/GetOperation",
         );
```

coryan:
Fixes #3125. Basically we needed to fix `sidekick` and regenerate the code.

### DIFF
--- a/.github/workflows/generator_dart.yaml
+++ b/.github/workflows/generator_dart.yaml
@@ -50,7 +50,7 @@ jobs:
       # the inputs should be pinned, including the generator version and the
       # googleapis SHA.
       - name: Regenerate existing clients
-        run: go run github.com/googleapis/librarian/cmd/sidekick@v0.1.2-0.20250812135937-eb593463c88c refreshall -project-root dart
+        run: go run github.com/googleapis/librarian/cmd/sidekick@v0.1.2-0.20250826011555-c0c043b7a0a6 refreshall -project-root dart
       - run: git diff --exit-code
 
   validate-packages:
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -401,7 +401,7 @@ jobs:
           sudo unzip -x /tmp/protoc.zip
           protoc --version
       - name: Regenerate all the code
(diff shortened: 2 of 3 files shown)


dbolduc (approved):

.github/workflows/generator_dart.yaml
```diff
@@ -50,7 +50,7 @@ jobs:
       # the inputs should be pinned, including the generator version and the
       # googleapis SHA.
       - name: Regenerate existing clients
-        run: go run github.com/googleapis/librarian/cmd/sidekick@v0.1.2-0.20250812135937-eb593463c88c refreshall -project-root dart
+        run: go run github.com/googleapis/librarian/cmd/sidekick@v0.1.2-0.20250826011555-c0c043b7a0a6 refreshall -project-root dart
```
devoncarew, line 53: Given a commit in googleapis/librarian, how do I know the right reference value to use here (i.e., `20250826011555-c0c043b7a0a6`)?

Also, do we want to keep the two librarian repo references here - in generator_dart.yaml and sdk.yaml - in sync?

coryan: > Given a commit in googleapis/librarian, how do I know the right reference value to use here (i.e., `20250826011555-c0c043b7a0a6`)?

That we just figured out today:

https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-generated-code-maintenance.md#special-cases

> Also, do we want to keep the two librarian repo references here - in generator_dart.yaml and sdk.yaml - in sync?

Good question ... let's see how it goes by trying to keep them in sync. If it becomes a problem we can change our minds.

Issue #2041, opened 2025-05-08
Add stubs and tracing to `storage::Storage`


coryan:
We need to structure the client to support tracing. Keep in mind that streaming reads and writes will need long-lived spans, so the instrumentation may be more manual.

PR #3137, opened 2025-08-26
refactor(storage)!: `ReadObjectResponse` uses pimpl

dbolduc:
Part of the work for #2041, reverts #2931 

Restore `ReadObjectResponse` to be a concrete object.

Implement it using a pimpl pattern. This will let us inject a fake implementation when we need to mock it as a return type.


dbolduc:
Had to rebase, PTAL

Issue #1342, opened 2025-02-17
Add support for Workload Identity Federation


moricho:
## Background
Currently, the auth library supports three authentication methods:
- User credentials (authorized_user)
- Service account key credentials (service_account)
- Metadata service credentials (for GCE/GKE)

However, it lacks support for Workload Identity Federation (WIF), which is Google Cloud's recommended way to access Google Cloud resources from non-Google Cloud environments (Github Actions, AWS, Azure, etc.) without using service account keys.

## Proposed Changes
Add support for Workload Identity Federation by:

1. Implementing external account credentials:
   - Add support for `type: "external_account"` in ADC files
   - Implement STS (Security Token Service) token exchange
   - Support credential source configurations for different providers

2. Supporting some providers:
   - Generic OIDC provider
   - AWS
   - Azure

## Example Configuration
```json
{
    "type": "external_account",
    "audience": "//iam.googleapis.com/projects/PROJECT_NUMBER/locations/global/workloadIdentityPools/POOL_ID/providers/PROVIDER_ID",
    "subject_token_type": "urn:ietf:params:oauth:token-type:jwt",
    "token_url":"https://sts.googleapis.com/v1/token",
    "credential_source": {
        "url": "***",
        "headers": {"Authorization":"***"},
        "format":{"type":"json","subject_token_field_name":"value"}
    },
    "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/***:generateAccessToken"
}
```

coryan:
> However, it lacks support for Workload Identity Federation, which is Google Cloud's recommended way to access Google Cloud resources from non-Google Cloud environments (Github Actions, AWS, Azure, etc.) without using service account keys.

We are aware and it is part of our roadmap.  Unfortunately, we are not at liberty of sharing timelines.

PR #3142, opened 2025-08-26
fix: workload identity federation service account impersonation

fujin:
Howdy folks, I came across https://github.com/googleapis/google-cloud-rust/issues/1342 and wanted to test, but wasn't able to get service account impersonation working; attempted patching it here. We've tested this works for us.

This pull request improves how external account credentials are parsed and handled, particularly for Kubernetes Workload Identity Federation (WIF) scenarios. The main focus is on making the `subject_token_field_name` optional, setting sensible defaults, and enhancing test coverage for Kubernetes identity configurations.

**Parsing and struct improvements:**

* Made the `subject_token_field_name` field in `CredentialSourceFormat` optional (`Option<String>`) to allow for configurations where it may be omitted.
* Updated construction logic in both `FileSourcedCredentials` and `UrlSourcedCredentials` to use an empty string as the default value for `subject_token_field_name` when not provided. [[1]](diffhunk://#diff-aeb4d0e86f16c6fe270b6e0a4d99a576b4bbb5dda53d80614694e936612d191dL37-R37) [[2]](diffhunk://#diff-1c1521b5f83d83d8f96a2609e7c133c887ed4c6cee713cb08b41ddb4bd6c3293L45-R45)

**Enum and test enhancements:**

* Reordered variants in the `CredentialSourceFile` enum to ensure correct deserialization for untagged enums, prioritizing the `Executable` variant.
* Added new tests for Kubernetes WIF direct identity and impersonation config parsing, verifying correct defaulting and field extraction for file-sourced credentials.

### DIFF
--- a/src/auth/src/credentials/external_account.rs
+++ b/src/auth/src/credentials/external_account.rs
@@ -140,7 +140,8 @@ const IAM_SCOPE: &str = "https://www.googleapis.com/auth/iam";
 pub(crate) struct CredentialSourceFormat {
     #[serde(rename = "type")]
     pub format_type: String,
-    pub subject_token_field_name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub subject_token_field_name: Option<String>,
 }
 
 #[derive(Serialize, Deserialize, Debug, Clone, Default)]
@@ -153,13 +154,14 @@ pub(crate) struct ExecutableConfig {
 #[derive(Serialize, Deserialize, Debug, Clone)]
 #[serde(untagged)]
 enum CredentialSourceFile {
+    // Most specific variants first for untagged enum
+    Executable {
+        executable: ExecutableConfig,
+    },
     Url {
         url: String,
         headers: Option<HashMap<String, String>>,
         format: Option<CredentialSourceFormat>,
-    },
-    Executable {
-        executable: ExecutableConfig,
     },
     File {
         file: String,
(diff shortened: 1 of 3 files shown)


coryan:
Thanks for the PR!  The bots would like you to run `cargo fmt -p google-cloud-auth` from the top-level directory do you mind making that change? I will start the next round of bots in the interim

fujin:
I've run `cargo fmt -p google-cloud-auth` as indicated, force pushed with lease. :)

coryan (approved):
Thanks for the PR.  Looks good to me, I need one of my colleagues to review too.

src/auth/src/credentials/external_account.rs
```diff
@@ -140,7 +140,8 @@ const IAM_SCOPE: &str = "https://www.googleapis.com/auth/iam";
 pub(crate) struct CredentialSourceFormat {
     #[serde(rename = "type")]
     pub format_type: String,
-    pub subject_token_field_name: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub subject_token_field_name: Option<String>,
```
dbolduc, line 144: the empty string business looks fishy to me. Is an empty string ever valid? If not, it makes me think this would have worked:

```rs
    #[serde(skip_serializing_if = "String::is_empty")]
```

dbolduc (approved):
I don't have time to think about this. Approved.

src/auth/src/credentials/external_account.rs
coryan: I agree, but I find this convincing:

https://google.aip.dev/auth/4117#determining-the-subject-token-in-file-sourced-credentials

The field is optional, this sounds like a better representation.  In any case, we can always change this later.

PR #3143, opened 2025-08-26
docs(storage): add upload and download from memory examples

### LINKED ISSUE #3077: storage_file_download_into_memory

### LINKED ISSUE #3078: storage_file_upload_from_memory

suzmue:
Fixes #3077 and fixes #3078

### DIFF
--- a/src/storage/examples/src/objects/file_upload_from_memory.rs
+++ b/src/storage/examples/src/objects/file_upload_from_memory.rs
@@ -17,7 +17,7 @@ use google_cloud_storage::client::Storage;
 
 pub async fn sample(client: &Storage, bucket: &str) -> Result<(), anyhow::Error> {
     const NAME: &str = "object-to-upload.txt";
-    let data = "Hello, world!";
+    let data: bytes::Bytes = bytes::Bytes::from("Hello, world!");
     let _result = client
         .write_object(format!("projects/_/buckets/{bucket}"), NAME, data)
         .send_unbuffered()


src/storage/examples/src/objects/file_upload_from_memory.rs
```diff
@@ -0,0 +1,29 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START storage_upload_from_memory]
+use google_cloud_storage::client::Storage;
+
+pub async fn sample(client: &Storage, bucket: &str) -> Result<(), anyhow::Error> {
+    const NAME: &str = "object-to-upload.txt";
+    let data = "Hello, world!";
```
coryan: Do you think we should use `bytes::Bytes` here to show that this consumes more than just strings?

coryan:
The code LGTM. The builds won't run until your rebase and resolve the conflicts.

src/storage/examples/src/objects/file_upload_from_memory.rs
suzmue: Done!

Hmm yeah I wonder if this sample should cover multiple Payload types.

coryan: I would keep it simple (to one type) we may want to cover the different upload types in our own user guide.

coryan (approved):

PR #3144, opened 2025-08-26
docs(user-guide): improve list operations guide

pcoet:
* Updated "Prerequisites" to match other updated guides that use Secret Manager.
* Updated first example (iterating over items) to mention `ItemPaginator` and `by_item`. It looks like an older version of the guide showed `Paginator` and `items`, and the code got updated but not the text.
* Updated second example (iterating over pages) to show how to bring `Paginator` into scope and to mention `by_page`.
* Documented `use` statement for streaming example.
* Made various nitty style updates, including user "list" instead of "List", unless we're talking about the API operation.
* Add two "What's next" links.

### DIFF
--- a/guide/src/pagination.md
+++ b/guide/src/pagination.md
@@ -131,8 +131,8 @@ token to resume paginating from a specific page.
 
 The standard [Google API List] method follows the pagination guideline defined
 by [AIP-158]. Each call to a `List` method for a resource returns a page of
-resource items (e.g. secrets) along with a next-page token that can be passed
-to the `List` method to retrieve the next page.
+resource items (e.g. secrets) along with a next-page token that can be passed to
+the `List` method to retrieve the next page.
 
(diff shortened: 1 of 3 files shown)


guide/src/pagination.md
```diff
@@ -14,61 +14,69 @@ See the License for the specific language governing permissions and
...
 ```shell
 cargo add google-cloud-secretmanager-v1
 ```
 
-## Iterating List methods
+## Iterating list methods
 
 To help iterate the items in a list method, the APIs return an implementation of
-the `Paginator` trait. We need to introduce it in scope via a `use` declaration:
+the `ItemPaginator` trait. Introduce it into scope via a `use` declaration:
```
coryan, line 51: Yeah, we definitely missed updating the text here.

coryan (approved):
The code LGTM, but you need another round of formatting. Pesky robots.

PR #3154, opened 2025-08-28
impl(storage): introduce stub interface

dbolduc:
Part of the work for #2041 

My plan is to
- put the stub in place
- introduce the `transport::Storage` stub and implement `read_object` with it
- implement `write_object` with the `transport::Storage` stub
- plumb the `transport::Storage` stub from the client to the builders
- replace the `transport::Storage` stub with a generic `T: crate::stub::Storage`
- add the `client::Storage::from_stub()` and mock tests. Make sure everything that needs to be public is public.

We could template the client and the builders with a placeholder earlier.

### DIFF
--- a/src/storage/src/storage/stub.rs
+++ b/src/storage/src/storage/stub.rs
@@ -1,0 +1,80 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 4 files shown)


src/storage/src/storage/stub.rs
```diff
@@ -0,0 +1,80 @@
...
+        "traits provide default implementations of each method. In the client ",
+        "libraries, all implementations of the traits override all methods. ",
+        "Therefore, this error should not appear in normal code using the ",
+        "client libraries. The only expected context for this error is test ",
+        "code mocking the client libraries. If that is how you got this ",
+        "error, verify that you have mocked all methods used in your test. ",
+        "Otherwise, please open a bug at ",
+        "https://github.com/googleapis/google-cloud-rust/issues"
+    ));
+}
```
coryan: Any reason the version from `gax-internal` does not work?  Is it because it has a `T: Send` restriction?

If so, can we refactor the text to a public constant in gax-internal?

dbolduc: It was the return type not lining up exactly. `gax::Result<gax::Response<T>>` vs. `gax::Result<T>`.

dbolduc:
> Can we fix that?

Yep. Fixed. PTAL

coryan (approved):

Issue #3100, opened 2025-08-25
Create storage bucket examples


suzmue:

coryan:
AFAICT this is done.

Issue #3098, opened 2025-08-25
storage_object_get_kms_key


suzmue:

nvnmandadhi:
I would like to work on this.

PR #3157, opened 2025-08-28
ci: create KMS resources for integration tests

coryan:
The integration tests for the storage examples will require a KMS crypto
key, which in turns lives in a KMS key ring. And we need the storage
service account to have access to the crypto key too.

part of the work for #3100 and #3101, #3098, #3097 and  probably several other examples.

### DIFF
--- a/.gcb/builds/resources/main.tf
+++ b/.gcb/builds/resources/main.tf
@@ -58,6 +58,31 @@ resource "google_firestore_database" "default" {
   type        = "FIRESTORE_NATIVE"
 }
 
+# Create a KMS Key Ring and key for the storage sample tests.
+resource "google_kms_key_ring" "us-central1" {
+  name     = "us-central1"
+  location = "us-central1"
+}
+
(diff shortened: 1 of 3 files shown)


.gcb/builds/resources/main.tf
```diff
@@ -58,6 +58,31 @@ resource "google_firestore_database" "default" {
...
+  name     = "us-central1"
+  location = "us-central1"
+}
+
+# A crypto key for the storage examples.
+resource "google_kms_crypto_key" "storage-examples" {
+  name     = "storage-examples"
+  key_ring = google_kms_key_ring.us-central1.id
+  # Rotate every 10 days
+  rotation_period = "864000s"
```
dbolduc, line 72: Q: does something need to rerun terraform to rotate?

coryan: AFAIK, KMS will rotate these automatically:

https://cloud.google.com/kms/docs/rotate-key#automatic

PR #3158, opened 2025-08-28
impl(storage): use stub for read_object

dbolduc:
Part of the work for #2041 

Introduce the `transport::Storage` stub, with an implementation for `read_object`.

Refactor `ReadObject` to write the implementation in terms of the `stub`, not the `inner`.

We start by naming the `transport::Storage` stub explicitly in `ReadObject`. Eventually it will get replaced by a `T: stub::Storage`, when we are ready.

`ReadObject` can't know about the `http_request_builder` any more. (That is only known by the specific transport stub implementation, which will eventually be a generic stub implementation). So in tests, we use the `ReadObject` builder to give us a request and options. We use those to construct something that knows about `http_request_builder`.

### DIFF
--- a/src/storage/src/storage/transport.rs
+++ b/src/storage/src/storage/transport.rs
@@ -1,0 +1,48 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::Result;
+use crate::model::ReadObjectRequest;
+use crate::read_object::ReadObjectResponse;
+use crate::storage::client::StorageInner;
+use crate::storage::read_object::{ReadObjectResponseImpl, Reader};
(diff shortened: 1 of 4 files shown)


src/storage/src/storage/transport.rs
```diff
@@ -0,0 +1,48 @@
...
+use crate::model::ReadObjectRequest;
+use crate::read_object::ReadObjectResponse;
+use crate::storage::client::StorageInner;
+use crate::storage::read_object::{ReadObjectResponseImpl, Reader};
+use crate::storage::request_options::RequestOptions;
+use std::sync::Arc;
+
+#[derive(Clone, Debug)]
+pub struct Storage {
+    inner: Arc<StorageInner>,
```
coryan, line 25: ETOOMANYARCS ... but maybe this is temporary and we will remove one level of Arc'ing before this is final?

coryan (commented):
If we will need all these Arcs forever we need to talk. Otherwise LGTM.

src/storage/src/storage/transport.rs
dbolduc: I don't have a good plan to remove a level of `Arc`s.

We could force the `stub::Storage` to be `Clone`, and hold a `stub: T` in the client. I don't really like that.

I don't think we can give the builders a `&stub`. That seems like a lifetime disaster.

I don't think the transport stub can hold an `inner: StorageInner`. We would have to clone `inner` or, use a reference and tie `Reader` to the lifetime of the transport stub... I hate that.

Observation: all of the `Arc`ing is done at client creation.

Do you have any suggestions?

coryan: > Do you have any suggestions?

This seems like something we can fix later if we can figure out how.

PR #3159, opened 2025-08-28
refactor(storage): use `WriteObjectRequest`

dbolduc:
Part of the work for #2041 

We have the `WriteObjectRequest` (which is just a `spec` and `params`), might as well use it to hold the `spec` and `params` fields.

### DIFF
--- a/src/storage/src/storage/write_object.rs
+++ b/src/storage/src/storage/write_object.rs
@@ -75,8 +75,7 @@ use crate::storage::checksum::details::{Checksum, Md5};
 /// ```
 pub struct WriteObject<T> {
     inner: std::sync::Arc<StorageInner>,
-    spec: crate::model::WriteObjectSpec,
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/write_object.rs
```diff
@@ -782,8 +782,8 @@ impl<T> WriteObject<T> {
         PerformUpload::new(
             self.payload,
             self.inner,
-            self.spec,
-            self.params,
+            self.request.spec,
+            self.request.params,
```
coryan, line 786: Seems like `PerformUpload` should keep these aggregated then?

coryan (approved):

src/storage/src/storage/write_object.rs
dbolduc: It could. I did not plan to change it though.

PR #3166, opened 2025-08-28
feat(storage): fake `ReadObjectResponse` for mocks

dbolduc:
Part of the work for #2041 

Offer a way for applications to create a `ReadObjectResponse`, so they can mock `client::Storage::read_object`.

My intention was to accept the same types as we do for the `write_object` payload.

### DIFF
--- a/src/storage/src/read_object.rs
+++ b/src/storage/src/read_object.rs
@@ -39,7 +39,7 @@ impl ReadObjectResponse {
 
     /// Create a ReadObjectResponse, given a data source.
     ///
-    /// Use this API to mock the return type of
+    /// Use this method to mock the return type of
     /// [Storage::read_object][crate::client::Storage::read_object].
     ///
     /// # Example


src/storage/src/read_object.rs
```diff
@@ -108,3 +134,64 @@ pub(crate) mod dynamic {
...
+{
+    fn object(&self) -> ObjectHighlights {
+        self.object.clone()
+    }
+
+    async fn next(&mut self) -> Option<Result<bytes::Bytes>> {
+        self.source
+            .next()
+            .await
+            .map(|r| r.map_err(gax::error::Error::io))
```
coryan, line 171: Ah, I get it now.

src/storage/src/read_object.rs
```diff
@@ -36,6 +37,31 @@ impl ReadObjectResponse {
         Self { inner }
     }
 
+    /// Create a ReadObjectResponse, given a data source.
+    ///
+    /// Use this API to mock the return type of
```
coryan: s/Use this API/Use this method/

"API" is terribly easy to confuse with protocol level things.

dbolduc: Fixed

coryan (approved):

PR #3169, opened 2025-08-28
feat(storage): mocking

### LINKED ISSUE #2041: Add stubs and tracing to `storage::Storage`

dbolduc:
Fixes #2041

### DIFF
--- a/src/storage/tests/mocking.rs
+++ b/src/storage/tests/mocking.rs
@@ -14,8 +14,112 @@
 
 #[cfg(test)]
 mod tests {
+    use gax::error::{
+        Error,
+        rpc::{Code, Status},
+    };
+    use gcs::Result;
+    use gcs::model::{Object, ReadObjectRequest};
+    use gcs::model_ext::{ObjectHighlights, WriteObjectRequest};
+    use gcs::read_object::ReadObjectResponse;
+    use gcs::request_options::RequestOptions;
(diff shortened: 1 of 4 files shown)


src/storage/tests/mocking.rs
```diff
@@ -14,9 +14,113 @@
...
+        let object = {
+            let mut o = ObjectHighlights::default();
+            o.etag = "custom-etag".to_string();
+            o
+        };
+        let object_clone = object.clone();
+
+        let mut mock = MockStorage::new();
+        mock.expect_read_object()
+            .return_once(move |_, _| Ok(ReadObjectResponse::from_source(object_clone, LAZY)));
```
coryan: `object_clone` is ugly. optional:

```suggestion
        let object_clone = object.clone();

        let mut mock = MockStorage::new();
        mock.expect_read_object()
            .return_once({ let clone = object.clone(); move |_, _| Ok(ReadObjectResponse::from_source(clone, LAZY)) });
```

coryan: I did not test that fix, use at your own risk.

dbolduc: Maybe better. Thanks.

PR #3184, opened 2025-08-29
docs(storage): copy file example

### LINKED ISSUE #3056: storage_copy_file

dbolduc:
Fixes #3056 

The associated cloud docs are [Copy an object between buckets](https://cloud.google.com/storage/docs/samples/storage-copy-file#whats-next), so...
- I think we need separate parameters for source and destination buckets.
- I think we need to use a rewrite loop. The buckets could be in different regions.

I tried to consolidate the print statement, but it is still super long.


coryan (approved):
FWIW, the example is intended for this RPC:

https://cloud.google.com/storage/docs/json_api/v1/objects/copy

Which does not have a gRPC equivalent. Rewrite is how you achieve the same effect. Rewrite can do things that `Copy` cannot, but it does require a more complex bit of code in the client.

PR #3195, opened 2025-09-02
impl(wkt): workaround for time==0.3.42

coryan:
It seems like time==0.3.42 has a regression in its Nanosecond -> Second
helpers. This workarounds that problem using a constant value to convert
from nanoseconds to seconds.

More details at: https://github.com/time-rs/time/issues/749

Unblocks #3193

### DIFF
--- a/src/wkt/src/timestamp.rs
+++ b/src/wkt/src/timestamp.rs
@@ -300,11 +300,9 @@ impl TryFrom<time::OffsetDateTime> for Timestamp {
     type Error = TimestampError;
 
     fn try_from(value: time::OffsetDateTime) -> Result<Self, Self::Error> {
-        use time::convert::{Nanosecond, Second};
-
+        const SCALE: i128 = 1_000_000_000_i128;
         let seconds = value.unix_timestamp();
-        let nanos = (value.unix_timestamp_nanos()
-            - seconds as i128 * Nanosecond::per(Second) as i128) as i32;
+        let nanos = (value.unix_timestamp_nanos() - seconds as i128 * SCALE) as i32;
         Self::new(seconds, nanos)
     }
 }
--- a/src/wkt/src/timestamp.rs
+++ b/src/wkt/src/timestamp.rs
@@ -300,9 +300,8 @@ impl TryFrom<time::OffsetDateTime> for Timestamp {
     type Error = TimestampError;
 
     fn try_from(value: time::OffsetDateTime) -> Result<Self, Self::Error> {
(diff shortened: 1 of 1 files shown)


src/wkt/src/timestamp.rs
```diff
@@ -300,11 +300,9 @@ impl TryFrom<time::OffsetDateTime> for Timestamp {
     type Error = TimestampError;
 
     fn try_from(value: time::OffsetDateTime) -> Result<Self, Self::Error> {
-        use time::convert::{Nanosecond, Second};
-
+        const SCALE: i128 = 1_000_000_000_i128;
         let seconds = value.unix_timestamp();
-        let nanos = (value.unix_timestamp_nanos()
-            - seconds as i128 * Nanosecond::per(Second) as i128) as i32;
```
dbolduc, line 307: Apparently we should say:

```rs
         let nanos = (value.unix_timestamp_nanos()
            - seconds as i128 * Nanosecond::per_t::<i128>(Second)) as i32;
```

I checked, and it still works against `time == 0.3.42`

src/wkt/src/timestamp.rs
```diff
@@ -300,11 +300,9 @@ impl TryFrom<time::OffsetDateTime> for Timestamp {
     type Error = TimestampError;
 
     fn try_from(value: time::OffsetDateTime) -> Result<Self, Self::Error> {
-        use time::convert::{Nanosecond, Second};
-
+        const SCALE: i128 = 1_000_000_000_i128;
```
dbolduc: moot if you take the other suggestion, but there is already a constant for this:

https://github.com/googleapis/google-cloud-rust/blob/e1a38539c4588a2805fb2ee09655c938a69e1bd9/src/wkt/src/timestamp.rs#L243

src/wkt/src/timestamp.rs
coryan: It does not work with `time == 0.3.41`... If we do that  we also need to bump the minimum required version for `time`, which means we will force everybody to pick 0.3.42 potentially breaking their code.  That may get fixed in a few days, but I would rather not risk it.

src/wkt/src/timestamp.rs
coryan: Fixed.

src/wkt/src/timestamp.rs
dbolduc: > It does not work with `time == 0.3.41`

Ah, I did not try this, obviously. Thanks

dbolduc (approved):

PR #3203, opened 2025-09-02
docs(storage): add sample for converting object from csek to cmek

### LINKED ISSUE #3097: storage_object_csek_to_cmek

suzmue:
Fixes #3097

### DIFF
--- a/src/storage/examples/src/objects/object_csek_to_cmek.rs
+++ b/src/storage/examples/src/objects/object_csek_to_cmek.rs
@@ -1,0 +1,55 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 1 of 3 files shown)


src/storage/examples/src/objects/object_csek_to_cmek.rs
```diff
@@ -0,0 +1,55 @@
...
+    // For more details on this loop, see the "Rewriting objects" section of the
+    // user guide:
+    // https://googleapis.github.io/google-cloud-rust/storage/rewrite_object.html
+    let updated = loop {
+        let resp = builder.clone().send().await?;
+        if resp.done {
+            break resp.resource;
+        }
+        builder = builder.set_rewrite_token(resp.rewrite_token);
+    };
```
coryan, line 48: Seems like we should add a helper to run this loop, something like:

```rust
pub async fn rewrite_all(builder: crate::builder::storage_control::RewriteObject) -> Result<crate::model::Object> {
    ... ... ...
}
```

We need to decide where this goes (what module) and we need to write documentation for it.

coryan (approved):

PR #3205, opened 2025-09-02
docs(storage): add retry configuration samples

### LINKED ISSUE #3055: storage_configure_retries

suzmue:
The backoff and retry policy types do not implement Clone, so the samples create new ones for each client.

Fixes #3055

### DIFF
--- a/src/storage/examples/src/client/configure_retries.rs
+++ b/src/storage/examples/src/client/configure_retries.rs
@@ -22,9 +22,9 @@ use std::time::Duration;
 pub async fn sample(bucket_id: &str) -> anyhow::Result<()> {
     // Retries all operations for up to 5 minutes, including any backoff time.
     let retry_policy = RetryableErrors.with_time_limit(Duration::from_secs(60 * 5));
-    // On error, it backs off for a random delay between [1, 3] seconds, then [3,
-    // 9] seconds, then [9, 27] seconds, etc. The backoff time never grows larger
-    // than 1 minute.
+    // On error, it backs off for a random delay between [0, 1] seconds, then
+    // [0, 3] seconds, then [0, 9] seconds, etc. The backoff time never grows
+    // larger than 1 minute.
     let backoff_policy = ExponentialBackoffBuilder::new()
         .with_initial_delay(Duration::from_secs(1))
         .with_maximum_delay(Duration::from_secs(60))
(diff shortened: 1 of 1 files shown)


src/storage/examples/src/client/configure_retries.rs
```diff
@@ -0,0 +1,73 @@
...
+use google_cloud_gax::retry_policy::RetryPolicyExt;
+use google_cloud_storage::client::{Storage, StorageControl};
+use google_cloud_storage::retry_policy::RetryableErrors;
+use std::time::Duration;
+
+pub async fn sample(bucket_id: &str) -> anyhow::Result<()> {
+    // Retries all operations for up to 5 minutes, including any backoff time.
+    let retry_policy = RetryableErrors.with_time_limit(Duration::from_secs(60 * 5));
+    // On error, it backs off for a random delay between [1, 3] seconds, then [3,
+    // 9] seconds, then [9, 27] seconds, etc. The backoff time never grows larger
```
coryan: nit: reflow so the `[3, 9]` interval is easier to read?

src/storage/examples/src/client/configure_retries.rs
```diff
@@ -0,0 +1,73 @@
...
+use google_cloud_gax::exponential_backoff::ExponentialBackoffBuilder;
+use google_cloud_gax::retry_policy::RetryPolicyExt;
+use google_cloud_storage::client::{Storage, StorageControl};
+use google_cloud_storage::retry_policy::RetryableErrors;
+use std::time::Duration;
+
+pub async fn sample(bucket_id: &str) -> anyhow::Result<()> {
+    // Retries all operations for up to 5 minutes, including any backoff time.
+    let retry_policy = RetryableErrors.with_time_limit(Duration::from_secs(60 * 5));
+    // On error, it backs off for a random delay between [1, 3] seconds, then [3,
```
coryan: This is not true I believe:

https://github.com/googleapis/google-cloud-rust/blob/0f870d1dce4af597602b961d500d100123742496/src/gax/src/exponential_backoff.rs#L210

The delay is first in the range $[0, 1]$, then in $[0, 3]$, then in $[0, 9]$, etc.

coryan (changes requested):

src/storage/examples/src/client/configure_retries.rs
suzmue: http://cloud/storage/docs/samples/storage-configure-retries Ah I had referenced the C++ sample, which I guess works very differently (or is also wrong I will go check)! Will update the comment, thanks for catching

src/storage/examples/src/client/configure_retries.rs
suzmue: Done.

coryan (approved):

src/storage/examples/src/client/configure_retries.rs
suzmue: C++ example looks correct for that implementation.

Fixed this sample to match our implementation.

coryan: https://github.com/googleapis/google-cloud-cpp/issues/8755 probably has the background for C++.

PR #3217, opened 2025-09-03
docs(rustdocfx): add borrowed ref for returns

PhongChuong:


doc/rustdocfx/generate.go
```diff
@@ -307,18 +307,8 @@ func processImplementation(c *crate, id string, page *docfxManagedReference, par
...
-			}
-			page.appendItem(assocConst)
-
-			reference, err := newDocfxReferenceFromDocfxItem(assocConst, parent)
-			if err != nil {
-				return fmt.Errorf("error processing item with id %s: %w", id, err)
-			}
-			parent.appendChildren(reference.Uid)
-			page.appendReference(reference)
+			// TODO: Add associated types.
```
dbolduc, line 310: nit: we should get in the habit of opening bugs for TODOs, like `TODO(#XXXX)`

doc/rustdocfx/workspace.go
```diff
@@ -468,25 +470,42 @@ func (t *typeEnum) toString() string {
...
+func (t *borrowedRef) toString() string {
 	if t == nil {
 		return ""
 	}
 	ret := ""
-	// TODO: Lifetime
+	// TODO: Error check for lifetime.
 	if t.IsMutable {
-		ret = "&mut "
+		ret = fmt.Sprintf("&mut %s", t.Type.toString())
```
dbolduc: nit: no need to define `ret`, could just return, with the code as is.

doc/rustdocfx/workspace.go
```diff
@@ -468,25 +470,42 @@ func (t *typeEnum) toString() string {
...
+	if t.BorrowedRef != nil {
+		return t.BorrowedRef.toString()
+	}
+	if len(t.ImplTrait) > 0 {
+		bounds := []string{}
+		for i := 0; i < len(t.ImplTrait); i++ {
+			if t.ImplTrait[i].TraitBound != nil {
+				bound := t.ImplTrait[i].TraitBound.Trait.toString()
+				bounds = append(bounds, bound)
+			} else if t.ImplTrait[i].Outlives != nil {
```
dbolduc, line 482: should this be an `if` instead of `else if`?

(idk, I would have to look into the data)

dbolduc (dismissed):

doc/rustdocfx/generate.go
PhongChuong: I'll update all the TODOs to the correct format once we're in better shape.

doc/rustdocfx/workspace.go
PhongChuong: From the json files, Outlives and TraitBound are mutually exclusive for our crates.
I'll verify in the error checking PR.

doc/rustdocfx/workspace.go
PhongChuong: Done.

dbolduc (approved):

PR #3225, opened 2025-09-03
feat(dart): add `google_cloud_ai_generativelanguage_v1` package

brianquinlan:
- Add a google_cloud_ai_generativelanguage_v1 package
- Create an example that uses it


coryan:
Closing, we like to keep our PR list tidy. Feel free to reopen if you want to merge something like this.  If you want to share the diff with a colleague, something like this works:

https://github.com/googleapis/google-cloud-rust/compare/main...brianquinlan:google-cloud-rust:generate_genai

No need to create a PR.

PR #3227, opened 2025-09-03
docs(storage): add object print acl example

nvnmandadhi:


src/storage/examples/src/lib.rs
```diff
@@ -516,6 +516,9 @@ pub async fn run_object_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
         .await?;
     tracing::info!("create test object for object ACL examples");
     let _ = make_object(&client, &id, "object-to-update").await;
+    let _ = make_object_with_acl(&client, &id, "object-with-acl").await;
+    tracing::info!("running print_file_acl example");
```
dbolduc: We don't need any of this fancy set up. It is nice to see an ACL being printed, but not necessary for our tests.

I think we should put this next to `objects::get_metadata::sample(...)` and reuse "object-to-read" as the `NAME`.

src/storage/examples/src/objects/print_file_acl.rs
```diff
@@ -0,0 +1,34 @@
...
+        .set_bucket(format!("projects/_/buckets/{bucket_id}"))
+        .set_object(NAME)
+        .send()
+        .await?;
+
+    let object_acl = object.acl;
+    println!(
+        "Object {NAME} in bucket {bucket_id} has the following ACL: {:?}",
+        object_acl
+    );
```
dbolduc: optional style nit: I would save a line by substituting `object.acl` in the print statement, instead of defining `let object_acl = ...`.

dbolduc (changes requested):

src/storage/examples/src/objects/print_file_acl.rs
```diff
@@ -0,0 +1,33 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START storage_print_file_acl]
+use google_cloud_storage::client::StorageControl;
+
+pub async fn sample(client: &StorageControl, bucket_id: &str) -> anyhow::Result<()> {
+    const NAME: &str = "object-with-acl";
```
dbolduc: I think the builds will fail because this object doesn't exist. `"object-to-read"` would work.

FYI, here is where we create the objects referenced in the samples:

https://github.com/googleapis/google-cloud-rust/blob/0e326e639a0299bc6cd213a3c127d01cd7094507/src/storage/examples/src/lib.rs#L389

nvnmandadhi: ah, that was a miss on my part.

src/storage/examples/src/lib.rs
```diff
@@ -28,7 +28,7 @@ use google_cloud_storage::model::bucket::{
     iam_config::UniformBucketLevelAccess,
     {HierarchicalNamespace, IamConfig},
 };
-use google_cloud_storage::model::{Bucket, Object};
+use google_cloud_storage::model::{Bucket, Object, ObjectAccessControl};
```
dbolduc: linter says `ObjectAccessControl` is unused

```sh
cargo clippy -p storage-samples --all-targets -- --deny warnings
```

Issue #2923, opened 2025-08-19
Create an exporter for Cloud Trace


coryan:
We would like to export traces to Cloud Trace.  We need a new crate (maybe `google-cloud-otel`).

fiadliel:
There is a current OTel implementation (but based directly on Tonic, and some pre-existing auth crates) here: https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-stackdriver - you may want to liaise with the group if planning an official replacement.

It probably wouldn't take much to convert it - the most annoying piece is probably just detecting the environment-local project ID, if you want to keep that feature.

Btw, I created a completely separate tracing implementation for fastrace here: https://github.com/fiadliel/fastrace-google-cloud (though not published to crates.io at this time). It might at least be useful for another point of view in considering what the tracing features should look like.

coryan:
"Stackdriver", I forgot to search for that.  Thanks!  This is an area where replacing existing things may not help that much, or may actually hurt.

dbolduc:
Also, Googlers should look at go/otel-gcp-trace-exporter-spec to make sure we cover everything the exporter is supposed to do.

coryan:
@westarle FYI

PR #3231, opened 2025-09-03
feat(wkt): conversion to/from `std::time::SystemTime`

### LINKED ISSUE #3230: Conversions to/from `std::time::SystemTime`
We often finding ourselves needing to convert `wkt::Timestamp` <---> `std::time::SystemTime`. Let's make it part of the library.

dbolduc:
Fixes #3230, would be nice for #2923 and maybe #3119

cc: @nvnmandadhi 

It seems that in practice, `SystemTime` cannot represent an out of bounds time. (e.g. something like `SystemTime::UNIX_EPOCH.checked_add(Duration::from_secs(i64::MAX as u64 + 1))`. The code has checks, because I do not know if that will always be the case. But there are no tests we can write.

### DIFF
--- a/src/wkt/src/timestamp.rs
+++ b/src/wkt/src/timestamp.rs
@@ -403,6 +403,46 @@ impl TryFrom<&String> for Timestamp {
     }
 }
 
+/// Converts from [std::time::SystemTime] to [Timestamp].
+///
+/// This conversion may fail if the [std::time::SystemTime] value is out of
+/// range.
+///
+/// # Example
+/// ```
+/// # use std::time::SystemTime;
+/// # use google_cloud_wkt::{Timestamp, TimestampError};
+/// let ts = Timestamp::try_from(SystemTime::now())?;
+/// println!("now={ts:?}");
+/// # Ok::<(), anyhow::Error>(())
+/// ```
+impl TryFrom<std::time::SystemTime> for Timestamp {
+    type Error = TimestampError;
+    fn try_from(value: std::time::SystemTime) -> Result<Self, Self::Error> {
+        match value.duration_since(std::time::SystemTime::UNIX_EPOCH) {
+            Ok(d) => {
+                let s = d.as_secs();
+                if s > (i64::MAX as u64) {
+                    return Err(TimestampError::OutOfRange);
+                }
+                Timestamp::new(s as i64, d.subsec_nanos() as i32)
+            }
+            Err(e) => {
+                // The error case essentially gives us a negative duration.
+                let d = e.duration();
+                let s = d.as_secs();
+                if s > (i64::MAX as u64) {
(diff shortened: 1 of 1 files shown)


src/wkt/src/timestamp.rs
```diff
@@ -403,6 +403,46 @@ impl TryFrom<&String> for Timestamp {
...
+                Timestamp::new(s as i64, d.subsec_nanos() as i32)
+            }
+            Err(e) => {
+                // The error case essentially gives us a negative duration.
+                let d = e.duration();
+                let s = d.as_secs();
+                if s > (i64::MAX as u64) {
+                    return Err(TimestampError::OutOfRange);
+                }
+                Ok(Timestamp::clamp(
```
coryan: Why is `clamp()` safe here?

src/wkt/src/timestamp.rs
```diff
@@ -447,6 +487,41 @@ impl TryFrom<Timestamp> for chrono::DateTime<chrono::Utc> {
...
+            let d = std::time::Duration::from_secs(s as u64);
+            ts = ts
+                .checked_add(d)
+                .ok_or_else(|| TimestampError::OutOfRange)?;
+        } else {
+            let d = std::time::Duration::from_secs(-s as u64);
+            ts = ts
+                .checked_sub(d)
+                .ok_or_else(|| TimestampError::OutOfRange)?;
+        }
```
coryan:
```suggestion
        let s = value.seconds();
        let ts = if s >= 0 {
            let d = std::time::Duration::from_secs(s as u64);
            [ts = ts](std::time::SystemTime::UNIX_EPOCH)
                .checked_add(d)
                .ok_or_else(|| TimestampError::OutOfRange)?
        } else {
            let d = std::time::Duration::from_secs(-s as u64);
            std::time::SystemTime::UNIX_EPOCH
                .checked_sub(d)
                .ok_or_else(|| TimestampError::OutOfRange)?
        };
```

I am not sure it is better...

src/wkt/src/timestamp.rs
dbolduc: It wasn't. Fixed.

src/wkt/src/timestamp.rs
dbolduc: potato potato. Done.

dbolduc:
FYI to the reviewer: I saw a windows failure where math relative to the `UNIX_EPOCH` (with no involvement of `now()`) produced incorrect results. Great. I might need to disable the tests on windows.

It might be due to the windows representation only having 100's of ns precision. I changed the tests to avoid data loss from rounding. I do not know if this fixed the problem, or if this may still flake.

coryan (approved):

PR #3233, opened 2025-09-03
docs(guide): friction log fixes

whuffman36:


guide/src/generate_text_using_the_vertex_ai_gemini_api.md
```diff
@@ -34,6 +34,20 @@ need:
 cargo add google-cloud-aiplatform-v1 --no-default-features --features prediction-service
 ```
 
+If you are using cloud shell (or have limited disk space), you might see a
```
whuffman36: https://b.corp.google.com/issues/429254171

guide/src/generate_text_using_the_vertex_ai_gemini_api.md
```diff
@@ -123,4 +137,9 @@ ______________________________________________________________________
 {{#include ../samples/src/gemini.rs:prompt-and-image}}
 ```
 
+Note: These code samples return `crate::Result<()>` for the sake of brevity.
```
whuffman36: https://b.corp.google.com/issues/429252054

guide/samples/tests/initialize_client.rs
```diff
@@ -44,6 +44,13 @@ pub async fn initialize_client(project_id: &str) -> Result {
 
     Ok(())
 }
+
```
whuffman36, line 45: https://b.corp.google.com/issues/429250702

guide/src/generate_text_using_the_vertex_ai_gemini_api.md
```diff
@@ -123,4 +137,9 @@ ______________________________________________________________________
 {{#include ../samples/src/gemini.rs:prompt-and-image}}
 ```
 
+Note: These code samples return `crate::Result<()>` for the sake of brevity. You
+will need to change the return type to
+`std::result::Result<(), Box<dyn std::error::Error>>` since your project
+structure is likely different from that of the samples.
+
```
coryan: Hmm... projects may use `anyhow::Result<()>` or the `Box` thing ... or their own error that wraps our error as a source.

Maybe we should change all the code to use `anyhow::Result<()>`

guide/src/generate_text_using_the_vertex_ai_gemini_api.md
```diff
@@ -34,6 +34,20 @@ need:
...
+```shell
+cargo clean
+```
+
+Alternatively, you can build in release mode, which should also use less disk
+space:
+
+```shell
+cargo build --release
+```
```
coryan: Feels like this should be a pre-requisite... "Make sure you have at least XX GiB of free disk space, or use `--release`" ?

whuffman36: By default cloud shell only gives you 5GB, if users choose that route it might be helpful to have the `cargo clean` tip. This happened to one of the friction log testers so it's likely to come up again from non-googlers.

coryan: yes, I am suggesting that it belongs in the previous section.

guide/src/generate_text_using_the_vertex_ai_gemini_api.md
whuffman36: Ah that's convenient, I replaced the `crate::Result<()>` with `anyhow::Result<()>`

guide/src/generate_text_using_the_vertex_ai_gemini_api.md
whuffman36: I see, I moved this section to the "Setting Up Rust on Cloud Shell" section.

PhongChuong:
This was merged as #3620.

Issue #3239, opened 2025-09-04
o12y: Detailed tracing for HTTP requests


westarle:
Thanks for stopping by to let us know something could be better!

**Is your feature request related to a problem? Please describe.**
When reviewing traces from my application, I only see the high-level operation of the client.  Beyond that, the operation of the client library (e.g. requests made, backoff and retries) are not present in tracing data.

I would like to see detailed spans (in [tracing](https://docs.rs/tracing/latest/tracing/)) for each individual HTTP request that is sent from the client.  These traces should ultimately have details like the ones defined by [Open Telemetry HTTP semantic conventions](https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span) (e.g. method, service name and port.)

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**

westarle:
We are adding OpenTelemetry-compatible tracing to the HTTP requests made by the
`google-cloud-rust` libraries. Each logical HTTP request attempt will emit a
`tracing` span with attributes aligned with OpenTelemetry semantic conventions.
This integration aims to provide rich, observable telemetry data without
imposing a direct dependency on the `opentelemetry` crate in the core client
libraries.

## Requirements

1.  **Span Emission:** Every HTTP request attempt made via `ReqwestClient` in
    `google-cloud-gax-internal` should emit a `tracing` span.
2.  **Conditional Tracing:** Tracing should only be active if enabled through
    `ClientConfig::tracing` or the `GOOGLE_CLOUD_RUST_LOGGING` environment
    variable (checked via `google_cloud_gax_internal::options::tracing_enabled()`).
3.  **Data Propagation:** Information required for spans, not available directly
    in the HTTP layer, must be passed from the generated client code. This
    includes static client information and per-request details like URL path
    templates.
4.  **Performant:** Avoid allocations or heavy computations when tracing is
    disabled.
5.  **Maintainable:** Attribute keys and recording logic should be easy to
    manage and consistent.
6.  **Alignment:** Span attributes should align with OpenTelemetry Semantic
    Conventions, particularly for HTTP client spans.

## Data Structures

1.  **`google_cloud_gax_internal::options::InstrumentationClientInfo`**:
    Defined in `src/gax-internal/src/options.rs`. Holds static client details.

    ```rust
    #[derive(Copy, Clone, Debug)]
    pub struct InstrumentationClientInfo {
        pub service_name: &'static str,
[...]

PR #3240, opened 2025-09-04
refactor(o12y): Extract inner reqwest code.

westarle:
Pulled out the inner request / response from `request_attempt` so that we can add instrumentation around it as neatly as possible.

A starting point for #3239

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -152,12 +152,27 @@ impl ReqwestClient {
             Ok(CacheableResource::New { data, .. }) => builder.headers(data),
             Ok(CacheableResource::NotModified) => unreachable!("headers are not cached"),
         };
-        let response = builder.send().await.map_err(Self::map_send_error)?;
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/http.rs
```diff
@@ -152,14 +152,29 @@ impl ReqwestClient {
...
         }
 
         self::to_http_response(response).await
     }
 
+    // This function is intended to be as close to the reqwest library call as possible.
+    async fn instrument_and_send(
+        &self,
+        client: reqwest::Client,
+        request: reqwest::Request,
```
coryan, line 172: I have not seen that many `x_and_y()` names for methods in Rust.  I think it may be more idiomatic to call this `execute_http_request()`.  But don't put too much weigh on my opinion. We can change the name at any point.

coryan:
Seems like this is not the path we are going to take.  Can we close it?

westarle:
yes, thanks!

PR #3250, opened 2025-09-05
cleanup(gax)!: future proof backoff policy

coryan:
Use a struct as an argument so we can add fields to it without breaking
the trait definition.


dbolduc (approved):
Good catch

PR #3262, opened 2025-09-06
doc(auth): Remove production usage warning on auth library

sai-sunder-s:


coryan:
We have a PR pending to remove all these warnings in all libraries at the same time:

https://github.com/googleapis/google-cloud-rust/pull/3186/files#diff-9188eb8444e8397a560ac683aeb93a7b7bdbe2a89954e748f6da63514abfdea1

Do you mind if we hold off until then?

sai-sunder-s:
> We have a PR pending to remove all these warnings in all libraries at the same time:
> 
> https://github.com/googleapis/google-cloud-rust/pull/3186/files#diff-9188eb8444e8397a560ac683aeb93a7b7bdbe2a89954e748f6da63514abfdea1
> 
> Do you mind if we hold off until then?

Ah ok, didn't know about that. I see that one is merged, so closing this.

PR #3277, opened 2025-09-08
ci: bump go version

coryan:


.github/workflows/generator_dart.yaml
```diff
@@ -41,6 +41,8 @@ jobs:
           protoc --version
       - name: Install Go
         uses: actions/setup-go@v5
+        with:
+          go-version: 1.24.4
```
codyoss, line 45: nit: for less churn here you can use: https://github.com/actions/setup-go?tab=readme-ov-file#getting-go-version-from-the-gomod-file

coryan: Yes, but I need to `go.mod` file from librarian, which is not readily available.  Do you know of a trick to do that?

PR #3283, opened 2025-09-08
docs(rustdocfx): gcb to upload reference docs to dev staging

PhongChuong:


doc/rustdocfx/preflight.go
```diff
@@ -21,13 +21,7 @@ import (
 )
 
 // preFlightTests() verifies all the required commands are available.
-func preFlightTests(upload bool) error {
-
-	// TODO: Preflight checks for:
-	// cargo workspaces
-	// cargo rustdoc
-	// docuploader
-	// Failfast if not installed.
```
coryan, line 30: Thanks, I forgot to clean those up.

.gcb/builds/triggers/main.tf
```diff
@@ -34,10 +34,15 @@ locals {
...
-  # Add to this list of you want to have more triggers.
-  builds = {
+  # Add to these lists of you want to have more triggers.
+  pr_builds = {
     integration = {}
   }
+
+  pm_builds = {
+    integration     = {}
+    referenceupload = {}
```
coryan, line 44: These new triggers are not enabled yet. Is that intentional?  Do you want to create the triggers first and test them before merging this code?  Both approaches have advantages and disadvantages.

PR #3284, opened 2025-09-08
impl(gax): prevent `ClientConfig` breaking changes

coryan:
This type is in the internal namespace, and `#[doc(hidden)]` so it is
not part of the public API for two reasons. However, nothing prevented
us from using the type in a way that can introduce breaking changes.

Making the type `#[non_exhaustive]` forces us to use the type in more
backwards compatible ways.

FWIW, I reviewed `RequestOptions` and it does not have public fields, so it is not subject to this problem.

### DIFF
--- a/src/gax-internal/src/options.rs
+++ b/src/gax-internal/src/options.rs
@@ -41,10 +41,9 @@ mod tests {
         let _e = ScopedEnv::remove(LOGGING_VAR);
         let config = ClientConfig::default();
         assert!(!tracing_enabled(&config), "expected tracing to be disabled");
-        let config = ClientConfig {
-            tracing: true,
-            ..ClientConfig::default()
-        };
+        let mut config = ClientConfig::default();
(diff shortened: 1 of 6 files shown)


src/gax-internal/src/options.rs
```diff
@@ -41,10 +41,9 @@ mod tests {
         let _e = ScopedEnv::remove(LOGGING_VAR);
         let config = ClientConfig::default();
         assert!(!tracing_enabled(&config), "expected tracing to be disabled");
-        let config = ClientConfig {
-            tracing: true,
-            ..ClientConfig::default()
```
dbolduc, line 46: comment: it annoys me that this code is not valid. I think it will be someday.

coryan: I kind of agree, but I am not a type systems expert that could argue with the Rust compiler folks about why it would be a good or bad idea.  🤷‍♂️

PR #3290, opened 2025-09-08
test: run storage-samples integration tests in ci

### LINKED ISSUE #3288: Add storage-samples integration tests to ci

suzmue:
Skipping all tests that cause failures in the CI and filing follow up bugs. These should be readded in follow up bugs, but at this time trying to get the base set running.

Fixes #3288


src/storage/examples/src/lib.rs
```diff
@@ -512,28 +513,29 @@ pub async fn run_object_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
...
+    //             )),
+    //     )
+    //     .send()
+    //     .await?;
+    // tracing::info!("create test object for object ACL examples");
+    // let _ = make_object(&client, &id, "object-to-update").await;
+    // tracing::info!("running add_file_owner example");
+    // objects::add_file_owner::sample(&control, &id, &service_account).await?;
+    // tracing::info!("running remove_file_owner example");
+    // objects::remove_file_owner::sample(&control, &id, &service_account).await?;
```
coryan: Consider something like this:

https://github.com/googleapis/google-cloud-rust/blob/284d68b81bb5e12ae335e7ce634d29772d8cc44c/src/storage/examples/src/lib.rs#L272-L278

Instead of commenting out the code.

suzmue: Oh I didn't notice you already had that feature created. Thanks I'll use that.

src/storage/examples/src/lib.rs
```diff
@@ -103,8 +103,11 @@ pub async fn run_bucket_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
     let client = control_client().await?;
     let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
     let service_account = std::env::var("GOOGLE_CLOUD_RUST_TEST_SERVICE_ACCOUNT")?;
-    let kms_ring = std::env::var("GOOGLE_CLOUD_RUST_TEST_STORAGE_KMS_RING")?;
-
+    #[cfg(feature = "skipped-integration-tests")]
+    {
+        // TODO(#3292): fix tests that use kms_ring.
+        let kms_ring = std::env::var("GOOGLE_CLOUD_RUST_TEST_STORAGE_KMS_RING")?;
+    }
```
coryan: I think you want:


```suggestion
    // TODO(#3292): fix tests that use kms_ring.
    #[cfg(feature = "skipped-integration-tests")]
    let kms_ring = std::env::var("GOOGLE_CLOUD_RUST_TEST_STORAGE_KMS_RING")?;
```

Or the variable is hidden within the scope and you won't be able to use it below.

src/storage/examples/src/lib.rs
```diff
@@ -366,7 +385,11 @@ pub async fn run_object_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
     let control = control_client().await?;
     let client = Storage::builder().build().await?;
     let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
-    let service_account = std::env::var("GOOGLE_CLOUD_RUST_TEST_SERVICE_ACCOUNT")?;
+    #[cfg(feature = "skipped-integration-tests")]
+    {
+        // TODO(#3289): reenable tests that need UBLA disabled.
+        let service_account = std::env::var("GOOGLE_CLOUD_RUST_TEST_SERVICE_ACCOUNT")?;
+    }
```
coryan: Ditto.  BTW, evidently we are missing a build of this with `--all-features`

.gcb/integration.yaml
```diff
@@ -42,3 +42,4 @@ steps:
       cargo build
       cargo test --package integration-tests --features run-integration-tests
       cargo test --package user-guide-samples --features run-integration-tests
+      cargo test --package storage-samples --features run-integration-tests
```
coryan, line 45: Maybe add:

```suggestion
      cargo test --package storage-samples --features run-integration-tests
      cargo build --package storage-samples --profile=test --all-features
```

To compile all the code, even if it does not run.

suzmue: Great suggestion, thanks.

src/storage/examples/src/lib.rs
suzmue: Done.

src/storage/examples/src/lib.rs
suzmue: Done.

.gcb/integration.yaml
suzmue: Done.

PR #3297, opened 2025-09-09
ci: trim docs build with `rust:nightly`

coryan:
The build is very slow for some reason. Maybe a regression in rustdoc.
In any case, I do not expect that building all the clients will catch
any problems vs. building a few clients

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -408,6 +408,13 @@ jobs:
       - name: Display rustc version
         run: rustup show active-toolchain -v
       - run: cargo doc --workspace
+        if: matrix.rust-version == 'rust:current'
+        env:
+          RUSTDOCFLAGS: "-D warnings"
+      - run: cargo doc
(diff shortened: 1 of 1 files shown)


.github/workflows/sdk.yaml
```diff
@@ -408,6 +408,13 @@ jobs:
...
+        if: matrix.rust-version == 'rust:current'
+        env:
+          RUSTDOCFLAGS: "-D warnings"
+      - run: cargo doc
+        # The nightly builds are slow for some reason, so we run a smaller set
+        # of documents for them. We expect these will catch most problems.
+        if: matrix.rust-version == 'rust:nightly'
         env:
           RUSTDOCFLAGS: "-D warnings"
       - run: cargo doc --package google-cloud-gax
```
dbolduc: while we're here, remove this?

coryan: Done.

coryan (commented):
PTAL

dbolduc (approved):

Issue #3226, opened 2025-09-03
Mocking storage samples + explain rationale


dbolduc:
We should show how to mock storage / how to write testable code using the storage client in the user guide.

And we should explain why storage client is templated on the stub, unlike the GAPICs.

And we should say that the generics are only necessary if you need to fake the client.

PR #3300, opened 2025-09-09
impl(rustdocfx): introduce markdown parsing for comments

dbolduc:
Part of the work for #3226 

Introduce a simple parser for markdown. We will use these on the doc strings.

---

For context, this is where we are going: https://github.com/dbolduc/google-cloud-rust/commit/84636a719ce3ed82e70f234546c98a54832619f9

I plan to add the node kinds one at a time to make the reviews easier. But some of the code along the way may seem like overkill.

### DIFF
--- a/doc/rustdocfx/go.mod
+++ b/doc/rustdocfx/go.mod
@@ -7,4 +7,3 @@ require (
 	github.com/google/go-cmp v0.7.0
         github.com/yuin/goldmark v1.7.13
 )
-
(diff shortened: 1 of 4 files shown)


coryan:
You can make these builds go faster if you approve #3297

doc/rustdocfx/go.mod
```diff
@@ -5,4 +5,5 @@ go 1.24.4
 require (
 	github.com/cbroglie/mustache v1.4.0
 	github.com/google/go-cmp v0.7.0
+        github.com/yuin/goldmark v1.7.13
```
dbolduc, line 8: I see that this has the wrong spaces. I will fix in a follow up.

coryan: I suspect my linters are not organized properly.  It seems they need to be part of the module that they lint.  Or at least a parent module.  Probably requires some code reorganization.

PR #3301, opened 2025-09-09
chore: use `preview` vs. `GA` in repo-metadata

coryan:


dbolduc (commented):
There is also `src/storage/.repo-metadata.json`

coryan:
> There is also `src/storage/.repo-metadata.json`

Thanks, fixed.

dbolduc (approved):

PR #3324, opened 2025-09-11
test(storage): run UBLA-disabled samples

### LINKED ISSUE #3289: Restore UBLA tests for storage samples
These tests will need to be disabled to run in the CI for #3288 because the project used to run the tests does not have appropriate capababilities. We will need to turn this back on once the project has the appropriate permissions.

dbolduc:
Fixes #3289


dbolduc:
> I renamed the PR

`test` is correct. I didn't consider it. I only considered `ci`, `chore`, `fix`, `docs`.

Issue #3204, opened 2025-09-02
Add a `rewrite_until_done()` helper


coryan:
Seems like we should add a helper to run this loop, something like:

```rust
pub async fn rewrite_until_done(builder: crate::builder::storage_control::RewriteObject) -> Result<crate::model::Object> {
    ... ... ...
}
```

We need to decide where this goes (what module) and we need to write documentation for it.

_Originally posted by @coryan in https://github.com/googleapis/google-cloud-rust/pull/3203#discussion_r2317201299_

suzmue:
#3328 implements this as an extension trait in `model_ext`.

Whatever implementation we choose should also come with an update to the docs: https://googleapis.github.io/google-cloud-rust/storage/rewrite_object.html

PR #3328, opened 2025-09-11
feat(storage): add rewrite_until_done helper as extension trait

suzmue:
Several examples for object rewrite operations required a manual loop to poll the operation until it was complete. This boilerplate is cumbersome for users and prone to error.

This change introduces a `RewriteObjectExt` extension trait, which provides `rewrite_until_done()` helper method on the `RewriteObject` builder. This function encapsulates the polling logic, automatically handling the rewrite_token and looping until the operation is finished.

This extension trait lives in the new `builder_ext` module. This is modeled after the `model_ext` module which contains other convenience functions and types that are built on top of generated types.

For #3204

### DIFF
--- a/src/storage/src/builder_ext.rs
+++ b/src/storage/src/builder_ext.rs
@@ -76,6 +76,7 @@ mod tests {
     use super::*;
     use crate::client::StorageControl;
     use crate::model::{Object, RewriteObjectRequest, RewriteResponse};
+    use gax::error::rpc::{Code, Status};
     use gax::options::RequestOptions;
     use gax::response::Response;
 
@@ -124,6 +125,28 @@ mod tests {
         let result = client.rewrite_object().rewrite_until_done().await?;
 
         assert_eq!(result, final_object);
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_rewrite_until_done_error() -> anyhow::Result<()> {
+        let mut mock = MockStorageControl::new();
+        mock.expect_rewrite_object()
+            .withf(|req: &RewriteObjectRequest, _| req.rewrite_token.is_empty())
+            .times(1)
+            .returning(|_, _| {
+                Err(gax::error::Error::service(
+                    Status::default().set_code(Code::Unavailable),
+                ))
+            });
+
+        let client = StorageControl::from_stub(mock);
+        let _ = client
+            .rewrite_object()
+            .rewrite_until_done()
+            .await
+            .unwrap_err();
+
         Ok(())
     }
 }


suzmue:
This PR still needs tests, but I welcome thoughts on  whether this is 1) the right location and 2) the right way to expose this helper.

dbolduc:
> thoughts

I really like the ergonomics of the trait extension. I think it is nicer than a fn of a builder.

I am cool with it in `model_ext`.

The only thing that isn't obvious to me is whether the loop should have some sort of exhaustion policy. Like if we are making progress too slowly.

I don't think we need a policy. I think even if we want such a policy, we would add it to the `storage::request_options::RequestOptions`. <-- oh wait this doesn't work. I mixed up my clients

coryan:
> > thoughts
> 
> I really like the ergonomics of the trait extension. I think it is nicer than a fn of a builder.

💯 

> I am cool with it in `model_ext`.

We are extending builders, so `builder_ext` seems slightly better to me?

> The only thing that isn't obvious to me is whether the loop should have some sort of exhaustion policy. Like if we are making progress too slowly.
> 
> I don't think we need a policy. I think even if we want such a policy, we would add it to the `storage::request_options::RequestOptions`. <-- oh wait this doesn't work. I mixed up my clients

I would keep this simple.  If we want a loop with exhaustion we can add another extension function (in the future) that transforms the rewrite into a stream.  Then they have all the power of streams to do things like:

```rust
    use futures::stream::StreamExt;
    let builder = client
        .rewrite_object()
        .set_source_bucket(format!("projects/_/buckets/{source_bucket_id}"))
        .set_source_object(SOURCE_NAME)
        .set_destination_bucket(format!("projects/_/buckets/{dest_bucket_id}"))
        .set_destination_name(DEST_NAME);

  let moved = builder
      .to_stream() // A new thing
      .take(100) // at most 100 attempts.
      .map(|v| { /* return some error if the stream is too slow */ })
      .try_fold(None, |object, response| if response.done { response.resource } else { object })
```

A simple function like the one provided in this PR will handle 90% of the cases. We can add the streams for the next 9%, and customers can fallback to an explicit loop for the remaining 1%.

suzmue:
> > I am cool with it in model_ext.

> We are extending builders, so builder_ext seems slightly better to me?

Created `builder_ext` and moved this there.

> This PR still needs tests

Tests are added!

src/storage/src/builder_ext.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// # }
+    /// ```
+    async fn rewrite_until_done(self) -> crate::Result<Option<crate::model::Object>>;
+}
+
+#[async_trait::async_trait]
+impl RewriteObjectExt for crate::builder::storage_control::RewriteObject {
+    async fn rewrite_until_done(mut self) -> crate::Result<Option<crate::model::Object>> {
+        loop {
+            let resp = self.clone().send().await?;
```
dbolduc, line 63: nit: add a test for the error case?

src/storage/src/builder_ext.rs
```diff
@@ -0,0 +1,126 @@
...
+/// An extension trait for `RewriteObject` to provide a convenient way
+/// to poll a rewrite operation until it is complete.
+#[async_trait::async_trait]
+pub trait RewriteObjectExt {
+    /// Sends the request and polls the operation until it is complete.
+    ///
+    /// This helper function simplifies the process of handling a
+    /// [StorageControl::rewrite_object][crate::client::StorageControl::rewrite_object]
+    /// operation, which may require multiple requests to complete for large objects.
+    /// It automatically handles the pagination logic of sending the
```
dbolduc: nit: I am not sure pagination is the right word. How about "continuation logic" instead?

src/storage/src/builder_ext.rs
```diff
@@ -0,0 +1,126 @@
...
+    ///     .await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    async fn rewrite_until_done(self) -> crate::Result<Option<crate::model::Object>>;
+}
+
+#[async_trait::async_trait]
+impl RewriteObjectExt for crate::builder::storage_control::RewriteObject {
+    async fn rewrite_until_done(mut self) -> crate::Result<Option<crate::model::Object>> {
```
dbolduc: This should return a `Result<Object>`. The examples didn't because I was lazy.

src/storage/src/builder_ext.rs
```diff
@@ -0,0 +1,126 @@
...
+        loop {
+            let resp = self.clone().send().await?;
+            if resp.done {
+                return Ok(resp.resource);
```
dbolduc: I think we can `resp.resource.expect("There is always an object returned when the rewrite operation is done")`

src/storage/src/builder_ext.rs
```diff
@@ -0,0 +1,126 @@
...
+                    Ok(Response::from(
+                        RewriteResponse::new()
+                            .set_done(true)
+                            .set_resource(obj.clone()),
+                    ))
+                }
+            });
+
+        let client = StorageControl::from_stub(mock);
+        let result = client.rewrite_object().rewrite_until_done().await.unwrap();
```
dbolduc: style nit: return `anyhow::Result<()>` and use `?` instead of `unwrap` ?

src/storage/src/builder_ext.rs
suzmue: sgtm

src/storage/src/builder_ext.rs
suzmue: done.

src/storage/src/builder_ext.rs
suzmue: done, with the expect changes.

src/storage/src/builder_ext.rs
suzmue: done.

PR #3341, opened 2025-09-16
feat(gax): Add InstrumentationOptions struct

westarle:
This struct will be used to pass client-level information necessary for instrumenting requests with tracing and metrics for #3239 -- this is client-level contract with Librarian to get tracing info from the generator to the client.


westarle:
I want to make this an 'internal' API.

PR #3344, opened 2025-09-16
impl(gax): Add InstrumentationOptions struct

westarle:
This struct will be used to pass client-level information necessary for instrumenting requests with tracing and metrics.  It's intended to be the main interface between sidekick and instrumentation of clients for observability.

For #3239

### DIFF
--- a/src/gax/src/client_builder.rs
+++ b/src/gax/src/client_builder.rs
@@ -386,6 +386,20 @@ impl<F, Cr> ClientBuilder<F, Cr> {
 pub mod internal {
     use super::*;
 
+    /// Options for instrumenting client behavior, primarily for tracing and metrics.
+    #[derive(Clone, Debug, PartialEq, Eq)]
+    #[non_exhaustive]
+    pub struct InstrumentationOptions {
+        /// The short service name, e.g., "appengine", "run", "firestore".
+        pub service_name: String,
+        /// The version of the client library.
+        pub client_version: String,
+        /// The name of the client library artifact (e.g., crate name).
+        pub client_artifact: String,
+        /// The default hostname of the service.
+        pub default_host: String,
+    }
+
     pub trait ClientFactory {
         type Client;
         type Credentials;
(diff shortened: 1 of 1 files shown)


src/gax/src/client_builder.rs
```diff
@@ -386,6 +386,20 @@ impl<F, Cr> ClientBuilder<F, Cr> {
 pub mod internal {
     use super::*;
 
+    /// Options for instrumenting client behavior, primarily for tracing and metrics.
+    #[derive(Clone, Debug, PartialEq, Eq)]
```
westarle, line 390: I will add derive Default to allow people to create these...

src/gax/src/client_builder.rs
```diff
@@ -386,6 +386,20 @@ impl<F, Cr> ClientBuilder<F, Cr> {
...
+    #[non_exhaustive]
+    pub struct InstrumentationOptions {
```
dbolduc, line 392: naming nit: s/Options/Config/ ?

unless this will grow to include option-like things.

dbolduc (approved):
I changed the PR title to use `impl` (as in implementation). We save `feat` for PRs that add customer-facing functionality. We use `impl` for PRs that add internal functionality.

src/gax/src/client_builder.rs
dbolduc: WDYM by people? I assume only our client crates should create this thing. (which is why it is in an `internal` mod).

westarle: Sorry about that, by "people" I meant "client crates".

dbolduc:
can this thing live in `gax-internal` instead of `gax` ?

We have a lot more freedom to change internal types in `gax-internal`

westarle:
> can this thing live in `gax-internal` instead of `gax` ?
> 
> We have a lot more freedom to change internal types in `gax-internal`

Good idea, I will try to pass it directly to ReqwestClient::new (like https://github.com/googleapis/librarian/pull/1981/files#diff-b2ada3703c74e76ca8634c8e1ae2b0686a685f8eba58f698887d4f6ac290fb78)

westarle:
Closing to move to a solution like X-Goog-Api-Client

src/gax/src/client_builder.rs
```diff
@@ -423,6 +437,7 @@ pub mod internal {
         pub retry_throttler: SharedRetryThrottler,
         pub polling_error_policy: Option<Arc<dyn PollingErrorPolicy>>,
         pub polling_backoff_policy: Option<Arc<dyn PollingBackoffPolicy>>,
+        pub instrumentation: Option<InstrumentationOptions>,
```
coryan, line 440: @dbolduc this type is non-exhaustive, so adding a new field is non-breaking. That is intentional to permit this type of addition.

PR #3345, opened 2025-09-16
docs(guide): add rewrite_until_done helper to rewrite guide

### LINKED ISSUE #3204: Add a `rewrite_until_done()` helper

suzmue:
Add a code sample to the user guide that shows how to use the rewrite_until_done helper to direct people to that method if it works for their use case.

Fixes #3204

### DIFF
--- a/guide/samples/tests/storage.rs
+++ b/guide/samples/tests/storage.rs
@@ -54,6 +54,15 @@ pub mod storage {
         }
 
         #[tokio::test(flavor = "multi_thread")]
+        async fn rewrite_object_until_done() -> anyhow::Result<()> {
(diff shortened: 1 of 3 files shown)


dbolduc (approved):
Nice!

PR #3346, opened 2025-09-16
impl(gax): add path_template to RequestOptions

westarle:
This sets up request-level details for #3239 to select a low cardinality path template to include in span names etc.; the path template is part of the contract for sidekick to populate.

### DIFF
--- a/src/gax/src/options.rs
+++ b/src/gax/src/options.rs
@@ -50,7 +50,6 @@ pub struct RequestOptions {
     polling_error_policy: Option<Arc<dyn PollingErrorPolicy>>,
     polling_backoff_policy: Option<Arc<dyn PollingBackoffPolicy>>,
     path_template: Option<String>,
-    prior_attempt_count: u32,
 }
 
 impl RequestOptions {
@@ -165,16 +164,6 @@ impl RequestOptions {
     pub(crate) fn set_path_template(&mut self, v: Option<String>) {
         self.path_template = v;
     }
-
-    /// Get the current prior attempt count.
-    pub(crate) fn prior_attempt_count(&self) -> u32 {
-        self.prior_attempt_count
-    }
-
-    /// Sets the prior attempt count for the request.
-    pub(crate) fn set_prior_attempt_count(&mut self, v: u32) {
-        self.prior_attempt_count = v;
-    }
 }
 
 /// Implementations of this trait provide setters to configure request options.
@@ -244,11 +233,6 @@ pub mod internal {
 
     pub fn get_path_template(options: &RequestOptions) -> Option<&str> {
         options.path_template()
-    }
-
-    pub fn increment_prior_attempt_count(mut options: RequestOptions) -> RequestOptions {
-        options.set_prior_attempt_count(options.prior_attempt_count() + 1);
-        options
     }
 }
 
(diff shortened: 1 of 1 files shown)


src/gax/src/options.rs
```diff
@@ -188,6 +210,12 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
 
     /// Sets the polling backoff policy configuration.
     fn with_polling_backoff_policy<V: Into<PollingBackoffPolicyArg>>(self, v: V) -> Self;
+
+    /// Sets the path template for the request URL.
+    fn with_path_template(self, v: Option<String>) -> Self;
+
+    /// Sets the prior attempt count for the request.
+    fn with_prior_attempt_count(self, v: u32) -> Self;
```
coryan: You can make this change non-breaking using:


```suggestion
    /// Sets the path template for the request URL.
    fn with_path_template(self, _v: Option<String>) -> Self { self }

    /// Sets the prior attempt count for the request.
    fn with_prior_attempt_count(self, _v: u32) -> Self { self }
```

But we should consider if these methods belong in the public API exposed to customers (I recall methods added here  automatically become part of the public API). I think they should not be public.  This is not an attribute of the request applications should be able to set or would want to set.

westarle: I'd prefer they are not in the public API -- I can follow the example for default_idempotency where accessors are in the ::internal module.

src/gax/src/options.rs
```diff
@@ -153,6 +155,26 @@ impl RequestOptions {
...
+
+    /// Get the current prior attempt count.
+    pub(crate) fn prior_attempt_count(&self) -> u32 {
+        self.prior_attempt_count
+    }
+
+    /// Sets the prior attempt count for the request.
+    pub(crate) fn set_prior_attempt_count(&mut self, v: u32) {
+        self.prior_attempt_count = v;
+    }
```
coryan: I am not sure this helps or not.  I believe you can keep this out of the `RequestOptions`.  The retry loop receives a mutable callable:

https://github.com/googleapis/google-cloud-rust/blob/f8103272373681e40170c0bb1ff4a028f2138e72/src/gax/src/retry_loop_internal.rs#L52

So you could change the call site(s) from this:

https://github.com/googleapis/google-cloud-rust/blob/f8103272373681e40170c0bb1ff4a028f2138e72/src/gax-internal/src/http.rs#L130-L135

To something more like this:

```rust
        let mut count = 0_u64;
        let inner = async move |d| {
            let attempts = { let c = count; count += 1; c};
            let builder = builder
                .try_clone()
                .expect("client libraries only create builders where `try_clone()` succeeds");
            this.request_attempt(builder, &options, attempts, d).await
        };
```

Or some variation of your liking, maybe with `Option<i32>`... 

To be clear, I am not saying my suggestion is better, just worth thinking about.

westarle: I will try it out!  For now, I removed it from the change -- I think I need this API for path_template.

westarle: I also had this maybe more correct but kind of ugly change that worked:

https://github.com/googleapis/google-cloud-rust/compare/main...westarle:google-cloud-rust:o12y-refactor-add-attempt-count

coryan: > I also had this maybe more correct but kind of ugly change that worked:
> 
> [main...westarle:google-cloud-rust:o12y-refactor-add-attempt-count](https://github.com/googleapis/google-cloud-rust/compare/main...westarle:google-cloud-rust:o12y-refactor-add-attempt-count)

I think that breaks an internal API, which requires a major version bump, even if it does not affect customer code. Something we *can* do, but should avoid.

coryan (approved):

PR #3347, opened 2025-09-16
impl(gax-internal): define InstrumentationClientInfo for tracing

westarle:
This structure forms an API to be populated and added by the generator (like the implementation of x-goog-api-client.)

It will be used by instrumentation to describe the client library itself, for #3239

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -39,6 +39,7 @@ pub struct ReqwestClient {
     retry_throttler: SharedRetryThrottler,
     polling_error_policy: Arc<dyn PollingErrorPolicy>,
     polling_backoff_policy: Arc<dyn PollingBackoffPolicy>,
+    instrumentation: Option<&'static crate::options::InstrumentationClientInfo>,
 }
 
 impl ReqwestClient {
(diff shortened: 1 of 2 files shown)


westarle:
@dbolduc fyi

src/gax-internal/src/http.rs
```diff
@@ -39,6 +39,7 @@ pub struct ReqwestClient {
     retry_throttler: SharedRetryThrottler,
     polling_error_policy: Arc<dyn PollingErrorPolicy>,
     polling_backoff_policy: Arc<dyn PollingBackoffPolicy>,
+    instrumentation: Option<&'static crate::options::InstrumentationClientInfo>,
```
coryan: Unless there is a really good reason, I think making a copy here is just fine.  a `ReqwestClient` should be long-lived, and typically wrapped in an `Arc`, so very few copies of it should be taking place.

src/gax-internal/src/options.rs
```diff
@@ -19,6 +19,19 @@ pub type ClientConfig = gax::client_builder::internal::ClientConfig<Credentials>
 
 pub(crate) const LOGGING_VAR: &str = "GOOGLE_CLOUD_RUST_LOGGING";
 
+/// Information about the client library used for instrumentation.
+#[derive(Clone, Debug)]
```
coryan: I believe we can also make this `#[derive(Clone, Copy, Debug)]` since it is just a collection of `static` references.  If we think copying may become expensive in some distant future then just `Clone` is safer.

coryan (commented):
Drive-by comments.

src/gax-internal/src/http.rs
westarle: Done

src/gax-internal/src/options.rs
westarle: Good idea, done!

Issue #3010, opened 2025-08-24
Generate topic admin client with no data-plane operations


coryan:
A generated client with the data plane operations excluded. Or rather, explicitly list the admin operations so the generator automatically generates the minimum set of messages to support them.

suzmue:
Fixed by https://github.com/googleapis/google-cloud-rust/pull/3126

PR #3349, opened 2025-09-16
fix(pubsub): rename Publisher and Subscription admin clients

suzmue:
The admin clients should be renamed to match their function. This is consistent with the Java and Go client libraries, though theirs are named `TopicAdminClient` and `SubscriptionAdminClient`. Since these types are within a `client` module, we do not use the suffix since its repetitive.

For #3011 and for #3010


coryan:
s/fix(storage)/fix(pubsub)/ in the PR title? 😁

suzmue:
> s/fix(storage)/fix(pubsub)/ in the PR title? 😁

LOL autopilot kicking in I guess 😅

PR #3351, opened 2025-09-17
test(bigquery/v2): update integration test with pagination

### LINKED ISSUE #1998: Handle pagination for non-standard pagination services
Some services such as SQL and BigQuery do not follow the pagination standard [AIP-4233](https://google.aip.dev/client-libraries/4233) such as using max_results instead of page_size and wrapping the field in an additional message layer.
As such, paginator is not generated for these List operations. We should look how to handle these special cases.

PhongChuong:
Regenerate bigquery/v2 with pagination and update integration test.

Fixes #1998


dbolduc:
We need to update the `librarian` version in the CI for the `regenerate` builds to pass.

https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-generated-code-maintenance.md#making-changes-to-sidekick

coryan:
> We need to update the `librarian` version in the CI for the `regenerate` builds to pass.
> 
> https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-generated-code-maintenance.md#making-changes-to-sidekick

Blegh, and I just merged my changes, which means it is harder to discover the right version.

dbolduc:
> Blegh, and I just merged my changes, which means it is harder to discover the right version.

@coryan - Send a PR updating librarian to latest and generate all the things.
@PhongChuong - after that goes through, rebase this, and change it to a `test(bigquery)` PR?

PhongChuong:
> > Blegh, and I just merged my changes, which means it is harder to discover the right version.
> 
> @coryan - Send a PR updating librarian to latest and generate all the things. @PhongChuong - after that goes through, rebase this, and change it to a `test(bigquery)` PR?

SGTM.

coryan:
> > > Blegh, and I just merged my changes, which means it is harder to discover the right version.
> > 
> > 
> > @coryan - Send a PR updating librarian to latest and generate all the things. @PhongChuong - after that goes through, rebase this, and change it to a `test(bigquery)` PR?
> 
> SGTM.

Will do.  For the record, the sidekick version with only @PhongChuong 's changes is: `v0.1.2-0.20250918140618-f7a7c8e2b04e`

src/integration-tests/src/bigquery.rs
```diff
@@ -151,16 +165,153 @@ async fn cleanup_stale_datasets(
...
+            .with_thread_ids(true)
+            .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+            .finish();
+
+        tracing::subscriber::set_default(subscriber)
+    };
+
+    let project_id = crate::project_id()?;
+    let client: bigquery::client::JobService = builder.build().await?;
+    cleanup_stale_jobs(&client, &project_id).await?;
```
alvarowolfx, line 207: I'm not sure if we want to keep cleaning jobs, as I don't remember exactly the behavior of the API if you try to delete a job that is still running, so that can turn this test suite flaky. I think we can create jobs just fine and when listing, we just filter by running ones or something like that. 

See some of the recent samples that we created in Go for jobs list: https://github.com/GoogleCloudPlatform/golang-samples/pull/5366/files#diff-03e30650743f3d65787bfc2f4431593b414d371cb329c2e49d94787c6188bf51R30

alvarowolfx (changes requested):
I think we can simplify the jobs service test by not worrying on cleaning up jobs (as this can lead to flaky test, as you might try to delete a running job) and also agree with @coryan that this should be renamed as `test(bigquery/v2)` PR

src/integration-tests/src/bigquery.rs
PhongChuong: If the concern is deleting jobs that is in the RUNNING state, I've updated the cleanup code to only delete jobs whose state == "DONE". I've also remove the delete job from the main test block so in the rare case that the query does not complete in time, we don't run into a flake.

The only race condition that might occur is if a test runs for greater than (60*60*48) secs (2 days) and the job that it created is deleted by another test execution. In this case, I would argue that the test should fail.

PTAL.

alvarowolfx: There is the concern of running jobs, which is addressed by filtering by `DONE` jobs, but we don't usually worry about cleaning up jobs in general, which ended up catching my attention. But might not hurt that much.

alvarowolfx (approved):

coryan:
> I think we can simplify the jobs service test by not worrying on cleaning up jobs (as this can lead to flaky test, as you might try to delete a running job) and also agree with @coryan that this should be renamed as `test(bigquery/v2)` PR

In some services the `list()` tests can get really slow if you don't cleanup things.... and you also may hit quota limits.  Neither of those concerns may apply to BQ.

src/integration-tests/src/bigquery.rs
PhongChuong: Thanks for the review.

PR #3354, opened 2025-09-17
WIP DRAFT: feat(pubsub): expose a publisher client

suzmue:
Starting on the publisher client, mostly focusing on trying to generate the Publish method without exposing it.

There is some weirdness with using the client template is that it ends up referencing types that we expose from the other generation, but we still have the generated types from that client. Might want to create a new template where we can handle that better (or one might exist I don't know about).

Also just using the http client for now because I didn't want to deal with adding grpc dependencies yet.

The quickstart was easy and works to publish a single message without batching. (the interface is not great as it is but was just created to have a simple pass through public client).


suzmue:
Generated using 
```sh
$ go run github.com/googleapis/librarian/cmd/sidekick@main refresh     -output src/pubsub/src/generated/gapic_internal/ && cargo fmt -p google-cloud-pubsub
```

and 
```sh
go run github.com/googleapis/librarian/cmd/sidekick@main refresh     -output src/pubsub/src/generated/gapic/ && cargo fmt -p google-cloud-pubsub
```

PR #3356, opened 2025-09-17
test(pubsub): add simple integration test

suzmue:
Adds a very simple integration test for the TopicAdmin client, adding some setup and cleanup functions that will be useful later when testing the publisher and subscriber.

There is still remaining work to cleanup stale topics that are not deleted because the test failed.

For #3355

### DIFF
--- a/src/integration-tests/src/pubsub.rs
+++ b/src/integration-tests/src/pubsub.rs
@@ -32,17 +32,16 @@ pub async fn basic_topic() -> Result<()> {
         tracing::subscriber::set_default(subscriber)
     };
 
-    let project = crate::project_id()?;
     let (client, topic) = create_test_topic().await?;
 
-    tracing::info!("testing list_topics()");
-    let topics = client
-        .list_topics()
-        .set_project(format!("projects/{project}"))
+    tracing::info!("testing get_topic()");
+    let get_topic = client
+        .get_topic()
+        .set_topic(topic.name.clone())
         .send()
         .await?;
-    tracing::info!("success with list_topics={topics:?}");
-    assert!(topics.topics.iter().any(|x| x.name == topic.name));
+    tracing::info!("success with get_topic={get_topic:?}");
+    assert_eq!(get_topic.name, topic.name);
 
     cleanup_test_topic(client, topic.name).await?;
 
@@ -51,7 +50,7 @@ pub async fn basic_topic() -> Result<()> {
 
 pub const TOPIC_ID_LENGTH: usize = 255;
 
-fn random_topic_id(project: String) -> String {
+fn random_topic_name(project: String) -> String {
     let prefix = "topic-";
     let topic_id: String = rand::rng()
         .sample_iter(&Alphanumeric)
@@ -67,7 +66,7 @@ pub async fn create_test_topic() -> Result<(TopicAdmin, Topic)> {
         .with_tracing()
         .build()
         .await?;
-    let topic_id = random_topic_id(project);
+    let topic_id = random_topic_name(project);
 
     tracing::info!("testing create_topic()");
     let topic = client
--- a/src/integration-tests/src/pubsub.rs
+++ b/src/integration-tests/src/pubsub.rs
@@ -66,12 +66,12 @@ pub async fn create_test_topic() -> Result<(TopicAdmin, Topic)> {
         .with_tracing()
         .build()
         .await?;
-    let topic_id = random_topic_name(project);
+    let topic_name = random_topic_name(project);
 
     tracing::info!("testing create_topic()");
     let topic = client
         .create_topic()
-        .set_name(topic_id)
+        .set_name(topic_name)
         .set_labels([("integration-test", "true")])
         .send()
         .await?;
--- a/src/integration-tests/src/pubsub.rs
+++ b/src/integration-tests/src/pubsub.rs
@@ -43,9 +43,9 @@ pub async fn basic_topic(builder: pubsub::builder::topic_admin::ClientBuilder) -
         .send()
         .await?;
     tracing::info!("success with list_topics={topics:?}");
-    assert!(topics.topics.iter().any(|x| x.name == topic.name.clone()));
+    assert!(topics.topics.iter().any(|x| x.name == topic.name));
 
-    cleanup_test_topic(topic_admin, topic.name.clone()).await?;
+    cleanup_test_topic(topic_admin, topic.name).await?;
 
     Ok(())
 }
@@ -68,12 +68,12 @@ pub async fn create_test_topic() -> Result<(TopicAdmin, Topic)> {
         .with_tracing()
         .build()
         .await?;
-    let topic_id = random_topic_id(project.clone());
+    let topic_id = random_topic_id(project);
 
     tracing::info!("testing create_topic()");
     let topic = client
         .create_topic()
-        .set_name(topic_id.clone())
+        .set_name(topic_id)
         .set_labels([("integration-test", "true")])
         .send()
         .await?;
--- a/src/integration-tests/src/pubsub.rs
+++ b/src/integration-tests/src/pubsub.rs
@@ -1,0 +1,90 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


coryan (approved):
I wonder if it is time to move the veneer integration tests to the veneer directories.

suzmue:
> I wonder if it is time to move the veneer integration tests to the veneer directories.

Filed #3358 for that

.gcb/integration.yaml
```diff
@@ -40,7 +40,7 @@ steps:
       #!/usr/bin/env bash
       set -e
       cargo build
-      cargo test --package integration-tests --features run-integration-tests
+      cargo test --package integration-tests --features run-integration-tests --features log-integration-tests pubsub
```
dbolduc: pubsub only? :monocle_face:

src/integration-tests/tests/driver.rs
```diff
@@ -47,6 +47,18 @@ mod driver {
             .map_err(integration_tests::report_error)
     }
 
+    #[test_case(pubsub::client::TopicAdmin::builder(); "default")]
+    #[test_case(pubsub::client::TopicAdmin::builder().with_tracing(); "with tracing enabled")]
+    #[test_case(pubsub::client::TopicAdmin::builder().with_retry_policy(retry_policy()); "with retry enabled")]
```
coryan: Consider simplifying this.  We have been cargo-culting these since I created the original integration tests for secret manager, it made sense at the time to make sure all the generated code paths were tested, but by now?  Probably one test with the retry policy enabled is good enough.

.gcb/integration.yaml
suzmue: This was for testing :) Already backed out

src/integration-tests/src/pubsub.rs
```diff
@@ -0,0 +1,90 @@
...
+    let (topic_admin, topic) = create_test_topic().await?;
+
+    tracing::info!("testing list_topics()");
+    let topics = client
+        .list_topics()
+        .set_project(format!("projects/{project}"))
+        .send()
+        .await?;
+    tracing::info!("success with list_topics={topics:?}");
+    assert!(topics.topics.iter().any(|x| x.name == topic.name.clone()));
```
dbolduc: This is assuming the new topic shows up in the first page of results for `ListTopics`. Can we iterate over `builder::ListTopics::by_item()`? e.g. like

https://github.com/googleapis/google-cloud-rust/blob/410578003c15bb512066386db6bdefb6baba51c9/src/integration-tests/src/secret_manager/protobuf.rs#L136

src/integration-tests/src/pubsub.rs
```diff
@@ -0,0 +1,90 @@
...
+
+    cleanup_test_topic(topic_admin, topic.name.clone()).await?;
+
+    Ok(())
+}
+
+pub const TOPIC_ID_LENGTH: usize = 255;
+
+fn random_topic_id(project: String) -> String {
```
dbolduc: nit: this is generating a topic name, not a topic ID.

The name is the full resource, which includes the project. The ID is just the end bit.

src/integration-tests/src/pubsub.rs
```diff
@@ -0,0 +1,90 @@
...
+    let client = pubsub::client::TopicAdmin::builder()
+        .with_tracing()
+        .build()
+        .await?;
+    let topic_id = random_topic_id(project.clone());
+
+    tracing::info!("testing create_topic()");
+    let topic = client
+        .create_topic()
+        .set_name(topic_id.clone())
```
dbolduc: nit: unnecessary `clone()`

src/integration-tests/src/pubsub.rs
```diff
@@ -0,0 +1,90 @@
...
+    format!("projects/{project}/topics/{prefix}{topic_id}")
+}
+
+pub async fn create_test_topic() -> Result<(TopicAdmin, Topic)> {
+    let project = crate::project_id()?;
+    let client = pubsub::client::TopicAdmin::builder()
+        .with_tracing()
+        .build()
+        .await?;
+    let topic_id = random_topic_id(project.clone());
```
dbolduc: same.

src/integration-tests/src/pubsub.rs
```diff
@@ -0,0 +1,90 @@
...
+            .with_thread_ids(true)
+            .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+            .finish();
+
+        tracing::subscriber::set_default(subscriber)
+    };
+
+    let project = crate::project_id()?;
+    let client = builder.build().await?;
+    let (topic_admin, topic) = create_test_topic().await?;
```
coryan: we usually add some kind of cleanup to delete stale topics from previous tests that crashed or otherwise failed to delete any topics they created. Should we do that here?

coryan (changes requested):

src/integration-tests/src/pubsub.rs
```diff
@@ -0,0 +1,90 @@
...
+    tracing::info!("testing list_topics()");
+    let topics = client
+        .list_topics()
+        .set_project(format!("projects/{project}"))
+        .send()
+        .await?;
+    tracing::info!("success with list_topics={topics:?}");
+    assert!(topics.topics.iter().any(|x| x.name == topic.name.clone()));
+
+    cleanup_test_topic(topic_admin, topic.name.clone()).await?;
```
dbolduc: same

src/integration-tests/src/pubsub.rs
dbolduc: And we shouldn't need to clone the field if we are just checking for equality. We can borrow it.

src/integration-tests/tests/driver.rs
dbolduc: nit: the clients all have a default retry policy now. so this test case is not sooo useful.

I can clean up the other existing cases.

dbolduc: I can clean this up for the existing test cases.

src/integration-tests/src/pubsub.rs
suzmue: I called that out as something I wanted to do in a followup PR in the description, but can do so in this PR if we want that.

src/integration-tests/tests/driver.rs
suzmue: Simplified to just use the default.

suzmue: Cleaned up to remove this and just use the client generated in the test.

src/integration-tests/src/pubsub.rs
suzmue: In the spirit of making this the most basic of tests, I switched to get_topic

src/integration-tests/src/pubsub.rs
suzmue: Done.

src/integration-tests/src/pubsub.rs
suzmue: The reason for delaying is that (AFAICT) there is not a field that I could use to just check the creation time on a topic so it wasn't quite as straightforward as with the others. And I was planning to add a create time label or something to hack around that

suzmue: Could also attach a schema that has a revision create_time

coryan: SGTM, sorry I missed the description.

src/integration-tests/src/pubsub.rs
```diff
@@ -0,0 +1,88 @@
...
+    format!("projects/{project}/topics/{prefix}{topic_id}")
+}
+
+pub async fn create_test_topic() -> Result<(TopicAdmin, Topic)> {
+    let project = crate::project_id()?;
+    let client = pubsub::client::TopicAdmin::builder()
+        .with_tracing()
+        .build()
+        .await?;
+    let topic_id = random_topic_name(project);
```
dbolduc: nit: s/topic_id/topic_name/

dbolduc (approved):

src/integration-tests/src/pubsub.rs
suzmue: oops done.

PR #3366, opened 2025-09-19
cleanup: reduce integration test matrix wrt client config

dbolduc:
Instead of testing every client with/without tracing, and with/without a custom retry policy, just test these dimensions in `secretmanager`, `storagecontrol`.

By default, enable tracing in the clients. We would like to know if anything goes wrong.

Note that we don't need to set the retry policy, as clients come with a default policy now.

### DIFF
--- a/src/integration-tests/src/bigquery.rs
+++ b/src/integration-tests/src/bigquery.rs
@@ -19,9 +19,7 @@ use rand::{Rng, distr::Alphanumeric};
 
 const INSTANCE_LABEL: &str = "rust-sdk-integration-test";
 
-pub async fn dataset_admin(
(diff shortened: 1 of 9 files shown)


coryan:
> Instead of testing every client with/without tracing, and with/without a custom retry policy, just test these dimensions in `secretmanager`, `storagecontrol`.

As you know, `secretmanager` does not have LROs via HTTP+JSON. Do you think the coverage drop is a problem?

dbolduc:
> LROs [...]  Do you think the coverage drop is a problem?

We were not setting custom **polling** policies to begin with. So no.

I don't think the individual RPCs made in a polling loop are special.

coryan (approved):

PR #3367, opened 2025-09-19
impl(gax-internal): Define OpenTelemetry key constants for HttpSpanInfo

westarle:
for #3239

### DIFF
--- a/src/gax-internal/src/observability.rs
+++ b/src/gax-internal/src/observability.rs
@@ -17,6 +17,50 @@
 use crate::options::InstrumentationClientInfo;
 use gax::options::RequestOptions;
 
+// OpenTelemetry Semantic Convention Keys
+// See https://opentelemetry.io/docs/specs/semconv/http/http-spans/
+
+/// Span Kind for OpenTelemetry interop. Always CLIENT for a span representing an outbound HTTP request.
+const KEY_OTEL_KIND: &str = "otel.kind";
+/// Span Name for OpenTelemetry interop. Recommended to be "{http.request.method} {url.template}" if url.template is available, otherwise "{http.request.method}".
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/observability.rs
```diff
@@ -17,6 +17,50 @@
 use crate::options::InstrumentationClientInfo;
 use gax::options::RequestOptions;
 
+// OpenTelemetry Semantic Convention Keys
+// See https://opentelemetry.io/docs/specs/semconv/http/http-spans/
+
+/// Span Kind for OpenTelemetry interop. Always CLIENT for a span representing an outbound HTTP request.
+const KEY_OTEL_KIND: &str = "otel.kind";
+/// Span Name for OpenTelemetry interop. Recommended to be "{http.request.method} {url.template}" if url.template is available, otherwise "{http.request.method}".
```
coryan: And I would avoid the passive voice, and format these differently:


```suggestion
/// Span Name for OpenTelemetry interop.
///
/// If `url.template` is available use "{http.request.method} {url.template}", otherwise use "{http.request.method}".
```

coryan (dismissed):

src/gax-internal/src/observability.rs
westarle: Thank you, I updated all the docs, let me know if it's overkill.

coryan (approved):

src/gax-internal/src/observability.rs
coryan: Before I forget, should we be using:

https://docs.rs/opentelemetry-semantic-conventions/latest/opentelemetry_semantic_conventions/trace/index.html

PR #3385, opened 2025-09-23
cleanup: consolidate CI envs

dbolduc:


coryan (dismissed):

.gcb/integration.yaml
```diff
@@ -19,7 +19,6 @@ options:
   env:
     - GOOGLE_CLOUD_PROJECT=${PROJECT_ID}
     - GOOGLE_CLOUD_RUST_TEST_SERVICE_ACCOUNT=set-iam-test-only@${PROJECT_ID}.iam.gserviceaccount.com
-    - GOOGLE_CLOUD_RUST_TEST_WORKFLOWS_RUNNER=test-workflow-runner@${PROJECT_ID}.iam.gserviceaccount.com
```
coryan, line 22: Ah, now I remember, I was trying to segregate the accounts by the permissions we need in them.

dbolduc: they actually have the same permissions in terms of what the SA's are able to do.

The difference is whether the `integration-test-runner` has permission to act as them.

I am sending a CL.

coryan: > they actually have the same permissions in terms of what the SA's are able to do.

I think `set-iam-test-only@` at least temporarily gains extra permissions to do other things. We add and remove that account to resources as part of the `SetIamPolicy()` tests.

I am not religious about permissions, but merging these two accounts seems like risk we do not need?

dbolduc: > seems like risk

I don't see the risk. The SA is disabled. It becomes an owner of empty test objects and buckets. And it is tied to short-lived workflows, but it does not even have permission to run the workflows.

PTAL

coryan: Agreee

coryan (approved):

PR #3395, opened 2025-09-23
cleanup: update generated code for OpenAPI

coryan:
We have not released any client based on an OpenAPI spec, so this change
is by definition non-breaking.


coryan:
I should break this down in two PRs... give me a minute.

coryan:
PTAL.

PR #3396, opened 2025-09-23
feat: Add package version constraints for all dependencies

brianquinlan:


coryan:
I suspect you will want to change this:

https://github.com/googleapis/google-cloud-rust/blob/8a362affed927930528caf753b488da34bad1e36/.github/workflows/generator_dart.yaml#L60

To use the version found with:

```shell
go list -m -u -f '{{.Version}}' github.com/googleapis/librarian@main
```

It also looks like a `taplo fmt dart/.sidekick.toml` may help.

brianquinlan:
> I suspect you will want to change this:
> 
> https://github.com/googleapis/google-cloud-rust/blob/8a362affed927930528caf753b488da34bad1e36/.github/workflows/generator_dart.yaml#L60
> 
> To use the version found with:
> 
> ```shell
> go list -m -u -f '{{.Version}}' github.com/googleapis/librarian@main
> ```
> 

I don't care that the generated code is out-of-date, I just want the configuration to be correct for https://github.com/googleapis/librarian/pull/2331

> It also looks like a `taplo fmt dart/.sidekick.toml` may help.

Done.

brianquinlan:
@devoncarew

dart/.sidekick.toml
```diff
@@ -33,6 +33,14 @@ version = "0.1.0"
 'package:google_cloud_gax' = '^0.1.0'
 'package:http'             = '^1.3.0'
 
+# Generated packages
```
devoncarew, line 36: I assume these are the ones that are currently referenced by other (generated) packages.

devoncarew (commented):
Did you mean to include the LICENSE files as well?

dart/.sidekick.toml
brianquinlan: Yes. Maybe
```suggestion
# Generated packages used as dependencies.
```

?

brianquinlan:
> Did you mean to include the LICENSE files as well?

For the CI to pass, I had to re-run the generator, which produced the LICENSE files.

dart/.sidekick.toml
devoncarew: I'm fine w/ either, I was just confirming this for my own understanding.

coryan:
> > Did you mean to include the LICENSE files as well?
> 
> For the CI to pass, I had to re-run the generator, which produced the LICENSE files.

It did not occur to me earlier, but you could change the CI build to skip checking of the LICENSE files.  The CI scripts are right here in this repo.

brianquinlan:
> > > Did you mean to include the LICENSE files as well?
> > 
> > 
> > For the CI to pass, I had to re-run the generator, which produced the LICENSE files.
> 
> It did not occur to me earlier, but you could change the CI build to skip checking of the LICENSE files. The CI scripts are right here in this repo.

Since the LICENSE files are generated, shouldn't they be checked just like everything else?

Hopefully this will be the last time that we need to update this code before we have a separate repo for dart but isn't the idea that no checkins should be allowed where the generated code doesn't match that produced by the CI?

coryan:
> 
> Since the LICENSE files are generated, shouldn't they be checked just like everything else?

Yes, you are right. Forget what I said.

Issue #3375, opened 2025-09-22
Default `Host` header does not work for regional endpoints


coryan:
Reportedly "gemini inference" (I assume that is one of the services in Vertex AI aka google-cloud-aiplatform-v1) does not work with regional endpoints because the `Host` header is not quite right.

scurkovic:
@coryan Just wanted to add more details. Indeed, this is referring to Vertex AI. I have a fine tuned Gemini model that I am running inference on. As per the docs, the endpoint is `https://europe-west1-aiplatform.googleapis.com`.

Unfortunately, this started failing (downgrading `google-cloud-aiplatform-v1` to `0.4.5` fixes the issue) with:

```Endpoint `projects/<project_id>/locations/europe-west1/endpoints/<endpoint_id>` not found.```

[This patch](https://github.com/cyanic-selkie/google-cloud-rust/tree/fix/regional-endpoint-routing) I made resolves my particular issue, but like you said, it won't work with some other services.

coryan:
Ugh, I had completely forgotten about the locational (`{region}-{service}.googleapis.com`) vs. regional (`{service}.{region}.rep.googleapis.com`) endpoints.

We have tests for regional endpoints, we do not have tests for locational endpoints.  That is the obvious place to start testing things.

dbolduc:
Hey @cyanic-selkie, thanks for the detailed bug report.

We were able to reproduce and fix the problem (#3402). We also added [a test](https://github.com/googleapis/google-cloud-rust/blob/4964d220d74b0c9d010bc5ccbb98323622057491/src/integration-tests/src/aiplatform.rs#L18-L32) to verify using locational endpoints with Vertex AI in our CI/CD.

The fix will be included in the next version (`1.1.0`) of `google-cloud-aiplatform-v1`, which we are planning to publish tomorrow.

scurkovic:
@dbolduc Thank you very much, I'll give it a try!

PR #3402, opened 2025-09-24
fix: support locational endpoints

dbolduc:
Part of the work for #3375. (The other half is gRPC).

When a client is configured to use a locational / regional endpoints, use them as the host, instead of the default host (global).

The client library detects these automatically. (Whereas, in C++, we decided to expose an `AuthorityOption`). If you have more test cases for custom endpoints, I will happily add them.

Adds an integration test that fails without the corresponding changes to the `gaxi` client.

I did not think about whether we should do the same thing for gRPC based clients. The answer is probably. It would be easy enough to refactor `host_from_endpoint` and the related tests.

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -49,12 +49,7 @@ impl ReqwestClient {
     ) -> gax::client_builder::Result<Self> {
         let cred = Self::make_credentials(&config).await?;
         let inner = reqwest::Client::new();
-        let uri = Uri::from_str(default_endpoint).map_err(BuilderError::transport)?;
-        let host = uri
-            .authority()
-            .ok_or_else(|| BuilderError::transport("missing authority in default endpoint"))?
-            .host()
-            .to_string();
+        let host = host_from_endpoint(config.endpoint.as_deref(), default_endpoint)?;
         let endpoint = config
             .endpoint
             .unwrap_or_else(|| default_endpoint.to_string());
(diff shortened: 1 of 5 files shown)


src/gax-internal/src/http.rs
```diff
@@ -451,4 +484,45 @@ mod tests {
...
+
+    #[test_case(None, "test.googleapis.com"; "default")]
+    #[test_case(Some("http://www.googleapis.com"), "test.googleapis.com"; "global")]
+    #[test_case(Some("http://private.googleapis.com"), "test.googleapis.com"; "VPC-SC private")]
+    #[test_case(Some("http://restricted.googleapis.com"), "test.googleapis.com"; "VPC-SC restricted")]
+    #[test_case(Some("http://test-my-private-ep.p.googleapis.com"), "test.googleapis.com"; "PSC custom endpoint")]
+    #[test_case(Some("https://us-central1-test.googleapis.com"), "us-central1-test.googleapis.com"; "locational endpoint")]
+    #[test_case(Some("https://test.us-central1.rep.googleapis.com"), "test.us-central1.rep.googleapis.com"; "regional endpoint")]
+    #[test_case(Some("https://test.my-universe-domain.com"), "test.googleapis.com"; "universe domain")]
+    #[test_case(Some("localhost:5678"), "test.googleapis.com"; "emulator")]
```
coryan, line 496: From go/rep-mtls, but I am not sure if these are finalized:

* `content-{service}.{region}.rep.mtls.googleapis.com`
* `{service}.{region}.rep.mtls.googleapis.com`

coryan (approved):
> I did not think about whether we should do the same thing for gRPC based clients. The answer is probably.

I agree.

> It would be easy enough to refactor host_from_endpoint and the related tests.

https://cloud.google.com/storage/docs/regional-endpoints

So maybe we should do that now?

coryan:
And I forgot these: https://cloud.google.com/storage/docs/locational-endpoints

src/gax-internal/src/http.rs
dbolduc: Ack. As it stands we will use the default host for these, which might just work. (In my testing, both the default host and the regional host worked for regional endpoints).

dbolduc:
> So maybe we should do that now?

Definitely. I will merge this PR because it adds functionality and isn't obviously wrong. Then probably send a refactor PR. Then send another PR that adds support to the gRPC client (with integration tests).

PR #3406, opened 2025-09-25
WIP: pubsub gapic_dataplane generation

PhongChuong:


src/pubsub/src/generated/gapic_dataplane/builder.rs
```diff
@@ -0,0 +1,118 @@
...
+    ///
+    /// let builder = prepare_request_builder();
+    /// let response = builder.send().await?;
+    /// # gax::Result::<()>::Ok(()) });
+    ///
+    /// fn prepare_request_builder() -> Publish {
+    ///   # panic!();
+    ///   // ... details omitted ...
+    /// }
+    /// ```
```
PhongChuong, line 59: Do not generate if HasVeneer?

src/pubsub/src/generated/gapic/model.rs
```diff
@@ -2679,6 +2679,110 @@ pub mod topic {
     }
 }
 
+/// A message that is published by publishers and consumed by subscribers. The
+/// message must contain either a non-empty data field or at least one attribute.
+/// Note that client libraries represent this object differently
+/// depending on the language. See the corresponding [client library
+/// documentation](https://cloud.google.com/pubsub/docs/reference/libraries) for
+/// more information. See [quotas and limits]
+/// (<https://cloud.google.com/pubsub/quotas>) for more information about message
```
coryan: no action required: the link is bad because it is broken across two lines, maybe https://github.com/googleapis/librarian/issues/1600 would fix such problems.

src/pubsub/src/generated/gapic/.sidekick.toml
```diff
@@ -18,6 +18,8 @@ service-config       = 'google/pubsub/v1/pubsub_v1.yaml'
 
 [source]
 included-ids = """\
+    .google.pubsub.v1.PublishRequest,\
+    .google.pubsub.v1.PublishResponse,\
```
coryan: Why add the messages to the control plane APIs?

src/pubsub/src/generated/gapic/model.rs
PhongChuong: Noted.

src/pubsub/src/generated/gapic/.sidekick.toml
PhongChuong: Removed.

PhongChuong:
Updated to use grpc instead for dataplane. PTAL.

src/pubsub/src/lib.rs
```diff
@@ -29,10 +29,22 @@ pub(crate) mod generated;
...
+    pub use crate::generated::gapic::builder::*;
+    pub use crate::generated::gapic_dataplane::builder::*;
+}
+pub mod model {
+    pub use crate::generated::gapic::model::*;
+    pub use crate::generated::gapic_dataplane::model::*;
+}
+pub mod client {
+    pub use crate::generated::gapic::client::*;
+    pub use crate::generated::gapic_dataplane::client::*;
```
suzmue, line 42: I think we want to keep this crate private for now. Probably all of the gapic_dataplane pub use need the same.


```
pub(crate) use crate::generated::gapic_dataplane::client::*;
```

PhongChuong: I had to keep the [client, model, builder] pub as comment references are causing doc tests to fail otherwise.
As the crate is not published and this will change when we have implement the Publisher veneer, I think it should be okay.

PR #3415, opened 2025-09-25
docs: Show actual enum values in setter samples.

amanda-tarafa:


amanda-tarafa:
This is the result of googleapis/librarian#2491

amanda-tarafa:
> I look forward to the release where all these examples are everywhere. I am more excited about the release with examples for all the RPCs!

Just one-off setters coming up, and then the ones for RPCs :)

PR #3419, opened 2025-09-26
feat(compute/v1): inline message fields

coryan:


src/generated/cloud/compute/v1/src/model.rs
```diff
@@ -646,6 +723,478 @@ impl wkt::message::Message for MachineTypeAggregatedList {
...
+pub mod machine_type_aggregated_list {
+    #[allow(unused_imports)]
+    use super::*;
+
+    /// The message type for the [warning][google.cloud.compute.v1.MachineTypeAggregatedList.warning] field.
+    ///
+    /// [google.cloud.compute.v1.MachineTypeAggregatedList.warning]: crate::model::machine_type_aggregated_list::Warning
+    #[derive(Clone, Default, PartialEq)]
+    #[non_exhaustive]
+    pub struct Warning {
```
coryan, line 736: This message type is duplicated (with slight variations in the comments) 184 times in the compute spec. Yuck.

src/generated/cloud/compute/v1/src/model.rs
```diff
@@ -646,6 +723,478 @@ impl wkt::message::Message for MachineTypeAggregatedList {
...
+        /// - New values appearing in future releases of the client library, **and**
+        /// - New values received dynamically, without application changes.
+        ///
+        /// Please consult the [Working with enums] section in the user guide for some
+        /// guidelines.
+        ///
+        /// [Working with enums]: https://google-cloud-rust.github.io/working_with_enums.html
+        #[derive(Clone, Debug, PartialEq)]
+        #[non_exhaustive]
+        pub enum Code {
```
coryan, line 852: This enum is duplicated  184 times in the compute specification.

I wonder if we should have a mechanism to dedup this code. It is going to be 300 x 184 lines of duplicative output... 

OTOH, I expect this will be less < 5% of the total code generated for compute.

dbolduc: > mechanism to dedup

Seems risky given that there are slight variations, even if they are only in the comments (for now).

If the compute folks say there is one `Warning`/`Code` as input into the generation process that fans out to 184 instances then I would be okay with it.

> duplicative output

It is also the ergonomics. With deduping applications could have one `fn ProcessWarning(...)` vs. 184 of them.

coryan: > > mechanism to dedup
> 
> Seems risky given that there are slight variations, even if they are only in the comments (for now).

The message has differences in the comments, the `code` enum does not.  But I cannot find an internal source of truth to determine if this is an accident or because there is a single source of truth.

> > duplicative output
> 
> It is also the ergonomics. With deduping applications could have one `fn ProcessWarning(...)` vs. 184 of them.

Macros + traits + generics can take care of that, but yes, we are exporting work to our customers.

PR #3420, opened 2025-09-26
feat(compute/v1): use per-service features

coryan:
This client library is going to be large. With per-service features we
can limit the build times.


coryan:
> Just forecasting, I wonder if we will need special handling of the LRO-type services which some of the other services depend on.

I expect we will insert a `get_operation()` method (just like we do for other services), so there won't be any need to call the service directly.

PR #3421, opened 2025-09-26
docs(storage): nicer landing page

dbolduc:


src/storage/README.md
```diff
@@ -1,10 +1,31 @@
...
+
+The client library types and functions are stable and not expected to change.
+Please note that Google Cloud services do change from time to time. The client
+libraries are designed to preserve backwards compatibility when the service
+changes in compatible ways. For example, adding RPCs, or fields to messages
+should not introduce breaking changes to the client libraries.
+
+## Quickstart
+
+The main types to work with this crate are the clients:
```
coryan: nit: we could provide a better quickstart for a veneer.

dbolduc: Fair. I stopped caring about this looking like a generated README.

I am afraid to paste code in that we don't build, so I linked to our user guide for storage.

PR #3425, opened 2025-09-26
feat(compute/v1): enable `instances` service

### LINKED ISSUE #3390: Write an integration test using enums
We want to verify the client library can send and receive enums.

coryan:
Fixes #3390.


coryan:
PTAL, I think this is working now.

src/integration-tests/src/compute.rs
```diff
@@ -101,27 +109,31 @@ pub async fn machine_types() -> Result<()> {
...
             continue;
         }
         latest = match &latest {
             None => Some(item),
-            Some(i) if item.family > i.family => Some(item),
+            Some(i)
+                if item
+                    .family
+                    .as_deref()
+                    .is_some_and(|f| f > i.family.as_deref().unwrap_or_default()) =>
```
dbolduc, line 133: nit: looking more closely, this is probably not always correct. e.g. `foo-9` > `foo-10`.

Nobody cares though.

dbolduc (approved):

src/integration-tests/src/compute.rs
coryan: Ack.

PR #3426, opened 2025-09-27
feat(compute/v1): singular fields are optional

coryan:


src/integration-tests/src/compute.rs
```diff
@@ -95,7 +95,10 @@ pub async fn machine_types() -> Result<()> {
         .set_machine_type("f1-micro")
         .send()
         .await?;
-    assert!(response.is_shared_cpu, "response={response:?}");
+    assert!(
```
dbolduc: nit: consider: `assert_eq!(response.is_shared_cpu, Some(true))`

dbolduc (dismissed):

src/integration-tests/src/compute.rs
coryan: Done.

dbolduc (approved):

PR #3431, opened 2025-09-29
fix: paginated response items

### LINKED ISSUE #3430: Wrong field in paginated responses
The following services use the wrong field in (some) paginated responses:

- `google-cloud-discoveryengine-v1`
- `google-cloud-retail-v2`

The bug is already fixed in the generator, this is a record to reference the problem in the generated code.

coryan:
This fixes a bug in the generated code for a few clients: in paginated
responses we should use the first eligible field for the items.

Fixes #3430

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2289,7 +2289,7 @@ dependencies = [
 
 [[package]]
 name = "google-cloud-discoveryengine-v1"
-version = "1.1.0"
+version = "2.0.0"
 dependencies = [
  "async-trait",
  "bytes",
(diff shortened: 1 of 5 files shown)


codyoss (dismissed):

src/generated/cloud/sql/v1/src/model.rs
```diff
@@ -4283,10 +4283,10 @@ impl wkt::message::Message for InstancesListResponse {
 
 #[doc(hidden)]
 impl gax::paginator::internal::PageableResponse for InstancesListResponse {
-    type PageItem = crate::model::DatabaseInstance;
+    type PageItem = crate::model::ApiWarning;
```
dbolduc: I don't think we should merge this as-is.

coryan: Agree, I implemented the override for SQL admin and regenerated the code, PTAL.

PR #3432, opened 2025-09-29
impl(gax-internal): Add disabled tracing logic to http.rs

westarle:
Introduces the core logic for emitting OpenTelemetry-compatible spans for HTTP requests within the `ReqwestClient::request_attempt`.

This change includes:
-   Instantiating `HttpSpanInfo` with request details.
-   Creating a tracing span.
-   Instrumenting the request future.
-   Recording response attributes to the span.

The tracing functionality is currently disabled by the `ENABLE_HTTP_TRACING` constant, which is set to `false`. This allows the logic to be merged without being active until integration tests are in place.

See #3239


src/gax-internal/src/http.rs
```diff
@@ -27,6 +27,13 @@ use http::{Extensions, Method};
 use std::sync::Arc;
 use std::time::Duration;
 
+use crate::observability;
+use crate::options;
+use tracing::Instrument;
+
+// TODO(#3239): Enable this flag to true once integration tests are in place.
+const ENABLE_HTTP_TRACING: bool = false;
```
coryan: I think the Rusty way to do this would be a feature, maybe a `_unstable-o12y` feature: the leading `_` makes it undocumented, and the `unstable-` prefix tells anybody naming the feature that they are about to enable something unstable.

Alternatives: `#[cfg(false)]` or `#[cfg(any())]`

src/gax-internal/src/http.rs
```diff
@@ -156,11 +175,30 @@ impl ReqwestClient {
...
+        let response_result = if self.tracing_enabled && ENABLE_HTTP_TRACING {
+            let mut span_info = observability::HttpSpanInfo::from_request(
+                &request,
+                options,
+                self.instrumentation,
+                attempt_count,
+            );
+            let span = span_info.create_span();
+            let result = self.inner.execute(request).instrument(span.clone()).await;
+            let _enter = span.enter();
```
coryan, line 188: Shouldn't you enter the span *before* calling `self.inner.execute()`?

src/gax-internal/src/http.rs
```diff
@@ -27,6 +27,13 @@ use http::{Extensions, Method};
 use std::sync::Arc;
 use std::time::Duration;
 
+use crate::observability;
```
coryan: It seems that you only use `HttpSpanInfo` from this module, in that case, I think it is more idiomatic to say: `use crate::observability::HttpSpanInfo;`

src/gax-internal/src/http.rs
```diff
@@ -27,6 +27,13 @@ use http::{Extensions, Method};
 use std::sync::Arc;
 use std::time::Duration;
```
coryan: nit: remove the blank line and `cargo fmt` can order these user declarations.

src/gax-internal/src/http.rs
```diff
@@ -39,6 +46,7 @@ pub struct ReqwestClient {
     polling_error_policy: Arc<dyn PollingErrorPolicy>,
     polling_backoff_policy: Arc<dyn PollingBackoffPolicy>,
     instrumentation: Option<&'static crate::options::InstrumentationClientInfo>,
```
coryan, line 43: nit: sometimes `crate::options` and sometimes `options::` is confusing. Pick one.

src/gax-internal/src/http.rs
```diff
@@ -131,11 +141,19 @@ impl ReqwestClient {
         let retry = self.get_retry_policy(&options);
         let backoff = self.get_backoff_policy(&options);
         let this = self.clone();
+
+        let mut attempt_count = 0u32;
```
coryan, line 140: no action required: there is something I do not quite like about having this counter here and in the retry loop.  But I like that we did not have to change existing APIs.

src/gax-internal/src/http.rs
coryan: or if there is a good reason why not: add a comment, it is not obvious to us o12y-cos-players.

src/gax-internal/src/http.rs
westarle: Here's what it looks like to pipe the definition of attempt_count:
 https://github.com/googleapis/google-cloud-rust/compare/main...westarle:google-cloud-rust:o12y-refactor-add-attempt-count#diff-f1eb073401433c5300befed4e6af602014b4b0b9bb3d40c159d3e0cdfc12d3fb

src/gax-internal/src/http.rs
westarle: I followed the convention in the rest of the module to say crate::...

src/gax-internal/src/http.rs
westarle: I wrote some comments to explain, the span is entered/exited through "instrument" and then after the await we want to enter a span to record the final results.

src/gax-internal/src/http.rs
westarle: I added a feature (I am going to wind up needing it for the next features) but kept the simple boolean to avoid a bunch of warnings about unused variables etc.

src/gax-internal/src/http.rs
westarle: I noticed we don't "use" any "crate:::" thing, so I inlined it below.

src/gax-internal/src/http.rs
westarle: done

src/gax-internal/src/http.rs
```diff
@@ -156,11 +174,32 @@ impl ReqwestClient {
...
+            // The instrument call ensures the span is entered/exited as the execute future is polled.
+            let result = self.inner.execute(request).instrument(span.clone()).await;
+            // Re-enter the span's context to record response attributes after the future has completed.
+            let _enter = span.enter();
+            span_info.update_from_response(&result);
+            span_info.record_response_attributes(&span);
+            result
+        } else {
+            self.inner.execute(request).await
+        };
```
coryan: I think you missed my point.  You want to say:

```suggestion
        #[cfg(not(feature = "_unstable-o12y"))]
        let response = self.inner.execute(request).await;
        #[cfg(feature = "_unstable-o12y")]
        let response_result = if self.tracing_enabled {
            let mut span_info = crate::observability::HttpSpanInfo::from_request(
                &request,
                options,
                self.instrumentation,
                attempt_count,
            );
            let span = span_info.create_span();
            // The instrument call ensures the span is entered/exited as the execute future is polled.
            let result = self.inner.execute(request).instrument(span.clone()).await;
            // Re-enter the span's context to record response attributes after the future has completed.
            let _enter = span.enter();
            span_info.update_from_response(&result);
            span_info.record_response_attributes(&span);
            result
        };
```

That guarantees the code is disabled at compile time.

westarle: thanks!  I was stressing a bit about correcting the unused warnings, so I was avoiding them with a flagged flag.  I think I've got a handle on them now.

coryan (approved):

Issue #3012, opened 2025-08-24
A minimal client to publish events


coryan:
We need a client for testing, something where the application calls "publish()" and the message is sent immediately.

We may remove this API once we implement a better alternative.

PR #3439, opened 2025-09-29
impl(pubsub): bootstrap publisher client

PhongChuong:
Implement a minimal publisher client and add an integration test that publish messages to a topic. Publisher is subject to change once the pubsub publisher design is more stable.

This is loosely based on the storage veneer implementations.

Fix: #3012

### DIFF
--- a/src/integration-tests/src/pubsub.rs
+++ b/src/integration-tests/src/pubsub.rs
@@ -20,20 +20,6 @@ use pubsub::model::{PubsubMessage, Topic};
 use rand::{Rng, distr::Alphanumeric};
 
 pub async fn basic_publisher() -> Result<()> {
-    // Enable a basic subscriber. Useful to troubleshoot problems and visually
-    // verify tracing is doing something.
-    #[cfg(feature = "log-integration-tests")]
-    let _guard = {
-        use tracing_subscriber::fmt::format::FmtSpan;
-        let subscriber = tracing_subscriber::fmt()
-            .with_level(true)
-            .with_thread_ids(true)
-            .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
-            .finish();
-
-        tracing::subscriber::set_default(subscriber)
-    };
-
     let (topic_admin, topic) = create_test_topic().await?;
 
     tracing::info!("testing publish()");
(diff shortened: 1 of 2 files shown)


src/integration-tests/src/pubsub.rs
```diff
@@ -15,9 +15,47 @@
...
+        PubsubMessage::new().set_data("World"),
+    ];
+    let resp = publisher
+        .publish()
+        .set_topic(&topic.name)
+        .set_messages(messages)
+        .send()
+        .await?;
+    assert_eq!(resp.message_ids.len(), 2);
+    tracing::info!("successfully published messages");
```
coryan: Consider printing the message ids or something?

src/pubsub/src/publisher/client.rs
```diff
@@ -0,0 +1,92 @@
...
+#[cfg(test)]
+mod tests {
+    use super::Publisher;
+
+    #[tokio::test]
+    async fn builder() -> anyhow::Result<()> {
+        let _ = Publisher::builder().build().await?;
+        Ok(())
+    }
+}
```
coryan, line 95: This is using ADC, which may not work in all environments.  Is there something else we can use to exercise the code? Test credentials?

coryan (approved):
The potential test problem worries me a bit, check with @dbolduc who ran into problems with ADC and unit tests. Otherwise, LGTM

src/pubsub/src/publisher/client.rs
dbolduc: +1 to `anonymous` credentials.

ADC in unit tests is fine, just not as hermetic as possible. I try to avoid them

PhongChuong:
Thanks for the review.
Updated with the suggestions. PTAL.

src/integration-tests/src/pubsub.rs
```diff
@@ -15,9 +15,47 @@
...
+    let _guard = {
+        use tracing_subscriber::fmt::format::FmtSpan;
+        let subscriber = tracing_subscriber::fmt()
+            .with_level(true)
+            .with_thread_ids(true)
+            .with_span_events(FmtSpan::NEW | FmtSpan::CLOSE)
+            .finish();
+
+        tracing::subscriber::set_default(subscriber)
+    };
```
coryan: We have refactored this code, look at the other tests.

coryan (changes requested):
Argh, I missed the refactoring to initialize the tracing subscriber in integration tests

src/integration-tests/src/pubsub.rs
PhongChuong: Updated.

PhongChuong:
Updated.
PTAL.

PhongChuong:
Ops. Auto merge was enable.
I'll ping the PR to @suzmue and address any concerns.

PR #3444, opened 2025-10-01
fix(storage): respect locational endpoints

### LINKED ISSUE #3375: Default `Host` header does not work for regional endpoints

dbolduc:
Fixes #3375 

Use locational endpoints as the `:authority` field in gRPC requests, instead of the default. An analog to #3402, but for gRPC.

I couldn't reproduce this with GCS. I think it was forgiving of trying to access locational resources from a global endpoint. And I did not want to [restrict global endpoint usage](https://cloud.google.com/assured-workloads/docs/restrict-endpoint-usage) in our test project.

I tested locally with `aiplatform`, and saw the same behavior we see over HTTP. https://github.com/dbolduc/google-cloud-rust/commit/381b5fe4130da304672df7b34a60187b7fb1e80b

### DIFF
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -191,7 +191,7 @@ impl Client {
     ) -> gax::client_builder::Result<InnerClient> {
         use tonic::transport::{ClientTlsConfig, Endpoint};
 
-        let origin = Uri::from_str(default_endpoint).map_err(BuilderError::transport)?;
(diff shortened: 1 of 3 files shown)


dbolduc:
I should rewrite this.

PR #3445, opened 2025-10-01
fix(storage): respect locational endpoints

### LINKED ISSUE #3375: Default `Host` header does not work for regional endpoints

dbolduc:
Fixes #3375 

Use locational endpoints as the `:authority` field in gRPC requests, instead of the default. An analog to #3402, but for gRPC.

I couldn't reproduce this with GCS. I think it was forgiving of trying to access locational resources from a global endpoint. And I did not want to [restrict global endpoint usage](https://cloud.google.com/assured-workloads/docs/restrict-endpoint-usage) in our test project.

I tested locally with `aiplatform`, and saw the same behavior we see over HTTP. https://github.com/dbolduc/google-cloud-rust/commit/381b5fe4130da304672df7b34a60187b7fb1e80b

### DIFF
--- a/src/gax-internal/src/host.rs
+++ b/src/gax-internal/src/host.rs
@@ -20,20 +20,29 @@ use std::str::FromStr;
 ///
 /// Notably, locational and regional endpoints are detected and used as the
 /// host. For VIPs and private networks, we need to use the default host.
-pub(crate) fn host_from_endpoint(
+///
+/// Accepting a generic function as a parameter lets us avoid dead code warnings
+/// when only one of HTTP, gRPC is enabled.
+///
+/// We could have returned a boolean that says whether to use the default
+/// endpoint vs. the custom endpoint. Note, however, that:
+///   (1) this function already computes the origin and host.
+///   (2) consumers of this function would have to recompute the values.
+pub(crate) fn from_endpoint<T>(
     endpoint: Option<&str>,
     default_endpoint: &str,
-) -> gax::client_builder::Result<String> {
-    let default_host = Uri::from_str(default_endpoint)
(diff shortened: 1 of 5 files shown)


src/gax-internal/src/host.rs
```diff
@@ -20,20 +20,29 @@ use std::str::FromStr;
...
 /// host. For VIPs and private networks, we need to use the default host.
-pub(crate) fn host_from_endpoint(
+///
+/// Accepting a generic function as a parameter lets us avoid dead code warnings
+/// when only one of HTTP, gRPC is enabled.
+///
+/// We could have returned a boolean that says whether to use the default
+/// endpoint vs. the custom endpoint. Note, however, that:
+///   (1) this function already computes the origin and host.
+///   (2) consumers of this function would have to recompute the values.
```
coryan, line 30: This is very weird, but I cannot think of a better approach in 10 minutes (something with a type that is convertible to `Uri` or `String`? Two different functions with the thing returning the tuple as a common implementation detail for both?)

I am fine with approving because the code in `pub(crate)`.

dbolduc: > This is very weird

Agree. We ugly up the code for a micro-optimization.

> Two different functions with the thing returning the tuple as a common implementation detail for both?

I originally had it as a type, but the compiler detected that one of the fields was not read when only one of HTTP / gRPC is enabled.

Also the two different functions would have to be feature gated. And if we are going to do that, might as well put the different functions in `http.rs` / `grpc.rs`.

PR #3446, opened 2025-10-01
feat(gax): Add Response Extensions

westarle:
This PR introduces a mechanism to pass extensions from transport layers (like HTTP and gRPC) up to the higher levels of the GAX crate, primarily for observability purposes.

**Key Changes:**

1.  **`Arc<Mutex<http::Extensions>>` in `Parts`:** The `google_cloud_gax::response::Parts` struct now includes an `extensions` field of type `Arc<Mutex<http::Extensions>>`. This allows storing arbitrary type-safe data. The `Arc<Mutex<>>` wrapper is used to maintain the `UnwindSafe` guarantee of `Response<T>`, as `http::Extensions` itself is not `UnwindSafe`. See issue #3463 for more details on this decision.
2.  **`extensions()` and `extensions_mut()` Methods:** New methods `extensions()` and `extensions_mut()` are added to `Response<T>`. These methods provide locked access to the underlying `http::Extensions` by returning a `MutexGuard`. They panic if the mutex is poisoned. These replace the previous `insert_extension` and `get_extension` methods.
3.  **Transport Layer Updates:**
    *   In `gax-internal/src/http.rs`, the `to_http_response` function is updated to move the extensions from the `reqwest::Response` into the new `Parts.extensions` field efficiently, avoiding unnecessary clones.
    *   In `gax-internal/src/grpc.rs`, the `to_gax_response` function is updated to move extensions from `tonic::Response` into `Parts.extensions`.

This change enables lower-level transport code to attach data (like observability summaries) to the response, which can then be accessed by higher-level wrappers or the end-user.

### DIFF
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -297,9 +297,12 @@ pub fn to_gax_response<T, G>(response: tonic::Response<T>) -> Result<gax::respon
 where
     T: crate::prost::FromProto<G>,
 {
-    let (metadata, body, _extensions) = response.into_parts();
+    let (metadata, body, extensions) = response.into_parts();
+    let mut parts = gax::response::Parts::default();
+    parts.headers = metadata.into_headers();
+    *parts.extensions.lock().unwrap() = extensions;
     Ok(gax::response::Response::from_parts(
-        gax::response::Parts::new().set_headers(metadata.into_headers()),
+        parts,
         body.cnv().map_err(Error::deser)?,
     ))
 }
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -300,7 +300,12 @@ where
     let (metadata, body, extensions) = response.into_parts();
     let mut parts = gax::response::Parts::default();
     parts.headers = metadata.into_headers();
-    *parts.extensions.lock().unwrap() = extensions;
+    let mut lock = parts
+        .extensions
+        .lock()
+        .expect("Extensions mutex is not poisoned");
+    *lock = extensions;
+    drop(lock); // Ensure the lock is dropped before parts is moved
     Ok(gax::response::Response::from_parts(
         parts,
         body.cnv().map_err(Error::deser)?,
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -24,7 +24,7 @@ use gax::response::{Parts, Response};
 use gax::retry_policy::{Aip194Strict as RetryAip194Strict, RetryPolicy, RetryPolicyExt as _};
 use gax::retry_throttler::SharedRetryThrottler;
 use http::{Extensions, Method};
-use std::sync::Arc;
+use std::sync::{Arc, Mutex};
 use std::time::Duration;
 #[cfg(feature = "_unstable-o12y")]
 use tracing::Instrument;
@@ -299,6 +299,7 @@ async fn to_http_response<O: serde::de::DeserializeOwned + Default>(
 ) -> Result<Response<O>> {
     // 204 No Content has no body and throws EOF error if we try to parse with serde::json
     let no_content_status = response.status() == reqwest::StatusCode::NO_CONTENT;
+    let extensions = Arc::new(Mutex::new(response.extensions().clone()));
     let response = http::Response::from(response);
     let (parts, body) = response.into_parts();
 
@@ -311,10 +312,10 @@ async fn to_http_response<O: serde::de::DeserializeOwned + Default>(
         content => serde_json::from_slice::<O>(&content).map_err(Error::deser)?,
     };
 
-    Ok(Response::from_parts(
-        Parts::new().set_headers(parts.headers),
-        response,
-    ))
+    let mut parts_out = Parts::default();
+    parts_out.headers = parts.headers;
+    parts_out.extensions = extensions;
+    Ok(Response::from_parts(parts_out, response))
 }
 
 #[cfg(test)]
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -299,7 +299,6 @@ async fn to_http_response<O: serde::de::DeserializeOwned + Default>(
 ) -> Result<Response<O>> {
     // 204 No Content has no body and throws EOF error if we try to parse with serde::json
     let no_content_status = response.status() == reqwest::StatusCode::NO_CONTENT;
-    let extensions = Arc::new(Mutex::new(response.extensions().clone()));
     let response = http::Response::from(response);
     let (parts, body) = response.into_parts();
 
@@ -314,7 +313,7 @@ async fn to_http_response<O: serde::de::DeserializeOwned + Default>(
 
     let mut parts_out = Parts::default();
     parts_out.headers = parts.headers;
-    parts_out.extensions = extensions;
+    parts_out.extensions = Arc::new(Mutex::new(parts.extensions));
     Ok(Response::from_parts(parts_out, response))
 }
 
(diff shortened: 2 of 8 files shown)


src/gax/src/response.rs
```diff
@@ -329,4 +385,28 @@ mod tests {
             Some(&http::HeaderValue::from_static("application/json"))
         );
     }
+
+    #[test]
+    fn response_extensions() {
+        #[derive(Clone, Debug, PartialEq)]
+        struct MyExtension(String);
```
dbolduc: nit: using a `&'static str` instead of a `String` would let us skip all the `.to_string()`s.

src/gax/src/response.rs
```diff
@@ -113,13 +113,28 @@
...
-///   
+///
 #[derive(Clone, Debug)]
 pub struct Response<T> {
     parts: Parts,
     body: T,
 }
 
+#[derive(Clone, Debug, Default)]
+struct ResponseExtensions(http::Extensions);
```
dbolduc: why the wrapper vs. using `http::Extensions` directly?

src/gax/src/response.rs
```diff
@@ -235,13 +260,24 @@ impl<T> Response<T> {
...
+///
 /// [tower]: https://github.com/tower-rs/tower
-#[derive(Clone, Debug, Default)]
+#[derive(Clone, Debug)]
 #[non_exhaustive]
 pub struct Parts {
     /// The HTTP headers or the gRPC metadata converted to HTTP headers.
     pub headers: http::HeaderMap<http::HeaderValue>,
+    /// Internal extensions for passing data between layers.
+    extensions: ResponseExtensions,
```
dbolduc: I think Carlos was suggesting we can make this public.

```rs
pub extensions: http::Extensions,
```

src/gax/src/response.rs
westarle: done!

src/gax/src/response.rs
westarle: removed wrapper...

src/gax/src/response.rs
westarle: done

westarle:
I need to use `Arc<Mutex<X>>` because http::Extensions is not `UnwindSafe` (it contains `Any`).  I'm open to suggestions for a better API, putting this here in case anyone has feedback.  Will revisit next week.  @coryan and @dbolduc

dbolduc:
> I need to use `Arc<Mutex<X>>` because http::Extensions is not `UnwindSafe`

Quick thought: If `http::Response` can use `http::Extensions`, why can't we? I think we are modeling our `gax::Response` off of `http::Response`.

https://docs.rs/http/latest/http/response/struct.Response.html#method.extensions

westarle:
> > I need to use `Arc<Mutex<X>>` because http::Extensions is not `UnwindSafe`
> 
> Quick thought: If `http::Response` can use `http::Extensions`, why can't we? I think we are modeling our `gax::Response` off of `http::Response`.
> 
> https://docs.rs/http/latest/http/response/struct.Response.html#method.extensions

Sorry I wasn't specific -- semver bot doesn't like the change from UnwindSafe.  We could alternatively accept the (technically) breaking change.  I will look more closely at http API thanks for this!

src/gax-internal/src/grpc.rs
```diff
@@ -297,9 +297,12 @@ pub fn to_gax_response<T, G>(response: tonic::Response<T>) -> Result<gax::respon
 where
     T: crate::prost::FromProto<G>,
 {
-    let (metadata, body, _extensions) = response.into_parts();
+    let (metadata, body, extensions) = response.into_parts();
+    let mut parts = gax::response::Parts::default();
+    parts.headers = metadata.into_headers();
+    *parts.extensions.lock().unwrap() = extensions;
```
coryan: I think I would write that as:

```suggestion
    let lock = parts.extensions.lock().expect("extensions is not poisoned");
    *lock = extensions;
```

src/gax-internal/src/http.rs
```diff
@@ -299,6 +299,7 @@ async fn to_http_response<O: serde::de::DeserializeOwned + Default>(
 ) -> Result<Response<O>> {
     // 204 No Content has no body and throws EOF error if we try to parse with serde::json
     let no_content_status = response.status() == reqwest::StatusCode::NO_CONTENT;
+    let extensions = Arc::new(Mutex::new(response.extensions().clone()));
```
coryan: I think you can avoid this `.clone()` (which is a deep copy) with something like:


```suggestion
    let response = http::Response::from(response);
    let (parts, body) = response.into_parts();
    let extensions = parts.extensions;
```

src/gax/src/response.rs
```diff
@@ -235,13 +247,15 @@ impl<T> Response<T> {
...
-///  
+///
 /// [tower]: https://github.com/tower-rs/tower
 #[derive(Clone, Debug, Default)]
 #[non_exhaustive]
 pub struct Parts {
     /// The HTTP headers or the gRPC metadata converted to HTTP headers.
     pub headers: http::HeaderMap<http::HeaderValue>,
+    /// Extensions for passing data from the transport.
+    pub extensions: Arc<Mutex<http::Extensions>>,
```
coryan, line 279: If we decide to go with this approach we should link a bug / document describing the tradeoffs and why we decided on this.

coryan (changes requested):

coryan:
> > I need to use `Arc<Mutex<X>>` because http::Extensions is not `UnwindSafe`
> 
> Quick thought: If `http::Response` can use `http::Extensions`, why can't we? 

We could, but @westarle is telling us this would be a breaking change (TIL, thanks @westarle).  By accident our `Response` is `UnwindSafe`:

https://docs.rs/google-cloud-gax/1.0.0/google_cloud_gax/response/struct.Response.html#impl-UnwindSafe-for-Response%3CT%3E

while `http::Response` is not:

https://docs.rs/http/latest/http/response/struct.Response.html#impl-UnwindSafe-for-Response%3CT%3E

And `cargo semver-checks` treats this as a breaking change:

https://github.com/obi1kenobi/cargo-semver-checks/blob/main/src/lints/function_abi_no_longer_unwind.ron

It seems unlikely that our libraries are used in the cases where this change is breaking, but we don't know.

I tried wrapping into an `Rc`, `Box`, a simple `Arc` and none of them does the trick.  `Arc<Mutex<>>`, as Wes proposes does the trick:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=7155d02d08e2e69ec691569af2cf7b9d

I think we have (maybe) three options:

1) Use `Arc<Mutex<Extensions>>` as long as  the public API only uses `Extensions`.
2) Silence the cargo-semver-checks warning. Ugly, potentially breaking for anybody using `catch_unwind()`. We can also use this if we use (1) first.
3) Bump everything to 2.0 and use plain `Extensions`. We can also do this if we use (1) first.

I am inclined to go with (1), but this is complicated enough to deserve some documentation (maybe a detailed bug?).

There are downsides to (1), specifically: we have to make more copies of the extensions than we would like, and initializing a response wrapper is more expensive.

dbolduc:
> semver bot doesn't like the change from `UnwindSafe`

Ahhh, sorry, I was unaware of `UnwindSafe` as a concept and didn't look it up. I assumed adding a field to a non-exhaustive struct would be safe.

> We could, but @westarle is telling us this would be a breaking change

Ack, I get it now. Thanks.

coryan:
> > semver bot doesn't like the change from `UnwindSafe`
> 
> Ahhh, sorry, I was unaware of `UnwindSafe` as a concept 


Me neither. And makes me wonder what else I am missing 😨

westarle:
Continues in #3479

PR #3447, opened 2025-10-01
feat(test): TestLayer for async tests

westarle:
This change introduces a dedicated test utility, `TestLayer`, to reliably capture and isolate tracing spans within individual test cases, especially in async contexts.

Key changes in `google-cloud-gax-internal/src/observability.rs`:

-   **New `test_layer` Module:** Encapsulates all test-specific tracing
    infrastructure, including the `TestLayer`, `CapturedSpanLog`, and
    `TestVisitor`. This module is marked `#[cfg(test)]`.
-   **`TestLayer::initialize()`:** Provides an RAII guard to install the
    `TestLayer` as a global subscriber for the duration of a test. It uses
    a unique `test_id` to isolate captured spans.
-   **`TestLayer::capture()`:** A new public API to retrieve spans
    captured for a specific `test_id`.
-   **Improved Isolation:** `CapturedSpanLog` and the global `SPAN_LOG`
    instance are now internal to the `test_layer` module, preventing
    direct access from tests.
-   **Refactoring:** Updated all existing observability tests to use the new
    `TestLayer::initialize()` and `TestLayer::capture()` API.

This new structure provides a cleaner, more robust, and idiomatic way to test tracing instrumentation in async environments.

### DIFF
--- a/src/gax-internal/Cargo.toml
+++ b/src/gax-internal/Cargo.toml
@@ -89,17 +89,18 @@ rpc  = { workspace = true, optional = true }
 wkt  = { workspace = true, optional = true }
 
 [dev-dependencies]
-anyhow.workspace      = true
-httptest.workspace    = true
-bytes.workspace       = true
-mockall.workspace     = true
-scoped-env.workspace  = true
-serde_with.workspace  = true
-serial_test.workspace = true
-test-case.workspace   = true
-tokio                 = { workspace = true, features = ["test-util"] }
-tracing               = { workspace = true, features = ["std"] }
-tracing-subscriber    = { workspace = true, features = ["registry", "std"] }
+anyhow.workspace                  = true
+bytes.workspace                   = true
+google-cloud-test-utils.workspace = true
+httptest.workspace                = true
+mockall.workspace                 = true
+scoped-env.workspace              = true
+serde_with.workspace              = true
+serial_test.workspace             = true
+test-case.workspace               = true
+tokio                             = { workspace = true, features = ["test-util"] }
+tracing                           = { workspace = true, features = ["std"] }
+tracing-subscriber                = { workspace = true, features = ["registry", "std"] }
 # Local crates
 echo-server = { path = "echo-server" }
 grpc-server = { path = "grpc-server" }
--- a/src/gax-internal/src/observability.rs
+++ b/src/gax-internal/src/observability.rs
@@ -231,7 +231,7 @@ impl HttpSpanInfo {
                     self.otel_status = OtelStatus::Ok;
                 } else {
                     self.otel_status = OtelStatus::Error;
-                    self.error_type = Some(response.status().to_string());
+                    self.error_type = Some(response.status().as_u16().to_string());
                 }
             }
             Err(err) => {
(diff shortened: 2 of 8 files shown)


westarle:
Let me know if this is getting too big for `observability.rs`.

src/gax-internal/src/observability.rs
```diff
@@ -647,46 +826,65 @@ mod tests {
...
         let expected: HashMap<String, String> = [
             ("new_attribute".to_string(), "new_value".to_string()),
             ("num_attribute".to_string(), "456".to_string()),
         ]
         .into_iter()
         .collect();
-        assert_eq!(attributes, expected);
+        assert_eq!(*attributes, expected);
     }
+    // GEMINI_TEST_MARKER
```
coryan: Uh?

coryan (commented):
(1) yes, this is getting too big for a single file (we have larger files, but we should do better).

(2) Is this supposed to be used outside `gax-internal`?  I worry about adding APIs, specially test+unstable things.  While we *can* break things, I don't know if we should be  making a lot of changes.  I might be worrying about nothing.

westarle:
> (1) yes, this is getting too big for a single file (we have larger files, but we should do better).

Thanks, I'll break it out!

> (2) Is this supposed to be used outside `gax-internal`? I worry about adding APIs, specially test+unstable things. While we _can_ break things, I don't know if we should be making a lot of changes. I might be worrying about nothing.

It is probably going to be used in `tests` and `integration-tests`, like this:

https://github.com/googleapis/google-cloud-rust/compare/main...westarle:google-cloud-rust:test/http-tracing-integration?expand=1#diff-5563268e35fbb4417967c90bdf58c131a9b651a4fb1514fd83d272f05c1040f3

The "TestLayer" is generic and doesn't have any interesting dependency, I can stick it anywhere if you have some ideas.  I think at least gax-internal and integration-tests will use it.  I'm not very happy publishing its API until I use it more.

coryan:
> > (2) Is this supposed to be used outside `gax-internal`? I worry about adding APIs, specially test+unstable things. While we _can_ break things, I don't know if we should be making a lot of changes. I might be worrying about nothing.
> 
> It is probably going to be used in `tests` and `integration-tests`, like this:
> 
> https://github.com/googleapis/google-cloud-rust/compare/main...westarle:google-cloud-rust:test/http-tracing-integration?expand=1#diff-5563268e35fbb4417967c90bdf58c131a9b651a4fb1514fd83d272f05c1040f3
> 
> The "TestLayer" is generic and doesn't have any interesting dependency, I can stick it anywhere if you have some ideas. I think at least gax-internal and integration-tests will use it. I'm not very happy publishing its API until I use it more.

My instinct would be to put this in `integration-tests` iterate on its API there, and move it to `gax-internal` only once it is sufficiently stable.

westarle:
not ready yet

westarle:
I think it's ready, I just put it in `integration-tests` for now @coryan

src/gax-internal/Cargo.toml
```diff
@@ -89,17 +89,18 @@ rpc  = { workspace = true, optional = true }
...
-serde_with.workspace  = true
-serial_test.workspace = true
-test-case.workspace   = true
-tokio                 = { workspace = true, features = ["test-util"] }
-tracing               = { workspace = true, features = ["std"] }
-tracing-subscriber    = { workspace = true, features = ["registry", "std"] }
+anyhow.workspace            = true
+bytes.workspace             = true
+httptest.workspace          = true
+integration-tests.workspace = true
```
westarle: added this one

src/gax-internal/src/observability.rs
westarle: bad robot :(

src/gax-internal/Cargo.toml
coryan: I am not sure this will work. It is adding a cyclic dependency, and I am not sure we can release a crate that depends on on an unreleased crate.

coryan: I think you need to move the tests too, at least until the TestLayer API is stable.

westarle: It's OK to depend on unpublished crates as long as you don't specify a version.  I don't know about circular dependencies, that does seem like it could cause a problem in the future.

Here I've addressed both issues by moving the tests and implementation out to integration-tests, with a kind of weird hack (`_test-internals`) to limit what needs to be `pub`.

It also hints at some clean interfaces for TestLayer...

src/gax-internal/src/lib.rs
```diff
@@ -39,6 +39,11 @@ pub mod http;
 #[cfg(feature = "_internal-http-client")]
 pub mod observability;
 
+#[cfg(all(feature = "_test-internals", feature = "_internal-http-client"))]
+pub mod observability_test_internals {
+    pub use crate::observability::test_internals::*;
+}
+
```
coryan: This seems like it saves one character? `observability::test_internals` works just as well?

coryan: Sorry I did not notice it before.

coryan (approved):

westarle:
Hopefully this fixes lint.  I added a couple tests and cleaned up TestLayer code -- it should be ready to go (if checks pass!)

Issue #3449, opened 2025-10-01
ID Token Credentials


alvarowolfx:
ID tokens provide proof of authentication and contain information about the user, enabling the client application to verify the user's identity. 

[AIP](https://google.aip.dev/auth/4116)

References and use cases
* https://cloud.google.com/run/docs/authenticating/service-to-service
* https://cloud.google.com/iap/docs/authentication-howto#iap_make_request-python

alvarowolfx:
Related to #915 and https://github.com/googleapis/google-cloud-rust/discussions/628

PR #3450, opened 2025-10-01
feat(auth): implement id tokens with MDS source

alvarowolfx:
Adds base ID Token credential trait to be followed by other implementations. Add MDS implementation as initial one.

Towards #3449

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -26,6 +26,7 @@ pub mod anonymous;
 pub mod api_key_credentials;
 pub mod external_account;
 pub(crate) mod external_account_sources;
+pub(crate) mod idtoken; 
 pub mod impersonated;
 pub(crate) mod internal;
 pub mod mds;
--- a/src/auth/src/credentials/idtoken.rs
+++ b/src/auth/src/credentials/idtoken.rs
@@ -1,0 +1,88 @@
+// Copyright 2024 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::Result;
+use crate::token::Token;
+use std::future::Future;
+use std::sync::Arc;
+
+/// Represents an `IDTokenCredentials` used to obtain an ID token.
+///
+/// `IDTokenCredentials` provide a way to obtain OIDC ID tokens, which are
+/// commonly used for authenticating to services like Cloud Run or Identity-Aware
+/// Proxy (IAP). Unlike access tokens, ID tokens are not used to authorize access
+// to Google Cloud APIs but to verify the identity of a principal.
+///
+/// This struct serves as a wrapper around different credential types that can
+/// produce ID tokens, such as service accounts or metadata server credentials.
+#[derive(Clone, Debug)]
+pub struct IDTokenCredentials {
+    pub(crate) inner: Arc<dyn dynamic::IDTokenCredentialsProvider>,
+}
+
+impl<T> From<T> for IDTokenCredentials
+where
+    T: IDTokenCredentialsProvider + Send + Sync + 'static,
+{
+    fn from(value: T) -> Self {
+        Self {
+            inner: Arc::new(value),
+        }
+    }
+}
+
+impl IDTokenCredentials {
+    /// Asynchronously retrieves an ID token.
+    ///
+    /// This method will contact the underlying credential source to obtain a
+    /// fresh ID token if a valid one is not already cached.
+    pub async fn id_token(&self) -> Result<Token> {
+        self.inner.id_token().await
+    }
+}
+
+/// A trait for credential types that can provide OIDC ID tokens.
+///
+/// Implement this trait to create custom ID token providers. This is useful for
+/// testing or for integrating with authentication systems not natively
+/// supported by this library.
+pub trait IDTokenCredentialsProvider: std::fmt::Debug {
+    /// Asynchronously retrieves an ID token.
+    fn id_token(&self) -> impl Future<Output = Result<Token>> + Send;
+}
+
+/// A module containing the dynamically-typed, object-safe version of the
+/// `IDTokenCredentialsProvider` trait. This is an internal implementation detail.
+pub(crate) mod dynamic {
+    use crate::Result;
+    use crate::token::Token;
+
+    /// A dyn-compatible, crate-private version of `IDTokenCredentialsProvider`.
+    #[async_trait::async_trait]
+    pub trait IDTokenCredentialsProvider: Send + Sync + std::fmt::Debug {
+        /// Asynchronously retrieves an ID token.
+        async fn id_token(&self) -> Result<Token>;
+    }
+
+    /// The public `IDTokenCredentialsProvider` implements the dyn-compatible `IDTokenCredentialsProvider`.
+    #[async_trait::async_trait]
+    impl<T> IDTokenCredentialsProvider for T
+    where
+        T: super::IDTokenCredentialsProvider + Send + Sync,
+    {
+        async fn id_token(&self) -> Result<Token> {
+            T::id_token(self).await
+        }
+    }
+}
--- a/src/auth/src/credentials/mds.rs
+++ b/src/auth/src/credentials/mds.rs
@@ -397,6 +397,138 @@ impl TokenProvider for MDSAccessTokenProvider {
     }
 }
 
+pub mod idtoken {
+    use std::sync::Arc;
+
+    use super::{
+        GCE_METADATA_HOST_ENV_VAR, MDS_DEFAULT_URI, METADATA_FLAVOR, METADATA_FLAVOR_VALUE,
+        METADATA_ROOT,
+    };
+    use crate::Result;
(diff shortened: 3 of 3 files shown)


src/auth/src/credentials/mds.rs
```diff
@@ -963,4 +1095,96 @@ mod tests {
         assert_eq!(universe_domain_response, DEFAULT_UNIVERSE_DOMAIN);
         Ok(())
     }
+
+    #[cfg(test)]
+    mod idtoken_tests {
```
coryan: nit: why a submodule?  And if you need one, does it need to be called `::tests::idtoken_tests`?

src/auth/src/credentials/idtoken.rs
```diff
@@ -0,0 +1,88 @@
+// Copyright 2024 Google LLC
```
coryan: nit: copyright year is wrong, probably cut&paste error?

src/auth/src/credentials.rs
```diff
@@ -26,6 +26,7 @@ pub mod anonymous;
 pub mod api_key_credentials;
 pub mod external_account;
 pub(crate) mod external_account_sources;
+pub(crate) mod idtoken;
```
coryan: I find the module structure of `google-cloud-auth` slightly odd...  it is deeper than I would expect 🤷  I am not sure if this needs to be nested or one lever up.

src/auth/src/credentials/mds.rs
alvarowolfx: removed the nesting

src/auth/src/credentials/idtoken.rs
alvarowolfx: solved

src/auth/src/credentials.rs
alvarowolfx: I don't have strong opinions, I was just following the design doc

alvarowolfx (commented):
addressed initial comments and added integration tests. Still missing some public structs and method documentation.

src/auth/integration-tests/src/lib.rs
```diff
@@ -267,6 +270,44 @@ pub async fn api_key() -> anyhow::Result<()> {
...
+    let client = reqwest::Client::new();
+    let expected_email = client
+        .get("http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/email")
+        .header("Metadata-Flavor", "Google")
+        .send()
+        .await
+        .expect("failed to get service account email from metadata server")
+        .text()
+        .await
+        .expect("failed to read service account email from metadata server response");
```
coryan: Without a retry loop this may flake at any time. Can we add something simple, like trying 3 times?

coryan: Feel free to do that in a followup PR.

src/auth/src/credentials/idtoken.rs
```diff
@@ -0,0 +1,88 @@
...
+            inner: Arc::new(value),
+        }
+    }
+}
+
+impl IDTokenCredentials {
+    /// Asynchronously retrieves an ID token.
+    ///
+    /// This method will contact the underlying credential source to obtain a
+    /// fresh ID token if a valid one is not already cached.
```
coryan:
```suggestion
    /// Obtains an ID token. If one is cached, returns the cached value.
```

coryan: All these functions are missing examples.

src/auth/src/credentials/idtoken.rs
```diff
@@ -0,0 +1,88 @@
...
+    pub async fn id_token(&self) -> Result<Token> {
+        self.inner.id_token().await
+    }
+}
+
+/// A trait for credential types that can provide OIDC ID tokens.
+///
+/// Implement this trait to create custom ID token providers. This is useful for
+/// testing or for integrating with authentication systems not natively
+/// supported by this library.
```
coryan: I found the comment a bit awkward, I am not sure if this is better:

```suggestion
/// Implement this trait to create custom ID token providers.
/// For example, if you are working with an authentication system not
/// supported by this crate. Or if you are trying to write a test and need
/// to mock the existing `IDTokenCredentialsProvider` implementations.
```

coryan (dismissed):
Nothing blocking.  The comments / examples could be better.

alvarowolfx:
will make improvements after this PR
* add retry logic to integration tests when hitting the MDS `/email` endpoint as proposed by @coryan 
* use `TokenCache` just internally to cache id tokens.

src/auth/src/credentials/idtoken.rs
```diff
@@ -0,0 +1,88 @@
...
+
+/// Represents an `IDTokenCredentials` used to obtain an ID token.
+///
+/// `IDTokenCredentials` provide a way to obtain OIDC ID tokens, which are
+/// commonly used for authenticating to services like Cloud Run or Identity-Aware
+/// Proxy (IAP). Unlike access tokens, ID tokens are not used to authorize access
+// to Google Cloud APIs but to verify the identity of a principal.
+///
+/// This struct serves as a wrapper around different credential types that can
+/// produce ID tokens, such as service accounts or metadata server credentials.
```
coryan: This is a public type, the bar for their documentation is higher. Sorry I did not notice this before.


```suggestion
/// Obtain OIDC ID tokens.
///
/// `IDTokenCredentials` provide a way to obtain OIDC ID tokens, which are
/// commonly used for authenticating to services like Cloud Run or Identity-Aware
/// Proxy (IAP). Unlike access tokens, ID tokens are not used to authorize access
// to Google Cloud APIs, but to verify the identity of a principal.
///
/// This struct serves as a wrapper around different credential types that can
/// produce ID tokens, such as service accounts or metadata server credentials.
```

src/auth/src/credentials/idtoken.rs
```diff
@@ -0,0 +1,88 @@
...
+// limitations under the License.
+
+use crate::Result;
+use crate::token::Token;
+use std::future::Future;
+use std::sync::Arc;
+
+/// Represents an `IDTokenCredentials` used to obtain an ID token.
+///
+/// `IDTokenCredentials` provide a way to obtain OIDC ID tokens, which are
```
coryan, line 22: The first time you use `OIDC ID token` please link a relevant document at cloud.google.com describing what these are.  Do not assume the reader knows.

src/auth/src/credentials/idtoken.rs
```diff
@@ -0,0 +1,88 @@
...
+/// commonly used for authenticating to services like Cloud Run or Identity-Aware
+/// Proxy (IAP). Unlike access tokens, ID tokens are not used to authorize access
```
coryan: This is slightly contradictory, Cloud Run is a Google Cloud service and you just said this is used for authentication.

I think you want to say something like "services *hosted* in Cloud Run can use ID token to authenticate incoming requests. Likewise, the Identity-Aware Proxy requires OIDC ID tokens to authenticate requests mediated by the proxy."

src/auth/src/credentials/idtoken.rs
```diff
@@ -0,0 +1,88 @@
...
+/// Implement this trait to create custom ID token providers.
+/// For example, if you are working with an authentication system not
+/// supported by this crate. Or if you are trying to write a test and need
+/// to mock the existing `IDTokenCredentialsProvider` implementations.
+pub trait IDTokenCredentialsProvider: std::fmt::Debug {
+    /// Asynchronously retrieves an ID token.
+    fn id_token(&self) -> impl Future<Output = Result<Token>> + Send;
+}
+
+/// A module containing the dynamically-typed, object-safe version of the
```
coryan: "object-safe" is no longer used in Rust documentation, it is now called "dyn-compatible".

src/auth/src/credentials/mds.rs
```diff
@@ -397,8 +397,249 @@ impl TokenProvider for MDSAccessTokenProvider {
...
+            // Determine the endpoint and whether it was overridden
+            if let Ok(host_from_env) = std::env::var(GCE_METADATA_HOST_ENV_VAR) {
+                // Check GCE_METADATA_HOST environment variable first
+                final_endpoint = format!("http://{host_from_env}");
+            } else if let Some(builder_endpoint) = self.endpoint {
+                // Else, check if an endpoint was provided to the mds::Builder
+                final_endpoint = builder_endpoint;
+            } else {
+                // Else, use the default metadata root
+                final_endpoint = METADATA_ROOT.to_string();
```
sai-sunder-s, line 511: Is it possible to reuse the logic from the access token creds?

src/auth/src/credentials/mds.rs
```diff
@@ -397,8 +397,249 @@ impl TokenProvider for MDSAccessTokenProvider {
...
+            let request = client
+                .get(format!("{}{}/identity", self.endpoint, MDS_DEFAULT_URI))
+                .header(
+                    METADATA_FLAVOR,
+                    HeaderValue::from_static(METADATA_FLAVOR_VALUE),
+                )
+                .query(&[
+                    ("audience", audience),
+                    ("format", format),
+                    ("licenses", licenses),
```
sai-sunder-s: are these required query parameters?

alvarowolfx: they are not required, but I'm passing their default values.

sai-sunder-s: Why not just skip them if user did not specify them? Feels unnecessary for us to set the default values, the service should be doing that already.

src/auth/src/credentials/mds.rs
alvarowolfx: I can add a private function on the module later, but in the access token logic, there are some extra variables being created to track if the endpoint was overrided or not

src/auth/src/credentials/idtoken.rs
alvarowolfx: added relevant docs and updated comment around usage with Cloud Run and IAP

PR #3452, opened 2025-10-02
chore(auth): bump version

coryan:
Minor testing changes, require a version bump or `sidekick` will do it
for us.


dbolduc:
Can we add `src/auth/.gcb` to the `ignored-changes` list?

I think `src/auth/integration-tests/` changes are already ignored (bc that has its own `Cargo.toml`).

Argh, #3440 might still trigger the version bump.

dbolduc (approved):

PR #3453, opened 2025-10-02
feat: Add support for APIs that use server-sent events (SSE)

brianquinlan:
The generator code that uses this looks like:
https://github.com/googleapis/librarian/compare/main...brianquinlan:librarian:generate_stream?expand=1

When we launch `package:google_cloud_ai_generativelanguage_v1`, the only method that supports streaming will look like:

```dart
  /// Generates a [streamed response](https://ai.google.dev/gemini-api/docs/text-generation?lang=python#generate-a-text-stream)
  /// from the model given an input `GenerateContentRequest`.
  Stream<GenerateContentResponse> streamGenerateContent(
    GenerateContentRequest request,
  ) {
    final url = Uri.https(_host, '/v1/${request.model}:streamGenerateContent');
    return _client
        .postStreaming(url, body: request)
        .map(GenerateContentResponse.fromJson);
  }
```

### DIFF
--- a/dart/packages/google_cloud_gax/lib/gax.dart
+++ b/dart/packages/google_cloud_gax/lib/gax.dart
@@ -66,83 +66,123 @@ class ServiceClient {
 
   ServiceClient({required this.client});
 
-  Future<Map<String, dynamic>> get(Uri url) async {
-    final response = await client.get(
-      url,
-      headers: {
-        _clientKey: _clientName,
-      },
-    );
-    return _processResponse(response);
-  }
+  Future<Map<String, dynamic>> get(Uri url) => _makeRequest(url, 'GET');
 
-  Future<Map<String, dynamic>> post(Uri url, {JsonEncodable? body}) async {
-    final response = await client.post(
-      url,
-      body: body?._asEncodedJson,
-      headers: {
-        _clientKey: _clientName,
-        if (body != null) _contentTypeKey: _typeJson,
-      },
-    );
-    return _processResponse(response);
-  }
+  Stream<Map<String, dynamic>> getStreaming(Uri url) =>
+      _makeStreamingRequest(url, 'GET');
 
-  Future<Map<String, dynamic>> put(Uri url, {JsonEncodable? body}) async {
-    final response = await client.put(
-      url,
-      body: body?._asEncodedJson,
-      headers: {
-        _clientKey: _clientName,
-        if (body != null) _contentTypeKey: _typeJson,
-      },
-    );
-    return _processResponse(response);
-  }
+  Future<Map<String, dynamic>> delete(Uri url) => _makeRequest(url, 'DELETE');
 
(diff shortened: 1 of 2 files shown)


brianquinlan:
@devoncarew @natebosch for Dart review

dart/packages/google_cloud_gax/lib/gax.dart
```diff
@@ -66,83 +66,123 @@ class ServiceClient {
...
+      _throwException(
+        response.statusCode,
+        response.reasonPhrase,
+        await response.stream.bytesToString(),
+      );
+    }
+
+    final lines = response.stream.toStringStream().transform(LineSplitter());
+    await for (final line in lines) {
+      // Google API only generate "data" events.
```
devoncarew:
```suggestion
      // Google APIs only generate "data" events.
```

dart/packages/google_cloud_gax/lib/gax.dart
```diff
@@ -66,83 +66,123 @@ class ServiceClient {
...
-      url,
-      body: body?._asEncodedJson,
-      headers: {
-        _clientKey: _clientName,
-        if (body != null) _contentTypeKey: _typeJson,
-      },
-    );
-    return _processResponse(response);
-  }
+  Stream<Map<String, dynamic>> deleteStreaming(Uri url) =>
```
devoncarew, line 76: Are these for completeness? I believe - or at least, assume - that the streaming apis only use POST when streaming; I could be mistaken. You may want to just expose the streaming variants we expect will exist.

brianquinlan:
> This looks good, but can you update the PR description to mention the apis we expect this to apply to?

Updated. Is that what you expected?

dart/packages/google_cloud_gax/lib/gax.dart
brianquinlan: I don't know what the complete set of methods that support streaming is and, AFAIK, that isn't documented anywhere.

I could just support POST for now since `streamGenerateContent` is the only method in our initial set of generated packages that supports SSE and it uses POST. But I don't see why GET and PUSH couldn't also support streaming and it doesn't cost us anything to support every method.

devoncarew:
> Updated. Is that what you expected?

Yup! Just mentioning that we expect to use it for the gemini API in order to stream generateContent results back.

dart/packages/google_cloud_gax/lib/gax.dart
devoncarew: Yeah, this is a non-blocking comment for me, just something for you to think about (it was strange seeing a `deleteStreaming` implementation). 'for completeness' seems like a reasonable reason.

brianquinlan:
> FWIW, there are not that many server-side streaming APIs in Google Cloud. And there are quite a few bidi streaming APIs, which are only supported via gRPC.

Yeah, I care specifically about the gemini `streamGenerateContent` function.

coryan (approved):

PR #3456, opened 2025-10-02
feat: Indicate the dart and client version in x-goog-api-client

brianquinlan:
The `'x-goog-api-client'` header now looks like `'gl-dart/3.9.2 gax/0.1.0'`

### DIFF
--- a/dart/packages/google_cloud_gax/lib/src/web.dart
+++ b/dart/packages/google_cloud_gax/lib/src/web.dart
@@ -1,0 +1,18 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


brianquinlan:
@devoncarew For Dart review.

dart/packages/google_cloud_gax/lib/src/web.dart
```diff
@@ -0,0 +1,18 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+/// Web-specific implementations.
+library;
+
+final String? dartVersion = null;
```
devoncarew: Consider having `dartVersion` be non-null, having the web impl. return `'0'`, and moving the disclaimer comment (`'0' is a special version string indicating that the version isn't known.`) here.

devoncarew (dismissed):
lgtm! One comment in-line (and, +1 to the 'Keep these versions in sync' reminder comments).

coryan (dismissed):

dart/packages/google_cloud_gax/lib/src/web.dart
brianquinlan: Done!

PR #3468, opened 2025-10-06
fix(gax-internal): better gRPC error messages

### LINKED ISSUE #3460: Improve error message for bad gRPC endpoints
When a bad endpoint is used for a gRPC-based client, the error surfaced by `tonic` is not actionable.

Developers need to google the error and find this comment deep in the thread to recognize the problem is with the endpoint. (I personally have done this twice): https://github.com/hyperium/tonic/issues/100#issuecomment-1193338079

For our `StorageControl` client, this is important. Applications may try to use locational or regional endpoints (which are not yet supported over gRPC).

coryan:
The Google Cloud load balancer can respond with HTML or XML messages if
the target endpoint is invalid, e.g. `not-a-service.googleapis.com`, or
if the target endpoint does not match the service, e.g. using
`pubsub.googleapis.com` to call Cloud Storage RPCs.

This change generates a better error message in these cases, which we
hope makes troubleshooting easier.

----
Fixes #3460

From:

```
ERROR the service reports an error with code INTERNAL described as: protocol error: received message with invalid compression flag: 60 (valid flags are 0 and 1) while receiving response with status: 400 Bad Request
```

To:

```
ERROR the transport reports an error: unexpected value in content-type header, should start with application/grpc. In Google Cloud, this is a common problem when using an invalid endpoint, or an endpoint that does not support the target gRPC service.

Caused by:
    0: unexpected value in content-type header, should start with application/grpc. In Google Cloud, this is a common problem when using an invalid endpoint, or an endpoint that does not support the target gRPC service.
    1: status: 'Internal error', self: "protocol error: received message with invalid compression flag: 60 (valid flags are 0 and 1) while receiving response with status: 400 Bad Request", metadata: {"content-type": "application/xml; charset=UTF-8", "x-guploader-uploadid": "AAwnv3KXwY3MPyo1Mv2r0U5jhIh9oftiE8kgrHNLgCM_vHJ1MCChboJ5lUjvIfDU2RSqFcgLnUQoKr8", "content-length": "203", "date": "Mon, 06 Oct 2025 13:58:59 GMT", "server": "UploadServer", "alt-svc": "h3=\":443\"; ma=2592000,h3-29=\":443\"; ma=2592000"}
```

### DIFF
--- a/src/gax-internal/src/grpc/from_status.rs
+++ b/src/gax-internal/src/grpc/from_status.rs
@@ -142,6 +142,9 @@ mod tests {
             .and_then(|e| e.downcast_ref::<tonic::Status>())
             .expect("want a tonic::Status as source().source()");
         assert_eq!(source.code(), tonic::Code::Internal);
+
+        let fmt = format!("{got}");
+        assert!(fmt.contains("should start with application/grpc"), "fmt={fmt}, got={got:?}");
     }
 
     #[test]
--- a/src/gax-internal/src/grpc/from_status.rs
+++ b/src/gax-internal/src/grpc/from_status.rs
@@ -52,8 +52,29 @@ pub fn to_gax_error(status: tonic::Status) -> Error {
         return Error::transport(headers, status);
     }
 
+    let content_type = headers.get("content-type").map(|v| v.as_bytes());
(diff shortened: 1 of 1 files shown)


dbolduc (dismissed):
Thanks!

coryan:
PTAL

PR #3469, opened 2025-10-06
refactor(test-utils): TestLayer creates an ID for the span

westarle:
This lightens up the test integrations: now they don't need to include the test name in a static string.


src/test-utils/src/test_layer.rs
```diff
@@ -181,26 +190,28 @@ impl TestLayer {
     /// Installs the `TestLayer` as a global subscriber if it hasn't been already.
     /// It clears any previously captured spans for the given `test_id`.
     ///
-    /// Returns an RAII guard (`tracing::span::EnteredSpan`). The layer will
-    /// only capture spans associated with this `test_id` while this guard is in scope.
-    pub fn initialize(test_id: &'static str) -> tracing::span::EnteredSpan {
+    /// Returns a `TestLayerGuard` which must be kept in scope for the duration of the test.
+    pub fn initialize() -> TestLayerGuard {
```
coryan, line 194: I think it would be more idiomatic to call this `new()`

https://rust-lang.github.io/api-guidelines/predictability.html?highlight=new()#constructors-are-static-inherent-methods-c-ctor

westarle: I gave it a weird name because it doesn't return a TestLayer, it returns a different type.

I think I might wind up updating the API again before I'm done so I'll leave it for now and come back to it, if that's OK with you!

PR #3470, opened 2025-10-06
test(gax-internal): Add tracing tests for HTTP retry loop

westarle:
Added tests for the retry code in #3239, using test-case.

The "tracing on" test is disabled if the _unstable-o12y flag is off.

### DIFF
--- a/src/gax-internal/tests/http_retry_loop.rs
+++ b/src/gax-internal/tests/http_retry_loop.rs
@@ -29,10 +29,13 @@ mod tests {
     use gax::retry_policy::{Aip194Strict, RetryPolicyExt};
     use google_cloud_gax_internal::http::ReqwestClient;
     use google_cloud_gax_internal::options::ClientConfig;
+    use google_cloud_test_utils::test_layer::TestLayer;
     use http::StatusCode;
     use httptest::{Expectation, Server, matchers::*, responders::*};
+    use opentelemetry_semantic_conventions::trace as semconv;
     use serde_json::json;
     use std::time::Duration;
+    use test_case::test_case;
 
     type Result<T> = std::result::Result<T, Box<dyn std::error::Error>>;
 
@@ -82,13 +85,24 @@ mod tests {
         Ok(())
     }
 
+    #[test_case(false; "tracing off")]
+    #[test_case(true; "tracing on")]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    async fn retry_loop_retry_success() -> Result<()> {
+    async fn retry_loop_retry_success(enable_tracing: bool) -> Result<()> {
+        if enable_tracing && !cfg!(feature = "_unstable-o12y") {
+            // Skip this test if the feature is disabled.
+            return Ok(());
+        }
+
+        let mut config = test_config();
+        config.tracing = enable_tracing;
+        let guard = TestLayer::initialize();
+
         // We create a server that will return two transient errors and then succeed.
         let server = start(vec![transient(), transient(), success()]);
         let endpoint = format!("http://{}", server.addr());
 
-        let client = ReqwestClient::new(test_config(), &endpoint).await?;
+        let client = ReqwestClient::new(config, &endpoint).await?;
         let builder = client.builder(reqwest::Method::GET, "/retry".into());
         let body = json!({});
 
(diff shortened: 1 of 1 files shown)


src/gax-internal/tests/http_retry_loop.rs
```diff
@@ -103,16 +117,53 @@ mod tests {
             .await;
         let response = response?.into_body();
         assert_eq!(response, json!({"status": "done"}));
+
+        let spans = TestLayer::capture(&guard);
+        if enable_tracing {
```
dbolduc, line 122: Can we have separate tests instead of using `test_case`? We try to keep unit test logic simple by avoiding `if` statements.

src/gax-internal/tests/http_retry_loop.rs
```diff
@@ -82,13 +85,24 @@ mod tests {
         Ok(())
     }
 
+    #[test_case(false; "tracing off")]
+    #[test_case(true; "tracing on")]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    async fn retry_loop_retry_success() -> Result<()> {
+    async fn retry_loop_retry_success(enable_tracing: bool) -> Result<()> {
+        if enable_tracing && !cfg!(feature = "_unstable-o12y") {
```
dbolduc, line 92: then we can put the `cfg(feature = ...)` gate outside the `#[tokio::test]`

src/gax-internal/tests/http_retry_loop.rs
```diff
@@ -103,16 +117,53 @@ mod tests {
...
+                Some(&"1".to_string())
+            );
+            // Span 3 (attempt 2)
+            assert_eq!(spans[2].name, "http_request");
+            assert_eq!(
+                spans[2].attributes.get(semconv::HTTP_REQUEST_RESEND_COUNT),
+                Some(&"2".to_string())
+            );
+        } else {
+            assert_eq!(spans.len(), 0, "Expected 0 spans when tracing is disabled");
```
dbolduc, line 144: nit: consider printing a debug string for `spans` in this case, so we learn more about its contents. "spans={spans:?}"

src/gax-internal/tests/http_retry_loop.rs
```diff
@@ -127,6 +178,29 @@ mod tests {
...
+
+        let spans = TestLayer::capture(&guard);
+        if enable_tracing {
+            assert_eq!(spans.len(), 3, "Expected 3 spans");
+            // Span 1 (attempt 0)
+            assert!(
+                !spans[0]
+                    .attributes
+                    .contains_key(semconv::HTTP_REQUEST_RESEND_COUNT)
+            );
```
coryan, line 190: If this fails you want to print more details, maybe `assert!(...., "{spans[0]:?}");` ?

src/gax-internal/tests/http_retry_loop.rs
```diff
@@ -127,6 +178,29 @@ mod tests {
...
+            // Span 2 (attempt 1)
+            assert_eq!(
+                spans[1].attributes.get(semconv::HTTP_REQUEST_RESEND_COUNT),
+                Some(&"1".to_string())
```
coryan, line 194: I find something like:


```suggestion
                spans[1].attributes.get(semconv::HTTP_REQUEST_RESEND_COUNT).as_deref(),
                Some("1")
```

more readable.

src/gax-internal/tests/http_retry_loop.rs
dbolduc: I think I would leave existing tests alone and add 2 new tests for:
- tracing disabled (any number of retries, up to you)
- tracing enabled (with 2 retries)

src/gax-internal/tests/http_retry_loop.rs
```diff
@@ -103,16 +117,53 @@ mod tests {
...
+        }
         Ok(())
     }
 
+    #[test_case(false; "tracing off")]
+    #[test_case(true; "tracing on")]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
-    async fn retry_loop_too_many_transients() -> Result<()> {
+    async fn retry_loop_too_many_transients(enable_tracing: bool) -> Result<()> {
+        if enable_tracing && !cfg!(feature = "_unstable-o12y") {
```
coryan, line 153: tests with a lot of logic like this are better off split into two tests. Feel free to refactor the common initialization to a helper function in the `mod tests { ... }` module.

src/gax-internal/tests/http_retry_loop.rs
coryan, line 92: Ditto, this is too much logic for a test. In addition, this suggests the test should be conditionally compiled.

coryan (changes requested):

dbolduc (commented):
I just have nits on piggy-backing existing tests vs. adding new tests. Feel free to push back.

Also, should we enable only the unstable observability feature in the CI?

https://github.com/googleapis/google-cloud-rust/blob/4964d220d74b0c9d010bc5ccbb98323622057491/.github/workflows/sdk.yaml#L117-L121

westarle:
Thank you for the feedback!  Moved to #3475

PR #3473, opened 2025-10-06
chore: update googleapis SHA circa 2025-10-06

suzmue:


src/generated/showcase/src/lib.rs
```diff
@@ -85,4 +85,14 @@ pub(crate) mod info {
             ac.rest_header_value()
         };
     }
+    lazy_static::lazy_static! {
+        pub(crate) static ref INSTRUMENTATION_CLIENT_INFO: gaxi::options::InstrumentationClientInfo = {
+            let mut info = gaxi::options::InstrumentationClientInfo::default();
```
coryan, line 90: I suspect `sidekick@main` and the sidekick version used in CI do not match at the moment.

src/generated/cloud/bigquery/reservation/v1/Cargo.toml
```diff
@@ -31,6 +31,7 @@ async-trait.workspace = true
 bytes.workspace      = true
 gax.workspace        = true
 gaxi                 = { workspace = true, features = ["_internal-http-client"] }
+iam_v1.workspace     = true
```
coryan, line 34: This is neat... I had forgotten we handle this automatically.

src/generated/showcase/src/lib.rs
suzmue: Ah yep. I'll send a PR to update the sidekick version separately.

coryan (approved):

PR #3474, opened 2025-10-06
chore: update sidekick version

suzmue:


src/generated/showcase/src/transport.rs
```diff
@@ -1476,7 +1482,13 @@ impl std::fmt::Debug for Echo {
 
 impl Echo {
     pub async fn new(config: gaxi::options::ClientConfig) -> gax::client_builder::Result<Self> {
+        let tracing_is_enabled = gaxi::options::tracing_enabled(&config);
         let inner = gaxi::http::ReqwestClient::new(config, crate::DEFAULT_HOST).await?;
+        let inner = if tracing_is_enabled {
+            inner.with_instrumentation(&crate::info::INSTRUMENTATION_CLIENT_INFO)
```
dbolduc, line 1488: cc: @westarle - **if** you wanted to add any tests, they might fit somewhere in: https://github.com/googleapis/google-cloud-rust/tree/main/src/integration-tests/src/showcase

PR #3479, opened 2025-10-07
feat(gax): Add TransportSummary trait and field to Parts

westarle:
This provides a way for transport implementations to pass data (like errors) up to higher layers to enrich spans that cover the entire request in a transport-agnostic way.

### DIFF
--- a/src/gax/src/response.rs
+++ b/src/gax/src/response.rs
@@ -62,6 +62,8 @@
 //! }
 //! ```
 
+use std::sync::{Arc, Mutex};
+
 /// Represents a Google Cloud service response.
 ///
 /// A response from a Google Cloud service consists of a body (potentially the
@@ -113,7 +115,7 @@
 ///     Ok(Response::from(body))
 /// }
 /// ```
-///   
+///
 #[derive(Clone, Debug)]
 pub struct Response<T> {
     parts: Parts,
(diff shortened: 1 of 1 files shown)


westarle:
For context, here is the sidekick change https://github.com/googleapis/librarian/compare/main...westarle:librarian:feat/librarian-t3-templates 

and an e2e prototype in Rust https://github.com/googleapis/google-cloud-rust/compare/main...westarle:impl/enable-t3-showcase

src/gax/src/response.rs
```diff
@@ -279,6 +286,54 @@ impl Parts {
...
+    /// This is intended for internal use by observability helpers.
+    pub fn get_transport_summary<T>(response: &Response<T>) -> Option<Box<dyn TransportSummary>> {
+        let guard = response
+            .parts
+            .transport_summary
+            .lock()
+            .expect("Mutex poisoned");
+        guard.clone()
+    }
+}
```
coryan, line 336: nit: missing blank line.

src/gax/src/response.rs
```diff
@@ -242,6 +244,11 @@ impl<T> Response<T> {
 pub struct Parts {
     /// The HTTP headers or the gRPC metadata converted to HTTP headers.
     pub headers: http::HeaderMap<http::HeaderValue>,
+    // Internal field for transport-specific observability data.
+    // Wrapped in Arc<Mutex<>> to maintain UnwindSafe for Response<T>,
```
coryan, line 248: This is a property you can unit test... Please add the corresponding test.

src/gax/src/response.rs
```diff
@@ -242,6 +244,11 @@ impl<T> Response<T> {
...
+    // as the boxed trait object may not be UnwindSafe.
+    // See https://github.com/googleapis/google-cloud-rust/issues/3463 for details.
+    transport_summary: Arc<Mutex<Option<Box<dyn internal::TransportSummary>>>>,
```
coryan, line 251: I thought you were going to add a concrete type so we could avoid all the `Arc<Mutex<...>` and still maintain `UnwindSafe`.  Did I misunderstand something?

coryan (changes requested):
We may need a more detailed design discussion here.  Using `Any` should be a last resort, that is very close to passing  `void*` around.

src/gax/src/response.rs
westarle: I wanted to, but I was (1) concerned about publishing the full span record to gax, and (2) it was convenient to use with https://github.com/googleapis/google-cloud-rust/compare/main...westarle:impl/enable-t3-showcase#diff-ff2816a0c00f3a7cc87ca8f6ec84e7888b5536f91f0a4ff5794b24be35cb4ee1R334-R353

I'll play around a bit more to try to avoid the Any, we can discuss options.

westarle: I think this could work better, thank you for the review.

https://github.com/googleapis/google-cloud-rust/pull/3482

westarle:
https://github.com/googleapis/google-cloud-rust/pull/3482

PR #3480, opened 2025-10-07
feat(storage): accept String payload for write_object

dbolduc:


coryan (approved):
> I guess one could also use `String::as_str`, but fewer utterances are better for our customers.

I don't think so. I think we only accept `&'static str`s. But yeah, they could say `bytes::Bytes::from(s)`.

I was just surprised `String` didn't work and the compiler suggestions relied on knowledge of `BytesSource` to be useful. :shrug:

PR #3481, opened 2025-10-07
impl(gax-internal): Update error reporting in HttpSpanInfo

westarle:
This PR enhances error reporting for HTTP client spans in `google-cloud-gax-internal` by introducing a structured `ErrorType` enum.

**Changes:**

*   Replaced the string-based `error_type` in `HttpSpanInfo` with the new `ErrorType` enum.
*   Implemented logic within `ErrorType` to classify errors from HTTP responses and `reqwest` client errors.
*   Added methods to `ErrorType` to generate values for `error.type`, `rpc.grpc.status_code`, and `grpc.status` attributes.
*   Updated span creation and attribute recording to use `ErrorType`, ensuring these new attributes are included on error spans.
*   Refactored tests to align with these changes.

This change improves the clarity and consistency of error information on OpenTelemetry spans.

For #3239

### DIFF
--- a/src/gax-internal/src/observability/errors.rs
+++ b/src/gax-internal/src/observability/errors.rs
@@ -62,21 +62,18 @@ impl ReqwestErrorDetails for reqwest::Error {
 }
 
 impl ErrorType {
-    pub(crate) fn from_reqwest_error(err: &dyn ReqwestErrorDetails) -> Self {
-        if err.is_timeout() {
-            ErrorType::ClientTimeout
-        } else if err.is_connect() {
-            ErrorType::ClientConnectionError
-        } else if err.is_request() {
-            ErrorType::ClientRequestError
-        } else if err.is_body() {
-            ErrorType::ClientRequestBodyError
-        } else if err.is_decode() {
-            ErrorType::ClientResponseDecodeError
-        } else if err.is_redirect() {
-            ErrorType::ClientRedirectError
-        } else {
-            ErrorType::Internal
+    pub(crate) fn from_reqwest_error<E>(err: &E) -> Self
+    where
+        E: ReqwestErrorDetails,
+    {
+        match err {
+            e if e.is_timeout() => ErrorType::ClientTimeout,
+            e if e.is_connect() => ErrorType::ClientConnectionError,
+            e if e.is_request() => ErrorType::ClientRequestError,
+            e if e.is_body() => ErrorType::ClientRequestBodyError,
+            e if e.is_decode() => ErrorType::ClientResponseDecodeError,
+            e if e.is_redirect() => ErrorType::ClientRedirectError,
+            _ => ErrorType::Internal,
         }
     }
 
@@ -134,6 +131,7 @@ impl ErrorType {
 pub(crate) mod tests {
     use super::*;
     use http::StatusCode;
+    use test_case::test_case;
 
     #[derive(Default)]
     pub struct MockReqwestError {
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/observability.rs
```diff
@@ -71,6 +77,105 @@ impl OtelStatus {
...
+    ClientRequestBodyError,
+    ClientResponseDecodeError,
+    ClientRedirectError,
+    ClientAuthenticationError,
+    Internal,
+}
+
+impl ErrorType {
+
+    pub(crate) fn from_reqwest_error(err: &reqwest::Error) -> Self {
```
westarle: Any ideas how to test this?  reqwest::Error is hard to mock or create.

coryan: It is indeed.  There are several approaches here

* The lazy person approach (what I do): create an http server that returns an error can make a call to it.
* Create a local trait that is implemented by `reqwest::Error` and some local type that also implements the trait and you use for testing.

src/gax-internal/src/observability/errors.rs
```diff
@@ -0,0 +1,388 @@
+// Copyright 2025 Google LLC
```
dbolduc, line 1: Code looks good, but we need to include `mod errors` in `observability.rs` to be sure

westarle: done!

westarle:
Thanks for the review @dbolduc please note the scope change (I integrated it with tracing code.)

src/gax-internal/src/observability/errors.rs
```diff
@@ -0,0 +1,329 @@
...
+    fn is_decode(&self) -> bool {
+        self.is_decode()
+    }
+    fn is_redirect(&self) -> bool {
+        self.is_redirect()
+    }
+}
+
+impl ErrorType {
+    pub(crate) fn from_reqwest_error(err: &dyn ReqwestErrorDetails) -> Self {
```
coryan: Prefer generics over `dyn`:


```suggestion
    pub(crate) fn from_reqwest_error<E>(err: &E) -> Self 
    where E: ReqwestErrorDetails {
```

src/gax-internal/src/observability/errors.rs
```diff
@@ -0,0 +1,329 @@
...
+        if err.is_timeout() {
+            ErrorType::ClientTimeout
+        } else if err.is_connect() {
+            ErrorType::ClientConnectionError
```
coryan: optional: I think I would have used a match:


```suggestion
       match err {
         e if e.is_timeout() => ErrorType::ClientTimeout.
         e if e.is_connect() => ErrorType::ClientConnectionError.
         ... ...
         _ => Error::Internal
```

src/gax-internal/src/observability/errors.rs
```diff
@@ -0,0 +1,329 @@
...
+
+    fn client_error(error_type: ErrorType, name: &str, code: Code) -> (ErrorType, String, Code) {
+        (error_type, name.to_string(), code)
+    }
+
+    #[test]
+    fn test_error_type_conversions() {
+        // Table of test case tuples.
+        // (error_type: ErrorType, expected_as_str: String, expected_grpc_code: Code)
+        let test_cases = vec![
```
coryan: Why not [test_case](https://docs.rs/test-case/3.3.1/test_case/)?  We use that in a lot of places with good results.

src/gax-internal/src/observability/errors.rs
```diff
@@ -0,0 +1,329 @@
...
+                expected_grpc_code.name().to_string(),
+                "grpc_status for {:?}",
+                error_type
+            );
+        }
+    }
+
+    #[test]
+    fn test_from_reqwest_error() {
+        let test_cases = vec![
```
coryan: Ditto, why not `test_case`?

PR #3482, opened 2025-10-07
impl(gax): Define concrete TransportSpanInfo struct

westarle:
Defines the TransportSpanInfo struct within the internal module to propagate transport-level details.

### DIFF
--- a/src/gax/src/response.rs
+++ b/src/gax/src/response.rs
@@ -113,7 +113,7 @@
 ///     Ok(Response::from(body))
 /// }
 /// ```
-///   
+///
 #[derive(Clone, Debug)]
 pub struct Response<T> {
     parts: Parts,
@@ -235,7 +235,7 @@ impl<T> Response<T> {
 ///     Some(&http::HeaderValue::from_static("application/json"))
 /// );
 /// ```
-///  
+///
 /// [tower]: https://github.com/tower-rs/tower
 #[derive(Clone, Debug, Default)]
 #[non_exhaustive]
@@ -243,7 +243,8 @@ pub struct Parts {
     /// The HTTP headers or the gRPC metadata converted to HTTP headers.
     pub headers: http::HeaderMap<http::HeaderValue>,
     // Internal field for transport-specific observability data.
-    pub(crate) client_request_span_info: Option<internal::ClientRequestSpanInfo>,
+    #[cfg(google_cloud_unstable_tracing)]
+    pub(crate) transport_span_info: Option<internal::TransportSpanInfo>,
 }
 
 impl Parts {
@@ -281,6 +282,7 @@ impl Parts {
     }
 }
 
+#[cfg(google_cloud_unstable_tracing)]
 #[cfg_attr(not(feature = "_internal-semver"), doc(hidden))]
 pub mod internal {
     //! This module contains implementation details. It is not part of the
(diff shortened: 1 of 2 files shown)


src/gax/src/response.rs
```diff
@@ -279,6 +281,89 @@ impl Parts {
...
+    #[derive(Debug, Clone, Default)]
+    pub struct ClientRequestSpanInfo {
+        // OpenTelemetry Semantic Conventions
+        /// The RPC system used.
+        ///
+        /// Examples: "grpc", "http"
+        pub rpc_system: Option<String>,
+        /// The full name of the service.
+        ///
+        /// Examples: "google.cloud.secretmanager.v1.SecretManagerService", "secretmanager.googleapis.com"
```
dbolduc: I think we should only ever use one of these.

dbolduc: I guess for something like a token refresh, we don't have an associated proto method. Meh.

src/gax/src/response.rs
```diff
@@ -279,6 +281,89 @@ impl Parts {
...
+        pub rpc_service: Option<String>,
+        /// The name of the method.
+        ///
+        /// Examples: "AccessSecretVersion", "GET"
```
dbolduc: Same. Seems like `http_request_method` is covering the `GET` case.

src/gax/src/response.rs
```diff
@@ -279,6 +281,89 @@ impl Parts {
...
+        pub url_domain: Option<String>,
+        /// A low-cardinality classification of the error.
+        ///
+        /// Examples: "404", "RATE_LIMIT_EXCEEDED", "CLIENT_TIMEOUT"
+        pub error_type: Option<String>,
+
+        // Custom GCP Attributes
+        /// The primary GCP resource being targeted.
+        ///
+        /// Example: //storage.googleapis.com/b/my-bucket
```
dbolduc: comment: Seems nice but difficult for us to compute

src/gax/src/response.rs
```diff
@@ -279,6 +281,89 @@ impl Parts {
...
+        pub gcp_resource_name: Option<String>,
+    }
+
+    /// Gets a reference to the ClientRequestSpanInfo from the response.
+    pub fn get_client_request_span_info<T>(
```
dbolduc: nit: drop the `get_`. It is not Rusty.

https://rust-lang.github.io/api-guidelines/naming.html#getter-names-follow-rust-convention-c-getter

src/gax/src/response.rs
```diff
@@ -279,6 +281,89 @@ impl Parts {
...
+    ) -> Option<&ClientRequestSpanInfo> {
+        response.parts.client_request_span_info.as_ref()
+    }
+
+    /// Sets the ClientRequestSpanInfo on the response.
+    pub fn set_client_request_span_info<T>(
+        response: &mut super::Response<T>,
+        info: Option<ClientRequestSpanInfo>,
+    ) {
+        response.parts.client_request_span_info = info;
```
dbolduc: comment: so we can only set all fields or nothing?

Some of the fields we know at compile time e.g. `rpc_service`, other fields like `url_domain` we know when the client is made, other fields like `grpc_status` we do not know until a request finishes.

is this a problem?

westarle: Good point, I think I have some extra information in here.  The real intent of this is to collect information from the response in a transport-agnostic way to update a span and ping the duration metric for end-to-end "single-RPC" latency.

I will reduce the scope to just info we need from the underlying transport at "response time".

src/gax/src/response.rs
westarle: I reduced the scope so I think this is obsolete.

src/gax/src/response.rs
westarle: this isn't present anymore.

src/gax/src/response.rs
westarle: done

src/gax/src/response.rs
westarle: reduced to just things that can change in the process of a request.

dbolduc (approved):

PR #3483, opened 2025-10-07
docs(storage): user guide for mocking `Storage`

### LINKED ISSUE #3226: Mocking storage samples + explain rationale

dbolduc:
Fixes #3226 

[mocking_storage.pdf](https://github.com/user-attachments/files/22754209/mocking_storage.pdf)


guide/src/storage/mocking.md
```diff
@@ -0,0 +1,157 @@
...
+Applications can use mocks to write controlled, reliable unit tests that do not
+involve network calls, and do not incur billing.
+
+In this guide, you will learn:
+
+- Why the design of the `Storage` client deviates from the design of other
+  Google Cloud clients
+- How to write testable interfaces using the `Storage` client
+- How to mock reads
+- How to mock writes
```
coryan: I think the why is important, but tutorials / guides should emphasize how to do things... Why is for "in-depth guides":
 
```suggestion
- How to write testable interfaces using the `Storage` client
- How to mock reads
- How to mock writes
- Why the design of the `Storage` client deviates from the design of other
  Google Cloud clients
```

guide/src/storage/mocking.md
```diff
@@ -0,0 +1,157 @@
...
+
+This guide is specifically for mocking the `Storage` client. For a generic
+mocking guide (which applies to the `StorageControl` client), see
+[How to write tests using a client](../mock_a_client.md).
+
+## Design rationale
```
coryan: Consider moving this to the end or even a separate page.

guide/src/storage/mocking.md
```diff
@@ -0,0 +1,157 @@
...
+
+### Other clients
+
+Most clients, such as `StorageControl` hold a boxed, `dyn`-compatible
+implementation of the stub trait internally. They use dynamic dispatch to
+forward requests from the client to their stub (which could be the real
+implementation or a mock).
+
+Because dynamic dispatch is used, the exact type of the stub does not need to be
```
coryan:
```suggestion
Because these clients use dynamic dispatch, the exact type of the stub does not need to be
```

guide/src/storage/mocking.md
```diff
@@ -0,0 +1,157 @@
...
+known by the compiler. The clients do not need to be templated on their stub.
```
coryan: "templated" is a C++-ism, "generic on"?

guide/src/storage/mocking.md
```diff
@@ -0,0 +1,157 @@
...
+
+Applications that do not need to test their code can simply write all interfaces
+in terms of `Storage`. The default `T` is the real implementation of the client.
+
+```rust,ignore,noplayground
+{{#rustdoc_include ../../samples/tests/storage/mocking.rs:prod-only-interface}}
+```
+
+Applications that need to test their code, should write their interfaces in
+terms of the generic `T`.
```
coryan:
```suggestion
Applications that need to test their code, should write their interfaces in
terms of the generic `T` with the correct constraints.
```

or maybe "constrained as needed"?

guide/src/storage/mocking.md
```diff
@@ -0,0 +1,157 @@
...
+
+```rust,ignore,noplayground
+{{#rustdoc_include ../../samples/tests/storage/mocking.rs:testable-interface}}
+```
+
+## Mocking reads
+
+Let's say we have an application function which downloads an object and counts
```
coryan: Check the style guide.  Google docs use the second person singular... "In this guide you will test an application function ..."

guide/src/storage/mocking.md
```diff
@@ -0,0 +1,157 @@
...
+how many newlines it contains.
+
+```rust,ignore,noplayground
+{{#rustdoc_include ../../samples/tests/storage/mocking.rs:count-newlines}}
+```
+
+We want to test our code against a known response from the server. We can do
+this by faking the `ReadObjectResponse`.
```
coryan: Here too, change to use "you".

guide/src/storage/mocking.md
```diff
@@ -0,0 +1,157 @@
...
+A `ReadObjectResponse` is essentially a stream of bytes. You can create a fake
+`ReadObjectResponse` in tests by supplying a payload to
+`ReadObjectResponse::from_source`. The library accepts the same payload types as
+`Storage::write_object`.
+
+```rust,ignore,noplayground
+{{#rustdoc_include ../../samples/tests/storage/mocking.rs:fake-read-object-resp}}
+```
+
+We define the mock as usual, using `mockall`.
```
coryan: I am not sure "as usual" is quite right.  It is usual for us in this project, applications may use something else.  Maybe "This guide uses the `mockall` crate to create a mock. You can use a different crate in your tests" ?

guide/src/storage/mocking.md
dbolduc: Done.

guide/src/storage/mocking.md
dbolduc: Done.

guide/src/storage/mocking.md
dbolduc: Done.

guide/src/storage/mocking.md
dbolduc: Fixed.

guide/src/storage/mocking.md
dbolduc: Done.

guide/src/storage/mocking.md
dbolduc: Used second person, singular.

I like saying "we" because it makes me feel like I am on the same team as the reader. It's fine though.

guide/src/storage/mocking.md
dbolduc: Fixed.

guide/src/storage/mocking.md
dbolduc: Done. FWIW, I think `mockall` is the main mocking framework in the ecosystem.

PR #3484, opened 2025-10-07
feat(auth): parse jwt id token as Token

alvarowolfx:
Towards #3449

### DIFF
--- a/src/auth/src/credentials/idtoken.rs
+++ b/src/auth/src/credentials/idtoken.rs
@@ -14,8 +14,14 @@
 
 use crate::Result;
 use crate::token::Token;
+use base64::{Engine as _, engine::general_purpose::URL_SAFE_NO_PAD};
+use gax::error::CredentialsError;
+use serde_json::Value;
+use std::collections::HashMap;
 use std::future::Future;
 use std::sync::Arc;
+use std::time::Duration;
+use tokio::time::Instant;
 
 /// Obtain [OIDC ID Tokens].
 ///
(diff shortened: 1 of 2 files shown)


src/auth/src/credentials/idtoken.rs
```diff
@@ -90,3 +96,80 @@ pub(crate) mod dynamic {
         }
     }
 }
+
+/// parse JWT ID Token string as google_cloud_auth::token::Token
+pub(crate) fn parse_id_token_from_str(token: String) -> Result<Token> {
```
coryan, line 100: You know that `Token` is not public, right?  Even if it was: how is this supposed to be used?  What would I do with one of those `Token`s parsed from a JWT ID Token?

alvarowolfx: oh, I haven't noticed that Token is not public. But in any case, there are two main purposes:
* to fill out the expires_at field that is used by the TokenCache internally and properly know when to fetch a new one
* for external users, they can access the claims like "aud", "email" and "email_verified" to check if the token has the expected target. But we can skip this, as we will need a provide a verify_token function anyway to properly verify the token signature and we can return the token claims so user can do the same verification. 

But now that I realized that Token is not public, I need to revisit the public IDTokenCredential trait to see what is going to be exposed to customers. Probably just return a String. cc @sai-sunder-s

sai-sunder-s: > for external users, they can access the claims like "aud", "email" and "email_verified" 

`Token` struct doesn't expose these directly. It just has the jwt string and the expiry we parse out from the jwt. Even though we have parsed the jwt, we are *currently* not adding them as fields in the Token struct and we need not do that as well.
The ID token is typically not used by the requestor themselves, they just use our library to get it and pass it along to another service like Cloud Run, who does the parsing and validation. (See the CUJs section in the design doc). So there is no reason for giving the caller a way to inspect the contents of it, they can just get the jwt string and send it to IAP or Cloud Run etc.

> we can return the token claims so user can do the same verification.

The user need not do the verification, they called us specifically to get the id token. It is the intended recipient (The service that the user is sending the id token to) who does the verification.

TLDR; lets just return the jwt string from the IDTokenCredential Trait, not Token struct. We will still parse it to get the expiry.

alvarowolfx: agree with all points @sai-sunder-s, my point was that we missed on the previous PR that the public trait should have returned just a String instead of Token. I've opened https://github.com/googleapis/google-cloud-rust/pull/3493 to address that. Also updated this PR to only focus on parsing the token and using just the `exp` field

src/auth/src/credentials/idtoken.rs
```diff
@@ -88,3 +95,87 @@ pub(crate) mod dynamic {
...
+        return Err(CredentialsError::from_msg(false, "invalid JWT token"));
+    }
+    let payload = URL_SAFE_NO_PAD
+        .decode(parts[1])
+        .map_err(|e| CredentialsError::from_source(false, e))?;
+
+    let claims: HashMap<String, Value> =
+        serde_json::from_slice(&payload).map_err(|e| CredentialsError::from_source(false, e))?;
+
+    let expires_at = claims["exp"].as_u64().and_then(instant_from_epoch_seconds);
```
sai-sunder-s, line 112: Is there some standard library that can do this parsing for us?

coryan: Maybe these can help:

https://github.com/googleapis/google-cloud-rust/blob/main/src/wkt/src/internal/uint64.rs

alvarowolfx: @sai-sunder-s you mean the whole jwt parsing or just the expiration and u64 -> Instant part ?

alvarowolfx: I did some experiments using https://docs.rs/jwt/latest/jwt/#verification, but felt like was not stricly need yet, but would be useful later when adding a verify function ( which is required per https://google.aip.dev/auth/4116#expected-behavior-for-verifying-id-tokens)

alvarowolfx: changes to use the `jwt` crate. Can't use the `jsonwebtoken` crate as it doesn't allow for parsing jwt without verifying the signature. 

```diff
diff --git a/Cargo.toml b/Cargo.toml
index 650ad5ce5..3d0f8348a 100644
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -328,6 +328,8 @@ uuid               = { default-features = false, version = "1", features = ["v4"
 clap               = { default-features = false, version = "4" }
 parse-size         = { default-features = false, version = "1" }
 humantime          = { default-features = false, version = "2" }
+jwt                = { default-features = false, version = "0.16.0" }
+hmac               = { default-features = false, version = "0.12" }
 
 opentelemetry-semantic-conventions = { default-features = false, version = "0.30.0", features = [
   "semconv_experimental",
diff --git a/src/auth/Cargo.toml b/src/auth/Cargo.toml
index 98f420d4a..756dd0273 100644
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -46,6 +46,9 @@ thiserror.workspace   = true
 time                  = { workspace = true, features = ["serde"] }
 tokio                 = { workspace = true, features = ["fs", "process"] }
 bon.workspace         = true
+sha2.workspace        = true
+jwt.workspace         = true
+hmac.workspace        = true
 # Local dependencies
 gax.workspace = true
 
diff --git a/src/auth/src/credentials/idtoken.rs b/src/auth/src/credentials/idtoken.rs
index 6bbc2201b..f2814c3db 100644
--- a/src/auth/src/credentials/idtoken.rs
+++ b/src/auth/src/credentials/idtoken.rs
@@ -14,8 +14,8 @@
 
 use crate::Result;
 use crate::token::Token;
-use base64::{Engine as _, engine::general_purpose::URL_SAFE_NO_PAD};
 use gax::error::CredentialsError;
+use jwt::{Token as JwtToken, header::Header};
 use serde_json::Value;
 use std::collections::HashMap;
 use std::future::Future;
@@ -98,16 +98,10 @@ pub(crate) mod dynamic {
 
 /// parse JWT ID Token string as google_cloud_auth::token::Token
 pub(crate) fn parse_id_token_from_str(token: String) -> Result<Token> {
-    let parts: Vec<&str> = token.split('.').collect();
-    if parts.len() != 3 {
-        return Err(CredentialsError::from_msg(false, "invalid JWT token"));
-    }
-    let payload = URL_SAFE_NO_PAD
-        .decode(parts[1])
-        .map_err(|e| CredentialsError::from_source(false, e))?;
-
-    let claims: HashMap<String, Value> =
-        serde_json::from_slice(&payload).map_err(|e| CredentialsError::from_source(false, e))?;
+    let jwt_token: JwtToken<Header, HashMap<String, Value>, _> =
+        JwtToken::parse_unverified(token.as_str())
+            .map_err(|e| CredentialsError::from_source(false, e))?;
+    let claims = jwt_token.claims();
 
     let expires_at = claims["exp"].as_u64().and_then(instant_from_epoch_seconds);
 
@@ -133,8 +127,10 @@ fn instant_from_epoch_seconds(secs: u64) -> Option<Instant> {
 pub(crate) mod tests {
     use super::parse_id_token_from_str;
     use super::*;
-    use base64::prelude::BASE64_URL_SAFE_NO_PAD;
+    use hmac::{Hmac, Mac};
+    use jwt::SignWithKey;
     use serial_test::parallel;
+    use sha2::Sha256;
     use std::time::{Duration, SystemTime, UNIX_EPOCH};
 
     type TestResult = anyhow::Result<()>;
@@ -152,10 +148,10 @@ pub(crate) mod tests {
             "iat": now.as_secs(),
         });
 
-        let json = serde_json::to_string(&claims).expect("failed to encode jwt claims");
-        let payload = BASE64_URL_SAFE_NO_PAD.encode(json.as_bytes());
+        let key: Hmac<Sha256> =
+            Hmac::new_from_slice(b"test-secret").expect("failed to create jwt secret");
 
-        format!("test_header.{}.test_signature", payload)
+        claims.sign_with_key(&key).expect("failed to sign jwt")
     }
 
     #[tokio::test]

```

sai-sunder-s: The diff looks good, this is what I meant.

PR #3485, opened 2025-10-07
test(auth): add integration tests for mds sourced id token

alvarowolfx:
Add integration tests for MDS sourced ID Tokens

Towards #3449 #3587

### DIFF
--- a/src/auth/integration-tests/src/lib.rs
+++ b/src/auth/integration-tests/src/lib.rs
@@ -20,12 +20,10 @@ use auth::credentials::{
         ProgrammaticBuilder as ExternalAccountProgrammaticBuilder,
     },
     impersonated::Builder as ImpersonatedCredentialsBuilder,
-    mds::idtoken::Builder as IDTokenMDSBuilder,
     service_account::Builder as ServiceAccountCredentialsBuilder,
     subject_token::{Builder as SubjectTokenBuilder, SubjectToken, SubjectTokenProvider},
 };
 use auth::errors::SubjectTokenProviderError;
-use base64::{Engine as _, engine::general_purpose::URL_SAFE_NO_PAD};
 use bigquery::client::DatasetService;
 use gax::error::rpc::Code;
 use httptest::{Expectation, Server, matchers::*, responders::*};
@@ -34,7 +32,6 @@ use language::client::LanguageService;
 use language::model::Document;
 use scoped_env::ScopedEnv;
 use secretmanager::{client::SecretManagerService, model::SecretPayload};
-use serde_json::Value;
 use std::sync::Arc;
 
 pub async fn service_account() -> anyhow::Result<()> {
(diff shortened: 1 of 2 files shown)


src/auth/integration-tests/src/lib.rs
```diff
@@ -525,3 +525,48 @@ impl SubjectTokenProvider for TestSubjectTokenProvider {
...
+        let audience = "https://example.com";
+
+        // Get the service account email from the metadata server directly
+        let client = reqwest::Client::new();
+        let expected_email = client
+            .get("http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/email")
+            .header("Metadata-Flavor", "Google")
+            .send()
+            .await
+            .expect("failed to get service account email from metadata server")
```
dbolduc, line 545: Hmm... Consider skipping the rest of the test if there is no MDS. Then the test is not environment dependent.

The downside is that we might write a bad request here, not notice, and then get no coverage of the ID token logic. :shrug:

src/auth/integration-tests/src/lib.rs
```diff
@@ -525,3 +525,48 @@ impl SubjectTokenProvider for TestSubjectTokenProvider {
...
+            .text()
+            .await
+            .expect("failed to read service account email from metadata server response");
+
+        // Only works when running on an env that has MDS.
```
dbolduc, line 550: I don't like this.

src/auth/integration-tests/src/lib.rs
```diff
@@ -525,3 +525,48 @@ impl SubjectTokenProvider for TestSubjectTokenProvider {
...
+        let id_token_creds = IDTokenMDSBuilder::new(audience)
+            .with_format("full")
+            .build()
+            .expect("failed to create id token credentials");
```
dbolduc, line 554: nit: why panic with `expect(...)` instead of surfacing error with the `?` operator?

(Here and elsewhere)

src/auth/integration-tests/src/lib.rs
dbolduc: Or you could copy @coryan's approach from #3614

alvarowolfx: we already have other tests in the auth integration test that depend on MDS. That change on using an env variable flag for that would need to be applied to other tests. Also, when developing from a VM, the MDS is available, so we would have to pass those extra flags. It would be nice if we can detect if MDS is available in a different way.

src/auth/integration-tests/src/lib.rs
alvarowolfx: me neither haha had to adapt since I came back working on auth stuff, since for building auth features, a lot of them have MDS related features.

src/auth/integration-tests/src/lib.rs
alvarowolfx: it's easier to see where it failed, I remember debugging some flaky issues because and without that, we only now the top level function that failed, but no exactly the line that caused it.

src/auth/integration-tests/src/lib.rs
dbolduc: Ok, I am cool with not bothering to make the auth tests run outside of GCP.

I haven't run them in months, and when I did, I used my cloudtop.

src/auth/integration-tests/src/lib.rs
dbolduc: Ack thanks. I guess that is a downside of the test driver / test library set up.

Issue #3477, opened 2025-10-07
Implement LRO helpers for Compute


coryan:
We will need to introduce new traits and new `Poller` implementations.

PR #3486, opened 2025-10-08
cleanup(lro): simplify AIP-151 implementation

coryan:
Motivated by #3477


src/lro/src/internal/aip151.rs
```diff
@@ -210,17 +210,7 @@ fn map_polling_metadata<R>(result: PollingResult<R, wkt::Empty>) -> PollingResul
 ///   It receives the name of the operation as its only input parameter. It
 ///   should have captured any stubs and request options.
 /// * `QF` - the type of future returned by `Q`.
```
dbolduc: we should clean up the comments too

coryan: Done.

coryan (commented):
PTAL

PR #3489, opened 2025-10-08
ci(auth): Cloud scheduler to refresh auth terraform once a week

PhongChuong:


src/auth/.gcb/builds/triggers/main.tf
```diff
@@ -87,11 +87,15 @@ resource "google_storage_bucket" "build-cache" {
   }
 }
 
-# This service account is created externally. It is used for all the builds.
+# These service account is created externally. It is used for all the builds.
```
coryan: nit:


```suggestion
# These service accounts are created externally. They are used in all the builds.
```

src/auth/.gcb/builds/triggers/main.tf
```diff
@@ -236,3 +244,43 @@ resource "google_cloudbuild_trigger" "post-merge" {
...
+  schedule    = "0 0 1 * *" # Midnight on 1st day of the month	
+
+  pubsub_target {
+    topic_name = google_pubsub_topic.terraform_runner_topic.id
+    data       = base64encode("sync")
+  }
+}
+
+resource "google_cloudbuild_trigger" "pubsub-trigger" {
+  for_each = tomap(local.pubsub_builds)
```
dbolduc: No need for this to be a loop.

We use a loop for the integration build so we can have a triggers that runs on PR and one that runs on merge.

src/auth/.gcb/terraform.yaml
```diff
@@ -0,0 +1,34 @@
...
+    script: |
+      #!/usr/bin/env bash
+      set -e
+      apt-get update && apt-get install -y gnupg software-properties-common wget
+      wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor | tee /usr/share/keyrings/hashicorp-archive-keyring.gpg > /dev/null
+      gpg --no-default-keyring --keyring /usr/share/keyrings/hashicorp-archive-keyring.gpg --fingerprint
+      echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(grep -oP '(?<=UBUNTU_CODENAME=).*' /etc/os-release || lsb_release -cs) main" | tee /etc/apt/sources.list.d/hashicorp.list
+      apt-get update && apt-get install -y terraform
+      echo "Running terraform"
+      cd /workspace/src/auth/.gcb/builds
```
coryan, line 31: Why not use the terraform docker image and save yourself the installation?

src/auth/.gcb/builds/triggers/main.tf
dbolduc: nit: "These service accounts"

src/auth/.gcb/builds/triggers/main.tf
```diff
@@ -165,6 +169,10 @@ locals {
   builds = {
     integration = {}
   }
+
```
coryan: Optional: Maybe we can simplify things if we believe only one build will run on a schedule?  Your call.

src/auth/.gcb/builds/triggers/main.tf
```diff
@@ -236,3 +244,43 @@ resource "google_cloudbuild_trigger" "post-merge" {
...
+resource "google_pubsub_subscription" "terraform_runner_sub" {
+  name  = "terraform-sub"
+  topic = google_pubsub_topic.terraform_runner_topic.name
+}
+
+resource "google_cloud_scheduler_job" "job" {
+  name        = "terraform-job"
+  description = "Periodically sync terraform build"
+  schedule    = "0 0 1 * *" # Midnight on 1st day of the month
```
dbolduc: Our keys go stale after 90 days, and we are set to rotate them after 60 days.

https://github.com/googleapis/google-cloud-rust/blob/ad638e9aca6f95fa31e3f856501f6fc08cd42fdd/src/auth/.gcb/builds/service_account_test/main.tf#L27

The window where the key can go stale is barely too short for this schedule.

We should either
(1) run terraform more often. Once every other week? Once a week?
(2) set the key to rotate when it is 45 days old.

I lean towards (1) because it can alert us to a bad terraform config sooner, but I am totally cool with either option.

src/auth/.gcb/builds/triggers/main.tf
```diff
@@ -165,6 +169,10 @@ locals {
   builds = {
     integration = {}
   }
+
+  pubsub_builds = {
+    terraform = {}
+  }
```
dbolduc: No need for this local. It should be simpler to inline it.

dbolduc (changes requested):
Looks good, just have a comment about the schedule and a simplification of the config

src/auth/.gcb/builds/triggers/main.tf
dbolduc: Actually, we should have separate comments. `integration-test-runner` is used for the integration build. `terraform-runner` is used for the terraform build.

src/auth/.gcb/builds/triggers/main.tf
PhongChuong: Updated to once a week.

src/auth/.gcb/builds/triggers/main.tf
PhongChuong: Updated comment.

src/auth/.gcb/builds/triggers/main.tf
PhongChuong: Simplified to a single build.

src/auth/.gcb/terraform.yaml
PhongChuong: IIRC, the terraform docker does not come with shell so we would have to use it within our own docker image (or build our own). I found it more simple to avoid docker within docker.

src/auth/.gcb/builds/triggers/main.tf
```diff
@@ -236,3 +241,42 @@ resource "google_cloudbuild_trigger" "post-merge" {
...
+  pubsub_target {
+    topic_name = google_pubsub_topic.terraform_runner_topic.id
+    data       = base64encode("sync")
+  }
+}
+
+resource "google_cloudbuild_trigger" "pubsub-trigger" {
+  location = var.region
+  name     = "gcb-pubsub-terraform"
+  filename = ".gcb/terraform.yaml"
```
dbolduc, line 268: should be `src/auth/.gcb/terraform.yaml`

PR #3490, opened 2025-10-08
refactor(gax-internal): Organize observability into sub-modules

westarle:
Restructures the observability.rs file into a directory module with sub-modules for attributes and http_tracing to improve organization. No functional changes.


src/gax-internal/src/observability.rs
coryan, line 1: We avoid `mod.rs` and use `observability.rs`.  A bit non-standard, but legal, and easier when opening files from the IDE.

westarle: done

PR #3492, opened 2025-10-08
docs: update generate instructions to use GOPROXY=direct

suzmue:
We want to be able to update the generated code immediately after submitting code to Github. The proxy will cache results for lists and branch resolutions. To bypass this, use GOPROXY=direct to bypass the proxy when getting the version and for `go run`.

The downloaded version of sidekick will still be verified against the checksum database.

It is possible that the versions will differ if there is a commit in between running the two commands, but our CI will catch that.

### DIFF
--- a/doc/contributor/howto-guide-generated-code-maintenance.md
+++ b/doc/contributor/howto-guide-generated-code-maintenance.md
@@ -142,18 +142,17 @@ go -C ../librarian run ./cmd/sidekick refreshall -project-root $PWD && cargo fmt
 Once the changes work then send a PR in the librarian repo to make your changes.
 Wait for the PR to be approved and merged.
 
+Then finish your PR in `google-cloud-rust` by running sidekick again:
+
+```bash
+GOPROXY=direct go run github.com/googleapis/librarian/cmd/sidekick@main refreshall && cargo fmt
+```
+
 Find out what is the new version of librarian:
 
 ```bash
 GOPROXY=direct go list -m -u -f '{{.Version}}' github.com/googleapis/librarian@main
 ```
-
-Then finish your PR in `google-cloud-rust` by running sidekick again using the new version of librarian. For example:
-
-```bash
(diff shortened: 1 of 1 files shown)


doc/contributor/howto-guide-generated-code-maintenance.md
```diff
@@ -142,18 +142,19 @@ go -C ../librarian run ./cmd/sidekick refreshall -project-root $PWD && cargo fmt
 Once the changes work then send a PR in the librarian repo to make your changes.
 Wait for the PR to be approved and merged.
 
-Then finish your PR in `google-cloud-rust` by running sidekick again:
+Find out what is the new version of librarian:
 
 ```bash
-go run github.com/googleapis/librarian/cmd/sidekick@main refreshall && cargo fmt
+GOPROXY=direct go list -m -u -f '{{.Version}}' github.com/googleapis/librarian@main
```
dbolduc: I liked that the old command was copy-paste-able. Consider something like:

```
LATEST=$(GOPROXY=direct go list -m -u -f '{{.Version}}' github.com/googleapis/librarian@main)
go run github.com/googleapis/librarian/cmd/sidekick@${LATEST} refreshall && cargo fmt
```

suzmue: I was considering doing, and probably not a big deal but technically its possible that:

We run to get the version for ci:
GOPROXY=direct go list -m -u -f '{{.Version}}' github.com/googleapis/librarian@main

Then `GOPROXY=direct go run github.com/googleapis/librarian/cmd/sidekick@main refreshall && cargo fmt`
and they are using two different versions. Meaning we want to both print and save the result of latest, so would be `export LATEST=` and it felt like more clutter to me. 

But this is probably overthinking it, librarian is not getting new commits every minute.

Will update to just use direct for both (this will still check that the bytes downloaded match the checksum at sum.golang.org so still has same safety guarantees).

dbolduc: Ah, I see. I think the PR you have is strictly better than the current instructions.

PR #3494, opened 2025-10-09
impl(compute/v1): LRO trait implementation

coryan:
Implement the `DiscoveryOperation` trait for `crate::model::Operation`.
This is the glue code between the generated types and the code in the
`google-cloud-lro` crate.

Part of the work for #3477


src/generated/cloud/compute/v1/src/operation.rs
```diff
@@ -0,0 +1,461 @@
...
+
+    #[test]
+    fn status_from_error() {
+        let input = Error::new();
+        let got = Status::from(&input);
+        assert_eq!(got, Status::default());
+
+        let input = Error::new().set_errors([
+            Errors::new().set_code("INTERNAL"),
+            Errors::new().set_code("UNAVAILABLE"),
```
dbolduc, line 229: Q: do we know if/when the service returns multiple errors? The loss of information is a bit sad.

src/generated/cloud/compute/v1/src/operation.rs
```diff
@@ -0,0 +1,461 @@
...
+        );
+    }
+
+    #[test]
+    fn status_details_from_localized_message() {
+        use rpc::model::LocalizedMessage as RpcLocalizedMessage;
+
+        let input = LocalizedMessage::new()
+            .set_locale("NYC")
+            .set_message("fuggedaboudit");
```
dbolduc, line 393: lol

src/generated/cloud/compute/v1/src/operation.rs
```diff
@@ -0,0 +1,461 @@
...
+        // The documentation for "quota_id" says that it is sometimes called "limit_name".
+        let r = value
+            .limit_name
+            .into_iter()
+            .fold(r, |r, v| r.set_quota_id(v));
+        let r = value
+            .metric_name
+            .into_iter()
+            .fold(r, |r, v| r.set_quota_metric(v));
+        // There is no way to represent representation for `value.rollout_status`
```
dbolduc: nit: s/way to represent //

coryan: Fixed.

src/generated/cloud/compute/v1/src/operation.rs
coryan: We don't know yet. Created #3501 to track this.

PR #3495, opened 2025-10-09
fix(auth): Read scope from impersonated json

sai-sunder-s:
If the impersonated credential json contains a `scope` field, use that instead of resorting to default scopes.

### DIFF
--- a/src/auth/src/credentials/impersonated.rs
+++ b/src/auth/src/credentials/impersonated.rs
@@ -436,60 +436,27 @@ impl Builder {
         TokenProviderWithRetry<ImpersonatedTokenProvider>,
         Option<String>,
     )> {
-        let (
-            source_credentials,
-            service_account_impersonation_url,
-            delegates,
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/impersonated.rs
```diff
@@ -432,46 +436,52 @@ impl Builder {
...
-                        extract_credential_type(&config.source_credentials)?;
-                    if source_credential_type == "impersonated_service_account" {
-                        return Err(BuilderError::parsing(
-                            "source credential of type `impersonated_service_account` is not supported. \
+        let (
+            source_credentials,
+            service_account_impersonation_url,
+            delegates,
+            quota_project_id,
+            scopes,
```
coryan: This tuple is getting too long. Can we use a struct with named fields?

coryan: Or a helper type with accessors?

PR #3496, opened 2025-10-09
test(auth): integration test for ADC flow and impersonated sourced id token

alvarowolfx:
Auth integration tests for id tokens.

Towards #3449 #3588 #3622 #3589

### DIFF
--- a/src/auth/integration-tests/tests/driver.rs
+++ b/src/auth/integration-tests/tests/driver.rs
@@ -90,6 +90,7 @@ mod driver {
 
     #[cfg(all(test, google_cloud_unstable_id_token))]
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[serial_test::serial]
     async fn run_id_token_adc() -> anyhow::Result<()> {
         auth_integration_tests::unstable::id_token_adc().await
     }


src/auth/integration-tests/src/lib.rs
```diff
@@ -525,3 +530,109 @@ impl SubjectTokenProvider for TestSubjectTokenProvider {
...
+
+        assert_eq!(claims["aud"], target_audience);
+        assert_eq!(claims["email"], expected_email);
+        assert_eq!(claims["email_verified"], true);
+
+        Ok(())
+    }
+
+    /// Generates a Google ID token using the impersonated::idtoken::Builder.
+    pub async fn generate_id_token(
```
sai-sunder-s: where is this being used?

src/auth/integration-tests/src/lib.rs
```diff
@@ -525,3 +530,109 @@ impl SubjectTokenProvider for TestSubjectTokenProvider {
...
+    }
+
+    fn parse_id_token(token: String) -> anyhow::Result<Value> {
+        // Decode the JWT and verify its claims
+        let parts: Vec<&str> = token.split('.').collect();
+        anyhow::ensure!(parts.len() == 3, "ID token is not a valid JWT");
+        let payload = URL_SAFE_NO_PAD.decode(parts[1])?;
+        let claims: Value = serde_json::from_slice(&payload)?;
+
+        Ok(claims)
```
coryan: nit: maybe 

```suggestion
        serde_json::from_slice(&payload).map_err(anyhow::Error::from)
```

src/auth/integration-tests/src/lib.rs
```diff
@@ -457,12 +456,18 @@ async fn get_secret_with_mds_creds(
     Ok(payload)
 }
 
+#[cfg(google_cloud_unstable_id_token)]
+use unstable::generate_id_token;
+
```
coryan: Why introduce `unstable::generate_id_token()` and then immediately define a function with the same name?

src/auth/integration-tests/Cargo.toml
```diff
@@ -30,6 +30,8 @@ scoped-env.workspace = true
 tempfile.workspace   = true
 serde_json.workspace = true
 httptest.workspace   = true
+base64.workspace     = true
+reqwest.workspace    = true
```
coryan, line 35: This does not seem to be used?

coryan (changes requested):
Largely fine, a few nites, and a question about `generate_id_token`. The last thing is what stops me from approving.

src/auth/integration-tests/src/lib.rs
alvarowolfx: when the integration tests for external account were written in the first place, we used the `IAMCredentials` client to generate id tokens. Now with the new features related to id token that was added to the auth package, we can just use those. But they are unstable features, so we keep using the previous method with the `iamcredentials` crate when the `google_cloud_unstable_id_token` is not set and use the new `impersonated::idtoken::Builder` when it's available. That's why it's showing two times cc @sai-sunder-s 

TL;DR - `lib::generated_id_token` is used on stable tests (which uses `IAMCredentials::Builder`) and `unstable::generate_id_token` is used on unstable tests (which uses `impersonated::idtoken::Builder`.

src/auth/integration-tests/src/lib.rs
alvarowolfx: see https://github.com/googleapis/google-cloud-rust/pull/3496/files#r2470225798

src/auth/integration-tests/Cargo.toml
alvarowolfx: those were being used on the MDS tests, this PR was broken down into smaller PRs. Now it's not showing as a diff here anymore.

src/auth/integration-tests/src/lib.rs
sai-sunder-s: Why not just continue using IAMCredentials for the 3P IDP until the id token feature is stabilized?

alvarowolfx: it is a way to have integration tests for impersonated::idtokens. But can be removed 🤷

sai-sunder-s (dismissed):

src/auth/integration-tests/tests/driver.rs
```diff
@@ -87,4 +87,22 @@ mod driver {
     async fn run_mds_id_token() -> anyhow::Result<()> {
         auth_integration_tests::unstable::mds_id_token().await
     }
+
+    #[cfg(all(test, google_cloud_unstable_id_token))]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn run_id_token_adc() -> anyhow::Result<()> {
+        auth_integration_tests::unstable::id_token_adc().await
+    }
```
coryan, line 96: This test needs to be marked `#[serial_test::serial]`  or it will flake, no?

alvarowolfx: good call. Added

PR #3498, opened 2025-10-10
refactor(ci): update CI for unstable flags

westarle:
This PR updates the GitHub Actions workflow to introduce a testing strategy for unstable features gated by --cfg flags.

   - Defines a central UNSTABLE_CFGS env variable for all active unstable flags.
   - Enables these flags in build, minimal-versions, and coverage jobs to test combined effects.
   - Adds a new test-unstable-cfg job to run key crate and integration tests with all unstable flags active.
   - Adds a new lint-unstable job to lint the workspace with all unstable flags active.
   - Other jobs test the stable configuration.

This ensures that unstable features are continuously tested without disrupting stable checks, and provides a clear signal on their impact.

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -24,6 +24,9 @@ on:
 env:
   GHA_RUST_VERSIONS: '{ "rust:msrv": "1.85", "rust:current": "1.90", "rust:nightly": "nightly" }'
   GHA_GO_VERSIONS: '{ "go:current": "1.25.0" }'
+  # Define the unstable flags once
+  UNSTABLE_CFGS: &unstable_cfgs "--cfg google_cloud_unstable_tracing"
+
 jobs:
   build:
     strategy:
@@ -34,6 +37,8 @@ jobs:
           - os: 'ubuntu-24.04'
             rust-version: 'rust:msrv'
     runs-on: ${{ matrix.os }}
+    env:
+      RUSTFLAGS: *unstable_cfgs
     steps:
       - uses: actions/checkout@v5
       - uses: actions/cache@v4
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -408,3 +408,6 @@ google-cloud-storage          = { version = "1", path = "src/storage" }
 google-cloud-test-utils = { path = "src/test-utils" }
 integration-tests       = { path = "src/integration-tests" }
 storage-samples         = { path = "src/storage/examples" }
+
+[workspace.lints.rust]
+unexpected_cfgs = { level = "deny", check-cfg = ['cfg(_google_cloud_unstable_tracing)'] }
```
westarle: I put "deny" to avoid typos.

I think this triggers when we put `#[cfg(not_in_list)]` in a rs file.

.github/workflows/sdk.yaml
```diff
@@ -30,24 +30,30 @@ jobs:
...
         os: ['macos-14', 'ubuntu-24.04', 'windows-2025']
         rust-version: ['rust:current']
+        unstable_flag: ['', '--cfg _google_cloud_unstable_tracing']
         include:
           - os: 'ubuntu-24.04'
             rust-version: 'rust:msrv'
+            unstable_flag: '' # Only run msrv without unstable flags
     runs-on: ${{ matrix.os }}
+    env:
+      RUSTFLAGS: ${{ matrix.unstable_flag }}
```
westarle: Alternatively we can put flags here.  WDYT?

.github/workflows/sdk.yaml
```diff
@@ -33,21 +33,29 @@ jobs:
         include:
           - os: 'ubuntu-24.04'
             rust-version: 'rust:msrv'
+          - os: 'ubuntu-24.04'
+            rust-version: 'rust:current'
+            unstable_flags: '--cfg _google_cloud_unstable_tracing'
```
coryan: Are we going to want many flags? or one?  Testing will growth as $2^{number of flags}$ no?

westarle: I think I want multiple flags to graduate items independently, but treat them largely as a bundle in our CI.

With o12y as an example, we can start generating docs and stabilizing APIs for tracing before we are ready to expose metrics...and we can do that behind the 'unstable' flag.

I plan to have build, coverage, minimal-versions unconditionally run with the UNSTABLE_FLAGS enabled to ensure the unstable code follows rules.

semver-checks will not have UNSTABLE_FLAGS (for now).

features, showcase, docs, and lint will use a matrix to run tests both with and without the UNSTABLE_FLAGS.

westarle:
docs/lint checks are in a weird state maybe because they changed names in this PR?

.github/workflows/sdk.yaml
westarle: OK I think I have it where I want it now, we should add new unstable flags to UNSTABLE_FLAGS.

.github/workflows/sdk.yaml
```diff
@@ -41,6 +43,8 @@ jobs:
           path: |
             ~/.cargo
           key: ${{ github.job }}-${{ runner.os }}-${{ matrix.rust-version }}-cargo-${{ hashFiles('Cargo.lock', '.github/workflows/sdk.yaml') }}
+      - name: Set RUSTFLAGS for the job
+        run: echo "RUSTFLAGS=${{ env.UNSTABLE_FLAGS }}" >> $GITHUB_ENV
```
westarle: This so we can collect the variable from `env` (that isn't available in the job's `env:` block.)

coryan:
> docs/lint checks are in a weird state maybe because they changed names in this PR?

Yes. do you gain a lot from using a matrix here?  The build runs on only one platform and with only one compiler. Maybe a different job would be less disruptive?

.github/workflows/sdk.yaml
coryan: I think if you had a separate job this would be easier.

westarle:
Thanks for the review, I'll give it more thought and revisit this on Tuesday!

> Yes. do you gain a lot from using a matrix here? The build runs on only one platform and with only one compiler. Maybe a different job would be less disruptive?

I think I can collect `docs`, `lint`, etc into a new job.

Jobs that can include flags always (i.e. `coverage`) I might still need the weird env stuff to avoid inlining the flags everywhere (and risk forgetting them.) Or alternatively we could decide on a single `unstable` flag to avoid the complexity.

`features` is complicated but would be a great test with unstable on / off to make sure published features work either way.  For now I could just (a) create a "unstable-features" that runs targeted tests for each feature (and keep it updated as features touch more things) or (b) update features test to set the flags inline.  In any case we'd add code so `google_cloud_unstable_tracing` would just run `gax-internal` tests for now, until we add the flag to `gax`, and so on.

coryan:
> Thanks for the review, I'll give it more thought and revisit this on Tuesday!

No rush.

> > Yes. do you gain a lot from using a matrix here? The build runs on only one platform and with only one compiler. Maybe a different job would be less disruptive?
> 
> I think I can collect `docs`, `lint`, etc into a new job.

Let's talk on your return ... I think your approach is correct, and certainly less cumbersome that what I hinted.  However, it is  doubling (more or less) the number of jobs. We want to stay within the free quota. We should check the usage metrics.

.github/workflows/sdk.yaml
```diff
@@ -34,6 +37,8 @@ jobs:
           - os: 'ubuntu-24.04'
             rust-version: 'rust:msrv'
     runs-on: ${{ matrix.os }}
+    env:
+      RUSTFLAGS: *unstable_cfgs
```
coryan, line 41: Fancy... It has been a few years since I saw references in YAML.

westarle:
Here's a test run that makes sure problems are caught:
https://github.com/googleapis/google-cloud-rust/actions/runs/18506622236

I introduced (1) "dead code" (unguarded) that is only used by code behind the flag; (2) lint exceptions guarded by a flag; (3) test failures guarded by the flag.

https://github.com/googleapis/google-cloud-rust/compare/7f351d1b651f7ab3a63403204ad79db2aaddd334..ad825ab5182c487104ab0462c117e89b17b64639

These issues are caught by multiple build rules, so there might be some room to trim things down if needed.

PR #3518, opened 2025-10-14
chore(compute/v1): enable all resources

coryan:


src/generated/cloud/compute/v1/Cargo.toml
```diff
@@ -28,27 +28,163 @@ rust-version.workspace = true
 
 [features]
 default = [
```
dbolduc, line 30: Just a thought, not for this PR. We could handpick the most common resources for this list so first time users do not have to build the whole thing.

(It is a decision we could always revert. Whereas we wouldn't be allowed to go from all features enabled to only a subset after 1.0)

coryan: Yes. I think it is time to implement support for "configure the default features".

coryan: https://github.com/googleapis/librarian/pull/2562

src/generated/cloud/compute/v1/.sidekick.toml
```diff
@@ -58,10 +40,13 @@ prefix    = 'compute/v1/'
...
-package-name-override = 'google-cloud-compute-v1'
-per-service-features  = 'true'
-extra-modules         = 'operation'
-'package:lro'         = 'force-used=true,package=google-cloud-lro'
-'package:rpc'         = 'force-used=true,package=google-cloud-rpc'
+version                   = '0.2.0'
+copyright-year            = '2025'
+package-name-override     = 'google-cloud-compute-v1'
+per-service-features      = 'true'
+default-features          = 'instances'
```
dbolduc, line 47: :+1:

PR #3521, opened 2025-10-14
ci(auth): remove iam perm from tf control

dbolduc:
My current theory is that this is what is breaking the GCB connection to the repository.

### DIFF
--- a/src/auth/.gcb/builds/triggers/main.tf
+++ b/src/auth/.gcb/builds/triggers/main.tf
@@ -127,14 +127,6 @@ resource "google_secret_manager_secret_iam_member" "test-api-key-secret-member" 
   member    = "serviceAccount:${data.google_service_account.integration-test-runner.email}"
 }
 
-# The default Cloud Build service account needs permission to act as the
-# integration test runner service account.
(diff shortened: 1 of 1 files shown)


alvarowolfx:
the GCB and GCB Auth integration tests are running now here and on https://github.com/googleapis/google-cloud-rust/pull/3520. It was fixed by https://github.com/googleapis/google-cloud-rust/pull/3519 ? cc @PhongChuong

dbolduc:
> It was fixed by ...

It was "fixed"[^1] by destroying and recreating the resources behind the scenes. And updating the tfstate accordingly with my local changes.

This PR syncs the configuration in our repo with the tfstate.

(Committing code does not run `terraform`... well, until this Sunday when @PhongChuong's scheduler job is put to the test.)

[^1]: I should say "maybe fixed". I claimed success too early yesterday. I will only claim success if it is still working tomorrow.

alvarowolfx:
ack @dbolduc, I wasn't sure if now we had terraform running on a schedule already. 
> 
> (Committing code does not run `terraform`... well, until this Sunday when @PhongChuong's scheduler job is put to the test.)
>

dbolduc:
For the record, this was a good clean up, but it fixed nothing.

PR #3522, opened 2025-10-14
refactor(auth): common function to load SA and API Key secrets

alvarowolfx:


src/auth/integration-tests/src/lib.rs
```diff
@@ -229,25 +176,8 @@ pub async fn impersonated() -> anyhow::Result<()> {
...
-            "projects/{project}/secrets/test-api-key/versions/latest",
-        ))
-        .send()
-        .await?;
-    let api_key = response
-        .payload
-        .expect("missing payload in test-api-key response")
-        .data;
-    let api_key = std::str::from_utf8(&api_key).unwrap();
+    let (_, secret) = get_project_and_api_key().await?;
```
dbolduc: why return the project at all?

src/auth/integration-tests/src/lib.rs
```diff
@@ -229,25 +176,8 @@ pub async fn impersonated() -> anyhow::Result<()> {
...
+    let api_key = std::str::from_utf8(&secret.data).unwrap();
```
dbolduc: if we always return the API key as a string, then should we do that in the helper function?

i.e. `get_api_key() -> Result<String>`

src/auth/integration-tests/src/lib.rs
```diff
@@ -489,6 +419,44 @@ pub async fn workload_identity_provider_programmatic_sourced() -> anyhow::Result
     Ok(())
 }
 
+async fn get_project_and_service_account() -> anyhow::Result<(String, SecretPayload)> {
```
dbolduc: same here. Consider returning `bytes::Bytes` instead of `SecretPayload`. Then we don't need to access `secret.data`...

src/auth/integration-tests/src/lib.rs
```diff
@@ -31,34 +31,16 @@ use iamcredentials::client::IAMCredentials;
...
-        .set_name(format!(
-            "projects/{project}/secrets/test-sa-creds-json/versions/latest"
-        ))
-        .send()
-        .await?;
-    let adc_json = response
-        .payload
-        .expect("missing payload in test-sa-creds-json response")
-        .data;
+    let (project, secret) = get_project_and_service_account().await?;
```
dbolduc: ... and we can give this a more descriptive name again like `(project, adc_json)`.

(I already know the type is a `SecretPayload`, so the name `secret` doesn't tell me much. I think it is more helpful to convey the contents of the secret with the name.)

alvarowolfx:
@dbolduc addressed all comments, all good points and applied all of them.

PR #3524, opened 2025-10-14
refactor(o11y): Switch from feature to cfg for unstable tracing

westarle:
This commit refactors the observability code to use `#[cfg(google_cloud_unstable_tracing)]` instead of the Cargo feature `_unstable-o12y` for gating unstable tracing features.


westarle:
semver-checks fails because we are now compiling out the observability module.

coryan:
> semver-checks fails because we are now compiling out the observability module.

I think this is going to continue even after you merge this PR.  How do we restore the build to green?

PR #3531, opened 2025-10-15
chore(dart): Remove the dart experiment

brianquinlan:
Dart has graduated to having its own repo at:
https://github.com/googleapis/google-cloud-dart

It was great being your guest!


coryan (approved):
It was a pleasure to host this project and working with you folks. Don't be strangers.

Issue #3533, opened 2025-10-15
Flaky test for compute samples


coryan:
I introduced a flake with the compute samples. Need to disable the test and then (maybe)

PR #3534, opened 2025-10-15
test(compute/v1): mitigate impact of flaky test

coryan:
Part of the work for #3533


dbolduc (approved):
comment: I was thinking about gating with a `skipped-integration-test` feature, but I don't think there would be a point to that.

https://github.com/googleapis/google-cloud-rust/blob/00cd7989c4fa73d3d6043dcb3e78cd18c8364981/src/storage/examples/Cargo.toml#L47

PR #3538, opened 2025-10-15
feat(auth): service account sourced id token

alvarowolfx:
Later will be integrated in the ADC flow from #3496 

Towards #3449 https://github.com/googleapis/google-cloud-rust/issues/3589

### DIFF
--- a/src/auth/src/credentials/service_account.rs
+++ b/src/auth/src/credentials/service_account.rs
@@ -502,7 +502,7 @@ pub(crate) mod idtoken {
     #[derive(Debug)]
     struct ServiceAccountTokenProvider {
         service_account_key: ServiceAccountKey,
-        audience: Option<String>,
+        audience: String,
         target_audience: String,
         token_server_url: String,
     }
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/service_account.rs
```diff
@@ -464,19 +464,169 @@ where
...
+    }
+
+    #[async_trait]
+    impl TokenProvider for ServiceAccountTokenProvider {
+        async fn token(&self) -> Result<Token> {
+            let audience = self.audience.clone();
+            let target_audience = Some(self.target_audience.clone());
+            let service_account_key = self.service_account_key.clone();
+            let tg = ServiceAccountTokenGenerator {
+                audience: audience.or(Some(OAUTH2_TOKEN_SERVER_URL.to_string())),
```
sai-sunder-s: audience is already defaulted to the constant in the builder, we should be able to make audience a required field and use it directly here?

sai-sunder-s (dismissed):

src/auth/src/credentials/service_account.rs
alvarowolfx: good catch, pushed a commit fixing this.

PR #3541, opened 2025-10-15
test: add minimal version testing on non-dev code

suzmue:
Our minimal version testing currently uses `test` which brings in dev dependencies. In this case I believe at least async-trait is resolved to different versions: 0.1.13 when using test, 0.1.0 when using check.

We also receive a new deprecation warning due to a transitive dependency (related issue https://github.com/RustCrypto/traits/issues/2036).

Fixes #3541.


src/storage/src/model_ext.rs
```diff
@@ -125,6 +125,7 @@ impl KeyAes256 {
 
 impl std::convert::From<KeyAes256> for crate::model::CommonObjectRequestParams {
     fn from(value: KeyAes256) -> Self {
+        #[allow(deprecated)]
```
coryan: Can you say more words here?  What are we using that is deprecated?

coryan (approved):

Issue #3511, opened 2025-10-13
Storage: read_object fails when object content-encoding is gzip


renato145:
#### Environment details

  - Programming language: rust
  - OS: ubuntu 22.04
  - Language runtime version: 1.90
  - Package version: 1.1.0

#### Steps to reproduce

  1. Following the example code to download an object:
```rust
storage
        .read_object("projects/_/buckets/xxx", "some-file.tar.gz")
        .send()
        .await;
```

I get the following error when the file have "gzip" as the `Content-Encoding` value in its metadata:
```
cannot deserialize the response the response is missing 'content-length', a required header
--------------------
Error {
    kind: Deserialization,
    source: Some(
        MissingHeader(
            "content-length",
        ),
    ),
}
```

When `Content-Encoding` is not present, there is no error. I tested by editing the metadata in https://console.cloud.google.com/storage/browser/

<img width="824" height="541" alt="Image" src="https://github.com/user-attachments/assets/59b34ef6-2c82-449b-b685-8dbb9f3c428f" />

coryan:
This is easy enough to repro (I did in my branch).  The problem is in the code to resume downloads, it assumes the `content-length' header is always there.  I forgot (or neglected to check) that the header is missing when reading gunzipped objects.  It also does not matter, it is not possible to resume uploads during downloads either.

We should ignore the header for gunzipped objects and also stop trying to resume such downloads.

Issue #2051, opened 2025-05-08
Decompressive transcoding


coryan:
Over HTTP+JSON Google Cloud Storage can automatically decompress objects.  We need a way to disable this decompression, to detect that is happening, and then all kinds of things do not work:

- Ranged reads (and therefore resumes) do not work well.
- Checksums do not work.

These things need to be disabled if we detect decompressive transcoding is happening.

PR #3550, opened 2025-10-16
fix(storage): disable default decompression

coryan:
This is the behavior we want by default. We are not breaking anything because automatic decompression (also known as "decompressive transcoding" is currently broken. A future PR will add an option to enable automatic decompression, at which time we will also fix the code to work with it.

Part of the work for #2051 and #3511

### DIFF
--- a/src/integration-tests/src/storage/read_gzip.rs
+++ b/src/integration-tests/src/storage/read_gzip.rs
@@ -71,8 +71,6 @@ pub async fn test(bucket: &storage::model::Bucket) -> anyhow::Result<()> {
         .await?;
     let highlights = response.object();
     tracing::info!("Compressed object read: {:?}", highlights);
-    println!("highlights = {highlights:?}");
-    println!("object = {object:?}");
     assert_eq!(highlights.content_encoding, "gzip", "{highlights:?}");
     assert_eq!(highlights.content_type, "text/plain", "{highlights:?}");
     assert_eq!(highlights.size as usize, compressed.len(), "{highlights:?}");
--- a/src/storage/src/storage/read_object.rs
+++ b/src/storage/src/storage/read_object.rs
@@ -429,6 +429,11 @@ impl Reader {
                 "x-goog-api-client",
                 reqwest::header::HeaderValue::from_static(&self::info::X_GOOG_API_CLIENT_HEADER),
             )
+            // Disable decompressive transcoding: https://cloud.google.com/storage/docs/transcoding
+            //
+            // The default is to decompress objects that have `contentEncoding == "gzip"`. This header
+            // tells Cloud Storage to disable automatic decompression. It has no effect on objects
+            // with a different `contentEncoding` value.
             .header(
                 "accept-encoding",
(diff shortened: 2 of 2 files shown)


src/integration-tests/src/storage/read_gzip.rs
```diff
@@ -0,0 +1,137 @@
...
+    tracing::info!("Reading compressed object");
+    let response = client
+        .read_object(&object.bucket, &object.name)
+        .set_generation(object.generation)
+        .send()
+        .await?;
+    let highlights = response.object();
+    tracing::info!("Compressed object read: {:?}", highlights);
+    println!("highlights = {highlights:?}");
+    println!("object = {object:?}");
```
dbolduc: remove debug prints?

src/storage/src/storage/read_object.rs
```diff
@@ -428,6 +428,10 @@ impl Reader {
             .header(
                 "x-goog-api-client",
                 reqwest::header::HeaderValue::from_static(&self::info::X_GOOG_API_CLIENT_HEADER),
+            )
+            .header(
+                "accept-encoding",
+                reqwest::header::HeaderValue::from_static("gzip"),
```
dbolduc, line 439: just for my understanding: this tells the server we can handle gzip compressed data. Does it change what the server returns to us?

src/storage/src/storage/read_object/resumable.rs
```diff
@@ -149,11 +149,16 @@ struct ReadRange {
...
     match response.status() {
         reqwest::StatusCode::OK => {
-            let header = parse_http_response::required_header(response, "content-length")?;
-            let limit = header
-                .parse::<u64>()
-                .map_err(|e| ReadError::BadHeaderFormat("content-length", e.into()))?;
-            Ok(ReadRange { start: 0, limit })
+            match (
+                read_limit("content-length", response),
+                read_limit("x-goog-stored-content-length", response),
```
dbolduc, line 154: :+1:

dbolduc (approved):

src/storage/src/storage/read_object.rs
dbolduc: Is the default gunzipped?

src/integration-tests/src/storage/read_gzip.rs
coryan: Done.

src/storage/src/storage/read_object.rs
coryan: Yes, the default is `gunzipped`. And this is how one disables that behavior:

https://cloud.google.com/storage/docs/transcoding#:~:text=If%20the%20request%20for%20the%20object%20includes%20an%20Accept%2DEncoding%3A%20gzip%20header%2C%20the%20object%20is%20served%20as%2Dis%20in%20that%20specific%20request%2C%20along%20with%20a%20Content%2DEncoding%3A%20gzip%20response%20header.

Added a comment to explain this.

coryan (commented):
PTAL

PR #3551, opened 2025-10-16
feat(auth): user account sourced id token

alvarowolfx:
Towards #3449 https://github.com/googleapis/google-cloud-rust/issues/3590

### DIFF
--- a/src/auth/src/credentials/user_account.rs
+++ b/src/auth/src/credentials/user_account.rs
@@ -425,7 +425,7 @@ impl TokenProvider for UserTokenProvider {
 
 const MSG: &str = "failed to refresh user access token";
 const MISSING_ID_TOKEN_MSG: &str = "UserCredentials can obtain an id token only when authenticated through \
-gcloud running 'gcloud auth application-default`";
+gcloud running 'gcloud auth application-default login`";
 
 /// Data model for a UserCredentials
 ///
@@ -496,14 +496,14 @@ struct Oauth2RefreshResponse {
 pub(crate) mod idtoken {
     /// Credentials for authenticating with [ID tokens] from a [user account].
     ///
-    /// This module provides a builder for creating [`IDTokenCredentials`] from
+    /// This module provides a builder for [`IDTokenCredentials`] from
     /// authorized user credentials, which are typically obtained by running
     /// `gcloud auth application-default login`.
     ///
-    /// These credentials, which are commonly used for [service to service authentication],
-    /// like when services are hosted in Cloud Run or mediated by Identity-Aware Proxy (IAP).
-    /// Unlike access tokens, ID tokens are not used to authorize access to
-    /// Google Cloud APIs but to verify the identity of a principal.    
+    /// These credentials are commonly used for [service to service authentication].
+    /// For example, when services are hosted in Cloud Run or mediated by Identity-Aware Proxy (IAP).
+    /// ID tokens are only used to verify the identity of a principal. Google Cloud APIs do not use ID tokens
+    /// for authorization, and therefore cannot be used to access Google Cloud APIs.
     ///
     /// [ID tokens]: https://cloud.google.com/docs/authentication/token-types#identity-tokens
     /// [user account]: https://cloud.google.com/docs/authentication#user-accounts
@@ -518,7 +518,6 @@ pub(crate) mod idtoken {
         },
         token::TokenProvider,
     };
-
     use async_trait::async_trait;
     use serde_json::Value;
     use std::sync::Arc;
@@ -541,8 +540,7 @@ pub(crate) mod idtoken {
         }
     }
 
-    /// A builder for constructing [`IDTokenCredentials`] instances that fetch ID tokens using
-    /// user accounts.
+    /// A builder for [`IDTokenCredentials`] instances backed by user account credentials.
     pub struct Builder {
         authorized_user: Value,
         token_uri: Option<String>,
@@ -554,8 +552,6 @@ pub(crate) mod idtoken {
         ///
         /// The `authorized_user` JSON is typically generated when a user
         /// authenticates using the [application-default login] process.
-        ///
-        /// The `target_audience` is not supported for user credentials.
         ///
         /// [application-default login]: https://cloud.google.com/sdk/gcloud/reference/auth/application-default/login
         pub fn new(authorized_user: Value) -> Self {


src/auth/src/credentials/user_account.rs
```diff
@@ -409,6 +424,10 @@ impl TokenProvider for UserTokenProvider {
 }
 
 const MSG: &str = "failed to refresh user access token";
+const MISSING_ID_TOKEN_MSG: &str = "UserCredentials can obtain an id token only when authenticated through \
+gcloud running 'gcloud auth login --update-adc' or 'gcloud auth application-default \
```
sai-sunder-s: `gcloud auth login --update-adc` is not recommended. We should not mention it here

src/auth/src/credentials/user_account.rs
```diff
@@ -409,6 +424,10 @@ impl TokenProvider for UserTokenProvider {
...
+login'. The latter form would not work for Cloud Run, but would still generate an \
```
sai-sunder-s: why does ADC not work for cloud run? I thought both 1P and 3P OAuth clients are special-cased in Cloud Run code

src/auth/src/credentials/user_account.rs
```diff
@@ -473,6 +494,126 @@ struct Oauth2RefreshResponse {
...
+#[allow(dead_code)]
+pub(crate) mod idtoken {
+    /// Credentials for authenticating with [ID tokens] from a [user account].
+    ///
+    /// This module provides a builder for creating [`IDTokenCredentials`] from
+    /// authorized user credentials, which are typically obtained by running
+    /// `gcloud auth application-default login`.
+    ///
+    /// These credentials can be used to authenticate with Google Cloud services
+    /// that require ID tokens for authentication, such as Cloud Run or Cloud Functions.
```
sai-sunder-s: IAP is another popular use case for ID tokens

src/auth/src/credentials/user_account.rs
alvarowolfx: I got this from the Java SDK - https://github.com/googleapis/google-auth-library-java/blob/a9bf3e1e390a629a1e68f07eb4669d99c3769571/oauth2_http/java/com/google/auth/oauth2/UserCredentials.java#L237
will update this msg

src/auth/src/credentials/user_account.rs
alvarowolfx: done

src/auth/src/credentials/user_account.rs
alvarowolfx: removed

alvarowolfx (commented):
updated msgs and docs based on review comments

src/auth/src/credentials/user_account.rs
```diff
@@ -473,6 +492,129 @@ struct Oauth2RefreshResponse {
     refresh_token: Option<String>,
 }
 
+#[allow(dead_code)]
+pub(crate) mod idtoken {
+    /// Credentials for authenticating with [ID tokens] from a [user account].
+    ///
+    /// This module provides a builder for creating [`IDTokenCredentials`] from
```
coryan: Omit superfluous words:

```suggestion
    /// This module provides a builder for [`IDTokenCredentials`] from
```

src/auth/src/credentials/user_account.rs
```diff
@@ -473,6 +492,129 @@ struct Oauth2RefreshResponse {
...
+    /// authorized user credentials, which are typically obtained by running
+    /// `gcloud auth application-default login`.
+    ///
+    /// These credentials, which are commonly used for [service to service authentication],
+    /// like when services are hosted in Cloud Run or mediated by Identity-Aware Proxy (IAP).
```
coryan:
```suggestion
    /// These credentials are commonly used for [service to service authentication].
    /// For example, when services are hosted in Cloud Run or mediated by Identity-Aware Proxy (IAP).
```

src/auth/src/credentials/user_account.rs
```diff
@@ -473,6 +492,129 @@ struct Oauth2RefreshResponse {
...
+    /// Unlike access tokens, ID tokens are not used to authorize access to
+    /// Google Cloud APIs but to verify the identity of a principal.
```
coryan: This is an slightly odd thing to say... access tokens are also used to verify the identity of a principal and *then* (via IAM) to authorize access to Google Cloud APIs.  Maybe:

```suggestion
    /// ID tokens are only used to verify the identity of a principal. Google Cloud APIs do not use ID tokens
    /// for authorization, and therefore cannot be used to access Google Cloud APIs.
```

src/auth/src/credentials/user_account.rs
```diff
@@ -473,6 +492,129 @@ struct Oauth2RefreshResponse {
...
+    use crate::constants::OAUTH2_TOKEN_SERVER_URL;
+    use crate::{
+        BuildResult, Result,
+        credentials::{
+            idtoken::{IDTokenCredentials, dynamic::IDTokenCredentialsProvider},
+            user_account::AuthorizedUser,
+        },
+        token::TokenProvider,
+    };
+
```
coryan: nit: we try to keep all the `use declarations` together so `cargo fmt` can order them for us.

src/auth/src/credentials/user_account.rs
```diff
@@ -425,9 +425,7 @@ impl TokenProvider for UserTokenProvider {
 
 const MSG: &str = "failed to refresh user access token";
 const MISSING_ID_TOKEN_MSG: &str = "UserCredentials can obtain an id token only when authenticated through \
-gcloud running 'gcloud auth login --update-adc' or 'gcloud auth application-default \
-login'. The latter form would not work for Cloud Run, but would still generate an \
-id token.";
+gcloud running 'gcloud auth application-default`";
```
sai-sunder-s: you are missing `login` at the end

src/auth/src/credentials/user_account.rs
```diff
@@ -473,6 +492,129 @@ struct Oauth2RefreshResponse {
...
+    impl<T> IDTokenCredentialsProvider for UserAccountCredentials<T>
+    where
+        T: TokenProvider,
+    {
+        async fn id_token(&self) -> Result<String> {
+            self.token_provider.token().await.map(|token| token.token)
+        }
+    }
+
+    /// A builder for constructing [`IDTokenCredentials`] instances that fetch ID tokens using
```
coryan:
```suggestion
    /// A builder for  [IDTokenCredentials] instances backed by user account credentials.
```

src/auth/src/credentials/user_account.rs
```diff
@@ -473,6 +492,129 @@ struct Oauth2RefreshResponse {
...
+    }
+
+    impl Builder {
+        /// Creates a new builder for `IDTokenCredentials` from a `serde_json::Value`
+        /// representing the authorized user credentials.
+        ///
+        /// The `authorized_user` JSON is typically generated when a user
+        /// authenticates using the [application-default login] process.
+        ///
+        /// The `target_audience` is not supported for user credentials.
```
coryan: Okay, but I assume there are millions of things that are not supported with user credentials?

coryan (approved):
Lots of nits, nothing substantial, so approved.

src/auth/src/credentials/user_account.rs
alvarowolfx: done

src/auth/src/credentials/user_account.rs
alvarowolfx: ugh, nice catch.

src/auth/src/credentials/user_account.rs
alvarowolfx: applied

src/auth/src/credentials/user_account.rs
alvarowolfx: done

src/auth/src/credentials/user_account.rs
alvarowolfx: that's a good point, I added explicitly here because all other ID Token providers require a target audience and most of the explanation around id token mentions that. But might be better to just remove

src/auth/src/credentials/user_account.rs
alvarowolfx: done

alvarowolfx (commented):
pushed more changes based on review comments

PR #3556, opened 2025-10-17
impl(storage): support gunzipped reads

coryan:
If we happen to receive a gunzipped read it will work. A future PR will enable
such reads, this is just adding infrastructure to support them.

### DIFF
--- a/src/storage/src/storage/read_object/gunzipped.rs
+++ b/src/storage/src/storage/read_object/gunzipped.rs
@@ -161,7 +161,7 @@ mod tests {
         let chunk = response.next().await;
         assert!(matches!(&chunk, Some(Err(_))), "{chunk:?}");
         let chunk = response.next().await;
-        assert!(matches!(&chunk, None), "{chunk:?}");
+        assert!(&chunk.is_none(), "{chunk:?}");
         Ok(())
     }
 }


src/storage/src/storage/read_object/gunzipped.rs
```diff
@@ -0,0 +1,167 @@
...
+    }
+
+    async fn next_attempt(&mut self) -> Option<Result<bytes::Bytes>> {
+        let response = self.response.as_mut()?;
+        response.chunk().await.map_err(Error::io).transpose()
+    }
+}
+
+#[async_trait::async_trait]
+impl crate::read_object::dynamic::ReadObjectResponse for GunzippedResponse {
```
dbolduc, line 60: comment: this was unexpectedly convenient.

src/storage/src/storage/read_object/gunzipped.rs
```diff
@@ -0,0 +1,167 @@
...
+use super::{Error, Result};
+use crate::model_ext::ObjectHighlights;
+
+#[derive(Debug)]
+pub struct GunzippedResponse {
+    response: Option<reqwest::Response>,
+    highlights: ObjectHighlights,
+}
+
+impl GunzippedResponse {
```
dbolduc, line 39: comment: these classes could be `NonResumableResponse` and `ResumableResponse` (instead of `GunzippedResponse`, `ReadObjectResponseImpl`).

Fine as is though.

dbolduc (approved):

src/storage/src/storage/read_object/gunzipped.rs
coryan: 🤔  that seems like a good cleanup after this is all working.

coryan:
Related to #2051 and to #3511

Issue #3497, opened 2025-10-10
gax-internal depends on gax 1, needs to depend on >= version of gax it needs to compile


fiadliel:
gax-internal depends on the version of gax defined in the workspace, which is "1". This would be fine, if gax-internal worked with any version of gax from 1.0.0 and upwards, but this is not the case.

Latest gax-internal needs to call `gax::options::internal::get_path_template`, which doesn't exist in gax 1.0.

While this is easy in the small scale to fix (cargo update), semver incompatibilities can build up and make it harder to have the cargo resolver "do the right thing". Also it makes it harder to use tools like dependabot/renovate if they update library-by-library.

Probably the simplest fix (is it the right one?) is to have gax-internal not inherit the workspace gax default, but specify more specifically what it needs.

suzmue:
Thanks for filing this issue.

Our current minimal versions tests are not detecting this problem so a solution for this should also update our CI: https://github.com/googleapis/google-cloud-rust/blob/9267a2825ac9284ee5152773654c57a32a696c8d/.github/workflows/sdk.yaml#L188-L219

coryan:
FWIW, we just released gax-internal v0.7.2 with the right dependency.

suzmue:
Closing, we have other issues tracking the work for improving our processes.

PR #3568, opened 2025-10-17
fix(gax-internal): backport version fixes

coryan:


coryan:
Part of the work for #3497

Issue #3505, opened 2025-10-12
Consider a explicit Operation to Error conversion


coryan:
The GCE status codes (at least for LROs) are very different from standard status codes:

https://cloud.google.com/compute/docs/reference/rest/v1/errors

Instead of automatically converting `Operation` to `Result<Operation>` maybe we should expose an explicit function, something like:

```rust
impl Operation {
  pub fn to_result(self) -> std::result::Result<Operation, OperationError> {
        if self.error.is_none() {
            return Ok(self)  
        }
        // Map the `error` and `error.errors` to an specific error type, preserving all the data.
  }
}
```

So an LRO becomes:

```rust
  let operation = client.insert().set_project(..).set_body(...).poller().until_done().await?;
  // here one can use operation.error.is_some() and get the error details manually, or maybe:
  let operation = operation.to_result()?; // automatically return when using `anyhow::Result` or a similar return type. 
```

coryan:
I think this is the right approach.  There are many errors in a compute long-running operation, the errors cannot be mapped to `google.rpc.Status` or `gax::error::Error` without data loss.  And it appears that more errors could be introduced in the future.

That means we will need to introduce a type that implements `std::error::Error`.  I think it should be something like this:

```rust
use crate::model::{
    InstancesBulkInsertOperationMetadata, SetCommonInstanceMetadataOperationMetadata,
};

#[derive(Clone, Debug)]
#[non_exhaustive]
pub enum OperationError {
    /// A HTTP error with additional details.
    Generic(GenericOperationError),
    /// A partial failure when inserting instances (VMs) in bulk.
    InstancesBulkInsert(InstancesBulkInsertOperationMetadata),
    /// A partial failure when setting the common metadata for all instances.
    SetCommonInstanceMetadata(SetCommonInstanceMetadataOperationMetadata),
}


#[derive(Clone, Debug, Default)]
#[non_exhaustive]
pub struct GenericOperationError {
    pub message: Option<String>,
    pub status_code: Option<i32>,
    pub details: Option<crate::model::operation::Error>,
}
```

PR #3581, opened 2025-10-20
impl(compute/v1): add `Operation::to_result()`

coryan:
We cannot convert `Operation` to `gax::error::Error` without data loss. This introduces a new error type to handle "the operation completed with an error", and a helper function to convert such as operation to errors.

Part of the work for #3505

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -590,7 +590,7 @@ jobs:
           sudo unzip -x /tmp/protoc.zip
           protoc --version
       - name: Regenerate all the code
-        run: go run github.com/googleapis/librarian/cmd/sidekick@v0.4.1-0.20251017142006-14a148302470 refreshall
+        run: go run github.com/googleapis/librarian/cmd/sidekick@v0.4.1-0.20251020204149-2c94a53f0fb5 refreshall
       - run: cargo fmt
         # If there is any difference between the generated code and the
         # committed code that is an error. All the inputs should be pinned,


src/generated/cloud/compute/v1/src/errors.rs
```diff
@@ -0,0 +1,429 @@
...
+    }
+
+    /// Set the [metadata][Self::metadata] field.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_compute_v1::errors::SetCommonInstanceMetadataOperationError;
+    /// use google_cloud_compute_v1::model::{
+    ///     SetCommonInstanceMetadataOperationMetadata, SetCommonInstanceMetadataOperationMetadataPerLocationOperationInfo,
+    ///     Status, set_common_instance_metadata_operation_metadata_per_location_operation_info::State};
```
dbolduc, line 226: these are a mouthful, but nothing we can do about it.

src/generated/cloud/compute/v1/src/errors.rs
```diff
@@ -0,0 +1,429 @@
...
+
+/// Possible errors returned by an operation.
+///
+/// The Compute API long running operations return different errors depending on
+/// the operation being called.
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub enum OperationError {
+    /// A HTTP error with additional details.
+    Generic(GenericOperationError),
```
dbolduc, line 76: so these branches are mutually exclusive in practice?

dbolduc (approved):

src/generated/cloud/compute/v1/src/errors.rs
coryan: AFAICT, yes.  If I am wrong, one can always skip calling `to_result()` and handle errors manually, or we can add a `GenericErrorWithFooBarMetadataDetails()` branch 😵‍💫 🤢

Issue #3567, opened 2025-10-17
Flake integration test for storage


coryan:
Created too many buckets, probably without the right retry policy:

```
2025-10-17T15:10:26.737575Z ERROR ThreadId(126) integration_tests: ERROR the service reports an error with code RESOURCE_EXHAUSTED described as: Create Bucket quota limit limit exceeded on the Google Cloud project. Please slow down execution and try again.
--
  |  
  | Caused by:
  | the service reports an error with code RESOURCE_EXHAUSTED described as: Create Bucket quota limit limit exceeded on the Google Cloud project. Please slow down execution and try again.
  | test driver::run_storage_upload_buffered::default ... FAILED

```

https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/3ee53345-ada3-44fa-b9c1-f901ff3d2a58?project=rust-sdk-testing&e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=monitoring_api_staging

coryan:
We seem to have mitigated this problem. Let's wait 90d and close this issue if there are no more problem

coryan:
Still [flaking](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/e4513f7d-79bd-4421-a6ca-7cfae9d8bb0c?project=rust-sdk-testing&e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=monitoring_api_staging):

```

---- storage::driver::run stdout ----
--
  | cleaning up 0 buckets
[...]

PR #3583, opened 2025-10-20
test(storage): deflake integration tests

coryan:
Create fewer buckets during the integration tests. That reduces the quota usage and (hopefully) deflakes the integration tests.

Refactored the code to make it easier to reuse buckets across tests.

Part of the work for #3567


dbolduc (approved):
Rubber stamp. The +/- line numbers are similar enough. That's as close as I looked. Let me know if you want a better review.

PR #3585, opened 2025-10-20
cleanup(storage): rename response implementations

coryan:
As discussed in https://github.com/googleapis/google-cloud-rust/pull/3556#discussion_r2445199863

### DIFF
--- a/src/storage/src/storage/read_object.rs
+++ b/src/storage/src/storage/read_object.rs
@@ -530,6 +530,18 @@ impl Reader {
     }
 
     fn is_gunzipped(response: &reqwest::Response) -> bool {
+        // Cloud Storage automatically [decompresses gzip-compressed][transcoding]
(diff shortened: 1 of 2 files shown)


src/storage/src/storage/read_object.rs
```diff
@@ -549,11 +549,11 @@ impl Reader {
         let response = self.clone().read().await?;
         if Self::is_gunzipped(&response) {
```
dbolduc, line 562: Consider relocating this comment...

https://github.com/googleapis/google-cloud-rust/blob/ec80604191846637fbbe39c58d07ba78bf46ac2d/src/storage/src/storage/read_object/non_resumable.rs#L15-L27

... inside the `is_gunzipped()` scope.

coryan: Done. PTAL.

PR #3586, opened 2025-10-20
cleanup(lro): simplify error handling

### LINKED ISSUE #3505: Consider a explicit Operation to Error conversion

coryan:
Fixes #3505

### DIFF
--- a/src/lro/src/internal/discovery.rs
+++ b/src/lro/src/internal/discovery.rs
@@ -23,8 +23,7 @@
 //! we can name directly.
 //!
 
-use crate::{Error, Poller, PollingBackoffPolicy, PollingErrorPolicy, PollingResult, Result};
-use gax::error::rpc::Status;
+use crate::{Poller, PollingBackoffPolicy, PollingErrorPolicy, PollingResult, Result};
 use gax::polling_state::PollingState;
 use gax::retry_result::RetryResult;
 use std::sync::Arc;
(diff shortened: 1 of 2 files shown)


src/lro/src/internal/discovery.rs
```diff
@@ -170,7 +162,7 @@ where
 {
     match result {
         Err(ref _e) => (None, PollingResult::Completed(result)),
-        Ok(o) if o.done() => (None, handle_polling_done(o)),
+        Ok(o) if o.done() => (None, PollingResult::Completed(Ok(o))),
```
dbolduc, line 165: I guess we can't return `o.to_result()` instead of `Ok(o)` because there is not a good mapping into a `gax::Error`. is that right?

coryan: Yes. Or at least I cannot find a lossless mapping to `gax::Error`.

PR #3594, opened 2025-10-20
test(bigquery): handle 404 err when cleaning stale datasets

### LINKED ISSUE #3572: Flake BigQuery integration integration test
```

failures:
--
  |  
  | ---- driver::run_bigquery_dataset_service stdout ----
  |  
  | thread 'driver::run_bigquery_dataset_service' panicked at /workspace/src/integration-tests/src/bigquery.rs:117:29:
  | called `Result::unwrap()` on an `Err` value: Error { kind: Service(ServiceDetails { status_code: Some(404), headers: Some({"vary": "X-Origin", "vary": "Referer", "vary": "Origin,Accept-Encoding", "content-type": "application/json; charset=UTF-8", "date": "Fri, 17 Oct 2025 21:16:56 GMT", "server": "ESF", "x-xss-protection": "0", "x-frame-options": "SAMEORIGIN", "x-content-type-options": "nosniff", "accept-ranges": "none", "transfer-encoding": "chunked"}), status: Status { code: NotFound, message: "Not found: Dataset rust-sdk-testing:rust_bq_test_dataset_9HIXkr5D", details: [] } }), source: None }
  | note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

```

alvarowolfx:
When multiple CI steps run in parallel, there is a change of then trying to delete the same dataset (BigQuery metadata service has an eventual consistency model). This PR checks if the error when trying to fetch a Dataset is a 404 and skip it.

Fixes #3572

### DIFF
--- a/src/integration-tests/src/bigquery.rs
+++ b/src/integration-tests/src/bigquery.rs
@@ -116,7 +116,7 @@ async fn cleanup_stale_datasets(
         .filter_map(|r| match r {
             Ok(dataset) => Some(dataset),
             Err(e) if e.status().is_some_and(|s| s.code == Code::NotFound) => None,
-            Err(e) => panic!("expected a successful get_dataset()"),
+            Err(_) => panic!("expected a successful get_dataset()"),
         })
         .filter_map(|dataset| {
             if dataset


src/integration-tests/src/bigquery.rs
```diff
@@ -114,16 +114,26 @@ async fn cleanup_stale_datasets(
...
+                }
+                Err(e) => {
+                    if e.status().is_some_and(|s| s.code == Code::NotFound) {
+                        return None;
+                    }
+                    Err(e)
+                }
             }
-            None
+            .expect("failed to fetch dataset")
```
coryan: This code is a bit ugly.  Maybe it can be simplified by breaking it down into more steps?


```suggestion
            match r {
                Ok(dataset) => Some(Ok(dataset)),
                Err(e) if e.status().is_some_and(|s| s.code == Code::NotFound) => None,
                Err(e) => panic!("expected a successful get_dataset()"),
            })
       .map_filter(|dataset| {
                    if dataset
                        .labels
                        .get(INSTANCE_LABEL)
                        .is_some_and(|v| v == "true")
                        && dataset.creation_time < stale_deadline
                    {
                        return Some(dataset);
                    }
                    None
                })
```

You could break down that `if dataset.labels && dataset.creation` into two `map_filter()` calls too.

alvarowolfx: thanks, code looks much better now. I was also not sure about using `panic!` here or not. But since I added `.except` any way, `panic!` seems to be the way to go and adds up to making it cleaner.

PR #3599, opened 2025-10-21
fix: improve error handling in gRPC client by wrapping transport erro…

atellou:
Improve the error message for transport client creation.

I am having issues in the production environment, and the only message I get for the error is "could not initialize transport client" for the Strage Control client.

### DIFF
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -209,9 +209,9 @@ impl Client {
             })?;
         let endpoint =
             Endpoint::from_shared(endpoint.unwrap_or_else(|| default_endpoint.to_string()))
-                .map_err(BuilderError::transport)?
+                .map_err(|e| BuilderError::transport(e))?
                 .tls_config(ClientTlsConfig::new().with_enabled_roots())
-                .map_err(BuilderError::transport)?
+                .map_err(|e| BuilderError::transport(e))?
                 .origin(origin);
         Ok(InnerClient::new(endpoint.connect_lazy()))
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/grpc.rs
```diff
@@ -209,9 +209,9 @@ impl Client {
             })?;
         let endpoint =
             Endpoint::from_shared(endpoint.unwrap_or_else(|| default_endpoint.to_string()))
-                .map_err(BuilderError::transport)?
+                .map_err(|e| BuilderError::transport(e))?
```
coryan, line 212: How is this any different?  I thought `.map_err(func)` is exactly equivalent to `.map_err(|e| func(e))` ?

coryan:
Separately, the CLA bot wants  email address in the commit message (something with `@fifthwallsolutions.com` in your case) with the address used when you signed the CLA (something with `@gmail.com` in your case, which does not match).

src/gax-internal/src/grpc.rs
atellou: Now that I see you are right, first time seeing this, not too long using Rust, sorry and thank you

coryan: No worries. Do let me know if `format!("{error:?}")` prints enough information or not.  If it does not, we have a bug somewhere.

atellou: It does work I get "Error(Transport(tonic::transport::Error(Transport, NativeCertsNotFound)))."

PR #3603, opened 2025-10-21
test(gax-internal): add http client redirect error tracing test

westarle:
Adds a test to verify span attributes when an HTTP client encounters a redirect loop.

### DIFF
--- a/src/gax-internal/tests/http_client_errors.rs
+++ b/src/gax-internal/tests/http_client_errors.rs
@@ -111,5 +111,69 @@ mod tests {
 
             Ok(())
         }
+
+        #[tokio::test]
+        async fn test_redirect_error_with_tracing_on() -> Result<()> {
+            use google_cloud_gax_internal::observability::attributes::error_type_values::CLIENT_REDIRECT_ERROR;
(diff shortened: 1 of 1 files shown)


src/gax-internal/tests/http_client_errors.rs
```diff
@@ -111,5 +111,69 @@ mod tests {
...
+                span.name, "http_request",
+                "Span name mismatch: {:?}, all attributes: {:?}",
+                span, attributes
+            );
+
+            let expected_error_type = CLIENT_REDIRECT_ERROR.to_string();
+            assert_eq!(
+                attributes.get(semconv::ERROR_TYPE),
+                Some(&expected_error_type),
+                "Span 0: {} mismatch, expected: {:?}, got: {:?}, all attributes: {:?}",
```
dbolduc, line 163: nit: the "expected", "got" should already be printed by `assert_eq!`. I would drop those from the print statement.

dbolduc (approved):

src/gax-internal/tests/http_client_errors.rs
westarle: https://github.com/googleapis/google-cloud-rust/pull/3613

PR #3604, opened 2025-10-21
feat(auth): ADC flow for id token creds

alvarowolfx:
Towards #3449 #3588

### DIFF
--- a/src/auth/src/credentials/idtoken.rs
+++ b/src/auth/src/credentials/idtoken.rs
@@ -289,28 +289,6 @@ pub(crate) mod tests {
 
     #[tokio::test]
     #[parallel]
-    async fn test_build_id_token_credentials_impersonated_service_account_not_supported()
-    -> TestResult {
-        let audience = "test_audience".to_string();
-        let json = serde_json::json!({
-            "type": "impersonated_service_account",
-            "source_credentials": {},
-            "subject_token_type": "urn:ietf:params:oauth:token-type:jwt",
-            "audience": "//iam.googleapis.com/projects/123/locations/global/workloadIdentityPools/my-pool/providers/my-provider",
-            "scope": "https://www.googleapis.com/auth/cloud-platform",
-            "lifetime": "3600s"
-        });
-
-        let result = build_id_token_credentials(audience, Some(json));
-        assert!(result.is_err());
-        let err = result.unwrap_err();
-        assert!(err.is_not_supported());
-        assert!(err.to_string().contains("impersonated_service_account"));
-        Ok(())
-    }
-
-    #[tokio::test]
-    #[parallel]
     async fn test_build_id_token_credentials_external_account_not_supported() -> TestResult {
         let audience = "test_audience".to_string();
         let json = serde_json::json!({


src/auth/src/credentials/idtoken.rs
```diff
@@ -96,6 +98,89 @@ pub(crate) mod dynamic {
     }
 }
 
+/// A builder for constructing [`IDTokenCredentials`] instances that
+/// fetch ID tokens using the loaded credential.
```
coryan: As in previous PRs, use a simpler description.

src/auth/src/credentials/idtoken.rs
```diff
@@ -96,6 +98,89 @@ pub(crate) mod dynamic {
...
+///
+/// This builder loads credentials according to the standard
+/// [Application Default Credentials (ADC)][ADC-link] strategy.
+/// ADC is the recommended approach for most applications and conforms to
+/// [AIP-4110]. If you need to load credentials from a non-standard location
+/// or source, you can use Builders on the specific credential types.
```
coryan:
```suggestion
/// or source, you can the builder for the desired credential type.
```

And we may want to add links to make them more discoverable?  (I imagine the same links would be useful in the OAuth credentials).

src/auth/src/credentials/idtoken.rs
```diff
@@ -96,6 +98,89 @@ pub(crate) mod dynamic {
...
+
+    /// Returns a [IDTokenCredentials] instance with the configured settings.
+    ///
+    /// # Errors
+    ///
+    /// Returns a [BuilderError] if a unsupported credential type is provided
+    /// or if the JSON value is either malformed
+    /// or missing required fields. For more information, on how to generate
+    /// json, consult the relevant section in the [application-default credentials] guide.
+    ///
```
coryan, line 140: Didn't one of the credential types did not support audience?  What does this do when you set an audience and the credential type does not support them?

src/auth/src/build_errors.rs
```diff
@@ -48,6 +48,11 @@ impl Error {
         matches!(self.0, ErrorKind::MissingField(_))
     }
 
+    /// The credential type is not supported for the given use case.
+    pub fn is_not_supported(&self) -> bool {
+        matches!(self.0, ErrorKind::NotSupported(_))
+    }
```
coryan, line 55: Should this be protected by the configuration flag too?  Or are you certain this API will not change or get removed?

If we are ready to add this: this is a new feature, we should bump the minor version of the auth at the same time.

src/auth/src/credentials/service_account.rs
```diff
@@ -466,20 +466,18 @@ where
...
     use crate::constants::{JWT_BEARER_GRANT_TYPE, OAUTH2_TOKEN_SERVER_URL};
     use crate::credentials::idtoken::dynamic::IDTokenCredentialsProvider;
     use crate::credentials::service_account::{ServiceAccountKey, ServiceAccountTokenGenerator};
     use crate::token::{Token, TokenProvider};
     use crate::{BuildResult, credentials::idtoken::IDTokenCredentials};
+    use async_trait::async_trait;
+    use gax::error::CredentialsError;
+    use reqwest::Client;
+    use serde_json::Value;
+    use std::sync::Arc;
```
suzmue, line 483: no action required: maybe we should be standardizing our import grouping throughout our repos (nightly config for https://crates.io/crates/cargo-group-imports)

src/auth/src/credentials/idtoken.rs
```diff
@@ -96,6 +98,89 @@ pub(crate) mod dynamic {
...
+/// or source, you can use Builders on the specific credential types.
+///
+/// [ADC-link]: https://cloud.google.com/docs/authentication/application-default-credentials
+/// [AIP-4110]: https://google.aip.dev/auth/4110
+pub struct Builder {
+    target_audience: String,
+}
+
+impl Builder {
+    /// Creates a new builder where id tokens will be obtained via [application-default login].
```
sai-sunder-s: prefix `application-default login` with `gcloud auth`.

Also call out somewhere that the user account created by that command is not supported by this builder and point to how to do for user account.

src/auth/src/credentials/idtoken.rs
alvarowolfx: `user_account::idtoken` doesn't support audience, but is also not gonna be supported via ADC flow.

src/auth/src/credentials/service_account.rs
alvarowolfx: yeah, I have been bitten a couple of times with imports formattings.

src/auth/src/build_errors.rs
alvarowolfx: good point, I'll put it behind the cfg flag after the PR is merged.

src/auth/src/credentials/idtoken.rs
alvarowolfx: any points on how to do that especifically ? need to use the `--update-adc` flag ?

sai-sunder-s: `gcloud auth application-default login` by default creates user account, which will not be supported here.

But they can provide additional flag [--impersonate-service-account](https://cloud.google.com/sdk/gcloud/reference#--impersonate-service-account) to generated impersonated cred

alvarowolfx: but impersonated credential is not supported yet for id token, which I might need to implement before merging this then 🤔

sai-sunder-s: You could add this doc comment along with impersonated id token change?

alvarowolfx: I'll implement the impersonated cred sourced id token impl separately and update this PR with that. This PR is suppose to be the last step that ties every impl together

coryan (approved):

PR #3607, opened 2025-10-21
refactor: change Token type to Arc<Token> for improved memory management

kasugamirai:


coryan:
Do you mind rebasing? Looks like something changed in the last few hours and the conflict stops the CI from starting.

And could you say something about what "improved memory management" means in this context?  Are you trying to save allocations? Do you have a benchmark you can share?

src/auth/src/token.rs
```diff
@@ -71,12 +72,12 @@ impl std::fmt::Debug for Token {
 
 #[async_trait::async_trait]
 pub(crate) trait TokenProvider: std::fmt::Debug + Send + Sync {
-    async fn token(&self) -> Result<Token>;
+    async fn token(&self) -> Result<Arc<Token>>;
```
coryan, line 75: Do we need to change this?  Seems like any implementation of this method would be called once an hour or so as the `CachedTokenProvider` ... well, caches the value?

src/auth/src/token.rs
```diff
@@ -71,12 +72,12 @@ impl std::fmt::Debug for Token {
...
 }
 
 #[async_trait::async_trait]
 pub(crate) trait CachedTokenProvider: std::fmt::Debug + Send + Sync {
-    async fn token(&self, extensions: Extensions) -> Result<CacheableResource<Token>>;
+    async fn token(&self, extensions: Extensions) -> Result<CacheableResource<Arc<Token>>>;
```
coryan, line 80: Can you add a comment explaining why it makes sense to use `Arc<>` here?  Maybe:


```suggestion
    /// Returns a token or an indication that there are no changes since the last call.
    ///
    /// This function returns `Arc<Token>` to avoid copying the (relatively large) token values. 
    ///
    /// See [CacheableResource] for more information about how to cache values computed
    /// from the `Token`.
    async fn token(&self, extensions: Extensions) -> Result<CacheableResource<Arc<Token>>>;
```

src/auth/src/token.rs
coryan: BTW, I am not sure, honest answer.

src/auth/src/token_cache.rs
```diff
@@ -100,26 +102,27 @@ where
...
-    token_provider: Arc<T>,
-    tx_token: watch::Sender<Option<Result<(Token, EntityTag)>>>,
-) where
+async fn refresh_task<T>(token_provider: Arc<T>, tx_token: watch::Sender<Option<TokenResult>>)
+where
     T: TokenProvider + Send + Sync + 'static,
 {
     loop {
         let token_result = token_provider.token().await;
-        let result = token_result.clone().map(|token| {
```
dbolduc, line 119: so I see we are saving a clone of `token` here...

src/auth/src/token_cache.rs
```diff
@@ -100,26 +102,27 @@ where
...
-            let entity_tag = EntityTag::new();
-            (token, entity_tag)
-        });
+        let result = token_result
+            .as_ref()
+            .map(|token| {
+                let entity_tag = EntityTag::new();
+                (Arc::clone(token), entity_tag)
+            })
+            .map_err(|e| e.clone());
```
dbolduc, line 125: .... but we still might clone an error

src/auth/src/token_cache.rs
```diff
@@ -100,26 +102,27 @@ where
...
 
         let _ = tx_token.send(Some(result));
```
dbolduc, line 128: ... even though the only fields we need in this `match` statement is `token.expires_at`, which is a `Option<std::time::Instant>` which implements `Copy`.

So I think we can do better. Consider:

```rs
    let expiry = token_result.as_ref().ok().map(|t| t.expires_at);
    let tagged = token_result.map(|token| {
      let entity_tag = EntityTag::new();
      (token, entity_tag)
    });

    let _ = tx_token.send(Some(tagged));
    
    match expiry {
        Some(Some(i)) => println!("Expires at: {i:?}"),
        Some(None) => println!("No expiry"),
        None => println!("Error")
    }
```

That doesn't introduce `clone`s or `Arc`s.

src/auth/src/token.rs
kasugamirai: ok

kasugamirai: I’m a little busy today. I’ll do it when I’m free.

dbolduc: > I’m a little busy today. I’ll do it when I’m free.

No worries at all. I sent #3677. I think we can close this PR in favor of that one.

Thanks for noticing and calling out the inefficient clone. We appreciate it!

PR #3616, opened 2025-10-22
fix(storage): `send_buffered()` fails for some object sizes

### LINKED ISSUE #3608: Malformed header in ResumableUpload when ObjectSize is multiple of Buffer size and no size hint is provided
When performing a send_buffered request using a Payload providing no SizeHint, if the object size is a multiple of the buffer size, the request fails with a malformed "Content-Range" header.

#### Environment details

  - Programming language:Rust
  - OS:Linux
  - Language runtime version: 1.90 (2024)
  - Package version: 1.1.0

#### Steps to reproduce
You can find a reproducer code here : https://gist.github.com/dbrasseur-aneo/c2ee7e8ece49dd4fb82fea73a5f9f43c
  1. Create a Storage.write_object request
  2. Set the source as a stream with no size_hint (or a size_hint with no upper)
  3. use the send_buffered function to send the stream to an object
  4. If the object size is a multiple of the buffer size, a request is sent with a malformed "Content-Range" header, leading to the failure

#### Potential location of the issue
After doing some digging, the issue may stem from https://github.com/googleapis/google-cloud-rust/blob/main/src/storage/src/storage/perform_upload/buffered/progress.rs#L162 used here https://github.com/googleapis/google-cloud-rust/blob/main/src/storage/src/storage/perform_upload/buffered.rs#L120

When uploading the first buffer, it gives : "bytes 0-'buffsize-1'/*"
But on the last buffer it gives: "bytes 'objectsize'-'objectsize-1'/'objectsize'" (since n==0), which is invalid.

Thanks!

coryan:
If the upload size is unknown and the stream is closed at a multiple of the buffer size the client library needs to send a final `PUT` request without a payload. The `Content-Range` header was not formatted correctly in this case.

Fixes #3608

### DIFF
--- a/src/integration-tests/src/storage.rs
+++ b/src/integration-tests/src/storage.rs
@@ -253,7 +253,7 @@ pub async fn create_test_hns_bucket() -> Result<(StorageControl, Bucket)> {
 
 async fn client_for_create_bucket() -> Result<StorageControl> {
     let client = StorageControl::builder()
-        // .with_tracing()
+        .with_tracing()
         .with_backoff_policy(
             gax::exponential_backoff::ExponentialBackoffBuilder::new()
                 .with_initial_delay(Duration::from_secs(2))
--- a/src/integration-tests/src/storage/write_object.rs
+++ b/src/integration-tests/src/storage/write_object.rs
@@ -16,18 +16,6 @@ use crate::Result;
 use storage::client::Storage;
 use storage::streaming_source::{Seek, SizeHint, StreamingSource};
 
-pub async fn repro(
-    builder: storage::builder::storage::ClientBuilder,
-    bucket_name: &str,
-) -> Result<()> {
-    // Run all the upload tests in parallel, using the same bucket.
-    // Creating a new bucket is rate-limited, and slow. Creating objects
-    // is relatively cheap.
-    let client = builder.build().await?;
-    repro_3608_buffered(&client, bucket_name).await?;
-    Ok(())
-}
-
 pub async fn run(
     builder: storage::builder::storage::ClientBuilder,
     bucket_name: &str,
--- a/src/integration-tests/tests/driver.rs
+++ b/src/integration-tests/tests/driver.rs
@@ -227,20 +227,6 @@ mod driver {
 
     #[test_case(Storage::builder(); "default")]
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
-    async fn run_storage_repro(
-        builder: storage::builder::storage::ClientBuilder,
-    ) -> integration_tests::Result<()> {
-        let _guard = integration_tests::enable_tracing();
-        let (control, bucket) = integration_tests::storage::create_test_bucket().await?;
-        let result = integration_tests::storage::write_object::repro(builder, &bucket.name).await;
-        if let Err(e) = storage_samples::cleanup_bucket(control, bucket.name.clone()).await {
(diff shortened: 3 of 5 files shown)


src/integration-tests/src/storage.rs
```diff
@@ -253,7 +253,7 @@ pub async fn create_test_hns_bucket() -> Result<(StorageControl, Bucket)> {
 
 async fn client_for_create_bucket() -> Result<StorageControl> {
     let client = StorageControl::builder()
-        .with_tracing()
+        // .with_tracing()
```
dbolduc: delete or uncomment

src/integration-tests/src/storage/write_object.rs
```diff
@@ -16,6 +16,18 @@ use crate::Result;
...
+    bucket_name: &str,
+) -> Result<()> {
+    // Run all the upload tests in parallel, using the same bucket.
+    // Creating a new bucket is rate-limited, and slow. Creating objects
+    // is relatively cheap.
+    let client = builder.build().await?;
+    repro_3608_buffered(&client, bucket_name).await?;
+    Ok(())
+}
+
```
dbolduc: delete, already handled in `run`

src/integration-tests/tests/driver.rs
```diff
@@ -225,6 +225,20 @@ mod driver {
         result
     }
 
+    #[test_case(Storage::builder(); "default")]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn run_storage_repro(
```
dbolduc: same, revert because already covered in `run_storage_write_object` ?

src/storage/src/storage/perform_upload/buffered.rs
```diff
@@ -118,6 +118,7 @@ where
         progress: &mut InProgressUpload,
     ) -> Result<reqwest::RequestBuilder> {
         let range = progress.range_header();
+        tracing::info!("  ***** range={range} progress={progress:?}");
```
dbolduc: revert debug logging

src/storage/src/storage/perform_upload/buffered/progress.rs
```diff
@@ -164,6 +164,7 @@ impl InProgressUpload {
             (0, 0, Some(len)) => format!("bytes */{len}"),
             (0, 0, None) => "bytes */0".to_string(),
```
dbolduc: nit: seems like we can remove this for the general `(0, o, None)` case.

dbolduc (commented):
The fix is good, but the PR needs some clean ups.

src/storage/src/storage/perform_upload/buffered/progress.rs
coryan: Done.

src/integration-tests/src/storage.rs
coryan: Uncommented.

src/integration-tests/src/storage/write_object.rs
coryan: Removed.

src/integration-tests/tests/driver.rs
coryan: Removed.

src/storage/src/storage/perform_upload/buffered.rs
coryan: removed.

Issue #3625, opened 2025-10-24
ci: periodically run terraform to update gcp resources


PhongChuong:
Use cloud schedular and gcb to periodically run terraform to update cloud-sdk-rust resouces.

PR #3633, opened 2025-10-24
ci: add cloud scheduler to refresh  terraform once a week

PhongChuong:
Fix: #3625


.gcb/terraform.yaml
```diff
@@ -0,0 +1,34 @@
...
+serviceAccount: 'projects/${PROJECT_ID}/serviceAccounts/terraform-runner@${PROJECT_ID}.iam.gserviceaccount.com'
+steps:
+  - name: 'ubuntu:latest'
+    script: |
+      #!/usr/bin/env bash
+      set -e
+      apt-get update && apt-get install -y gnupg software-properties-common wget
+      wget -O- https://apt.releases.hashicorp.com/gpg | gpg --dearmor | tee /usr/share/keyrings/hashicorp-archive-keyring.gpg > /dev/null
+      gpg --no-default-keyring --keyring /usr/share/keyrings/hashicorp-archive-keyring.gpg --fingerprint
+      echo "deb [arch=$(dpkg --print-architecture) signed-by=/usr/share/keyrings/hashicorp-archive-keyring.gpg] https://apt.releases.hashicorp.com $(grep -oP '(?<=UBUNTU_CODENAME=).*' /etc/os-release || lsb_release -cs) main" | tee /etc/apt/sources.list.d/hashicorp.list
```
coryan: I still don't understand why we use this image instead of one of the images with terraform pre-installed.

Is there a particular reason or is it that it is just convenient?

.gcb/terraform.yaml
```diff
@@ -0,0 +1,34 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+options:
+  dynamic_substitutions: true
+  substitutionOption: 'ALLOW_LOOSE'
+  logging: CLOUD_LOGGING_ONLY
+serviceAccount: 'projects/${PROJECT_ID}/serviceAccounts/terraform-runner@${PROJECT_ID}.iam.gserviceaccount.com'
+steps:
+  - name: 'ubuntu:latest'
```
coryan: Separately: `ubuntu:latest` is prone to breakage whenever a new version of Ubuntu is released. Consider pinning to the latest LTS (`ubuntu:24.04` as I write this).

.gcb/terraform.yaml
PhongChuong: It is  just for convenient.
I couldn't find an image with terraform preinstalled along with shell* so the alternative is to split the init/plan/apply into multiple gcb steps.
Or we can build our own docker image with terraform preinstalled.

[*] hashicorp/terraform does not come with shell.

.gcb/terraform.yaml
PhongChuong: Thanks. Updated.

.gcb/terraform.yaml
coryan: > I couldn't find an image with terraform preinstalled along with shell* 

I think you just need to override `--entrypoint`:

```bash
docker run -it --rm --entrypoint /bin/sh hashicorp/terraform:latest 
/ # ls
bin    dev    etc    home   lib    media  mnt    opt    proc   root   run    sbin   srv    sys    tmp    usr    var
```

> alternative is to split the init/plan/apply into multiple gcb steps.

That works too, you may need to use the volume feature from GCB to pass data between the steps.

dbolduc (approved):

.gcb/terraform.yaml
PhongChuong: Thanks. Using entry point worked.
I've updated the image to hashicorp/terraform.
PTAL.

PhongChuong:
@dbolduc , I've made some changes to the terraform.yaml. PTAL.

.gcb/scripts/terraform.sh
```diff
@@ -0,0 +1,39 @@
...
+      exit 1
+      ;;
+  esac
+done
+
+echo "Running terraform on ${DIR}"
+cd ${DIR}
+terraform init
+terraform plan -out /tmp/bootstrap.tplan
+terraform apply /tmp/bootstrap.tplan
```
dbolduc: nit: newline at end of file?

Do we need to start running `shfmt` in our `typos` build?

.gcb/scripts/terraform.sh
```diff
@@ -0,0 +1,39 @@
...
+      ;;
+    ?)
+      echo "Invalid option: -${OPTARG}."
+      exit 1
+      ;;
+  esac
+done
+
+echo "Running terraform on ${DIR}"
+cd ${DIR}
```
dbolduc: Don't love that if this is not given a `-d` argument, we will navigate to the home directory and continue executing.

Consider exiting if no `-d`. It might be cleanest to expect an extra arg which is the directory.

.gcb/terraform.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+options:
+  dynamic_substitutions: true
+  substitutionOption: 'ALLOW_LOOSE'
+  logging: CLOUD_LOGGING_ONLY
+serviceAccount: 'projects/${PROJECT_ID}/serviceAccounts/terraform-runner@${PROJECT_ID}.iam.gserviceaccount.com'
+steps:
+  - name: 'hashicorp/terraform:latest'
+    entrypoint: '/bin/sh'
+    args: ['/workspace/.gcb/scripts/terraform.sh', '-d', '/workspace/.gcb/builds']
```
coryan: You may be able to use [dir](https://cloud.google.com/build/docs/build-config-file-schema#dir) option.

coryan: maybe:


```suggestion
    entrypoint: '/bin/sh'
    dir: '/workspace/.gcb/builds'
    args: ['-c', 'terraform init && terraform plan -out /tmp/bootstrap.tplan && terraform apply /tmp/bootstrap.tplan']
```

But you should test that idea, I am not 100% sure it works.

PhongChuong: This works. Thanks.

PhongChuong: This works. Thanks.

.gcb/scripts/terraform.sh
PhongChuong: Switched to use dir and inlined arg so the shell script is no longer needed.

Issue #3015, opened 2025-08-24
Implement the publisher client


coryan:
The list of requirements will come out of the design, put probably batching (time and size based), retries, and support for exactly once semantics will be needed.

PR #3634, opened 2025-10-25
feat(pubsub)!: update API for interacting with the publisher

suzmue:
This updates the API to be closer to our desired surface. The `publish` call has been updated to accept a message and return a Future that will resolve to the message id.

The `PublisherClient` is also now responsible for creating `Publisher`s for specific topics.

For #3015.

### DIFF
--- a/src/pubsub/src/publisher/client.rs
+++ b/src/pubsub/src/publisher/client.rs
@@ -91,12 +91,15 @@ impl PublisherClient {
     /// let message_id = publisher.publish(PubsubMessage::new().set_data("Hello, World")).await?;
     /// # Ok(()) }
     /// ```
-    pub fn publisher(&self, topic: String) -> Publisher {
+    pub fn publisher<T>(&self, topic: T) -> Publisher
+    where
+        T: Into<String>,
+    {
         // In order to configure the Publisher, we will probably want
         // to return a builder.
         Publisher {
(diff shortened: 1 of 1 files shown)


src/pubsub/src/publisher/client.rs
```diff
@@ -52,44 +55,121 @@ pub(crate) mod client_builder {
...
+    topic: String,
+}
+
+impl Publisher {
+    /// Publishes a message to the topic.
+    ///
+    /// ```
+    /// # use google_cloud_pubsub::client::Publisher;
+    /// # async fn sample(publisher: Publisher) -> anyhow::Result<()> {
+    /// # use google_cloud_pubsub::model::PubsubMessage;
```
coryan, line 133: nit: for a future PR, google-cloud_**pubsub**::model::**Pubsub**Message is a bit repetitive. But I guess we inherited the name? Is there any reason to hide that type into a hand-crafted type?

suzmue: We could always rename to be less repetitive. 

Looking at the [type definition](https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#pubsubmessage), it seems like a hand-crafted type _could_ make it easier to construct an appropriate messages since there are some rules that we could direct users to do correctly (making sure to have data or >=1 attribute for example).

Issue #3362, opened 2025-09-19
Add simpler credentials header accessor


coryan:
The current API is correct, and powerful, but a bit hard to use.  We should offer a simpler version...

> I've been using, but been quite nervous about, using `unreachable!()` for this code path. Could you also add some explicit clarification/guarantees on under what conditions a `NotModified` won't be returned?
> 
> Yes.  In addition, I think we should provide an extension trait that simplifies the API and makes this clear via the type system. Something like:
> 
> ```rust
> pub trait CredentialsExt {
>     async fn cached_headers(&self) -> Result<HeaderMap> { .... }
> ```
> 
>  
> (and I'm also very very happy this exists, to replace the hacky code we had to use before now - also appreciate the huge amount of work that's obviously gone into it)
> 
> Thanks for saying that.
>  

 _Originally posted by @coryan in [#3361](https://github.com/googleapis/google-cloud-rust/issues/3361#issuecomment-3312698375)_

alvarowolfx:
related to #2929 and potentially duplication.

alvarowolfx:
solved by #3642

Issue #3361, opened 2025-09-19
How should google-cloud-auth CacheableResource<> be used?


sxlijin:
Excited to have an official Rust SDK, so thanks!

I'm trying to figure out how to use the CacheableResouce<> type in google-cloud-auth and have no idea how I'm supposed to use the returned `EntityTag` to cache the returned auth headers. Is it deliberate that there's no user-visible timeout on the auth headers? Or is there an API that I'm supposed to be able to use the `EntityTag` with to determine if my headers are still sufficiently recent (I don't see anything on CredentialsProvider that I can use)?

Links:
- https://docs.rs/google-cloud-auth/1.0.0/google_cloud_auth/credentials/struct.Credentials.html
- https://docs.rs/google-cloud-auth/1.0.0/google_cloud_auth/credentials/trait.CredentialsProvider.html
- https://docs.rs/google-cloud-auth/1.0.0/google_cloud_auth/credentials/enum.CacheableResource.html

coryan:
> Excited to have an official Rust SDK, so thanks!

We are too.

> 
> I'm trying to figure out how to use the `CacheableResource<>` type in google-cloud-auth and have no idea how I'm supposed to use the returned `EntityTag` to cache the returned auth headers.

Yes, we are missing some user-guides / tutorials for auth.

> Is it deliberate that there's no user-visible timeout on the auth headers?

Yes.

> Or is there an API that I'm supposed to be able to use the `EntityTag` with to determine if my headers are still sufficiently recent (I don't see anything on CredentialsProvider that I can use)?

The credentials will automatically refresh any token in the background.  And they automatically cache the headers until the token is about to expire. If you are going to use the headers as they are, without any additional transformations, then you can ignore the `EntityTag` and pass the headers to your HTTP or gRPC client.

```rust
struct MyClient {
   creds: google_cloud_auth::credentials::Credentials,
   inner: ... some client that uses http::Headers directly ...
}

impl MyClient {
    pub async fn thing(&self) -> Result<Thing> {
       let headers = match self.creds.headers(http::Extensions())? {
           CacheableResource::NotModified => unreachable!(),
           CacheableResource::New { entity_tag, data } => data,
       };
       self.inner.do_some_call_using_headers(headers)
   }
```

If you perform some additional computations based on the headers, then the entity tag can help you avoid the recomputation.  The pseudo-code would be something like:

```rust
use google_cloud_auth::credentials::CacheableResource;

struct MyClient {
[...]

PR #3642, opened 2025-10-29
feat(auth): allow consuming access tokens directly

### LINKED ISSUE #2929: Add "access token" method to credentials
We have at couple of requests to fetch the access token. A few observations:

- The token should include the string value and the expiration time.
- The token should be wrapped in a `CacheableResource<>`.
- We should return an error for API keys.

----

> * we need direct access to an access token to integrate with existing libraries like [`object_store`](https://docs.rs/object_store/latest/object_store/). `object_store` has pretty rudimentary support for how it does gcloud auth, we were aiming to use your first party auth library to make it more complete/customized for our specific environments.
> 
> @sai-sunder-s now that we have a clear need, we should restore the second method in `Credentials` to fetch the token. It would need to be something like:
> 
> ```rust
> impl Credentials {
>   async pub fn access_token(&self, ...) -> Result<CacheableResource<AccessToken>>;
> ```
> 
> because it can fail to fetch the token (or the credential type may not support tokens at all) and you may want to cache a derived value and avoid recomputation.  And `AccessToken` would be something like:
> 
> ```rust
> struct AccessToken { /* all private */}
> impl AccessToken {
>   pub fn value(&self) -> String;
>   pub fn type(&self) -> &'static str; // typically `Bearer`
>   pub fn expiration_time(&self) -> std::time::Instant /* ??? SystemTime ??? */
> }
> ```
> 
> We also need to consider how to create tokens for mocking.
> 
> 
>  

 _Originally posted by @coryan in [#2879](https://github.com/googleapis/google-cloud-rust/issues/2879#issuecomment-3190465594)_

alvarowolfx:
Fixes #2929 and #3362

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -768,6 +768,7 @@ pub(crate) mod tests {
     use std::sync::LazyLock;
     use test_case::test_case;
     use tokio::time::Duration;
+    use tokio::time::Instant;
 
     pub(crate) fn find_source_error<'a, T: Error + 'static>(
         error: &'a (dyn Error + 'static),


alvarowolfx:
This is a WIP and is missing all documentation for the new interface.

Pros:
* `AccessTokenCredentials` is compatible with `Credentials` type (via automatic conversion to `CredentialProvider` -> `Credentials`)
* Reuse all current `Builder`s and can selective pick which `Credentials` types would support that (we don't want API Key creds to support this)
Cons:
* I don't like the `build_access_token_credentials()` method name. Too long and not sure if is a good pattern to add another final branch to a Builder. 
* Introduces new AccessTokenCredentials which is *almost* the same as `Credentials`. Perhaps it can be simplified ? 
* `AccessTokenCredentialsProvider` looks really similar to `TokenProvider`, can this be reused ? 
   *  Probably not, Token is private and we need a new AccessToken type to expose that.

coryan:
I don't know what is the exact level of test coverage we want, but 46% is too low.

alvarowolfx:
> I don't know what is the exact level of test coverage we want, but 46% is too low.

this is also missing tests, is just a draft for the public API.

src/auth/src/credentials.rs
```diff
@@ -137,6 +139,78 @@ impl Credentials {
...
+            CacheableResource::New { data, .. } => Ok(data.into()),
+            CacheableResource::NotModified => Err(errors::CredentialsError::from_msg(
+                false,
+                "Expecting token to be present",
+            )),
+        }
+    }
+}
+
+impl std::convert::From<Token> for AccessToken {
```
dbolduc, line 216: Q: why do we have `Token` at all?

The answer might be caching.

Are we using the `expires_at` to determine when a token goes stale? If so, `trait ExpiringThing` with impls for `AccessToken` / `IdToken` might be cleaner.

I know this is all internal, so how we organize things is not a big deal. I am mostly just curious what the plan is.

tests/integration-auth/src/lib.rs
```diff
@@ -81,7 +82,8 @@ pub async fn service_account_with_audience() -> anyhow::Result<()> {
                 "https://secretmanager.googleapis.com/",
             ),
         )
-        .build()?;
+        //TODO: revert - this is just to show that AccessTokenCredentials is compatible with Credentials
+        .build_access_token_credentials()?;
```
dbolduc, line 84: FWIW, I think `build_foo()` is not that crazy.

For example, for 99% of our builders, we have a `send()`, but in storage, our `builder::WriteObject` has `send_buffered()` and `send_unbuffered()`.

Q: do we recommend applications use `build()` or `build_access_token_credentials()` ?

src/auth/src/credentials.rs
```diff
@@ -137,6 +139,78 @@ impl Credentials {
...
+    }
+}
+
+impl AccessTokenCredentials {
+    pub async fn token(&self) -> Result<AccessToken> {
+        self.inner.token().await
+    }
+}
+
+impl CredentialsProvider for AccessTokenCredentials {
```
dbolduc, line 179: comment: I see this is how `AccessTokenCredentials` becomes compatible with `Credentials`. LGTM.

src/auth/src/credentials.rs
```diff
@@ -137,6 +139,78 @@ impl Credentials {
...
+    async fn headers(&self, extensions: Extensions) -> Result<CacheableResource<HeaderMap>> {
+        self.inner.headers(extensions).await
+    }
+
+    async fn universe_domain(&self) -> Option<String> {
+        self.inner.universe_domain().await
+    }
+}
+
+pub struct AccessToken {
```
dbolduc, line 191: nit: documentation for public types, but I think you are just sketching the interfaces for now.

dbolduc (commented):
FWIW, this approach makes a lot of sense to me. It is lamentable that we are already making backwards compatibility sacrifices in version `1.1.0` of a library, but that is where we are.

I think I messed up by making `Credentials` an opaque object instead of a trait. I think at the time I was afraid of all the generics that entailed... but now I realize that is just what flexible Rust looks like.

src/auth/src/credentials.rs
alvarowolfx: Those two `impl` From trait related to AccessToken from Token and CacheableResource are there to simplify `impl AccessTokenCredentialsProvider`, so all conversion from `Token` to `AccessToken` is handled in one place:
* CacheableResource<Token> -> Token.into() -> AccessToken: https://github.com/googleapis/google-cloud-rust/pull/3642/files#diff-d02d1aa503dc8c573dee51988fb8e5156b401a6383c717a33ebd172870013f92R188
* Token.into() -> AccessToken: https://github.com/googleapis/google-cloud-rust/pull/3642/files#diff-d02d1aa503dc8c573dee51988fb8e5156b401a6383c717a33ebd172870013f92R200

Final code in the Credential types are all the same because of that:

* SA impl - https://github.com/googleapis/google-cloud-rust/pull/3642/files#diff-0bbe1ea5b1bfeb349c2c9771de22722e41b746a97a0fabaf72b7e77cf6b2f8dcR474
* User Account impl - https://github.com/googleapis/google-cloud-rust/pull/3642/files#diff-0bbe1ea5b1bfeb349c2c9771de22722e41b746a97a0fabaf72b7e77cf6b2f8dcR470

tests/integration-auth/src/lib.rs
alvarowolfx: I think they should still use `build()` by default.

coryan:
Marking as draft. If / when it is ready for review please let us know.

alvarowolfx:
> Marking as draft. If / when it is ready for review please let us know.

@coryan it was ready to review, but I think I'll split into smaller PRs to make it easier to review. Unless you think this is in a reasonable size.

coryan:
@dbolduc can you take another look?

src/auth/src/credentials.rs
```diff
@@ -404,6 +526,34 @@ impl Builder {
     ///
     /// [application-default credentials]: https://cloud.google.com/docs/authentication/application-default-credentials
     pub fn build(self) -> BuildResult<Credentials> {
+        Ok(self.build_access_token_credentials()?.into())
+    }
+
+    /// Returns a [Credentials] instance with the configured settings.
```
dbolduc: nit: s/Credentials/AccessTokenCredentials/

src/auth/src/credentials.rs
```diff
@@ -404,6 +526,34 @@ impl Builder {
...
+    ///     .build_access_token_credentials()?;
+    /// let access_token = credentials.access_token().await?;
+    /// println!("Token: {}", access_token.token);
+    /// # Ok::<(), anyhow::Error>(())
+    /// # });
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// Returns a [CredentialsError] if a unsupported credential type is provided
```
dbolduc: nit: s/a unsupported/an unsupported/

src/auth/src/credentials.rs
```diff
@@ -404,6 +526,34 @@ impl Builder {
...
+    /// or if the JSON value is either malformed
+    /// or missing required fields. For more information, on how to generate
+    /// json, consult the relevant section in the [application-default credentials] guide.
```
dbolduc: nit: s/json/the JSON for a credential/ ?

src/auth/src/credentials/external_account.rs
```diff
@@ -643,6 +645,22 @@ impl Builder {
     ///
     /// [external_account_credentials]: https://google.aip.dev/auth/4117#configuration-file-generation-and-usage
     pub fn build(self) -> BuildResult<Credentials> {
+        Ok(self.build_access_token_credentials()?.into())
+    }
+
+    /// Returns a [AccessTokenCredentials] instance with the configured settings.
```
dbolduc: nit: s/Returns a/Returns an/

src/auth/src/credentials/external_account.rs
```diff
@@ -643,6 +645,22 @@ impl Builder {
...
+    ///
+    /// # Errors
+    ///
+    /// Returns a [BuilderError] if the `external_account_config`
+    /// provided to [`Builder::new`] cannot be successfully deserialized into the
+    /// expected format for an external account configuration. This typically happens if the
+    /// JSON value is malformed or missing required fields. For more information,
+    /// on the expected format, consult the relevant section in the
+    /// [external_account_credentials] guide.
```
dbolduc: nit: reflow?

we do not have any sort of tool. I just manully cut lines off at or before 80 characters. :shrug: 

Also consider a paragraph break before "For more information..."

src/auth/src/credentials.rs
```diff
@@ -404,6 +526,34 @@ impl Builder {
...
+    /// println!("Token: {}", access_token.token);
+    /// # Ok::<(), anyhow::Error>(())
+    /// # });
+    /// ```
+    ///
+    /// # Errors
+    ///
+    /// Returns a [CredentialsError] if a unsupported credential type is provided
+    /// or if the JSON value is either malformed
+    /// or missing required fields. For more information, on how to generate
```
dbolduc: optional nit: paragraph break before "For more information..."

src/auth/src/credentials/external_account.rs
```diff
@@ -1261,6 +1281,17 @@ where
     }
 }
 
+#[async_trait::async_trait]
+impl<T> AccessTokenCredentialsProvider for ExternalAccountCredentials<T>
+where
+    T: CachedTokenProvider,
+{
+    async fn access_token(&self) -> Result<AccessToken> {
+        let token = self.token_provider.token(Extensions::new()).await?;
```
dbolduc, line 1292: observation, for all cred types: we are deciding not to play the `CacheableResource` game for `AccessToken`s? Why is that?

src/auth/src/credentials.rs
```diff
@@ -137,6 +139,108 @@ impl Credentials {
...
+/// A trait for credential types that can provide direct access to an access token.
+///
+/// This trait is primarily intended for interoperability with other libraries that
+/// require a raw access token, or for calling Google Cloud APIs that are not yet
+/// supported by the SDK.
+pub trait AccessTokenCredentialsProvider: CredentialsProvider + std::fmt::Debug {
+    /// Asynchronously retrieves an access token.
+    fn access_token(&self) -> impl Future<Output = Result<AccessToken>> + Send;
+}
+
```
dbolduc, line 231: hmmm... I was questioning whether we need this `pub trait`.

We needed a public trait for `Credentials` mainly because `gax` unit tests needed a way to mock `Credentials`.

I guess I buy that apps built on top of these `AccessTokenCredentials` need something similar for their unit tests.

If so, we should add a quick mocking test to appease the coverage bot. Something like: https://github.com/googleapis/google-cloud-rust/blob/6709447d1934c9dad742f065cb9bca8b46bf7194/src/auth/tests/credentials.rs#L441-L465

dbolduc: feel free to defer to a follow up PR

src/auth/src/credentials/external_account.rs
alvarowolfx: the CacheableResource is the biggest source of friction for customers, since you have to add some handling for the `NotModified` enum that usually leads to some extra and unreachable code. This was mentioned in the issue https://github.com/googleapis/google-cloud-rust/issues/3361

src/auth/src/credentials.rs
alvarowolfx: that was a good point and I found out that the new trait was not mock friendly, since `AccessToken` had some private fields, so I decided to remove them for now.

alvarowolfx:
> The code looks really good, and I would symbolically approve the PR. I think my comments are mostly grammar nits.
> 
> I need an explicit answer on why we are abandoning `Extensions` / `CacheableResource` for the `AccessToken` before I can approve though.

pushed some updates addressing the nits in the docs and also made some improvements so customers can mock the new trait

src/auth/src/credentials/external_account.rs
dbolduc: Ack, and I think you are making the right call.

I think we could always add a `fn access_token_if_modified` later, and provide a default implementation in terms of the existing `access_token` (so we don't break anyone who has their own trait impls).

PR #3643, opened 2025-10-29
test(gax-internal): parameterize HTTP error/success tests

westarle:
This unit-tests more cases within the core HTTP tracing implementation.

### DIFF
--- a/src/gax-internal/src/observability/http_tracing.rs
+++ b/src/gax-internal/src/observability/http_tracing.rs
@@ -133,8 +133,9 @@ mod tests {
     use google_cloud_test_utils::test_layer::{AttributeValue, TestLayer};
     use http::Method;
     use opentelemetry_semantic_conventions::{attribute as otel_attr, trace as otel_trace};
-    use reqwest;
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -329,10 +351,16 @@ mod tests {
...
+            (
+                otel_trace::HTTP_RESPONSE_STATUS_CODE,
+                (status_code.as_u16() as i64).into(),
+            ),
+            (otel_trace::ERROR_TYPE, status_code.as_str().into()),
+            (
+                otel_attr::RPC_GRPC_STATUS_CODE,
+                (expected_grpc_code as i64).into(),
+            ),
+            (KEY_GRPC_STATUS, expected_grpc_code.to_string().into()),
```
westarle: I'm going to fix this test so it doesn't just agree with the implementation (i.e. if Code::to_string() changes, the test should fail.)

coryan (approved):

PR #3647, opened 2025-10-29
test(gax-internal): Add http_observability error tests

westarle:


src/gax-internal/tests/http_observability.rs
```diff
@@ -146,4 +144,77 @@ mod tests {
...
+            "otel.status mismatch, attrs: {:?}",
+            attrs
+        );
+
+        assert_eq!(
+            attrs.get(otel_attr::RPC_GRPC_STATUS_CODE),
+            Some(&expected_grpc_code.into()),
+            "rpc.grpc.status_code mismatch, attrs: {:?}",
+            attrs
+        );
```
coryan, line 211: I missed this in previous reviews: why do we always have a RPC_GRPC_STATUS_CODE?  What if the client is Compute or Storage where gRPC plays no role?

westarle: No reason, I had some wires crossed between this and our metrics requirement to standardize on gRPC code.

#3649 brings this up to speed with the current PRD.

PR #3648, opened 2025-10-30
feat(pubsub): add topic publisher builder and batching options

PhongChuong:
Adding a publisher builder allows users to fine grain configure options on each publisher. This is in contrast to the client builder which shares options across multiple publishers.

This PR includes a batching options for publisher.

Toward: #3015

### DIFF
--- a/src/pubsub/src/publisher/publisher.rs
+++ b/src/pubsub/src/publisher/publisher.rs
@@ -147,7 +147,19 @@ mod tests {
         let publisher = client
             .publisher("projects/my-project/topics/my-topic".to_string())
             .build();
-        assert_ne!(publisher.batching_options.message_count_threshold, 1_u32);
+
+        assert_eq!(
+            publisher.batching_options.message_count_threshold,
+            BatchingOptions::default().message_count_threshold
+        );
+        assert_eq!(
+            publisher.batching_options.byte_threshold,
+            BatchingOptions::default().byte_threshold
+        );
+        assert_eq!(
+            publisher.batching_options.delay_threshold,
+            BatchingOptions::default().delay_threshold
+        );
         Ok(())
     }
 }


src/pubsub/src/publisher/client.rs
```diff
@@ -159,6 +157,60 @@ impl Publisher {
...
+            inner: self.inner,
+            topic: self.topic,
+            batching_settings: self.batching_settings,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub struct BatchingSettings {
```
dbolduc: public types need documentation

also, I think we have been calling things more `FooOptions`. I don't think we have any `FooSettings`.

does this belong in some sort of `options` mod?

src/pubsub/src/publisher/client.rs
```diff
@@ -159,6 +157,60 @@ impl Publisher {
...
+        topic: String,
+    ) -> Self {
+        Self {
+            inner: client,
+            topic,
+            batching_settings: BatchingSettings::default(),
+        }
+    }
+
+    // Change the message batching settings.
```
dbolduc: For public documentation, we need three slashes `///`.

Also, these functions should have examples. Your plan was probably to defer writing the examples until we have the exact interfaces we want. That makes sense to me.

src/pubsub/src/publisher/client.rs
```diff
@@ -159,6 +157,60 @@ impl Publisher {
...
+            topic: self.topic,
+            batching_settings: self.batching_settings,
+        }
+    }
+}
+
+#[derive(Clone, Debug)]
+#[non_exhaustive]
+pub struct BatchingSettings {
+    // To turn off batching, set the value of max_messages to 1.
```
dbolduc: Stale comment. There is no `max_messages`

src/pubsub/src/publisher/client.rs
```diff
@@ -159,6 +157,60 @@ impl Publisher {
...
+    pub byte_threshold: Option<u32>,
+    pub delay_threshold_ms: Option<std::time::Duration>,
+}
+
+impl std::default::Default for BatchingSettings {
+    fn default() -> Self {
+        Self {
+            message_count_threshold: Some(100u32),
+            byte_threshold: Some(1000u32),
+            delay_threshold_ms: Some(std::time::Duration::from_millis(10)),
```
dbolduc: why make these options? They always have a value it seems.

src/pubsub/src/publisher/client.rs
```diff
@@ -159,6 +157,60 @@ impl Publisher {
     }
 }
 
+#[derive(Clone, Debug)]
+pub struct PublisherBuilder {
```
dbolduc: same comment about public types needing documentation

src/pubsub/src/publisher/client.rs
```diff
@@ -159,6 +157,60 @@ impl Publisher {
...
+    ) -> Self {
+        Self {
+            inner: client,
+            topic,
+            batching_settings: BatchingSettings::default(),
+        }
+    }
+
+    // Change the message batching settings.
+    pub fn with_batching(mut self, settings: BatchingSettings) -> PublisherBuilder {
```
dbolduc: tests for this?

we could have the field be `pub(crate)` on the client and inspect it in a test.

src/pubsub/src/publisher/client.rs
suzmue: +1 I think these will always have some value, even if its the default one we choose

src/pubsub/src/publisher/client.rs
```diff
@@ -159,6 +157,60 @@ impl Publisher {
...
+        Self {
+            message_count_threshold: Some(100u32),
+            byte_threshold: Some(1000u32),
+            delay_threshold_ms: Some(std::time::Duration::from_millis(10)),
+        }
+    }
+}
+
 #[cfg(test)]
 mod tests {
```
suzmue, line 104: Existing test needs to be updated to build the builder.

Also tests for:
- unconfigured Builder uses the correct defaults
- configuration works

src/pubsub/src/publisher/client.rs
```diff
@@ -159,6 +157,60 @@ impl Publisher {
...
 }
 
+#[derive(Clone, Debug)]
+pub struct PublisherBuilder {
+    pub(crate) inner: crate::generated::gapic_dataplane::client::Publisher,
+    topic: String,
+    batching_settings: BatchingSettings,
+}
+
+impl PublisherBuilder {
```
suzmue: For crate organization, this should be exported in pubsub::builder::*

src/pubsub/src/publisher/client.rs
```diff
@@ -159,6 +157,60 @@ impl Publisher {
...
+    pub message_count_threshold: Option<u32>,
+    pub byte_threshold: Option<u32>,
+    pub delay_threshold_ms: Option<std::time::Duration>,
+}
+
+impl std::default::Default for BatchingSettings {
+    fn default() -> Self {
+        Self {
+            message_count_threshold: Some(100u32),
+            byte_threshold: Some(1000u32),
```
coryan: nitty, nitty, nit: I find these more readable:


```suggestion
            message_count_threshold: Some(100_u32),
            byte_threshold: Some(1000_u32),
```

suzmue:
Quick commit message feedback:
- nit: I think this is PR type "feat"
- Would be nice to include quick description on why we are adding a builder(). For example "There are several batching options that need to be configured on each Publisher, not on the client. A publisher's batching settings are configurable only before publishing begins, so we use a builder to allow this configuration on setup." ....or something better than that

src/pubsub/src/publisher/client.rs
PhongChuong: Updated to Options.

src/pubsub/src/publisher/client.rs
PhongChuong: Fixed.

src/pubsub/src/publisher/client.rs
PhongChuong: Removed Options.

src/pubsub/src/publisher/client.rs
PhongChuong: Added test.

src/pubsub/src/publisher/client.rs
PhongChuong: Done.

src/pubsub/src/publisher/client.rs
PhongChuong: Done.

src/pubsub/src/publisher/client.rs
PhongChuong: Updated to include tests.

src/pubsub/src/publisher/client.rs
PhongChuong: Added documentation.

src/pubsub/src/publisher/client.rs
PhongChuong: Updated to use /// and added a TODO.

PhongChuong:
Thanks for the review. I've updated the PR with the feedback. I also moved options and publisher to its own file. PTAL.

src/pubsub/src/publisher/publisher.rs
```diff
@@ -0,0 +1,153 @@
...
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn default_batching() -> anyhow::Result<()> {
+        let client = PublisherClient::builder().build().await?;
+        let publisher = client
+            .publisher("projects/my-project/topics/my-topic".to_string())
+            .build();
+        assert_ne!(publisher.batching_options.message_count_threshold, 1_u32);
```
suzmue: This test condition doesn't make sense to me. Can we compare against BatchingOptions::default() instead or something?

suzmue (approved):

src/pubsub/src/publisher/publisher.rs
PhongChuong: Updated.
Note: I'm not a big fan of testing default configuration values. Ideally, we should be testing the default behavior but that is not yet possible.

suzmue (approved):

PhongChuong:
Thanks for the review everyone.

PR #3649, opened 2025-10-30
fix(gax-internal): Remove gRPC status attributes from HTTP error spans

westarle:
Aligns T4 HTTP span attributes with go/clo:product-requirements-v1. Removes rpc.grpc.status_code and grpc.status from being recorded for errors originating from HTTP status codes.

### DIFF
--- a/src/gax-internal/src/observability/errors.rs
+++ b/src/gax-internal/src/observability/errors.rs
@@ -12,7 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use gax::error::rpc::Code;
 use http::StatusCode;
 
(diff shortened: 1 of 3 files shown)


src/gax-internal/src/observability/errors.rs
```diff
@@ -94,39 +93,6 @@ impl ErrorType {
             ErrorType::Internal => INTERNAL.to_string(),
         }
     }
-
-    pub(crate) fn grpc_code(&self) -> Code {
```
westarle, line 98: I will probably need this for the metrics (where we are hoping to emit a standard code) and gRPC tracing of course, but I can add it back then.

PR #3650, opened 2025-10-30
ci(auth): make all integration test run in serial

alvarowolfx:
Fix flaky tests in auth, where tests that change set the env var `GOOGLE_APPLICATION_CREDENTIALS` to test ADC flow can affect other tests starting in parallel.


dbolduc (dismissed):

dbolduc:
I think the other strategy is to move the tests that need to be in serial into a different binary, e.g. `driver_serial.rs`.

Then we would still run the other tests in parallel. I think overall it would be slightly faster.

dbolduc (approved):

Issue #3626, opened 2025-10-24
Bootstrap a storage client with support for bidi streaming reads


coryan:
We want to add methods to the Cloud Storage client to support bidi streaming reads.  It is unclear if these methods should live in a new client, or an existing client.  It is clear that it will take multiple PRs to implement them, and potentially multiple breaking changes (to the new methods, the existing APIs should remain stable) as we write them. We should incubate these methods separate from `google_cloud_storage::client::Storage` until their API is stabilized and we are ready to release.

We will need to initialize a `gaxi::grpc::Client`, which means we need to have a `ClientConfig`.

We probably do not want to create a new `ClientBuilder` because maintaining it would be expensive.

PR #3652, opened 2025-10-30
refactor(storage): consolidate test bucket code

coryan:
Move the code to create and cleanup test buckets to a single place.

Motivated by #3626

### DIFF
--- a/src/storage/examples/src/lib.rs
+++ b/src/storage/examples/src/lib.rs
@@ -50,44 +50,29 @@ pub async fn run_anywhere_cache_examples(buckets: &mut Vec<String>) -> anyhow::R
         tracing::subscriber::set_default(subscriber)
     };
 
-    let client = control_client().await?;
-    let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
-
-    let id = random_bucket_id();
-    buckets.push(id.clone());
     let zone = "us-central1-f";
     tracing::info!("Create bucket for anywhere cache examples");
-    let _bucket = client
-        .create_bucket()
-        .set_parent("projects/_")
-        .set_bucket_id(&id)
-        .set_bucket(
-            Bucket::new()
-                .set_project(format!("projects/{project_id}"))
-                .set_location("us-central1")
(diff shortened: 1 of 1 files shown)


src/storage/examples/src/lib.rs
```diff
@@ -67,7 +69,7 @@ pub async fn run_anywhere_cache_examples(buckets: &mut Vec<String>) -> anyhow::R
                 .set_iam_config(IamConfig::new().set_uniform_bucket_level_access(
```
dbolduc: `create_test_hns_bucket` ?

src/storage/examples/src/lib.rs
```diff
@@ -67,7 +69,7 @@ pub async fn run_anywhere_cache_examples(buckets: &mut Vec<String>) -> anyhow::R
...
                     UniformBucketLevelAccess::new().set_enabled(true),
                 ))
-                .set_labels([("integration-test", "true")]),
+                .set_labels([("integration-tests", "true")]),
```
dbolduc: Pick one. Looks like singular is more common?

suzmue (approved):

src/storage/examples/src/lib.rs
```diff
@@ -650,7 +652,7 @@ async fn create_bucket_kms_key(
             google_cloud_storage::model::Bucket::new()
                 .set_project(format!("projects/{project_id}"))
                 .set_location("US-CENTRAL1")
```
dbolduc, line 639: optional: This could probably be `create_test_bucket` if we refactor the `id` business

src/storage/examples/src/lib.rs
coryan: Done.

src/storage/examples/src/lib.rs
coryan: Good catch, look like I reversed so other changes. Fixed.

src/storage/examples/src/lib.rs
coryan: I am going to leave this alone.

Issue #3640, opened 2025-10-29
DNS and connection errors should be retryable


coryan:
> 
> Once I got a DNS error (I am running this from home, on my laptop):
> 
> ```
> ---- parallel stdout ----
> Error: the transport reports an error: error sending request for url (https://storage.googleapis.com/upload/storage/v1/b/coryan-test-bucket/o?uploadType=multipart&name=dc0d9e8b-b534-487a-b285-5f40ca473bcd%2Fobject-000)
> 
> Caused by:
>     0: error sending request for url (https://storage.googleapis.com/upload/storage/v1/b/coryan-test-bucket/o?uploadType=multipart&name=dc0d9e8b-b534-487a-b285-5f40ca473bcd%2Fobject-000)
>     1: client error (Connect)
>     2: dns error
>     3: failed to lookup address information: nodename nor servname provided, or not known
> ```
> 
> The [retry loop](https://cloud.google.com/storage/docs/retry-strategy) can handle such errors, but requires idempotent requests, and therefore  [request preconditions](https://cloud.google.com/storage/docs/request-preconditions).

I am wrong in that last bit. The DNS + connection errors are always retryable, even if the operation is not idempotent.  No request data never left the local computer.

This affects both the `Aip194Strict` policy and the default policy for storage.

 _Originally posted by @coryan in [#3574](https://github.com/googleapis/google-cloud-rust/issues/3574#issuecomment-3426572454)_

dbolduc:
I think these errors are already retried for the `Storage` client.

I used @coryan's repro from https://github.com/googleapis/google-cloud-rust/issues/3574#issuecomment-3426572454

Plus I added a logging retry policy:

```rs
    use google_cloud_gax::error::Error;
    use google_cloud_gax::retry_policy::LimitedElapsedTime;
    use google_cloud_gax::retry_policy::RetryPolicy;
    use google_cloud_gax::retry_policy::RetryPolicyExt;
    use google_cloud_gax::retry_result::RetryResult;
    use google_cloud_gax::retry_state::RetryState;
    use google_cloud_storage::retry_policy::RetryableErrors;
    use std::time::Duration;

    #[derive(Clone, Debug)]
    pub struct LoggingRetry<T>
    where
        T: RetryPolicy,
    {
        inner: T,
    }

    impl<T> LoggingRetry<T>
    where
        T: RetryPolicy,
    {
        pub fn new(inner: T) -> Self {
            Self { inner }
        }
    }

[...]

PR #3654, opened 2025-10-31
feat(gax): introduce connect error

dbolduc:
Part of the work for #3640

Introduce a connection error kind, which is distinct from `is_io()`.

Note that if we think `is_connect()` should imply `is_transport()`, we can always change the internal representation later.

### DIFF
--- a/src/gax/src/error/core_error.rs
+++ b/src/gax/src/error/core_error.rs
@@ -580,6 +580,36 @@ impl Error {
 
     /// Not part of the public API, subject to change without notice.
     ///
+    /// A problem connecting with the endpoint.
+    ///
+    /// Examples include DNS errors and broken connections.
+    #[cfg_attr(not(feature = "_internal-semver"), doc(hidden))]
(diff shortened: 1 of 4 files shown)


src/gax/src/error/core_error.rs
```diff
@@ -578,6 +578,36 @@ impl Error {
...
+    #[cfg_attr(not(feature = "_internal-semver"), doc(hidden))]
+    pub fn connect<T: Into<BoxError>>(source: T) -> Self {
+        Self {
+            kind: ErrorKind::Connect,
+            source: Some(source.into()),
+        }
+    }
+
+    /// Not part of the public API, subject to change without notice.
+    ///
```
coryan, line 595: Separately, we should think about how many of these should become public...

coryan (approved):

src/gax/src/error/core_error.rs
dbolduc: Yes, and expose the policy traits

PR #3659, opened 2025-10-31
refactor(gax-internal): Record HTTP span attributes on gax::Error

westarle:


src/gax-internal/src/observability/errors.rs
```diff
@@ -28,53 +29,46 @@ pub enum ErrorType {
...
-            e if e.is_request() => ErrorType::ClientRequestError,
-            e if e.is_body() => ErrorType::ClientRequestBodyError,
-            e if e.is_decode() => ErrorType::ClientResponseDecodeError,
-            e if e.is_redirect() => ErrorType::ClientRedirectError,
-            _ => ErrorType::Internal,
+            e if e.is_exhausted() => ErrorType::ResourceExhausted,
+            e if e.is_serialization() => ErrorType::ClientRequestError,
+            e if e.is_deserialization() => ErrorType::ClientResponseDecodeError,
+            e if e.is_authentication() => ErrorType::ClientAuthenticationError,
+            e if e.is_io() || e.is_connect() => {
```
westarle: I'm not 100% on this, on one hand it is nice to be offering these detailed errors, on the other hand I don't know about the stability of reqwest errors and their mapping to our errors.

src/gax-internal/src/observability/errors.rs
```diff
@@ -25,57 +26,48 @@ pub enum ErrorType {
...
+                }
+            }
+        }
+    }
+
+    fn from_io_error(err: &gax::error::Error) -> Self {
+        // Attempt to detect redirect error from inner reqwest error
+        let mut source = err.source();
+        while let Some(s) = source {
+            if let Some(reqwest_err) = s.downcast_ref::<reqwest::Error>() {
```
westarle: *topic 1*: do we want to provide these more detailed errors in tracing?

If yes, are we OK with a hard dependency on Reqwest or should we put them in gax Error?

src/gax-internal/src/http.rs
```diff
@@ -180,26 +180,39 @@ impl ReqwestClient {
...
+                    self::to_http_response(response).await
+                }
+            }
+            Err(err) => Err(map_send_error(err)),
+        };
+
+        #[cfg(google_cloud_unstable_tracing)]
+        if let Some(ref s) = span {
+            let _enter = s.enter(); // Re-enter span context
+            record_http_response_attributes(s, &gax_result, status_code);
```
westarle: *Topic 2*: Collecting error results down here lets us (makes us) handle deser problems in HTTP span.

First, we have to decide if that is good.

Then, do we want to move span creation to capture auth errors (no?) and build request errors (yes?).

westarle:
Open for discussion and feedback on the 2 topics I posted.

src/gax-internal/src/observability/errors.rs
coryan: No.  If anything, you want to use the `is_connect()`, `is_timeout()`, etc. predicates from `gax::error::Error`.

westarle: I think I want to delete it and stick to our pub Error API.

src/gax-internal/src/http.rs
westarle: Looking at OTel specs:

>HTTP client spans SHOULD end sometime after the HTTP response headers are fully read (or when they fail to be read). This may or may not include reading the response body.

src/gax-internal/src/http.rs
```diff
@@ -180,26 +180,36 @@ impl ReqwestClient {
...
+        #[cfg(google_cloud_unstable_tracing)]
+        if let Some(s) = span {
+            record_http_response_attributes(&s, &intermediate_result);
+            // Span 's' is dropped here, ending it before decoding starts.
+        }
+
+        match intermediate_result {
+            Ok(response) => {
+                if !response.status().is_success() {
+                    self::to_http_error(response).await
```
westarle, line 206: I want to move this into the span.

westarle:
This has changed sufficiently from where it started that I'm going to start a new PR.

PR #3660, opened 2025-10-31
fix(pubsub): cleanup exported types in pubsub crate

suzmue:
This change removes some builders that were exported in the client module.


coryan (dismissed):

suzmue:
Removed the change that drops the Publish request builder because it breaks the doc tests. Need to figure out a solution for that, but will do in a follow up.

PR #3664, opened 2025-11-02
impl(pubsub): add background worker

### LINKED ISSUE #3684: Add asynchronous publishing in the Publisher

suzmue:
Add a background worker to the Publisher to publish messages asynchronously. This change does not add batching. Each message is sent in its own request. 

Fixes #3684

### DIFF
--- a/src/pubsub/src/publisher/model_ext.rs
+++ b/src/pubsub/src/publisher/model_ext.rs
@@ -23,7 +23,7 @@ use tokio::sync::oneshot;
 /// result of the publish call: either a server-assigned message ID `String`
 /// or an `Error` if the publish failed.
 pub struct PublishHandle {
-    pub(crate) rx: oneshot::Receiver<Result<String, crate::Error>>,
+    pub(crate) rx: oneshot::Receiver<crate::Result<String>>,
 }
 
 impl Future for PublishHandle {
@@ -31,11 +31,8 @@ impl Future for PublishHandle {
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let result = ready!(Pin::new(&mut self.rx).poll(cx));
-        Poll::Ready(match result {
-            Ok(res) => res,
-            // This error will only occur if the sender of the self.rx was dropped,
-            // which would be a bug.
-            Err(_) => Err(crate::Error::deser("unable to get message id")),
-        })
+        // An error will only occur if the sender of the self.rx was dropped,
+        // which would be a bug.
+        Poll::Ready(result.expect("the client library should not release the sender"))
     }
 }
--- a/src/pubsub/src/publisher/publisher.rs
+++ b/src/pubsub/src/publisher/publisher.rs
@@ -170,7 +170,7 @@ impl Worker {
                 let result = request
                     .send()
                     .await
-                    .map(|response| response.message_ids.first().cloned().unwrap_or_default());
+                    .map(|mut response| response.message_ids.pop().unwrap_or_default());
 
                 // The user may have dropped the handle, so it is ok if this fails.
                 let _ = msg.tx.send(result);


src/pubsub/src/publisher/model_ext.rs
```diff
@@ -0,0 +1,41 @@
...
+impl Future for PublishHandle {
+    type Output = crate::Result<String>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let result = ready!(Pin::new(&mut self.rx).poll(cx));
+        Poll::Ready(match result {
+            Ok(res) => res,
+            // This error will only occur if the sender of the self.rx was dropped,
+            // which would be a bug.
+            Err(_) => Err(crate::Error::deser("unable to get message id")),
```
coryan: Arguably, if this is a bug then we should do:

```suggestion
            // This error will only occur if the sender of the self.rx was dropped,
            // which would be a bug.
            Err(e) => panic!("the client library should not release the sender: {e:?}"),
```

src/pubsub/src/publisher/model_ext.rs
```diff
@@ -0,0 +1,41 @@
...
+use std::task::{Context, Poll, ready};
+use tokio::sync::oneshot;
+
+/// A handle that represents an in-flight publish operation.
+///
+/// This struct is a `Future`. You can `.await` it to get the final
+/// result of the publish call: either a server-assigned message ID `String`
+/// or an `Error` if the publish failed.
+pub struct PublishHandle {
+    pub(crate) rx: oneshot::Receiver<Result<String, crate::Error>>,
```
coryan: How would one mock this class?

coryan: Why  `Result<.., crate::Error>` here, but `crate::Result<_>` below?

src/pubsub/src/publisher/model_ext.rs
```diff
@@ -0,0 +1,41 @@
...
+}
+
+impl Future for PublishHandle {
+    type Output = crate::Result<String>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let result = ready!(Pin::new(&mut self.rx).poll(cx));
+        Poll::Ready(match result {
+            Ok(res) => res,
```
coryan: If we go the `panic!()` route this can become:


```suggestion
        Poll::Ready(result.expect("$reason"))
```

If we go the error route, then consider `result.unwrap_or_else(|| Err(crate::Error::deser("..."))`

src/pubsub/src/publisher/model_ext.rs
coryan: `Error::deser()` does not seem quite right.

src/pubsub/src/publisher/publisher.rs
```diff
@@ -118,17 +103,167 @@ impl PublisherBuilder {
...
+                let request = client
+                    .publish()
+                    .set_topic(topic)
+                    .set_messages(vec![msg.msg]);
+
+                // Handle the response by extracting the message ID on success.
+                let result = request
+                    .send()
+                    .await
+                    .map(|response| response.message_ids.first().cloned().unwrap_or_default());
```
coryan: Since you are expecting at most 1 element:


```suggestion
                    .map(|response| response.message_ids.pop().unwrap_or_default());
```

src/pubsub/src/publisher/model_ext.rs
suzmue: Mocking is definitely still a TODO for the Publisher overall.

I could imagine we could provide a helper to make a `PublishHandle` from a `crate::Result<String>`. I'd prefer to put that in another PR though.

suzmue: Just autocomplete behaving differently and I missed cleaning it up to crate::Result in this spot :)

src/pubsub/src/publisher/model_ext.rs
suzmue: Done.

suzmue: Yeah I agree, its a little awkward... Swapped to panic in this instance.

src/pubsub/src/publisher/model_ext.rs
suzmue: Done!

src/pubsub/src/publisher/publisher.rs
suzmue: Done.

coryan (approved):

Issue #3360, opened 2025-09-18
Set up testing of hand crafted examples


suzmue:
Set up the initial directories, etc.

Add the tests to the integration test runner: https://github.com/googleapis/google-cloud-rust/blob/main/.gcb/integration.yaml

For an initial sample to use to run, [pubsub_quickstart_create_topic/pubsub_create_topic](http://cloud/pubsub/docs/samples/pubsub-create-topic?hl=en#pubsub_create_topic) would be a good candidate.

PR #3670, opened 2025-11-03
test(pubsub): add first example for pubsub

suzmue:
Set up the samples crate for all our samples for google-cloud-pubsub code can go. This also refactors the helpers to move to the pubsub-samples crate.

For #3360

### DIFF
--- a/src/pubsub/examples/src/lib.rs
+++ b/src/pubsub/examples/src/lib.rs
@@ -22,18 +22,15 @@ pub async fn run_topic_examples(topic_names: &mut Vec<String>) -> anyhow::Result
     let client = TopicAdmin::builder().build().await?;
     let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
 
-    // TODO: use a real random topic and project.
-    let name = random_topic_name(project_id);
-    topic_names.push(name.clone());
-    topic::create_topic::sample(&client, &name).await?;
+    let id = random_topic_id();
+    topic_names.push(format!("projects/{project_id}/topics/{id}"));
+    topic::create_topic::sample(&client, &project_id, &id).await?;
 
     Ok(())
 }
 
 pub async fn cleanup_test_topic(client: &TopicAdmin, topic_name: String) -> anyhow::Result<()> {
-    tracing::info!("testing delete_topic()");
     client.delete_topic().set_topic(topic_name).send().await?;
-    tracing::info!("success on delete_topic");
     Ok(())
 }
 
@@ -46,17 +43,16 @@ pub async fn create_test_topic() -> anyhow::Result<(TopicAdmin, Topic)> {
 
     cleanup_stale_topics(&client, &project_id).await?;
 
-    let topic_name = random_topic_name(project_id);
+    let topic_id = random_topic_id();
     let now = chrono::Utc::now().timestamp().to_string();
 
-    tracing::info!("testing create_topic()");
     let topic = client
         .create_topic()
-        .set_name(topic_name)
+        .set_name(format!("projects/{project_id}/topics/{topic_id}"))
         .set_labels([("integration-test", "true"), ("create-time", &now)])
         .send()
         .await?;
-    tracing::info!("success on create_topic: {topic:?}");
+    println!("success on create_topic(): {topic:?}");
 
     Ok((client, topic))
 }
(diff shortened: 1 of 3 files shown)


src/pubsub/examples/src/topic/create_topic.rs
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// [START pubsub_create_topic]
+use google_cloud_pubsub::{client::TopicAdmin, model::Topic};
+
+pub async fn sample(client: &TopicAdmin, topic_name: &str) -> anyhow::Result<()> {
+    let topic: Topic = client.create_topic().set_name(topic_name).send().await?;
```
coryan: What is the motivation to use a type annotation here? And should we make it easier to grok what is the structure of a topic name?


```suggestion
    let topic = client.create_topic().set_name(format!("projects/{project_id}/topics/{topic_id}")).send().await?;
```

src/pubsub/examples/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+
+use google_cloud_gax::paginator::ItemPaginator as _;
+use google_cloud_pubsub::{client::TopicAdmin, model::Topic};
+use rand::{Rng, distr::Alphanumeric};
+
+pub async fn run_topic_examples(topic_names: &mut Vec<String>) -> anyhow::Result<()> {
+    let client = TopicAdmin::builder().build().await?;
+    let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
+
+    // TODO: use a real random topic and project.
```
coryan: Is this TODO stale?

src/pubsub/examples/tests/integration.rs
```diff
@@ -0,0 +1,32 @@
...
+
+    #[tokio::test]
+    async fn topic_examples() -> anyhow::Result<()> {
+        let client = TopicAdmin::builder().build().await?;
+
+        let mut topics = Vec::new();
+        let result = run_topic_examples(&mut topics).await;
+        // Ignore cleanup errors.
+        for name in topics.into_iter() {
+            let _ = cleanup_test_topic(&client, name).await;
```
coryan: Consider:


```suggestion
            if let Err(e) = cleanup_test_topic(&client, name.clone()).await {
                println!("Error cleaning up test topic {name}: {e:?}");
            }
```

src/pubsub/examples/src/lib.rs
```diff
@@ -0,0 +1,117 @@
...
+    let name = random_topic_name(project_id);
+    topic_names.push(name.clone());
+    topic::create_topic::sample(&client, &name).await?;
+
+    Ok(())
+}
+
+pub async fn cleanup_test_topic(client: &TopicAdmin, topic_name: String) -> anyhow::Result<()> {
+    tracing::info!("testing delete_topic()");
```
coryan: This trace message seems wrong now?

src/pubsub/examples/src/lib.rs
suzmue: yes.

src/pubsub/examples/src/lib.rs
suzmue: done.

src/pubsub/examples/src/topic/create_topic.rs
suzmue: Just wanted to show what the type was since the example was so short anyway.

Updated to use the format.

src/pubsub/examples/tests/integration.rs
suzmue: done.

src/pubsub/examples/src/lib.rs
```diff
@@ -0,0 +1,113 @@
...
+                .get("create-time")
+                .and_then(|v| v.parse::<i64>().ok())
+                .and_then(|s| chrono::DateTime::from_timestamp(s, 0))
+                .is_some_and(|create_time| create_time < stale_deadline)
+        {
+            let client = client.clone();
+            let name = topic.name.clone();
+            pending.push(tokio::spawn(async move {
+                cleanup_test_topic(&client, name).await
+            }));
```
coryan, line 87: optional: I wonder ... since you are going to await for these anyway, maybe there is no need to spawn tasks for them?

suzmue: True, though this way we can fire off multiple requests at the same time.

Will consider in a follow up :)

PR #3673, opened 2025-11-03
refactor(gax-internal): Record HTTP span attributes after error parsing

westarle:
This PR refactors how HTTP span attributes are recorded in `google-cloud-gax-internal`.

Previously, attributes were recorded immediately after receiving the `reqwest::Response` (or error), before any body parsing or error conversion took place. This meant that
for non-2xx responses, the span would only contain the raw HTTP status code, but not the richer error details (like `google.rpc.ErrorInfo` reason) that are available after
parsing the error body.

This change moves the attribute recording logic to *after* the error has been parsed (via `to_http_error`), but *before* the success response body is fully read and
deserialized (via `to_http_response`).

Key changes:
- `ReqwestClient::request_attempt` now parses HTTP errors (non-2xx responses) *within* the span's lifetime.
- `record_http_response_attributes` now accepts `Result<&reqwest::Response, &gax::error::Error>`, allowing it to extract details from the fully parsed `gax::Error`.
- `ErrorType::from_gax_error` is updated to handle more error kinds from `gax::Error` (e.g., `Binding`, `Exhausted` renamed to `ClientRetryExhausted`).
- `RESOURCE_EXHAUSTED` constant renamed to `CLIENT_RETRY_EXHAUSTED` to better reflect its meaning and avoid confusion with server-side 429s.

This ensures that error spans contain the most accurate `error.type` and other attributes, aligned with the public API's error reporting.

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -194,24 +194,17 @@ impl ReqwestClient {
 
         let intermediate_result = reqwest_result.map_err(map_send_error);
         let intermediate_result = match intermediate_result {
-            Ok(response) if !response.status().is_success() => {
-                self::to_http_error(response).await
-            }
+            Ok(response) if !response.status().is_success() => self::to_http_error(response).await,
             other => other,
         };
 
         // Record span before parsing result, after parsing error.
         #[cfg(google_cloud_unstable_tracing)]
-        if let Some(s) = &span {
-            record_http_response_attributes(s, intermediate_result.as_ref());
+        if let Some(s) = span {
+            record_http_response_attributes(&s, intermediate_result.as_ref());
         }
-        #[cfg(google_cloud_unstable_tracing)]
-        drop(span);
 
-        match intermediate_result {
-            Ok(response) => self::to_http_response(response).await,
-            Err(err) => Err(err),
-        }
+        self::to_http_response(intermediate_result?).await
     }
 
     fn get_retry_policy(&self, options: &gax::options::RequestOptions) -> Arc<dyn RetryPolicy> {
--- a/src/gax-internal/src/observability/errors.rs
+++ b/src/gax-internal/src/observability/errors.rs
@@ -22,13 +22,14 @@ pub enum ErrorType {
         code: StatusCode,
         reason: Option<String>,
     },
+    RpcError(gax::error::rpc::Code),
     ClientTimeout,
     ClientConnectionError,
     ClientRequestError,
     ClientResponseDecodeError,
     ClientAuthenticationError,
     ClientRetryExhausted,
-    Internal,
+    Unknown,
 }
 
 impl ErrorType {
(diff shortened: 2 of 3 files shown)


src/gax-internal/src/observability/errors.rs
```diff
@@ -25,56 +25,33 @@ pub enum ErrorType {
...
+            e => {
+                if let Some(status) = e.http_status_code() {
+                    ErrorType::HttpError {
+                        code: http::StatusCode::from_u16(status)
+                            .unwrap_or(http::StatusCode::INTERNAL_SERVER_ERROR),
+                        reason: None,
+                    }
+                } else {
+                    ErrorType::Internal
+                }
```
coryan: I think you want to create a branch for `ErrorType::Unknown` and only use the http status if `e.is_transport()`.

You should also consider the case where `e.status().is_some()`: the most detailed error information arrived from the service.

src/gax-internal/src/http.rs
```diff
@@ -180,26 +180,38 @@ impl ReqwestClient {
...
+            other => other,
+        };
+
+        // Record span before parsing result, after parsing error.
+        #[cfg(google_cloud_unstable_tracing)]
+        if let Some(s) = &span {
+            record_http_response_attributes(s, intermediate_result.as_ref());
+        }
+        #[cfg(google_cloud_unstable_tracing)]
+        drop(span);
```
coryan: I think:


```suggestion
        span.map(|s| record_http_response_attributes(s, intermediate_result.as_ref()) );
```

Has the same effect.

src/gax-internal/src/http.rs
```diff
@@ -180,26 +180,38 @@ impl ReqwestClient {
...
 
-        let response = response_result.map_err(map_send_error)?;
-        if !response.status().is_success() {
-            return self::to_http_error(response).await;
+        match intermediate_result {
+            Ok(response) => self::to_http_response(response).await,
+            Err(err) => Err(err),
```
coryan: maybe:

```suggestion
        let response = intermediate_result?;
        self::to_http_response(response).await
```

??

src/gax-internal/src/observability/errors.rs
```diff
@@ -41,17 +43,13 @@ impl ErrorType {
...
-                    ErrorType::Internal
-                }
-            }
+            e if e.status().is_some() => ErrorType::RpcError(e.status().unwrap().code),
+            e if e.is_transport() => e
+                .http_status_code()
+                .and_then(|s| http::StatusCode::from_u16(s).ok())
+                .map(|code| ErrorType::HttpError { code, reason: None })
+                .unwrap_or(ErrorType::Unknown),
+            _ => ErrorType::Internal,
```
coryan: Maybe `Unknown` would be the best choice here?

src/gax-internal/src/observability/errors.rs
westarle: Thank you, I took care of that.

src/gax-internal/src/http.rs
westarle: Done!

src/gax-internal/src/http.rs
westarle: Done

src/gax-internal/src/observability/errors.rs
westarle: Done, thanks for the review!

PR #3683, opened 2025-11-03
fix(storage): `StorageImpl` nameable outside crate

### LINKED ISSUE #3663: storage: template-based stub makes it hard to create a generic type holding a `Storage`
The `Storage` client type takes a generic for its stub implementation, which is well-documented. However, the default implementation used for the real client doesn't seem to be nameable outside the crate, and so one can't define a type holding a `Storage` like

```
struct App<
   StorStub = google_cloud_storage::storage::transport::Storage
> where StorStub: google_cloud_storage::stub::Storage + 'static
{
    // ...
    storage: Storage<StorStub>
}
```

Specifically, you can't give the correct default and thus can't do something like a `MockableApp<StorStub>` with the real application using a `type App = MockableApp<gcs::storage::transport::Storage>`. It then becomes very hard to send that application state around anything long-lived, because you can't rely on type inference to solve the problem.

dbolduc:
Fixes #3663 

Make the default stub implementation name-able from outside the crate. This lets applications provide a default implementation for their own generics.

Note that this type is not constructible from outside the crate. It has no public fields. Its only public fns are its impl of `stub::Storage`.

I think this type belongs in the `stub` module. I called it `StorageImpl`.

If we are happy with these names, I will clean up the internal wiring, either in this PR or in a follow up.

### DIFF
--- a/src/storage/src/lib.rs
+++ b/src/storage/src/lib.rs
@@ -69,7 +69,7 @@ pub mod model_ext;
 pub mod stub {
     pub use crate::control::stub::*;
     pub use crate::storage::stub::*;
-    pub use crate::storage::transport::Storage as StorageImpl;
+    pub use crate::storage::transport::Storage as DefaultStorage;
 }
 
 #[allow(dead_code)]
--- a/src/storage/src/storage.rs
+++ b/src/storage/src/storage.rs
@@ -20,7 +20,7 @@ pub(crate) mod read_object;
 pub mod request_options;
 pub mod streaming_source;
 pub mod stub;
-pub mod transport;
+pub(crate) mod transport;
 pub(crate) mod v1;
 pub(crate) mod write_object;
 
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -89,7 +89,7 @@ use std::sync::Arc;
 /// [Private Google Access with VPC Service Controls]: https://cloud.google.com/vpc-service-controls/docs/private-connectivity
 /// [Application Default Credentials]: https://cloud.google.com/docs/authentication#adc
 #[derive(Clone, Debug)]
-pub struct Storage<S = crate::stub::StorageImpl>
+pub struct Storage<S = crate::stub::DefaultStorage>
 where
     S: crate::stub::Storage + 'static,
 {
--- a/src/storage/src/storage/transport.rs
+++ b/src/storage/src/storage/transport.rs
@@ -34,8 +34,8 @@ use std::sync::Arc;
 /// ```
 /// # async fn sample() -> anyhow::Result<()> {
 /// use google_cloud_storage::client::Storage;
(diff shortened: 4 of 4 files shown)


src/storage/src/storage/client.rs
```diff
@@ -89,9 +89,9 @@ use std::sync::Arc;
 /// [Private Google Access with VPC Service Controls]: https://cloud.google.com/vpc-service-controls/docs/private-connectivity
 /// [Application Default Credentials]: https://cloud.google.com/docs/authentication#adc
 #[derive(Clone, Debug)]
-pub struct Storage<S = crate::storage::transport::Storage>
+pub struct Storage<S = crate::stub::StorageImpl>
```
coryan: Maybe `crate::stub::DefaultStorageStub` ?

src/storage/src/lib.rs
```diff
@@ -69,6 +69,7 @@ pub mod model_ext;
 pub mod stub {
     pub use crate::control::stub::*;
     pub use crate::storage::stub::*;
+    pub use crate::storage::transport::Storage as StorageImpl;
```
coryan: This makes the type public, which means it goes into the documentation, and it current has no documentation.  Luckily the constructor is `pub(crate)`, so at least there are no new methods to document.

src/storage/src/storage.rs
```diff
@@ -20,7 +20,7 @@ pub(crate) mod read_object;
 pub mod request_options;
 pub mod streaming_source;
 pub mod stub;
-pub(crate) mod transport;
+pub mod transport;
```
coryan: Do we need to make this change?  It makes the module public and it would be only too easy to add a public symbol by accident in that module too?

src/storage/src/storage/client.rs
dbolduc: Thanks for the suggestion. I will go with `crate::stub::DefaultStorage` to not repeat "stub"

src/storage/src/lib.rs
dbolduc: > and it currently has no documentation

This PR adds documentation.

image

src/storage/src/storage.rs
dbolduc: > Do we need to make this change?

Apparently not. Thanks for catching. Reverted.

> It makes the module public

I didn't think so. It is within the non-public `mod storage`:

https://github.com/googleapis/google-cloud-rust/blob/b0dcf168dea463b802f4fe109053629a0e0d950f/src/storage/src/lib.rs#L43

> it would be only too easy to add a public symbol by accident in that module too?

We would have to accidentally export it too.

src/storage/src/lib.rs
dbolduc: You help me realize that the `pub mod stub` did not have documentation. Sent #3690 for that.

coryan (approved):

Issue #3687, opened 2025-11-03
Support Pub/Sub batch limits by configuring max delay for sending a batch


suzmue:
See https://cloud.google.com/pubsub/docs/batch-messaging#use_batch_messaging for batch messaging documentation.

Issue #3679, opened 2025-11-03
Add minimal batching to the Publisher


suzmue:

suzmue:
Batching using the message count and delay threshold is implemented.

Remaining work to respect the byte threshold (and the overall max byte threshold) is covered by #3748

PR #3693, opened 2025-11-03
feat(pubsub): implement batching for message count

suzmue:
Add the initial batching implementation to the Publisher. `Batch` is just a simple wrapper around a vector for now, but in the future this will also support setting limits on max_bytes.

This PR includes both message count and delay settings for batching. Without the delay, we cannot be sure partial batches will eventually send. Without message count being supported, we cannot test delay since everything is sent immediately.

Flush will be added in a future PR.

For #3679 , #3687

### DIFF
--- a/src/pubsub/src/publisher/options.rs
+++ b/src/pubsub/src/publisher/options.rs
@@ -58,7 +58,6 @@ impl BatchingOptions {
     /// Set the [BatchingOptions][Self::delay_threshold] field.
     ///
     /// # Example
-    ///
     /// ```
     /// # use google_cloud_pubsub::options::publisher::BatchingOptions;
     /// let options = BatchingOptions::new().set_delay_threshold(std::time::Duration::from_millis(10));
--- a/src/pubsub/src/publisher/publisher.rs
+++ b/src/pubsub/src/publisher/publisher.rs
@@ -14,7 +14,6 @@
 
 use crate::generated::gapic_dataplane::client::Publisher as GapicPublisher;
 use crate::publisher::options::BatchingOptions;
-use std::time::Duration;
 use std::sync::Arc;
 use std::time::Duration;
 use tokio::sync::mpsc::UnboundedSender;
@@ -178,11 +177,7 @@ impl Worker {
 
     async fn run(mut self) {
         let mut batch = Batch::new();
-        // Set max delay at 1 year to avoid overflow. This should probably be in the setting itself.
-        // Would be reasonable to make it much less.
-        let delay = self.batching_options.delay_threshold.min(
-            Duration::from_secs(60 * 60 * 24 * 365), // 1 year.
-        );
+        let delay = self.batching_options.delay_threshold;
         let message_limit = self.batching_options.message_count_threshold;
 
         let timer = tokio::time::sleep(delay);
@@ -193,7 +188,7 @@ impl Worker {
                 // Handle timer events.
                 // This branch will only be checked when there is a non-empty batch,
                 // so this will not fire continuously.
-                _ = &mut timer, if !batch.batch.is_empty() => {
+                _ = &mut timer, if !batch.is_empty() => {
                     let batch_to_send = std::mem::take(&mut batch);
                     tokio::spawn(batch_to_send.send(self.client.clone(), self.topic_name.clone()));
                 }
@@ -202,11 +197,11 @@ impl Worker {
                     match msg {
                         Some(msg) => {
                             // Reset the timer if this is the first message to be added to the batch.
-                            if batch.batch.is_empty() {
+                            if batch.is_empty() {
                                 timer.as_mut().reset(tokio::time::Instant::now() + delay);
                             }
                             batch.push(msg);
-                            if batch.batch.len() as u32 >= message_limit {
+                            if batch.len() as u32 >= message_limit {
                                 let batch_to_send = std::mem::take(&mut batch);
                                 // In the future, we may also want to keep track of JoinHandles in order to
                                 // flush the results.
@@ -215,7 +210,13 @@ impl Worker {
                             }
                         },
                         None => {
-                            // The sender has been dropped, stop running.
+                            // The sender has been dropped send batch and stop running.
+                            // This isn't guaranteed to execute if a user does not .await on the
+                            // corresponding PublishHandles for the batch and the program ends.
+                            if !batch.is_empty() {
+                                let _handle =
+                                        tokio::spawn(batch.send(self.client.clone(), self.topic_name.clone()));
+                            }
                             break;
                         }
                     }
@@ -224,6 +225,7 @@ impl Worker {
         }
     }
 }
+
 #[derive(Debug)]
 struct Batch {
     // TODO(#3686): A batch should also keep track of its total size
@@ -242,6 +244,14 @@ impl Batch {
         Batch { batch: Vec::new() }
     }
 
+    fn is_empty(&self) -> bool {
+        self.batch.is_empty()
+    }
+
+    fn len(&self) -> usize {
+        self.batch.len()
+    }
+
     fn push(&mut self, msg: BundledMessage) {
         self.batch.push(msg);
     }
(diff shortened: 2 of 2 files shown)


src/pubsub/src/publisher/options.rs
coryan, line 15: s/configurating/configuring/

Or better-er:

/// Configure publisher batching behavior.

src/pubsub/src/publisher/options.rs
```diff
@@ -44,18 +44,21 @@ impl BatchingOptions {
...
+    #[allow(dead_code)]
+    pub(crate) fn set_byte_threshold<V: Into<u32>>(mut self, v: V) -> Self {
         self.byte_threshold = v.into();
         self
     }
 
     /// Set the [BatchingOptions][Self::delay_threshold] field.
     ///
     /// # Example
+    ///
```
coryan: We typically do not include a blank line here. Is it needed?

src/pubsub/src/publisher/publisher.rs
```diff
@@ -153,28 +156,100 @@ impl Worker {
...
+                            // The sender has been dropped, stop running.
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+    }
+}
+#[derive(Debug)]
```
coryan, line 229: nit: blank line.

src/pubsub/src/publisher/publisher.rs
```diff
@@ -153,28 +156,100 @@ impl Worker {
...
+
+        // Handle the response by extracting the message ID on success.
+        match request.send().await {
+            Err(e) => {
+                let e = Arc::new(e);
+                txs.into_iter().for_each(|tx| {
+                    // The user may have dropped the handle, so it is ok if this fails.
+                    // TODO(#3689): The error type for this is incorrect, will need to handle
+                    // this error propagation more fully.
+                    let _ = tx.send(Err(gax::error::Error::io(Arc::clone(&e))));
```
coryan: Maybe:

```suggestion
                let e = Arc::new(e);
                txs.into_iter().for_each(|tx| move {
                    // The user may have dropped the handle, so it is ok if this fails.
                    // TODO(#3689): The error type for this is incorrect, will need to handle
                    // this error propagation more fully.
                    let _ = tx.send(Err(gax::error::Error::io(e.clone())));
```

src/pubsub/src/publisher/publisher.rs
```diff
@@ -153,28 +156,100 @@ impl Worker {
...
+        let timer = tokio::time::sleep(delay);
+        // Pin the timer to the stack.
+        tokio::pin!(timer);
+        loop {
+            tokio::select! {
+                // Handle timer events.
+                // This branch will only be checked when there is a non-empty batch,
+                // so this will not fire continuously.
+                _ = &mut timer, if !batch.batch.is_empty() => {
+                    let batch_to_send = std::mem::take(&mut batch);
```
coryan, line 192: TIL

src/pubsub/src/publisher/publisher.rs
```diff
@@ -153,28 +156,100 @@ impl Worker {
...
+                                // In the future, we may also want to keep track of JoinHandles in order to
+                                // flush the results.
+                                let _handle =
+                                    tokio::spawn(batch_to_send.send(self.client.clone(), self.topic_name.clone()));
+                            }
+                        },
+                        None => {
+                            // The sender has been dropped, stop running.
+                            break;
+                        }
```
coryan, line 221: What about any messages that are still in `batch.batch`?

src/pubsub/src/publisher/publisher.rs
```diff
@@ -153,28 +156,100 @@ impl Worker {
...
+                _ = &mut timer, if !batch.batch.is_empty() => {
+                    let batch_to_send = std::mem::take(&mut batch);
+                    tokio::spawn(batch_to_send.send(self.client.clone(), self.topic_name.clone()));
+                }
+                // Handle receiving a message from the channel.
+                msg = self.rx.recv() => {
+                    match msg {
+                        Some(msg) => {
+                            // Reset the timer if this is the first message to be added to the batch.
+                            if batch.batch.is_empty() {
```
coryan: I find the mixing of `batch.push()` and `batch.batch.is_empty()` a bit confusing. let's try to make this consistent?

coryan (changes requested):

src/pubsub/src/publisher/publisher.rs
suzmue: I was intentionally not sending these since we actually can't guarantee that these will successfully be able to send without the full program shutting down first, but I think we should and updated to do so.

If a user is waiting on the PublishHandles, it will prevent the program from shutting down, and we want to be able to send the messages then.

If the user is not waiting on those handles, dropping the Publisher may result in these not successfully sending but I think that is fine (and what the `flush()` function will be useful for). It wouldn't hurt to try to send them regardless, but I was saving reasoning about shutdown to the flush PR which is the way to guarantee things are sent

src/pubsub/src/publisher/options.rs
suzmue: Nope.

src/pubsub/src/publisher/publisher.rs
suzmue: done.

src/pubsub/src/publisher/publisher.rs
suzmue: done.

src/pubsub/src/publisher/publisher.rs
suzmue: Done

src/pubsub/src/publisher/options.rs
suzmue: Nice catch. Cleaning up this documentation in another PR now.

src/pubsub/src/publisher/publisher.rs
coryan: SG.  Feels like we could refactor some of the the code here, maybe to a `batch.flush()` function?

src/pubsub/src/publisher/publisher.rs
```diff
@@ -174,28 +176,111 @@ impl Worker {
...
+                                timer.as_mut().reset(tokio::time::Instant::now() + delay);
+                            }
+                            batch.push(msg);
+                            if batch.len() as u32 >= message_limit {
+                                let batch_to_send = std::mem::take(&mut batch);
+                                // In the future, we may also want to keep track of JoinHandles in order to
+                                // flush the results.
+                                let _handle =
+                                    tokio::spawn(batch_to_send.send(self.client.clone(), self.topic_name.clone()));
+                            }
```
coryan, line 210: It feels like all the `tokio::spawn()` calls could be done inside a `batch.flush()` call?

src/pubsub/src/publisher/publisher.rs
```diff
@@ -286,6 +460,53 @@ mod tests {
...
+            .with_batching(BatchingOptions {
+                delay_threshold: std::time::Duration::from_millis(10),
+                // Maximizing byte and message count threshold to make sure that
+                // delay is used.
+                byte_threshold: u32::MAX,
+                message_count_threshold: u32::MAX,
+            })
+            .build();
+
+        // Test that messages eventually all send.
```
coryan, line 490: This is a decent test, but you can do better.  With Tokio you can control the timers:

https://tokio.rs/tokio/topics/testing

And verify the messages are *not* sent before the 10ms, and *are* sent at 10ms. We already have several tests that do something similar.

coryan (approved):
There are several smaller improvements you could make. I would be Okay if you merged this as-is to unblock other work and started on the clean ups separately.

src/pubsub/src/publisher/publisher.rs
suzmue: Yeah I think some refactoring could be nice. Especially since it is more difficult to edit code within the `tokio::select!` macro.

suzmue:
> There are several smaller improvements you could make. I would be Okay if you merged this as-is to unblock other work and started on the clean ups separately.

SG will merge

PR #3694, opened 2025-11-03
chore: bump version numbers circa 2025-11-03

dbolduc:


src/generated/cloud/compute/v1/.sidekick.toml
```diff
@@ -40,7 +40,7 @@ prefix    = 'compute/v1/'
 method-id = '.google.cloud.compute.v1.globalOrganizationOperations.get'
 
 [codec]
-version                   = '0.2.0'
+version                   = '0.3.0'
```
coryan: I think `rust-bump-versions` got this wrong... the changes are backwards compatible, this should be 0.2.1.  Thoughts?

src/auth/Cargo.toml
```diff
@@ -18,7 +18,7 @@ name = "google-cloud-auth"
 #     version of all downstream dependencies. For details see:
 #         https://github.com/googleapis/google-cloud-rust/issues/3237
 #         https://github.com/googleapis/google-cloud-rust/issues/3265
-version     = "1.1.0"
+version     = "1.2.0"
```
coryan: I would override this to `1.1.1`. No customer visible changes yet.

coryan (changes requested):

src/generated/cloud/compute/v1/.sidekick.toml
dbolduc: Ack. Fixed.

src/auth/Cargo.toml
dbolduc: Ack, my mistake.

I set it to `1.1.1`, but maybe we would be better off reverting the bump altogether. Thoughts?

There would be less risk of calling a minor bump a patch bump (if we make a change that requires a minor bump before the release). And we will probably run `rust-bump-versions` before the next release anyway (so I doubt we will miss the bump altogether).

coryan: >  but maybe we would be better off reverting the bump altogether. Thoughts?

Probably the safest thing, but also kind of annoying.  We will need to remember this until we are ready for a release.

FWIW, your changes in #3677 are arguably worthy of a release, but maybe not of a minor bump.

src/generated/cloud/compute/v1/.sidekick.toml
coryan: I think you missed this line and only fixed the Cargo file.

coryan (changes requested):

src/auth/Cargo.toml
dbolduc: I will leave it in as a patch version. I think we are reasonably confident that we will not add public APIs to auth that aren't feature gated in the next ~week

src/generated/cloud/compute/v1/.sidekick.toml
dbolduc: Correct.

dbolduc: Fixed now.

coryan (approved):

PR #3700, opened 2025-11-04
refactor(test-utils): Add StaticString to AttributeValue

westarle:
Introduces a `StaticString` variant to `AttributeValue` and implements a custom `PartialEq` to allow comparison between `String` and `StaticString` variants. This enables tests to assert on span attributes using string literals without unnecessary allocations.

### DIFF
--- a/src/test-utils/src/test_layer.rs
+++ b/src/test-utils/src/test_layer.rs
@@ -256,7 +256,7 @@ fn find_test_id<S: Subscriber + for<'b> tracing_subscriber::registry::LookupSpan
 ///     assert_eq!(captured.len(), 1);
 ///     let span = &captured[0];
 ///     assert_eq!(span.name, "my_operation");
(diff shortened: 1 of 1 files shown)


src/test-utils/src/test_layer.rs
```diff
@@ -27,15 +27,33 @@ pub struct TestLayerGuard {
 }
 
 /// Represents the value of a captured tracing attribute.
-#[derive(Debug, Clone, PartialEq)]
+#[derive(Debug, Clone)]
 pub enum AttributeValue {
     String(String),
+    StaticString(&'static str),
```
coryan: I should have thought of this, `Cow<>` may be a better fit here: 

https://doc.rust-lang.org/std/borrow/enum.Cow.html

That would combine `String` and `StaticString` into a single variant.

westarle: much nicer, thank you!

coryan (approved):

PR #3703, opened 2025-11-04
feat(pubsub): clamp batching options to service limits

suzmue:
There are particular service limits that the client library should be enforcing for batching defined at https://cloud.google.com/pubsub/quotas. Clamp the user limits when creating the `Publisher`.

The delay threshold is not a service defined limit, but one we can choose for convenience within our library. Setting delays has a maximum value, clamping our max delay to something smaller communicates the ability we have to set delays, which will be less than `Duration::MAX`. The max_delay_threshold isn't a limit that we cannot violate, but we will attempt to send if possible. For example, if an ordered batch is still being attempted, other batches with that ordering_key will continue to wait. 

We can also consider adding a method to expose the new batching values to users for inspection. User's can manually see them by printing with debug.

### DIFF
--- a/src/pubsub/src/publisher/publisher.rs
+++ b/src/pubsub/src/publisher/publisher.rs
@@ -14,8 +14,14 @@
 
 use crate::generated::gapic_dataplane::client::Publisher as GapicPublisher;
 use crate::publisher::options::BatchingOptions;
+use std::time::Duration;
 use tokio::sync::mpsc::UnboundedSender;
 use tokio::sync::{mpsc, oneshot};
+
+const MAX_DELAY: Duration = Duration::from_secs(60 * 60 * 24); // 1 day
+// These limits come from https://cloud.google.com/pubsub/quotas.
+const MAX_MESSAGES: u32 = 1000;
+const MAX_BYTES: u32 = 1e7 as u32; // 10MB
 
 /// Publishes messages to a single topic.
 ///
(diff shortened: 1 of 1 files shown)


src/pubsub/src/publisher/publisher.rs
```diff
@@ -14,9 +14,15 @@
...
 use crate::generated::gapic_dataplane::client::Publisher as GapicPublisher;
 use crate::publisher::options::BatchingOptions;
+use std::time::Duration;
 use tokio::sync::mpsc::UnboundedSender;
 use tokio::sync::{mpsc, oneshot};
 
+const MAX_DELAY: Duration = Duration::from_secs(60 * 60 * 24); // 1 day
+// These limits come from https://cloud.google.com/pubsub/quotas.
+const MAX_MESSAGES: u32 = 1000;
+const MAX_BYTES: u32 = 1e7 as u32; // 10MB
```
coryan, line 24: What happens when / if the service increases these values?  I guess we cut a new release, and otherwise the application cannot take advantage of the new limits?

coryan: Does it violate https://google.aip.dev/client-libraries/4223 ?

src/pubsub/src/publisher/publisher.rs
```diff
@@ -103,16 +109,31 @@ impl PublisherBuilder {
     }
 
     pub fn build(self) -> Publisher {
+        // Enforce limits by clamping the user-provided options.
+        let batching_options = BatchingOptions::new()
+            .set_delay_threshold(std::cmp::min(
+                self.batching_options.delay_threshold,
+                MAX_DELAY,
+            ))
```
coryan, line 117: Does this work? Is it any better?


```suggestion
            .set_delay_threshold(
                self.batching_options.delay_threshold.clamp(Duration::ZERO,
                MAX_DELAY,
            ))
```

coryan (approved):
Nothing blocking, a few questions to think about.

src/pubsub/src/publisher/publisher.rs
suzmue: Perhaps this link is better: https://cloud.google.com/pubsub/docs/batch-messaging#quotas_and_limits_on_batch_messaging. It says the high level client libraries ensure these limits are enforced. (will do in follow up because want to rebase other PR on top of this one).

src/pubsub/src/publisher/publisher.rs
suzmue: Yeah that also works! had that in a previous iteration of the PR. If durations can be negative we definitely want this.

PR #3704, opened 2025-11-04
impl(gax-internal): Construct TransportSpanInfo in HTTP transport

westarle:
- Add create_transport_span_info function to http_tracing.rs
- Update http.rs to use create_transport_span_info
- Add tests for TransportSpanInfo population in http_observability.rs

This for T3 (Client Request) spans.

### DIFF
--- a/src/gax-internal/src/observability/http_tracing.rs
+++ b/src/gax-internal/src/observability/http_tracing.rs
@@ -128,21 +128,15 @@ pub(crate) fn create_transport_span_info(
     response: &reqwest::Response,
     prior_attempt_count: u32,
 ) -> gax::response::internal::TransportSpanInfo {
-    let mut info = gax::response::internal::TransportSpanInfo::default();
-    info.request_resend_count = if prior_attempt_count > 0 {
-        Some(prior_attempt_count as i64)
-    } else {
-        None
-    };
-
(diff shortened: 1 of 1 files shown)


westarle:
@coryan for the equivalent on the `Error` path I might need to add `TransportSpanInfo` to gax `Error` corresponding to the field in `Response` with similar `pub mod internal` accessors.  WDYT?

coryan (approved):

coryan:
> @coryan for the equivalent on the `Error` path I might need to add `TransportSpanInfo` to gax `Error` corresponding to the field in `Response` with similar `pub mod internal` accessors. WDYT?

This is starting to look ugly.  I wonder if this is why the `tracing` library attaches the span to the future trait.  Maybe we can attach the span info to the future? like they do here:

https://docs.rs/tracing/latest/tracing/trait.Instrument.html

westarle:
> This is starting to look ugly. I wonder if this is why the `tracing` library attaches the span to the future trait. Maybe we can attach the span info to the future?

I will look into this -- I also think we might be able to use `Span::current()` directly and avoid `SpanInfo` if we can manage to have the right span in scope at the right time.

westarle:
Replaced with https://github.com/googleapis/google-cloud-rust/pull/3761

PR #3708, opened 2025-11-04
docs(pubsub): improve batching configuration docs

suzmue:
For #3707

### DIFF
--- a/src/pubsub/src/publisher/options.rs
+++ b/src/pubsub/src/publisher/options.rs
@@ -26,10 +26,10 @@
 ///     .set_delay_threshold(Duration::from_millis(20));
 /// ```
 ///
-/// Disable batching
+/// Set batch size to 1 to disable batching
 ///
 /// ```
-/// // Disable batching by setting batch size to 1.
+/// # use google_cloud_pubsub::options::publisher::BatchingOptions;
 /// let options = BatchingOptions::new()
 ///     .set_message_count_threshold(1_u32);
 /// ```
--- a/src/pubsub/src/publisher/publisher.rs
+++ b/src/pubsub/src/publisher/publisher.rs
@@ -108,7 +108,7 @@ impl PublisherBuilder {
     /// Configure message count and delay
     ///
     /// ```
-    ///  # async fn sample() -> anyhow::Result<()> {
+    /// # async fn sample() -> anyhow::Result<()> {
     /// # use google_cloud_pubsub::*;
     /// # use builder::publisher::ClientBuilder;
     /// # use client::PublisherFactory;


src/pubsub/src/publisher/options.rs
```diff
@@ -58,7 +67,8 @@ impl BatchingOptions {
     /// # Example
     /// ```
     /// # use google_cloud_pubsub::options::publisher::BatchingOptions;
-    /// let options = BatchingOptions::new().set_delay_threshold(std::time::Duration::from_millis(10));
+    /// let options =
+    ///     BatchingOptions::new().set_delay_threshold(std::time::Duration::from_millis(10));
```
coryan, line 80:
```suggestion
    /// # use std::time::Duration;
    /// let options =
    ///     BatchingOptions::new().set_delay_threshold(Duration::from_millis(10));
```

coryan: I am not sure we want to hide `use std::time::Duration`, but we can decide that later.

src/pubsub/src/publisher/publisher.rs
```diff
@@ -101,9 +101,28 @@ impl PublisherBuilder {
         }
     }
 
-    /// Change the message batching options.
+    /// Configure publisher batching behavior.
+    ///
+    /// To turn off batching, provide [BatchingOptions] with message_count_threshold
+    /// set to 1.
+    /// # Example
```
coryan:
```suggestion
    /// To turn off batching, provide [BatchingOptions] with message_count_threshold
    /// set to 1.
    ///
    /// # Example
```

coryan: Alternatively, you can create a `# Example: disable batching` section instead of describing how to disable batching.

coryan (approved):
A few nits. Maybe unneeded, check how the example render and then make a decision.

src/pubsub/src/publisher/options.rs
suzmue: I think it probably helps to show. Made it visible.

src/pubsub/src/publisher/publisher.rs
suzmue: Added disabling as a separate example.

"disable" is technically not accutate since it is possible that we will end up batching messages together in the case of ordering keys (if a batch cannot be sent because there are outstanding publishes for other ordering keys, but it is not "full" wrt the service maximums, we should add to the existing batch instead of creating new batches).

src/pubsub/src/publisher/publisher.rs
```diff
@@ -101,9 +101,50 @@ impl PublisherBuilder {
...
 
-    /// Change the message batching options.
+    /// Configure publisher batching behavior.
+    ///
+    /// # Examples
+    ///
+    /// Configure message count and delay
+    ///
+    /// ```
+    ///  # async fn sample() -> anyhow::Result<()> {
```
coryan: Nit:


```suggestion
    /// # async fn sample() -> anyhow::Result<()> {
```

src/pubsub/src/publisher/options.rs
```diff
@@ -12,9 +12,27 @@
...
+/// use std::time::Duration;
+/// let options = BatchingOptions::new()
+///     .set_message_count_threshold(100_u32)
+///     .set_delay_threshold(Duration::from_millis(20));
+/// ```
+///
+/// Disable batching
+///
+/// ```
+/// // Disable batching by setting batch size to 1.
```
coryan:
```suggestion
/// Set batch size to 1 to disable batching
///
/// ```
```

??

coryan (approved):

PR #3714, opened 2025-11-05
refactor(gax-internal): move test helpers

coryan:
Keeping them outside the crate simplifies version bumps.


dbolduc (approved):
FWIW, I think you already handled them for version bumps.

https://github.com/googleapis/librarian/blob/2c94a53f0fb5f8c76c4639721337f08453b658e8/internal/sidekick/internal/rust_release/find_cargo_manifests_test.go#L38

And there will still be things like `src/storage/examples` and `src/auth/integration-tests`.

I have no issues with the PR. Just sayin.

coryan:
> FWIW, I think you already handled them for version bumps.
> 
> https://github.com/googleapis/librarian/blob/2c94a53f0fb5f8c76c4639721337f08453b658e8/internal/sidekick/internal/rust_release/find_cargo_manifests_test.go#L38
> 
> And there will still be things like `src/storage/examples` and `src/auth/integration-tests`.
> 
> I have no issues with the PR. Just sayin.

You are right. Closing.

PR #3715, opened 2025-11-05
impl(storage): handle redirect errors

coryan:
In the bidi streaming protocol the stream may be closed to redirect the client to a new handle and routing header. This PR introduces a helper function to convert a `grpc::Status` to `gax::error::Error` **and** also update the read spec (part of the initial bidi read request) with any redirect data.

Parsing the error requires several storage-specific types. It seems better to keep this in the storage library.

Part of the work for #3626

### DIFF
--- a/src/storage/src/storage/bidi/redirect.rs
+++ b/src/storage/src/storage/bidi/redirect.rs
@@ -28,6 +28,7 @@ pub fn handle_redirect(
                 let mut guard = spec.lock().expect("never poisoned");
                 guard.routing_token = redirect.routing_token;
                 guard.read_handle = redirect.read_handle;
+                break;
             }
         }
     }


src/storage/src/storage.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#[allow(dead_code)]
```
dbolduc, line 15: I know there are no public interfaces, but did you want to introduce the `--cfg google_cloud_unstable_storage_bidi` gate now?

src/storage/src/storage/bidi/redirect.rs
```diff
@@ -0,0 +1,105 @@
...
+pub fn handle_redirect(
+    spec: Arc<Mutex<BidiReadObjectSpec>>,
+    status: tonic::Status,
+) -> crate::Error {
+    if let Ok(status) = RpcStatus::decode(status.details()) {
+        for d in status.details {
+            if let Ok(redirect) = d.to_msg::<BidiReadObjectRedirectedError>() {
+                let mut guard = spec.lock().expect("never poisoned");
+                guard.routing_token = redirect.routing_token;
+                guard.read_handle = redirect.read_handle;
```
dbolduc, line 30: `break`?

I assume we expect exactly one `BidiReadObjectRedirectedError`.

src/storage/src/storage/bidi/redirect.rs
```diff
@@ -0,0 +1,105 @@
...
+// limitations under the License.
+
+use crate::google::rpc::Status as RpcStatus;
+use crate::google::storage::v2::{BidiReadObjectRedirectedError, BidiReadObjectSpec};
+use gaxi::grpc::from_status::to_gax_error;
+use prost::Message;
+use std::sync::{Arc, Mutex};
+
+pub fn handle_redirect(
+    spec: Arc<Mutex<BidiReadObjectSpec>>,
```
dbolduc, line 22: this can be written by multiple threads?

are there N streams per spec? or something else is going on?

src/storage/src/storage/bidi/redirect.rs
dbolduc, line 30: Q: just checking: `None` is supposed to override the current value in `spec`?

What does `None` even mean?

src/storage/src/storage.rs
coryan: I think I will use that gate only when public APIs are needed, such as for integration tests.

src/storage/src/storage/bidi/redirect.rs
coryan: I expect zero contention in that mutex.  With that said,  the code I wrote has multiple writers. It is possible that code gets simplified and we can eliminate the Arc<Mutex<>> thing.

Basically two pieces of code will need to hold on to the spec and change it: the thing that reads from a stream, and then thing that reconnects a new stream *after* the connection succeeds.  Maybe I can refactor those two things to a single type and all of this goes away... Hmmm.

Let's merge this and I will refactor later if it makes sense.

src/storage/src/storage/bidi/redirect.rs
coryan: I assume that a redirect with a `None` means: "go back to the CFE and start over from just the bucket and object name".  Possibly it is a "can't happen" thing, but I figured we would handle the case.

coryan: Done. I don't think we expect more than one and we save a wee-bit of time.

coryan: @cjc25 in case he has opinions or thoughts.

cjc25: tl;dr: clearing in the request is fine if either field in the error is `None`.

The service should never set an empty routing token in practice, but if it does, then clearing the field in the request (and the `x-goog-request-params` header) is fine and we will eventually route to an endpoint that can serve the request.

The service should never set `read_handle` to empty in a redirect error if it already returned a `read_handle` in some other response. It _may_ return an empty `read_handle` if no response was ever delivered. Regardless, it is always safe, but slower, to send a read retry without a `read_handle` in the request, so clearing here if we hit that case is safe.

Issue #2808, opened 2025-08-01
Enable setter sample generation for all APIs


amanda-tarafa:

amanda-tarafa:
For Secret Manager and Workflows: https://github.com/googleapis/google-cloud-rust/pull/3369

amanda-tarafa:
Setter sample generation is now enabled by default. We are explicitly disabling for the following packages, as there were some build failures for those. I'll look into them.

- src/firestore/src/generated/gapic/.sidekick.toml
- src/generated/cloud/aiplatform/v1/.sidekick.toml
- src/generated/cloud/compute/v1/.sidekick.toml
- src/generated/cloud/speech/v2/.sidekick.toml
- src/generated/openapi-validation/.sidekick.toml
- src/protojson-conformance/src/generated/test_protos/.sidekick.toml
- src/pubsub/src/generated/gapic/.sidekick.toml
- src/storage/src/generated/gapic/.sidekick.toml
- src/storage/src/generated/gapic_control/.sidekick.toml
- src/wkt/src/generated/.sidekick.toml
- src/wkt/tests/common/src/generated/.sidekick.toml

amanda-tarafa:
And from @coryan:

You know this, but just recording my opinion. These are surprising:

- src/generated/cloud/speech/v2/.sidekick.toml
- src/generated/openapi-validation/.sidekick.toml

These are unsurprising, but need fixing.

- src/generated/cloud/aiplatform/v1/.sidekick.toml
- src/generated/cloud/compute/v1/.sidekick.toml
- src/pubsub/src/generated/gapic/.sidekick.toml
- src/storage/src/generated/gapic/.sidekick.toml
- src/storage/src/generated/gapic_control/.sidekick.toml

These we don't need to fix immediately:

- src/firestore/src/generated/gapic/.sidekick.toml
- src/wkt/src/generated/.sidekick.toml

These we don't need to fix ever:

- src/protojson-conformance/src/generated/test_protos/.sidekick.toml
- src/wkt/tests/common/src/generated/.sidekick.toml

amanda-tarafa:
Special cases fixed in googleapis/librarian#3179 with generation results in #4017 . I've tested #4017 locally wich is passing by running:

```
export RUSTDOCFLAGS='--cfg google_cloud_unstable_tracing --cfg google_cloud_unstable_id_token --cfg google_cloud_unstable_signed_url --cfg google_cloud_unstable_storage_bidi'
[...]

PR #3718, opened 2025-11-05
chore: bump all versions circa 2025-11-05

coryan:
Motivated by #2808

### DIFF
--- a/src/generated/api/apikeys/v2/.sidekick.toml
+++ b/src/generated/api/apikeys/v2/.sidekick.toml
@@ -17,6 +17,6 @@ specification-source = 'google/api/apikeys/v2'
 service-config = 'google/api/apikeys/v2/apikeys_v2.yaml'
 
 [codec]
-version        = '1.1.0'
+version        = '1.2.0'
 copyright-year = '2025'
 package-name-override = 'google-cloud-apikeys-v2'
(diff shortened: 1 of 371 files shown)


src/generated/api/apikeys/v2/.sidekick.toml
```diff
@@ -17,6 +17,6 @@ specification-source = 'google/api/apikeys/v2'
 service-config = 'google/api/apikeys/v2/apikeys_v2.yaml'
 
 [codec]
-version        = '1.1.0'
+version        = '1.2.0'
```
dbolduc, line 20: FWIW, adding samples seems like a patch bump to me. :shrug: 

If we don't want to do a patch bump because of the manual effort and likelihood of mistakes, that is fine with me.

dbolduc: Counterpoint: when services only make a documentation change, I would prefer to do a minor version bump over manually inspecting every googleapis SHA update + manually doing a patch bump.

This case is a little different in that we know which changes we made.

Just thinking out loud... I would approve this PR.

coryan: I went for "it is the easiest thing to do and it is a customer-visible change".

coryan:
Yup, we should redo this with the new samples.

PR #3720, opened 2025-11-05
refactor(pubsub): add closure for flushing a batch

suzmue:
Minor refactor to share code. Making the flush a closure instead of moving onto the batch type keeps the batch type a little simpler (it doesn't have to hold the client or topic).

### DIFF
--- a/src/pubsub/src/publisher/publisher.rs
+++ b/src/pubsub/src/publisher/publisher.rs
@@ -221,6 +221,10 @@ impl Worker {
         let delay = self.batching_options.delay_threshold;
         let message_limit = self.batching_options.message_count_threshold;
 
+        let flush = |b: Batch| {
+            tokio::spawn(b.send(self.client.clone(), self.topic_name.clone()));
+        };
+
         let timer = tokio::time::sleep(delay);
         // Pin the timer to the stack.
         tokio::pin!(timer);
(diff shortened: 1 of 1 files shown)


src/pubsub/src/publisher/publisher.rs
```diff
@@ -221,6 +221,10 @@ impl Worker {
         let delay = self.batching_options.delay_threshold;
         let message_limit = self.batching_options.message_count_threshold;
 
+        let flush = |b: Batch| {
+            tokio::spawn(b.send(self.client.clone(), self.topic_name.clone()));
+        };
```
coryan, line 226: optional:  consider:

```rust
impl Batch {
...
  fn flush(&mut self, client: GapicPublisher, topic: String) {
      let b = Self { messages: self.messages.drain(); }
      tokio::spawn(b.send(self.client.clone(), self.topic_name.clone()));
  }
```

then you can always write:

```rust
batch.flush()
```

coryan (approved):

src/pubsub/src/publisher/publisher.rs
suzmue: Nice! Going to submit as is, running in to some move errors in my next PR anyway so will likely be re-refactoring this :)

PR #3725, opened 2025-11-06
test(o11y): add auth interceptor for otlp integration

westarle:
## Goal
Implement the authentication component required for the OpenTelemetry OTLP exporter to securely communicate with the Google Cloud Telemetry API (`telemetry.googleapis.com`).

## Changes
*   Created the `observability` module in `integration-tests`, guarded by the `google_cloud_unstable_tracing` flag.
*   Implemented `CloudTelemetryAuthInterceptor`: A custom `tonic::service::Interceptor` that injects OAuth2 access tokens and the quota project ID into gRPC requests.
*   Implemented a background `tokio` task to asynchronously refresh credentials using `google-cloud-auth` and push updates to the interceptor via a `watch` channel.
*   Added comprehensive unit tests for the interceptor and the background refresh loop, including handling of `CacheableResource::NotModified` for efficiency.

## Context & Rationale
Standard OpenTelemetry OTLP exporters do not natively support Google Cloud authentication. We need a custom mechanism to inject the required headers.

Since `tonic` interceptors must be synchronous, we use a background task and a shared `watch` channel to bridge the gap with the asynchronous `google-cloud-auth` library, ensuring the interceptor never blocks the async runtime.

### DIFF
--- a/src/integration-tests/src/observability/auth.rs
+++ b/src/integration-tests/src/observability/auth.rs
@@ -13,16 +13,16 @@
 // limitations under the License.
 
 use auth::credentials::{CacheableResource, Credentials, EntityTag};
-use http::HeaderMap;
 use tokio::sync::watch;
 use tokio::time::{Duration, sleep};
+use tonic::metadata::MetadataMap;
 use tonic::service::Interceptor;
 use tonic::{Request, Status};
 
-const REFRESH_INTERVAL: Duration = Duration::from_secs(300); // 5 minutes
+const REFRESH_INTERVAL: Duration = Duration::from_secs(5);
 const ERROR_RETRY_DELAY: Duration = Duration::from_secs(10);
 
-/// A `tonic` interceptor that injects Google Cloud authentication headers into
+/// A [tonic] interceptor that injects Google Cloud authentication headers into
 /// every gRPC request.
 ///
 /// This is a special-purpose interceptor for the Cloud Telemetry API. We chose
@@ -33,13 +33,18 @@ const ERROR_RETRY_DELAY: Duration = Duration::from_secs(10);
 ///     hardcodes the underlying `Channel` type, making it difficult to inject
 ///     generic middleware. It does, however, provide a simple way to add a
 ///     `tonic` interceptor.
-/// 2.  **Sync/Async Bridge:** `tonic` interceptors must be synchronous. To
-///     handle asynchronous token refreshing, we use a background `tokio` task
-///     that pushes the latest valid headers to a shared `watch` channel. The
-///     interceptor then synchronously and instantly retrieves them.
+/// 3.  **Performance:** The background task pre-converts the `HeaderMap` into
+///     a `tonic::metadata::MetadataMap`. This ensures the critical path in the
+///     interceptor's `call` method is fast, as it only needs to clone the
+///     pre-converted metadata.
+///
+/// [tonic]: https://docs.rs/tonic/latest/tonic/service/trait.Interceptor.html
 #[derive(Clone)]
 pub struct CloudTelemetryAuthInterceptor {
-    rx: watch::Receiver<Option<HeaderMap>>,
+    // Our plan of record is to migrate to Google's gRPC at some point. However,
+    // this is for integration with `opentelemetry-otlp` which uses Tonic and may
+    // never migrate.
+    rx: watch::Receiver<Option<MetadataMap>>,
 }
 
 impl CloudTelemetryAuthInterceptor {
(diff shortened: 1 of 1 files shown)


src/integration-tests/src/observability/auth.rs
```diff
@@ -0,0 +1,322 @@
...
+        tokio::spawn(refresh_task(credentials, tx));
+        Self { rx }
+    }
+}
+
+impl Interceptor for CloudTelemetryAuthInterceptor {
+    fn call(&mut self, mut request: Request<()>) -> Result<Request<()>, Status> {
+        // Read the latest headers from the watch channel.
+        let rx_ref = self.rx.borrow();
+        if let Some(headers) = rx_ref.as_ref() {
```
coryan: optional:


```suggestion
        let headers = rx_ref.as_ref().ok_or_else(|| Err(Status::unauthenticated(...))?;
```

src/integration-tests/src/observability/auth.rs
```diff
@@ -0,0 +1,322 @@
...
+            for (name, value) in headers.iter() {
+                let key = tonic::metadata::MetadataKey::from_bytes(name.as_str().as_bytes())
+                    .map_err(|e| Status::internal(format!("invalid header name: {e}")))?;
+                let val = tonic::metadata::MetadataValue::try_from(value.as_bytes())
+                    .map_err(|e| Status::internal(format!("invalid header value: {e}")))?;
+                request.metadata_mut().insert(key, val);
+            }
```
coryan: Can we move this transformation to the background task? That was you do it once, when the new values appear?

westarle: done

src/integration-tests/src/observability/auth.rs
westarle: done

westarle:
> Cool stuff. Eventually this can become part of a helper library I assume?

Yes, please review as if it would be library code at some point.
 
> I think we can save some time in the critical path by transforming to `tonic::metadata::MetadataMap` only once in the background task.

It looks better now!

src/integration-tests/src/observability/auth.rs
```diff
@@ -0,0 +1,335 @@
...
+                        // (e.g., non-ASCII characters in values without -bin suffix).
+                    }
+                }
+
+                if tx.send(Some(metadata)).is_err() {
+                    // Receiver dropped (all interceptors are gone), stop task.
+                    break;
+                }
+                last_etag = Some(entity_tag);
+                sleep(REFRESH_INTERVAL).await;
```
coryan, line 118: The refresh interval seems too long. If the credentials have expired the requests will fail for 5 minutes?

src/integration-tests/src/observability/auth.rs
```diff
@@ -0,0 +1,335 @@
...
+use tokio::sync::watch;
+use tokio::time::{Duration, sleep};
+use tonic::metadata::MetadataMap;
+use tonic::service::Interceptor;
+use tonic::{Request, Status};
+
+const REFRESH_INTERVAL: Duration = Duration::from_secs(300); // 5 minutes
+const ERROR_RETRY_DELAY: Duration = Duration::from_secs(10);
+
+/// A `tonic` interceptor that injects Google Cloud authentication headers into
```
coryan: a nit:

```suggestion
/// A [tonic] interceptor that injects Google Cloud authentication headers into
```

And then define the link later:

```
/// [tonic]: https://....
```

coryan: I think assuming "tonic" here makes sense, but we should document the rationale. Something like:

```
// Our plan of record is to migrate to Google's gRPC at some point.  However, this is for integration with 
// `opentelemetry-otlp` which uses Tonic and may never migrate.
```

I am not sure if that should be a public or internal-only comment.

src/integration-tests/src/observability/auth.rs
coryan: Maybe make the time shorter for now, pinging this every 5 seconds should cost almost nothing.  

Possibly the API for our credentials cache is not good enough, and we need a way to know for how long the credentials are valid.

Alternatively, we may need to monitor for errors (in a different type of interceptor?) and invalidate the cache in that case?

coryan (changes requested):
The refresh interval seems too long. Otherwise, LGTM.

src/integration-tests/src/observability/auth.rs
westarle: Done.

@sai-sunder-s any thoughts on this interval or some hints from the cache?

src/integration-tests/src/observability/auth.rs
westarle: done

coryan (approved):

PR #3727, opened 2025-11-06
test(o11y): implement OTLP tracer provider

westarle:


westarle:
I plan to create an in-memory OTLP test server in another PR, to improve coverage.  Once I've got tracing -> otlp in place.

src/integration-tests/src/observability/otlp.rs
```diff
@@ -0,0 +1,131 @@
...
+// limitations under the License.
+
+use super::auth::CloudTelemetryAuthInterceptor;
+use auth::credentials::Credentials;
+use opentelemetry_otlp::{WithExportConfig, WithTonicConfig};
+use opentelemetry_sdk::trace::{SdkTracerProvider, TraceError};
+use tonic::transport::ClientTlsConfig;
+
+const GCP_OTLP_DOMAIN_NAME: &str = "telemetry.googleapis.com";
+const GCP_OTLP_ENDPOINT: &str = "https://telemetry.googleapis.com:443";
```
coryan: Is the `:443` needed?  That seems surprising given that it is `https`?

coryan: Does this work?


```suggestion
const GCP_OTLP_DOMAIN_NAME: &str = "telemetry.googleapis.com";
const GCP_OTLP_ENDPOINT: &str = concat!("https://", GCP_OTLP_DOMAIN_NAME);
```

src/integration-tests/src/observability/otlp.rs
```diff
@@ -0,0 +1,131 @@
...
+/// Returns a [`TraceError`] if:
+/// * The OTLP exporter cannot be built (e.g., TLS configuration fails).
+///
+/// # Examples
+///
+/// ```no_run
+/// use integration_tests::observability::otlp::init_tracer_provider;
+/// use google_cloud_auth::credentials::Credentials;
+///
+/// # async fn example() -> Result<(), Box<dyn std::error::Error>> {
```
coryan: I think we started using:


```suggestion
/// # async fn example() -> anyhow::Result<()> {
```

May require adding `anyhow` to the dev dependencies.

src/integration-tests/src/observability/otlp.rs
```diff
@@ -0,0 +1,131 @@
...
+
+const GCP_OTLP_DOMAIN_NAME: &str = "telemetry.googleapis.com";
+const GCP_OTLP_ENDPOINT: &str = "https://telemetry.googleapis.com:443";
+const OTEL_KEY_GCP_PROJECT_ID: &str = "gcp.project_id";
+const OTEL_KEY_SERVICE_NAME: &str = "service.name";
+
+/// Configures and initializes an OpenTelemetry `SdkTracerProvider` optimized for Google Cloud Trace.
+///
+/// This function sets up a complete OTLP export pipeline using gRPC with TLS, specifically
+/// targeting the Google Cloud Trace endpoint (`telemetry.googleapis.com`). It automatically
```
coryan: Would we want to make that configurable for VPC-SC, TPC, or REP?  I think we should use a builder to initialize this thing, even if we hard-code a few parameters for now.  With a builder it is easier to extend the configuration later.

src/integration-tests/src/observability/otlp.rs
westarle: done

westarle: done

src/integration-tests/src/observability/otlp.rs
westarle: Done

PR #3728, opened 2025-11-06
feat(pubsub): add `flush` to the publisher

suzmue:
This adds a flush command to the Publisher, providing a way for users to force all messages to send without waiting for batching limits. This is implemented by sending the flush request through the same channel as messages are requested, ensuring that any publish request received before flush are included in the flush.

For #3015

### DIFF
--- a/src/pubsub/src/publisher/publisher.rs
+++ b/src/pubsub/src/publisher/publisher.rs
@@ -341,7 +341,7 @@ impl Worker {
                             // also need to keep track of any pending flushes, and make sure
                             // all of those resolve as well.
                             let mut flushing = std::mem::take(&mut inflight);
-                            while let Some(_) = flushing.next().await {}
+                            while flushing.next().await.is_some() {}
                             let _ = tx.send(());
                         },
                         None => {


src/pubsub/src/publisher/publisher.rs
```diff
@@ -62,11 +64,60 @@ impl Publisher {
...
+    /// This method is `async` and returns only after all publish attempts for the
+    /// messages in the snapshot have completed. A "completed" attempt means the
+    /// message has either been successfully sent, or has failed permanently after
+    /// exhausting any applicable retry policies.
+    ///
+    /// After flush()` returns, the final result of each individual publish
+    /// operation (i.e., a success with a message ID or a terminal error) will
+    /// be available on its corresponding [PublishHandle](crate::model_ext::PublishHandle).
+    ///
+    /// Messages published after `flush()` is called will be buffered for a
```
coryan, line 93: nit:


```suggestion
    /// Messages published after calling `flush()` are be buffered for a
```

src/pubsub/src/publisher/publisher.rs
```diff
@@ -62,11 +64,60 @@ impl Publisher {
...
         }
         crate::model_ext::PublishHandle { rx }
     }
+
+    /// Flushes all outstanding messages.
+    ///
+    /// This method sends any messages that have been published but not yet sent,
+    /// regardless of the configured batching options (`delay_threshold`, etc.).
+    ///
+    /// This method is `async` and returns only after all publish attempts for the
```
coryan, line 84: "returns" is not quite right, "the future is ready"?

src/pubsub/src/publisher/publisher.rs
```diff
@@ -297,6 +380,27 @@ impl Batch {
...
+    /// Drains the batch and spawns a task to send the messages.
+    ///
+    /// This method mutably drains the messages from the current batch, leaving it
+    /// empty, and returns a `JoinHandle` for the spawned send operation. This allows
+    /// the `Worker` to immediately begin creating a new batch while the old one is
+    /// being sent in the background.
+    fn flush(
+        &mut self,
+        client: GapicPublisher,
+        topic: String,
```
coryan, line 403: note: most of (all?) the methods in this type get that `client` and `topic` parameters, that suggest they should be part of `self` ?

src/pubsub/src/publisher/publisher.rs
```diff
@@ -216,50 +275,74 @@ impl Worker {
...
                         },
+                        Some(ToWorker::Flush(tx)) => {
+                            batch.flush(self.client.clone(), self.topic_name.clone(), &mut inflight);
+                            // Wait on all the tasks that exist right now.
+                            // We could instead tokio::spawn this as well so the publisher
+                            // can keep working on additional messages. The worker would
+                            // also need to keep track of any pending flushes, and make sure
+                            // all of those resolve as well.
+                            let mut flushing = std::mem::take(&mut inflight);
+                            while flushing.next().await.is_some() {}
```
coryan, line 344: I spent too much time thinking about the tradeoffs between this vs. `futures::future::join_all()`.  Instead of wasting more of your time: please consider that as an alternative. If you did already, let me know why this is better?

src/pubsub/src/publisher/publisher.rs
```diff
@@ -62,11 +64,60 @@ impl Publisher {
 
         // If this fails, the worker is gone, which indicates something bad has happened.
         // The PublishHandle will automatically receive an error when `tx` is dropped.
-        if self.tx.send(BundledMessage { msg, tx }).is_err() {
+        if self
+            .tx
+            .send(ToWorker::Publish(BundledMessage { msg, tx }))
+            .is_err()
+        {
```
coryan, line 73: For a future PR: Maybe we should log these, or add metrics / counters for them.

suzmue: Ack. Good idea.

src/pubsub/src/publisher/publisher.rs
suzmue: Fair point. I've got a few refactors in mind so will include this then.

src/pubsub/src/publisher/publisher.rs
suzmue: I looked into it a bit, and based on the join_all documentation (https://docs.rs/futures/latest/futures/future/fn.join_all.html) FuturesOrdered and FuturesUnordered have more performance optimizations that make them better for handling a large number of futures so I would be inclined to leaving it like this instead of `join_all(inflight.collect())` (or whatever the correct code may be).

If it turns out join_all is better we can always change it over.

src/pubsub/src/publisher/publisher.rs
suzmue: switched to "completes"

src/pubsub/src/publisher/publisher.rs
coryan: SG, maybe a comment or update one of the design docs.

PR #3731, opened 2025-11-06
docs(pubsub): clean up documentation for the builders

suzmue:
For #3707

### DIFF
--- a/src/pubsub/src/publisher/client.rs
+++ b/src/pubsub/src/publisher/client.rs
@@ -14,7 +14,7 @@
 
 use crate::publisher::publisher::PublisherBuilder;
 
-/// A factory for creating [`Publisher`](super::publisher::Publisher) instances.
+/// Creates [`Publisher`](super::publisher::Publisher) instances.
 ///
 /// This is the main entry point for the publisher API. A single `PublisherFactory`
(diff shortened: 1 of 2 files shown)


src/pubsub/src/publisher/client.rs
```diff
@@ -14,7 +14,32 @@
 
 use crate::publisher::publisher::PublisherBuilder;
 
-/// Client for publishing messages to Pub/Sub topics.
+/// A factory for creating [`Publisher`](super::publisher::Publisher) instances.
```
coryan:
```suggestion
/// Creates [`Publisher`](super::publisher::Publisher) instances.
```

src/pubsub/src/publisher/publisher.rs
```diff
@@ -69,7 +71,11 @@ impl Publisher {
     }
 }
 
-/// A builder for [Publisher].
+/// A builder for creating a `Publisher`.
```
coryan:
```suggestion
/// A builder to create `Publisher`s.
```

Or just "Creates `Publisher` instances" as before?

coryan (approved):
Small nits, consider fixing them before merging.

PR #3738, opened 2025-11-07
fix(auth): SA sourced id token response uses OAuth2 format

alvarowolfx:
The naive JWT parsing was missing the fact that the OAuth endpoint returns a JSON with `id_token` field, not the JWT token as text. 

Towards #3449


alvarowolfx:
this issue was found by doing the changes on PR  #3739

Issue #3638, opened 2025-10-28
Create a preview release for pubsub


suzmue:
This should be called 0.1.0-preview.

Targeting the mid-November client library release. This is targeting the Publisher.

Tasks to complete before publishing:

coryan:
Closing, we created this release.

PR #3745, opened 2025-11-09
docs(pubsub): add module level docs

suzmue:
Add module level docs for all modules where they were missing. This skips options since that is about to be deleted.

To add module level documentation for the admin clients, we add them individually to lib.rs.

For #3638

### DIFF
--- a/src/pubsub/src/lib.rs
+++ b/src/pubsub/src/lib.rs
@@ -48,29 +48,42 @@ pub(crate) mod publisher;
 pub use gax::Result;
 pub use gax::error::Error;
 
+/// Request and client builders.
 pub mod builder {
-    pub use crate::generated::gapic::builder::*;
+    /// Request and client builders for the [Publisher][crate::client::Publisher] client.
     pub mod publisher {
         #[doc(hidden)]
         pub use crate::generated::gapic_dataplane::builder::publisher::*;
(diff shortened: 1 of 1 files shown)


src/pubsub/src/lib.rs
```diff
@@ -48,32 +48,46 @@ pub(crate) mod publisher;
...
+    pub use crate::generated::gapic::builder::topic_admin;
 }
+
+/// The messages and enums that are part of this client library.
 pub mod model {
     pub use crate::generated::gapic::model::*;
     pub use crate::generated::gapic_dataplane::model::*;
 }
 
+/// Extends [model] with types that improve type safety and/or ergonomics.
```
coryan, line 74: Something does not read right here...  ergonomics is a noun, and "type safety" sounds like and adjective to me... In any case, better than before but maybe needs another copy-edit pass.

coryan (dismissed):

src/pubsub/src/lib.rs
suzmue: SG. This one was copied directly from storage so both should be updated when we do so.

dbolduc: > "type safety" sounds like an adjective to me

FWIW, I think it is a noun and the adjective would be "type safe"

PR #3747, opened 2025-11-09
docs(pubsub): add complete publishing example to client module

suzmue:
This example shows how to use the pubsub API to publish messages.

For #3707

### DIFF
--- a/src/pubsub/src/lib.rs
+++ b/src/pubsub/src/lib.rs
@@ -103,11 +103,6 @@ pub mod model_ext {
 ///     handles.push(publisher.publish(msg));
 /// }
 ///
-/// // Optionally flush the messages from the publisher. This bypasses
-/// // the configured batching settings to send the queued messages
-/// // as soon as possible.
-/// publisher.flush().await;
-///
 /// // The handles are futures that resolve to the server-assigned message IDs.
 /// // You can await them to get the results. Messages will still be sent even
 /// // if the handles are dropped.


src/pubsub/src/lib.rs
```diff
@@ -66,6 +66,47 @@ pub mod model_ext {
...
+/// let mut handles = Vec::new();
+/// for i in 0..10 {
+///     let msg = PubsubMessage::new().set_data(format!("message {}", i));
+///     handles.push(publisher.publish(msg));
+/// }
+///
+/// // Optionally flush the messages from the publisher. This bypasses
+/// // the configured batching settings to send the queued messages
+/// // as soon as possible.
+/// publisher.flush().await;
```
coryan: nit: I would leave that for a separate example or just let the method example speak for itself. Feel free to do that in a future PR.

src/pubsub/src/lib.rs
```diff
@@ -66,6 +66,47 @@ pub mod model_ext {
...
+/// ```
+/// # async fn sample() -> anyhow::Result<()> {
+/// use google_cloud_pubsub::client::PublisherFactory;
+/// use google_cloud_pubsub::model::PubsubMessage;
+///
+/// // Create a factory for creating publishers.
+/// let factory = PublisherFactory::builder().build().await?;
+///
+/// // Create a publisher that handles batching for a specific topic.
+/// let publisher = factory.publisher("projects/my-project/topics/my-topic").build();
```
coryan, line 96: For a future PR: I think we want a variant where we do:

```suggestion
/// let publisher: BatchedPublisher =
///     Publisher::builder("projects/my-project/topics/my-topic").build().await?;
```

And separately:

```rust
/// let publisher: Publisher = Publisher::shared().build().await?;
/// let batched: BatchedPublisher =
///     publisher.builder("projects/my-project/topics/my-topic").build();
```

and someday:

```rust
/// let publisher: Publisher = Publisher::shared().build().await?;
/// publisher.unbatched_publish("projects/my-project/topics/my-topic", message).send().await?;
```

src/pubsub/src/lib.rs
suzmue: I'll drop it in this one. Need to rebase anyway.

PR #3750, opened 2025-11-10
docs(pubsub): add example to PublishHandle

suzmue:
Add an example to the `PublishHandle` and document the output type to indicated the value of string is a message id.

For #3707

### DIFF
--- a/src/pubsub/src/publisher/model_ext.rs
+++ b/src/pubsub/src/publisher/model_ext.rs
@@ -36,8 +36,8 @@ use tokio::sync::oneshot;
 ///
 /// // The handle can be awaited later to get the result.
 /// match handle.await {
-///     Ok(message_id) => println!("Message published with ID: {}", message_id),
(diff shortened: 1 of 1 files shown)


src/pubsub/src/publisher/model_ext.rs
```diff
@@ -22,11 +22,34 @@ use tokio::sync::oneshot;
...
+/// # use google_cloud_pubsub::client::Publisher;
+/// # use google_cloud_pubsub::model::PubsubMessage;
+/// # async fn sample(publisher: Publisher) -> anyhow::Result<()> {
+/// // publish() returns a handle immediately.
+/// let handle = publisher.publish(PubsubMessage::new().set_data("hello world"));
+///
+/// // The handle can be awaited later to get the result.
+/// match handle.await {
+///     Ok(message_id) => println!("Message published with ID: {}", message_id),
+///     Err(e) => eprintln!("Failed to publish message: {}", e),
```
dbolduc: nit: use `{e:?}` for full error

PR #3756, opened 2025-11-10
test(o11y): implement tracing subscriber helper

westarle:


src/integration-tests/src/observability/tracing.rs
```diff
@@ -0,0 +1,137 @@
...
+
+/// The instrumentation scope name for the Google Cloud Rust client libraries.
+/// This identifies spans generated by these libraries in the exported telemetry.
+const INSTRUMENTATION_SCOPE_NAME: &str = "google-cloud-rust";
+
+/// Creates a `tracing` layer that exports spans to the provided OpenTelemetry `SdkTracerProvider`.
+///
+/// The layer is filtered by the `RUST_LOG` environment variable (defaulting to "info").
+///
+/// # Arguments
```
coryan: nit: I think Rust calls these "Parameters", and we have been doing that in our docs.

It would be useful if you said something about the `S` type parameter.

src/integration-tests/src/observability/tracing.rs
```diff
@@ -0,0 +1,137 @@
...
+///
+/// * `tracer_provider` - The configured OpenTelemetry `SdkTracerProvider`.
```
coryan: maybe "the destination for the tracing spans"?

src/integration-tests/src/observability/tracing.rs
westarle: done

src/integration-tests/src/observability/tracing.rs
westarle: done

PR #3760, opened 2025-11-10
impl(storage): bidi retry and resume decorators

coryan:
I need some retry and resume decorators to deal with redirect errors. Notably, these decorators ignore permanent errors (and **only** permanent errors) as defined by the inner policy. The application will always be able to set the inner policy. The decorator will be harcoded in the client.

Part of the work for #3626

### DIFF
--- a/src/storage/src/storage/bidi/redirect.rs
+++ b/src/storage/src/storage/bidi/redirect.rs
@@ -156,6 +156,8 @@ mod tests {
         )),
         false
     )]
+    #[test_case(deep_redirect("r2", 4), true)]
+    #[test_case(deep_redirect("r2", 64), false)]
     fn redirect(input: Error, want: bool) {
         assert_eq!(is_redirect(&input), want, "{input:?}");
     }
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/bidi/resume_redirect.rs
```diff
@@ -0,0 +1,78 @@
...
+use gax::retry_result::RetryResult;
+use std::sync::Arc;
+
+/// Decorate the resume policy to continue on redirect errors.
+///
+/// The bidi streaming read API uses errors to redirect requests. We want to
+/// ignore these errors in the retry loop while respecting any limits set by the
+/// application.
+///
+/// The client library uses this policy to decorate whatever policy set by the
```
dbolduc: s/whatever/any/

src/storage/src/storage/bidi/retry_redirect.rs
```diff
@@ -0,0 +1,97 @@
...
+use gax::throttle_result::ThrottleResult;
+use std::sync::Arc;
+
+/// Decorate the retry policy to continue on redirect errors.
+///
+/// The bidi streaming read API uses errors to redirect requests. We want to
+/// ignore these errors in the retry loop while respecting any limits set by the
+/// application.
+///
+/// The client library uses this policy to decorate whatever policy set by the
```
dbolduc: s/whatever/any/

src/storage/src/storage/bidi/redirect.rs
```diff
@@ -35,8 +38,31 @@ pub fn handle_redirect(
...
 }
 
+/// Determine if an error is a redirect error.
+pub fn is_redirect(error: &Error) -> bool {
+    if error.status().is_none_or(|s| s.code != Code::Aborted) {
+        return false;
+    }
+    let Some(status) = error
+        .status()
+        .and_then(|_| error.source())
```
dbolduc: Q: should this be a loop?

dbolduc (approved):

src/storage/src/storage/bidi/redirect.rs
coryan: Yes, fixed.

coryan:
PTAL

PR #3761, opened 2025-11-10
feat(gax-internal): record transport attributes in HTTP transport

westarle:
This change implements the enrichment of the Client Request Span (T3) with transport-level attributes derived from the final HTTP attempt.

*   Adds `record_intermediate_client_request` to `observability/http_tracing.rs`.
    This function uses `tracing::Span::current()` to access the active T3 span
    and records attributes such as `server.address`, `server.port`, `url.full`,
    `http.response.status_code`, and `http.request.resend_count`.
*   Includes a safety check to only enrich spans marked with `gax.client.span`,
    preventing accidental pollution of user-defined spans.
*   Updates `ReqwestClient::execute` in `http.rs` to call this enrichment function
    upon receiving a response (success or failure) before the response body is
    consumed and converted into a `gax::Response` or `gax::error::Error`.

The T3 span will be created and finalized (setting `otel.status_code` and `error.type`) in the generated client code just before returning the final `gax::Response` (replacing current tracing macro in clients.)

### DIFF
--- a/src/gax-internal/src/observability/http_tracing.rs
+++ b/src/gax-internal/src/observability/http_tracing.rs
@@ -137,11 +137,10 @@ pub(crate) fn record_intermediate_client_request(
 
     // Only enrich spans that are explicitly marked as GAX client spans.
     // This prevents accidental enrichment of user-provided spans that happen to have the same fields.
-    if let Some(metadata) = span.metadata() {
-        if metadata.fields().field("gax.client.span").is_none() {
-            return;
-        }
-    } else {
+    if span
+        .metadata()
+        .is_none_or(|m| m.fields().field("gax.client.span").is_none())
+    {
         return;
     }
 


src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -122,6 +122,65 @@ pub(crate) fn record_http_response_attributes(
...
+
+    // Only enrich spans that are explicitly marked as GAX client spans.
+    // This prevents accidental enrichment of user-provided spans that happen to have the same fields.
+    if let Some(metadata) = span.metadata() {
+        if metadata.fields().field("gax.client.span").is_none() {
+            return;
+        }
+    } else {
+        return;
+    }
```
coryan:
```suggestion
    if span.metadata().is_none_or(|m| m.fields().field("gax.client.span").is_none()) {
        return;
    }
```

westarle: done

PR #3763, opened 2025-11-10
impl(auth): id token adc flow should allow including email

alvarowolfx:
Allow to specify if customers want to include email in the token claims when using ADC flow. This will enable email claim on sources that have to be set up for that, like MDS and Impersonated flows.

Towards #3449

### DIFF
--- a/src/auth/integration-tests/tests/driver.rs
+++ b/src/auth/integration-tests/tests/driver.rs
@@ -102,7 +102,7 @@ mod driver {
     #[cfg(all(test, google_cloud_unstable_id_token))]
     #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
     #[serial_test::serial]
-    // verify that include_email via ADC flow is passed down to the impersonated 
+    // verify that include_email via ADC flow is passed down to the impersonated
     // builder and email claim is included in the token.
     async fn run_id_token_adc_impersonated() -> anyhow::Result<()> {
         let with_impersonation = true;


src/auth/src/credentials/idtoken.rs
```diff
@@ -193,9 +194,16 @@ impl Builder {
     pub fn new<S: Into<String>>(target_audience: S) -> Self {
         Self {
             target_audience: target_audience.into(),
+            include_email: false,
         }
     }
 
+    /// Sets whether the ID token should include the `email` claim of the user in the token.
+    pub fn with_include_email(mut self, include_email: bool) -> Self {
```
coryan: There are no tests covering this new function. Probably trivial, but please fix.

src/auth/src/credentials/idtoken.rs
```diff
@@ -214,18 +222,20 @@ impl Builder {
...
 fn build_id_token_credentials(
     audience: String,
+    include_email: bool,
     json: Option<Value>,
 ) -> BuildResult<IDTokenCredentials> {
     match json {
         None => {
             // TODO(#3587): pass context that is being built from ADC flow.
-            mds::Builder::new(audience).with_format("full").build()
+            let format = if include_email { "full" } else { "standard" };
```
sai-sunder-s: is email the only difference between standard and full?

src/auth/src/credentials/idtoken.rs
```diff
@@ -234,9 +244,9 @@ fn build_id_token_credentials(
                     "{cred_type}, use idtoken::user_account::Builder directly."
                 ))),
                 "service_account" => service_account::Builder::new(audience, json).build(),
```
sai-sunder-s: does service account provide email always?

src/auth/src/credentials/idtoken.rs
```diff
@@ -234,9 +244,9 @@ fn build_id_token_credentials(
...
-                "impersonated_service_account" => {
-                    impersonated::Builder::new(audience, json).build()
-                }
+                "impersonated_service_account" => impersonated::Builder::new(audience, json)
+                    .with_include_email(include_email)
+                    .build(),
                 "external_account" => {
```
sai-sunder-s, line 286: email for byoid?

src/auth/src/credentials/idtoken.rs
alvarowolfx: yes

src/auth/src/credentials/idtoken.rs
alvarowolfx: idtokens for byoid is not supported

src/auth/src/credentials/idtoken.rs
alvarowolfx: No, there is more information being returned. From the docs:
```
the optional parameter that specifies whether or not the project and instance details are included in the payload. Specify full to include this information in the payload or standard to omit the information from the payload. The default value is standard. For more information, see [Identity token format](https://docs.cloud.google.com/compute/docs/instances/verifying-instance-identity#token_format).
```

The docs don't talk about the email claim, but I have tested on a VM and the `email` claim is only there with format=full.

Ref: https://docs.cloud.google.com/compute/docs/instances/verifying-instance-identity

src/auth/src/credentials/idtoken.rs
alvarowolfx: added tests for it. And @sai-sunder-s and I had some discussions that setting `with_include_email(false)` would be weird, since in some sources (e.g. service account) you can't disable that the `email` is going to be included.

alvarowolfx:
will be updated after some PR like #3792 and #3789

src/auth/integration-tests/tests/driver.rs
```diff
@@ -95,7 +95,16 @@ mod driver {
...
     async fn run_id_token_adc() -> anyhow::Result<()> {
-        auth_integration_tests::unstable::id_token_adc().await
+        let with_impersonation = false;
+        auth_integration_tests::unstable::id_token_adc(with_impersonation).await
+    }
+
+    #[cfg(all(test, google_cloud_unstable_id_token))]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    #[serial_test::serial]
+    async fn run_id_token_adc_impersonated() -> anyhow::Result<()> {
```
coryan, line 107: These changes are about impersonated credentials, but the PR title is about emails?

src/auth/src/credentials/idtoken.rs
```diff
@@ -193,9 +194,19 @@ impl Builder {
     pub fn new<S: Into<String>>(target_audience: S) -> Self {
         Self {
             target_audience: target_audience.into(),
+            include_email: false,
         }
     }
 
+    /// Sets whether the ID token should include the `email` claim of the user in the token.
+    /// For some credentials sources like Metadata Server and Impersonated Credentials, the default is
```
coryan, line 203:
```suggestion
    /// Sets whether the ID token should include the `email` claim of the user in the token.
    ///
    /// For some credentials sources like Metadata Server and Impersonated Credentials, the default is
```

src/auth/src/credentials/idtoken.rs
```diff
@@ -424,11 +446,67 @@ pub(crate) mod tests {
...
+
+    #[tokio::test]
+    #[parallel]
+    async fn test_build_id_token_include_email_mds() -> TestResult {
+        let audience = "test_audience".to_string();
+
+        // Test with include_email = true and no source credentials (MDS Fallback)
+        let creds_true = build_id_token_credentials(audience.clone(), true, None)?;
+        let fmt = format!("{:?}", creds_true.inner);
+        assert!(fmt.contains("MDSCredentials"));
```
coryan: yuck, but I have done the same in other places.

src/auth/src/credentials/idtoken.rs
```diff
@@ -424,11 +446,67 @@ pub(crate) mod tests {
...
+        assert!(fmt.contains("format: Some(Full)"));
```
coryan: maybe too much yuck...  You can fix this, but requires non-trivial coding:

- Break `build_id_token_credentials()` into two functions:
- `build_id_token_credentials_internal() -> BuildResult<EnumWithVariantsForEachCredentialType>` where you can examine the credential provider and its contents.
- `build_id_token_credentials()` calls the `_internal()` version and wraps it.

Thoughts?

src/auth/integration-tests/tests/driver.rs
alvarowolfx: the PR is about allowing customer to force including the `email` claim in the ID Token when using the ID Token ADC Builder, and that requires to pass down to Impersonated and MDS sources to include such claim. That's why I added the test here with ADC that resolves to an impersonated source

coryan: The maybe a comment:

```rust
// Verify email claims work.
//
// Impersonated credentials with MDS sources require an email claim.
```

src/auth/src/credentials/idtoken.rs
alvarowolfx: I liked the idea and just pushed an update doing that. Changes were not that bad.

Issue #3418, opened 2025-09-26
Feature: Add Client-Side Tracing for Unary gRPC Network Requests


westarle:
Thanks for stopping by to let us know something could be better!

**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.

**Is your feature request related to a problem? Please describe.**
Currently, the `google-cloud-rust` client libraries do not provide tracing for the underlying unary gRPC network requests. This makes it difficult to get a complete picture of request latency and to diagnose
network-level issues, as only the higher-level client operation is traced. We need visibility into each gRPC attempt.

**Describe the solution you'd like**
Implement client-side tracing for unary gRPC requests made by the client libraries. This involves creating a span for each gRPC network attempt.

The span should include:
*   Standard RPC attributes (`rpc.system`, `rpc.service`, `rpc.method`, `net.peer.name`, `net.peer.port`).
*   gRPC specific attributes (`rpc.grpc.status_code`).
*   Error information on failure.
*   Client library-specific attributes (`gcp.client.service`, `gcp.client.version`, etc.) to identify the calling library.

This will be achieved by adding a custom Tonic interceptor (`tower::Layer`) within the `google-cloud-gax-internal` crate. Tracing for streaming gRPC calls is out of scope for this issue.

**Describe alternatives you've considered**
1.  **Using existing Tonic tracing crates:** Existing crates lack the flexibility to add our custom `gcp.client.*` attributes.
2.  **Modifying Tonic directly:** Upstreaming changes would be time-consuming and might not align with Tonic's goals.
3.  **Passing all metadata via request extensions:** This would require more extensive changes to the code generator and add complexity.

The chosen interceptor approach allows for targeted changes within our GAX layer.

**Additional context**
This work is part of a larger effort to enhance observability within the Google Cloud client libraries for Rust. This specific feature request addresses the network span requirement for unary gRPC.

haphungw:
@westarle can we close this yet?

PR #3766, opened 2025-11-11
impl(o11y): add basic gRPC tracing infrastructure

westarle:
This PR establishes the foundational infrastructure for tracing of gRPC requests in the `google-cloud-rust` client libraries.

**Changes in this PR:**
*   Introduces `TracingTowerLayer` and `TracingTowerService` in `google-cloud-gax-internal`.
*   Integrates this layer into the `grpc::Client` construction pipeline.
*   Refactors `Client::make_inner` to conditionally apply the tracing layer based on configuration, using `tower::util::Either` to avoid runtime overhead when tracing is disabled.
*   Adds a basic integration test ensuring the layer is active when configured.

**Next Steps:**
*  URI Parsing and Standard RPC Attributes (`rpc.service`, `rpc.method`, etc.).
*  Client Library Metadata (`gcp.client.*`).
*  Response Status Handling (mapping `tonic::Code` to OTel status).
*  Cancellation and Edge Cases.

For #3418

### DIFF
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -33,21 +33,16 @@ use std::sync::Arc;
 use std::time::Duration;
 
 #[cfg(not(google_cloud_unstable_tracing))]
-pub type InnerClient = tonic::client::Grpc<tonic::transport::Channel>;
+pub type GrpcService = tonic::transport::Channel;
 
 #[cfg(google_cloud_unstable_tracing)]
-/// The inner gRPC client type.
-///
-/// When tracing is enabled (via the `google_cloud_unstable_tracing` config flag),
-/// this type uses `tower::util::Either` to dynamically dispatch between a
-/// traced client (wrapped with `TracingTowerService`) and a raw client. This allows
-/// tracing to be enabled or disabled at runtime based on configuration.
-pub type InnerClient = tonic::client::Grpc<
-    tower::util::Either<
-        crate::observability::grpc_tracing::TracingTowerService<tonic::transport::Channel>,
-        tonic::transport::Channel,
-    >,
+pub type GrpcService = tower::util::Either<
+    crate::observability::grpc_tracing::TracingTowerService<tonic::transport::Channel>,
+    tonic::transport::Channel,
 >;
+
+/// The inner gRPC client type.
+pub type InnerClient = tonic::client::Grpc<GrpcService>;
 
 #[derive(Clone, Debug)]
 pub struct Client {
--- a/src/gax-internal/src/observability.rs
+++ b/src/gax-internal/src/observability.rs
@@ -20,7 +20,7 @@
 #[cfg(google_cloud_unstable_tracing)]
 pub mod attributes;
 
-#[cfg(google_cloud_unstable_tracing)]
+#[cfg(all(google_cloud_unstable_tracing, feature = "_internal-http-client"))]
 mod errors;
 
 #[cfg(all(google_cloud_unstable_tracing, feature = "_internal-http-client"))]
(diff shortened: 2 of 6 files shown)


src/gax-internal/src/observability.rs
```diff
@@ -23,10 +23,13 @@ pub mod attributes;
...
 mod http_tracing;
 
-#[cfg(google_cloud_unstable_tracing)]
+#[cfg(all(google_cloud_unstable_tracing, feature = "_internal-http-client"))]
 pub(crate) use http_tracing::{
     create_http_attempt_span, record_http_response_attributes, record_intermediate_client_request,
 };
+
+#[cfg(all(google_cloud_unstable_tracing, feature = "_internal-grpc-client"))]
+pub mod grpc_tracing;
```
coryan, line 35: I think you need to change `lib.rs` too:

https://github.com/googleapis/google-cloud-rust/blob/467f409f3f3a2283cb527ed0e902b96e7cd9b057/src/gax-internal/src/lib.rs#L39-L40

Otherwise only clients that turn on `_internal-http-client` *and* `_internal-grpc-client` will get this new thing.

src/gax-internal/tests/grpc_observability.rs
```diff
@@ -0,0 +1,73 @@
...
+            .await?;
+
+        let spans = TestLayer::capture(&guard);
+        let grpc_spans: Vec<_> = spans.iter().filter(|s| s.name == "grpc.request").collect();
+        assert_eq!(
+            grpc_spans.len(),
+            1,
+            "Should capture one grpc.request span: {:?}",
+            grpc_spans
+        );
```
coryan, line 69: Can we say something about the contents of the span?

src/gax-internal/src/grpc.rs
```diff
@@ -32,8 +32,23 @@ use http::HeaderMap;
...
 
+#[cfg(google_cloud_unstable_tracing)]
+/// The inner gRPC client type.
+///
+/// When tracing is enabled (via the `google_cloud_unstable_tracing` config flag),
+/// this type uses `tower::util::Either` to dynamically dispatch between a
+/// traced client (wrapped with `TracingTowerService`) and a raw client. This allows
+/// tracing to be enabled or disabled at runtime based on configuration.
+pub type InnerClient = tonic::client::Grpc<
+    tower::util::Either<
```
coryan: TIL, and you just became my tower consultant.

Should we refactor this to something like:

```rust
pub type InnerClient = tonic::client::Grpc<GrpcService>;
```

where only `GrpcService` depends on the setting for `google_cloud_unstable_tracing`?

src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -0,0 +1,78 @@
...
+    S: Service<http::Request<B>, Response = http::Response<ResBody>> + Clone + Send + 'static,
+    S::Future: Send + 'static,
+    S::Error: std::fmt::Display,
+{
+    type Response = S::Response;
+    type Error = S::Error;
+    // We use `Box<dyn Future...>` (type erasure) here to simplify the type signature.
+    // Without this, we would need to explicitly name the complex type returned by
+    // `.instrument()` (and any implementation changes in `call`), which can be verbose and brittle.
+    type Future = Pin<Box<dyn Future<Output = Result<Self::Response, Self::Error>> + Send>>;
```
coryan, line 70: This is probably the right call, can you add some comments explaining what is the cost of this extra abstraction? Are we doing an extra allocation per RPC? Or an extra allocation per message in a streaming RPC?

src/gax-internal/src/observability.rs
westarle: done, thanks!

src/gax-internal/tests/grpc_observability.rs
westarle: updated to assert it is empty, so we'll know to update the test when we add stuff.

src/gax-internal/src/grpc.rs
westarle: looks better, thanks!

src/gax-internal/src/observability/grpc_tracing.rs
westarle: done, let me know if you want more detail -- essentially "call" creates the stream (or the single RPC) and the future returns when initial response headers are received.

PR #3770, opened 2025-11-11
impl(storage): connect bidi streaming reads

coryan:
Part of the work for #3626

### DIFF
--- a/src/storage/src/storage/bidi/connector.rs
+++ b/src/storage/src/storage/bidi/connector.rs
@@ -303,6 +303,8 @@ mod tests {
         options.retry_policy = Arc::new(NeverRetry);
         let mut connector = Connector::new(spec, options, client);
         let start = connector.connect(Vec::new());
+        // Verify the result is `Send`. Eventually this will be called from
+        // functions that need to be `send()` themselves.
         need_send(&start);
 
         let err = start.await.unwrap_err();
@@ -314,8 +316,9 @@ mod tests {
     #[tokio::test]
     #[test_case("")]
     #[test_case("my-bucket")]
-    async fn binding(bucket_name: &str) -> Result<()> {
+    async fn binding_error(bucket_name: &str) -> Result<()> {
         let mut mock = MockTestClient::new();
+        // Binding errors are detected before a request is sent.
         mock.expect_start().never();
         let client = SharedMockClient::new(mock);
 
@@ -527,8 +530,9 @@ mod tests {
         Ok(())
     }
 
+    // Verify redirects *after* the stream opens work as expected.
     #[tokio::test]
-    async fn start_redirect_open_then_redirect() -> Result<()> {
+    async fn start_redirect_open_with_redirect_then_error() -> Result<()> {
         let (tx, rx) = tokio::sync::mpsc::channel::<tonic::Result<BidiReadObjectResponse>>(5);
         let stream = tonic::Response::from(rx);
 


coryan:
/FYI @cjc25 @BenWhitehead this is the first PR that does anything "storage-y".  In particular, the `x-goog-request-params` header.

src/storage/src/storage/bidi/connector.rs
```diff
@@ -0,0 +1,842 @@
...
+        spec: Arc<Mutex<BidiReadObjectSpec>>,
+        tx: Sender<BidiReadObjectRequest>,
+        response: tonic::Result<tonic::Response<T::Stream>>,
+    ) -> Result<(BidiReadObjectResponse, Connection<T::Stream>)> {
+        let response = match response {
+            Ok(r) => r,
+            Err(status) => return Err(handle_redirect(spec, status)),
+        };
+        let (_metadata, mut stream, _) = response.into_parts();
+        match stream.next_message().await {
```
dbolduc: (comment for my understanding): Ok, the first message is special

src/storage/src/storage/bidi/connector.rs
```diff
@@ -0,0 +1,842 @@
...
+    use gax::error::binding::{BindingError, SubstitutionFail};
+    use gax::retry_policy::NeverRetry;
+    use static_assertions::assert_impl_all;
+    use std::error::Error as _;
+    use std::sync::Arc;
+    use test_case::test_case;
+
+    #[test]
+    fn assertions() {
+        assert_impl_all!(Connector: Clone, std::fmt::Debug, Send, Sync);
```
dbolduc: TIL

src/storage/src/storage/bidi/connector.rs
```diff
@@ -0,0 +1,842 @@
...
+        let retry = Arc::new(RetryRedirect::new(self.options.retry_policy.clone()));
+        let backoff = self.options.backoff_policy.clone();
+        let client = self.client.clone();
+        let request = BidiReadObjectRequest {
+            read_object_spec: Some((*self.spec.lock().expect("never poisoned")).clone()),
+            read_ranges: ranges,
+        };
+        let options = self.options.clone();
+        let spec = self.spec.clone();
+        let inner = async move |_| {
```
suzmue, line 98: nit: This is where the function get's interesting, everything before this is clone. Adding a comment here would make it easier for maintainers to find the interesting part of the function.

src/storage/src/storage/bidi/connector.rs
```diff
@@ -0,0 +1,842 @@
...
+    pub rx: S,
+}
+
+impl<S> Connection<S> {
+    pub fn new(tx: Sender<BidiReadObjectRequest>, rx: S) -> Self {
+        Self { tx, rx }
+    }
+}
+
+/// Establishes connections to gRPC for bidi streaming reads.
```
suzmue: nit: More words in this doc comment might be nice for coming back to this in the future about what all the Connector is responsible for / to orient devs

Maybe (might be too detailed): 
```
/// A `Connector` is reponsible for the initial connection and handling reconnections when a stream breaks.
/// It updates the spec with new generation IDs or routing tokens as the stream progresses and manages the retry and resume policies that govern how it handles failures.

File: src/storage/src/storage/bidi/connector.rs [hunk 5]
```diff
@@ -0,0 +1,842 @@
...
+                rx,
+                options,
+                &X_GOOG_API_CLIENT_HEADER,
+                &x_goog_request_params,
+            )
+            .await?;
+        Self::started(spec, tx, response).await
+    }
+
+    async fn started(
```
suzmue: nit: also could use a comment here, that this reads the first message from the stream.

src/storage/src/storage/bidi/connector.rs
```diff
@@ -0,0 +1,842 @@
...
+    pub async fn reconnect(
+        &mut self,
+        status: tonic::Status,
+        ranges: Vec<ProtoRange>,
+    ) -> Result<(BidiReadObjectResponse, Connection<T::Stream>)> {
+        use crate::read_resume_policy::ReadResumePolicy;
+
+        let error = handle_redirect(self.spec.clone(), status);
+        self.reconnect_attempts += 1;
+        let policy = ResumeRedirect::new(self.options.read_resume_policy());
```
suzmue, line 114: reconnect_attempts is increased even though policy.on_error may result in an error. Is this the desired behavior?

src/storage/src/storage/bidi/connector.rs
```diff
@@ -0,0 +1,842 @@
...
+            bucket: "projects/_/buckets/test-only-bucket".into(),
+            object: "test-only-object".into(),
+            ..BidiReadObjectSpec::default()
+        };
+
+        let mut options = test_options();
+        options.retry_policy = Arc::new(NeverRetry);
+        let mut connector = Connector::new(spec, options, client);
+        let start = connector.connect(Vec::new());
+        need_send(&start);
```
suzmue, line 308: Why is this necessary?

src/storage/src/storage/bidi/connector.rs
```diff
@@ -0,0 +1,842 @@
...
+        let err = start.await.unwrap_err();
+        assert!(err.is_connect(), "{err:?}");
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    #[test_case("")]
+    #[test_case("my-bucket")]
+    async fn binding(bucket_name: &str) -> Result<()> {
```
suzmue: nit: binding_error is slightly more informative of the set of tests. Was initially confused as to why expect_start was never.

src/storage/src/storage/bidi/connector.rs
```diff
@@ -0,0 +1,842 @@
...
+        assert_eq!(err.status(), permanent_error().status(), "{err:?}");
+        let guard = connector.spec.lock().expect("never poisoned");
+        assert_eq!(guard.routing_token.as_deref(), Some("r1"));
+        assert_eq!(guard.read_handle, Some(redirect_handle()));
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn start_redirect_open_then_redirect() -> Result<()> {
```
suzmue: nit: this test seems like it is start_redirect_open_then_error?

src/storage/src/storage/bidi/connector.rs
coryan: Done.

src/storage/src/storage/bidi/connector.rs
coryan: Yes. If it returns an error then the resume loop stops and `reconnect_attempts` is never used. This is consistent with how we used this in `src/storage/src/storage/read_object/resumable.rs`.

```rust
    /// The number of times the read request has been interrupted already.
    pub attempt_count: u32,
```

We could change this, provide `self.reconnect_attempts + 1` to the query and then increment on `Continue`, but seems more complicated than needed.

src/storage/src/storage/bidi/connector.rs
coryan: Joined the functions and put some comments at the top.

src/storage/src/storage/bidi/connector.rs
coryan: I will use the future in other functions that need to be `Send` themselves.

src/storage/src/storage/bidi/connector.rs
coryan: Renamed and added a comment.

src/storage/src/storage/bidi/connector.rs
coryan: renamed to "open_with_redirect" + "then_error".

src/storage/src/storage/bidi/connector.rs
coryan: Move the code and included some comments.

PR #3773, opened 2025-11-11
impl(storage): normalize read ranges

coryan:
Part of the work for #3626

### DIFF
--- a/src/storage/src/model_ext.rs
+++ b/src/storage/src/model_ext.rs
@@ -260,6 +260,40 @@ impl ReadRange {
             limit: count,
         })
     }
+
+    /// Maps the range to a `(offset, limit)` pair.
+    ///
+    /// Once the size of the object is known it is possible to normalize the
+    /// range. This is useful in bidi streaming reads because the size is always
+    /// known when the range is requested.
+    #[allow(dead_code)]
+    pub(crate) fn normalize(&self, object_size: i64) -> (i64, i64) {
+        use crate::model_ext::Range;
+        match self.0 {
+            Range::All => (0, object_size),
+            Range::Offset(o) => {
+                let o = o.clamp(0, i64::MAX as u64) as i64;
+                (
+                    o.clamp(0, object_size),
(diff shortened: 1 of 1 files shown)


src/storage/src/model_ext.rs
```diff
@@ -260,6 +260,40 @@ impl ReadRange {
...
         })
     }
+
+    /// Maps the range to a `(offset, limit)` pair.
+    ///
+    /// Once the size of the object is known it is possible to normalize the
+    /// range. This is useful in bidi streaming reads because the size is always
+    /// known when the range is requested.
+    #[allow(dead_code)]
+    pub(crate) fn normalize(&self, object_size: i64) -> (i64, i64) {
```
BenWhitehead, line 270: Both java and go had to rip out this kind of range logic because of appendable objects being able to grow over time, without the client knowing what the new size is. To make this blind read more tolerable, the server will respond with an `ObjectRangeData` with an empty `checksummed_data` and `range_end: true`.

coryan: Blegh.  I guess if the service is broken and delivering bad messages there isn't much to do in that case.

cjc25: The service resolves a size at read time. Issuing a read range (https://github.com/googleapis/googleapis/blob/master/google/storage/v2/storage.proto#L1301) with fields `read_offset: -10 read_length: 0` will resolve a tail size and return `ObjectRangeData` with `ReadRange` (https://github.com/googleapis/googleapis/blob/master/google/storage/v2/storage.proto#L1320-L1333) normalized to positive offset and length values.

coryan: > The service resolves a size at read time. 

Yes, I was trying to avoid messy arithmetic when receiving responses. Oh well.

dbolduc (approved):

coryan:
I am going to follow a different approach since the size may change with Rapid buckets. Closing for now.

PR #3778, opened 2025-11-12
impl(o11y): implement standard RPC attributes for gRPC tracing

westarle:


src/gax-internal/tests/grpc_observability.rs
```diff
@@ -64,16 +67,194 @@ mod tests {
...
+                "Echo",
+            ));
+            e
+        };
+        let request = google::test::v1::EchoRequest {
+            message: "test message".into(),
+            ..Default::default()
+        };
+        let _ = client
+            .execute::<_, google::test::v1::EchoResponse>(
```
coryan, line 137: I think you want to add a test for the (new) `.bidi_stream()` requests. It can be in a separate PR if you want.

Issue #3776, opened 2025-11-11
google-cloud-auth v1.1.1 relies on num-bigint-dig v0.8.5 which will be rejected in future Rust versions


brannondorsey:
#### Environment details

  - Programming language: Rust
  - OS: MacOS 15.6.1 
  - Language runtime version: Rust stable (v1.91.1)
  - Package version: google-cloud-auth v1.1.1

#### Steps to reproduce

```bash
rustc --version # rustc 1.91.1 (ed61e7d7e 2025-11-07) which is latest stable rust
 
cargo new foo
cd foo
cargo add google-cloud-auth@=1.1.1
cargo check 
```

You can see the warning produced by `cargo check` below.

```text
warning: the following packages contain code that will be rejected by a future version of Rust: num-bigint-dig v0.8.5
note: to see what the problems were, use the option `--future-incompat-report`, or run `cargo report future-incompatibilities --id 1`
```

To resolve the warning, install the package at 1.1.0 instead.

```bash
# still inside foo
cargo add google-cloud-auth@=1.1.0
cargo check
```

You will see the warning is gone.

It seems this issue was introduced because 1.1.1 relies on `jsonwebtoken`, which relies on `rsa`, which relies on v0.8.5 of `num-bigint-dig`, where this warning originates.

Thanks!

coryan:
Blegh... We missed the warning.  Apparently, these do not result in errors even with  `cargo clippy -- -Dwarnings` and we rely on the CI buildings for this kind of thing. 

Regardless. We also have a direct `dev-dependency` on `rsa`, so that is still a problem.

coryan:
Seems like this is fixed, the is a 0.8.6 version of `num-bigint-dig` out, compatible with future versions of Rust, and our `Cargo.lock` file picked it up.

PR #3786, opened 2025-11-13
cleanup(auth): make jsonwebtoken a dep only with cfg_google_cloud_unstable_id_token

alvarowolfx:
Towards #3776


alvarowolfx:
> I am not sure I would call this "ci", maybe "cleanup"? no biggie either way.

I was on a fence here, went back and forth on ci and cleanup.

PR #3792, opened 2025-11-13
impl(auth): make mds format an enum

alvarowolfx:
Towards #3449

### DIFF
--- a/src/auth/src/credentials/idtoken/mds.rs
+++ b/src/auth/src/credentials/idtoken/mds.rs
@@ -118,11 +118,11 @@ pub enum Format {
 }
 
 impl Format {
-    fn to_string(&self) -> String {
+    fn as_str(&self) -> &str {
         match self {
-            Format::Standard => "standard".to_string(),
-            Format::Full => "full".to_string(),
-            Format::UnknownValue(value) => value.clone(),
+            Format::Standard => "standard",
+            Format::Full => "full",
+            Format::UnknownValue(value) => value.as_str(),
         }
     }
 }
@@ -245,7 +245,7 @@ impl TokenProvider for MDSTokenProvider {
             .query(&[("audience", audience)]);
 
         let request = self.format.iter().fold(request, |builder, format| {
-            builder.query(&[("format", format.to_string())])
+            builder.query(&[("format", format.as_str())])
         });
         let request = self.licenses.iter().fold(request, |builder, licenses| {
             builder.query(&[("licenses", licenses)])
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/idtoken/mds.rs
```diff
@@ -106,12 +106,34 @@ where
     }
 }
 
+// Parameter that specifies whether or not the project and instance details are included in the ID Token
+// fetched from the Metadata Service.
```
coryan:
```suggestion
/// Specifies what assertions are included in ID Tokens fetched from the Metadata Service.
```

src/auth/src/credentials/idtoken/mds.rs
```diff
@@ -106,12 +106,34 @@ where
...
+#[derive(Debug, Clone)]
+pub enum Format {
+    // Omit project and instance details from the payload. It's the default value.
```
coryan:
```suggestion
    /// Omit project and instance details from the payload. It's the default value.
```

src/auth/src/credentials/idtoken/mds.rs
```diff
@@ -106,12 +106,34 @@ where
...
+    Standard,
+    // Include project and instance details in the payload.
```
coryan:
```suggestion
    /// Include project and instance details in the payload.
```

src/auth/src/credentials/idtoken/mds.rs
```diff
@@ -106,12 +106,34 @@ where
...
+    Full,
+    // Use this variant to handle new values that are not yet known to this library.
```
coryan:
```suggestion
    /// Use this variant to handle new values that are not yet known to this library.
```

src/auth/src/credentials/idtoken/mds.rs
```diff
@@ -106,12 +106,34 @@ where
...
+
+impl Format {
+    fn as_str(&self) -> &str {
+        match self {
+            Format::Standard => "standard",
+            Format::Full => "full",
+            Format::UnknownValue(value) => value.as_str(),
+        }
+    }
+}
```
coryan, line 128: Need some tests for this.

alvarowolfx: used test_case to test with the other enums and therefore, this method. Also moved to use `&str` to `String`, I was running into some lifetime issues.

alvarowolfx: nvm, came back to `as_str()`.

coryan (approved):

PR #3802, opened 2025-11-13
chore: update sidekick version

westarle:


.github/workflows/sdk.yaml
```diff
@@ -592,7 +592,7 @@ jobs:
           sudo unzip -x /tmp/protoc.zip
           protoc --version
       - name: Regenerate all the code
-        run: go run github.com/googleapis/librarian/cmd/sidekick@v1.0.1-0.20251107152048-d798500afbbd refreshall
+        run: go run github.com/googleapis/librarian/cmd/sidekick@v0.6.1-0.20251113184719-000d3a0de636 refreshall
```
dbolduc, line 595: aside (nothing to do with your work): monotonic versioning would be nice.

coryan: The 1.0 stuff was an accident. They decided to withdraw the 1.0 release instead of living with the accident. SGTM.

westarle, line 595: I'm not certain but I think the version bumped *back* because of https://github.com/googleapis/librarian/pull/2895

PR #3805, opened 2025-11-13
impl(storage): normalize and update read ranges

coryan:
Implement some types to normalize and update read ranges in bidi streaming reads. We need to model both requested ranges (which include ranges with a negative offset) and ranges updated from the first (and subsequent) responses. The latter always have positive offsets.

Part of the work for #3626

### DIFF
--- a/src/storage/src/error.rs
+++ b/src/storage/src/error.rs
@@ -137,29 +137,79 @@ pub enum ReadError {
     UnrecoverableBidiReadInterrupt(#[source] std::sync::Arc<crate::Error>),
 
     /// A bidi read received an invalid offset.
+    ///
+    /// # Troubleshooting
+    ///
+    /// This indicates a bug in the service or a corrupted message in
+    /// transit Please contact [Google Cloud support] with as much detail as
+    /// possible.
+    ///
+    /// [open an issue]: https://github.com/googleapis/google-cloud-rust/issues/new/choose
+    /// [Google Cloud support]: https://cloud.google.com/support
     #[cfg(google_cloud_unstable_storage_bidi)]
     #[error("invalid offset in bidi response: {0}")]
     BadOffsetInBidiResponse(i64),
 
     /// A bidi read received an invalid length.
+    ///
+    /// # Troubleshooting
+    ///
+    /// This indicates a bug in the service or a corrupted message in
+    /// transit. Please contact [Google Cloud support] with as much detail as
+    /// possible.
+    ///
+    /// [open an issue]: https://github.com/googleapis/google-cloud-rust/issues/new/choose
+    /// [Google Cloud support]: https://cloud.google.com/support
     #[cfg(google_cloud_unstable_storage_bidi)]
     #[error("invalid length in bidi response: {0}")]
     BadLengthInBidiResponse(i64),
 
     /// A bidi read without a valid range.
+    ///
+    /// # Troubleshooting
+    ///
+    /// This indicates a bug in the service or a corrupted message in
+    /// transit Please contact [Google Cloud support] with as much detail as
+    /// possible.
+    ///
+    /// [open an issue]: https://github.com/googleapis/google-cloud-rust/issues/new/choose
+    /// [Google Cloud support]: https://cloud.google.com/support
     #[cfg(google_cloud_unstable_storage_bidi)]
     #[error("missing range in bidi response")]
     MissingRangeInBidiResponse,
 
     /// An out of order bidi read.
+    ///
+    /// # Troubleshooting
+    ///
+    /// The client library received an out-of-sequence range of data. This
+    /// indicates a bug in the service or the client library.
+    ///
+    /// Please [open an issue] with as much detail as possible or contact
+    /// [Google Cloud support].
+    ///
+    /// [open an issue]: https://github.com/googleapis/google-cloud-rust/issues/new/choose
+    /// [Google Cloud support]: https://cloud.google.com/support
     #[cfg(google_cloud_unstable_storage_bidi)]
     #[error("out of order bidi response, expected offset={expected}, got={got}")]
     OutOfOrderBidiResponse { got: i64, expected: i64 },
 
-    /// An unexpected range id.
+    /// The service returned a range id unknown to the client library.
+    ///
+    /// # Troubleshooting
+    ///
+    /// In bidi reads the application may issue multiple concurrent reads for
+    /// different ranges of the same object. The client library assigns ids to
+    /// each range. This indicates a bug in the service or the client library.
+    ///
+    /// Please [open an issue] with as much detail as possible or contact
+    /// [Google Cloud support].
+    ///
+    /// [open an issue]: https://github.com/googleapis/google-cloud-rust/issues/new/choose
+    /// [Google Cloud support]: https://cloud.google.com/support
     #[cfg(google_cloud_unstable_storage_bidi)]
     #[error("unknown range id in bidi response: {0}")]
-    UnknownRange(i64),
+    UnknownBidiRangeId(i64),
 }
 
 /// An unrecoverable problem in the upload protocol.
--- a/src/storage/src/storage/bidi/normalized_range.rs
+++ b/src/storage/src/storage/bidi/normalized_range.rs
@@ -18,7 +18,7 @@ use crate::google::storage::v2::ReadRange as ProtoRange;
 type ReadResult<T> = std::result::Result<T, ReadError>;
 
 /// A normalized range represents a range of bytes normalized with a positive
-/// offset.
+/// offset and length.
 ///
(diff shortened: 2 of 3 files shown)


src/storage/src/error.rs
```diff
@@ -130,6 +130,36 @@ pub enum ReadError {
         &'static str,
         #[source] Box<dyn std::error::Error + Send + Sync + 'static>,
     ),
+
+    /// A bidi read was interrupted with an unrecoverable error.
```
dbolduc, line 134: comment: we are going all in on "bidi" in the public API names. I don't have a better name. I am only slightly concerned because "Bidi" does not show up in the GCS cloud docs. (It is in the name of the RPCs.... so that makes it more ok in my opinion).

src/storage/src/error.rs
```diff
@@ -130,6 +130,36 @@ pub enum ReadError {
...
+
+    /// A bidi read received an invalid offset.
+    #[cfg(google_cloud_unstable_storage_bidi)]
+    #[error("invalid offset in bidi response: {0}")]
+    BadOffsetInBidiResponse(i64),
+
+    /// A bidi read received an invalid length.
+    #[cfg(google_cloud_unstable_storage_bidi)]
+    #[error("invalid length in bidi response: {0}")]
+    BadLengthInBidiResponse(i64),
```
dbolduc, line 165: Q: why is a length signed? is that what the proto uses?

src/storage/src/error.rs
```diff
@@ -130,6 +130,36 @@ pub enum ReadError {
...
+    #[cfg(google_cloud_unstable_storage_bidi)]
+    #[error("missing range in bidi response")]
+    MissingRangeInBidiResponse,
+
+    /// An out of order bidi read.
+    #[cfg(google_cloud_unstable_storage_bidi)]
+    #[error("out of order bidi response, expected offset={expected}, got={got}")]
+    OutOfOrderBidiResponse { got: i64, expected: i64 },
+
+    /// An unexpected range id.
```
dbolduc: comment: hmm.. I don't know what a range ID is.

src/storage/src/storage/bidi/normalized_range.rs
```diff
@@ -0,0 +1,284 @@
...
+/// This can be tedious because applications may request ranges with negative
+/// offsets (representing the last bytes of the object), and with or without a
+/// length limit.
+///
+/// After the first response arrives these requested ranges can be normalized to
+/// have a positive offset. This struct represent such normalized changes.
+#[derive(Clone, Copy, Debug, PartialEq)]
+pub struct NormalizedRange {
+    offset: i64,
+    length: Option<i64>,
```
dbolduc, line 43: same Q: on length being negative

src/storage/src/error.rs
dbolduc: Maybe we calculate a negative number and that is the problem?

src/storage/src/storage/bidi/normalized_range.rs
```diff
@@ -0,0 +1,284 @@
...
+        }
+    }
+
+    pub fn matching_offset(&self, requested_offset: u64) -> bool {
+        self.offset as u64 == requested_offset
+    }
+
+    pub fn update(&mut self, response: ProtoRange) -> ReadResult<()> {
+        let update = NormalizedRange::from_proto(response)?;
+        if update.offset != self.offset {
```
dbolduc, line 97: Q: so `update.offset < self.offset` is invalid?

src/storage/src/storage/bidi/normalized_range.rs
```diff
@@ -0,0 +1,284 @@
...
+                    got: got as u64,
+                    expected: expected as u64,
+                });
+            }
+        };
+        self.offset = update.offset + update.length().unwrap_or_default();
+        Ok(())
+    }
+
+    pub fn handle_empty(&self, end: bool) -> ReadResult<()> {
```
dbolduc, line 118: comment: strange.

src/storage/src/storage/bidi/normalized_range.rs
```diff
@@ -0,0 +1,284 @@
...
+        match (end, self.length) {
+            (true, Some(l)) if l > 0 => Err(ReadError::ShortRead(l as u64)),
+            _ => Ok(()),
```
dbolduc, line 124: are there not cases where you are `end == false`, and `length` is greater than the remainder of the bytes in the object?

if there are no cases, why can it happen when `end == true`?

src/storage/src/storage/bidi/remaining_range.rs
```diff
@@ -0,0 +1,225 @@
...
+use super::normalized_range::NormalizedRange;
+use crate::error::ReadError;
+use crate::google::storage::v2::ReadRange as ProtoRange;
+use crate::model_ext::RequestedRange;
+
+type ReadResult<T> = std::result::Result<T, ReadError>;
+
+/// Tracks the remaining range.
+///
+/// [PendingRange][super::pending_range::PendingRange] is initialized with
```
dbolduc: `PendingRange` coming soon? if so, ignore

dbolduc (dismissed):

src/storage/src/storage/bidi/normalized_range.rs
coryan: The proto may return negative values. This type never constructs a negative length value.

src/storage/src/storage/bidi/normalized_range.rs
coryan: Yes. Returning earlier data is a violation of the protocol. I could ignore that data, but then why would I trust anything the service says?

src/storage/src/storage/bidi/remaining_range.rs
coryan: Yes, though it will arrive with a different name. Updated the comment.

src/storage/src/storage/bidi/normalized_range.rs
coryan: I included a larger comment explaining why. This type is intended to be internal-only btw, the comments are too long as they are (and lacking examples if we were to make this public).

src/storage/src/storage/bidi/normalized_range.rs
coryan: A bit, left a comment.

src/storage/src/storage/bidi/normalized_range.rs
coryan: The calling code is always:

```
    pub(super) async fn handle_data(
        &mut self,
        data: Option<ChecksummedData>,
        received_range: ProtoRange,
        end: bool,
        ) -> ReadResult<()> {
            self.state.update(received_range)?;
            let Some(data) = data else {
                return self.state.handle_empty(end);
            };
```

So the question is: if we receive an empty range, and it is marked as the end of the read, did we miss any data?  Empty ranges in the middle are weird, but can be tolerated.

src/storage/src/error.rs
coryan: A bit of a mess I think. As you say, the reference docs do mention "bidi":

https://cloud.google.com/storage/docs/reference/rpc/google.storage.v2#bidireadobjectrequest

The service guides do not. I think the product name is "Rapid". But that is not quite right either (Rapid implies Rapid Buckets, and this protocol can be used with other buckets too).

Maybe @BenWhitehead has some guidance, but I will merge these changes as they are (they are all private and easy to change) to make progress.

src/storage/src/error.rs
coryan: No, the proto may send a negative number (or too large of a length) and I want to return something.

src/storage/src/error.rs
coryan: Good point, improved the comments and renamed the variant.

coryan (commented):
PTAL

src/storage/src/storage/bidi/requested_range.rs
```diff
@@ -0,0 +1,65 @@
...
+
+use crate::error::ReadError;
+use crate::google::storage::v2::ReadRange as ProtoRange;
+use crate::model_ext::RequestedRange;
+
+type ReadResult<T> = std::result::Result<T, ReadError>;
+
+/// Additional functions to use `RequestedRange` in pending requests.
+///
+/// [PendingRange][super::pending_range::PendingRange] is initialized with
```
dbolduc, line 23: There is another `ActiveRead`

dbolduc (approved):

src/storage/src/storage/bidi/requested_range.rs
coryan: Ack, let's see if this makes more sense after the next PR.

Issue #3793, opened 2025-11-13
Create a patch release for crates depending on google-cloud-longrunning


suzmue:
In our last release, there were new features added in `longrunning` that our other crates depend on, but they still declare minimum version as "1".

#3782 updates the dependency in the main branch, but we should do a patch release to pick up this version requirement and possibly yank the bad versions after.

suzmue:
This is done. See https://github.com/googleapis/google-cloud-rust/releases/tag/release-20251114

PR #3806, opened 2025-11-13
chore: bump all crates depending on longrunning

suzmue:
This may be too eager, but we bumped the version of longrunning in the workspace Cargo.toml and many of these crates depend on features in the upgraded version.

There are 122 files containing "longrunning.workspace". 119 of those are changed in this PR. The other files are the user-guide-samples, integration-tests, and google-cloud-showcase-v1beta1 crates, and these are marked as publish=false so did not need a change.

For #3793

### DIFF
--- a/src/generated/cloud/cloudsecuritycompliance/v1/Cargo.toml
+++ b/src/generated/cloud/cloudsecuritycompliance/v1/Cargo.toml
@@ -16,7 +16,7 @@
 
 [package]
 name                   = "google-cloud-cloudsecuritycompliance-v1"
-version                = "2.1.0"
+version                = "2.1.1"
 description            = "Google Cloud Client Libraries for Rust - Cloud Security Compliance API"
 edition.workspace      = true
 authors.workspace      = true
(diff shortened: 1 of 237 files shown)


dbolduc:
The `.sidekick.toml` are the source of truth for the versions. Gotta do them too.

Aside: I forget why we skip formatting on generated things. It might be simpler if we didn't?

https://github.com/googleapis/google-cloud-rust/blob/80b732a06afddad88d8f8d944d3ec1e4cc096878/.github/workflows/lint.yaml#L73-L74

suzmue:
> The `.sidekick.toml` are the source of truth for the versions. Gotta do them too.

Right thank you 🙏

src/generated/cloud/cloudsecuritycompliance/v1/Cargo.toml
```diff
@@ -16,7 +16,7 @@
 
 [package]
 name                   = "google-cloud-cloudsecuritycompliance-v1"
-version                = "2.1.0"
+version                = "2.1.1"
```
coryan, line 19: nice

PR #3808, opened 2025-11-14
impl(o11y): add status handling for gRPC network spans

westarle:
- Implement `ResponseFuture` to wrap the inner future and record status.
- This records the span after the headers arrive (before the body is read and decoded)
- We will read the body (for large payloads e.g.) and handling errors (cancelation e.g.) in a future PR

#3766

### DIFF
--- a/src/gax-internal/src/observability/grpc_tracing.rs
+++ b/src/gax-internal/src/observability/grpc_tracing.rs
@@ -152,7 +152,7 @@ fn record_response_status<ResBody>(span: &tracing::Span, response: &http::Respon
     if response
         .headers()
         .get("grpc-status")
-        .map_or(true, |v| v == "0")
+        .is_none_or(|v| v == "0")
     {
         span.record(otel_attr::RPC_GRPC_STATUS_CODE, 0_i64);
         return;


src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -114,10 +114,75 @@ where
...
+where
+    F: Future<Output = Result<http::Response<ResBody>, Error>>,
+    Error: std::fmt::Display,
+{
+    type Output = Result<http::Response<ResBody>, Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.project();
+        let _guard = this.span.enter();
+        let result = futures_util::ready!(this.inner.poll(cx));
```
coryan, line 139: This works, but I had to read the code of `futures_util::ready!` to convince myself.  Maybe with more experience I will not care, but it is not obvious that this "leaves" the span when the inner future is not ready.

Maybe a small comment?

src/gax-internal/src/observability/grpc_tracing.rs
coryan: Since we are not calling `.instrument()` can we simplify this code / reduce the overhead?

src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -114,10 +114,75 @@ where
...
+        match &result {
+            Ok(response) => record_response_status(this.span, response),
+            Err(e) => record_error_status(this.span, e),
+        }
+        Poll::Ready(result)
+    }
+}
+
+fn record_response_status<ResBody>(span: &tracing::Span, response: &http::Response<ResBody>) {
+    match tonic::Status::from_header_map(response.headers()) {
```
coryan: Does this function get called twice then? One for this layer and once in the real tonic layer?  Do you have a sense of how expensive it is for the "success" case?   Am I connect in assuming it is fairly fast because it returns immediately (as the header does not exist)? 

https://docs.rs/tonic/latest/src/tonic/status.rs.html#465

It may be fine if it is a bit expensive, just wan to document the tradeoff.

src/gax-internal/src/observability/grpc_tracing.rs
westarle: done

src/gax-internal/src/observability/grpc_tracing.rs
westarle: done, thanks!

src/gax-internal/src/observability/grpc_tracing.rs
westarle: I put a guard above to make sure it is as fast as possible in the success case.

coryan (approved):
Thanks

PR #3811, opened 2025-11-14
Switch to rustls

casibbald:
Replace native-tls (OpenSSL) with rustls-tls in the workspace reqwest
dependency to enable cross-compilation to musl targets without requiring
OpenSSL cross-compilation toolchain.

This change adds `rustls-tls` and `rustls-tls-webpki-roots` features to
the reqwest workspace dependency, which propagates to all crates that
use reqwest via the workspace, including google-cloud-gax-internal.

Benefits:
- Eliminates OpenSSL dependency and cross-compilation complexity
- Enables musl target builds without vendored OpenSSL or cargo-zigbuild
- Reduces binary size (rustls is pure Rust, no C dependencies)
- Improves musl compatibility and static linking support

The change maintains API compatibility as reqwest's rustls backend
provides equivalent functionality to native-tls for HTTP client operations.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -325,7 +325,7 @@ thiserror             = { default-features = false, version = "2" }
 time                  = { default-features = false, version = "0.3.44" }
 tokio                 = { default-features = false, version = "1" }
 tokio-metrics         = { default-features = false, version = "0.4.5" }
-tonic                 = { default-features = false, version = "0.14.1", features = ["tls-native-roots", "tls-ring"] }
+tonic                 = { default-features = false, version = "0.14.1", features = ["tls-ring", "tls-webpki-roots"] }
 tonic-prost           = { default-features = false, version = "0.14.2" }
 tonic-prost-build     = { default-features = false, version = "0.14.1" }
 tower                 = { default-features = false, version = "0.5", features = ["util"] }


Cargo.toml
```diff
@@ -314,7 +314,7 @@ prost                 = { default-features = false, version = "0.14", features =
 prost-build           = { default-features = false, version = "0.14" }
 prost-types           = { default-features = false, version = "0.14" }
 rand                  = { default-features = false, version = "0.9.2" }
-reqwest               = { default-features = false, version = "0.12.24", features = ["json"] }
```
coryan, line 317: Hmmm... my understanding is that you can enable these features in your own `Cargo.toml` file, and if we enable them here no downstream dependency could disable them.  Is that not the case?

Cargo.toml
```diff
@@ -325,7 +325,7 @@ thiserror             = { default-features = false, version = "2" }
 time                  = { default-features = false, version = "0.3.44" }
 tokio                 = { default-features = false, version = "1" }
 tokio-metrics         = { default-features = false, version = "0.4.5" }
-tonic                 = { default-features = false, version = "0.14.1", features = ["tls-native-roots", "tls-ring"] }
```
coryan, line 328: This could be a problem.  But we should fix it by reducing the number of default features.

coryan:
I should have mentioned. If you think I am wrong just say so.  I am sure our handling of TLS dependencies is far from perfect.

coryan:
Any thoughts on my comments?

coryan:
Closing due to lack of response. Please feel free to reopen or create a bug with more details about the problem you are trying to solve, ideally with a minimal repro, so we can help.

PR #3814, opened 2025-11-17
impl(storage): a type for active reads

coryan:
This type holds the pending remaining range (for resends) and a `mpsc` sender (to deliver data and errors to the application).

The worker task will keep a collection of these objects and deliver data and errors to the application through them.

Part of the work for #3626

### DIFF
--- a/src/storage/src/storage/bidi/active_read.rs
+++ b/src/storage/src/storage/bidi/active_read.rs
@@ -64,7 +64,7 @@ impl ActiveRead {
 
     pub(super) async fn handle_error(&mut self, error: ReadError) {
         if let Err(e) = self.sender.send(Err(error)).await {
-            tracing::error!("cannot notify sender about read error: {e:?}");
+            tracing::error!("cannot notify reader (dropped?) about: {e:?}");
         }
     }
 
@@ -74,7 +74,7 @@ impl ActiveRead {
             .send(Err(ReadError::UnrecoverableBidiReadInterrupt(error)))
             .await
         {
-            tracing::error!("cannot notify sender about unrecoverable error: {e:?}");
+            tracing::error!("cannot notify reader (dropped?) about: {e:?}");
         }
     }
 
@@ -88,6 +88,8 @@ mod tests {
     use super::super::tests::{permanent_error, proto_range};
     use super::*;
     use crate::model_ext::ReadRange;
+    use std::sync::Mutex;
+    use tracing::Subscriber;
 
     #[tokio::test]
     async fn normal() -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/bidi/active_read.rs
```diff
@@ -0,0 +1,229 @@
...
+            content: bytes::Bytes::from_owner(content.clone()),
+            ..ChecksummedData::default()
+        };
+        range.handle_data(Some(data), response, false).await?;
+        assert_eq!(range.as_proto(0), proto_range(25, 0));
+
+        let recv = rx.recv().await;
+        assert!(matches!(recv, Some(Ok(ref b)) if *b == content), "{recv:?}");
+
+        assert_eq!(range.as_proto(0), proto_range(25, 0), "{range:?}");
```
dbolduc: nit: we already check this on L104, and `as_proto` is a const function, so I don't expect the assertion to fail.

src/storage/src/storage/bidi/active_read.rs
```diff
@@ -0,0 +1,229 @@
...
+            .await;
+    }
+
+    #[tokio::test]
+    async fn interrupted() {
+        let (tx, mut rx) = tokio::sync::mpsc::channel(1);
+        let requested = ReadRange::all().0;
+        let mut range = ActiveRead::new(tx, requested);
+        let error = Arc::new(permanent_error());
+        range.interrupted(error.clone()).await;
```
dbolduc, line 243: Q: should this test verify that some sort of logging occurs?

We could set up a tracing_subscriber and see what it catches I think.

src/storage/src/storage/bidi/active_read.rs
```diff
@@ -0,0 +1,229 @@
...
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn handle_error() {
+        let (tx, mut rx) = tokio::sync::mpsc::channel(1);
+        let requested = ReadRange::all().0;
+        let mut range = ActiveRead::new(tx, requested);
+        range
+            .handle_error(ReadError::MissingRangeInBidiResponse)
```
dbolduc, line 211: verify that `rx` receives this error?

src/storage/src/storage/bidi/active_read.rs
coryan: Removed.

src/storage/src/storage/bidi/active_read.rs
coryan: Done.

src/storage/src/storage/bidi/active_read.rs
coryan: > Q: should this test verify that some sort of logging occurs?

Done.

> We could set up a tracing_subscriber and see what it catches I think.

I think we really want counters for this sort of thing.  The messages may be too much noise.

dbolduc: > counters

err... I think we still want to log the error's contents. I am not sure a counter can capture that.

I think this is a case of "why not both?"

PR #3762, opened 2025-11-10
docs(auth): id token user guide

alvarowolfx:
Towards #3449

PR #3818, opened 2025-11-17
docs(auth): add builder examples to id token builder

alvarowolfx:
Splitting work on #3762 that can already be merged.

### DIFF
--- a/src/auth/src/credentials/idtoken/impersonated.rs
+++ b/src/auth/src/credentials/idtoken/impersonated.rs
@@ -89,6 +89,25 @@ use serde_json::Value;
 use std::sync::Arc;
 
 /// A builder for constructing Impersonated Service Account [IDTokenCredentials] instance.
+///
+/// # Example
+/// ```
+/// # use google_cloud_auth::credentials::idtoken;
+/// # tokio_test::block_on(async {
+/// let impersonated_credential = serde_json::json!({
+///     "type": "impersonated_service_account",
+///     "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/test-principal:generateAccessToken",
+///     "source_credentials": {
+///         "type": "authorized_user",
+///         "client_id": "test-client-id",
+///         "client_secret": "test-client-secret",
+///         "refresh_token": "test-refresh-token"
+///     }
+/// });
+///
+/// let audience = "https://my-service.a.run.app";
+/// let credentials = idtoken::impersonated::Builder::new(impersonated_credential).build();
+/// # });
 pub struct Builder {
     source: BuilderSource,
     delegates: Option<Vec<String>>,
@@ -167,16 +186,8 @@ impl Builder {
     /// ```
     /// # use google_cloud_auth::credentials::idtoken;
     /// # use serde_json::json;
-    /// # let impersonated_credential = json!({
-    /// #     "type": "impersonated_service_account",
-    /// #     "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/test-principal:generateAccessToken",
-    /// #     "source_credentials": {
-    /// #         "type": "authorized_user",
-    /// #         "client_id": "test-client-id",
-    /// #         "client_secret": "test-client-secret",
-    /// #         "refresh_token": "test-refresh-token"
-    /// #     }
-    /// # });
+    /// let impersonated_credential = json!({ /* add details here */ });
+    ///
     /// let audience = "https://my-service.a.run.app";
     /// let credentials = idtoken::impersonated::Builder::new(audience, impersonated_credential)
     ///     .with_include_email()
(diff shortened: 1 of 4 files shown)


src/auth/src/credentials/idtoken/impersonated.rs
```diff
@@ -144,12 +161,56 @@ impl Builder {
...
+    /// # let impersonated_credential = json!({
+    /// #     "type": "impersonated_service_account",
+    /// #     "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/test-principal:generateAccessToken",
+    /// #     "source_credentials": {
+    /// #         "type": "authorized_user",
+    /// #         "client_id": "test-client-id",
+    /// #         "client_secret": "test-client-secret",
+    /// #         "refresh_token": "test-refresh-token"
+    /// #     }
+    /// # });
```
coryan: Why the full details here and not in the previous example?  Should we (at least) summarize the source credentials:

```suggestion
    /// # let impersonated_credential = json!({
    /// #     "type": "impersonated_service_account",
    /// #     "service_account_impersonation_url": "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/test-principal:generateAccessToken",
    /// #     "source_credentials": { /* details here */ }
    /// # });
```

And maybe link documentation for the format, e.g.:

https://docs.cloud.google.com/iam/docs/keys-create-delete#iam-service-account-keys-create-console

??

src/auth/src/credentials/idtoken/impersonated.rs
```diff
@@ -144,12 +161,56 @@ impl Builder {
...
+    /// let audience = "https://my-service.a.run.app";
+    /// let credentials = idtoken::impersonated::Builder::new(audience, impersonated_credential)
```
coryan, line 212: Here and the previous examples, consider at least making the type visible:


```suggestion
    /// let audience = "https://my-service.a.run.app";
    /// let credentials = idtoken::impersonated::Builder::new(audience, impersonated_credentials("test-principal"))

    /// fn impersonated_credentials(principal: &str) -> serde_json::Value {
    /// serde_json::json!({
    ///         "type": "impersonated_service_account",
    ///         "service_account_impersonation_url": format!( "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{principal}:generateAccessToken"),
    ///       "source_credentials": {
    ///         // See https://docs.cloud.google.com/iam/docs/keys-create-delete#iam-service-account-keys-create-console for possible formats
    ///        }
    /// })
    /// }
```

src/auth/src/credentials/idtoken/mds.rs
```diff
@@ -178,6 +201,18 @@ impl Builder {
...
     ///
     /// [license codes]: https://cloud.google.com/compute/docs/reference/rest/v1/images/get#body.Image.FIELDS.license_code
+    ///
+    /// # Example
+    ///
+    /// ```
+    /// # use google_cloud_auth::credentials::idtoken;
+    /// let audience = "https://my-service.a.run.app";
+    /// let credentials = idtoken::mds::Builder::new(audience)
+    ///     .with_format("full") // licenses only works with format = "full"
```
coryan: I thought format used enums now?

src/auth/src/credentials/idtoken/user_account.rs
```diff
@@ -112,6 +113,23 @@ impl Builder {
...
+    ///
+    /// ```
+    /// # use google_cloud_auth::credentials::idtoken;
+    /// # use serde_json::json;
+    /// # let authorized_user = json!({
+    /// #     "client_id": "YOUR_CLIENT_ID.apps.googleusercontent.com",
+    /// #     "client_secret": "YOUR_CLIENT_SECRET",
+    /// #     "refresh_token": "YOUR_REFRESH_TOKEN",
+    /// #     "type": "authorized_user",
+    /// # });
```
coryan: Adding all these details and then hiding them makes no sense to me.  Either say something like `let authorized_user: serde_json::Value = hidden_function();` or show the details.

src/auth/src/credentials/idtoken/verifier.rs
```diff
@@ -38,6 +38,7 @@
 
 use crate::credentials::internal::jwk_client::JwkClient;
 use jsonwebtoken::Validation;
+/// Represents the claims in an ID token.
 pub use serde_json::{Map, Value};
```
coryan: If you want to re-export and comment you should use different comments for each, no?

coryan (changes requested):
Adding a lot of details in the example and then hiding the details does not make sense to me. We should do one or the other.

alvarowolfx:
> Adding a lot of details in the example and then hiding the details does not make sense to me. We should do one or the other.

@coryan good points, I moved the full example with the credential format in the Builder example. The Builder methods have a shorter version of it, with just `add details here`. This is more in line with the docs on the other `Credentials`. Also fixed the MDS thing (I thought that this was going to be captured by docs being build 🤔 )

PR #3819, opened 2025-11-17
docs(auth): add idtoken samples

alvarowolfx:
Splitting work on https://github.com/googleapis/google-cloud-rust/pull/3762 that can already be merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7548,7 +7548,7 @@ dependencies = [
  "integration-tests",
  "mockall",
  "rand 0.9.2",
- "reqwest",
+ "reqwest 0.12.24",
  "serde_json",
  "storage-samples",
  "tempfile",


guide/samples/tests/driver.rs
```diff
@@ -24,6 +24,17 @@ mod driver {
         Ok(())
     }
 
+    #[cfg(all(test, google_cloud_unstable_id_token))]
+    #[tokio::test(flavor = "multi_thread")]
+    async fn id_token() -> anyhow::Result<()> {
+        let audience = "https://my-service.a.run.app";
+        let id_token =
+            user_guide_samples::authentication::request_id_token::sample(audience).await?;
+        user_guide_samples::authentication::verify_id_token::sample(id_token.as_str(), audience)
```
dbolduc: nit: save some characters with `&id_token`?

guide/samples/src/authentication/request_id_token.rs
```diff
@@ -0,0 +1,53 @@
...
+// * `credentials`: The IDTokenCredentials to use for authentication.
+pub async fn api_call_with_id_token(
+    target_url: &str,
+    credentials: &google_cloud_auth::credentials::idtoken::IDTokenCredentials,
+) -> anyhow::Result<()> {
+    use reqwest;
+
+    let id_token = credentials.id_token().await?;
+    let client = reqwest::Client::new();
+    client.get(target_url).bearer_auth(id_token).send().await?;
```
dbolduc, line 48: I remain skeptical about using an ID token in a `Authorization: Bearer <id_token>` header.

Remove this fn for now until we can call it from the CI?

guide/samples/src/authentication/request_id_token.rs
```diff
@@ -0,0 +1,53 @@
...
+// # Parameters
+// * `audience`: The audience for the ID token.
+pub async fn sample(audience: &str) -> anyhow::Result<String> {
+    // [END rust_auth_request_id_token_parameters] ANCHOR_END: request_id_token_parameters
+    // [START rust_auth_request_id_token_use] ANCHOR: request_id_token_use
+    use google_cloud_auth::credentials::idtoken::Builder;
+    // [END rust_auth_request_id_token_use] ANCHOR_END: request_id_token_use
+
+    // [START rust_auth_request_id_token_credentials] ANCHOR: request_id_token_credentials
+    let credentials = Builder::new(audience).build()?;
```
dbolduc, line 26: comment: the credentials have a single audience, but the verifier accepts a list of audiences

(Probably makes sense. I definitely missed the discussion. Just calling it out)

alvarowolfx: that's correct, we can only generate ID Tokens for a given audience, but a verifier can verify for multiple audiences.

guide/samples/src/authentication/request_id_token.rs
alvarowolfx: why you are skeptical about that ? The main use case for ID Tokens is to call a target service and use the ID Token in the `Authorization` header. Samples and SDK from other language even have a custom client that puts the ID Token in the header automatically. See example in go https://cloud.google.com/run/docs/authenticating/service-to-service#run-service-to-service-example-go

dbolduc: > See example in go https://cloud.google.com/run/docs/authenticating/service-to-service#run-service-to-service-example-go

Thanks for sharing. I think my confusion was that Google Cloud services will never accept an ID token like that. But as you say, some other service might.

PR #3821, opened 2025-11-17
feat(auth): make idtoken stable

alvarowolfx:
Towards #3449

### DIFF
--- a/guide/samples/tests/storage.rs
+++ b/guide/samples/tests/storage.rs
@@ -29,6 +29,7 @@ pub mod storage {
     mod driver {
         use super::*;
 
+        #[ignore = "TODO(#3916) - disabled because it is flaky"]
         #[tokio::test(flavor = "multi_thread")]
         async fn quickstart() -> anyhow::Result<()> {
             let project_id = std::env::var("GOOGLE_CLOUD_PROJECT").unwrap();
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -15,13 +15,13 @@
 use crate::build_errors::Error as BuilderError;
 use crate::constants::GOOGLE_CLOUD_QUOTA_PROJECT_VAR;
 use crate::errors::{self, CredentialsError};
+use crate::token::Token;
 use crate::{BuildResult, Result};
 use http::{Extensions, HeaderMap};
 use serde_json::Value;
 use std::future::Future;
 use std::sync::Arc;
 use std::sync::atomic::{AtomicU64, Ordering};
-
 pub mod anonymous;
 pub mod api_key_credentials;
 pub mod external_account;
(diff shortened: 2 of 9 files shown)


src/auth/src/credentials/idtoken/mds.rs
```diff
@@ -161,11 +161,13 @@ impl Builder {
     ///
     /// ```
     /// # use google_cloud_auth::credentials::idtoken;
+    /// # tokio_test::block_on(async {
```
coryan: Why the async block if there are no `.await` calls? And do you really need to compile and **run** these examples? Or would it be enough to put them in an asynchronous function so they are compiled?

src/auth/src/credentials/idtoken.rs
```diff
@@ -53,22 +53,24 @@
 //! ## Example: Verifying an ID token
 //!
 //! Within the receiving private service, you can parse the authorization header to
-//! receive the information being sent by the Bearer token and use the [Verifier] to
+//! receive the information being sent by the Bearer token and use the Verifier to
```
coryan: If you drop the link, then either "\`Verifier\`" or "verifiier" ?

coryan: Could also be `[Verifier][verifier::Verifier]` if you need the link to work.

src/auth/Cargo.toml
```diff
@@ -46,6 +46,7 @@ thiserror.workspace   = true
 time                  = { workspace = true, features = ["serde"] }
 tokio                 = { workspace = true, features = ["fs", "process"] }
 bon.workspace         = true
+jsonwebtoken          = { workspace = true, features = ["rust_crypto"] }
```
coryan, line 49: Does this reintroduce the warnings about `rsa` and big numbers being unsupported in future versions of Rust?

If so, what are we planning to do until `rsa` adopts a new bigint library in a release (excluding preview releases)?

alvarowolfx: I think the issue was fixed by `num-bigint-dig`, they release a patch release v0.8.6 with a fix:
https://github.com/dignifiedquire/num-bigint/commit/e6096dde470d41d419a9f72ec8c19cbfe00052db

src/auth/src/credentials/idtoken.rs
alvarowolfx: TIL

src/auth/src/credentials/idtoken/mds.rs
alvarowolfx: removed some of the async blocks and replaced with `no_run` and simple functions when needed

src/auth/src/credentials/idtoken/impersonated.rs
```diff
@@ -138,19 +138,19 @@ impl Builder {
     /// Target principal is the email of the service account to impersonate.
     ///
     /// # Example
-    /// ```
+    /// ```no_run
```
coryan: FWIW, `no_run` and putting the example inside the function is overkill.  The code will "run" but the function is not called, it is just defined.

coryan: And if the code does not need `async` there is no need to put this into a function.

src/auth/src/credentials/idtoken/impersonated.rs
```diff
@@ -138,19 +138,19 @@ impl Builder {
...
     /// # use google_cloud_auth::credentials::idtoken;
     /// # use google_cloud_auth::credentials::user_account;
     /// # use serde_json::json;
     /// #
-    /// # tokio_test::block_on(async {
+    /// # fn example() -> Result<(), anyhow::Error> {
```
coryan, line 146:
```suggestion
```

src/auth/src/credentials/idtoken/impersonated.rs
```diff
@@ -138,19 +138,19 @@ impl Builder {
...
     /// let source_credentials = user_account::Builder::new(json!({ /* add details here */ })).build()?;
     ///
     /// let audience = "https://my-service.a.run.app";
     /// let credentials = idtoken::impersonated::Builder::from_source_credentials(audience, "test-principal", source_credentials)
     ///     .build();
     /// # Ok::<(), anyhow::Error>(())
-    /// # });
+    /// # }
```
coryan, line 153:
```suggestion
```

src/auth/src/credentials/idtoken/impersonated.rs
alvarowolfx: I removed the `no_run`, but had to leave the function because this line is raising the Err with `?`:
```
let source_credentials = user_account::Builder::new(json!({ /* add details here */ })).build()?
```

.gcb/builds/triggers/main.tf
```diff
@@ -36,7 +36,6 @@ locals {
 
   unstable_flags = join(" ", [
     "--cfg google_cloud_unstable_tracing",
-    "--cfg google_cloud_unstable_id_token",
```
coryan, line 39: Remember to update the triggers (`terraform apply`) after merging these changes.

src/auth/src/credentials/idtoken/verifier.rs
```diff
@@ -20,17 +20,17 @@
 //!
 //! ## Example: Verifying an ID token
 //!
-//! ```no_run
+//! ```
 //! # use google_cloud_auth::credentials::idtoken;
+//! # use google_cloud_auth::credentials::idtoken::verifier::Verifier;
 //! # use std::time::Duration;
 //! let audience = "https://my-service.a.run.app";
 //! let verifier = idtoken::verifier::Builder::new([audience]).build();
```
coryan, line 28: It seems like this test executes some non-trivial code, please merge and (if needed) clean up afterwards .

src/auth/src/credentials/idtoken.rs
```diff
@@ -53,22 +53,24 @@
...
-//! check if is valid.
+//! receive the information being sent by the Bearer token and use the
+//! [Verifier][verifier::Verifier] to check if the token is valid.
 //!
-//! ```no_run
+//! ```
 //! # use google_cloud_auth::credentials::idtoken;
+//! # use google_cloud_auth::credentials::idtoken::verifier::Verifier;
 //! # use std::time::Duration;
+//! # tokio_test::block_on(async {
```
coryan, line 63: Ditto. Please fix (if needed) afterwards.

PR #3822, opened 2025-11-17
test(o11y): End-to-end HTTP to Cloud Trace integration test

westarle:


src/integration-tests/src/observability/cloud_trace.rs
```diff
@@ -0,0 +1,164 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use auth::credentials::{CacheableResource, Credentials, EntityTag};
+use http::Extensions;
+use std::time::Duration;
+
+/// A client for the Google Cloud Trace API (v1 REST).
+///
+/// It uses the v1 rest API for simplicity since there is no generated client that can get traces.
```
dbolduc: https://github.com/googleapis/google-cloud-rust/pull/3823

src/integration-tests/src/observability/cloud_trace.rs
```diff
@@ -0,0 +1,164 @@
...
+
+        let mut cached_token: Option<String> = None;
+        let mut last_etag: Option<EntityTag> = None;
+
+        for i in 1..=max_retries {
+            let token = self.get_token(&mut cached_token, &mut last_etag).await?;
+
+            let resp = self
+                .http_client
+                .get(&url)
```
dbolduc: comment: I remember this sucking.

As you noticed, it takes a while for the trace to become available, and the RPC has a quota on it, so we cannot just retry willy-nilly.

This is the loop I ended up with in C++, which starts with an initial sleep of 10s (yuck!):

https://github.com/googleapis/google-cloud-cpp/blob/f928005d4b02cec2d1896fe3575ec44170a0042d/google/cloud/opentelemetry/integration_tests/trace_exporter_integration_test.cc#L99-L107

westarle: Thank you, added that to https://github.com/googleapis/google-cloud-rust/pull/3822/commits/3445e75f44a19271d74315813e1ec9bd458e78bc

src/integration-tests/src/observability/cloud_trace.rs
westarle: Thank you!! Deleted!

src/integration-tests/tests/telemetry.rs
```diff
@@ -0,0 +1,173 @@
...
+        // Force flush
+        for _ in 0..5 {
+            let _ = provider.force_flush();
+            tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+        }
+        println!("Spans flushed.");
+
+        // 5. Verify (Poll Cloud Trace API)
+        // Configure retry policy for NOT_FOUND errors
+        let retry_policy = RetryNotFound;
```
dbolduc: FYI: you can extend this and limit the attempt count and/or operation timeout with `RetryPolicyExt`

https://github.com/googleapis/google-cloud-rust/blob/dcd8ce07ec48d9fcc4761c226de6aefca008fab7/guide/samples/src/pagination.rs#L29-L33

src/integration-tests/tests/telemetry.rs
```diff
@@ -0,0 +1,173 @@
...
+        // Configure backoff policy (initial 10s, max 120s, scaling 6.0)
+        // This roughly matches the previous manual loop: 10, 60, 120...
```
dbolduc: FYI: it's not quite the same. It makes the first attempt after 0 seconds, vs. after waiting 10s.

And the subsequent backoffs are picked uniformly from [0, 10s], [0, 60s], [0, 120s], [0, 120s] vs. picking the max of that range.

It might be easiest to use the default retry, backoff policies, but with your own retry loop around the GAPIC.

westarle: ooh thanks, updated and its way easier to understand.

src/integration-tests/tests/telemetry.rs
westarle: removed for a simple loop

westarle:
This passes now (needed retry on 500's and maybe less wonky retry policy.)

src/integration-tests/tests/telemetry.rs
```diff
@@ -0,0 +1,167 @@
...
+        let _guard = tracing_subscriber::Registry::default()
+            .with(integration_tests::observability::tracing::layer(
+                provider.clone(),
+            ))
+            .set_default();
+
+        // Wait for credentials to be refreshed in the background task
+        println!("Waiting for credentials to initialize...");
+        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+
```
dbolduc: Yuck, we are going to ask applications to do this?

https://github.com/googleapis/google-cloud-rust/blob/ce7bb8ec559fb55bb96b9730b7e1d5cbfbb92350/src/integration-tests/src/observability/auth.rs#L53-L57

^ Maybe this should be async and await an initial token, which we send on the channel.

Then we hand the credentials and sender off to the refresh task.

src/integration-tests/tests/telemetry.rs
```diff
@@ -0,0 +1,167 @@
...
+        let span_name = "e2e-showcase-test";
+
+        // Start a root span
+        let root_span = tracing::info_span!("e2e_root", "otel.name" = span_name);
+        let trace_id = {
+            let _enter = root_span.enter();
+            let otel_ctx = root_span.context();
+            let otel_span_ref = otel_ctx.span();
+            let span_context = otel_span_ref.span_context();
+            let trace_id = span_context.trace_id().to_string();
```
dbolduc: style nit:

```rs
let trace_id = root_span.context().span().span_context().trace_id().to_string();
```

src/integration-tests/tests/telemetry.rs
```diff
@@ -0,0 +1,167 @@
...
+        tokio::time::sleep(std::time::Duration::from_secs(5)).await;
+
+        // 3. Generate Trace
+        let span_name = "e2e-showcase-test";
+
+        // Start a root span
+        let root_span = tracing::info_span!("e2e_root", "otel.name" = span_name);
+        let trace_id = {
+            let _enter = root_span.enter();
+            let otel_ctx = root_span.context();
```
dbolduc: TIL: `opentelemetry::trace::TraceContextExt`

src/integration-tests/tests/telemetry.rs
```diff
@@ -0,0 +1,167 @@
...
+            let otel_span_ref = otel_ctx.span();
+            let span_context = otel_span_ref.span_context();
+            let trace_id = span_context.trace_id().to_string();
+
+            // Initialize showcase client pointing to local mock server
+            // We use anonymous credentials for the *client* because it's talking to httptest
```
dbolduc: nit: the anonymous creds comment seems better anchored above `.with_credentials(...)`.

I would probably just omit it

src/integration-tests/tests/telemetry.rs
```diff
@@ -0,0 +1,167 @@
...
+            let endpoint = format!("http://{}", echo_server.addr());
+
+            let client = showcase::client::Echo::builder()
+                .with_endpoint(endpoint)
```
dbolduc: style nit: inline the endpoint?

```rs
            let client = showcase::client::Echo::builder()
                .with_endpoint(format!("http://{}", echo_server.addr()))
```

src/integration-tests/tests/telemetry.rs
```diff
@@ -0,0 +1,167 @@
...
+            trace_id, project_id
+        );
+
+        // 4. Flush
+        // Force flush
+        for _ in 0..5 {
+            let _ = provider.force_flush();
+            tokio::time::sleep(std::time::Duration::from_secs(1)).await;
+        }
+        println!("Spans flushed.");
```
dbolduc: This is ghastly. Why do we need 5 flushes?

Is it because any given flush might fail? If so, it looks like `force_flush` returns a `Result`. We could break if it is successful.

Is it because spans take a while to propagate? If so, we might as well sleep for all 5 seconds and then flush once. I still hate that, but it seems better.

src/integration-tests/tests/telemetry.rs
```diff
@@ -0,0 +1,167 @@
...
+
+        // 5. Verify (Poll Cloud Trace API)
+        let client = TraceService::builder().build().await?;
+
+        println!("Polling for trace...");
+        let backoff_delays = [10, 60, 120, 120, 120];
+        let mut trace = None;
+
+        for delay in backoff_delays {
+            println!("Waiting {}s before polling...", delay);
```
dbolduc: nit: add a comment saying why we are doing this. There is actually a good reason.

```rs
// Because we are limited by quota, start with a backoff.
```

tests/integration/tests/telemetry.rs
```diff
@@ -0,0 +1,167 @@
...
+        println!("Spans flushed.");
+
+        // 5. Verify (Poll Cloud Trace API)
+        let client = TraceService::builder().build().await?;
+
+        println!("Polling for trace...");
+        let backoff_delays = [10, 60, 120, 120, 120];
+        let mut trace = None;
+
+        for delay in backoff_delays {
```
dbolduc, line 103: nit: add a comment explaining why the custom loop.

```rs
// Traces can take several minutes to propagate after they have been written. Implement a generous retry loop to account for this.
```

src/integration-tests/tests/telemetry.rs
dbolduc: We at least need a comment explaining why we are doing this.

dbolduc (commented):
Looks mostly good, but some things are strange... mainly the initial sleep and the multiple flushes.

src/integration-tests/tests/telemetry.rs
westarle: great idea, looks much nicer now.

src/integration-tests/tests/telemetry.rs
westarle: done

src/integration-tests/tests/telemetry.rs
westarle: deleted this and some other comment fluff.

src/integration-tests/tests/telemetry.rs
westarle: done

src/integration-tests/tests/telemetry.rs
westarle: This was an artifact from my debugging -- thanks for catching it; now we rely on the built-in delay before first read.

tests/integration/tests/telemetry.rs
westarle: done, thanks!

tests/integration/src/observability/auth.rs
```diff
@@ -50,9 +50,15 @@ pub struct CloudTelemetryAuthInterceptor {
...
-    pub fn new(credentials: Credentials) -> Self {
-        let (tx, rx) = watch::channel(None);
+    pub async fn new(credentials: Credentials) -> Self {
+        let (tx, mut rx) = watch::channel(None);
         tokio::spawn(refresh_task(credentials, tx));
+
+        // Wait for the first refresh to complete.
+        // We ignore the result because if the sender is dropped (unlikely),
+        // the interceptor will just fail requests, which is the correct behavior.
+        let _ = rx.changed().await;
```
dbolduc, line 60: We only send headers in the case of a success, so this hangs on bad credentials, right?

https://github.com/googleapis/google-cloud-rust/blob/79bc61ad09d680f74ad436fe3e6492677f600edf/src/integration-tests/src/observability/auth.rs#L129-L132

So should we:

```rs
match headers {
  // omitted...
  Err(e) if e.is_transient() => {
     tracing::warn!("Failed to refresh GCP credentials: {e:?}"); 
     sleep(ERROR_RETRY_DELAY).await; 
  },
  Err(e) => break,
}
```

dbolduc: Also, seems like this can't get hit anymore:

https://github.com/googleapis/google-cloud-rust/blob/4d8249a77272ee2cb4c40a1e87009f9b84a8d68e/src/integration-tests/src/observability/auth.rs#L65-L68

dbolduc: This is getting complicated enough that we should probably defer it in a separate PR.

src/integration-tests/tests/telemetry.rs
```diff
@@ -0,0 +1,152 @@
...
+        // explicitly drop the span to end it
+        drop(root_span);
+
+        println!(
+            "View generated trace in Console: https://console.cloud.google.com/traces/explorer;traceId={}?project={}",
+            trace_id, project_id
+        );
+
+        // 4. Force flush to ensure spans are sent.
+        let _ = provider.force_flush();
```
dbolduc: Q: can we use `?`. Are you trying to avoid workaround `Timeout` errors? https://docs.rs/opentelemetry_sdk/latest/opentelemetry_sdk/error/enum.OTelSdkError.html#variant.Timeout

dbolduc (approved):
I think the initial headers for the interceptor needs more thought, but maybe not for this PR.

coryan:
@westarle this is not going to auto-merge, it has conflicts. Please rebase and resolve.

westarle:
Sorry for the delay, the change to handle retryable errors triggered an infinite loop in a test that was meant to fail on error credentials (changed retryable arg from true to false).

src/integration-tests/tests/telemetry.rs
westarle: done

tests/integration/src/observability/auth.rs
westarle: I took your advice in this PR.

tests/integration/src/observability/auth.rs
```diff
@@ -139,6 +149,12 @@ mod tests {
     #[tokio::test]
     async fn test_interceptor_injects_headers() {
         let (tx, rx) = watch::channel(None);
+        // Manually construct because new() spawns a task we don't want here,
+        // or we could just use new() and let it spawn.
+        // But since we want to control the channel, we construct manually as before.
+        // Wait, the previous test code manually constructed it:
+        // let mut interceptor = CloudTelemetryAuthInterceptor { rx };
+        // So we don't need to change this test if it doesn't use new().
```
dbolduc, line 157: can you clean up this comment? :shaking_face:

westarle: oops, done.

PR #3823, opened 2025-11-17
feat(trace/v1): generate library

dbolduc:
This client can list traces. The trace v2 client cannot do that. We need it.


src/generated/devtools/cloudtrace/v1/.sidekick.toml
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+[general]
+specification-source = 'google/devtools/cloudtrace/v1'
+service-config = 'google/devtools/cloudtrace/v1/cloudtrace_v1.yaml'
+
+[codec]
+version        = '1.0.0'
+copyright-year = '2025'
+package-name-override = 'google-cloud-trace-v1'
```
coryan, line 22: Should we make this a "not for publication" library?  If the only motivation is our own tests.

src/generated/devtools/cloudtrace/v1/.repo-metadata.json
```diff
@@ -0,0 +1,11 @@
+{
+    "api_id": "cloudtrace.googleapis.com",
+    "api_shortname": "cloudtrace",
+    "client_documentation": "https://docs.rs/google-cloud-trace-v1/latest",
+    "distribution_name": "google-cloud-rust",
+    "language": "rust",
+    "library_type": "GAPIC_AUTO",
+    "name_pretty": "Stackdriver Trace API",
```
westarle, line 8: I think it should be Cloud Trace API (https://docs.cloud.google.com/trace/docs/reference/v1/rest)

src/generated/devtools/cloudtrace/v1/.sidekick.toml
dbolduc: > Should we make this a "not for publication" library?

No.

> If the only motivation is our own tests.

Our customers should also be able to list/get traces. No?

src/generated/devtools/cloudtrace/v1/.repo-metadata.json
dbolduc: Agreed.

Note that this is generated from the service yaml:

https://github.com/googleapis/googleapis/blob/7c0dcbba70fc5dd64655a77e74dbbf8aaf04c1bf/google/devtools/cloudtrace/v1/cloudtrace_v1.yaml#L4

I think I am too lazy to send a CL.

coryan: You can override this in the .sidekick.toml file.

src/generated/devtools/cloudtrace/v1/.sidekick.toml
coryan: Ack.

src/generated/devtools/cloudtrace/v1/.repo-metadata.json
dbolduc: Ack, Cloud Trace v2 could also use this treatment. I will merge this PR as is, then send a follow up to rename both.

dbolduc: #3825

PR #3826, opened 2025-11-17
docs(guide): move chapters to c.g.c

coryan:


guide/book.toml
```diff
@@ -21,3 +21,8 @@ src          = "src"
 
 [rust]
 edition = "2021"
+
+[output.html.redirect]
```
dbolduc, line 25: Seems like there is not a good way back to the rest of the user guide. Do we care?

dbolduc (approved):

guide/book.toml
coryan: I don't think we care. Eventually (soon?) all the content moves to `docs.cloud.google.com/rust` and we just leave a few redirects.

Issue #3543, opened 2025-10-15
Update minimal-versions tests to check against workspace dependencies as well


suzmue:
The minimal versions of the workspace crates use the most recent, even when the minimal specified version is lower.

Update the minimal version tests to work against the actual minimal versions specified. The `--detach-path-deps` flag on `cargo minimal-versions` may be helpful.

suzmue:
We have added a step before running `cargo minimal-versions check` that:
- Removes the path = "..." entries from the Cargo.toml for the duration of the CI run.
- Identifies which local crates are specified in the workspace Cargo.toml but are not yet published
- Adds these crates to a .cargo/config.toml file with a [[patch]](https://doc.rust-lang.org/cargo/reference/overriding-dependencies.html) section to point to the local files.
- Checks that the version for these patched packages matches the version in the Cargo.toml for that package.

This allows us to selectively swap out our unpublished workspace dependencies with the local code and verify that the version that is being tested against is the version that is planned to be released.

PR #3830, opened 2025-11-17
ci: add helper tool for minimal versions

suzmue:
This PR introduces the `minimal-version-helper` tool to streamline our CI process to more closely simulate a release environment: its `prepare` command temporarily removes `path` dependencies, generates `[patch.crates-io]` for unpublished crates, and enforces a "minimal match" versioning convention.

The minimal versions tests are also updated to not run against local testing crates and unstable configurations. It doesn't make sense to run these against the testing and sample crates, since we are not publishing these on crates.io, so minimal version selection is unimportant. We also don't make guarantees for our unstable flags, so we want to exclude these from these checks as well.

For #3543

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -27,54 +27,10 @@ dependencies = [
 ]
 
 [[package]]
-name = "anstream"
-version = "0.6.21"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "43d5b281e737544384e969a5ccad3f1cdd24b48086a0fc1b2a5262a26b8f4f4a"
-dependencies = [
- "anstyle",
- "anstyle-parse",
- "anstyle-query",
- "anstyle-wincon",
- "colorchoice",
- "is_terminal_polyfill",
- "utf8parse",
-]
-
-[[package]]
 name = "anstyle"
 version = "1.0.13"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
-
-[[package]]
-name = "anstyle-parse"
-version = "0.2.7"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4e7644824f0aa2c7b9384579234ef10eb7efb6a0deb83f9630a49594dd9c15c2"
-dependencies = [
- "utf8parse",
-]
-
-[[package]]
-name = "anstyle-query"
-version = "1.1.5"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
-dependencies = [
- "windows-sys 0.61.2",
-]
-
-[[package]]
-name = "anstyle-wincon"
-version = "3.0.11"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
-dependencies = [
- "anstyle",
- "once_cell_polyfill",
- "windows-sys 0.61.2",
-]
 
 [[package]]
 name = "anyhow"
@@ -385,10 +341,8 @@ version = "4.5.52"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "02576b399397b659c26064fbc92a75fede9d18ffd5f80ca1cd74ddab167016e1"
 dependencies = [
- "anstream",
  "anstyle",
  "clap_lex",
- "strsim",
 ]
 
 [[package]]
@@ -408,12 +362,6 @@ name = "clap_lex"
 version = "0.7.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "a1d728cc89cf3aee9ff92b05e62b19ee65a02b5702cff7d5a377e32c6ae29d8d"
-
-[[package]]
-name = "colorchoice"
-version = "1.0.4"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "b05b61dc5112cbb17e4b6cd61790d9845d13888356391624cbe7e41efeac1e75"
 
 [[package]]
 name = "common"
@@ -5598,7 +5546,7 @@ dependencies = [
  "libc",
  "percent-encoding",
  "pin-project-lite",
- "socket2 0.6.1",
+ "socket2 0.5.10",
  "tokio",
  "tower-service",
  "tracing",
@@ -5829,12 +5777,6 @@ dependencies = [
  "memchr",
  "serde",
 ]
-
-[[package]]
-name = "is_terminal_polyfill"
-version = "1.70.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a6cb138bb79a146c1bd460005623e142ef0181e3d0219cb493e02f7d08a35695"
 
 [[package]]
 name = "itertools"
@@ -6159,12 +6101,6 @@ name = "once_cell"
 version = "1.21.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "42f5e15c9953c5e4ccceeb2e7382a716482c34515315f7b03532b8b4e8393d2d"
-
-[[package]]
-name = "once_cell_polyfill"
-version = "1.70.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "384b8ab6d37215f3c5301a95a4accb5d64aa607f1fcb26a11b5303878451b4fe"
 
 [[package]]
 name = "openssl"
@@ -6594,7 +6530,7 @@ dependencies = [
  "quinn-udp",
  "rustc-hash",
  "rustls",
- "socket2 0.6.1",
+ "socket2 0.5.10",
  "thiserror 2.0.17",
  "tokio",
  "tracing",
@@ -6631,7 +6567,7 @@ dependencies = [
  "cfg_aliases",
  "libc",
  "once_cell",
- "socket2 0.6.1",
+ "socket2 0.5.10",
  "tracing",
  "windows-sys 0.60.2",
 ]
@@ -8032,12 +7968,6 @@ name = "utf8_iter"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b6c140620e7ffbb22c2dee59cafe6084a59b5ffc27a8859a5f0d494b5d52b6be"
-
-[[package]]
-name = "utf8parse"
-version = "0.2.2"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "06abde3611657adf66d383f00b093d7faecc7fa57071cce2578660c9f1010821"
 
 [[package]]
 name = "uuid"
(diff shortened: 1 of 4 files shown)


tools/minimal-version-helper/src/main.rs
```diff
@@ -0,0 +1,229 @@
...
+    let mut root_manifest = fs::read_to_string(&root_manifest_path)?.parse::<Document>()?;
+
+    if is_local {
+        // 1. Backup original files
+        println!("--local flag detected. Backing up files.");
+        let cargo_backup_path = root_manifest_path.with_extension("toml.bak");
+        if !cargo_backup_path.exists() {
+            fs::copy(&root_manifest_path, &cargo_backup_path)?;
+            println!("Backed up Cargo.toml to Cargo.toml.bak");
+        }
```
coryan, line 68: This is fine, but we also have version control? `git restore Cargo.toml` can do the trick?

tools/minimal-version-helper/Cargo.toml
```diff
@@ -0,0 +1,13 @@
+[package]
+name = "minimal-version-helper"
+version = "0.0.0"
+edition = "2021"
```
coryan: Why not 2024?

Cargo.lock
coryan, line 1: Looks like a lot of changes, do we understand why?

Cargo.toml
```diff
@@ -34,6 +34,7 @@ default-members = [
   "src/storage",
   "src/wkt",
   "tools/check-copyright",
+  "tools/minimal-version-helper",
```
coryan, line 37: This does not need to be in the `default-members` list (for that matter, neither does `check-copyright`).

tools/minimal-version-helper/Cargo.toml
```diff
@@ -0,0 +1,13 @@
...
+publish = false
+
+[dependencies]
+anyhow = "1.0.75"
+cargo_metadata = "0.18.0"
+clap = { version = "4.4.8", features = ["derive"] }
+crates_io_api = "0.9.0"
+semver = "1.0.21"
+toml_edit = "0.21.0"
```
coryan: At least some of these are in the top-level `Cargo.toml` should we move all of them there?

tools/minimal-version-helper/src/main.rs
```diff
@@ -0,0 +1,255 @@
...
+                                .to_string(),
+                            path: dep_table.get("path").unwrap().as_str().unwrap().to_string(),
+                        },
+                    );
+                    // Remove path from the Cargo.toml.
+                    dep_table.remove("path");
+                }
+            }
+        }
+    }
```
coryan, line 119: Refactor this to a function?

tools/minimal-version-helper/src/main.rs
```diff
@@ -0,0 +1,255 @@
...
+    println!("Removed path dependencies from Cargo.toml");
+
+    // 5. Generate a Patch File for unpublished crates
+    println!("Querying crates.io for unpublished crates...");
+    let client = crates_io_api::SyncClient::new(
+        "google-cloud-rust-ci (https://github.com/googleapis/google-cloud-rust)",
+        std::time::Duration::from_millis(1000),
+    )?;
+
+    let mut patch_content = String::new();
```
coryan, line 132: Seems most of one of these numbered steps can be separate functions?

tools/minimal-version-helper/Cargo.toml
```diff
@@ -0,0 +1,27 @@
...
+edition = "2021"
+publish = false
+
+[dependencies]
+anyhow         = "1.0.75"
+cargo_metadata = "0.18.0"
+clap           = { version = "4.4.8", features = ["derive"] }
+crates_io_api  = "0.9.0"
+semver         = "1.0.21"
+toml_edit      = "0.21.0"
```
coryan: Can we control these from the top-level Cargo file?

tools/minimal-version-helper/README.md
```diff
@@ -0,0 +1,76 @@
...
+   ```
+
+   *(Replace `your-crate-name` with the crate you are testing.)*
+
+1. **Revert the changes:** Once you are finished with your testing, run the
+   revert command to restore your workspace to its normal development state.
+
+   ```bash
+   cargo run -p minimal-version-helper -- revert
+   ```
```
coryan, line 76:
```suggestion
   ```

## Why not use Go?

Normally all tools are written in Go, but this needs to access the `crates.io` API, and the Cargo metadata. There are good Rust crates to build such a tool, not so many helpers to write it in Go.
```

tools/minimal-version-helper/src/main.rs
```diff
@@ -0,0 +1,255 @@
...
+    let mut path_deps = HashMap::new();
+    if let Some(deps) = root_manifest
+        .get_mut("workspace")
+        .and_then(|w| w.as_table_like_mut())
+        .and_then(|wt| wt.get_mut("dependencies"))
+        .and_then(|d| d.as_table_like_mut())
+    {
+        for (key, value) in deps.iter_mut() {
+            if let Some(dep_table) = value.as_table_like_mut() {
+                if dep_table.contains_key("path") && dep_table.contains_key("version") {
```
coryan, line 91:
```suggestion
            let Some(dep_table) = value.as_table_like_mut() else { continue; }
            if !dep_table.contains_key("path") || !dep_table.contains_key("version") { continue; }
            // and your code is not so deeply indented.
```

tools/minimal-version-helper/src/main.rs
```diff
@@ -0,0 +1,255 @@
...
+                    path_deps.insert(
+                        package_name.clone(),
+                        Package {
+                            version: dep_table
+                                .get("version")
+                                .unwrap()
+                                .as_str()
+                                .unwrap()
+                                .to_string(),
+                            path: dep_table.get("path").unwrap().as_str().unwrap().to_string(),
```
coryan: Maybe document the `unwrap()` with `expect()` explaining why these are safe.

tools/minimal-version-helper/src/main.rs
```diff
@@ -0,0 +1,255 @@
...
+    for (name, Package { version, path }) in &path_deps {
+        let required = VersionReq::parse(version)?;
+        match client.get_crate(name) {
+            Ok(crate_info) => {
+                if !crate_info.versions.iter().any(|v| {
+                    Version::parse(&v.num).is_ok_and(|remote_v| required.matches(&remote_v))
+                }) {
+                    println!("Found unpublished crate: {} v{}", name, version);
+                    patched.push(name);
+                    patch_content.push_str(&format!("{} = {{ path = \"{}\" }}\n", name, path));
```
coryan: optional:


```suggestion
                    patch_content.push_str(&format!(r#"{} = {{ path = "{}" }}\n"#, name, path));
```

assuming I did not botch the closing `"#` vs. `#"`

tools/minimal-version-helper/src/main.rs
```diff
@@ -0,0 +1,255 @@
...
+                }
+            }
+            Err(_) => {
+                // If crate doesn't exist at all, it's unpublished
+                println!("Found new crate: {} v{}", name, version);
+                patch_content.push_str(&format!("{} = {{ path = \"{}\" }}\n", name, path));
+            }
+        }
+    }
+    if !patch_content.is_empty() {
```
coryan, line 155: In a function you could do:


```suggestion
    if patch_content.is_empty() {
       println!("....");
       return;
     }
```

tools/minimal-version-helper/src/main.rs
```diff
@@ -0,0 +1,255 @@
...
+        //
+        // Patch version requirement, must match exactly
+        // - Cargo.toml = 1.4.1, package = 1.4.1 => true
+        // - Cargo.toml = 1.4.1, package = 1.4.2 => false
+        // Minor version requirement, patch is zero.
+        // - Cargo.toml = 1.4, package = 1.4.0 => true
+        // - Cargo.toml = 1.4, package = 1.4.1 => false
+        // Major version requirement, minor and patch are zero.
+        // - Cargo.toml = 2, package = 2.0.0 => true
+        // - Cargo.toml = 2, package = 2.0.1 => false
```
coryan, line 198: These look like a good set of tests? And the code below could be refactored to a function?

tools/minimal-version-helper/src/main.rs
```diff
@@ -0,0 +1,255 @@
...
+        // the updated version. If we are patching the crate, we want
+        // newly published crates to depend on this version. We want to
+        // make sure it is the minimum version (without requiring more
+        // specificity than necessary).
+        //
+        // Examples:
+        //
+        // Patch version requirement, must match exactly
+        // - Cargo.toml = 1.4.1, package = 1.4.1 => true
+        // - Cargo.toml = 1.4.1, package = 1.4.2 => false
```
coryan, line 192: Do we want to enforce anything about `0.x` versions?  Maybe they always need to be in `0.x.y` form? (because the `x` is effectively a major version?)

suzmue:
> If you want, we can merge as-is and refactor later. We need this tool working before the next release.

I would prefer to push refactoring to a future PR. This can definitely be better tested (and better written), but I think highest priority is to get it in to avoid another breaking release and we can keep iterating.

I'll clean up a couple of your comments since I need to rebase anyway to get the CI fix, but will leave the version testing (for example) as a future exercise.

tools/minimal-version-helper/src/main.rs
suzmue: This works if you have a clean workspace. But often when I was testing I would also edit the file and then try to test it and when I restored I dropped my change I had made.

I'll leave it for now, we can delete later if we want.

tools/minimal-version-helper/Cargo.toml
suzmue: Because copy-paste from check-copyright (aka no reason at all)

Cargo.toml
suzmue: I'll file an issue follow up, the commands we use in CI will need to be updated when we fix this. #3839

tools/minimal-version-helper/Cargo.toml
suzmue: inherited from workspace now.

tools/minimal-version-helper/src/main.rs
suzmue: Mostly this is "safe" because it is our internal tool and if it fails our Cargo.toml is wrong in some way so we should fix it and panicking is fine. added some minimal commentary in expect blocks

tools/minimal-version-helper/src/main.rs
suzmue: TY!

tools/minimal-version-helper/src/main.rs
suzmue: nice, would be a good refactor. Leaving it to a future code cleanup.

tools/minimal-version-helper/src/main.rs
suzmue: agreed, leaving it to the future.

tools/minimal-version-helper/src/main.rs
coryan: FWIW, you can edit the file, `git add` and then `git restore` will go back to the state when you called `git add` 🤷

tools/minimal-version-helper/src/main.rs
suzmue: Filed https://github.com/googleapis/google-cloud-rust/issues/3841

tools/minimal-version-helper/README.md
suzmue: Huh didn't see this in my review for some reason. Will add in a fast follow.

tools/minimal-version-helper/Cargo.toml
suzmue: Done.

PR #3832, opened 2025-11-17
test: WIP fake a change that breaks the minimal version testing

suzmue:
Adds a fake new feature to google-cloud-gax to see if our CI can catch this change.

DO NOT SUBMIT.


suzmue:
Didn't need the fake change, there is already a real one using gax-internal.

PR #3833, opened 2025-11-17
test: update gax-internal version

suzmue:
WIP.

Verifying that this would fix the breakage in https://github.com/googleapis/google-cloud-rust/actions/runs/19448186097/job/55647350073?pr=3830


suzmue:
This was for testing. Got the result of the minimal-versions run which was needed to confirm things were working as expected.

PR #3837, opened 2025-11-18
impl(storage): bidi streaming read worker

coryan:
Implement the background task reading from a bidi streaming read RPC, reconnecting to it on failures, and sending new range requests when needed.

Part of the work for #3626

### DIFF
--- a/src/storage/src/storage/bidi/worker.rs
+++ b/src/storage/src/storage/bidi/worker.rs
@@ -341,7 +341,7 @@ mod tests {
         assert!(matches!(got, Some(Ok(ref b)) if *b == content), "{got:?}");
         // The stream is not closed, so it does not have a `None` waiting.
         // We are willing to tolerate false positives here, it is fine if this
-        // succeeeds when it shouldn't due to race conditions.
+        // succeeds when it shouldn't due to race conditions.
         assert!(reader.is_empty());
 
         drop(tx);


coryan:
/FYI: @cjc25 @BenWhitehead this is the main loop reading data from the stream, sending new ranges to the stream, and reconnecting the stream as needed.

src/storage/src/storage/bidi/worker.rs
```diff
@@ -0,0 +1,733 @@
...
+            let mut pending = ranges
+                .lock()
+                .await
+                .remove(&range.read_id)
+                .ok_or(ReadError::UnknownBidiRangeId(range.read_id))?;
+            pending
+                .handle_data(response.checksummed_data, range, true)
+                .await
+        } else {
+            let mut guard = ranges.lock().await;
```
suzmue, line 199: aside: Can maybe consider using a more finegrained lock instead of blocking the whole map here

src/storage/src/storage/bidi/worker.rs
```diff
@@ -0,0 +1,733 @@
...
+        self.next_range_id += 1;
+
+        let request = reader.as_proto(id);
+        self.ranges.lock().await.insert(id, reader);
+        let request = BidiReadObjectRequest {
+            read_ranges: vec![request],
+            ..BidiReadObjectRequest::default()
+        };
+        // Any errors here are recovered by the main background loop.
+        if let Err(e) = tx.send(request).await {
```
suzmue, line 179: Errors here are recovered, does that insert has already happened? How do those end up resolving?

src/storage/src/storage/bidi/worker.rs
```diff
@@ -0,0 +1,733 @@
...
+        // return immediately.
+        loop {
+            tokio::select! {
+                m = rx.next_message() => {
+                    match self.handle_response(m).await {
+                        None => return Ok(()),
+                        Some(Err(e)) => return Err(e),
+                        Some(Ok(None)) => {},
+                        Some(Ok(Some(connection))) => {
+                            (rx, tx) = (connection.rx, connection.tx);
```
suzmue, line 75: This is clever I, like it.

nit: maybe comment to say this is a reconnection?

src/storage/src/storage/bidi/worker.rs
```diff
@@ -0,0 +1,733 @@
...
+        Some(Ok(Some(connection)))
+    }
+
+    async fn close_readers(&mut self, error: Arc<crate::Error>) {
+        let mut guard = self.ranges.lock().await;
+        let closing: Vec<_> = guard
+            .iter_mut()
+            .map(|(_, pending)| pending.interrupted(error.clone()))
+            .collect();
+        let _ = futures::future::join_all(closing).await;
```
suzmue: no action required: Since ordering doesn't matter and if this could be a large number of futures, [FuturesUnordered](https://docs.rs/futures/0.3.31/futures/prelude/stream/struct.FuturesUnordered.html) may be more efficient

src/storage/src/storage/bidi/worker.rs
```diff
@@ -0,0 +1,733 @@
...
+        let ranges = self.ranges.clone();
+        let pending = data
+            .into_iter()
+            .map(|r| Self::handle_range_data(ranges.clone(), r))
+            .collect::<Vec<_>>();
+        let _ = futures::future::join_all(pending)
+            .await
+            .into_iter()
+            .collect::<ReadResult<Vec<_>>>()
+            .map_err(crate::Error::io)?; // TODO: think about the error type
```
suzmue: This maps the first error found, do we care about any other errors that happened?

src/storage/src/storage/bidi/worker.rs
```diff
@@ -0,0 +1,733 @@
...
+
+        let connector = mock_connector(mock);
+        let worker = Worker::new(connector);
+        drop(tx);
+        worker.run(connection, rx).await?;
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn run_partial_read() -> anyhow::Result<()> {
```
suzmue, line 316: This test and run_full_read are very similar. It looks like the segment is different and the full read checks that the we expect to get None as a response after.

Can we combine these to share the code? Or possible add the comment about not expecting the stream to be done (would be great if we could confirm not None but that seems hard).

src/storage/src/storage/bidi/worker.rs
coryan: Done

src/storage/src/storage/bidi/worker.rs
coryan: (1) No, but I am adding some comments to that effect.

(2) I just realized this loop is doing the wrong thing and may deliver messages out of order.

coryan: Created #3848 to track the improvements.

src/storage/src/storage/bidi/worker.rs
coryan: I am not sure the number of futures is likely to be large. But it is easy enough to use.

src/storage/src/storage/bidi/worker.rs
coryan: Fixed the comment.

src/storage/src/storage/bidi/worker.rs
coryan: We can talk more about this because I may be misunderstand what you mean by fine-grained.  We need a mutex around the whole map, it is unsafe to read the map if some other thread may write to it.  

Maybe you meant that the critical section can be smaller? That I agree with. Done. With that change the lock is only held for the time required to update the data structure and clone the sender, maybe a few nanos.

src/storage/src/storage/bidi/worker.rs
coryan: > Can we combine these to share the code? 

Done.

> Or possible add the comment about not expecting the stream to be done (would be great if we could confirm not None but that seems hard).

A comment and a (weak) test.

src/storage/src/storage/bidi/worker.rs
suzmue: Yeah you're right. I was wondering if there were a way that we could do a per key mutex for certain operations (likely in addition to one for the whole map?). But seems like it wouldn't help us much even if it worked since we would still need to prevent writes to the whole map.

src/storage/src/storage/bidi/worker.rs
suzmue: nice.

PR #3846, opened 2025-11-19
impl(o11y): full body instrumentation for gRPC tracing

westarle:
implements full body instrumentation for gRPC network spans, extending the span duration to cover the entire response stream and enabling the capture of delayed gRPC status codes from trailers.

Design Rationale: I chose to use tower::util::Either (via TracedOrRawBody) to unify the return types of the tracing and non-tracing paths, rather than passing a boolean flag to a single TracingTowerLayer or using dynamic dispatch (Box<dyn Body>).   It did add a lot of kind of ugly "boilerplate" code.  

vs. a flag: NoTracingTowerService gives us confidence there's no tracing code on the no-tracing setup and the compiler might optimize it out.

vs. a Box: avoids dynamic dispatch and heap allocations.

### DIFF
--- a/src/gax-internal/src/observability/grpc_tracing.rs
+++ b/src/gax-internal/src/observability/grpc_tracing.rs
@@ -32,7 +32,7 @@ pub struct AttemptCount(pub i64);
 ///
 /// * `Either::Left`: The body is wrapped in `InstrumentedBody` (tracing enabled).
 /// * `Either::Right`: The body is the raw `B` (tracing disabled).
-pub type TracedOrRawBody<B> = http_body_util::Either<InstrumentedBody<B>, B>;
+pub type OptionallyTracedBody<B> = http_body_util::Either<InstrumentedBody<B>, B>;
 
 /// A wrapper around the response body that keeps the span active while streaming.
 #[pin_project::pin_project]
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -25,6 +25,59 @@ use tower::{Layer, Service};
...
 pub struct AttemptCount(pub i64);
 
+/// A type alias for the response body that can be either an instrumented body or a raw body.
+///
+/// This allows us to return a single type from both the tracing and non-tracing paths
+/// without using `Box<dyn Body>`, which avoids heap allocation and dynamic dispatch overhead.
+///
+/// * `Either::Left`: The body is wrapped in `InstrumentedBody` (tracing enabled).
+/// * `Either::Right`: The body is the raw `B` (tracing disabled).
+pub type TracedOrRawBody<B> = http_body_util::Either<InstrumentedBody<B>, B>;
```
coryan: `OptionallyTraced` ? `MaybeTraced` ?

src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -116,6 +172,78 @@ where
     }
 }
 
+/// A wrapper service that wraps the response body in `Either::Right` to match the `TracedOrRawBody` type.
```
coryan:
```suggestion
/// A service that wraps the response body in `Either::Right` to match the `TracedOrRawBody` type.
```

src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -170,6 +308,7 @@ fn record_response_status<ResBody>(span: &tracing::Span, response: &http::Respon
                 otel_trace::ERROR_TYPE,
                 ErrorType::from_gax_error(&gax_error).as_str(),
             );
+            span.record(OTEL_STATUS_DESCRIPTION, gax_error.to_string());
```
coryan: This is new?  Unrelated to the current change?

src/gax-internal/src/observability/grpc_tracing.rs
westarle: `OptionallyTraced` won!

src/gax-internal/src/observability/grpc_tracing.rs
westarle: done

src/gax-internal/src/observability/grpc_tracing.rs
westarle: removed!

westarle:
> I agree with your decisions here, I wonder if we should document them in a brief one pager.

[go/cloud-rust:grpc-tracing](http://goto.google.com/cloud-rust:grpc-tracing)

PR #3847, opened 2025-11-19
feat(aiplatform/v1): generate `schema` crates

### LINKED ISSUE #983: Generate libraries for `google/cloud/aiplatform/v1/schema/*`
These cannot go under `src/generated/cloud/aiplatform/v1` because that is not how Cargo works. We can put them in `src/generated/cloud/aiplatform/schema/v1` or in `src/generated/cloud/aiplatform/v1-schema` or any number of other locations.

@codyoss asked a more general question: how do we deal with all these corner cases. I am hoping that there are only a handful or so.

coryan:
Fixes #983


src/generated/cloud/aiplatform/schema/trainingjob/definition/.sidekick.toml
```diff
@@ -0,0 +1,20 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[general]
+specification-source = 'google/cloud/aiplatform/v1/schema/trainingjob/definition'
+service-config = 'google/cloud/aiplatform/v1/schema/aiplatform_v1.yaml'
+
+[codec]
```
dbolduc, line 19: comment: oh nice, I forgot that the `nosvc` template was automatically used by `sidekick`.

https://github.com/googleapis/librarian/blob/5888f4557c5ba7dccaad03e23f900d68b83c871c/internal/sidekick/rust/generate.go#L87

src/generated/cloud/aiplatform/schema/predict/instance/.sidekick.toml
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+[general]
+specification-source = 'google/cloud/aiplatform/v1/schema/predict/instance'
+service-config = 'google/cloud/aiplatform/v1/schema/aiplatform_v1.yaml'
```
dbolduc, line 17: comment: Hm, not sure if I have seen the yaml outside the proto directory before.

coryan: You are right.  It happens to work, which I was glad to see. I think the *intent* was to use that single service config yaml for *all* the protos in its directory (recursively). In Rust we require "one protobuf namespace per crate", so we need to reuse it for more then one crate.

PR #3851, opened 2025-11-20
ci: retry flaky downloads

coryan:
Downloading things somethings fails, using a retry loop to avoid the most common flakes.


coryan:
> just curious: any rhyme or reason behind `--retry-all-errors` or not?

I got an error about `--retry-all-errors` being an unknown flag, and the default is to just retry transient errors (including HTTP 5xx), so that seemed easier.

PR #3852, opened 2025-11-20
cleanup: fix links, newlines in docs

dbolduc:
First commit includes some generated changes, as an example. Second commit has the rest of the generated changes.


PhongChuong (approved):

src/generated/showcase/src/transport.rs
```diff
@@ -934,6 +934,16 @@ impl super::stub::Compliance for Compliance {
             })??;
         #[cfg(google_cloud_unstable_tracing)]
         let options = gax::options::internal::set_path_template(options, _path_template);
+        #[cfg(google_cloud_unstable_tracing)]
```
dbolduc, line 937: cc: @westarle

Issue #3843, opened 2025-11-18
Run minimal version tests against *all* crates that are scheduled to be published


suzmue:
This could be a part of the CI, or can be done as part of the release. But it probably should be done.

Something like:

```shell
$ cargo run -p minimal-version-helper prepare
$ cargo workspaces plan --skip-published | xargs -I PKG cargo minimal-versions check -p PKG
```

This should also be run with all features enabled most likely.

suzmue:
Running all minimal versions takes ~45 minutes on gcb: https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/02bb7ada-5c92-4b54-8e1d-026473188b39?project=rust-sdk-testing&e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=monitoring_api_staging

This is still not appropriate for a pre-submit, but perhaps a post merge test (and we already have a few )

coryan:
> Running all minimal versions takes ~45 minutes on gcb

I just clicked the retry build button. We use a (distributed) build cache on GCB and maybe it is not that bad the second time (I doubt it, but I can hope and we should try).

> but perhaps a post merge test (and we already have a few )

That seems fine. We do need a way to get notified of failures or some human needs to check their status from time to time.

coryan:
> > Running all minimal versions takes ~45 minutes on gcb
> 
> I just clicked the retry build button. We use a (distributed) build cache on GCB and maybe it is not that bad the second time (I doubt it, but I can hope and we should try).

The second build took 40m... I do not call that much of an improvement.

PR #3853, opened 2025-11-20
test: run all workspaces for minimal versions

suzmue:
Do not submit. Checking how long it takes to check everything in our ci.

For #3843


suzmue:
Closing this was for a test. If we want to do this gcb is probably more appropriate

PR #3855, opened 2025-11-21
impl(o11y): implement cancellation detection for gRPC spans

westarle:


src/gax-internal/tests/grpc_observability.rs
```diff
@@ -554,4 +554,90 @@ mod tests {
...
+        use opentelemetry_semantic_conventions::{attribute as otel_attr, trace as otel_trace};
+        use std::time::Duration;
+
+        let guard = TestLayer::initialize();
+
+        let mut config = google_cloud_gax_internal::options::ClientConfig::default();
+        config.tracing = true;
+        config.cred = Some(test_credentials());
+
+        // Use a a real but non-existentaddress to ensure the request hangs and stays Pending,
```
westarle:
```suggestion
        // Use a a real but non-existent address to ensure the request hangs and stays Pending,
```

PR #3902, opened 2025-11-24
ci: skip tools in MSRV builds

coryan:
We really do not need the tools to compile with the MSRV, and we are running into problems with package updates.


dbolduc (approved):

.gcb/scripts/test.sh
```diff
@@ -21,8 +21,16 @@ rustup show active-toolchain -v
 
 echo "RUSTFLAGS in test: $RUSTFLAGS"
 
+excluded=()
+if [[ "${GCB_TRIGGER_NAME:-}" == "*-test-msrv" ]]; then
```
coryan, line 25: Made a mistake here, fixed in #3903

PR #3905, opened 2025-11-24
feat(pubsub)!: Rename PublisherFactory to Client

PhongChuong:
This new naming is consistent with the Publisher design: go/cloud-rust:pubsub-publisher-design


PhongChuong:
> Approved, but FWIW, I think the current names make more sense.
> 
> The `client::Client` does not feel like a client to me. It is something that can make clients.
> 
> If we changed `client::publisher` (a noun) to something like `client::batch_publish` (a verb), it might feel more like a client.
> 
> Your call on how to proceed.

@dbolduc , noted. Thanks for the review. I'll merge this for now. We can revisit before the next release when Suzy is back.

PR #3909, opened 2025-11-24
refactor(pubsub): Move publisher Worker and Batch into its own file

PhongChuong:


src/pubsub/src/publisher/publisher.rs
coryan, line 1: I am not sure why, but `publisher/publisher.rs` seems odd to me.  No action needed in this PR.

PhongChuong: Agreed. I think it's also why pubsub::client::Client::... feels odd. We'll find a better naming convention before the next release.

PR #3911, opened 2025-11-24
tests: Mark some examples as compile_fail

amanda-tarafa:
We will mark generated examples as ignore because they take a long time to build. Still we want to test them weekly using `--ignored` which builds and runs all ignored examples.

The examples on this PR were previously marked as ignore because they don't build by design. If we keep them as are they will fail when we use `--ignored` to run ignore samples, so we mark them as compile_fail instead.

Towards #2808

### DIFF
--- a/src/gax-internal/src/observability/client_tracing.rs
+++ b/src/gax-internal/src/observability/client_tracing.rs
@@ -26,7 +26,7 @@ use tracing::{Span, field};
 ///
 /// # Example
 ///
-/// ```ignore
+/// ```compile_fail
 /// let span = create_client_request_span(
 ///     "google_cloud_storage::client::Client::upload_chunk",
 ///     "upload_chunk",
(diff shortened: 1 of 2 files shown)


amanda-tarafa:
With this change in, I've ran:

```
export RUSTDOCFLAGS='--cfg google_cloud_unstable_tracing --cfg google_cloud_unstable_id_token --cfg google_cloud_unstable_signed_url --cfg google_cloud_unstable_storage_bidi'
export RUSTFLAGS='--cfg google_cloud_unstable_tracing --cfg google_cloud_unstable_id_token --cfg google_cloud_unstable_signed_url --cfg google_cloud_unstable_storage_bidi'
cargo test --doc --workspace --all-features -- --ignored --show-output
```

and no failures are reported. We currently don't have any `ignore` documentation examples that would fail build or run if tested.

PR #3922, opened 2025-11-25
docs: remove scary comments about `nosvc` libraries

coryan:
The libraries without a service still had the pre-GA scary comments.
Fixed the generator to use the same comments we use for normal crates
with services.


coryan:
Blegh, I created double blank lines with this PR. Will fix sidekick, but I think this PR improves things as-is.

dbolduc (approved):

Issue #3748, opened 2025-11-09
Add byte threshold batching settings to publisher


suzmue:
The publisher will need to be able to:
1. Calculate the size of a message
2. Check if a batch exceeds the limits
3. Ensure that the MAX limit is not exceeded by a batch (a message cannot be added to the batch if it would make it larger than the quota limits).
4. 
See https://cloud.google.com/pubsub/docs/batch-messaging#quotas_and_limits_on_batch_messaging

PR #3925, opened 2025-11-26
feat(pubsub): add byte threshold to publisher

PhongChuong:
Implement byte threshold batching in Publisher. The batch size is the accumulation of the topic, messages data, ordering key and attribute length. All other fields are currently omitted in the size calculation.

Fixes: #3748

### DIFF
--- a/src/pubsub/src/publisher/batch.rs
+++ b/src/pubsub/src/publisher/batch.rs
@@ -52,10 +52,11 @@ impl Batch {
         // This is only an estimate and not the wire length.
         // TODO(#3963): If we move on to use protobuf crate, then it may be
         // possible to use compute_size to find the wire length.
-        msg
-            .attributes
+        msg.attributes
             .iter()
-            .fold(msg.data.len() + msg.ordering_key.len(), |acc, (k, v)| acc + k.len() + v.len())
+            .fold(msg.data.len() + msg.ordering_key.len(), |acc, (k, v)| {
+                acc + k.len() + v.len()
+            })
     }
 
     /// Drains the batch and spawns a task to send the messages.
@@ -115,11 +116,13 @@ impl Batch {
 
 #[cfg(test)]
 mod tests {
-    use crate::generated::gapic_dataplane::client::Publisher as GapicPublisher;
-    use crate::model::PubsubMessage;
-    use crate::publisher::batch::Batch;
-    use crate::publisher::worker::BundledMessage;
-    use futures::stream::FuturesUnordered;
+    use crate::{
+        generated::gapic_dataplane::client::Publisher as GapicPublisher,
+        model::{PublishResponse, PubsubMessage},
+        publisher::batch::Batch,
+        publisher::worker::BundledMessage,
+    };
+    use futures::{StreamExt, stream::FuturesUnordered};
 
     mockall::mock! {
         #[derive(Debug)]
@@ -131,7 +134,7 @@ mod tests {
 
     #[tokio::test]
     async fn test_push_and_flush_batch() {
-        let mut batch: Batch = Batch::new("topic");
+        let mut batch = Batch::new("topic");
         assert!(batch.is_empty());
 
         let (message_a, _rx_a) = create_bundled_message_from_bytes("hello");
@@ -146,12 +149,21 @@ mod tests {
         batch.push(message_c);
         assert_eq!(batch.len(), 3);
 
-        let mock = MockGapicPublisher::new();
+        let mut mock = MockGapicPublisher::new();
+        mock.expect_publish().return_once({
+            |r, _| {
+                assert_eq!(r.topic, "topic");
+                assert_eq!(r.messages.len(), 3);
+                Ok(gax::response::Response::from(PublishResponse::new()))
+            }
+        });
         let client = GapicPublisher::from_stub(mock);
         let mut inflight = FuturesUnordered::new();
         batch.flush(client, "topic".to_string(), &mut inflight);
-        assert_eq!(inflight.len(), 1);
         assert_eq!(batch.len(), 0);
+        assert_eq!(inflight.len(), 1);
+        // Apply pressure on inflight to validate the mock.
+        inflight.next().await.unwrap().unwrap();
     }
 
     #[tokio::test]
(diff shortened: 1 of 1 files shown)


src/pubsub/src/publisher/batch.rs
```diff
@@ -37,7 +37,17 @@ impl Batch {
...
 
+    pub(crate) fn size(&self) -> u32 {
+        self.messages_byte_size
+    }
+
     pub(crate) fn push(&mut self, msg: BundledMessage) {
+        use gaxi::prost::ToProto;
+        use prost::Message;
+        // TODO(NOW): Yuk. Can we avoid the clone here?
+        let proto = msg.msg.clone().to_proto().unwrap();
```
dbolduc: could just use the size of data... Or if we must, we could iterate over the attributes (as a ref) and add that onto the size in bytes

PhongChuong: Thanks for the suggestions. PTAL.

src/pubsub/src/publisher/batch.rs
```diff
@@ -37,10 +39,29 @@ impl Batch {
...
+    }
+
     pub(crate) fn push(&mut self, msg: BundledMessage) {
+        self.messages_byte_size += Self::message_size(&msg.msg) as u32;
         self.messages.push(msg);
     }
 
+    fn message_size(msg: &crate::model::PubsubMessage) -> usize {
+        // This is only an estimate and not the wire length.
+        // TODO(#3963): If we move on to use protobuf crate, then it may be
```
dbolduc, line 53: comment: I think if an application configures a batch size and we are within like 5% of that, that is probably good enough.

We should just remember that it is probably an under-estimate, which might come into play if there are request size limits.

src/pubsub/src/publisher/batch.rs
```diff
@@ -96,41 +119,102 @@ impl Batch {
...
+        impl crate::generated::gapic_dataplane::stub::Publisher for GapicPublisher {
+            async fn publish(&self, req: crate::model::PublishRequest, _options: gax::options::RequestOptions) -> gax::Result<gax::response::Response<crate::model::PublishResponse>>;
+        }
+    }
 
     #[tokio::test]
-    async fn test_push_batch() {
-        let mut batch = Batch::new();
+    async fn test_push_and_flush_batch() {
+        let mut batch: Batch = Batch::new("topic");
```
dbolduc: optional style nit: I don't tend to write the type inline unless I have to.

src/pubsub/src/publisher/batch.rs
```diff
@@ -96,41 +119,102 @@ impl Batch {
...
+        let (message_b, _rx_b) = create_bundled_message_from_bytes(", ");
         batch.push(message_b);
         assert_eq!(batch.len(), 2);
 
-        let (message_c, _rx_c) = create_bundled_message_helper("world".to_string());
+        let (message_c, _rx_c) = create_bundled_message_from_bytes("world");
         batch.push(message_c);
         assert_eq!(batch.len(), 3);
+
+        let mock = MockGapicPublisher::new();
```
dbolduc: should I be surprised that there are no expectations on the mock?

would we need to `expect_publish(...)` if we awaited the `inflight` futures? If so, we should probably do that.

Sorry, I am confused here.

src/pubsub/src/publisher/batch.rs
```diff
@@ -96,41 +119,102 @@ impl Batch {
...
+        let mut expected_encoded_len = topic.len();
+        assert_eq!(batch.size(), expected_encoded_len as u32);
+
+        let (message_data_only, _rx) = create_bundled_message_from_bytes("message_data_only");
+        expected_encoded_len += message_data_only.msg.data.len();
+        batch.push(message_data_only);
+        assert_eq!(batch.size(), expected_encoded_len as u32);
+
+        let (message_with_ordering, _rx) = create_bundled_message_from_pubsub_message(
+            PubsubMessage::new().set_ordering_key("ordering_key".to_string()),
```
dbolduc: nit: `to_string()` is unnecessary

src/pubsub/src/publisher/batch.rs
```diff
@@ -96,41 +119,102 @@ impl Batch {
...
+
+        let attributes = HashMap::from([("k1", "v1"), ("k2", "v2")]);
+        let (message_with_attributes, _rx) = create_bundled_message_from_pubsub_message(
+            PubsubMessage::new().set_attributes(attributes),
+        );
+        expected_encoded_len += message_with_attributes
+            .msg
+            .attributes
+            .iter()
+            .fold(0, |acc, (k, v)| acc + k.len() + v.len());
```
dbolduc: comment: This is bordering on a change detector test. It is kind of a smell to repeat function code in the test fixture. Consider something like:

```rust
fn test_data(len: usize) -> String {
  "A".repeat(len)
}
let attributes = HashMap::from([(test_data(10), test_data(1000)), (test_data(10), test_data(1000))]
```

Then comparing that the size is 2020 or something. (basically do the math by hand)

src/pubsub/src/publisher/publisher.rs
```diff
@@ -539,8 +559,52 @@ mod tests {
         }
     }
 
+    #[tokio::test]
+    async fn test_batching_send_on_byte_threshold() {
```
dbolduc, line 563: nice

dbolduc (commented):
Looks good. I think the only thing that ought to change is the mock in the `test_push_and_flush_batch` test.

But lmk if I am misunderstanding something.

src/pubsub/src/publisher/batch.rs
```diff
@@ -37,10 +39,29 @@ impl Batch {
...
+        // TODO(#3963): If we move on to use protobuf crate, then it may be
+        // possible to use compute_size to find the wire length.
+        let mut size = 0;
+        size += msg.data.len();
+        size += msg.ordering_key.len();
+        size += msg
+            .attributes
+            .iter()
+            .fold(0, |acc, (k, v)| acc + k.len() + v.len());
+        size
```
coryan: Maybe too clever?

```suggestion
        msg
            .attributes
            .iter()
            .fold(msg.data.len() + msg.ordering_key.len(), |acc, (k, v)| acc + k.len() + v.len())
```

src/pubsub/src/publisher/batch.rs
PhongChuong: Fixed.

src/pubsub/src/publisher/batch.rs
PhongChuong: Updated to include expectations. PTAL.

src/pubsub/src/publisher/batch.rs
PhongChuong: Fixed.

src/pubsub/src/publisher/batch.rs
PhongChuong: I updated to validate on a hand calculated size. I didn't include the test data method. PTAL.

PhongChuong:
Thanks for the review. PTAL.

dbolduc (approved):

src/pubsub/src/publisher/batch.rs
dbolduc: FTR, this comment was to avoid a race, where if the batch somehow flushed before the mock is dropped, the mock would have an unrecognized expectation. Super unlikely, but should still be avoided.

PR #3929, opened 2025-11-27
impl(storage): `open_object()` works

### LINKED ISSUE #3626: Bootstrap a storage client with support for bidi streaming reads

coryan:
The new function works and it has an integration tests. The function is not mockable, and some constructors and additional traits will be needed to support mocking. The types are good enough to write tests, and start refactoring towards public (and stable) APIs.

Fixes #3626

### DIFF
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -240,7 +240,7 @@ where
     /// # async fn sample(client: &Storage) -> anyhow::Result<()> {
     /// use google_cloud_storage::model_ext::ReadRange;
     /// let descriptor = client
-    ///     .object_object("projects/_/buckets/my-bucket", "my-object")
+    ///     .open_object("projects/_/buckets/my-bucket", "my-object")
     ///     .send()
     ///     .await?;
     /// // Print the object metadata


coryan:
@googleapis/rust-cloud-sdk PTAL

tests/integration/src/storage/bidi_read.rs
```diff
@@ -0,0 +1,51 @@
...
+        .write_object(
+            bucket_name,
+            "basic/source.txt",
+            String::from_iter((0..100_000).map(|_| 'a')),
+        )
+        .set_if_generation_match(0)
+        .send_unbuffered()
+        .await?;
+
+    println!("created bidi client: {client:?}");
```
PhongChuong: nit. why not "tracing::info!"?

src/storage/src/storage/client.rs
```diff
@@ -226,6 +230,44 @@ where
...
+    #[cfg(google_cloud_unstable_storage_bidi)]
+    /// Opens an object to read its contents using concurrent ranged reads.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_storage::client::Storage;
+    /// # async fn sample(client: &Storage) -> anyhow::Result<()> {
+    /// use google_cloud_storage::model_ext::ReadRange;
+    /// let descriptor = client
+    ///     .object_object("projects/_/buckets/my-bucket", "my-object")
```
PhongChuong: s/object_object/open_object?

I think doc test might not have picked this up because it's behind google_cloud_unstable_storage_bidi?

tests/integration/src/storage/bidi_read.rs
coryan: Fixed, thanks.

src/storage/src/storage/client.rs
coryan: Fixed. I agree there is something broken here:  #3937

PR #3930, opened 2025-11-27
impl(storage): extend read resume policy for gRPC

### LINKED ISSUE #3772: Extend recommended resume policy to support gRPC
We can close this issue when the `google_cloud_storage::read_resume_policy::Recommended` correctly supports gRPC reads (or at least bidi streaming reads).

## Background

When using JSON the only possible error *after* a read starts is `Error::io`.

When using gRPC we get more error types, including `Error::service()`.

The default read resume policy should handle the additional errors.

coryan:
Fixes #3772

### DIFF
--- a/src/storage/src/read_resume_policy.rs
+++ b/src/storage/src/read_resume_policy.rs
@@ -260,72 +260,80 @@ mod tests {
     #[test]
     fn recommended() {
         let policy = Recommended;
-        let r = policy.on_error(&ResumeQuery::new(0), transient());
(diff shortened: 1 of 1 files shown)


src/storage/src/read_resume_policy.rs
PhongChuong: nit. renaming this to be more specific might help.

coryan: thanks, PTAL

Issue #3934, opened 2025-12-01
Support enabling the 'aws_lc_rs' (instead of 'rust_cypto') feature on 'jsonwebtoken'


Aaron1011:
**Is your feature request related to a problem? Please describe.**
The `google-cloud-auth` crate currently unconditionally enables the `rust_crypto` feature on the `jsonwebtoken` crate:
https://github.com/googleapis/google-cloud-rust/blob/c77d92a4d32d9ec89ae5040d2b2ff67d4fc2cba1/src/auth/Cargo.toml#L49

Unfortunately, this ends up pulling in the `rsa` crate, which has a timing attack vulnerability (https://rustsec.org/advisories/RUSTSEC-2023-0071
See https://github.com/Keats/jsonwebtoken/issues/463 for more details.

Concretely, this results in an error from `cargo deny` when adding `google-cloud-auth` as a dependency, via the chain `google-cloud-auth -> jsonwebtoken -> rsa`

**Describe the solution you'd like**
It would be very useful for `google-cloud-auth` to expose features to control the backend used by the `jsonwebtoken` crate. Specifically, adding `rsa` and `aws_lc_rs` features, which forward to the corresponding `jsonwebtoken` features, would allow direct consumers of `google-cloud-auth` to avoid pulling in a dependency on `rsa`.

**Describe alternatives you've considered**
The `rsa` error from `cargo deny` could be explicitly ignored, or we could wait for a fix in the RSA crate. However, the vulnerability was reported in 2023, so it would be very useful to have a way of avoiding a dependency with a known timing attack.

coryan:
Thanks for the bug report.  We will figure out a way to avoid the `rsa` dependency. I also filed #3936 to prevent problems like this in the future.

coryan:
Closing.  With the changes related to #4207 it is now possible to use `google-cloud-auth` with `default-features = false` and `features = ["idtoken"]`. The application needs to enable the right jsonwebtoken backend in this case.

PR #3938, opened 2025-12-01
impl(auth): allow to change crypto backend on jwt lib

alvarowolfx:
Add `rsa` and `aws_lc` features in the auth crate, that changes crypto backend for checking JWTs.

Towards https://github.com/googleapis/google-cloud-rust/issues/3934


alvarowolfx:
not gonna work with `jsonwebtoken`, because our CI workflows test the case of enabling all features, which causes compile issues on the `jsonwebtoken` crate.

Issue #3615, opened 2025-10-22
Feature parity between `Storage` client and other clients


dbolduc:
The `google_cloud_storage::client::Storage` uses a `reqwest::Client` directly in its implementation instead of using `gaxi::http::ReqwestClient` like the GAPICs do.

I think I missed the discussion on why we chose to do this, but at a guess it is because `Storage` is the only client that does HTTP streaming, so there was no need to build that into the common gaxi layer.

The consequence is that we are missing some common features handled by the gaxi layer. Such as:
- Calculating and setting the `Host:` header.
  - This is **required** to use a client in a private network such as VPC-SC.
- Cloud client library instrumentation for HTTP transport.
- `User-Agent:` header.
- etc. (this list may grow over time too)

We will need to do some level of refactoring to resolve this. It looks like either:
- add streaming support to `gaxi::http::ReqwestClient`, and use that from within `google-cloud-storage`
- OR factor out configuration of a reqwest builder given options, and call into it from `google-cloud-storage`

joshuatants:
As indicated in https://github.com/googleapis/google-cloud-rust/pull/4839#pullrequestreview-3866472847 , closing this issue now that setting the user header is possible.

PR #3946, opened 2025-12-02
refactor: add streaming execution to ReqwestClient

westarle:
Adds `execute_streaming` to `ReqwestClient`. This allows executing requests
without the internal retry loop and supports streaming bodies, returning
`Response<impl Stream>`. Non-2xx responses (including 308) are converted
to Errors.

For #3615

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -306,7 +306,6 @@ crates_io_api         = { default-features = false, version = "0.12" }
 clap                  = { default-features = false, version = "4" }
 crc32c                = { default-features = false, version = "0.6.8" }
 futures               = { default-features = false, version = "0.3" }
-futures-util          = { default-features = false, version = "0.3" }
 http                  = { default-features = false, version = "1", features = ["std"] }
 http-body             = { default-features = false, version = "1" }
 http-body-util        = { default-features = false, version = "0.1.3" }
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/http.rs
```diff
@@ -110,17 +110,48 @@ impl ReqwestClient {
...
+            builder = builder.json(&body);
+        }
+        self.retry_loop::<O>(builder, options).await
+    }
+
+    pub async fn execute_streaming(
+        &self,
+        mut builder: reqwest::RequestBuilder,
+        options: gax::options::RequestOptions,
+    ) -> Result<Response<impl futures_util::Stream<Item = Result<bytes::Bytes>>>> {
```
coryan: Elsewhere, we use `futures::Stream<>` instead of `futures_util::Stream<>`, e.g.:

https://github.com/googleapis/google-cloud-rust/blob/71a6b620fa3913f4189696c1358a24717a6c4c7a/src/gax/src/paginator.rs#L225

Structurally both traits are identical, but they have different types. Any reason to prefer one over the other?

westarle: no real reason.  Also removed `futures-util` crate since we have `futures` in our workspace.

PR #3951, opened 2025-12-02
test(storage): bidi reads in random access benchmark

### LINKED ISSUE #3950: Extend the `storage-random` benchmark to use `open_object()`

coryan:
Fixes #3950

### DIFF
--- a/src/storage/benchmarks/random/src/bidi.rs
+++ b/src/storage/benchmarks/random/src/bidi.rs
@@ -1,0 +1,77 @@
+// Copyright 2025 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


src/storage/benchmarks/random/src/bidi.rs
```diff
@@ -0,0 +1,77 @@
...
+        };
+        let mut reader = descriptor
+            .read_range(ReadRange::segment(range.read_offset, range.read_offset))
+            .await;
+        let mut ttfb = None;
+        while reader.next().await.transpose()?.is_some() {
+            let _ = ttfb.get_or_insert(Instant::now() - start);
+        }
+        let ttlb = Instant::now() - start;
+        let ttfb = ttfb.unwrap_or(ttlb);
```
dbolduc, line 74: comment: I was wondering when the `or` happens, but I think the answer is an empty range, which could be requested, I guess.

dbolduc (approved):
I am kind of impressed by how minimal this PR is. Good job on the set up.

src/storage/benchmarks/random/src/bidi.rs
coryan: Maybe I am just paranoid and lazy, and yes, you are right.

PR #3955, opened 2025-12-02
impl(storage): optimize read range messages

### LINKED ISSUE #3940: Batch multiple `read_range()` requests to the service
Where possible, we should send a single message to the service for multiple `read_range()` requests. Saving networking overhead is a good thing.

We can close this bug when the code in:

https://github.com/googleapis/google-cloud-rust/blob/95bdc62f7448ffb183aada62de1be5a704e54a8c/src/storage/src/storage/bidi/worker.rs#L79-L85

uses some kind of `recv_many()` to fetch multiple ranges from the request and send a single message to the service.

coryan:
Fixes #3940

### DIFF
--- a/src/storage/src/storage/bidi/worker.rs
+++ b/src/storage/src/storage/bidi/worker.rs
@@ -404,7 +404,7 @@ mod tests {
 
         // Send 3 read requests in quick succession. The first one will block
         // until we pull a message from `response_tx`. That will force the second
-        // and third requests to be coallesced into a single BidiReadObjectRequest.
+        // and third requests to be coalesced into a single BidiReadObjectRequest.
         //
         // It is possible that all three reads are collaesced into a single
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/bidi/worker.rs
```diff
@@ -388,6 +393,90 @@ mod tests {
...
+            response_tx,
+            mut request_rx,
+        } = set_up_mock().await;
+
+        // Send 3 read requests in quick succession. The first one will block
+        // until we pull a message from `response_tx`. That will force the second
+        // and third requests to be coalesced into a single BidiReadObjectRequest.
+        //
+        // It is possible that all three reads are collaesced into a single
+        // `BidiReadObjectRequest` message.
```
dbolduc, line 410: thanks for the expected behavior write up

dbolduc (approved):

Issue #3947, opened 2025-12-02
Make google-cloud-auth safer by removing dependencies flagged as vulnerable by rustsec


alvarowolfx:
Based on https://github.com/googleapis/google-cloud-rust/issues/3934, we need to make the auth library clean of security warnings. This is currently being caused mainly by the id token feature (more specifically the id token verification feature), which pull the `jsonwebtoken` crate, which has `rsa` in it's dependency chain.  Vulnerability report: https://rustsec.org/advisories/RUSTSEC-2023-0071

We also include `rsa` as dev-dependency for tests, to generate a real certificate based on some test prime numbers.  I think we can replace this with `ring` https://docs.rs/ring/latest/ring/rsa/struct.KeyPair.html#method.from_components. We already use `ring` as a crypto provider in the `rustls` crate.

Current plan:

- [x] Gate `idtoken` behind a rust feature. https://github.com/googleapis/google-cloud-rust/pull/3961
- [x] ~Use `aws_lc` backend instead of `rust_crypto` for `jsonwebtoken`. This removes dependency on `rsa`, but customers don't like the `aws_lc` backend, since is not written in pure rust, so this is just temporary and gated behind the `idtoken` feature. https://github.com/googleapis/google-cloud-rust/pull/3962~ We are gonna wait for `rsa` to be fixed. For the time being, `idtoken` is gated behind a feature.
- [ ] Replace `rsa` usage with `ring` for tests.
- [ ] Move away from `jsonwebtoken`. We don't have a candidate crate yet, but maybe some code can be written to check for signatures using `ring`. https://github.com/googleapis/google-cloud-rust/pull/3960

alvarowolfx:
This can block release of the google-cloud-auth crate, which is going to happen next week. Tagging this as P1

alvarowolfx:
`idtoken` is now gated behing a rust feature and only users enabling the feature will see the warning. `rsa` next release will fix the security issue, so hopefully by the next release of our SDK, the warning will be gone. Moving to p2

Bitdragon50:
Here is a polished and clear rewrite of your message:

The rustls-pemfile crate has been flagged by RustSec as unmaintained. To address this, I have submitted a PR to update google-cloud-auth to use the rustls-pki-types crate instead. You can review the changes here: https://github.com/googleapis/google-cloud-rust/pull/4109

houseme:
[...]

PR #3962, opened 2025-12-02
impl(auth): change jsonwebtoken crypto provider

alvarowolfx:
Towards #3947


alvarowolfx:
Moving away from `jsonwebtoken` is preferable, so will leave as draft for now.

alvarowolfx:
we are gonna keep using `rust_crypto` and wait for `rsa` crate to be updated

PR #3966, opened 2025-12-02
test(storage): fixes in random access benchmark

coryan:
Report the range offset, in case it impacts the results (does not seem to). Also report the bytes transferred, which helped be find a bug on the transfer length.

### DIFF
--- a/src/storage/benchmarks/random/src/bidi.rs
+++ b/src/storage/benchmarks/random/src/bidi.rs
@@ -64,14 +64,16 @@ impl Runner {
             );
         };
         let mut reader = descriptor
-            .read_range(ReadRange::segment(range.read_offset, range.read_offset))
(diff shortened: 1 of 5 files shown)


src/storage/benchmarks/random/src/bidi.rs
```diff
@@ -64,14 +64,16 @@ impl Runner {
             );
         };
         let mut reader = descriptor
-            .read_range(ReadRange::segment(range.read_offset, range.read_offset))
+            .read_range(ReadRange::segment(range.read_offset, range.read_length))
```
dbolduc, line 67: lol

coryan: I LOL-ed too, or at least I was LOL-ing in my head

dbolduc: LIMH

Issue #3645, opened 2025-10-29
auth: Signing and Signed URL


alvarowolfx:
To support generating signed URL for GCS, we need to implement support for calling IAM signing endpoints via some supported Credential types in the Rust SDK.

Some ideas:
* Have a given `Credentials` builder provide a  `Signer` trait (via `build_signer()` method) that can later be used on a `bucket.signed_url(object_name, signer)` method in GCS.

https://docs.cloud.google.com/storage/docs/access-control/signed-urls

Sydney-o9:
Please make it clear in the documentation of `cloud_storage` that signed url is unstable

<img width="663" height="54" alt="Image" src="https://github.com/user-attachments/assets/e9ae3347-a99d-4e38-9241-abbf664fe72c" />

Most people are switching to this library expecting feature parity from Yoshida's crate, which is not the case. Thank you

alvarowolfx:
> Please make it clear in the documentation of `cloud_storage` that signed url is unstable
> Most people are switching to this library expecting feature parity from Yoshida's crate, which is not the case. Thank you

thanks for reaching out @Sydney-o9. We are still not actively recommending customers to use the unstable features of this SDK since we introduced such features to mark things that are work in progress and might be in a state where they are not even working. Is not the case for signed urls, and the good news is that is getting close to be marked stable (#4185) and is very likely that is gonna be out in the next release.

PR #3828, opened 2025-11-17
WIP: signed urls 

alvarowolfx:
Towards #3645

PR #3969, opened 2025-12-02
impl(auth): add SigningProvider trait

alvarowolfx:
Towards #3645

### DIFF
--- a/src/auth/src/signer.rs
+++ b/src/auth/src/signer.rs
@@ -1,4 +1,4 @@
-// Copyright 2024 Google LLC
+// Copyright 2025 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
@@ -58,18 +58,17 @@ impl Signer {
 }
 
 /// A trait for types that can sign content.
-#[async_trait::async_trait]
-pub trait SigningProvider: Send + Sync + std::fmt::Debug {
+pub trait SigningProvider: std::fmt::Debug {
     /// Returns the email address of the authorizer.
     ///
     /// It is typically the Google service account client email address from the Google Developers Console
-    /// in the form of "xyz@developer.gserviceaccount.com". Required.
-    async fn client_email(&self) -> Result<String>;
+    /// in the form of "xxx@developer.gserviceaccount.com". Required.
+    fn client_email(&self) -> impl Future<Output = Result<String>> + Send;
 
     /// Signs the content.
     ///
-    /// Returns the signature in hex format.
-    async fn sign(&self, content: &[u8]) -> Result<String>;
+    /// Returns the signature.
+    fn sign(&self, content: &[u8]) -> impl Future<Output = Result<String>> + Send;
 }
 
 pub(crate) mod dynamic {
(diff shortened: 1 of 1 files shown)


alvarowolfx:
See #3828  for big picture

src/auth/src/signer.rs
```diff
@@ -0,0 +1,172 @@
+// Copyright 2024 Google LLC
```
coryan: nit: 2025

src/auth/src/signer.rs
```diff
@@ -0,0 +1,172 @@
...
+    pub async fn sign<T>(&self, content: T) -> Result<String>
+    where
+        T: AsRef<[u8]> + Send + Sync,
+    {
+        self.inner.sign(content.as_ref()).await
+    }
+}
+
+/// A trait for types that can sign content.
+#[async_trait::async_trait]
```
coryan: If this is a public API, then we don't want to be using `async_trait::async_trait`.  That is why we have a `dynamic` namespace, to keep the 0.x trait out of our public API.

Maybe you can change my mind about this, but until then please remove.

src/auth/src/signer.rs
```diff
@@ -0,0 +1,172 @@
...
+pub trait SigningProvider: Send + Sync + std::fmt::Debug {
```
coryan: Typically we expose public traits like this so the type (in this case `Signer`) can be mocked, but I don't see a clear need, nor a way to initial a signer with a mock.  Thoughts?

src/auth/src/signer.rs
alvarowolfx: done

src/auth/src/signer.rs
alvarowolfx: this was an oversight, just fixed it and removed the `async_trait`

src/auth/src/signer.rs
alvarowolfx: Customers can provide their own way of signing blob (not sure yet how this is used, but the other SDKs provide that), so the use case is both for mocking like you mentioned, but also for letting customers implement their own logic to sign things.

coryan: > Customers can provide their own way of signing blob (not sure yet how this is used, but the other SDKs provide that), so the use case is both for mocking like you mentioned, but also for letting customers implement their own logic to sign things.

Fine, but I still cannot create a `Signer` from a signer provide, can I? That is, if I write:

```rust
impl SigningProvider for MyCoolSigner { ... }
```

how do I get that inside a `Signer` so I can use it to sign URLs or something?

src/auth/src/signer.rs
```diff
@@ -0,0 +1,223 @@
...
+/// from any type that implements [SigningProvider].
+#[derive(Clone, Debug)]
+pub struct Signer {
+    pub(crate) inner: Arc<dyn dynamic::SigningProvider>,
+}
+
+impl<T> std::convert::From<T> for Signer
+where
+    T: SigningProvider + Send + Sync + 'static,
+{
```
coryan, line 31: Ah, I see the `From` thing. Yup.

src/auth/src/signer.rs
```diff
@@ -0,0 +1,223 @@
...
+    /// This is typically the service account email.
+    pub async fn client_email(&self) -> Result<String> {
+        self.inner.client_email().await
+    }
+
+    /// Signs the provided content using the underlying provider.
+    ///
+    /// The content is typically a string-to-sign generated by the caller.
+    /// Returns the signature as a base64 encoded string (or other format depending on implementation,
+    /// but typically hex or base64).
```
coryan, line 51: nit: reflow.

src/auth/src/signer.rs
```diff
@@ -0,0 +1,223 @@
...
+        self.inner.sign(content.as_ref()).await
+    }
+}
+
+/// A trait for types that can sign content.
+pub trait SigningProvider: std::fmt::Debug {
+    /// Returns the email address of the authorizer.
+    ///
+    /// It is typically the Google service account client email address from the Google Developers Console
+    /// in the form of "xxx@developer.gserviceaccount.com". Required.
```
coryan, line 65: nit: reflow.

sai-sunder-s (approved):

alvarowolfx:
@coryan I'll fix the comments on coming PRs

Issue #3957, opened 2025-12-02
Implement subscriber


dbolduc:
We need to implement the internals for the subscriber. I may spin out these tasks into their own issues, with more details.

- [x] Generate internals owned by sidekick
- [x] Introduce `LeaseState` to track at-least-once acks/nacks/leases
- [x] Introduce `Leaser` trait to the `LeaseState`
- [x] Introduce lease loop task
- [x] Introduce internal `Stub`
- [x] Implement an internal `Transport` for the `Stub` (actually sends RPCs)
- [x] Implement `DefaultLeaser` using `Stub`
- [x] Add keepalive task for stream
- [x] Open subscribe stream
- [x] Add real impl for `Handler`
- [x] Add real impl for `Session`
  - [x] Stream messages and serve them
  - [x] Add keepalives
  - [x] Add lease management

PR #3971, opened 2025-12-03
impl(pubsub): nascent lease management

dbolduc:
Part of the work for #3957 

Add a crate private struct to help track the lease state.

A follow up PR will introduce a `trait Leaser`, and start taking actual actions (on `flush()`).

The plan is to keep at-least-once and exactly-once delivery mostly separate. The code in this PR only applies to at-least-once delivery.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -55,11 +55,10 @@ impl LeaseState {
         }
     }
 
-    /// Flush pending acks/nacks and extend leases
+    /// Flush pending acks/nacks
     async fn flush(&mut self) {
         let _to_ack = std::mem::take(&mut self.to_ack);
         let _to_nack = std::mem::take(&mut self.to_nack);
-        let _under_lease: Vec<String> = self.under_lease.iter().cloned().collect();
     }
 }
 
--- a/src/pubsub/src/subscriber.rs
+++ b/src/pubsub/src/subscriber.rs
@@ -12,4 +12,4 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub(crate) mod leaser;
+pub(crate) mod lease_state;


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -0,0 +1,158 @@
...
+struct LeaseState {
+    // TODO(#3957) - support message expiry
+    under_lease: HashSet<String>,
+    to_ack: Vec<String>,
+    to_nack: Vec<String>,
+    // TODO(#3964) - support exactly once acks
+}
+
+impl LeaseState {
+    fn new() -> Self {
```
dbolduc, line 27: FYI: this will grow to something where we cannot `#[derive(Default)]`

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -0,0 +1,157 @@
...
+// limitations under the License.
+
+use std::collections::HashSet;
+
+#[derive(Debug, PartialEq)]
+struct LeaseState {
+    // TODO(#3957) - support message expiry
+    under_lease: HashSet<String>,
+    to_ack: Vec<String>,
+    to_nack: Vec<String>,
```
coryan, line 22: Food for tought: delaying nacks (even for batching) stops the message from being processed elsewhere.

dbolduc: True. I figured it would be fine to flush on something like a 1s delay when extensions already take minimum 10s.

With that being said, I don't have a good feel for when an application might nack a message. I could see them doing it if they were under load. :shrug:

coryan: I think the typical use-case is something like "I receive messages and I save them to a database / GCS object / something".   If the saving part fails you want to nack the message so some other process gets a second try.

dbolduc: Ah thanks.

PR #3978, opened 2025-12-03
chore: update rust version for ref upload

dbolduc:
Tested locally with:

```sh
cd google-cloud-rust
# make PR edits...
gcloud --project rust-sdk-testing builds submit --config .gcb/referenceupload.yaml .
```

https://pantheon.corp.google.com/cloud-build/builds;region=global/465dd694-0a74-4782-b638-021966b673b0;step=1?e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=monitoring_api_staging&project=rust-sdk-testing

### DIFF
--- a/.gcb/referenceupload.yaml
+++ b/.gcb/referenceupload.yaml
@@ -31,7 +31,7 @@ steps:
     env:
       - _SCCACHE_VERSION=${_SCCACHE_VERSION}
       - _SCCACHE_SHA256=${_SCCACHE_SHA256}
-  - name: 'rust:1.85-bookworm'
+  - name: 'rust:${_RUST_VERSION}-bookworm'
     script: |
       #!/usr/bin/env bash
       set -e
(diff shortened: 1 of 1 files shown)


.gcb/referenceupload.yaml
```diff
@@ -46,5 +46,6 @@ steps:
       cargo install --locked cargo-workspaces
```
coryan, line 46: This will break in a few months, when the latest version of `cargo-workspaces` does not work with rustc 1.91.  Maybe pin `cargo-workspaces` to an specific version so we know it works with the compiler version?

dbolduc: I assumed we would update this when we periodically update to the latest rust version. WDYT of that plan?

coryan: I do not like that plan.

This build breaks when there is a new release of `cargo-workpaces` (with new requirements), that happens outside our control, whether we update the rustc version or not.

dbolduc: Understood. I assume we were using `0.4.1`, previously. I will try pinning to that. If this build gets past the `cargo install`, I will update this PR.

https://pantheon.corp.google.com/cloud-build/builds;region=global/26f664b3-075c-4257-820d-2463b1a0ca04;step=1?e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=monitoring_api_staging&project=rust-sdk-testing

PR #3981, opened 2025-12-03
impl(pubsub): a task to keepalive streams

dbolduc:
Part of the work for #3957 

The new way to do keepalives in Pub/Sub is to ping the stream with empty writes. This PR adds a task to do that.

It is not responsible for opening the stream with the initial write. It just does the pinging.

### DIFF
--- a/src/pubsub/src/subscriber/keepalive.rs
+++ b/src/pubsub/src/subscriber/keepalive.rs
@@ -69,7 +69,7 @@ mod tests {
             r.subscription,
             "projects/my-project/subscriptions/my-subscription"
         );
-        assert_eq!(Instant::now() - start, Duration::from_secs(30));
+        assert_eq!(Instant::now() - start, KEEPALIVE_PERIOD);
 
         // Wait for the second keepalive
         let r = request_rx.recv().await.unwrap();
@@ -77,7 +77,7 @@ mod tests {
             r.subscription,
             "projects/my-project/subscriptions/my-subscription"
         );
-        assert_eq!(Instant::now() - start, Duration::from_secs(60));
+        assert_eq!(Instant::now() - start, KEEPALIVE_PERIOD * 2);
 
         // Wait for the third keepalive
         let r = request_rx.recv().await.unwrap();
@@ -85,7 +85,7 @@ mod tests {
             r.subscription,
             "projects/my-project/subscriptions/my-subscription"
         );
-        assert_eq!(Instant::now() - start, Duration::from_secs(90));
+        assert_eq!(Instant::now() - start, KEEPALIVE_PERIOD * 3);
     }
 
     #[tokio::test(start_paused = true)]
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/keepalive.rs
```diff
@@ -0,0 +1,115 @@
...
+/// `CancellationToken` and `await`ing the returned handle.
+///
+/// Callers can also just drop the returned handle to shutdown.
+fn spawn(
+    request: StreamingPullRequest,
+    request_tx: Sender<StreamingPullRequest>,
+    shutdown: CancellationToken,
+) -> JoinHandle<()> {
+    tokio::spawn(async move {
+        let mut keepalive = interval_at(Instant::now() + KEEPALIVE_PERIOD, KEEPALIVE_PERIOD);
```
coryan, line 38: TIL

src/pubsub/src/subscriber/keepalive.rs
```diff
@@ -0,0 +1,115 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::google::pubsub::v1::StreamingPullRequest;
+use tokio::sync::mpsc::Sender;
+use tokio::task::JoinHandle;
+use tokio::time::{Duration, Instant, interval_at};
+use tokio_util::sync::CancellationToken;
```
coryan, line 19: I hoped there would be a way to do this without an extra dep  (`tokio::sync::Notify` ?), but seems like a pretty standard thing. SGTM.

coryan (approved):

src/pubsub/src/subscriber/keepalive.rs
dbolduc: I was originally using `Notify`, but was having issues either: (1) missing a notify in between clones. (2) managing lifetimes of an `OwnedNotified`.

`CancellationToken`, on the other hand, made the code very natural to write.

> seems like a pretty standard thing

I learned it from the `gcloud-pubsub` subscriber quickstart. I think it is a .NET concept too.

src/pubsub/src/subscriber/keepalive.rs
```diff
@@ -0,0 +1,115 @@
...
+) -> JoinHandle<()> {
+    tokio::spawn(async move {
+        let mut keepalive = interval_at(Instant::now() + KEEPALIVE_PERIOD, KEEPALIVE_PERIOD);
+        loop {
+            tokio::select! {
+                    _ = shutdown.cancelled() => break,
+                    _ = keepalive.tick() => {
+                    let _ = request_tx.send(request.clone()).await;
+                }
+            }
```
PhongChuong, line 45: The indentation here looks weird to me, shouldn't it be:
```
            tokio::select! {
                _ = shutdown.cancelled() => break,
                _ = keepalive.tick() => {
                    let _ = request_tx.send(request.clone()).await;
                }
            }
```
instead?

src/pubsub/src/subscriber/keepalive.rs
```diff
@@ -0,0 +1,115 @@
...
+        let shutdown = CancellationToken::new();
+        let _handle = spawn(request, request_tx, shutdown);
+
+        // Wait for the first keepalive
+        let r = request_rx.recv().await.unwrap();
+        assert_eq!(
+            r.subscription,
+            "projects/my-project/subscriptions/my-subscription"
+        );
+        assert_eq!(Instant::now() - start, KEEPALIVE_PERIOD);
```
PhongChuong, line 72: I think you can use [elapsed()](https://docs.rs/tokio/latest/tokio/time/struct.Instant.html#method.elapsed) instead of `now() - start`.

src/pubsub/src/subscriber/keepalive.rs
dbolduc: Ah, yeah, thanks. `rustfmt` does not work within macros, and my manual indentation was not doing any favors for readability. Fixed in #3986

src/pubsub/src/subscriber/keepalive.rs
dbolduc: Ah, TIL, thanks. Fixed in #3986.

PR #3984, opened 2025-12-04
impl(auth): add IAM API based signing provider

alvarowolfx:
This uses the IAM signBlob API for signatures. Will be used by MDS and Impersonated sourced credentials types.

Towards #3645

### DIFF
--- a/src/auth/src/signer/iam.rs
+++ b/src/auth/src/signer/iam.rs
@@ -109,7 +109,7 @@ async fn sign_blob_call_with_retry(
     let retry_throttler: RetryThrottlerArg = AdaptiveThrottler::default().into();
     let retry_throttler: SharedRetryThrottler = retry_throttler.into();
 
-    let response = retry_loop(
+    retry_loop(
         async move |_| {
             let source_headers = credentials
                 .headers(Extensions::new())
@@ -125,9 +125,7 @@ async fn sign_blob_call_with_retry(
         backoff_policy,
     )
     .await
-    .map_err(SigningError::transport)?;
-
-    Ok(response)
+    .map_err(SigningError::transport)
 }
 
 async fn sign_blob_call(


src/auth/src/signer/iam.rs
```diff
@@ -0,0 +1,213 @@
...
+            endpoint: "https://iamcredentials.googleapis.com".to_string(),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl SigningProvider for IamSigner {
+    async fn client_email(&self) -> Result<String> {
+        Ok(self.client_email.clone())
+    }
```
coryan, line 62: No tests for this function, but probably easy to fix?

src/auth/src/signer/iam.rs
```diff
@@ -0,0 +1,213 @@
...
+            }
+        };
+
+        let client_email = self.client_email.clone();
+        let url = format!(
+            "{}/v1/projects/-/serviceAccounts/{}:signBlob",
+            self.endpoint, client_email
+        );
+
+        let client = Client::new();
```
coryan: Creating a client on the fly seems like a waste.  All the caching will be  lost. Can we save the client?

src/auth/src/signer/iam.rs
```diff
@@ -0,0 +1,213 @@
...
+        let response = client
+            .post(url)
+            .header("Content-Type", "application/json")
+            .headers(source_headers)
+            .json(&body)
+            .send()
+            .await
+            .map_err(SigningError::transport)?;
+
+        if !response.status().is_success() {
```
coryan, line 78: Should we have a retry loop here?  And at that point: should we try to use the IAM client?

I can see the cyclic dependency, could we fix that with a trait? And maybe a bit of hand-crafted code to implement the trait in the client?

Cargo.toml
```diff
@@ -307,6 +307,7 @@ clap                  = { default-features = false, version = "4" }
 crc32c                = { default-features = false, version = "0.6.8" }
 futures               = { default-features = false, version = "0.3" }
 http                  = { default-features = false, version = "1", features = ["std"] }
+hex                   = { default-features = false, version = "0.4.3" }
```
coryan: Skip the patch version to reduce churn:

```suggestion
hex                   = { default-features = false, version = "0.4" }
```

src/auth/src/signer/iam.rs
```diff
@@ -0,0 +1,213 @@
...
+        let res = response
+            .json::<SignBlobResponse>()
+            .await
+            .map_err(SigningError::transport)?;
+
+        let signature = BASE64_STANDARD
+            .decode(res.signed_blob)
+            .map_err(SigningError::transport)?;
+
+        let signature = hex::encode(signature);
```
coryan, line 92: Why not return `bytes::Bytes` and let the caller figure out if they want to turn that into `hex` or some other format?

src/auth/src/signer/iam.rs
alvarowolfx: the whole auth crate would benefit a LOT of using the IAM Client. But I tried that some time ago and ran into the cyclic dependency issue that I could not figure out how to get away with that. Will explore some of ideas that you gave.

coryan: In more detail:

define a trait in the `g-c-auth` library, write the implementation in terms of that trait.  Then implement the trait in the client, maybe using a bit of hand-crafted code as in:

https://github.com/googleapis/google-cloud-rust/blob/e885645fca9472e7ec8706b452d98c930f50d38b/src/generated/cloud/compute/v1/src/operation.rs#L15-L24

trick the generator into importing this hand-crafted code:

https://github.com/googleapis/google-cloud-rust/blob/e885645fca9472e7ec8706b452d98c930f50d38b/src/generated/cloud/compute/v1/.sidekick.toml#L48

You may need to trick the generator into taking a direct dependency on `google-cloud-auth` too.  Maybe via a `force-used` package?

https://github.com/googleapis/google-cloud-rust/blob/e885645fca9472e7ec8706b452d98c930f50d38b/.sidekick.toml#L57

src/auth/src/signer/iam.rs
alvarowolfx: done

src/auth/src/signer/iam.rs
alvarowolfx: good point, but I have to change the SigningProvider trait. I'll check that in #3828 and send a separated PR to change the trait.

src/auth/src/signer/iam.rs
alvarowolfx: will try this separately. For now I added retry support using gax_internal::retry::retry_loop

src/auth/src/signer/iam.rs
```diff
@@ -0,0 +1,304 @@
...
+) -> Result<reqwest::Response> {
+    let sleep = async |d| tokio::time::sleep(d).await;
+
+    let retry_policy: RetryPolicyArg = Aip194Strict.with_attempt_limit(3).into();
+    let backoff_policy: BackoffPolicyArg = ExponentialBackoff::default().into();
+    let backoff_policy: Arc<dyn BackoffPolicy> = backoff_policy.into();
+    let retry_throttler: RetryThrottlerArg = AdaptiveThrottler::default().into();
+    let retry_throttler: SharedRetryThrottler = retry_throttler.into();
+
+    let response = retry_loop(
```
coryan: nit: why do you need a named variable here?  Consider:


```suggestion
    retry_loop(
```

src/auth/src/signer/iam.rs
```diff
@@ -0,0 +1,304 @@
...
+        sleep,
+        true, // signBlob is idempotent
+        retry_throttler,
+        retry_policy.into(),
+        backoff_policy,
+    )
+    .await
+    .map_err(SigningError::transport)?;
+
+    Ok(response)
```
coryan: And also:

```suggestion
    .map_err(SigningError::transport)
```

src/auth/src/signer/iam.rs
alvarowolfx: good point. I resolved some conflict issues after merging another PR related to signing and applied this idea. Thanks

PR #3992, opened 2025-12-05
fix: whitespace alignment issues in Cargo.toml files

julieqiu:
Ran `taplo format src/generated/**/Cargo.toml`.

### DIFF
--- a/src/generated/api/apikeys/v2/Cargo.toml
+++ b/src/generated/api/apikeys/v2/Cargo.toml
@@ -28,18 +28,18 @@ rust-version.workspace = true
 
 [dependencies]
 async-trait.workspace = true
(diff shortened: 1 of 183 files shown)


coryan:
We have not been running `taplo fmt` for generated `Cargo.toml` files.  If we want to start doing this, we should change `sidekick` to automate the call *or* update the [howto](https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-generated-code-maintenance.md#update-the-code-with-new-googleapis-protos) to say that we should run it manually.

Separately, you need to rebase and resolve the conflicts.

julieqiu:
I'll close this for now and let's discuss what we want to do on the librarian side.

PR #3994, opened 2025-12-05
refactor(discovery): simplify error handling

kasugamirai:
…x from handle_polling_error function call


coryan:
Thanks for the cleanup!

PR #3997, opened 2025-12-05
impl(auth): add service account key based signing provider

alvarowolfx:
Signs blobs with service account private key. 

Towards #3645

### DIFF
--- a/src/auth/src/credentials/service_account.rs
+++ b/src/auth/src/credentials/service_account.rs
@@ -939,6 +939,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[cfg(google_cloud_unstable_signed_url)]
     async fn get_service_account_signer() -> TestResult {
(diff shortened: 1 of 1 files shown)


src/auth/src/signer.rs
```diff
@@ -47,8 +49,7 @@ impl Signer {
     /// Signs the provided content using the underlying provider.
     ///
     /// The content is typically a string-to-sign generated by the caller.
-    /// Returns the signature as a base64 encoded string (or other format depending on implementation,
-    /// but typically hex or base64).
+    /// Returns the signature as a hex encoded string.
```
coryan, line 52: As before, returning `hex` seems odd.

coryan (approved):

src/auth/src/signer.rs
alvarowolfx: agree, I'll change the trait and the merged implementations

PR #4000, opened 2025-12-05
chore: update googleapis SHA circa 2025-12-05

coryan:
Motivated by #3999


PhongChuong (approved):
I'm a bit confused as #3999 is to onboard cloud/maintenance but this only regenerates firestore.

coryan:
> I'm a bit confused as #3999 is to onboard cloud/maintenance but this only regenerates firestore.

We need to update the googleapis SHA first.  cloud/maintenance was not available in the previous SHA.

PR #4006, opened 2025-12-05
cleanup: prefer `Instance::elapsed()`

coryan:
Prefer `start.elapsed()` over `Instant::now() - start`. The former makes the intent clearer and nicely saturates to 0.


coryan:
@PhongChuong thanks for pointing out this function.

PR #4007, opened 2025-12-05
impl(auth): change SigningProvider trait to return Bytes

alvarowolfx:
Makes more sense to return raw bytes of the signed blob instead of a more opinionated hex format.

Towards #3645

### DIFF
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -47,7 +47,6 @@ thiserror.workspace   = true
 time                  = { workspace = true, features = ["serde"] }
 tokio                 = { workspace = true, features = ["fs", "process"] }
 bon.workspace         = true
-hex                   = { workspace = true, features = ["std"] }
 jsonwebtoken          = { workspace = true, features = ["rust_crypto"], optional = true }
 # Local dependencies
 gax.workspace = true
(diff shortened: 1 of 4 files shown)


alvarowolfx:
This also fixes a test in the IAM signBlob test suite, where the retry test was not being executed and was not working as intended.

src/auth/Cargo.toml
```diff
@@ -36,6 +36,7 @@ rustc_version.workspace = true
 [dependencies]
 async-trait.workspace = true
 base64.workspace      = true
+bytes.workspace       = true
```
coryan, line 39: You also added a number of `hex.workspace` dependencies and (as I recall) also introduced `hex` to the top-level Cargo? Can we remove these?

src/auth/src/signer/iam.rs
```diff
@@ -141,14 +139,30 @@ async fn sign_blob_call(
...
-    client
+    let response = client
         .post(url)
         .header("Content-Type", "application/json")
         .headers(source_headers.clone())
         .json(&body)
         .send()
         .await
-        .map_err(|e| gax::error::Error::transport(source_headers, e))
+        .map_err(gax::error::Error::io)?;
```
coryan: Seems like this should go to a separate PR, and probably deserves a test (using `httptest` should not be that hard).

coryan (changes requested):

src/auth/Cargo.toml
alvarowolfx: just removed `hex` as it's not used yet. Only `storage` will need later.

Issue #4012, opened 2025-12-05
Pubsub: Support message ordering in Publisher


PhongChuong:
Support [message ordering](https://docs.cloud.google.com/pubsub/docs/ordering) in Publisher.

The publisher must:
1. Within key ordering
2. Batching
4. Correctly handle [retry](https://docs.cloud.google.com/pubsub/docs/retry-requests#retry_ordering)

suzmue:
This is completed.

The open PR is an optimization.

PR #4016, opened 2025-12-05
feat(pubsub): Initial support message ordering in Publisher

PhongChuong:
Add initial support for message ordering in Publisher. This feature does not retry and does not respect message ordering in the case of failure. These will be supported in a later PR.

Towards: #4012

### DIFF
--- a/src/pubsub/src/publisher/worker.rs
+++ b/src/pubsub/src/publisher/worker.rs
@@ -84,7 +84,7 @@ impl Worker {
         // A dictionary of ordering key to outstanding publish operations.
         // We batch publish operations on the same ordering key together.
         // Publish without ordering keys are treated as having the key "".
-        // TODO(#4012): Remove pending_batches entries when there are no oustanding publish.
+        // TODO(#4012): Remove pending_batches entries when there are no outstanding publish.
         let mut pending_batches: HashMap<String, OutstandingPublishes> = HashMap::new();
         let delay = self.batching_options.delay_threshold;
         let message_limit = self.batching_options.message_count_threshold;


src/pubsub/src/publisher/worker.rs
```diff
@@ -92,40 +98,48 @@ impl Worker {
...
                 _ = inflight.next(), if !inflight.is_empty() => {},
                 // Handle timer events.
-                // This branch will only be checked when there is a non-empty batch,
-                // so this will not fire continuously.
-                _ = &mut timer, if !batch.is_empty() => {
-                    batch.flush(self.client.clone(), self.topic_name.clone(), &mut inflight);
+                _ = &mut timer => {
+                    for (_, outstanding) in pending_batches.iter_mut() {
+                        if !outstanding.pending_batch.is_empty() {
+                            outstanding.pending_batch.flush(self.client.clone(), self.topic_name.clone(), &mut inflight);
```
suzmue, line 105: Need to double check with rest of PR, but I don't think this takes into account that there may be a batch already out for that ordering key?

I think this will be easier to do with the refactoring that we talked about where more of the control for flushing batches is moved to the OutstandingPublishes, so I am okay with doing in a follow up, but please add a TODO for it.

src/pubsub/src/publisher/worker.rs
```diff
@@ -78,7 +81,10 @@ impl Worker {
     /// The loop terminates when the `rx` channel is closed, which happens when all
     /// `Publisher` clones have been dropped.
     pub(crate) async fn run(mut self) {
-        let mut batch = Batch::new(&self.topic_name);
+        // A dictionary of ordering key to outstanding publish operations.
+        // We batch publish operations on the same ordering key together.
+        // Publish without ordering keys are treated as having the key "".
+        let mut pending_batches: HashMap<String, OutstandingPublishes> = HashMap::new();
```
suzmue, line 88: One thing to think about, do we ever remove ordering keys from this map? Or will it grow indefinitely?

src/pubsub/src/publisher/worker.rs
```diff
@@ -92,40 +98,48 @@ impl Worker {
...
-                            // Wait on all the tasks that exist right now.
-                            // We could instead tokio::spawn this as well so the publisher
-                            // can keep working on additional messages. The worker would
-                            // also need to keep track of any pending flushes, and make sure
-                            // all of those resolve as well.
-                            let mut flushing = std::mem::take(&mut inflight);
-                            while flushing.next().await.is_some() {}
+                            for (_, outstanding) in pending_batches.iter_mut() {
+                                if !outstanding.pending_batch.is_empty() {
+                                    outstanding.pending_batch.flush(self.client.clone(), self.topic_name.clone(), &mut inflight);
```
suzmue, line 126: I would at least add a comment / TODO here for the behavior of flush when there are ordering keys. This will involve sending multiple batches *sequentially* since they need to be ordered. So it is not as simple as just flushing the existing messages and waiting on the response.

src/pubsub/src/publisher/worker.rs
```diff
@@ -135,3 +149,18 @@ impl Worker {
         }
     }
 }
+
+#[derive(Debug, Default)]
+pub(crate) struct OutstandingPublishes {
```
suzmue, line 157: I'm not sure I see the necessity for adding a new type that just wraps the Batch. Can we just use Batch directly?

src/pubsub/src/publisher/worker.rs
PhongChuong: Yep. This does not take into account if there is a batch inflight.
The TODO on worker.rs:156 is to handle this.

src/pubsub/src/publisher/worker.rs
PhongChuong: It will grow indefinitely, I've added a TODO.

src/pubsub/src/publisher/worker.rs
PhongChuong: We can. I decided to move the refactoring into its own PR to keep things simple here. I can also move OutstandingPublishes into the refactor PR if you like.

src/pubsub/src/publisher/worker.rs
PhongChuong: Added a TODO here with more explicit callout.

PhongChuong:
Thanks for the review. PTAL.

src/pubsub/src/publisher/worker.rs
suzmue: If you have a use for it in the next PRs thats fine.

suzmue (approved):
With the TODOs, I'm fine to submit this to keep making forward progress.

Issue #3952, opened 2025-12-02
Capture benchmark results for bidi streaming reads


coryan:
Capture both throughput (TTLB for largish reads 2MiB and 100MiB) and random access latency (TTFB for smallish reads 8KiB) benchmark results, using both regional and zonal buckets.

coryan:
Done in go/cloud-rust:capture-for-3952

PR #4026, opened 2025-12-08
test(storage-random): pick range size from list

coryan:
Sometimes it is easier to explain the results if one uses a small number of fixed range size values.  A few numbers such as "latency for 8KiB, 1MiB and 16MiB reads" are easier to explain than "the curve grows approximately linearly with a slope of X".

Part of the work for #3952

### DIFF
--- a/src/storage/benchmarks/random/src/random_size.rs
+++ b/src/storage/benchmarks/random/src/random_size.rs
@@ -140,4 +140,24 @@ mod tests {
         assert!(got.is_err(), "{got:?}");
         Ok(())
     }
+
+    #[test]
+    fn sample_range() {
+        let sampler = RandomSize::Range(20, 30);
+        for _ in 0..100 {
+            let got = sampler.sample(&mut rand::rng());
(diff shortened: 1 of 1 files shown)


src/storage/benchmarks/random/src/random_size.rs
```diff
@@ -0,0 +1,163 @@
...
+fn parse_size_arg(arg: &str) -> anyhow::Result<u64> {
+    let value = parse_size::parse_size(arg)?;
+    Ok(value)
+}
+
+pub fn parse_random_size_arg(arg: &str) -> anyhow::Result<RandomSize> {
+    if let Some(sep) = arg.find('-') {
+        let min = parse_size_arg(&arg[..sep])?;
+        let max = parse_size_arg(&arg[sep + 1..])?;
+        if min > max {
```
PhongChuong, line 76: nit. Not sure if min == max should also be a valid value. I guess in that case, Values can be used instead.

coryan: Ack. Leaving as-is for now.

PR #4030, opened 2025-12-08
impl(auth): add MDS based signing provider

alvarowolfx:
Towards #3645

### DIFF
--- a/src/auth/src/signer/mds.rs
+++ b/src/auth/src/signer/mds.rs
@@ -51,9 +51,9 @@ impl SigningProvider for MDSSigner {
     async fn client_email(&self) -> Result<String> {
         if self.client_email.get().is_none() {
             let email = self.fetch_client_email().await?;
-            self.client_email
-                .set(email.clone())
-                .map_err(|_e| SigningError::transport("failed to cache client email"))?;
+            // Ignore error if we can't set the client email.
+            // Might be due to multiple tasks trying to set value
+            let _ = self.client_email.set(email.clone());
             return Ok(email);
         }
 


src/auth/src/signer/mds.rs
```diff
@@ -0,0 +1,188 @@
...
+        self
+    }
+}
+
+#[async_trait::async_trait]
+impl SigningProvider for MDSSigner {
+    async fn client_email(&self) -> Result<String> {
+        let mut client_email = self
+            .client_email
+            .try_write()
```
coryan: Why use a RWLock if you always acquire a write lock?  That is actually slower than a normal lock.

src/auth/src/signer/mds.rs
```diff
@@ -0,0 +1,188 @@
...
+            .map_err(|_e| SigningError::transport("failed to obtain lock to read client email"))?;
+
+        if client_email.is_empty() {
+            let email = self.fetch_client_email().await?;
+            *client_email = email.clone();
+        }
```
coryan: I think you may be able to use `SetOnce` instead of managing your own lock:

https://docs.rs/tokio/latest/tokio/sync/struct.SetOnce.html

```suggestion
        if self.client_email.get().is_none() {
            let email = self.fetch_client_email().await?;
            let _ = self.client_email.set(email.clone());
            return Ok(email)
        }
        
        Ok(self.client_email.get().clone().unwrap())
```

src/auth/src/signer/mds.rs
```diff
@@ -50,17 +49,15 @@ impl MDSSigner {
...
-            .try_write()
-            .map_err(|_e| SigningError::transport("failed to obtain lock to read client email"))?;
-
-        if client_email.is_empty() {
+        if self.client_email.get().is_none() {
             let email = self.fetch_client_email().await?;
-            *client_email = email.clone();
+            self.client_email
+                .set(email.clone())
+                .map_err(|_e| SigningError::transport("failed to cache client email"))?;
```
coryan: I think this is a race condition if multiple tasks call this function at the same time. No? So maybe ignore the error instead?

coryan (changes requested):

src/auth/src/signer/mds.rs
alvarowolfx: good point, I was trying to solve a `clippy` issue and missed the point on why this could throw an error. Added a comment to our future selves looking at this.

coryan (approved):

PR #4035, opened 2025-12-09
impl(o11y): add resource_name support for gRPC tracing

westarle:


src/gax-internal/src/grpc.rs
```diff
@@ -490,4 +495,71 @@ mod tests {
         // We can't easily assert the internal state without exposing more internals,
         // but this verifies the method exists and runs.
     }
+
+    #[derive(Debug)]
+    struct NoAuth;
+
+    impl auth::credentials::CredentialsProvider for NoAuth {
```
coryan: We have the anonymous credentials:

https://docs.rs/google-cloud-auth/1.2.0/google_cloud_auth/credentials/anonymous/struct.Builder.html

Why not use those?

src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -25,6 +25,10 @@ use tower::{Layer, Service};
 #[derive(Clone, Copy, Debug)]
 pub struct AttemptCount(pub i64);
 
+/// A wrapper for the resource name to be stored in request extensions.
+#[derive(Clone, Debug)]
+pub struct ResourceName(pub String);
```
coryan: This is a nit:


```suggestion
pub struct ResourceName(String);

impl ResourceName {
  pub fn new(value: String) -> Self { Self(value) }
  pub fn as_str(&self) -> &str { self.0.as_str() }
}
```

I prefer wrapper types to hide things...

coryan (approved):
Small improvements (maybe, I am not sure, look at them and tell me what you think). None of them blocking.

src/gax-internal/src/grpc.rs
westarle: done!

src/gax-internal/src/observability/grpc_tracing.rs
westarle: done

coryan (approved):

PR #4037, opened 2025-12-09
chore: bump version numbers circa 2025-12-09

dbolduc:
`gax-internal` gets a patch bump because there is a fix, but no new APIs have been added (outside of unstable feature gates).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -389,8 +389,8 @@ tokio-test        = { default-features = false, version = "0.4" }
 # Local packages used as dependencies
 auth                          = { version = "1.3", path = "src/auth", package = "google-cloud-auth" }
 google-cloud-auth             = { version = "1.3", path = "src/auth" }
-gax                           = { version = "1.3", path = "src/gax", package = "google-cloud-gax" }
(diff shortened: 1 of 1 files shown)


src/gax/Cargo.toml
```diff
@@ -18,7 +18,7 @@ name = "google-cloud-gax"
 #     version of all downstream dependencies. For details see:
 #         https://github.com/googleapis/google-cloud-rust/issues/3237
 #         https://github.com/googleapis/google-cloud-rust/issues/3265
-version     = "1.3.1"
+version     = "1.4.0"
```
coryan, line 21: Should we update the top-level Cargo.toml file to depend on the latest version? It seems that every time we try to avoid that we run into trouble later.

It can happen in a follow-up PR.

coryan (approved):

src/gax/Cargo.toml
dbolduc: `gcb-pr-minimal-versions` says we must do it now.

Nice job and thank you, @suzmue!

PR #4043, opened 2025-12-10
impl(auth): add impersonated based signing provider

alvarowolfx:
Towards #3645

### DIFF
--- a/src/auth/src/credentials/impersonated.rs
+++ b/src/auth/src/credentials/impersonated.rs
@@ -123,7 +123,7 @@ pub(crate) const IMPERSONATED_CREDENTIAL_TYPE: &str = "imp";
 pub(crate) const DEFAULT_LIFETIME: Duration = Duration::from_secs(3600);
 pub(crate) const MSG: &str = "failed to fetch token";
 
-#[derive(Clone)]
+#[derive(Debug, Clone)]
 pub(crate) enum BuilderSource {
     FromJson(Value),
     FromCredentials(Credentials),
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/impersonated.rs
```diff
@@ -476,6 +477,49 @@ impl Builder {
...
+                    return Ok(signer);
+                }
+                let components = build_components_from_json(json)?;
+                let client_email =
+                    extract_client_email(&components.service_account_impersonation_url)?;
+                let creds = self.build()?;
+                let signer = crate::signer::iam::IamSigner::new(client_email, creds, iam_endpoint);
+                Ok(crate::signer::Signer {
+                    inner: Arc::new(signer),
+                })
```
coryan, line 505: The converage build says there are no tests for this, can you add some tests?

src/auth/src/credentials/impersonated.rs
```diff
@@ -123,6 +123,7 @@ pub(crate) const IMPERSONATED_CREDENTIAL_TYPE: &str = "imp";
 pub(crate) const DEFAULT_LIFETIME: Duration = Duration::from_secs(3600);
 pub(crate) const MSG: &str = "failed to fetch token";
 
+#[derive(Clone)]
```
coryan: `#[derive(Clone, Debug)]` if that is possible.

src/auth/src/credentials/impersonated.rs
```diff
@@ -550,6 +594,51 @@ pub(crate) fn build_components_from_json(
...
+        .collect();
+    if parts.len() != 2 {
+        return Err(BuilderError::parsing(
+            "invalid service account impersonation URL",
+        ));
+    }
+    let client_email = parts[1]
+        .trim_end_matches(":generateAccessToken")
+        .to_string();
+    Ok(client_email)
```
coryan: It bothers me that we need to materialize the vector...  I think you could do this:


```suggestion
    let parts = service_account_impersonation_url.split("serviceAccounts/");
    match (parts.nth(1), parts.next()) {
        (Some(email), None) => Ok(email.trim_end_matches(":generateAccessToken").to_string()),
        (_, _) => Err(BuilderError::parsing(
            "invalid service account impersonation URL",
        )),
    }
```

I realize that is is a silly optimization, and I am not sure the code is any more readable.

src/auth/src/credentials/impersonated.rs
```diff
@@ -550,6 +594,51 @@ pub(crate) fn build_components_from_json(
...
+        };
+        return Ok(Some(signer));
+    }
+    Ok(None)
+}
+
+#[cfg(google_cloud_unstable_signed_url)]
+fn extract_client_email(service_account_impersonation_url: &str) -> BuildResult<String> {
+    let parts: Vec<&str> = service_account_impersonation_url
+        .split("serviceAccounts/")
```
coryan: Shouldn't this be `/serviceAccounts/` with a leading slash?

src/auth/src/credentials/impersonated.rs
alvarowolfx: added tests for that and also for when the client email can't be extracted

src/auth/src/credentials/impersonated.rs
alvarowolfx: good catch

src/auth/src/credentials/impersonated.rs
alvarowolfx: at first I was confused about the new code, but now it makes sense and I think it's better, specially as `nth(1)` returns an option and seems safer to access data in a index position.

alvarowolfx (commented):
addressed review comments and increase test coverage

Issue #4044, opened 2025-12-10
Make number of gRPC subchannels configurable


coryan:
When using gRPC we want multiple sockets to get good throughput and latency. Otherwise the queuing on the same HTTP/2 request makes everything too slow.

To close this issue we need to:

- [x] Add a configuration parameter to `GrpcClient` that controls the number of subchannels.
- [ ] Set the default value for the storage client to something like "number of CPUs".

PR #4045, opened 2025-12-10
impl(gax): gRPC clients can have multiple subchannels

coryan:
Part of the work for #4044


coryan:
PTAL, I decided to remove some public APIs.

PR #4051, opened 2025-12-12
refactor(pubsub): move worker logic into OutstandingPublishes

PhongChuong:
Refactor by: 
- moving batch size logic from Worker to OutstandingPublishes
- simplify Batch by passing initial message size instead of the topic string
- simplify Batch by returning a JoinHandle instead of appending to inflight FuturesUnordered data structure

Towards: #4012

### DIFF
--- a/src/pubsub/src/publisher/worker.rs
+++ b/src/pubsub/src/publisher/worker.rs
@@ -74,6 +74,8 @@ impl Worker {
     /// 2. A timer is continuously fired to flush all pending batches.
     /// 3. A `Flush` command from the `Publisher` causes pending batches to
     ///    be sent immediately.
+    /// 4. The `inflight` set is continuously polled to remove `JoinHandle`s for
+    ///    send tasks that have completed, preventing the set from growing indefinitely.
     ///
     /// The loop terminates when the `rx` channel is closed, which happens when all
     /// `Publisher` clones have been dropped.


src/pubsub/src/publisher/worker.rs
```diff
@@ -112,27 +102,21 @@ impl Worker {
...
-                            }
+                            let outstanding_publishes =
+                                pending_batches
+                                    .entry(ordering_key)
+                                    .or_insert(OutstandingPublishes::new(
+                                        &self.topic_name,
+                                        message_limit,
+                                        byte_threshold,
+                                    ));
+                            outstanding_publishes.push(msg, self.client.clone(), self.topic_name.clone());
```
suzmue: Can we make the outstanding_publish already hold a copy of these so we don't need to clone on every message?

src/pubsub/src/publisher/worker.rs
```diff
@@ -112,27 +102,21 @@ impl Worker {
...
-                                }
-
-                                // Wait on all the tasks that exist right now.
-                                // We could instead tokio::spawn this as well so the publisher
-                                // can keep working on additional messages. The worker would
-                                // also need to keep track of any pending flushes, and make sure
-                                // all of those resolve as well.
-                                let mut flushing = std::mem::take(&mut inflight);
-                                while flushing.next().await.is_some() {}
+                                outstanding.flush(self.client.clone(), self.topic_name.clone());
```
suzmue: This seems to me like it has changed the behavior so that it is no longer waiting on all of the batches to have returned, which is a regression. Flush just sends the existing batch, but it doesn't wait on any of those to resolve.

Notably this doesn't appear to be caught in the tests so if we can add a test for this too that would be great.

src/pubsub/src/publisher/worker.rs
```diff
@@ -112,27 +102,21 @@ impl Worker {
...
+                                    .or_insert(OutstandingPublishes::new(
+                                        &self.topic_name,
+                                        message_limit,
+                                        byte_threshold,
+                                    ));
+                            outstanding_publishes.push(msg, self.client.clone(), self.topic_name.clone());
                         },
                         Some(ToWorker::Flush(tx)) => {
+                            // TODO(#4012): To guarantee ordering, we should wait for the
+                            // inflight batch to complete so that messages are publish in order.
```
suzmue, line 125: This comment feels like it belongs within OutstandingPublish NOT here, since this is no longer responsible for deciding when to send the message.

src/pubsub/src/publisher/worker.rs
```diff
@@ -155,15 +139,36 @@ impl Worker {
 
 #[derive(Debug, Default)]
 pub(crate) struct OutstandingPublishes {
-    pub(crate) pending_batch: Batch,
+    message_limit: u32,
+    byte_threshold: u32,
+    pending_batch: Batch,
+    inflight: Option<tokio::task::JoinHandle<()>>,
```
suzmue: inflight is unused?

src/pubsub/src/publisher/worker.rs
```diff
@@ -70,13 +68,10 @@ impl Worker {
...
-    ///    and added to the current `batch`.
-    /// 2. A timer is armed when the first message is added to a batch.
-    ///    If that timer fires, the batch is sent.
-    /// 3. A `Flush` command from the `Publisher` causes the current batch to be
-    ///    sent immediately, and all in-flight send tasks to be awaited.
-    /// 4. The `inflight` set is continuously polled to remove `JoinHandle`s for
-    ///    send tasks that have completed, preventing the set from growing indefinitely.
+    ///    and added to the message's ordering key OutstandingPublishes.
+    /// 2. A timer is continuously fired to flush all pending batches.
+    /// 3. A `Flush` command from the `Publisher` causes ordering key's batch to
```
suzmue:
```suggestion
    /// 3. A `Flush` command from the `Publisher` causes pending batches to
```

src/pubsub/src/publisher/worker.rs
```diff
@@ -155,15 +139,36 @@ impl Worker {
...
+        self.pending_batch.push(msg);
+        if self.pending_batch.len() as u32 >= self.message_limit
+            || self.pending_batch.size() >= self.byte_threshold
+        {
+            self.inflight = Some(self.pending_batch.flush(client, topic));
+        }
+    }
+
+    pub(crate) fn flush(&mut self, client: GapicPublisher, topic: String) {
+        if !self.pending_batch.is_empty() {
```
suzmue, line 201: I think this is where the comment about waiting for inflight to return should go.

suzmue (commented):
Can you also update the commit message description to describe a bit about the change that you are making (what has changed / why). Thanks!

src/pubsub/src/publisher/worker.rs

src/pubsub/src/publisher/worker.rs
PhongChuong: I moved inflight back up to Worker. I think having the task control logic purely in Worker is easier to follow then having it distributed across Worker and OutstandingPublishes.
PTAL.

src/pubsub/src/publisher/worker.rs
PhongChuong: See comment regarding moving inflight back into Worker.

In that case, I think having the comment here makes more sense.

src/pubsub/src/publisher/worker.rs
PhongChuong: See other comment.

src/pubsub/src/publisher/worker.rs
PhongChuong: I think there was a misunderstanding on flush on my part. In my head, flush is to signal to the publisher to send out all the pending batches/publish but not await until it finishes.

If the contract of flush is to await, then are we waiting for all pending publishes? Or just the pending batch?

PhongChuong:
@suzmue , Thanks for the review.
I've added inflight back to worker as having all the task control logic within Worker turns out to be better.
I think we need to discuss the Flush contract in a bit more detail in regards to what it would await on (pending batch vs pending publishes) or if it should await at all.

PTAL.

src/pubsub/src/publisher/batch.rs
```diff
@@ -69,17 +68,13 @@ impl Batch {
...
-    ) {
-        if self.is_empty() {
-            return;
-        }
+    ) -> tokio::task::JoinHandle<()> {
         let batch_to_send = Self {
             messages: self.messages.drain(..).collect(),
             messages_byte_size: self.messages_byte_size,
         };
         self.messages_byte_size = 0;
```
suzmue: should this be initial_size?

src/pubsub/src/publisher/worker.rs
suzmue: I still don't think this main loop is handling keeping track of which batches have outstanding publishes. But since this is a TODO, I am fine leaving the comment here. But it seems likely this sort of decision should be handled by the OutstandingPublish.

src/pubsub/src/publisher/worker.rs
```diff
@@ -153,17 +150,55 @@ impl Worker {
...
+            byte_threshold,
+        }
+    }
+
+    pub(crate) fn push(
+        &mut self,
+        msg: BundledMessage,
+        inflight: &mut FuturesUnordered<tokio::task::JoinHandle<()>>,
+    ) {
+        self.pending_batch.push(msg);
```
suzmue, line 189: This can be in a future PR, but what if adding the msg increases the size > the limit of the size of a single message (not just the configured max size but the actual max size we are allowed to send). I think we should:
1. flush the existing batch
2. start a new batch with the message.

src/pubsub/src/publisher/worker.rs
suzmue: This conditional still seems like a good place to decide whether we send the message. But can wait for the implementation.

src/pubsub/src/publisher/worker.rs
```diff
@@ -70,13 +70,10 @@ impl Worker {
     /// This method concurrently handles four main events:
     ///
     /// 1. Messages from the `Publisher` are received from the `rx` channel
-    ///    and added to the current `batch`.
-    /// 2. A timer is armed when the first message is added to a batch.
-    ///    If that timer fires, the batch is sent.
-    /// 3. A `Flush` command from the `Publisher` causes the current batch to be
-    ///    sent immediately, and all in-flight send tasks to be awaited.
-    /// 4. The `inflight` set is continuously polled to remove `JoinHandle`s for
```
suzmue, line 78: 4 should be added back in.

src/pubsub/src/publisher/worker.rs
PhongChuong: Added a TODO.

src/pubsub/src/publisher/worker.rs
PhongChuong: It does not handle outstanding publish yet as this is simply a refactoring.

Regarding where to handle awaiting the inflight batch, I actually think it is better handled by the Worker as it has it avoids spawning additional tasks in OutstandingPublishes to push pending batches along.

Lets discuss offline outside of this PR.

src/pubsub/src/publisher/worker.rs
PhongChuong: Done.

src/pubsub/src/publisher/batch.rs
PhongChuong: Fixed.

Issue #4052, opened 2025-12-12
Preserve error that terminates a worker


coryan:
The worker task for an object descriptor may terminate because the underlying connection had an unrecoverable error.

In this case, we send an error to all *active* ranged reads (which could be "none"). Future ranged reads get a more obscure error. We should preserve the original I/O error so we can deliver it to future reads.

That is, calling `ObjectDescriptor::read_range()` when the worker is terminated should return the error that terminated the worker.

Issue #3953, opened 2025-12-02
Run the `storage-random` test for one billion requests


coryan:
This is a smoke test to catch leaks, unexpected race conditions, deadlocks, etc.

coryan:
With far fewer than 1B requests I got this error:

```
2025-12-03T01:44:28.914357Z ERROR ThreadId(08) storage_random: error on range 0: the transport reports an error: cannot recover from an underlying read error: the service reports an error with code UNKNOWN described as: error reading a body from connection

Caused by:
    0: cannot recover from an underlying read error: the service reports an error with code UNKNOWN described as: error reading a body from connection
    1: the service reports an error with code UNKNOWN described as: error reading a body from connection
    2: status: 'Unknown error', self: "error reading a body from connection"
    3: error reading a body from connection
    4: stream error received: unexpected internal error encountered
```

That sounds like a problem in gRPC that should be recoverable with a retry / resume of some sort.

coryan:
I am also getting zero bytes transferred, but with more than zero bytes requested.  The benchmark does not have enough information to troubleshoot this.

coryan:
I am also getting:

```
2025-12-07T23:18:32.785051Z ERROR ThreadId(326) google_cloud_storage::read_resume_policy: unknown error in read resume: Error { kind: Service(ServiceDetails { status_code: None, headers: Some({}), status: Status { code: Unknown, message: "error reading a body from connection", details: [] } }), source: Some(Status { code: Unknown, message: "error reading a body from connection", source: Some(hyper::Error(Body, Error { kind: Reset(StreamId(5), INTERNAL_ERROR, Remote) })) }) }
```

That seems like a recoverable problem.

coryan:
The latest run got about 450,000 errors:

```
grep -c ,ERROR bm-${TS}.txt
442999
``

The only error that was not recoverable is documented in #4069

PR #4053, opened 2025-12-12
impl(storage): handle closed object descriptors

coryan:
The object descriptor may be closed if the worker finds an unrecoverable error. In that case, we return a reader that immediately returns an error.

Part of the work for #4052, found while working on #3953

### DIFF
--- a/src/storage/src/error.rs
+++ b/src/storage/src/error.rs
@@ -100,6 +100,8 @@ pub enum KeyAes256Error {
     InvalidLength,
 }
 
+type BoxedSource = Box<dyn std::error::Error + Send + Sync + 'static>;
+
 /// Represents an error that can occur when reading response data.
 #[derive(thiserror::Error, Debug)]
 #[non_exhaustive]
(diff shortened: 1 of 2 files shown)


src/storage/src/error.rs
```diff
@@ -210,6 +209,18 @@ pub enum ReadError {
...
     UnknownBidiRangeId(i64),
+
+    /// A `read_range()` request failed because the object descriptor is closed.
+    ///
+    /// # Troubleshooting
+    ///
+    /// The object descriptor closes only when there is an unrecoverable I/O
+    /// error. Consider using a more lenient [ReadResumePolicy].
+    ///
+    /// [ReadResumePolicy]: [crate::read_resume_policy::ReadResumePolicy]
```
dbolduc, line 220: optional nit: I don't think the brackets around `crate::read_resume_policy::ReadResumePolicy` are necessary, but I think it works either way

coryan: ack.

PR #4054, opened 2025-12-12
impl(storage): canonical url for signed urls

alvarowolfx:
Towards #3645

### DIFF
--- a/src/storage/src/storage/signed_url.rs
+++ b/src/storage/src/storage/signed_url.rs
@@ -12,32 +12,14 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use crate::{error::SigningError, signed_url::UrlStyle};
+use crate::{error::SigningError, signed_url::UrlStyle, storage::client::ENCODED_CHARS};
 use auth::signer::Signer;
 use percent_encoding::{AsciiSet, utf8_percent_encode};
 use std::collections::BTreeMap;
 
-/// https://cloud.google.com/storage/docs/request-endpoints#encoding
-/// !, #, $, &, ', (, ), *, +, ,, /, :, ;, =, ?, @, [, ]
-const PATH_ENCODE_SET: AsciiSet = AsciiSet::EMPTY
-    .add(b' ')
-    .add(b'!')
-    .add(b'#')
-    .add(b'$')
-    .add(b'&')
-    .add(b'\'')
-    .add(b'(')
-    .add(b')')
-    .add(b'*')
-    .add(b'+')
-    .add(b',')
-    .add(b':')
-    .add(b';')
-    .add(b'=')
-    .add(b'?')
-    .add(b'@')
-    .add(b'[')
-    .add(b']');
+/// Same encoding set as used in https://cloud.google.com/storage/docs/request-endpoints#encoding
+/// but for signed URLs, we do not encode '/'.
+const PATH_ENCODE_SET: AsciiSet = ENCODED_CHARS.remove(b'/');
 
 /// Creates [Signed URLs].
 ///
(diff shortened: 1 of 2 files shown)


src/storage/src/storage/signed_url.rs
```diff
@@ -14,8 +14,30 @@
 
 use crate::{error::SigningError, signed_url::UrlStyle};
 use auth::signer::Signer;
+use percent_encoding::{AsciiSet, utf8_percent_encode};
 use std::collections::BTreeMap;
-use tokio::time::Duration;
+
+/// https://cloud.google.com/storage/docs/request-endpoints#encoding
+/// !, #, $, &, ', (, ), *, +, ,, /, :, ;, =, ?, @, [, ]
+const PATH_ENCODE_SET: AsciiSet = AsciiSet::EMPTY
```
coryan: This is suspiciously similar to:

https://github.com/googleapis/google-cloud-rust/blob/0b42085164397827ea1a5705c63bb4380960289d/src/storage/src/storage/client.rs#L699-L718

We probably want to share both definitions.  And I am 99.44% sure you want to percent-encode `\n` and `\r` (which are legal characters in an object name) too.

src/storage/src/storage/signed_url.rs
```diff
@@ -306,6 +395,17 @@ impl SignedUrlBuilder {
...
 
+    fn resolve_endpoint(&self) -> String {
+        if let Some(endpoint) = self.endpoint.clone() {
+            if !endpoint.starts_with("http") {
+                return format!("https://{}", endpoint);
+            }
+            return endpoint;
+        }
+
+        format!("https://storage.{}", self.universe_domain.clone())
```
coryan: Why clone?

```suggestion
        format!("https://storage.{}", self.universe_domain)
```

src/storage/src/storage/signed_url.rs
```diff
@@ -306,6 +395,17 @@ impl SignedUrlBuilder {
         self
     }
 
+    fn resolve_endpoint(&self) -> String {
+        if let Some(endpoint) = self.endpoint.clone() {
+            if !endpoint.starts_with("http") {
+                return format!("https://{}", endpoint);
+            }
+            return endpoint;
+        }
```
coryan: hmm what about `http.foo.bar.example.com`?  And why clone when you do not need to?

```suggestion
        match self.endpoint.as_ref() {
            Some(e) if e.starts_with("https://") => return e.clone(),
            Some(e) if e.starts_with("http://") => return e.clone(),
            Some(e) => return format!("https://{e}"),
            None => (),
        };
```

src/storage/src/storage/signed_url.rs
```diff
@@ -63,6 +85,73 @@ enum SigningScope {
...
+    fn bucket_endpoint(&self, endpoint: &str, url_style: UrlStyle) -> String {
+        let bucket_name = self.bucket_name();
+        let scheme = if endpoint.starts_with("http://") {
+            "http"
+        } else {
+            "https"
+        };
+        let endpoint = endpoint
+            .trim_start_matches("http://")
+            .trim_start_matches("https://");
```
coryan: Seems to me that you should be use `Url::parse` instead of manually parsing the endpoint.

src/storage/src/storage/signed_url.rs
```diff
@@ -63,6 +85,73 @@ enum SigningScope {
...
+                format!("{scheme}://{bucket_name}.{endpoint}")
+            }
+        }
+    }
+
+    fn canonical_uri(&self, url_style: UrlStyle) -> String {
+        let bucket_name = self.bucket_name();
+        match self {
+            SigningScope::Object(_, object) => {
+                let encoded_object = utf8_percent_encode(object, &PATH_ENCODE_SET).to_string();
```
coryan: `to_string()` for this type just calls `Display::fmt`.  So I think you can avoid allocations and reformatting.

here you would write:


```suggestion
                let encoded_object = utf8_percent_encode(object, &PATH_ENCODE_SET);
```

PR #4064, opened 2025-12-12
impl(storage): preserve bidi streaming errors

### LINKED ISSUE #4052: Preserve error that terminates a worker

coryan:
When the underlying streaming RPC has an unrecoverable error we need to preserve the error and return it on future `read_range()` requests.

Fixes #4052

### DIFF
--- a/src/storage/src/storage/bidi/worker.rs
+++ b/src/storage/src/storage/bidi/worker.rs
@@ -85,6 +85,8 @@ where
                 },
             }
         };
+        drop(rx);
+        drop(tx);
         let Some(e) = error else {
             // A successfully closed stream *and* there are no more readers.
             return Ok(());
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/bidi/worker.rs
```diff
@@ -79,16 +79,24 @@ where
...
             }
+        };
+        drop(rx);
+        drop(tx);
+        let Some(e) = error else {
+            // A successfully closed stream *and* there are no more readers.
+            return Ok(());
+        };
+        // Return errors for any future readers.
+        while let Some(mut r) = requests.recv().await {
```
dbolduc, line 95: comment for myself: ah, so as long as there is a sender, we keep this thing running in case we need to return them an error. It breaks when the sender(s) are all dropped

PR #4074, opened 2025-12-15
impl(pubsub): introduce subscriber stream

dbolduc:
Part of the work for #3957 

Introduce an **internal** `Stub` abstraction and a function to open a `StreamingPull` bidirectional stream

For context, in future PRs we will add:
- `acknowledge` and `modify_ack_deadline` fns to the `Stub` trait
- a concrete `Transport` implementation of `Stub` which uses a `gaxi::grpc::Client` to send RPCs.

### DIFF
--- a/src/pubsub/src/subscriber/stream.rs
+++ b/src/pubsub/src/subscriber/stream.rs
@@ -16,7 +16,7 @@ use super::stub::Stub;
 use crate::google::pubsub::v1::StreamingPullRequest;
 use crate::{Error, Result};
 use gax::options::RequestOptions;
-use tokio::sync::mpsc::{Sender, channel};
+use tokio::sync::mpsc;
 
 /// Open a stream for the `StreamingPull` RPC.
 ///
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/stream.rs
```diff
@@ -0,0 +1,143 @@
...
+pub(crate) async fn open_stream<T>(
+    inner: T,
+    initial_req: StreamingPullRequest,
+) -> Result<(<T as Stub>::Stream, Sender<StreamingPullRequest>)>
+where
+    T: Stub,
+{
+    // The only writes we perform are keepalives, which are sent so infrequently
+    // that we don't fear any back pressure on this channel.
+    let (request_tx, request_rx) = channel(1);
```
suzmue: nit: light preference to keep mpsc::channel and mspc::Sender to make it really easy to see what kind of channel and sender we are using.

dbolduc: mehh I don't really agree, but don't feel strongly, so done. :shrug: 

It doesn't help distinguish from [`std::sync::mpsc::channel`](https://doc.rust-lang.org/stable/std/sync/mpsc/fn.channel.html) (which has confused me in a review before, lol)

And we often have to look at the use statements to know which `Result`/`Error` type the code is talking about.

PR #4079, opened 2025-12-16
fix: Add `lazy_static` workspace dependency.

westarle:


dbolduc (approved):

westarle:
not needed, the problem was generating info for clients with no services.

PR #4088, opened 2025-12-17
docs(compute): document all features

### LINKED ISSUE #4048: The documentation for some crates is incomplete
I think the documentation for `google-cloud-compute-v1` is generated with (only) the default features:

https://docs.rs/google-cloud-compute-v1/latest/google_cloud_compute_v1/client/index.html

We need to enable all the features. That may make the documentation expensive to generate, but it is kind of useless as it is.

dbolduc:
Fixes #4048 (well, when we publish the crate)

Note that `compute` is the only crate for which we exclude features from the default set.


src/generated/cloud/compute/v1/Cargo.toml
```diff
@@ -254,6 +254,9 @@ zone-operations = []
 # Enables `client::Zones` and all the types it depends on.
 zones = []
 
+[package.metadata.docs.rs]
+all-features = true
```
coryan, line 258: Can you check the memory requirements? https://github.com/rust-lang/docs.rs/issues/3011 has some commands to get that info.

We may need to ask for more RAM.

dbolduc:
```
        Command being timed: "cargo +nightly docs-rs -p google-cloud-compute-v1"
        Maximum resident set size (kbytes): 4850372
```

so we will probably need to ask for 8GB. I am not sure if it is better to ask before or after releasing the crate.

dbolduc:
Closing, because this is still blocked, and I missed other changes.

PR #4100, opened 2025-12-18
impl(storage): return values from `ObjectDescriptor`

coryan:
Returning values from `ObjectDescriptor::object()` and `ObjectDescriptor::headers()` will make it easier to change these dynamically. For example if we want to use the headers of the last request vs. the headers of the first request.

### DIFF
--- a/src/storage/src/object_descriptor.rs
+++ b/src/storage/src/object_descriptor.rs
@@ -122,8 +122,8 @@ impl ObjectDescriptor {
     /// # async fn sample() -> anyhow::Result<()> {
     /// let descriptor = open();
     /// // Often useful when troubleshooting problems with Google Support.
-    /// let header = descriptor.headers().get("x-guploader-uploadid");
-    /// println!("debugging header = {:?}", header);
+    /// let headers = descriptor.headers();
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/bidi/transport.rs
```diff
@@ -59,8 +59,8 @@ impl ObjectDescriptorTransport {
 }
 
 impl ObjectDescriptor for ObjectDescriptorTransport {
-    fn object(&self) -> &Object {
-        self.object.as_ref()
+    fn object(&self) -> Object {
+        self.object.as_ref().clone()
```
dbolduc, line 66: nit: `as_ref()` unnecessary?

dbolduc (dismissed):

src/storage/src/storage/bidi/transport.rs
coryan: That would clone the `Arc`, but we need to clone the thing inside the `Arc`.  Seems sufficiently obscure to add a comment.

dbolduc: My mistake. I didn't look at the type.

dbolduc (approved):

PR #4102, opened 2025-12-18
impl(storage): future proof `ObjectDescriptor`

coryan:


coryan:
This is in preparation of `open_and_read()` (see #4104).  I have convinced myself that it works and we can make this changes to stabilize the APIs while I work on tests and so forth.

/FYI: @cjc25 @BenWhitehead

src/storage/src/storage/open_object.rs
```diff
@@ -63,7 +63,8 @@ where
     /// # Ok(()) }
     /// ```
     pub async fn send(self) -> Result<ObjectDescriptor> {
```
dbolduc, line 65: Will this return type change in a follow up?

I expected `Result<(ObjectDescriptor, Vec<ReadObjectResponse>)>`

coryan: I am planning to add:

```rust
pub async fn send_and_read(self, ReadRange) -> Result<(ObjectDescriptor, ReadObjectResponse)>
```

and (maybe):

```rust
pub async fn send_and_read_many(self, Vec<ReadRange>) -> Result<(ObjectDescriptor, Vec<ReadObjectResponse>)>
```

dbolduc: Consider a builder method like:

```rs
fn with_initial_reads<T>(ranges: T)
where T: IntoIterator<Item = ReadRange> -> Self;
```

with the one `send()` method.

Con:

- empty vector when no initial reads can be clunky
- useless if reading many is not a real workflow

Pro:

- one API

dbolduc: @BenWhitehead - any opinions?

is reading many initial ranges when you open an object even a real world use case?

PR #4115, opened 2025-12-22
impl(pubsub): timed ack/nack flushes, lease extensions

dbolduc:
Part of the work for #3957

Add basic batching logic for Acks and lease extensions. Currently the batches fire on a timer.

This logic belongs in the `LeaseState`, so we can consider batch size / outstanding RPCs / etc. in the future.

Interfaces are made `pub(crate)` because the lease loop (coming shortly) will need them.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -14,9 +14,32 @@
 
 use super::leaser::Leaser;
 use std::collections::HashSet;
+use tokio::time::{Duration, Instant, Interval, interval_at};
+
+pub(crate) struct LeaseOptions {
+    /// How often we flush acks/nacks
+    pub(crate) flush_period: Duration,
+    /// How long we wait for the initial flush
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -110,6 +170,13 @@ pub(crate) mod tests {
     use super::super::leaser::tests::MockLeaser;
     use super::*;
 
+    fn test_interval() -> Interval {
```
suzmue, line 173: Are these chosen so that they never fire?

dbolduc: That's what I was originally thinking, but it doesn't matter, because nothing awaits them. Any `Interval` would do.

dbolduc: Merged, because I had a PR blocked on this one.

Just sent #4119 to answer your question with a comment in the code.

PR #4116, opened 2025-12-22
impl(pubsub): default leaser, based on `Stub`

dbolduc:
Part of the work for #3957 

Implement a `Leaser` in terms of a `Stub`. Basically we are mapping leaser actions to RPCs. The real implementation will use a transport stub as the `T`.

We are deferring observability for failed requests. For now we just eat up the errors.

### DIFF
--- a/src/pubsub/src/subscriber/leaser.rs
+++ b/src/pubsub/src/subscriber/leaser.rs
@@ -12,6 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::stub::Stub;
+use crate::model::{AcknowledgeRequest, ModifyAckDeadlineRequest};
+use gax::options::RequestOptions;
+use gax::retry_policy::NeverRetry;
+use std::sync::Arc;
+
 /// A trait representing leaser actions
 ///
 /// We stub out the interface, in order to test the lease management.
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/leaser.rs
```diff
@@ -39,4 +103,82 @@ pub(crate) mod tests {
...
+    async fn ack() {
+        let mut mock = MockStub::new();
+        mock.expect_acknowledge().times(1).return_once(|r, o| {
+            assert_eq!(
+                r.subscription,
+                "projects/my-project/subscriptions/my-subscription"
+            );
+            assert_eq!(r.ack_ids, test_ids(0..10));
+            assert!(
+                format!("{o:?}").contains("NeverRetry"),
```
suzmue, line 117: Surprised there is not a more direct way to compare these. Maybe a comparison function could be useful for tests in the future (not necessary for this PR and might be overkill).

suzmue (approved):

src/pubsub/src/subscriber/leaser.rs
dbolduc: I think we could use the retry policy and verify that its behavior is identical to `NeverRetry`, e.g. it would not retry an UNAVAILABLE.

buuuuuuut I am lazy, and there is already a precedence for being lazy:

https://github.com/googleapis/google-cloud-rust/blob/df77ee8173df791e24c70b40131e5beda73fb65d/src/storage/src/storage/request_options.rs#L153

Issue #3941, opened 2025-12-01
Set up subscriber surface


dbolduc:
Googlers can see: go/cloud-rust:subscriber-api (not approved/final circa 2025-12-01)

- [ ] Add message data type
- [ ] Add handler and ack traits
- [ ] Add `Session`
  - with a way to create sessions from a `futures::Stream<(Message, Handler)>`
- [ ] Add streaming pull builder
- [ ] Add client

For the handler and session, we do not need the concrete implementation yet.

dbolduc:
I am going to track the documentation cleanups in a separate issue: #4508

So this is done.

PR #4118, opened 2025-12-22
impl(pubsub): `StreamingPull` builder

dbolduc:
Part of the work for #3941 

Add a builder for a streaming pull bidi RPC. This class will eventually be public, and thus it must have documentation.

For now, this thing just sets options. It will gain a `fn start(self) -> Result<Session>` later, when `Session` is a thing.

### DIFF
--- a/src/pubsub/src/subscriber/builder.rs
+++ b/src/pubsub/src/subscriber/builder.rs
@@ -15,6 +15,8 @@
 use super::stub::Stub;
 use std::sync::Arc;
 
+const MIB: i64 = 1024 * 1024;
+
 /// Builder for the `client::Subscriber::streaming_pull` method.
 pub struct StreamingPull<S>
 where
@@ -38,7 +40,7 @@ where
             subscription,
             ack_deadline_seconds: 10,
             max_outstanding_messages: 1000,
-            max_outstanding_bytes: 100 * 1024 * 1024,
+            max_outstanding_bytes: 100 * MIB,
         }
     }
 
@@ -48,12 +50,14 @@ where
     /// incoming message. Note that this value is independent of the deadline
     /// configured on the server-side subscription.
     ///
+    /// If the server does not hear back from the client within this deadline
+    /// (e.g. if an application crashes), it will resend any unacknowledged
+    /// messages to another subscriber.
+    ///
     /// The minimum deadline you can specify is 10 seconds. The maximum deadline
     /// you can specify is 600 seconds (10 minutes).
     ///
     /// The default value is 10 seconds.
-    ///
-    /// To use the server-side subscription deadline, specify a value of `0`.
     ///
     /// # Example
     ///
@@ -81,6 +85,8 @@ where
     ///
     /// Use a value <= 0 to set no limit on the number of outstanding messages.
     ///
+    /// The default value is 1000 messages.
+    ///
     /// # Example
     ///
     /// ```no_rust
@@ -88,7 +94,7 @@ where
     /// # async fn sample() -> anyhow::Result<()> {
     /// # let client = Subscriber::builder().build().await?;
     /// let session = client.streaming_pull("projects/my-project/subscriptions/my-subscription")
-    ///     .set_max_outstanding_messages(100000)
+    ///     .set_max_outstanding_messages(2000)
     ///     .start();
     /// # Ok(()) }
     /// ```
@@ -107,14 +113,17 @@ where
     ///
     /// Use a value <= 0 to set no limit on the number of outstanding bytes.
     ///
+    /// The default value is 100 MiB.
+    ///
     /// # Example
     ///
     /// ```no_rust
     /// # use google_cloud_pubsub::client::Subscriber;
     /// # async fn sample() -> anyhow::Result<()> {
     /// # let client = Subscriber::builder().build().await?;
+    /// const MIB: i64 = 1024 * 1024;
     /// let session = client.streaming_pull("projects/my-project/subscriptions/my-subscription")
-    ///     .set_max_outstanding_bytes(1024 * 1024 * 1024)
+    ///     .set_max_outstanding_bytes(200 * MIB)
     ///     .start();
     /// # Ok(()) }
     /// ```
@@ -129,6 +138,8 @@ mod tests {
     use super::super::stub::tests::MockStub;
     use super::*;
 
+    const KIB: i64 = 1024;
+
     #[test]
     fn reasonable_defaults() {
         let mock = MockStub::new();
@@ -142,14 +153,14 @@ mod tests {
         );
         assert_eq!(builder.ack_deadline_seconds, 10);
         assert!(
-            1000000 > builder.max_outstanding_messages && builder.max_outstanding_messages > 100,
-            "{}",
+            100_000 > builder.max_outstanding_messages && builder.max_outstanding_messages > 100,
+            "max_outstanding_messages={}",
             builder.max_outstanding_messages
         );
         assert!(
-            builder.max_outstanding_bytes > 100000,
-            "{}",
-            builder.max_outstanding_messages
+            builder.max_outstanding_bytes > 100 * KIB,
+            "max_outstanding_bytes={}",
+            builder.max_outstanding_bytes
         );
     }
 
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// ```no_rust
+    /// # use google_cloud_pubsub::client::Subscriber;
+    /// # async fn sample() -> anyhow::Result<()> {
+    /// # let client = Subscriber::builder().build().await?;
+    /// let session = client.streaming_pull("projects/my-project/subscriptions/my-subscription")
+    ///     .set_max_outstanding_bytes(1024 * 1024 * 1024)
+    ///     .start();
+    /// # Ok(()) }
+    /// ```
+    pub fn set_max_outstanding_bytes<T: Into<i64>>(mut self, v: T) -> Self {
```
suzmue, line 130: thoughts on this being a u64 vs i64? We already use 0 to indicate no limit so negative provides no additional value.

src/pubsub/src/subscriber/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    /// Flow control settings for the maximum number of outstanding messages.
+    ///
+    /// The server will stop sending messages to a client when this many
+    /// messages are outstanding (i.e. that have not been acked or nacked).
+    ///
+    /// The server resumes sending messages when the outstanding message count
+    /// drops below this value.
+    ///
+    /// Use a value <= 0 to set no limit on the number of outstanding messages.
```
suzmue, line 86: nit: could be nice to include the default in this doc comment.

Based on the tests it seems omitting this was intentional. Why is ack_deadline_seconds different from the other two?

src/pubsub/src/subscriber/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+    /// Use a value <= 0 to set no limit on the number of outstanding bytes.
+    ///
+    /// # Example
+    ///
+    /// ```no_rust
+    /// # use google_cloud_pubsub::client::Subscriber;
+    /// # async fn sample() -> anyhow::Result<()> {
+    /// # let client = Subscriber::builder().build().await?;
+    /// let session = client.streaming_pull("projects/my-project/subscriptions/my-subscription")
+    ///     .set_max_outstanding_bytes(1024 * 1024 * 1024)
```
suzmue: nit: Might be nice to include a comment with the human readable version of this number  (1 GiB) in the example. Maybe in the example header?

src/pubsub/src/subscriber/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+    ///
+    /// The server will stop sending messages to a client when this many bytes
+    /// of messages are outstanding (i.e. that have not been acked or nacked).
+    ///
+    /// The server resumes sending messages when the outstanding byte count
+    /// drops below this value.
+    ///
+    /// Use a value <= 0 to set no limit on the number of outstanding bytes.
+    ///
+    /// # Example
```
suzmue, line 118: nit: Same comment about default for this setting.

src/pubsub/src/subscriber/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+        assert_eq!(builder.ack_deadline_seconds, 10);
+        assert!(
+            1000000 > builder.max_outstanding_messages && builder.max_outstanding_messages > 100,
+            "{}",
+            builder.max_outstanding_messages
+        );
+        assert!(
+            builder.max_outstanding_bytes > 100000,
+            "{}",
+            builder.max_outstanding_messages
```
suzmue:
```suggestion
            builder.max_outstanding_bytes
```

src/pubsub/src/subscriber/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+            "projects/my-project/subscriptions/my-subscription".to_string(),
+        );
+        assert_eq!(
+            builder.subscription,
+            "projects/my-project/subscriptions/my-subscription"
+        );
+        assert_eq!(builder.ack_deadline_seconds, 10);
+        assert!(
+            1000000 > builder.max_outstanding_messages && builder.max_outstanding_messages > 100,
+            "{}",
```
suzmue: nit: Since we are adding a custom error message, it would be nice to provide more info in the message.

Maybe even just add which field failed, the test name indicates that the problem is a non-reasonable default so its likely enough context to understand:
```suggestion
            "max_outstanding_messages = {}",
```

src/pubsub/src/subscriber/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+            builder.max_outstanding_messages
+        );
+        assert!(
+            builder.max_outstanding_bytes > 100000,
```
suzmue: nit: Could make sense to use a number of KiB instead, since we are mostly dealing with bytes in terms of increments of 1024 (maybe use 100 KiB = 100 * 1024)?

src/pubsub/src/subscriber/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+    }
+
+    /// Sets the ack deadline to use for the stream.
+    ///
+    /// This value represents how long the application has to ack or nack an
+    /// incoming message. Note that this value is independent of the deadline
+    /// configured on the server-side subscription.
+    ///
+    /// The minimum deadline you can specify is 10 seconds. The maximum deadline
+    /// you can specify is 600 seconds (10 minutes).
```
suzmue, line 58: What do you mean by "can" here? It seems like you can set to any values. Will something error? Will it get truncated?

suzmue (dismissed):

src/pubsub/src/subscriber/builder.rs
```diff
@@ -0,0 +1,174 @@
...
+            subscription,
+            ack_deadline_seconds: 10,
+            max_outstanding_messages: 1000,
+            max_outstanding_bytes: 100 * 1024 * 1024,
+        }
+    }
+
+    /// Sets the ack deadline to use for the stream.
+    ///
+    /// This value represents how long the application has to ack or nack an
```
PhongChuong, line 49: Nit. Perhaps clarify what happens in the case where the application exceed the deadline for acking/nacking.

src/pubsub/src/subscriber/builder.rs
dbolduc: Done.

src/pubsub/src/subscriber/builder.rs
dbolduc: D'oh. Fixed, thanks

src/pubsub/src/subscriber/builder.rs
dbolduc: Fixed here and below

src/pubsub/src/subscriber/builder.rs
dbolduc: Added here and below.

> Why is ack_deadline_seconds different from the other two?

The ack deadline default seemed more relevant to the application. Like if they need more than 10s to process a message, they should be setting the option.

The flow control options are more about what your CPU can handle / going fast. That is probably still useful for an application developer to know. I am convinced.

src/pubsub/src/subscriber/builder.rs
dbolduc: > What do you mean by "can" here? It seems like you can set to any values.

Lol. Good point.

FWIW I was copying the proto documentation:

https://github.com/googleapis/googleapis/blob/60f1e6ad747f1a603902b0d87867539c7df22a3b/google/pubsub/v1/pubsub.proto#L2164-L2167

> Will something error?

Quick testing says no... :thinking: 

> Will it get truncated?

It might on the server side. I am not sure. Let me ask the Pub/Sub folks.

dbolduc: Sorry, let me take a TODO on this one: #4134

If we need to clamp the input, it will be a super easy PR.

src/pubsub/src/subscriber/builder.rs
dbolduc: Done.

src/pubsub/src/subscriber/builder.rs
dbolduc: Added

src/pubsub/src/subscriber/builder.rs
dbolduc: Done.

src/pubsub/src/subscriber/builder.rs
dbolduc: > We already use 0 to indicate no limit so negative provides no additional value.

Agreed. Wish you were there to review the service protos when they were originally published:

https://github.com/googleapis/googleapis/blob/60f1e6ad747f1a603902b0d87867539c7df22a3b/google/pubsub/v1/pubsub.proto#L2199

I think it's best to match the proto types.

Also aside on `ack_deadline_seconds`: It is a code smell to have types like `foo_seconds: i32`. Typically you would just call it `foo: Duration`. But given that the server only accepts integer values, I went ahead with the integer type with the unit (seconds) encoded in the name. :shrug:

dbolduc (commented):
PTAL

src/pubsub/src/subscriber/builder.rs
suzmue: Sounds reasonable to me.

PR #4119, opened 2025-12-23
cleanup(pubsub): `test_interval()` review comments

dbolduc:
This could be simpler, and a comment could explain that its value is meaningless.

In response to: https://github.com/googleapis/google-cloud-rust/pull/4115#discussion_r2641577200


dbolduc:
> It currently reads that this PR is going to add some tests, but it does not. 

Ah. Removed the ambiguity.

Aside: I use `test:` not `cleanup:` for PRs that add tests

PR #4120, opened 2025-12-23
impl(pubsub): lease loop

dbolduc:
Part of the work for #3957 

Add a lease loop. The loop runs on end, performing any of the following actions:

- handle a shutdown
- flush acks/nacks
- extend leases
- accept a new message
- handle incoming acks/nacks

I combined these things into a struct, but I may change my mind later.

## On testing...

TIL: `mock.checkpoint()`. It means check all the expectations at this exact moment.

I found it easiest to use an `Arc<Mutex<Mock>>` which allowed me to give the mock to the lease loop AND set expectations on the mock outside of the lease loop.

I also tried to recover the time at which events happen from within the mock expectations, but decided that was more complicated.

Because `tokio::test` is single threaded (if you want to `start_paused = true`), I found that I needed to yield the current execution so tokio would actually launch the lease loop. Otherwise it would not launch the lease loop until the first `.await` point, and the timing was all messed up.

### DIFF
--- a/src/pubsub/src/subscriber/leaser.rs
+++ b/src/pubsub/src/subscriber/leaser.rs
@@ -106,6 +106,19 @@ pub(crate) mod tests {
         }
     }
 
+    #[async_trait::async_trait]
+    impl Leaser for Arc<Mutex<MockLeaser>> {
+        async fn ack(&self, ack_ids: Vec<String>) {
+            self.lock().await.ack(ack_ids).await
+        }
+        async fn nack(&self, ack_ids: Vec<String>) {
+            self.lock().await.nack(ack_ids).await
+        }
+        async fn extend(&self, ack_ids: Vec<String>) {
+            self.lock().await.extend(ack_ids).await
+        }
+    }
+
     #[tokio::test]
     async fn ack() {
         let mut mock = MockStub::new();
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_loop.rs
```diff
@@ -0,0 +1,399 @@
...
+// limitations under the License.
+
+use super::handler::AckResult;
+use super::lease_state::{LeaseEvent, LeaseOptions, LeaseState};
+use super::leaser::Leaser;
+use tokio::sync::mpsc::{UnboundedSender, unbounded_channel};
+use tokio::task::JoinHandle;
+use tokio_util::sync::CancellationToken;
+
+/// A convenience struct that groups the components of the lease loop
```
suzmue: nit: all these comments should end with periods.

> Comments should usually be complete sentences. Start with a capital letter, end with a period (.). - https://doc.rust-lang.org/stable/style-guide/#comments

src/pubsub/src/subscriber/lease_loop.rs
```diff
@@ -0,0 +1,399 @@
...
+        // Ack 10 messages
+        for i in 0..10 {
+            lease_loop.ack_tx.send(AckResult::Ack(test_id(i)))?;
+        }
+
+        // Confirm initial state
+        mock.lock().await.checkpoint();
+
+        // Advance to and validate the first flush
+        {
```
suzmue, line 121: Just curious, is the block providing additional function, or is this just organizational?

suzmue (dismissed):

src/pubsub/src/subscriber/lease_loop.rs
dbolduc: organizational. I struggled to write the test succinctly.

src/pubsub/src/subscriber/lease_loop.rs
dbolduc: Fixed

PR #4121, opened 2025-12-23
test: generate diff - do not merge

zhumin8:
This is a test PR to show differences in `librarian generate --all` from `sidekick refresh-all`.
For more context https://github.com/googleapis/librarian/issues/3330

PR #4122, opened 2025-12-23
chore: remove placeholder file from longrunning

zhumin8:
This file looks like a leftover from https://github.com/googleapis/google-cloud-rust/pull/198. Other placeholder files created in the same PR were removed, e.g. src/generated/type/src/main.rs, src/generated/cloud/secretmanager/v1/src/main.rs.

Encountered this file when comparing generate results, see https://github.com/googleapis/google-cloud-rust/pull/4121.

Let me know if I am misunderstanding or missing context.


dbolduc (approved):
Thanks!

PR #4124, opened 2025-12-23
impl(pubsub): message ordering in Publisher

PhongChuong:
Support messaging ordering in Publisher by creating a new BatchWorker which handles that batching and message ordering logic.

When the Worker receives an Publish operation, it demultiplex it into the corresponding BatchWorker which ensures that there is at most one inflight batch at any given time.
When the Worker receives a Flush operation, it flushes all BatchWorker pending message and waits until the operation is complete.

As of now, the set of BatchWorkers grows indefinitely. This is be address in the next PR.

### DIFF
--- a/src/pubsub/src/publisher/worker.rs
+++ b/src/pubsub/src/publisher/worker.rs
@@ -94,6 +94,7 @@ impl Worker {
         let mut batch_workers: HashMap<String, mpsc::UnboundedSender<ToBatchWorker>> =
             HashMap::new();
         let delay = self.batching_options.delay_threshold;
+        let batch_worker_error_msg = "Batch worker should not close the channel";
 
         let timer = tokio::time::sleep(delay);
         // Pin the timer to the stack.
@@ -108,7 +109,7 @@ impl Worker {
                         let (tx, _) = oneshot::channel();
                         batch_worker
                             .send(ToBatchWorker::Flush(tx))
-                            .expect("Batch worker should not close the channel");
+                            .expect(batch_worker_error_msg);
                     }
                     timer.as_mut().reset(tokio::time::Instant::now() + delay);
                 }
@@ -133,7 +134,7 @@ impl Worker {
                                 });
                             batch_worker
                                 .send(ToBatchWorker::Publish(msg))
-                                .expect("Batch worker should not close the channel");
+                                .expect(batch_worker_error_msg);
                         },
                         Some(ToWorker::Flush(tx)) => {
                             let mut flush_set = JoinSet::new();
@@ -141,7 +142,7 @@ impl Worker {
                                 let (tx, rx) = oneshot::channel();
                                 batch_worker
                                     .send(ToBatchWorker::Flush(tx))
-                                    .expect("Batch worker should not close the channel");
+                                    .expect(batch_worker_error_msg);
                                 flush_set.spawn(rx);
                             }
                             // Wait on all the tasks that exist right now.
@@ -161,7 +162,7 @@ impl Worker {
                                 let (tx, rx) = oneshot::channel();
                                 batch_worker
                                     .send(ToBatchWorker::Flush(tx))
-                                    .expect("Batch worker should not close the channel");
+                                    .expect(batch_worker_error_msg);
                                 flush_set.spawn(rx);
                             }
                             flush_set.join_all().await;
@@ -183,7 +184,7 @@ pub(crate) struct BatchWorker {
     batching_options: BatchingOptions,
     rx: mpsc::UnboundedReceiver<ToBatchWorker>,
     pending_batch: Batch,
-    pending_msges: VecDeque<BundledMessage>,
+    pending_msgs: VecDeque<BundledMessage>,
 }
 
 impl BatchWorker {
@@ -199,7 +200,7 @@ impl BatchWorker {
             client,
             batching_options,
             rx,
-            pending_msges: VecDeque::new(),
+            pending_msgs: VecDeque::new(),
         }
     }
 
@@ -210,9 +211,9 @@ impl BatchWorker {
             || self.pending_batch.size() >= self.batching_options.byte_threshold
     }
 
-    // Move pending publishes to the pending batch respecting batch thresholds.
+    // Move pending messages to the pending batch respecting batch thresholds.
     pub(crate) fn move_to_batch(&mut self) {
-        while let Some(publish) = self.pending_msges.pop_front() {
+        while let Some(publish) = self.pending_msgs.pop_front() {
             self.pending_batch.push(publish);
             if self.at_batch_threshold() {
                 break;
@@ -238,6 +239,8 @@ impl BatchWorker {
         // While it is possible to use Some(JoinHandle) here as there is at max
         // a single inflight task at any given time, the use of JoinSet
         // simplify the managing the inflight JoinHandle.
+        // TODO(#4012): There is no inflight restriction when the ordering key is "".
+        // Add handling of this special case.
         let mut inflight = JoinSet::new();
         loop {
             tokio::select! {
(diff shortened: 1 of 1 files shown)


src/pubsub/src/publisher/worker.rs
```diff
@@ -69,38 +78,37 @@ impl Worker {
...
-                _ = inflight.next(), if !inflight.is_empty() => {},
-                // Handle timer events.
+                // Currently, the batch worker periodically flush on a shared timer. If needed,
+                // this can be moved into the batch worker such that each are running on a
+                // separate timer.
                 _ = &mut timer => {
-                    for (_, outstanding) in pending_batches.iter_mut() {
-                        outstanding.flush(&mut inflight);
+                    for (_, batch_worker) in batch_workers.iter_mut() {
+                        let (tx, _) = oneshot::channel();
```
suzmue, line 109: I'm confused, why do we need this channel?

src/pubsub/src/publisher/worker.rs
```diff
@@ -109,39 +117,54 @@ impl Worker {
...
-                            // TODO(#4012): To guarantee ordering, we should wait for the
-                            // inflight batch to complete so that messages are publish in order.
-                            for (_, outstanding) in pending_batches.iter_mut() {
-                                outstanding.flush(&mut inflight);
+                            let mut flush_set = JoinSet::new();
+                            for (_, batch_worker) in batch_workers.iter_mut() {
+                                let (tx, rx) = oneshot::channel();
+                                batch_worker
+                                    .send(ToBatchWorker::Flush(tx))
+                                    .expect("Batch worker should not close the channel");
```
suzmue: nit: Consider using a helper message or helper function to avoid the duplication of this error message all over.

src/pubsub/src/publisher/worker.rs
```diff
@@ -152,57 +175,107 @@ impl Worker {
     }
 }
 
+/// A background worker that continuously handle Publisher commands for a specific ordering key.
```
suzmue:
```suggestion
/// A background worker that continuously handles Publisher commands for a specific ordering key.
```

src/pubsub/src/publisher/worker.rs
```diff
@@ -152,57 +175,107 @@ impl Worker {
...
     topic: String,
     client: GapicPublisher,
-    message_limit: u32,
-    byte_threshold: u32,
+    batching_options: BatchingOptions,
+    rx: mpsc::UnboundedReceiver<ToBatchWorker>,
     pending_batch: Batch,
-    // TODO(#4012): Track pending messages as within key message ordering is
-    // not currently respected during a failure.
+    pending_msges: VecDeque<BundledMessage>,
```
suzmue:
```suggestion
    pending_msgs: VecDeque<BundledMessage>,
```

suzmue: TIL: VecDeque

src/pubsub/src/publisher/worker.rs
```diff
@@ -152,57 +175,107 @@ impl Worker {
...
+    /// 2. A `Flush` command from the `Publisher` causes all the pending messages
+    ///    to be flushed respecting the configured batch size and message ordering.
+    /// 4. A `inflight` batch completion causes the next batch to send if it satisfies
+    ///    the configured batch threshold.
+    ///
+    /// The loop terminates when the `rx` channel is closed, which happens when the
+    /// `Worker` drops the Sender.
+    pub(crate) async fn run(mut self) {
+        // While it is possible to use Some(JoinHandle) here as there is at max
+        // a single inflight task at any given time, the use of JoinSet
```
suzmue, line 240: This isn't true when ordering key is "".

src/pubsub/src/publisher/worker.rs
```diff
@@ -152,57 +175,107 @@ impl Worker {
...
+                            }
+                        },
+                        Some(ToBatchWorker::Flush(tx)) => {
+                            if !inflight.is_empty() {
+                                inflight.join_next().await;
+                            }
+                            loop {
+                                if self.pending_batch.is_empty() && self.pending_msges.is_empty() {
+                                    break;
+                                }
```
suzmue: This can be a while loop.
```suggestion
                            while !self.pending_batch.is_empty() || !self.pending_msges.is_empty() {
```

src/pubsub/src/publisher/worker.rs
```diff
@@ -152,57 +175,107 @@ impl Worker {
...
-        {
-            inflight.push(
-                self.pending_batch
-                    .flush(self.client.clone(), self.topic.clone()),
-            );
+        self.pending_batch.len() as u32 >= self.batching_options.message_count_threshold
+            || self.pending_batch.size() >= self.batching_options.byte_threshold
+    }
+
+    // Move pending publishes to the pending batch respecting batch thresholds.
```
suzmue:
```suggestion
    // Move pending messages to the pending batch respecting batch thresholds.
```

suzmue (changes requested):
As far as ordering, this looks pretty good. However, I think this removed support for non-ordered messages where we can have multiple inflight batches. 

Please let me know if I'm mistaken.

src/pubsub/src/publisher/worker.rs
suzmue: Oh I see its because you don't want to wait on the result.

if this doesn't use a lot of resources seems fine.

src/pubsub/src/publisher/worker.rs
```diff
@@ -152,57 +175,107 @@ impl Worker {
...
+                        Some(ToBatchWorker::Flush(tx)) => {
+                            if !inflight.is_empty() {
+                                inflight.join_next().await;
+                            }
+                            loop {
+                                if self.pending_batch.is_empty() && self.pending_msges.is_empty() {
+                                    break;
+                                }
+                                self.move_to_batch();
+                                self.pending_batch.flush(self.client.clone(), self.topic.clone(), &mut inflight);
```
suzmue, line 268: using flush for the timer seems slightly inefficient since if we have a max batch size of 1000 and we have a batch out and 1001 messages in the queue, then we we will send that single message as a batch of 1 even if more come in. I guess that is a possibility with having the timeout anyway, so we maybe its not a real problem and we don't need it to be changed necessarily.

This also makes we wonder actually if we have an outstanding batch, would it be better to just leave all messages in the channel? to avoid having to copy things around as much? I'm just curious about this. Something to maybe try out later.

suzmue (changes requested):

src/pubsub/src/publisher/worker.rs
PhongChuong: Moved to a helper message for now. It's likely to change when we implement the batch worker cleanup PR.

src/pubsub/src/publisher/worker.rs
PhongChuong: I added a TODO to handle the empty ordering key special case.

src/pubsub/src/publisher/worker.rs
PhongChuong: Noted. Lets discuss during our meeting tomorrow. It largely depends on how strict we want to respect the user configured batch size.
Note: In the worst case, we still need to batch if the batch size is greater than the backend limit.

PhongChuong:
@suzmue , thanks for the review.
In this implementation, we do not handle the special case of "" ordering key. I've added a TODO and handle that case in a later PR.
PTAL.

src/pubsub/src/publisher/worker.rs
```diff
@@ -238,6 +239,8 @@ impl BatchWorker {
         // While it is possible to use Some(JoinHandle) here as there is at max
         // a single inflight task at any given time, the use of JoinSet
         // simplify the managing the inflight JoinHandle.
+        // TODO(#4012): There is no inflight restriction when the ordering key is "".
```
suzmue, line 242: I am a little concerned about delaying this to a future PR since this is a performance regression, but since we are still in preview I think its alright if it improves dev velocity for this to be submitted.

PR #4126, opened 2025-12-25
docs(guide): fix broken LRO links in compute.md

gadekarnitesh:
docs(guide): fix broken LRO links in compute.md

Update links for "Working with long-running operations" and
"Configuring polling policies" to point to correct relative paths
in lro.md instead of non-existent absolute paths.


dbolduc (approved):
Thanks!

PR #4141, opened 2026-01-02
impl(storage): logic for signed urls

alvarowolfx:
Add core logic for generating signed URLs for GCS. Tested against GCS conformance tests.

Towards #3645

### DIFF
--- a/src/storage/src/storage/signed_url.rs
+++ b/src/storage/src/storage/signed_url.rs
@@ -81,11 +81,10 @@ impl SigningScope {
     }
 
     fn bucket_host(&self, host: &str, url_style: UrlStyle) -> String {
-        let bucket_name = self.bucket_name();
         match url_style {
             UrlStyle::PathStyle => host.to_string(),
             UrlStyle::BucketBoundHostname => host.to_string(),
-            UrlStyle::VirtualHostedStyle => format!("{bucket_name}.{host}"),
+            UrlStyle::VirtualHostedStyle => format!("{}.{host}", self.bucket_name()),
         }
     }
 
@@ -325,7 +324,7 @@ impl SignedUrlBuilder {
 
     /// Sets the endpoint for the signed URL. The default is "https://storage.googleapis.com".
     ///
-    /// Settings an endpoint takes precedence over using `with_universe_domain`.
+    /// Setting an endpoint takes precedence over using `with_universe_domain`.
     ///
     /// # Example
     ///


src/storage/src/storage/signed_url.rs
```diff
@@ -482,6 +616,24 @@ mod tests {
             .with_universe_domain("googleapis.com")
             .with_client_email("test@example.com")
             .with_url_style(UrlStyle::PathStyle)
+            .sign_with(&signer)
+            .await;
+
+        assert!(res.is_ok());
```
dbolduc: If this is an error, we do not get an informative log.

You can use `?`, as in: `let _ = blah_blah_blah().await?;`

src/storage/src/storage/signed_url.rs
```diff
@@ -718,7 +873,10 @@ mod tests {
         }
         println!("{}/{} tests passed", total_passed, total_tests);
 
-        // TODO: for now, all tests will fail
-        Ok(())
+        if failed {
+            Err(anyhow::anyhow!("Some tests failed"))
+        } else {
+            Ok(())
+        }
```
dbolduc, line 867:
```suggestion
        if failed {
            anyhow::bail!("Some tests failed")
        }
        Ok(())
```

seems slightly nicer to me

src/storage/src/storage/signed_url.rs
```diff
@@ -393,37 +394,170 @@ impl SignedUrlBuilder {
...
-            Some(e) if e.starts_with("https://") => e.clone(),
-            Some(e) => format!("https://{}", e),
-            None => format!("https://storage.{}", self.universe_domain),
+            Some(e) if e.starts_with("http://") => return e.clone(),
+            Some(e) if e.starts_with("https://") => return e.clone(),
+            Some(e) => return format!("https://{}", e),
+            None => {}
+        }
+
+        let emulator_host = std::env::var("STORAGE_EMULATOR_HOST").ok();
```
dbolduc: environment variables should override programmatic options.

src/storage/src/storage/signed_url.rs
```diff
@@ -393,37 +394,170 @@ impl SignedUrlBuilder {
...
+        match emulator_host {
+            Some(host) if host.starts_with("http://") => return host.clone(),
+            Some(host) if host.starts_with("https://") => return host.clone(),
+            Some(host) if !host.is_empty() => return format!("http://{}", host),
```
dbolduc: I am not a fan of handling empty as a special case of a bad endpoint. We do not do it if someone calls `.set_endpoint("")`, so why do it here?

dbolduc: genuine question: All of our clients have a `with_endpoint` function, and none of them add in the scheme. So why do it here? Is it necessary for the url builder thingy?

src/storage/src/storage/signed_url.rs
```diff
@@ -393,37 +394,170 @@ impl SignedUrlBuilder {
...
+        // the host and port pair should be maintained, even if is a known
+        // port like 80/443. The url crate omits the port if it is the default
+        // port for the scheme.
+        let path = url.path();
+        let scheme = format!("{}://", url.scheme());
+        let host = endpoint.trim_start_matches(&scheme).trim_end_matches(path);
+        Ok((url, host.to_string()))
     }
 
     fn resolve_endpoint(&self) -> String {
```
dbolduc, line 414: If we must add the scheme, consider something like this:

```rs
fn add_scheme(url: String) -> String {
  if url.starts_with("http://") || url.starts_with("https://") {
    return url;
  }
  format!("https://{url}")
}

fn resolve_endpoint(&self) -> String {
  std::env::var("STORAGE_EMULATOR_HOST").ok()
      .or_else(|| self.endpoint.clone())
      .map(Self::add_scheme)
      .unwrap_or_else(|| format!("https://storage.{}", self.universe_domain))
}
```

and if not, just:

```rs
fn resolve_endpoint(&self) -> String {
  std::env::var("STORAGE_EMULATOR_HOST").ok()
      .or_else(|| self.endpoint.clone())
      .unwrap_or_else(|| format!("https://storage.{}", self.universe_domain))
}
```

dbolduc: Aside, for another PR: seems like conflicting values can be supplied for the endpoint and universe_domain. Should we mention in the documentation that `with_endpoint` wins over `with_universe_domain`?

src/storage/src/storage/signed_url.rs
```diff
@@ -393,37 +394,170 @@ impl SignedUrlBuilder {
...
-        let host = url
-            .host_str()
-            .ok_or_else(|| SigningError::invalid_parameter("endpoint", "invalid endpoint host"))?;
-        Ok((endpoint_url, host.to_string()))
+        // the host and port pair should be maintained, even if is a known
+        // port like 80/443. The url crate omits the port if it is the default
+        // port for the scheme.
+        let path = url.path();
+        let scheme = format!("{}://", url.scheme());
+        let host = endpoint.trim_start_matches(&scheme).trim_end_matches(path);
```
dbolduc: comment: seems fishy, because we re-parse the URL below, and use `url::Url::host_str()` without a problem...

src/storage/src/storage/signed_url.rs
```diff
@@ -393,37 +394,170 @@ impl SignedUrlBuilder {
...
+            signer.client_email().await.map_err(SigningError::signing)?
+        };
+        let credential = format!("{client_email}/{credential_scope}");
+
         let (endpoint_url, host) = self.resolve_endpoint_url()?;
         let scheme = endpoint_url.scheme();
-        let _canonical_url = self
+        let canonical_url = self
             .scope
             .canonical_url(scheme, &host, self.url_style.clone());
```
dbolduc: `UrlStyle` should `derive(Copy)`, and we should not clone it here.

src/storage/src/storage/signed_url.rs
```diff
@@ -393,37 +394,170 @@ impl SignedUrlBuilder {
...
+            .map_err(|e| SigningError::invalid_parameter("endpoint", e))?;
+        let endpoint_host = endpoint_url
+            .host_str()
+            .ok_or_else(|| SigningError::invalid_parameter("endpoint", "invalid endpoint host"))?;
+
+        let mut headers = self.headers;
+        headers.insert("host".to_string(), endpoint_host.to_string());
+
+        let mut sorted_headers = headers.keys().collect::<Vec<_>>();
+        sorted_headers.sort_by_key(|k| k.to_lowercase());
```
dbolduc: Seems like we should transform the key to lowercase on insert, instead of during signing. I think this code gets simpler, and we don't need to re-sort (the `BTreeMap<>` handles that.)

src/storage/src/storage/signed_url.rs
```diff
@@ -393,37 +394,170 @@ impl SignedUrlBuilder {
...
+
+        let signed_headers = sorted_headers.iter().fold("".to_string(), |acc, k| {
+            format!("{acc}{};", k.to_lowercase())
+        });
+        let signed_headers = signed_headers.trim_end_matches(';').to_string();
```
dbolduc: `.join(";")` will save you a step

https://doc.rust-lang.org/std/slice/trait.Join.html

src/storage/src/storage/signed_url.rs
```diff
@@ -393,37 +394,170 @@ impl SignedUrlBuilder {
...
+
+        // If the user provides a value for X-Goog-Content-SHA256, we must use
+        // that value in the request string. If not, we use UNSIGNED-PAYLOAD.
+        let signature = "UNSIGNED-PAYLOAD".to_string();
+        let signature = headers.iter().fold(signature, |acc, (k, v)| {
+            if k.to_lowercase().eq("x-goog-content-sha256") {
+                return v.clone();
+            }
+            acc
+        });
```
dbolduc: This is more direct:

```rs
let signature = headers.get("x-goog-content-sha256")
    .cloned()
    .unwrap_or_else(|| "UNSIGNED-PAYLOAD".to_string());
```

src/storage/src/storage/signed_url.rs
```diff
@@ -393,37 +394,170 @@ impl SignedUrlBuilder {
...
+
+        let mut canonical_query = url::form_urlencoded::Serializer::new("".to_string());
+        let mut sorted_query_parameters_keys = query_parameters.keys().collect::<Vec<_>>();
+        sorted_query_parameters_keys.sort_by_key(|k| k.to_string());
+        sorted_query_parameters_keys.iter().for_each(|k| {
+            let value = query_parameters.get(k.as_str());
+            if value.is_none() {
+                return;
+            }
+            let value = value.unwrap();
```
dbolduc: I have no idea what this is trying to do.

What is wrong with saying:

```rs
for (k, v) in &query_parameters {
    canonical_query.append_pair(k, v);
}
```

src/storage/src/storage/signed_url.rs
```diff
@@ -393,37 +394,170 @@ impl SignedUrlBuilder {
...
+        let canonical_query = canonical_query
+            .replace("%7E", "~") // rollback to ~
+            .replace("+", "%20"); // missing %20 in +
+
+        let canonical_headers = sorted_headers.iter().fold("".to_string(), |acc, k| {
+            let header_value = headers.get(k.as_str());
+            if header_value.is_none() {
+                return acc;
+            }
+            let header_value = Self::canonicalize_header_value(header_value.unwrap());
```
dbolduc: Same. Iterating over a map, then looking up its elements is bizarre. Please clean this up.

src/storage/src/storage/signed_url.rs
```diff
@@ -718,7 +873,10 @@ mod tests {
         }
         println!("{}/{} tests passed", total_passed, total_tests);
 
-        // TODO: for now, all tests will fail
```
dbolduc, line 721: FYI for the future: all TODOs should have an issue number associated with them like `TODO(#XXXX)`

dbolduc (changes requested):
You need to convert the keys to lowercase on insert to the maps and revisit this logic.

And note that we are using `BTreeMap`s as the data structure, so we get sorting for free. You should be suspicious of any additional sorting. It probably means the code could be simpler.

src/storage/src/storage/signed_url.rs
alvarowolfx: you're right and that was the backing decision on using BTreeMap in the first place, to keep data sorted by the keys. The problem is that for generating the canonical request, we can't change the provide headers or query parameters (like lowercasing it). But the order of the keys needs to be done with the items lowercased. So we either keep two copies (one ordered and lowercases and the other with original data) or we sort when needed. 

I just tried to change the code to save it lowercased and it failed on conformance tests due to some of them checking that the canonical request needs to keep the case informed by the customer.

* https://github.com/googleapis/google-cloud-rust/blob/1830c777feb25a7d8197d7019ba3af1d01b11e36/src/storage/src/storage/conformance/v4_signatures.json#L187
* https://github.com/googleapis/google-cloud-rust/blob/1830c777feb25a7d8197d7019ba3af1d01b11e36/src/storage/src/storage/conformance/v4_signatures.json#L202

src/storage/src/storage/signed_url.rs
alvarowolfx: good catch, this was a leftover before I moved to using BTree to try to reduce the amount of sorts in the code.

src/storage/src/storage/signed_url.rs
alvarowolfx: because the header keys need to be sorted by their key name lowercased and appended in that order. If we just iterate on the `headers` is not gonna be sorted by their lowercased value. If I don't this way, it fails in this conformance test:

* https://github.com/googleapis/google-cloud-rust/blob/1830c777feb25a7d8197d7019ba3af1d01b11e36/src/storage/src/storage/conformance/v4_signatures.json#L218

src/storage/src/storage/signed_url.rs
alvarowolfx: but is not gonna handle the case where the `x-goog-content-sha256` is not lowercased, or am I missing something ? in the conformance test it's informed as `X-Goog-Content-SHA256` 

* https://github.com/googleapis/google-cloud-rust/blob/1830c777feb25a7d8197d7019ba3af1d01b11e36/src/storage/src/storage/conformance/v4_signatures.json#L233

src/storage/src/storage/signed_url.rs
alvarowolfx: I was adding the scheme manually before on previous PRs, but then I found this issue where the `url` crate omits known port numbers and the code got a bit confuse. I feel like I should go back and use `Url::parse` strictly where needed.

alvarowolfx: and good point on the usage of `with_endpoint` and storage emulator host, I'll improve docs.

src/storage/src/storage/signed_url.rs
alvarowolfx: Like I mention in another part related to this. I was previously on some PRs doing some manual string manipulation and then moved to the `url` crate, but then I found this issue where it omits known port numbers and the code got a bit confuse. I think I'll get rid of most code using `url` since for GCS signed url requirements are strict on keeping the explicit port if they were informed in the original endpoint and the `url` crate removes it.

src/storage/src/storage/signed_url.rs
alvarowolfx: good call

src/storage/src/storage/signed_url.rs
alvarowolfx: this empty handling is for when the env var `STORAGE_EMULATOR_HOST` is empty, which happens when is not even informed if I understood correctly since I used the `std::env::var("something").ok()` method.

alvarowolfx: > genuine question: All of our clients have a `with_endpoint` function, and none of them add in the scheme. So why do it here? Is it necessary for the url builder thingy?

the GCS signed url conformance test has a lot of entries where the scheme is missing and it requires to fill it up if missing.

src/storage/src/storage/signed_url.rs
alvarowolfx: the host headers keys needs to be lowercased. To use `join`, I'll have to create a new array with the lowercased keys, which doesn't make the code that much different.

```
let signed_headers = sorted_headers.iter().map(|k| {
    k.to_lowercase()
}).collect::<Vec<_>>();
let signed_headers = signed_headers.join(";");
 ```

File: src/storage/src/storage/signed_url.rs (hunk 5 above)
Comment (alvarowolfx) [reply src/storage/src/storage/signed_url.rs:414]: actually if customers use `with_endpoint`, it takes precedence over emulator options (even env var). This is also checked on conformance tests:

* Endpoint on client takes precedence over emulator
* Hostname takes precedence over endpoint and emulator

Comment (alvarowolfx) [reply src/storage/src/storage/signed_url.rs:414]: also, on the `add_scheme` method, the rules for adding scheme is different for emulator and endpoints. For endpoints it should default to `https` if missing, and for emulator it defaults to `http`.

Comment (coryan) [reply src/storage/src/storage/signed_url.rs:414]: > actually if customers use `with_endpoint`, it takes precedence over emulator options (even env var). This is also checked on conformance tests:
> 
> * Endpoint on client takes precedence over emulator
> * Hostname takes precedence over endpoint and emulator

This is a philosophical disagreement.  Some people believe that "the code" is the last thing our customers change and therefore is the last override and should win.  They are wrong.  Our customers also include the DevOps and SRE teams, which control "the environment" and that gets set *after* the code.  Environment settings should override the code.

We have not made a decision for the Rust SDK.  I am inclined to always use the environment as the last  override.

Comment (alvarowolfx) [reply src/storage/src/storage/signed_url.rs:414]: I agree with the philosophy of env vars taking precedence, but I think this is an special case for GCS and emulators, which is asserted by the conformance tests. If we don't open an exception here, we would have to skip related tests.

**Comment by alvarowolfx** [comment]
@dbolduc I managed to get rid of the confusing code around sorting and lowercasing headers. In the end the only part that needs to keep ordering and casing is for Query Params, so for headers we can lowercase on insert and keep sort of BTreeMap.

File: src/storage/src/storage/signed_url.rs (hunk 5 above)
Comment (coryan) [reply src/storage/src/storage/signed_url.rs:414]: > I agree with the philosophy of env vars taking precedence, but I think this is an special case for GCS and emulators,

You could convince me that we need to do something different here, if you can articulate why this case is special beyond "we need this to pass the tests".

Arguably `EMULATOR_HOST` should play no role in signed URLs. AFAIK we don't  support `EMULATOR_HOST` for the main client, so why use them for the signed URLs types?  And I am not sure how signed URLs would be used with the emulator.

> which is asserted by the conformance tests.

That is not an argument. That is (arguably) a bug in the conformance tests.

>  If we don't open an exception here, we would have to skip related tests.

I can live with skipping the tests.

Comment (alvarowolfx) [reply src/storage/src/storage/signed_url.rs:414]: > You could convince me that we need to do something different here, if you can articulate why this case is special beyond "we need this to pass the tests".

I'm searching the reasoning behind the Conformance Tests asserting that specifically, they should have a reason since two tests in the suite are related to that. I would argue that ignoring the expectation of a given conformance tests is to ignore the reasoning behind what the GCS team defined when writing such tests, I was treating them almost like requirements to the feature, but documented as tests.

> Arguably EMULATOR_HOST should play no role in signed URLs. AFAIK we don't support EMULATOR_HOST for the main client, so why use them for the signed URLs types?

Signed URLs is not dependent on the client, so I don't see why EMULATOR_HOST needs to be supported on both at the same time. This will just put the client behind on supporting it.

Comment (coryan) [reply src/storage/src/storage/signed_url.rs:414]: > Signed URLs is not dependent on the client, so I don't see why EMULATOR_HOST needs to be supported on both at the same time. This will just put the client behind on supporting it.

Suppose I set `EMULATOR_HOST` to something and then created a signed URL that uses the value.  What do I do with that signed URL? I cannot send it to the production service. I am not sure I can send it to the emulator either... or if I do, I am not sure it serves any purpose: the emulator cannot verify the signed URL is valid. It cannot check the hashes, it does not have access to the information needed to do so.

Comment (alvarowolfx) [reply src/storage/src/storage/signed_url.rs:414]: > Suppose I set `EMULATOR_HOST` to something and then created a signed URL that uses the value. What do I do with that signed URL? I cannot send it to the production service. I am not sure I can send it to the emulator either... or if I do, I am not sure it serves any purpose: the emulator cannot verify the signed URL is valid. It cannot check the hashes, it does not have access to the information needed to do so.

`EMULATOR_HOST` is not meant for production services. You can, for example, generated a signed URL (with `EMULATOR_HOST` set) to allow a frontend app running locally to upload a file to the emulator directly, the main client doesn't need to be involved. The emulator indeed doesn't validate signatures, but the overall workflow would be similar to how it would work in production.

 ```rust
let url = SignedUrlBuilder::for_object("my-bucket", "my-object.txt")
     .with_method(http::Method::PUT)
     .with_expiration(Duration::from_secs(3600))
     .with_header("content-type", "application/octet-stream")
     .sign_with(signer)
     .await?;
```

I've already used such feature in Node to test direct uploads from the frontend and have used the emulator for local development. Such use case can be seen here: https://levelup.gitconnected.com/uploading-images-from-react-frontend-to-google-cloud-storage-using-firebase-cloud-functions-ee1fd5079717

alvarowolfx: removed `EMULATOR_HOST` handling and skipped tests that inform it

src/storage/src/storage/signed_url.rs
```diff
@@ -325,6 +326,9 @@ impl SignedUrlBuilder {
 
     /// Sets the endpoint for the signed URL. The default is "https://storage.googleapis.com".
     ///
+    /// Settings an endpoint takes precedence over using `with_universe_domain` and
```
dbolduc: nit: s/Settings/Setting/

src/storage/src/storage/signed_url.rs
```diff
@@ -325,6 +326,9 @@ impl SignedUrlBuilder {
...
+    /// setting an emulator host via STORAGE_EMULATOR_HOST environment variable.
```
dbolduc: nit: remove mention of `STORAGE_EMULATOR_HOST` until it means something

src/storage/src/storage/signed_url.rs
```diff
@@ -298,7 +299,7 @@ impl SignedUrlBuilder {
     /// # }
     /// ```
     pub fn with_header<K: Into<String>, V: Into<String>>(mut self, key: K, value: V) -> Self {
-        self.headers.insert(key.into(), value.into());
+        self.headers.insert(key.into().to_lowercase(), value.into());
```
dbolduc, line 300: sooooooo much cleaner.

src/storage/src/storage/signed_url.rs
BenWhitehead: I very much agree with removing EMULATOR_HOST and skipping those conformance tests.

It's a trivial (single line?) for a customer to plumb an env var into `.with_endpoint` if they want/need that for their app.

IMO, the fact that rust has the signed url builder separate from the client makes things less ambiguous, fewer surprises (transparently trying to send 1k QPS to the IAM service has bitten multiple customers in multiple languages in the past -- explicitly providing the signer removes this surprise) and more simplified.

src/storage/src/storage/signed_url.rs
```diff
@@ -393,11 +397,13 @@ impl SignedUrlBuilder {
         let url = url::Url::parse(&endpoint)
             .map_err(|e| SigningError::invalid_parameter("endpoint", e))?;
 
-        let endpoint_url = url.clone();
-        let host = url
-            .host_str()
-            .ok_or_else(|| SigningError::invalid_parameter("endpoint", "invalid endpoint host"))?;
-        Ok((endpoint_url, host.to_string()))
+        // the host and port pair should be maintained, even if is a known
```
dbolduc: nit: s/the host/The host/

nit: s/even if is a/even if it is a/

src/storage/src/storage/signed_url.rs
```diff
@@ -80,21 +80,20 @@ impl SigningScope {
...
-            }
-            UrlStyle::BucketBoundHostname => {
-                format!("{scheme}://{host}")
-            }
-            UrlStyle::VirtualHostedStyle => {
-                format!("{scheme}://{bucket_name}.{host}")
-            }
+            UrlStyle::PathStyle => host.to_string(),
+            UrlStyle::BucketBoundHostname => host.to_string(),
+            UrlStyle::VirtualHostedStyle => format!("{bucket_name}.{host}"),
```
dbolduc: nit: we only need to create the `bucket_name` String in this branch.

`format!("{}.{host}", self.bucket_name())`

src/storage/src/storage/signed_url.rs
```diff
@@ -326,8 +325,7 @@ impl SignedUrlBuilder {
 
     /// Sets the endpoint for the signed URL. The default is "https://storage.googleapis.com".
     ///
-    /// Settings an endpoint takes precedence over using `with_universe_domain` and
-    /// setting an emulator host via STORAGE_EMULATOR_HOST environment variable.
+    /// Settings an endpoint takes precedence over using `with_universe_domain`.
```
dbolduc: nit: s/Settings/Setting/

alvarowolfx:
> This is much easier on the eyes. Thanks.

thanks for the help on getting this into a much better state. Just pushed some updates to address the final nits.

PR #4143, opened 2026-01-02
chore: run refresh-all with latest sidekick

ldetmer:
In order to prep for migration to Librarian we need to run sidekick refresh-all to get the latest changes made to sidekick [example](https://github.com/googleapis/librarian/pull/33120).

For https://github.com/googleapis/librarian/issues/3405

### DIFF
--- a/src/firestore/src/generated/gapic/client.rs
+++ b/src/firestore/src/generated/gapic/client.rs
@@ -126,6 +126,22 @@ impl Firestore {
     }
 
     /// Gets a single document.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_firestore::client::Firestore;
+    /// async fn sample(
+    ///    client: &Firestore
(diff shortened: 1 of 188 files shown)


dbolduc:
Argh, the `pr-regenerate` build is failing because we need to update the `.sidekick-version.txt` with the latest version:

```
GOPROXY=direct go list -m -u -f '{{.Version}}' github.com/googleapis/librarian@main >.sidekick-version.txt
```

But the `test-*` builds are failing because of Rust reasons. :unamused:

Give me a few mins to come up with a fix for those.

dbolduc:
> Give me a few mins to come up with a fix for those.

I just sent https://github.com/googleapis/librarian/pull/3426. I think we should merge that, then you will want to update the sidekick version and regenerate. There might not be any changes. (The changes in this PR are now gated by a codec option that is not enabled)

PR #4153, opened 2026-01-05
chore: gen all

ldetmer:
test librarian generate --all


dbolduc:
FYI @ldetmer, this will need to be migrated too:

https://github.com/googleapis/librarian/blob/aee33b7a411e891fb3d780229c9f3f508ce4deb7/internal/sidekick/sidekick/refresh.go#L82-L94

Basically this `rust_storage` configuration depends on other configurations. We need to load them somehow. We can continue to hardcode the paths (there is only one use case at the moment, it would be too early to generalize).

Can you look into it?

Issue #4085, opened 2025-12-17
Run 1B iterations of the "scenarios" test without errors.


coryan:
I am writing a new stress test for the client library. This test performs different scenarios for the new bidi streaming reads, including:

- Call `open_object()` and discard the result without making any reads.
- Call `open_object()` then call `read_range()` on the result and read all the data.
- Call `open_object()` then call `read_range()` read some data and then discard the reader and the descriptor.
- Call `open_object()` then call `read_range()` then drop the object descriptor and then read all the data.

The idea is to shake out problems (deadlocks, leaks, race conditions) by trying all these things concurrently, multiple times.  We should be able to complete about 1B operations without errors.

coryan:
I am getting several different errors:

```
bm-1765912602.txt:447,83741,14125855910,open_read_after_drop,334319711,,1wMOsOatnBm8Q60Oyt3Ong9bwaQ0LkTf,ERROR=the service reports an error with code UNAVAILABLE described as: 502:Bad Gateway;;Caused by:;    status: 'The service is currently unavailable'; self: "502:Bad Gateway"; metadata: {"content-type": "application/grpc"; "date": "Tue; 16 Dec 2025 23:18:14 GMT"; "content-length": "0"}
```

```
bm-1765912602.txt:261,87342,14126379567,open_read,360002144,,eMVtWLHCatx9b6ssdSKmLYmPwCBVUUFe,ERROR=the service reports an error with code UNKNOWN described as: 408:Request Timeout;;Caused by:;    status: 'Unknown error'; self: "408:Request Timeout"; metadata: {"content-type": "application/grpc"; "date": "Tue; 16 Dec 2025 23:18:41 GMT"; "content-length": "0"}
```

```
[...]

PR #4158, opened 2026-01-05
test(storage-scenarios): better policy logging

coryan:
Need to log the details of the retry loop status, including the time it has been running. That let me diagnose some `Unknown` problems as server-side timeouts reported with an unexpected code.

Motivated by #4085

### DIFF
--- a/src/storage/tests/scenarios/src/retry_policy.rs
+++ b/src/storage/tests/scenarios/src/retry_policy.rs
@@ -69,6 +69,9 @@ where
 
     fn remaining_time(&self, state: &RetryState) -> Option<Duration> {
         let result = self.inner.remaining_time(state);
+        // This function is called on every retry attempt, reduce the noise by
+        // just printing something if most (more than 75%) of the retry time
+        // limit has been used.
         if result.is_some_and(|d| d < state.start.elapsed() / 4) {
             tracing::info!("retry policy remaining time: {result:?}");
         }


src/storage/tests/scenarios/src/retry_policy.rs
```diff
@@ -54,18 +58,18 @@ where
...
+                tracing::info!("retry policy exhausted on throttle, state: {state:?}, error: {e:?}")
             }
         }
         result
     }
 
     fn remaining_time(&self, state: &RetryState) -> Option<Duration> {
         let result = self.inner.remaining_time(state);
-        if result.is_some_and(|d| d < Duration::from_secs(150)) {
+        if result.is_some_and(|d| d < state.start.elapsed() / 4) {
```
dbolduc, line 75: why divided by 4?

dbolduc (dismissed):

src/storage/tests/scenarios/src/retry_policy.rs
coryan: To reduce noise. It deserves a comment, done.

src/storage/tests/scenarios/src/retry_policy.rs
```diff
@@ -69,6 +69,9 @@ where
 
     fn remaining_time(&self, state: &RetryState) -> Option<Duration> {
         let result = self.inner.remaining_time(state);
+        // This function is called on every retry attempt, reduce the noise by
+        // just printing something if most (more than 75%) of the retry time
```
dbolduc, line 73: nit: 80%

dbolduc (approved):

src/storage/tests/scenarios/src/retry_policy.rs
coryan: $100 - 100 / 4 = 75$, no?

dbolduc: Yes, but that is not what we are calcuating.

```
progress = elapsed / total_time = elapsed / (elapsed + remaining_time)
```

subbing in `remaining_time * 4 = elapsed` yields 4/5 = 80%

(I am too lazy to deal with inequalities)

PR #4160, opened 2026-01-05
impl(storage): `open_object()` needs timeouts

coryan:
Without a per-attempt timeout the service sometimes returns a `Unknown` error after 300s. This is (a) too slow, and (b) an unrecoverable error. Setting a default timeout avoids both problems.

This is the key finding from #4085

### DIFF
--- a/src/storage/src/storage/open_object.rs
+++ b/src/storage/src/storage/open_object.rs
@@ -614,11 +614,7 @@ mod tests {
             .await
             .unwrap_err();
         assert!(err.is_timeout(), "{err:?}");
-        let elapsed = start.elapsed();
-        assert!(
-            elapsed >= target && elapsed < target + Duration::from_secs(5),
-            "elapsed={elapsed:?}"
-        );
+        assert_eq!(start.elapsed(), target);
 
         Ok(())
     }


src/storage/src/storage/bidi/connector.rs
```diff
@@ -91,10 +92,16 @@ where
...
-        let inner = async move |_| {
-            Self::connect_attempt(client.clone(), spec.clone(), ranges.clone(), &options).await
+        let inner = async move |d: Option<Duration>| {
+            let attempt_timeout = std::cmp::min(default_timeout, d.unwrap_or(default_timeout));
+            let attempt =
+                Self::connect_attempt(client.clone(), spec.clone(), ranges.clone(), &options);
+            match tokio::time::timeout(attempt_timeout, attempt).await {
+                Ok(r) => r,
+                Err(e) => Err(Error::timeout(e)),
+            }
```
dbolduc, line 104: optional: this looks like `foo.map_err(Error::timeout).flatten()`, if you prefer that.

src/storage/src/storage/open_object.rs
```diff
@@ -326,6 +327,31 @@ impl<S> OpenObject<S> {
...
+    ///     .send()
+    ///     .await?;
+    /// # Ok(()) }
+    /// ```
+    ///
+    /// The Cloud Storage client library times out `open_object()` attempts by
+    /// default (with a 60s timeout). Applications may want to set a different
+    /// value depending on how they are deployed.
+    ///
+    /// Note that the per-attempt timeout is subject to the overall
```
dbolduc: ...

src/storage/src/storage/open_object.rs
```diff
@@ -549,6 +582,48 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test(start_paused = true)]
+    async fn timeout() -> anyhow::Result<()> {
```
dbolduc, line 589: this test could cover more if it sets `with_attempt_timeout` and verifies how much time has elapsed when the result is yielded.

src/storage/src/storage/open_object.rs
```diff
@@ -549,6 +582,48 @@ mod tests {
...
+            .with_credentials(Anonymous::new().build())
+            .with_endpoint(endpoint.clone())
+            .with_retry_policy(NeverRetry)
+            .build()
+            .await?;
+
+        // This will timeout because we never send the initial message over `_tx`.
+        let response = client
+            .open_object("projects/_/buckets/test-bucket", "test-object")
+            .send();
```
dbolduc: Q: can we just `.await` the response? instead of the below

dbolduc (commented):
The documentation comment needs fixing, everything else is at your discretion

src/storage/src/storage/bidi/connector.rs
coryan: I do prefer that, but it was introduced in Rust 1.89... cannot use it yet.

src/storage/src/storage/open_object.rs
coryan: Fixed.

src/storage/src/storage/open_object.rs
coryan: Fixed.

src/storage/src/storage/open_object.rs
coryan: TIL: I think we need to simplify a lot of our other tests where I copied these ideas from.

src/storage/src/storage/open_object.rs
```diff
@@ -549,6 +585,44 @@ mod tests {
...
+            .with_attempt_timeout(target)
+            .send()
+            .await
+            .unwrap_err();
+        assert!(err.is_timeout(), "{err:?}");
+        let elapsed = start.elapsed();
+        assert!(
+            elapsed >= target && elapsed < target + Duration::from_secs(5),
+            "elapsed={elapsed:?}"
+        );
```
dbolduc: Is the delta necessary? If so, do you know where it comes from?

I expected `elapsed == target`

src/storage/src/storage/bidi/connector.rs
dbolduc: :shrug:

dbolduc (approved):

src/storage/src/storage/open_object.rs
coryan: I guess you are saying that as long as we use the tokio timer that is true. TIL x 2

coryan (commented):
PTAL

src/storage/src/storage/bidi/connector.rs
dbolduc: > but it was introduced in Rust 1.89

Ah. I must have confused it with `Option::flatten`

dbolduc (approved):

PR #4166, opened 2026-01-05
test(storage-scenarios): rampup tasks slowly

coryan:
Avoid transient errors by slowly ramping up the worker threads.

Motivated by #4085 where I had to run 1024 tasks at a time.

### DIFF
--- a/src/storage/tests/scenarios/src/dataset.rs
+++ b/src/storage/tests/scenarios/src/dataset.rs
@@ -12,8 +12,6 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use std::time::Duration;
-
 use super::args::Args;
 use super::names::random_object_name;
 use google_cloud_auth::credentials::Credentials;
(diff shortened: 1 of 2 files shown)


src/storage/tests/scenarios/src/dataset.rs
```diff
@@ -67,8 +73,16 @@ async fn existing(args: &Args, control: StorageControl) -> anyhow::Result<Vec<Ob
...
 
-async fn create(args: &Args, client: &Storage) -> anyhow::Result<Object> {
-    let data = bytes::Bytes::from_owner((32..128_u8).cycle().take(SIZE).collect::<Vec<_>>());
+async fn create(
+    task: usize,
+    data: bytes::Bytes,
+    args: &Args,
+    client: &Storage,
+) -> anyhow::Result<Object> {
+    tokio::time::sleep(Duration::from_millis(10) * task as u32).await;
```
dbolduc: s/Duration::from_millis(10)/args.rampup_period/ ?

dbolduc (dismissed):

src/storage/tests/scenarios/src/dataset.rs
coryan: Fixed by using a different argument.

dbolduc (approved):

PR #4171, opened 2026-01-06
impl(pubsub): publisher support multiple inflight batches for empty ordering key

PhongChuong:
Allow Publisher to send multiple inflight batches for the special case when ordering key is "".

### DIFF
--- a/src/pubsub/src/publisher/worker.rs
+++ b/src/pubsub/src/publisher/worker.rs
@@ -275,9 +275,8 @@ impl BatchWorker {
                                     self.move_to_batch();
                                     self.pending_batch.flush(self.client.clone(), self.topic.clone(), &mut inflight);
                                 }
-                                while !inflight.is_empty() {
-                                    inflight.join_next().await;
-                                }
+                                inflight.join_all().await;
+                                inflight = JoinSet::new();
                             } else {
                                 // Send batches sequentially.
                                 if !inflight.is_empty() {


src/pubsub/src/publisher/publisher.rs
```diff
@@ -867,6 +867,84 @@ mod tests {
...
+        let msg1_handle = publisher.publish(messages.get(0).unwrap().clone());
+        let msg2_handle = publisher.publish(messages.get(1).unwrap().clone());
+        assert_eq!(msg2_handle.await.expect("expected message id"), "hello 2");
+        assert_eq!(
+            start.elapsed(),
+            Duration::from_millis(0),
+            "the second batch of messages should have sent without any delay"
+        );
+        // Also validate the content of the first publish.
+        assert_eq!(msg1_handle.await.expect("expected message id"), "hello 1");
```
suzmue, line 945: not necessary, but could be worth checking the time at this point in the test.

This test also relies on the batches going out in order, which I think is fine, but that would mean this test will be still pass if we just mix up the batch ordering.

Possibly could add additional verification by adding a short delay between sends, and adding checks of the time in expect blocks. (first send occurs at 0, second send occurs at 1, result for second is at 1, result for second is at 11).

src/pubsub/src/publisher/worker.rs
```diff
@@ -255,18 +264,30 @@ impl BatchWorker {
...
-                                self.move_to_batch();
-                                self.pending_batch.flush(self.client.clone(), self.topic.clone(), &mut inflight);
-                                inflight.join_next().await;
+                            if unlimited_inflight {
+                                // Send all the batches concurrently.
+                                while !self.pending_batch.is_empty() || !self.pending_msgs.is_empty() {
+                                    self.move_to_batch();
+                                    self.pending_batch.flush(self.client.clone(), self.topic.clone(), &mut inflight);
+                                }
+                                while !inflight.is_empty() {
```
suzmue: This can be simplified to use join_all.

src/pubsub/src/publisher/worker.rs
```diff
@@ -239,12 +243,17 @@ impl BatchWorker {
         // While it is possible to use Some(JoinHandle) here as there is at max
         // a single inflight task at any given time, the use of JoinSet
         // simplify the managing the inflight JoinHandle.
-        // TODO(#4012): There is no inflight restriction when the ordering key is "".
-        // Add handling of this special case.
         let mut inflight = JoinSet::new();
+        // For empty ordering key, we do not need to limit the number of inflight batches.
+        let unlimited_inflight = self.ordering_key.is_empty();
         loop {
```
suzmue, line 249: I'd have to see it, but I think maybe just having two separate tokio::selects for the two cases may be easier to parse. So instead of having multiple `if unlimited inflight`, we would just have a single one and then choose the appropriate loop.

Could be a future refactor.

suzmue (dismissed):

src/pubsub/src/publisher/publisher.rs
PhongChuong: I think the issue is that because the test execution is sequential, so the ordering which we capture the elapsed time is critical in the validation. Let me think about this some more and see if there's a good way to validate this. One option is to perform the another test with a reversed order of await/validation but maybe there's a better solution.

src/pubsub/src/publisher/worker.rs

PR #4172, opened 2026-01-06
test(storage): add integration test for signed url

alvarowolfx:
Towards #3645

### DIFF
--- a/tests/integration/src/storage.rs
+++ b/tests/integration/src/storage.rs
@@ -115,9 +115,11 @@ pub async fn signed_urls(
 
     // Download the contents of the object using the signed URL.
     let client = reqwest::Client::new();
-    let res = client.get(signed_url)
+    let res = client
+        .get(signed_url)
         .header("Content-Type", "text/plain")
-        .send().await?;
+        .send()
+        .await?;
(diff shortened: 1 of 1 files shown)


tests/integration/tests/driver.rs
```diff
@@ -201,6 +201,24 @@ mod driver {
         result
     }
 
+    #[cfg(all(test, google_cloud_unstable_signed_url))]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
+    async fn run_storage_signed_urls() -> integration_tests::Result<()> {
+        let _guard = integration_tests::enable_tracing();
+        let (control, bucket) = integration_tests::storage::create_test_bucket().await?;
```
coryan, line 208: nit: creating buckets is not free (in time / flakiness), can we fold this test into another existing test?

alvarowolfx: will check a bit before merging this if I can add it to another test

alvarowolfx:
will merge as is for now @coryan, I was making some changes to make it part of the `run_storage_objects` integration test, but is starting to diverge significantly from the original PR.

Issue #4170, opened 2026-01-06
Provide feature flags to control the crypto provider


coryan:
I am moving this discussion to a separate issue, it was getting off-topic.

> You can enable aws-lc-rs in your own Cargo.toml file. Then you can (if you are writing an application) change the default provider using [CryptoProvider::install_default()](https://docs.rs/rustls/latest/rustls/crypto/struct.CryptoProvider.html#method.install_default). Does that work for you?
> 
> This will then exist in the entire dependency tree.
> 
>  

 _Originally posted by @houseme in [#3947](https://github.com/googleapis/google-cloud-rust/issues/3947#issuecomment-3715072845)_

Yes, can you help me understand why is that a problem?  Is it the code size?

houseme:
Hi @coryan,


Three possible scenarios:


1. **Package size issue**: After introducing `aws-lc-rs`, redundant dependencies of the `ring` crate may cause the compiled package size to increase.

2. **Dependency conflict issue**: Although the possibility is low, it still needs to be considered for potential conflicts with other dependency libraries.

3. **Auditing Issue**: When using the `ring` crate, it may not comply with certain security standards.


Thank you!

houseme:
This can be achieved through feature flags, with the `ring` feature enabled by default, and `aws-lc-rs` as optional, which allows excluding the simultaneous introduction of both the `aws-lc-rs` and `ring` crates. The changes required are:

https://github.com/googleapis/google-cloud-rust/blob/42aadd2253d5d0e57006d890b1f190221627ffaa/src/auth/src/credentials/service_account.rs#L376

houseme:
@coryan 
I would like to apply for the implementation of this modification, using feature flags to independently control ring and aws-lc-rs. This will allow users to choose to enable one of them or both as needed, thereby addressing potential package size, dependency conflicts, and audit issues. Please let me know if I can start working.

coryan:
FWIW, we are shooting for all these changes to go in the next release, maybe as early as 2026-01-28. In general, the approach will be:

- With the default features one gets a default crypto provider: `aws-lc-rs`.
- With the default features disabled one does not get a default crypto provider **and**:
   - The crypto providers are pruned from the dependency tree
   - The application must call `rustls::CryptoProvider::install_default()`

coryan:
@Dig-Doug I think this will affect you (based on your request in #1077). We simply cannot find another solution that works for all customers and allows us to keep our dependencies up to date.

[...]

PR #4174, opened 2026-01-06
feat: make CryptoProvider selectable via crate features

houseme:
feat: make CryptoProvider selectable via crate features

- Add two new optional features:
  - `rustls-ring` – enables `rustls/ring`, re-exports `rustls::crypto::ring::default_provider`
  - `rustls-aws-lc-rs` – enables `rustls/aws-lc-rs`, re-exports `rustls::crypto::aws_lc_rs::default_provider`
- Default feature set is `["rustls-ring"]`, keeping the current behaviour unchanged.
- The two features are mutually exclusive; enabling both raises a clear
  compile-time error.
- README updated to show how to switch provider:
    cargo add <crate> --no-default-features --features rustls-aws-lc-rs

Downstream crates can now pick their rustls crypto provider without
patching or forking.

#4170

### DIFF
--- a/src/storage/Cargo.toml
+++ b/src/storage/Cargo.toml
@@ -62,7 +62,7 @@ mime.workspace = true
 # Local crates
 google-cloud-auth     = { workspace = true, optional = true }
 gax.workspace         = true
-gaxi                  = { workspace = true, features = ["_internal-common", "_internal-grpc-client"] }
+gaxi                  = { workspace = true, optional = true }
 gtype.workspace       = true
 iam_v1.workspace      = true
 longrunning.workspace = true
@@ -72,8 +72,8 @@ wkt.workspace         = true
 
 [features]
 default = ["rustls-ring"]
-rustls-aws-lc-rs = ["dep:google-cloud-auth","google-cloud-auth?/rustls-aws-lc-rs", "dep:reqwest", "reqwest?/rustls"]
-rustls-ring = ["dep:google-cloud-auth","google-cloud-auth?/rustls-ring", "dep:reqwest", "reqwest?/rustls-no-provider"]
+rustls-aws-lc-rs = ["dep:google-cloud-auth","google-cloud-auth?/rustls-aws-lc-rs", "dep:reqwest", "reqwest?/rustls", "dep:gaxi","gaxi?/_internal-grpc-client-aws"]
+rustls-ring = ["dep:google-cloud-auth","google-cloud-auth?/rustls-ring", "dep:reqwest", "reqwest?/rustls-no-provider", "dep:gaxi","gaxi?/_internal-grpc-client"]
 unstable-stream = ["reqwest/stream"]
 
 [dev-dependencies]


coryan:
> 
> Downstream crates can now pick their rustls crypto provider without patching or forking.

How does this work for something like `google-cloud-storage` or `google-cloud-secretmanager-v1`?  If those crates use the default features of `google-cloud-auth` how does one select the `aws-lc-rs` feature? If they don't select the default features for `google-cloud-auth`, how do they work for by default?

houseme:
> > Downstream crates can now pick their rustls crypto provider without patching or forking.
> 
> How does this work for something like `google-cloud-storage` or `google-cloud-secretmanager-v1`? If those crates use the default features of `google-cloud-auth` how does one select the `aws-lc-rs` feature? If they don't select the default features for `google-cloud-auth`, how do they work for by default?

Thank you for your guidance in case you are ignored and left unprocessed. In addition to the above three crates, which other related crates need to be processed together. Thank you!

coryan:
> > > Downstream crates can now pick their rustls crypto provider without patching or forking.
> > 
> > 
> > How does this work for something like `google-cloud-storage` or `google-cloud-secretmanager-v1`? If those crates use the default features of `google-cloud-auth` how does one select the `aws-lc-rs` feature? If they don't select the default features for `google-cloud-auth`, how do they work for by default?
> 
> Thank you for your guidance in case you are ignored and left unprocessed. In addition to the above three crates, which other related crates need to be processed together.

Basically every crate in this repo. There are about 200 of them.  And most of them are generated.  Even the `Cargo.toml` file is generated.

This is why we have hesitated to just add some features. We need to design the features so they work for all crates, even crates that depend on each other. For example: `google-cloud-kms-inventory-v1` depends on `google-cloud-kms-v1` and many crates depend on `google-cloud-iam-v1` , `google-cloud-location` and/or `google-cloud-longrunning`.

We like what you did with the default provider in this PR, but we need to design something that works across all crates, including the changes in the code generator.

houseme:
@coryan Upgrade reqwest to 0.13.x, okay?

houseme:
> > > > Downstream crates can now pick their rustls crypto provider without patching or forking.
> > > 
> > > 
> > > How does this work for something like `google-cloud-storage` or `google-cloud-secretmanager-v1`? If those crates use the default features of `google-cloud-auth` how does one select the `aws-lc-rs` feature? If they don't select the default features for `google-cloud-auth`, how do they work for by default?
> > 
> > 
> > Thank you for your guidance in case you are ignored and left unprocessed. In addition to the above three crates, which other related crates need to be processed together.
> 
> Basically every crate in this repo. There are about 200 of them. And most of them are generated. Even the `Cargo.toml` file is generated.
> 
> This is why we have hesitated to just add some features. We need to design the features so they work for all crates, even crates that depend on each other. For example: `google-cloud-kms-inventory-v1` depends on `google-cloud-kms-v1` and many crates depend on `google-cloud-iam-v1` , `google-cloud-location` and/or `google-cloud-longrunning`.
> 
> We like what you did with the default provider in this PR, but we need to design something that works across all crates, including the changes in the code generator.

@coryan After query, only 'google-cloud-storage' crate relies on 'google-cloud-auth' crate.

coryan:
> > Basically every crate in this repo. There are about 200 of them. And most of them are generated. Even the `Cargo.toml` file is generated.
> > This is why we have hesitated to just add some features. We need to design the features so they work for all crates, even crates that depend on each other. For example: `google-cloud-kms-inventory-v1` depends on `google-cloud-kms-v1` and many crates depend on `google-cloud-iam-v1` , `google-cloud-location` and/or `google-cloud-longrunning`.
> > We like what you did with the default provider in this PR, but we need to design something that works across all crates, including the changes in the code generator.
> 
> @coryan After query, only 'google-cloud-storage' crate relies on 'google-cloud-auth' crate.

But they all depend on `google-cloud-gax-internal` and that depends on `google-cloud-auth`.

coryan:
I want to thank you for the PR, but I think we will use a different approach.   I think we will just use a feature (enabled by default) to use the default provider in `reqwest` or we will disable all default crypto providers.

Please give me a couple of days and I will ping you once my proposal is ready for review.

coryan:
FWIW, I think #4182 is closer to what we need.  That is still missing the changes to the generated crates, and more tests, but seems (to me) easier to maintain and test.

houseme:
> FWIW, I think #4182 is closer to what we need. That is still missing the changes to the generated crates, and more tests, but seems (to me) easier to maintain and test.

Support users can choose to enable different engines. As well as the current 'google-cloud-storage' relies on 'google-cloud-auth' and 'gaxi', we need to consider the enabling of 'google-cloud-storage' relying on 'features'. It would be best if 'aws-lc-rs' is finally enabled by default. thank

PR #4176, opened 2026-01-06
impl(storage): atomic `open_object()` + read

### LINKED ISSUE #3996: Support an immediate read as part of `open_object()`
It is possible to both open an object and then read some data.  Maybe do something like:

```
// Open and read the trailer...
let (descriptor, reader) = client.open_object("...bucket", "my-object.parquet").send_with_read(ReadRange::last(32)).await?;

// ^^^ that has lower latency than the following:
let descriptor = client.open_object("...bucket", "my-object.parquet").send().await?;
let reader = descriptor.read_range(ReadRange::last(32)).await;
```

I think we can support this in the client and request builders "easily", but it may affect the mocks and therefore we should at least design for it before releasing this thing.

coryan:
Support for performing a read and open in a single RPC.

Fixes #3996

### DIFF
--- a/src/storage/src/storage/open_object.rs
+++ b/src/storage/src/storage/open_object.rs
@@ -86,7 +86,7 @@ where
     ///
     /// This method allows applications to open an object and issue a read
     /// request in the same RPC, which is typically faster than opening an
-    /// object and then issuing a `read_range()` cal. This may be useful when
+    /// object and then issuing a `read_range()` call. This may be useful when
     /// opening objects that have metadata information in a footer or header.
     pub async fn send_and_read(
         mut self,


src/storage/src/storage/open_object.rs
```diff
@@ -66,6 +67,39 @@ where
...
+    ///     .open_object("projects/_/buckets/my-bucket", "my-object.parquet")
+    ///     .send_and_read(ReadRange::tail(32))
+    ///     .await?;
+    /// println!("object metadata={:?}", descriptor.object());
+    /// let data = reader.next().await.transpose()?;
+    /// # Ok(()) }
+    /// ```
+    ///
+    /// This method allows applications to open an object and issue a read
+    /// request in the same RPC. This may be useful when opening objects
```
dbolduc: optional: explicitly say that "in the same RPC" is faster for those in the back of the class?

"... in the same RPC, which is faster than opening an object then reading from it."

src/storage/src/storage/open_object.rs
```diff
@@ -585,6 +622,69 @@ mod tests {
...
+                }),
+                range_end: true,
+                checksummed_data: Some(ChecksummedData {
+                    content: payload.clone(),
+                    crc32c: None,
+                }),
+            }],
+            ..BidiReadObjectResponse::default()
+        };
+        tx.send(Ok(initial.clone())).await?;
```
dbolduc, line 652: optional: could also test error case

src/storage/src/storage/open_object.rs
```diff
@@ -585,6 +622,69 @@ mod tests {
...
+        let (tx, rx) = tokio::sync::mpsc::channel::<tonic::Result<BidiReadObjectResponse>>(1);
+        let payload = Vec::from_iter((0..32).map(|i| i as u8));
+        let initial = BidiReadObjectResponse {
+            metadata: Some(ProtoObject {
+                bucket: "projects/_/buckets/test-bucket".to_string(),
+                name: "test-object".to_string(),
+                generation: 123456,
+                ..ProtoObject::default()
+            }),
+            object_data_ranges: vec![ObjectRangeData {
```
dbolduc, line 639: Q: just curious, is this ever empty? and you have to keep reading to get the content.

If so, I assume you already have test coverage of this case. No need to dup the tests here.

src/storage/tests/scenarios/src/sample/run.rs
```diff
@@ -111,6 +111,36 @@ pub async fn open_read(client: &Storage, objects: &[Object]) -> Attempt {
...
+        .expect("at least one object");
+    let start = Instant::now();
+    let result = client
+        .open_object(&object.bucket, &object.name)
+        .set_generation(object.generation)
+        .send_and_read(ReadRange::head(4 * MIB as u64))
+        .await;
+    match result {
+        Ok((d, reader)) => {
+            let open_latency = start.elapsed();
```
dbolduc, line 126: just curious: is `open_latency` a bit higher for this case because of the initial read?

tests/integration/src/storage/bidi_read.rs
```diff
@@ -48,3 +54,40 @@ pub async fn run(bucket_name: &str) -> anyhow::Result<()> {
...
+    assert_eq!(got, want);
+
+    let mut count = 0_usize;
+    let mut data = Vec::new();
+    while let Some(r) = reader.next().await.transpose()? {
+        tracing::info!("received {} bytes", r.len());
+        count += r.len();
+        data.extend_from_slice(&r);
+    }
+    assert_eq!(count, 100_usize);
```
dbolduc: nit: I'd probably delete the `count` stuff if we are verifying the full payload.

dbolduc (approved):
Nice

src/storage/src/storage/open_object.rs
coryan: Done.

src/storage/src/storage/open_object.rs
coryan: Yes, I believe the existing tests cover this.  Any response may include a subset of the ranges.

src/storage/tests/scenarios/src/sample/run.rs
coryan: I suppose it depends on how much you read. Every MiB is costs you about 4ms, the maximum size of a response is 2MiB, so it could add up to 8ms for reads over 2MiB.  In the use-cases I expect to happen the most (headers and footers) this would be adding a few KiB (so maybe a few µsecs?).

tests/integration/src/storage/bidi_read.rs
coryan: Done.

src/storage/src/storage/open_object.rs
```diff
@@ -66,6 +67,40 @@ where
...
+    ///     .send_and_read(ReadRange::tail(32))
+    ///     .await?;
+    /// println!("object metadata={:?}", descriptor.object());
+    /// let data = reader.next().await.transpose()?;
+    /// # Ok(()) }
+    /// ```
+    ///
+    /// This method allows applications to open an object and issue a read
+    /// request in the same RPC, which is typically faster than opening an
+    /// object and then issuing a `read_range()` cal. This may be useful when
```
dbolduc: s/cal./call./

dbolduc (dismissed):

src/storage/src/storage/open_object.rs
coryan: Fixed, thanks.

PR #4177, opened 2026-01-07
impl(pubsub): introduce `Session`

dbolduc:
Part of the work for #3957

Introduce a `Session`, which will be a public-facing type. Applications interact with this to stream messages.

The streamed item is a `Result<(Message, Handler)>`. Applications are forced to handle stream errors, if any.

As of 2026-01-06, this is not reflected in go/cloud-rust:subscriber-api. I need to do an update on that thing. I want to get something, anything working first.

I tried to keep this PR minimal. That means it is missing things like:
1. keepalives
1. lease management
1. graceful shutdown
1. stream retries
1. exactly once messages
1. etc.

I have plans for 1-3. I am sure 4-6 can be done.

### DIFF
--- a/src/pubsub/src/subscriber/session.rs
+++ b/src/pubsub/src/subscriber/session.rs
@@ -164,7 +164,7 @@ mod tests {
     use std::sync::Arc;
     use tokio::sync::mpsc::channel;
     use tokio::task::JoinHandle;
-    use tokio::time::{Duration, Instant};
+    use tokio::time::Duration;
 
     fn test_data(v: i32) -> bytes::Bytes {
         bytes::Bytes::from(format!("data-{}", test_id(v)))
@@ -293,8 +293,12 @@ mod tests {
         // In this test, we verify the case where an application asks for a
         // message, but a response is not immediately available on the stream.
 
-        let start_t = Instant::now();
         let (response_tx, response_rx) = channel(10);
+        let handle: JoinHandle<anyhow::Result<()>> = tokio::spawn(async move {
+            tokio::time::sleep(Duration::from_millis(20)).await;
+            response_tx.send(Ok(test_response(1..2))).await?;
+            Ok(())
+        });
 
         let mut mock = MockSubscriber::new();
         mock.expect_streaming_pull()
@@ -306,13 +310,6 @@ mod tests {
             "projects/p/subscriptions/s".to_string(),
         );
         let mut session = Session::new(builder).await?;
-
-        let handle: JoinHandle<anyhow::Result<()>> = tokio::spawn(async move {
-            tokio::time::sleep(Duration::from_millis(20)).await;
-            response_tx.send(Ok(test_response(1..2))).await?;
-            Ok(())
-        });
-
         let (m, Handler::AtLeastOnce(h)) = session
             .next()
             .await
@@ -320,7 +317,6 @@ mod tests {
             .expect("stream should wait for a message");
         assert_eq!(m.data, test_data(1));
         assert_eq!(h.ack_id, test_id(1));
-        assert_eq!(start_t.elapsed(), Duration::from_millis(20));
 
         handle.await??;
 


src/pubsub/src/subscriber/session.rs
```diff
@@ -0,0 +1,384 @@
...
+    /// (ack) or rejecting (nack) the message.
+    ///
+    /// If the underlying stream encounters a permanent error, an `Error` is
+    /// returned instead.
+    ///
+    /// `None` represents the end of a stream, typically triggered by a
+    /// `close()`, or an unrecoverable internal error.
+    ///
+    /// # Example
+    /// ```no_rust
```
suzmue, line 103: Love this example. Maybe we can duplicate it on the high level session object or add something similar?

src/pubsub/src/subscriber/session.rs
```diff
@@ -0,0 +1,384 @@
...
+use tokio::sync::mpsc::{Sender, UnboundedSender, unbounded_channel};
+use tokio_util::sync::CancellationToken;
+
+/// Represents an open subscribe session.
+///
+/// This is a stream-like struct for serving messages to an application.
+#[derive(Debug)]
+pub struct Session {
+    stream: <Transport as Stub>::Stream,
+    pool: VecDeque<(PubsubMessage, Handler)>,
```
suzmue, line 55: Can you add a comment for developers for pool and why we have it?

src/pubsub/src/subscriber/session.rs
```diff
@@ -0,0 +1,384 @@
...
+    pub(crate) async fn stream_next(&mut self) -> Option<Result<()>> {
+        let resp = match self.stream.next_message().await.transpose()? {
+            Ok(resp) => resp,
+            Err(e) => return Some(Err(to_gax_error(e))),
+        };
+        for rm in resp.received_messages {
+            // TODO(#3957) - add message to lease management
+            let pb = match rm
+                .message
+                .expect("The message field is always present.")
```
suzmue: Since these are coming from the server and we can't verify that this true for sure, I think we should consider handling the error where this isn't included. Worst case it never gets returned, but best case we don't kill the stream if there is a single broken message returned by the server.

src/pubsub/src/subscriber/session.rs
```diff
@@ -0,0 +1,384 @@
...
+
+        let (endpoint, _server) = start("0.0.0.0:0", mock).await?;
+        let transport = test_transport(endpoint).await?;
+        let builder = StreamingPull::new(
+            Arc::new(transport),
+            "projects/p/subscriptions/s".to_string(),
+        )
+        .set_ack_deadline_seconds(20)
+        .set_max_outstanding_messages(2000)
+        .set_max_outstanding_bytes(200 * MIB);
```
suzmue, line 244: I didn't know this is how it would get formatted when you have the parens ending left aligned. Interesting...

src/pubsub/src/subscriber/session.rs
```diff
@@ -0,0 +1,384 @@
...
+            assert_eq!(h.ack_id, test_id(i));
+        }
+        let end = session.next().await.transpose()?;
+        assert!(end.is_none(), "Received extra message: {end:?}");
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn delayed_responses() -> anyhow::Result<()> {
```
suzmue, line 292: Is this test expected to behave differently? Is there something this test is likely to catch that basic_success will not?

suzmue (approved):

src/pubsub/src/subscriber/session.rs
dbolduc: Good idea. I documented the fields that aren't going to go away in the next PR or two.

src/pubsub/src/subscriber/session.rs
dbolduc: > Love this example.

Ay.

> Maybe we can duplicate it on the high level session object or add something similar?

Yeah... I agree. Moved this example to the struct docs, and assumed a session in this example.

FYI: I am taking a TODO on going through the docs. I will need to do a pass to remove all the `no_rust` stuff. I added a line item to #3941 to make sure we have examples for everything.

src/pubsub/src/subscriber/session.rs
dbolduc: Done. I ignored the `ReceivedMessage` instead of error-ing. Added a `handles_missing_message_field` test.

src/pubsub/src/subscriber/session.rs
dbolduc: > Is this test expected to behave differently?

No.

> Is there something this test is likely to catch that basic_success will not?

In all the other tests I seeded the channel with responses. If we accidentally wrote the code with `try_recv()` instead of `recv().await`, none of the other tests would notice.

It could be simpler though. I made it simpler.

---

I actually also wanted to verify we do not drain the full stream (which ends in our tests, but not in practice) before serving the first message to an application.

But the test would not have caught that. D'oh.

So I added a `serves_messages_immediately` test.

suzmue (approved):

PR #4178, opened 2026-01-07
docs(storage): improve signed url docs

alvarowolfx:
Towards #3645

### DIFF
--- a/src/storage/src/storage/signed_url.rs
+++ b/src/storage/src/storage/signed_url.rs
@@ -39,7 +39,7 @@ const PATH_ENCODE_SET: AsciiSet = ENCODED_CHARS.remove(b'/');
 ///
 /// This is the recommended way for most applications. It automatically finds credentials from the environment.
 ///
-/// ```rust,no_run
+/// ```
 /// use auth::credentials::Builder;
 /// use auth::signer::Signer;
 ///
@@ -55,7 +55,7 @@ const PATH_ENCODE_SET: AsciiSet = ENCODED_CHARS.remove(b'/');
 /// Service account based signers work by local signing and do not make network requests, which can be
 /// useful in environments where network access is restricted and performance is critical.
 ///
-/// ```rust,no_run
+/// ```
 /// use auth::credentials::service_account::Builder;
 /// use auth::signer::Signer;
 ///


src/storage/src/storage/signed_url.rs
```diff
@@ -29,7 +29,47 @@ const PATH_ENCODE_SET: AsciiSet = ENCODED_CHARS.remove(b'/');
...
+///
+/// ## Using Application Default Credentials (ADC)
+///
+/// This is the recommended way for most applications. It automatically finds credentials from the environment.
+///
+/// ```rust,no_run
+/// use auth::credentials::Builder;
+/// use auth::signer::Signer;
+///
+/// # fn build_signer() -> Result<Signer, Box<dyn std::error::Error>> {
```
coryan: nit:

```suggestion
/// # fn build_signer() -> anyhow::Result<()> {
```

src/storage/src/storage/signed_url.rs
```diff
@@ -29,7 +29,47 @@ const PATH_ENCODE_SET: AsciiSet = ENCODED_CHARS.remove(b'/');
...
+/// let signer = Builder::default().build_signer()?;
+/// # Ok(signer)
```
coryan: nit:

```suggestion
/// # Ok(())
```

src/storage/src/storage/signed_url.rs
```diff
@@ -29,7 +29,47 @@ const PATH_ENCODE_SET: AsciiSet = ENCODED_CHARS.remove(b'/');
...
+///
+/// You can use `google-cloud-auth` to create a `Signer`.
+///
+/// ## Using Application Default Credentials (ADC)
+///
+/// This is the recommended way for most applications. It automatically finds credentials from the environment.
+///
+/// ```rust,no_run
+/// use auth::credentials::Builder;
+/// use auth::signer::Signer;
```
coryan, line 44: No action needed: can we use `google_cloud_auth` instead of `auth`?  The customers may not know about our convenience aliases.

src/storage/src/storage/signed_url.rs
```diff
@@ -29,7 +29,47 @@ const PATH_ENCODE_SET: AsciiSet = ENCODED_CHARS.remove(b'/');
...
+/// # }
+/// ```
+///
+/// ## Using a Service Account Key File
+///
+/// This is useful when you have a specific service account key file (JSON) and want to use it directly.
+/// Service account based signers work by local signing and do not make network requests, which can be
+/// useful in environments where network access is restricted and performance is critical.
+///
+/// ```rust,no_run
```
coryan: Same nits, and I think you can omit the `,no_run` in both examples.

src/storage/src/storage/signed_url.rs
```diff
@@ -47,7 +87,29 @@ const PATH_ENCODE_SET: AsciiSet = ENCODED_CHARS.remove(b'/');
...
 /// ```
+///
+/// ## Generating a Signed URL for Uploading an Object (PUT)
+///
+/// ```no_run
+/// use google_cloud_storage::builder::storage::SignedUrlBuilder;
+/// use std::time::Duration;
+/// # use auth::signer::Signer;
+///
+/// # async fn run(signer: &Signer) -> Result<(), Box<dyn std::error::Error>> {
```
coryan: Similar nits.

src/storage/src/storage/signed_url.rs
coryan: Maybe use a more useful example such as `private.googleapis.com`.

src/storage/src/storage/signed_url.rs
```diff
@@ -324,6 +379,8 @@ impl SignedUrlBuilder {
 
     /// Sets the endpoint for the signed URL. The default is "https://storage.googleapis.com".
     ///
+    /// This is useful for testing (e.g. using the emulator) or when using a custom domain.
```
coryan: Maybe revert this, I believe customers may use a custom domain:


```suggestion
    /// This is useful when using a custom domain, or when testing with some Cloud Storage emulators.
```

coryan (dismissed):
Only nits.

src/storage/src/storage/signed_url.rs
alvarowolfx: I feel like would be nice to bring the Signer examples here from auth, but we can't reference `google_cloud_auth` as is, since we use an alias in the crate. I tried adding `google_cloud_auth` again without the alias, but of course `cargo` doesn't allow duplicate entries. Do you think a comment saying that `auth` is an alias to `google_cloud_auth` makes things better ? Not particularly happy about that 

```rust
/// // auth crate is an alias for google-cloud-auth
/// use auth::credentials::Builder;
/// use auth::signer::Signer;
```

coryan: Does this work:

```rust
/// # use auth as google_cloud_auth;
/// use google_cloud_auth::credentials::Builder;
```

??

coryan (approved):

src/storage/src/storage/signed_url.rs
alvarowolfx: Sorry, missed this comment. It doesn't work: 
```
error[E0433]: failed to resolve: use of unresolved module or unlinked crate `google_cloud_auth`
 --> src/storage/src/storage/signed_url.rs:44:5
  |
4 | use google_cloud_auth::signer::Signer;
  |     ^^^^^^^^^^^^^^^^^ use of unresolved module or unlinked crate `google_cloud_auth`
  |
  = help: if you wanted to use a crate named `google_cloud_auth`, use `cargo add google_cloud_auth` to add it to your `Cargo.toml`
```

PR #4180, opened 2026-01-07
docs(auth): improve signer docs

alvarowolfx:
Towards #3645

### DIFF
--- a/src/auth/src/signer.rs
+++ b/src/auth/src/signer.rs
@@ -14,8 +14,9 @@
 
 //! Abstraction for signing arbitrary bytes using Google Cloud [Credentials][crate::credentials::Credentials].
 //!
-//! A [`Signer`] is used to sign data, typically for authentication or authorization purposes.
-//! One primary use case in the Google Cloud ecosystem is generating [Signed URLs] for Google Cloud Storage.
+//! A [`Signer`] is used to sign data, typically for authentication or
+//! authorization purposes. One primary use case in the Google Cloud
+//! ecosystem is generating [Signed URLs] for Google Cloud Storage.
 //!
 //! The main type in this module is [Signer]. This is an opaque type
 //! that implements the [SigningProvider] trait and can be used to
@@ -24,12 +25,13 @@
 //!
 //! ## Example: Creating a Signer using Application Default Credentials (ADC)
 //!
-//! This is the recommended way for most applications. It automatically finds credentials from the environment.
+//! This is the recommended way for most applications. It automatically finds
+//! credentials from the environment.
 //!
 //! ```
 //! use google_cloud_auth::credentials::Builder;
 //! use google_cloud_auth::signer::Signer;
-//! 
+//!
 //! # tokio_test::block_on(async {
 //! let signer: Signer = Builder::default().build_signer()?;
 //! # Ok::<(), anyhow::Error>(())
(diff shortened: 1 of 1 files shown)


src/auth/src/signer.rs
```diff
@@ -12,6 +12,62 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Abstraction for signing arbitrary bytes using Google Cloud [Credentials][crate::credentials::Credentials].
+//!
+//! A [`Signer`] is used to sign data, typically for authentication or authorization purposes.
```
dbolduc: nit: can you re-flow the text lines to be <= 80 characters long?

src/auth/src/signer.rs
```diff
@@ -12,6 +12,62 @@
...
+//! use google_cloud_auth::credentials::Builder;
+//! use google_cloud_auth::signer::Signer;
+//!
+//! # fn build_signer() -> Result<Signer, Box<dyn std::error::Error>> {
+//! let signer = Builder::default().build_signer()?;
+//! # Ok(signer)
+//! # }
+//! ```
+//!
+//! ## Example: Creating a Signer using a Service Account Key File
```
dbolduc, line 41: I thought we want to discourage people from downloading these

src/auth/src/signer.rs
```diff
@@ -12,6 +12,62 @@
...
+//! useful in environments where network access is restricted and performance is critical.
+//!
+//! ```rust,no_run
+//! use google_cloud_auth::credentials::service_account::Builder;
+//! use google_cloud_auth::signer::Signer;
+//!
+//! # async fn build_signer() -> Result<Signer, Box<dyn std::error::Error>> {
+//! // In a real application, you would read this from a file.
+//! let service_account_key = serde_json::json!({
+//!     "type": "service_account",
```
dbolduc: Can we make this minimal by removing the fields that are ignored by the builder? I think it only cares about these:

https://github.com/googleapis/google-cloud-rust/blob/807266b02e90fe5ac83ab08308ec6c0aad0603d3/src/auth/src/credentials/service_account.rs#L356-L369

src/auth/src/signer.rs
```diff
@@ -12,6 +12,62 @@
...
+//! One primary use case in the Google Cloud ecosystem is generating [Signed URLs] for Google Cloud Storage.
+//!
+//! The main type in this module is [Signer]. This is an opaque type
+//! that implements the [SigningProvider] trait and can be used to
+//! sign content. Use [crate::credentials::Builder::build_signer]
+//! to create a `Signer` from loaded credentials.
+//!
+//! ## Example: Creating a Signer using Application Default Credentials (ADC)
+//!
+//! This is the recommended way for most applications. It automatically finds credentials from the environment.
```
dbolduc: Maybe add that "If your ADC are [service account credentials], the signing will be local and not make any network requests. This can be useful in environments where network access is restricted and performance is critical."

Then we might not need the next section?

Maybe. Your call.

src/auth/src/signer.rs
dbolduc: (if we keep it)

src/auth/src/signer.rs
dbolduc: If you remove the other section, this could just be:

```md
Applications can use their Application Default Credentials (ADC) to create a signer.

If your ADC are [service account credentials], the signing will be local and not
make any network requests. This can be useful in environments where network access
is restricted and performance is critical.

# Example
// blah blah blah
// let signer = Builder::default().build_signer()?;
// blah blah blah
```

dbolduc (dismissed):
This PR is an obvious improvement to the docs. Feel free to merge and address any of my nits in a follow up

src/auth/src/signer.rs
alvarowolfx: removed the extra info, indeed was not needed.

Issue #3972, opened 2025-12-03
Respect subscriber RPC request size limits


dbolduc:
There is a size limit on requests for the `Acknowledge` and `ModifyAckDeadline` RPCs.

https://cloud.google.com/pubsub/quotas#resource_limits

We need to at least make sure that when we send these RPCs they are within the size limit. (This may look like breaking up a batch into multiple requests).

We should also consider flushing acks/nacks as soon as an RPC is full (instead of only on a timer).

Issue #3975, opened 2025-12-03
Optimize subscriber lease actions


dbolduc:
For example, in `LeaseState::shutdown()`, we ack messages, and we nack messages. This can take two different RPCs. We should spawn tasks for each of these actions, and await them concurrently.

This also applies to `LeaseState::flush()`, which should not block the lease event loop.

dbolduc:
We should hold pending operations in a `JoinSet`.

Another optimization strategy is to flush acks when (1) there are >= 1 acks and (2) there are <= N pending operations.

This would give us ideal latency for a subscription with low traffic. (i.e. if the messages are published at a rate less than 1 Hz).

It also keeps the pipes flowing for a subscription with high traffic. We would fill up our batches while earlier acknowledge RPCs are in transit.

suzmue:
Possibly related: Probably need to keep ordered + exactly once in mind as we refactor this, since these acks need to be in order: https://docs.cloud.google.com/pubsub/docs/exactly-once-delivery#exactly-once_delivery_and_ordered_subscriptions. Though this could just be handled in user code by making sure that the `confirmed_ack` returns, before acking the next message, but that becomes very low throughput

dbolduc:
<img width="1401" height="411" alt="Image" src="https://github.com/user-attachments/assets/3066d177-12b7-4275-8067-8ba94231a647" />

We are seeing similar top-end performance after these changes. That's good.

PR #4184, opened 2026-01-08
impl(pubsub): respect ack/nack rpc size quotas

dbolduc:
Part of the work for #3957, #3972

Flush acks/nacks if they start approaching the RPC size quota.

I hate hate hate how I decided to flush acks and nacks as one event, instead of as separate events. I will clean that up but not in this PR. In this PR, we kick the can.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -99,6 +99,8 @@ where
     /// statement.
     pub(crate) async fn next_event(&mut self) -> LeaseEvent {
         if self.to_ack.len() >= ACK_IDS_PER_RPC || self.to_nack.len() >= ACK_IDS_PER_RPC {
+            // This is an OR because `Acknowledge` and `ModifyAckDeadline` are
+            // separate RPCs, with separate limits.
             return LeaseEvent::Flush;
         }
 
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -92,8 +98,9 @@ where
     /// hold one mutable reference to `LeaseState` within its `select!`
     /// statement.
     pub(crate) async fn next_event(&mut self) -> LeaseEvent {
-        // TODO(#3972) - flush on size if an `Acknowledge` or
-        // `ModifyAckDeadline` RPC is full.
+        if self.to_ack.len() >= ACK_IDS_PER_RPC || self.to_nack.len() >= ACK_IDS_PER_RPC {
```
coryan, line 101: Since the limit is per request, shouldn't this be:

```suggestion
        if self.to_ack.len() + self.to_nack_len() >= ACK_IDS_PER_RPC  {
```

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -417,4 +428,88 @@ pub(crate) mod tests {
         assert_eq!(state.next_event().await, LeaseEvent::Flush);
         assert_eq!(start.elapsed(), Duration::from_secs(7));
     }
+
+    #[tokio::test(start_paused = true)]
+    async fn limit_size_of_ack_batch() {
```
coryan, line 435: Seems like you need a combined ack + nack message limit?

coryan (changes requested):
aside: I never implemented ack batching in C++, I am not sure we will it.

src/pubsub/src/subscriber/lease_state.rs
dbolduc: They are different RPCs. One is `Acknowledge`, the other is `ModifyAckDeadline`.

Added a comment and a test that verifies we use OR.

src/pubsub/src/subscriber/lease_state.rs
dbolduc: Nope, because the RPCs are separate.

dbolduc:
> I never implemented ack batching in C++, I am not sure we will need it.

Batching acks should give us better throughput.

coryan (approved):

coryan:
> > I never implemented ack batching in C++, I am not sure we will need it.
> 
> Batching acks should give us better throughput.

Lower client-CPU / network costs per Ack: yes. Better throughput? You are delaying the service sending you more data while you batch the acks.  That could _lower_ the overall throughput, depending on how the service handles larger acks vs. lots of smaller acks.

dbolduc:
> > Batching acks should give us better throughput.
> 
> Lower client-CPU / network costs per Ack: yes. Better throughput? You are delaying the service sending you more data while you batch the acks. That could _lower_ the overall throughput, depending on how the service handles larger acks vs. lots of smaller acks.

Good points. I agree and appreciate the insight.

I have an issue to track optimizing this stuff, and just pinged it with my plan, which is kinda the happy middle between acking immediately and filling batches: https://github.com/googleapis/google-cloud-rust/issues/3975#issuecomment-3726132207

(If the design is familiar, that is because I stole it from the C++ bigtable mutation batcher)

My first priority is making a Subscriber available in the next `google-cloud-pubsub` preview that runs and is correct. Then I will start working on making it fast.

PR #4185, opened 2026-01-08
feat(storage): mark signed url stable

alvarowolfx:
Towards #3645

### DIFF
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -18,7 +18,7 @@ name = "google-cloud-auth"
 #     version of all downstream dependencies. For details see:
 #         https://github.com/googleapis/google-cloud-rust/issues/3237
 #         https://github.com/googleapis/google-cloud-rust/issues/3265
-version     = "1.5.0"
+version     = "1.4.0"
 description = "Google Cloud Client Libraries for Rust - Authentication"
 build       = "build.rs"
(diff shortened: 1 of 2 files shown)


src/auth/Cargo.toml
```diff
@@ -18,7 +18,7 @@ name = "google-cloud-auth"
 #     version of all downstream dependencies. For details see:
 #         https://github.com/googleapis/google-cloud-rust/issues/3237
 #         https://github.com/googleapis/google-cloud-rust/issues/3265
-version     = "1.4.0"
+version     = "1.5.0"
```
dbolduc: auth was already bumped in https://github.com/googleapis/google-cloud-rust/pull/4152

alvarowolfx: I'm trying to find why minimal-version check was failing https://github.com/googleapis/google-cloud-rust/runs/59931178068. I can't repro locally

dbolduc: Your change to make the workspace version of auth `1.4` in `Cargo.toml` is probably the fix.

The build was failing because it tried to build storage against auth `1.3` which didn't have any of the new signer stuff.

alvarowolfx: yeah, I thought that bumping was gonna solve the issue, I just missed that auth was already bumped before. Thanks

PR #4189, opened 2026-01-08
chore: create librarian.yaml as first step to migrating google-cloud-rust to librarian

ldetmer:
This maps the list of libraries into a librarian.yaml file that librarian tool uses as source of configuration values.

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -266,10 +266,7 @@ libraries:
           match: |2
                 - `folders/<folder-number>`
                 - `organizations/<organization-number>`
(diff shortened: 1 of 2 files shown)


librarian.yaml
coryan, line 1: The formatting robots do not like the new file. Consider reformatting or excluding the file here:

https://github.com/googleapis/google-cloud-rust/blob/980aa25187528adf8ed26cc517558ee72166247a/.gcb/format.yaml#L69

coryan (approved):

PR #4195, opened 2026-01-08
impl(pubsub): respect modack size limit

### LINKED ISSUE #3972: Respect subscriber RPC request size limits

dbolduc:
Part of the work for #3957, Fixes #3972 

Make sure we break up our lease extensions into requests that do not exceed the service's RPC size limits. (The service rejects any request over the limit).

TIL: `chunks()`

### DIFF
--- a/src/pubsub/src/subscriber/lease_loop.rs
+++ b/src/pubsub/src/subscriber/lease_loop.rs
@@ -371,12 +371,7 @@ mod tests {
 
             // Confirm that no messages are under lease management.
             {
-                mock.lock()
-                    .await
-                    .expect_extend()
-                    .times(1)
-                    .withf(|v| v.is_empty())
-                    .returning(|_| ());
+                mock.lock().await.expect_extend().times(0);
                 tokio::time::advance(Duration::from_millis(100)).await;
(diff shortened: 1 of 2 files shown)


src/pubsub/src/subscriber/lease_loop.rs
```diff
@@ -371,12 +371,7 @@ mod tests {
 
             // Confirm that no messages are under lease management.
             {
```
dbolduc, line 373: Note to the reviewer: this test changed because we went from calling `leaser.extend(Vec::new())` to not calling `leaser.extend(...)` at all. (Because an empty vector turns into 0 chunks).

coryan: Ack.

PR #4197, opened 2026-01-09
docs(storage): improve signed url builder docs

alvarowolfx:
Improvements to SignedUrlBuilder docs:
* Hide use google_cloud_storage::builder::storage::SignedUrlBuilder
* Add warning about using Service Account
* Point to the `http:Method::GET` on `with_method`
* Show signed url generation examples before Signer examples
* Document the behavior of `with_header` and `with_query_param` that they override value if same key is set
* Link to what is ADC.
* Show wrapper func on examples so developer knows where the `signer` comes from

Towards #3645

### DIFF
--- a/src/storage/src/storage/signed_url.rs
+++ b/src/storage/src/storage/signed_url.rs
@@ -127,7 +127,6 @@ pub struct SignedUrlBuilder {
     headers: BTreeMap<String, String>,
     query_parameters: BTreeMap<String, String>,
     endpoint: Option<String>,
-    universe_domain: String,
     client_email: Option<String>,
     timestamp: DateTime<Utc>,
     url_style: UrlStyle,
@@ -215,7 +214,6 @@ impl SignedUrlBuilder {
             headers: BTreeMap::new(),
             query_parameters: BTreeMap::new(),
             endpoint: None,
-            universe_domain: "googleapis.com".to_string(),
             client_email: None,
             timestamp: Utc::now(),
             url_style: UrlStyle::PathStyle,
@@ -390,8 +388,6 @@ impl SignedUrlBuilder {
     ///
     /// This is useful when using a custom domain, or when testing with some Cloud Storage emulators.
     ///
-    /// Setting an endpoint takes precedence over using `with_universe_domain`.
-    ///
     /// # Example
     ///
     /// ```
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/signed_url.rs
```diff
@@ -219,9 +227,8 @@ impl SignedUrlBuilder {
     /// # Example
     ///
     /// ```
-    /// use google_cloud_storage::builder::storage::SignedUrlBuilder;
+    /// # use google_cloud_storage::builder::storage::SignedUrlBuilder;
     /// # use auth::signer::Signer;
-    ///
     /// # async fn run(signer: &Signer) -> anyhow::Result<()> {
```
coryan: We also discussed doing this:

```suggestion
    /// async fn run(signer: &Signer) -> anyhow::Result<()> {
```

So the application developer knows where the magic `signer` comes from...

src/storage/src/storage/signed_url.rs
```diff
@@ -266,13 +272,14 @@ impl SignedUrlBuilder {
         self
     }
 
-    /// Sets the HTTP method for the signed URL. The default is "GET".
+    /// Sets the HTTP method for the signed URL. The default is [http::Method::GET][crate::signed_url::http::Method::GET].
```
coryan: I think this may be enough?  

```suggestion
    /// Sets the HTTP method for the signed URL. The default is [GET][crate::signed_url::http::Method::GET].
```

src/storage/src/storage/signed_url.rs
```diff
@@ -377,7 +387,7 @@ impl SignedUrlBuilder {
         self
     }
 
-    /// Sets the endpoint for the signed URL. The default is "https://storage.googleapis.com".
+    /// Sets the endpoint for the signed URL. The default is "<https://storage.googleapis.com>".
```
coryan: Do you want that link to be clickable?  Maybe:


```suggestion
    /// Sets the endpoint for the signed URL. The default is `"https://storage.googleapis.com"`.
```

coryan (approved):
A few nits.

src/storage/src/storage/signed_url.rs
alvarowolfx: I was going to do in another PR since I ended up bundling too many changes here. But will add that too.

src/storage/src/storage/signed_url.rs
alvarowolfx: good call

src/storage/src/storage/signed_url.rs
alvarowolfx: I'm dumb, I was trying to get rid of the docs build error and forgot that a simple backtick could solve.

coryan (approved):

PR #4198, opened 2026-01-09
impl(storage): move from tokio::sync::SetOnce to std::sync::OnceLock

alvarowolfx:
tokio::sync::SetOnce required newer version of tokio. Move to `std` lib alternative.

Towards #3645  and fix minimal-version check on https://github.com/googleapis/google-cloud-rust/pull/4185


coryan (approved):
Approved for realsies

PR #4199, opened 2026-01-09
impl(storage): signed url builder should accept FQN bucket name

alvarowolfx:
The SignedUrlBuilder is already stripping `projects/_buckets/` and working with the main client, but the format was not being enforced.

Towards #3645

### DIFF
--- a/src/storage/src/storage/signed_url.rs
+++ b/src/storage/src/storage/signed_url.rs
@@ -246,7 +246,7 @@ impl SignedUrlBuilder {
     /// # use google_cloud_storage::builder::storage::SignedUrlBuilder;
     /// # use auth::signer::Signer;
     /// async fn run(signer: &Signer) -> anyhow::Result<()> {
-    ///     let url = SignedUrlBuilder::for_object("my-bucket", "my-object.txt")
+    ///     let url = SignedUrlBuilder::for_object("projects/_/buckets/my-bucket", "my-object.txt")
     ///         .sign_with(signer)
     ///         .await?;
     /// # Ok(())
@@ -268,7 +268,7 @@ impl SignedUrlBuilder {
     /// # use google_cloud_storage::builder::storage::SignedUrlBuilder;
     /// # use auth::signer::Signer;
     /// async fn run(signer: &Signer) -> anyhow::Result<()> {
-    ///     let url = SignedUrlBuilder::for_bucket("my-bucket")
+    ///     let url = SignedUrlBuilder::for_bucket("projects/_/buckets/my-bucket")
     ///         .sign_with(signer)
     ///         .await?;
     /// # Ok(())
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/signed_url.rs
```diff
@@ -663,6 +684,26 @@ mod tests {
...
+    async fn test_signed_url_error_bucket() -> TestResult {
+        let mut mock = MockSigner::new();
+        mock.expect_client_email()
+            .return_once(|| Ok("test@example.com".to_string()));
+        mock.expect_sign()
+            .return_once(|_content| Ok(bytes::Bytes::from("test-signature")));
+
         let signer = Signer::from(mock);
         let err = SignedUrlBuilder::for_object("b", "o")
             .with_endpoint("invalid url")
```
dbolduc: I don't think you mean to set an invalid endpoint in this test.

src/storage/src/storage/signed_url.rs
```diff
@@ -663,6 +684,26 @@ mod tests {
...
+    #[tokio::test]
+    async fn test_signed_url_error_bucket() -> TestResult {
+        let mut mock = MockSigner::new();
+        mock.expect_client_email()
+            .return_once(|| Ok("test@example.com".to_string()));
+        mock.expect_sign()
+            .return_once(|_content| Ok(bytes::Bytes::from("test-signature")));
+
         let signer = Signer::from(mock);
         let err = SignedUrlBuilder::for_object("b", "o")
```
dbolduc: Consider something that jumps at the reader like `for_object("invalid-bucket-name", "o")`

src/storage/src/storage/signed_url.rs
```diff
@@ -663,6 +684,26 @@ mod tests {
...
             .return_once(|_content| Ok(bytes::Bytes::from("test-signature")));
 
+        let signer = Signer::from(mock);
+        let err = SignedUrlBuilder::for_object("projects/_/buckets/b", "o")
+            .with_endpoint("invalid url")
+            .sign_with(&signer)
+            .await
+            .unwrap_err();
+
+        assert!(err.is_invalid_parameter());
```
dbolduc, line 701: comment: hopefully something tells me which parameters are invalid.

optional: you could check that the error debug string contains "invalid url" and/or "endpoint"

dbolduc (approved):
Approved, but please look at `test_signed_url_error_bucket` again

src/storage/src/storage/signed_url.rs
alvarowolfx: ugh, that's what happen when you code too late through the night

src/storage/src/storage/signed_url.rs
alvarowolfx: added extra check on debug string

PR #4201, opened 2026-01-09
docs(auth): improve signer docs with devsite refs

alvarowolfx:
Improvements to Signer docs:
* Add warning about using Service Account
* Link to what is ADC.
 
Towards #3645

### DIFF
--- a/src/auth/src/signer.rs
+++ b/src/auth/src/signer.rs
@@ -23,10 +23,11 @@
 //! sign content. Use [crate::credentials::Builder::build_signer]
 //! to create a `Signer` from loaded credentials.
 //!
-//! ## Example: Creating a Signer using Application Default Credentials (ADC)
+//! ## Example: Creating a Signer using [Application Default Credentials] (ADC)
 //!
 //! This is the recommended way for most applications. It automatically finds
-//! credentials from the environment.
+//! credentials from the environment. See how [Application Default Credentials]
+//! works.
 //!
 //! ```
 //! use google_cloud_auth::credentials::Builder;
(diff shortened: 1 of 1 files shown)


dbolduc (approved):

src/auth/src/signer.rs
```diff
@@ -46,6 +47,12 @@
 //! environments where network access is restricted and performance is
 //! critical.
 //!
+//! <div class="warning">
+//!     <strong>Caution:</strong> Service account keys are a security risk if not managed correctly.
+//!     See <a href="https://docs.cloud.google.com/iam/docs/best-practices-for-managing-service-account-keys">
+//!     Best practices for managing service account keys</a> for more information.
```
coryan, line 53: Can you use markdown inside the `<div>`?


```suggestion
//! <div class="warning">
//!     **Caution:** Service account keys are a security risk if not managed correctly.
//!     See [Best practices for managing service account keys] for more information.
```

And then add this at the bottom?

```
/// [Best practices for managing service account keys]: https://docs.cloud.google.com/iam/docs/best-practices-for-managing-service-account-keys
```

alvarowolfx: it doesn't render markdown inside divs

coryan: Apparently it can happen with some caveats:

https://doc.rust-lang.org/rustdoc/how-to-write-documentation.html#adding-a-warning-block

alvarowolfx: oh, interesting, I tried to use the markdown syntax and just assumed was not supported. Will try that trick from the docs

alvarowolfx: hmm, I wonder if there a version of rust that supports that, because I tried this and it is still not rendering:

```diff
//! <div class="warning">
//!
//!     **Caution:** Service account keys are a security risk if not managed correctly.
//!     See [Best practices for managing service account keys] for more information.
//!
//! </div>
```
Screenshot 2026-01-09 at 09 39 53

coryan: To get closer to the rendering at the `docs.rs` site use:

```shell
rustyp update nightly
cargo install --locked cargo-docs-rs
cargo +nightly docs-rs --package google-cloud-storage
```

Image

But then again, it is not that important.

PR #4203, opened 2026-01-09
impl(pubsub): lease loop drains acks on shutdown

dbolduc:
Part of the work for #3957 

If the lease loop is about to shutdown, drain all the acks received from the application first.

Actually set expectations in the test that verifies shutdown behavior.

### DIFF
--- a/src/pubsub/src/subscriber/lease_loop.rs
+++ b/src/pubsub/src/subscriber/lease_loop.rs
@@ -15,7 +15,7 @@
 use super::handler::AckResult;
 use super::lease_state::{LeaseEvent, LeaseOptions, LeaseState};
 use super::leaser::Leaser;
-use tokio::sync::mpsc::{UnboundedSender, unbounded_channel};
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_loop.rs
```diff
@@ -49,7 +49,16 @@ impl LeaseLoop {
...
                         match message {
-                            None => break state.shutdown().await,
+                            None => {
+                                while let Ok(r) = ack_rx.try_recv() {
+                                    if let AckResult::Ack(ack_id) = r {
+                                        // Process any acks that we already know about.
+                                        state.ack(ack_id);
+                                    }
+                                }
+                                state.shutdown().await;
```
coryan: Refactor to a function?

dbolduc: Done. Probably wise to avoid heavy code in a macro like `tokio::select!`.

I preferred to keep it outside of the `LeaseState` class. I don't think that class needs to know about channels.

coryan (approved):

PR #4208, opened 2026-01-09
test(storage): non-default crypto providers

coryan:
Verify storage works with the default crypto provider and with non-default crypto providers. Later I will expand these tests to verify we can eliminate the dependency on the default crypto provider.

Part of the work for #4170

### DIFF
--- a/tests/crypto-providers/test-storage/src/lib.rs
+++ b/tests/crypto-providers/test-storage/src/lib.rs
@@ -31,7 +31,7 @@ pub async fn run() -> anyhow::Result<()> {
         .list_buckets()
         .set_parent(format!("projects/{project_id}"))
         .by_item();
-    let _ = buckets.next().await.transpose()?.expect("expected at least one bucket");
+    let _ = buckets.next().await.expect("expected at least one bucket")?;
 
     // Verify JSON requests work as expected:
     let storage = Storage::builder().build().await?;


tests/crypto-providers/test-storage/src/lib.rs
```diff
@@ -0,0 +1,54 @@
...
+    let mut buckets = client
+        .list_buckets()
+        .set_parent(format!("projects/{project_id}"))
+        .by_item();
+    let mut count = 0;
+    while let Some(b) = buckets.next().await.transpose()? {
+        println!("found bucket {}", b.name);
+        count += 1;
+    }
+    assert_ne!(count, 0);
```
dbolduc: optional nit: Maybe just this (but also, who cares):

```rs
let _ = buckets.next().await.expect("expected at least one bucket")?;
```

tests/crypto-providers/test-storage/src/lib.rs
```diff
@@ -0,0 +1,49 @@
...
+/// The caller may need to configure the default crypto provider.
+pub async fn run() -> anyhow::Result<()> {
+    let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
+    // Verify gRPC requests work as expected:
+    let client = StorageControl::builder().build().await?;
+    let mut buckets = client
+        .list_buckets()
+        .set_parent(format!("projects/{project_id}"))
+        .by_item();
+    let _ = buckets.next().await.transpose()?.expect("expected at least one bucket");
```
dbolduc: nit: the order of operations doesn't matter, so you can skip the `transpose()`.

```rs
let _ = buckets.next().await.expect("expected at least one bucket")?;
```

dbolduc (approved):

tests/crypto-providers/storage-with-ring/Cargo.toml
```diff
@@ -0,0 +1,37 @@
...
+categories   = ["network-programming"]
+rust-version = "1.92.0"
+
+# This test needs to be its own workspace so it is isolated from any features
+# enabled in the top-level Cargo.toml file (and its dependencies).
+[workspace]
+
+[dependencies]
+anyhow        = { default-features = false, version = "1", features = ["std"] }
+rustls        = { default-features = false, version = "0.23", features = ["aws-lc-rs"] }
```
amanda-tarafa, line 34: I think this needs to be ring?

coryan: Yes. Already caught this in:

https://github.com/googleapis/google-cloud-rust/pull/4299/changes#diff-c1d7ad015e014abafaa8d14e2b8308d50e91caac26ff6b0bf06ffbaac19af005

PR #4214, opened 2026-01-09
impl(pubsub): Session includes lease management

### LINKED ISSUE #3957: Implement subscriber

dbolduc:
Fixes #3957 

Plug in lease management to the session.

In these unit tests, we do enough to verify that the lease loop is running. We do not recreate every single test of the lease loop in the session.

### DIFF
--- a/src/pubsub/src/subscriber/session.rs
+++ b/src/pubsub/src/subscriber/session.rs
@@ -71,13 +71,12 @@ pub struct Session {
     /// when it is dropped.
     _keepalive_guard: DropGuard,
 
-    #[cfg(test)]
     /// A handle on the lease loop task.
     ///
     /// We hold onto this handle so we can await pending lease operations. While awaiting pending
     /// lease operations is useful for setting expectations in our unit tests, it is not that
     /// helpful to applications in practice.
-    lease_loop: tokio::task::JoinHandle<()>,
+    _lease_loop: tokio::task::JoinHandle<()>,
 }
 
 impl Session {
@@ -113,8 +112,7 @@ impl Session {
             message_tx,
             ack_tx,
             _keepalive_guard: shutdown.drop_guard(),
-            #[cfg(test)]
-            lease_loop: _lease_loop,
+            _lease_loop,
         })
     }
 
@@ -195,7 +193,7 @@ impl Session {
         drop(self.message_tx);
 
         // Wait for the lease management task to complete.
-        self.lease_loop.await?;
+        self._lease_loop.await?;
 
         Ok(())
     }


src/pubsub/src/subscriber/leaser.rs
```diff
@@ -31,7 +31,7 @@ pub(crate) trait Leaser {
     async fn extend(&self, ack_ids: Vec<String>);
 }
 
-struct DefaultLeaser<T>
+pub(crate) struct DefaultLeaser<T>
```
coryan, line 34: `pub(super)` is also an alternative.

src/pubsub/src/subscriber/lease_loop.rs
```diff
@@ -21,11 +21,11 @@ use tokio::task::JoinHandle;
 /// A convenience struct that groups the components of the lease loop.
 pub(crate) struct LeaseLoop {
     /// A handle to the task running the lease loop.
-    handle: JoinHandle<()>,
+    pub(crate) handle: JoinHandle<()>,
```
coryan, line 24: Ditto re: `pub(super)`.

src/pubsub/src/subscriber/session.rs
```diff
@@ -56,26 +59,46 @@ pub struct Session {
...
     /// when it is dropped.
     _keepalive_guard: DropGuard,
+
+    #[cfg(test)]
+    /// A handle on the lease loop task.
+    ///
+    /// We hold onto this handle so we can await pending lease operations. While awaiting pending
+    /// lease operations is useful for setting expectations in our unit tests, it is not that
+    /// helpful to applications in practice.
+    lease_loop: tokio::task::JoinHandle<()>,
```
coryan:
```suggestion
    /// A handle on the lease loop task.
    ///
    /// We hold onto this handle so we can await pending lease operations. While awaiting pending
    /// lease operations is useful for setting expectations in our unit tests, it is not that
    /// helpful to applications in practice.
    _lease_loop: tokio::task::JoinHandle<()>,
```

If you want less conditional code.

src/pubsub/src/subscriber/session.rs
```diff
@@ -286,10 +341,115 @@ mod tests {
...
+
+        Ok(())
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn basic_lease_management() -> anyhow::Result<()> {
+        let (response_tx, response_rx) = channel(10);
+        let (ack_tx, mut ack_rx) = unbounded_channel();
+        let (nack_tx, mut nack_rx) = unbounded_channel();
+        let (extend_tx, mut extend_rx) = unbounded_channel();
```
coryan, line 363: Clever, I should use more of these.

coryan (approved):
Just small nits for you to consider.

src/pubsub/src/subscriber/lease_loop.rs
dbolduc: Ah. I did not think of that. That would be a better fit with the flat `mod subscriber` I have set up...

I don't want to change things now, because there are like 99 other instances of `pub(crate)` that warrant the same treatment. I'll maybe send a PR to change them en masse later.

src/pubsub/src/subscriber/leaser.rs
dbolduc: See above.

src/pubsub/src/subscriber/session.rs
dbolduc: Thanks @suzmue for teaching me about them. They are very convenient from inside the synchronous mockall `expect_` functions.

src/pubsub/src/subscriber/session.rs
dbolduc: Yeah, that is nicer. Thanks.

src/pubsub/src/subscriber/lease_loop.rs
dbolduc: #4223

PR #4216, opened 2026-01-09
refactor(pubsub): separate BatchWorker loop logic for ordering key

PhongChuong:


src/pubsub/src/publisher/worker.rs
```diff
@@ -240,20 +240,59 @@ impl BatchWorker {
...
+            self.run_without_ordering_key().await;
+        } else {
+            self.run_with_ordering_key().await;
+        }
+    }
+
+    async fn run_without_ordering_key(&mut self) {
         // While it is possible to use Some(JoinHandle) here as there is at max
         // a single inflight task at any given time, the use of JoinSet
         // simplify the managing the inflight JoinHandle.
```
suzmue: This comment I think is for with_ordering_key not without.

src/pubsub/src/publisher/worker.rs
```diff
@@ -240,20 +240,59 @@ impl BatchWorker {
...
             tokio::select! {
                 _ = inflight.join_next(), if !inflight.is_empty() => {
-                    if unlimited_inflight {
-                        // When there are no limits to the number of inflight, we can skip sending the
-                        // next batch as it is handled by the Publish and Flush operations.
-                        continue;
+                    // When there are no limits to the number of inflight, we can skip sending the
+                    // next batch as it is handled by the Publish and Flush operations.
+                    continue;
+                }
```
suzmue, line 257: Consider removing this comment, I think it makes it more confusing.

src/pubsub/src/publisher/worker.rs
```diff
@@ -240,20 +240,59 @@ impl BatchWorker {
...
+                            }
+                        },
+                        Some(ToBatchWorker::Flush(tx)) => {
+                            // Send all the batches concurrently.
+                            while !self.pending_batch.is_empty() || !self.pending_msgs.is_empty() {
+                                self.move_to_batch();
+                                self.pending_batch.flush(self.client.clone(), self.topic.clone(), &mut inflight);
+                            }
+                            inflight.join_all().await;
+                            inflight = JoinSet::new();
```
suzmue, line 274: This line is unnecessary, join_all will have emptied it.
```suggestion
```

src/pubsub/src/publisher/worker.rs
```diff
@@ -240,20 +240,59 @@ impl BatchWorker {
...
+                            let _ = tx.send(());
+                        },
+                        None => {
+                            break;
```
suzmue, line 279: Can you add TODO here to think about shutdown behavior for the BatchWorkers?

src/pubsub/src/publisher/worker.rs
```diff
@@ -240,20 +240,59 @@ impl BatchWorker {
...
+                    continue;
+                }
+                msg = self.rx.recv() => {
+                    match msg {
+                        Some(ToBatchWorker::Publish(msg)) => {
+                            self.pending_msgs.push_back(msg);
+                            self.move_to_batch();
+                            if self.at_batch_threshold() {
+                                self.pending_batch.flush(self.client.clone(), self.topic.clone(), &mut inflight);
+                            }
```
suzmue, line 265: nit:  Don't need to address in this CL, but I think it would be nice to refactor this into a function or combine them. in some way. We never do self.pending_msgs.push_back without move_to_batch which signals to me that we should maybe be combining them.

src/pubsub/src/publisher/worker.rs
```diff
@@ -264,29 +303,19 @@ impl BatchWorker {
...
-                                if !inflight.is_empty() {
-                                    inflight.join_next().await;
-                                }
-                                while !self.pending_batch.is_empty() || !self.pending_msgs.is_empty() {
-                                    self.move_to_batch();
-                                    self.pending_batch.flush(self.client.clone(), self.topic.clone(), &mut inflight);
-                                    inflight.join_next().await;
-                                }
+                            // Send batches sequentially.
+                            if !inflight.is_empty() {
```
suzmue: I think for this invocation of inflight.join_next(), we don't need to check that it is non-empty first, since it will return None if the set was empty. https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.join_next

It is still useful on the tokio::select join_next, because we would rather the other branches execute.

src/pubsub/src/publisher/worker.rs
```diff
@@ -240,20 +240,59 @@ impl BatchWorker {
...
+            self.run_with_ordering_key().await;
+        }
+    }
+
+    async fn run_without_ordering_key(&mut self) {
         // While it is possible to use Some(JoinHandle) here as there is at max
         // a single inflight task at any given time, the use of JoinSet
         // simplify the managing the inflight JoinHandle.
         let mut inflight = JoinSet::new();
         // For empty ordering key, we do not need to limit the number of inflight batches.
```
suzmue: nit: for messages without an ordering key.
```suggestion
        // For messages without an ordering key, we can have multiple inflight batches concurrently.
```

or something like that

src/pubsub/src/publisher/worker.rs
PhongChuong: This is needed. [join_all](https://docs.rs/tokio/latest/tokio/task/struct.JoinSet.html#method.join_all) consumes the JoinSet.

src/pubsub/src/publisher/worker.rs

src/pubsub/src/publisher/worker.rs
PhongChuong: Done.

src/pubsub/src/publisher/worker.rs
suzmue: TIL. thanks

PR #4217, opened 2026-01-09
refactor(storage): aggregate http mod re-exports

alvarowolfx:
Per discussions on how to re-export types from other crates.

### DIFF
--- a/src/storage/tests/mocking.rs
+++ b/src/storage/tests/mocking.rs
@@ -25,8 +25,8 @@ mod tests {
     use gcs::request_options::RequestOptions;
     use gcs::streaming_source::{BytesSource, Payload, Seek, StreamingSource};
     use gcs::{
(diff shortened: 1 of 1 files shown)


src/storage/src/object_descriptor.rs
```diff
@@ -14,15 +14,13 @@
...
+use crate::http::HeaderMap;
 use crate::model::Object;
 use crate::model_ext::ReadRange;
 use crate::read_object::ReadObjectResponse;
 use crate::storage::bidi::stub::dynamic::ObjectDescriptor as ObjectDescriptorStub;
 use std::sync::Arc;
 
-/// The `open_object()` metadata attributes.
-pub use http::HeaderMap;
-
```
coryan, line 25: Note to self: this is Okay to remove because the feature has not been released yet.

coryan (approved):

PR #4220, opened 2026-01-09
feat(auth): prune crypto provider dependencies

coryan:
With the default features enabled `google-cloud-auth` selects a crypto
provider automatically. Note that whatever features are enabled, the
applications can always override the crypto provider, but before this
change it was impossible to prune the default provider from the
dependency tree.

With this change, it is possible to compile `google-cloud-auth` without
linking the default crypto provider for rustls. If the library is
compiled without a default crypto provider, applications **must**
configure the provider using
`rustls::CryptoProvider::install_default()`.

Note that most other `google-cloud-*` crates enable the default crypto
provider. Future PRs will add a similar `default-tls` feature to all
downstream crates. This feature will be enabled by default, but if
disabled the downstream crates will not require a provider from
`google-cloud-auth` either.

Part of the work for #4170

### DIFF
--- a/src/auth/README.md
+++ b/src/auth/README.md
@@ -16,7 +16,7 @@ also describes the common terminology used with authentication, such as
 
 # Features
 
-- `default-rustls-provider`: enabled by default. This feature select a default
+- `default-rustls-provider`: enabled by default. This feature selects a default
   crypto provider and trusted root certificate selection for TLS. Applications
   that have specific requirements for TLS (such as exclusively using the
   [aws-lc-rs], or [ring] crates) should disable this default and configure the
@@ -30,7 +30,7 @@ also describes the common terminology used with authentication, such as
   for this backend should not rely on the current default. To control the
   backend selection:
   - Configure this crate with `default-features = false`, and
-    `features = ["jsonwebtoken"]`
+    `features = ["idtoken"]`
   - Select the desired backend for `jsonwebtoken`.
 
 [authentication methods at google]: https://cloud.google.com/docs/authentication
--- a/src/auth/src/lib.rs
+++ b/src/auth/src/lib.rs
@@ -26,7 +26,7 @@
 //!
 //! # Features
 //!
-//! - `default-rustls-provider`: enabled by default. This feature select a default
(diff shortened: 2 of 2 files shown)


src/auth/README.md
```diff
@@ -25,14 +30,16 @@ also describes the common terminology used with authentication, such as
   for this backend should not rely on the current default. To control the
   backend selection:
   - Configure this crate with `default-features = false`, and
-    `features = ["idtoken"]`
-  - Configure the `jsonwebtoken` crate to use the desired backend.
+    `features = ["jsonwebtoken"]`
```
dbolduc: Revert. Rebase error?

src/auth/README.md
```diff
@@ -16,6 +16,11 @@ also describes the common terminology used with authentication, such as
 
 # Features
 
+- `default-rustls-provider`: enabled by default. This feature select a default
```
dbolduc: s/select/selects/

src/auth/src/lib.rs
```diff
@@ -35,9 +40,11 @@
 //!   for this backend should not rely on the current default. To control the
 //!   backend selection:
 //!   - Configure this crate with `default-features = false`, and
-//!     `features = ["idtoken"]`
-//!   - Configure the `jsonwebtoken` crate to use the desired backend.
+//!     `features = ["jsonwebtoken"]`
```
dbolduc: Same

src/auth/src/lib.rs
```diff
@@ -26,6 +26,11 @@
 //!
 //! # Features
 //!
+//! - `default-rustls-provider`: enabled by default. This feature select a default
```
dbolduc: s/select/selects/

src/auth/README.md
coryan: Fixed.

src/auth/README.md
coryan: Fixed.

src/auth/src/lib.rs
coryan: Fixed, also fixed the "select -> selects" above.

PR #4221, opened 2026-01-09
idtoken: cache `user_account` token provider

howardjohn:
The 3 other types are cached but this one is not. Its simple to cache it so seems appropriate to do so.


coryan:
Thanks for the PR!  I wonder if it is possible to add a unit test to verify caching is working?

alvarowolfx:
approving for as is and I can add the unit tests

coryan (approved):

PR #4225, opened 2026-01-09
impl(pubsub): introduce streaming_pull RPC

dbolduc:
Part of the work for #3941 

Add the `streaming_pull` RPC to the subscriber surface. Aside: maybe it should be called `subscribe`... but let's review that later. I am OK with renaming things from a preview release of this client.

While we are here, normalize and fix some of the examples. The examples are repetitive, but I think that is close to what we want.

Note that this is not the final pass on documentation. But I want to get this in to unblock making the structs public and adding an integration test against production.

### DIFF
--- a/src/pubsub/src/subscriber/builder.rs
+++ b/src/pubsub/src/subscriber/builder.rs
@@ -12,7 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::session::Session;
 use super::transport::Transport;
+use crate::Result;
 use std::sync::Arc;
 
 const MIB: i64 = 1024 * 1024;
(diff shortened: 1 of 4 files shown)


suzmue (dismissed):

dbolduc:
argh, sorry, had to learn a lesson about version control.

@suzmue PTAL

src/pubsub/src/subscriber/builder.rs
```diff
@@ -37,6 +39,26 @@ impl StreamingPull {
         }
     }
 
+    /// Creates a new session to receive messages from the subscription.
+    ///
+    /// # Example
+    /// ```no_rust
```
coryan, line 45: Why `no_rust`?  I could understand `no_run`, maybe, but this is Rust code, no?

dbolduc: None of the types are public outside the crate. The next PR is the `feat:` PR that makes them public and removes all the `no_rust` from these examples. :slightly_smiling_face:

PR #4227, opened 2026-01-09
feat(pubsub): add `Subscriber`

dbolduc:
Part of the work for #3941 

Add the `Subscriber` client, which can receive messages from Cloud Pub/Sub.

An integration test against production is coming in the next PR.

Add a little coverage to the docs. They are still scant. This is not the final pass on them. Docs look like this: [docs.zip](https://github.com/user-attachments/files/24535978/docs.zip). (download, unzip, and open `doc/google_cloud_pubsub/index.html`)

### DIFF
--- a/src/pubsub/src/subscriber/session.rs
+++ b/src/pubsub/src/subscriber/session.rs
@@ -35,7 +35,7 @@ use tokio_util::sync::{CancellationToken, DropGuard};
 /// This is a stream-like struct for serving messages to an application.
 ///
 /// # Example
-/// ```no_rust
+/// ```
 /// # use google_cloud_pubsub::client::Subscriber;
 /// # async fn sample(client: Subscriber) -> anyhow::Result<()> {
 /// let mut session = client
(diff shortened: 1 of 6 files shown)


src/pubsub/src/subscriber/session.rs
```diff
@@ -35,7 +35,7 @@ use tokio_util::sync::{CancellationToken, DropGuard};
 /// This is a stream-like struct for serving messages to an application.
 ///
 /// # Example
-/// ```no_rust
+/// ```
 /// # use google_cloud_pubsub::client::Subscriber;
 /// # async fn sample(client: Subscriber) -> anyhow::Result<()> {
```
coryan, line 40: This context is useful, if I am looking at the documentation for `Subscriber` I may not know what kind of `client` you have hidden in the `# ` comments.

coryan (approved):
Good enough for a preview. Could be better.

src/pubsub/src/subscriber/session.rs
dbolduc: Ack.

dbolduc:
> Good enough for a preview. Could be better.

Agreed. I am missing a lot of practical context, and practical examples. I plan to make incremental progress on this.

PR #4229, opened 2026-01-10
test(pubsub): basic subscriber integration test

### LINKED ISSUE #4228: Basic subscriber integration test
We need a test using the subscriber against production. We can start with something simple like a roundtrip publishing N messages then receiving N messages.

dbolduc:
Fixes #4228 

Add an integration test that reads messages off of a subscription.

We need to seed these messages, so might as well hijack our basic publisher integration test and turn it into a roundtrip test.

### DIFF
--- a/tests/integration/tests/driver.rs
+++ b/tests/integration/tests/driver.rs
@@ -135,8 +135,13 @@ mod driver {
             .await
             .map_err(integration_tests::report_error)?;
 
-        pubsub_samples::cleanup_test_subscription(&sub_admin, sub.name).await?;
-        pubsub_samples::cleanup_test_topic(&topic_admin, topic.name).await
+        if let Err(e) = pubsub_samples::cleanup_test_subscription(&sub_admin, sub.name).await {
+            tracing::info!("Error cleaning up test subscription: {e:?}");
+        }
+        if let Err(e) = pubsub_samples::cleanup_test_topic(&topic_admin, topic.name).await {
+            tracing::info!("Error cleaning up test topic: {e:?}");
+        }
+        Ok(())
     }
 
     #[test_case(sm::client::SecretManagerService::builder(); "default")]


src/pubsub/examples/src/lib.rs
```diff
@@ -81,21 +78,16 @@ pub async fn cleanup_stale_topics(client: &TopicAdmin, project_id: &str) -> anyh
...
         {
             let client = client.clone();
-            let name = topic.name.clone();
-            pending.push(tokio::spawn(async move {
-                cleanup_test_topic(&client, name).await
-            }));
-            names.push(topic.name);
+            pending.spawn(async move {
+                let name = topic.name.clone();
+                (topic.name, cleanup_test_topic(&client, name).await)
```
coryan, line 83: Interesting way to avoid the `.zip()`.

tests/integration/tests/driver.rs
```diff
@@ -123,11 +123,22 @@ mod driver {
...
+        integration_tests::pubsub::basic_publisher(topic.name.clone())
             .await
-            .map_err(integration_tests::report_error)
+            .map_err(integration_tests::report_error)?;
+        integration_tests::pubsub::basic_subscriber(sub.name.clone())
+            .await
+            .map_err(integration_tests::report_error)?;
+
+        let _ = pubsub_samples::cleanup_test_topic(&topic_admin, topic.name).await?;
+        let _ = pubsub_samples::cleanup_test_subscription(&sub_admin, sub.name).await?;
```
coryan: Maybe the opposite order and some error reporting?


```suggestion
        if let Err(e) = pubsub_samples::cleanup_test_subscription(&sub_admin, sub.name).await {
            println!("Error cleaning up test subscription: {e:?}");
        }
        if let Err(e) = pubsub_samples::cleanup_test_topic(&topic_admin, topic.name).await {
            println!("Error cleaning up test topic: {e:?}");
        }
```

dbolduc: Done

PR #4226, opened 2026-01-09
chore: update librarian version to commit c02145f

ldetmer:
This includes fix for creating release PR and including the copyright in the librarian.yaml header.

PR #4230, opened 2026-01-10
fix: Sample generation flags are only overwritten when necessary

amanda-tarafa:
This is the counterpart to https://github.com/googleapis/librarian/pull/3518, after which we only need to overwrite sample generation flags that are different to defaults.

Generating with these changes in results in no changes on generated code, as expected.


amanda-tarafa:
Note the generation failure is expected because some of the flag instances changed from `bool` to `string` in https://github.com/googleapis/librarian/pull/3518, so until that is merged and we update the librarian version here, we'll see the following error with this new `librarian.yaml`:

```
ERROR librarian command failed err="yaml: unmarshal errors:\n  line 132: cannot unmarshal !!str `false` into bool\n  line 1159: cannot unmarshal !!str `false` into bool""
```

amanda-tarafa:
Closes #4226 beacause it updates librarian to a more recent commit.

amanda-tarafa:
This is now ready for review.

Issue #4207, opened 2026-01-09
idtoken: allow configuring crypto backend


howardjohn:
**Is your feature request related to a problem? Please describe.**
Enabling the `idtoken` feature of google_cloud_auth unconditionally pulls in `jsonwebtoken` with the `rust_crypto` feature. `jsonwebtoken` has two features, `rust_crypto` and `aws_lc_rust`; they MUST have only one set otherwise compilation fails.

For various reasons, I am already using `jsonwebtoken` with `aws_lc_rust`. As such, I would like to configure `google_cloud_auth` to also use the `aws_lc_rust` feature so that it can compile.

**Describe the solution you'd like**
I would like to configure `google_cloud_auth` to also use the `aws_lc_rust` feature so that it can compile.

**Describe alternatives you've considered**
use `rust_crypto` which is not compliant with my security requirements.

**Additional context**

houseme:
https://github.com/googleapis/google-cloud-rust/pull/4182

coryan:
I am breaking down the solution into multiple steps:

-  #4231 allows applications that *only* use `google-cloud-auth` (and no other `google-cloud-*` crate) to pick a different `jsonwebtoken` backend.
- It should be possible to use `google-cloud-gax-internal`  without enabling `*-auth`'s default features.
- It should be possible to use `google-cloud-storage`  without enabling `*-auth`'s default features.
- It should be possible to use `google-cloud-*`  (all the generated libraries) without enabling `*-auth`'s default features.

#4182 is about prototyping a solution for the second and third steps.

coryan:
Closing. All the changes to prune either rust crypto and/or aws-lc-rs dependencies are in.

howardjohn:
thank you!!

PR #4231, opened 2026-01-10
feat(auth): make `jsonwebtoken` backend optional

coryan:
With this change applications that only use `google-cloud-auth` can select a different backend for `jsonwebtoken, the crate `google-cloud-auth` uses to verify OIDC ID Tokens.

Part of the work for #4207

### DIFF
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -66,8 +66,6 @@ url.workspace         = true
 mutants.workspace     = true
 
 [features]
-# `default-idtoken-backend` is enabled by default, but has no effect unless
-# `idtoken` is also enabled. It
 default = ["default-idtoken-backend"]
 # The `idtoken` feature enables support to create and validate OIDC ID Tokens.
 # See the create top-level documentation for more information.


tests/crypto-providers/idtoken-with-rust-crypto/src/main.rs
```diff
@@ -0,0 +1,23 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use google_cloud_auth::credentials::idtoken::verifier::Builder;
+
+fn main() -> anyhow::Result<()> {
+    test_metadata::idtoken_has_rust_crypto_backend()?;
```
dbolduc, line 18: Q: should we also check that `test_metadata::idtoken_has_aws_lc_rs_backend().is_err()` ?

Because features are additive. Seems like if we always used `aws_lc_rs` and `rust_crypto`, these tests would pass. But that is not what we want.

tests/crypto-providers/idtoken-with-default/src/main.rs
```diff
@@ -0,0 +1,23 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use google_cloud_auth::credentials::idtoken::verifier::Builder;
+
+fn main() -> anyhow::Result<()> {
+    test_metadata::idtoken_has_default_backend()?;
+
+    // Verify the google-cloud-auth library is usable too.
+    let _ = Builder::new(["https://my-service.a.run.app"]).build();
```
dbolduc: what is this verifying? we are not looking at the result.

src/auth/Cargo.toml
```diff
@@ -67,8 +67,16 @@ url.workspace         = true
...
+default = ["default-idtoken-backend"]
+# The `idtoken` feature enables support to create and validate OIDC ID Tokens.
+# See the create top-level documentation for more information.
+idtoken = ["dep:jsonwebtoken"]
+# By default this crate enables the `rust_crypto` backend. Applications can
+# link `google-cloud-auth` with `default-features = false, feature = ["idtoken"]
+# to select the backend themselves.
+default-idtoken-backend = ["jsonwebtoken?/rust_crypto"]
+# Do not use, this was a mistake in the 1.3 release.
+jsonwebtoken = ["dep:jsonwebtoken"]
```
dbolduc, line 78: I am confused...

src/auth/README.md
```diff
@@ -14,9 +14,24 @@ also describes the common terminology used with authentication, such as
...
+
+- `idtoken`: disabled by default, this feature enables support to create and
+  verify [OIDC ID Tokens]. This feature depends on the [jsonwebtoken] crate.
+- `default-jsonwebtoken-backend`: enabled by default, this feature enables a
+  default backend for the `jsonwebtoken` crate. Currently the default is
+  `rust_crypto`, but we may change the default backend at any time, applications
+  that have specific needs for this backend should not rely on the current
+  default. To control the backend selection:
+  - Configure this crate with `default-features = false`, and
+    `features = ["jsonwebtoken"]`
```
dbolduc: This should be `idtoken`, right?

src/auth/Cargo.toml
dbolduc: Ah, ok, I see. This was an implicit feature flag because we did not use `dep:`.

src/auth/src/lib.rs
```diff
@@ -24,6 +24,22 @@
...
+//! - `idtoken`: disabled by default, this feature enables support to create
+//!   and verify [OIDC ID Tokens]. This feature depends on the [jsonwebtoken]
+//!   crate.
+//! - `default-jsonwebtoken-backend`: enabled by default, this feature enables
+//!   a default backend for the `jsonwebtoken` crate. Currently the default is
+//!   `rust_crypto`, but we may change the default backend at any time,
+//!   applications that have specific needs for this backend should not rely
+//!   on the current default. To control the backend selection:
+//!   - Configure this crate with `default-features = false`, and
+//!     `features = ["jsonwebtoken"]`
```
dbolduc: same, this should be `idtoken`

tests/crypto-providers/idtoken-with-default/src/main.rs
alvarowolfx: When the jsonwebtoken doesn't have any crypto backend, it fails to compile. I think that's what is being check here if I understood correctly

src/auth/Cargo.toml
alvarowolfx: oh no, sorry for this. I missed the dep: thing 😢

src/auth/src/lib.rs
coryan: Fixed.

src/auth/README.md
coryan: yes. fixed.

tests/crypto-providers/idtoken-with-default/src/main.rs
coryan: What Alvaro said. Improved the comments here and elsewhere.

tests/crypto-providers/idtoken-with-rust-crypto/src/main.rs
coryan: `jsonwebtoken` is weird in that enabling both features results in a compile-time error.  I have documented this in the `idtoken_has_*()` functions.

tests/crypto-providers/idtoken-with-default/src/main.rs
dbolduc: Ah thanks for the context. Seems strange though..

tests/crypto-providers/test-metadata/src/lib.rs
```diff
@@ -43,6 +43,44 @@ pub fn no_default_crypto_provider() -> anyhow::Result<()> {
...
+/// This function returns an error if the jsonwebtoken crate is not configured
+/// with the `rust_crypto` backend.
+pub fn idtoken_has_rust_crypto_backend() -> anyhow::Result<()> {
+    let metadata = metadata()?;
+    let features = find_jsonwebtoken_features(&metadata)?;
+    if !features.contains(&FeatureName::new("rust_crypto".to_string())) {
+        bail!("jsonwebtoken should have rust_crypto enabled: {features:?}")
+    }
+    if features.contains(&FeatureName::new("aws_lc_rs".to_string())) {
+        // The jsonwebtoken library would not compile if both features are
```
dbolduc, line 61: comment: (not our crate, but) this is not how features are supposed to work. They are supposed to be additive. Oh well.

coryan:
Rebased and resolved conflicts, PTAL.

Issue #4222, opened 2026-01-09
Subscriber needs to set client_id field


dbolduc:
The client ID is unique to a **client**. Pub/Sub team confirms that multiple sessions for the same client get the same `client_id` (even for different subscriptions).

We need to generate a string (I will just use a uuid) and pass it from the client into the `StreamingPull` builder into the initial request.

PR #4233, opened 2026-01-10
refactor(pubsub): use client in unit tests

dbolduc:
Motivated by #4222 

A pure unit test refactor. Use a client directly instead of going through the builder. (We didn't have a `streaming_pull` method on the client when these tests were written).

I need to change the `StreamingPull::new()` to accept a client ID. All of this code would change.

### DIFF
--- a/src/pubsub/src/subscriber/session.rs
+++ b/src/pubsub/src/subscriber/session.rs
@@ -202,13 +202,13 @@ impl Session {
 
 #[cfg(test)]
 mod tests {
+    use super::super::client::Subscriber;
     use super::super::keepalive::KEEPALIVE_PERIOD;
     use super::super::lease_state::tests::{test_id, test_ids};
-    use super::super::transport::tests::test_transport;
     use super::*;
+    use auth::credentials::anonymous::Builder as Anonymous;
     use pubsub_grpc_mock::google::pubsub::v1;
     use pubsub_grpc_mock::{MockSubscriber, start};
-    use std::sync::Arc;
     use tokio::sync::mpsc::{channel, unbounded_channel};
     use tokio::task::JoinHandle;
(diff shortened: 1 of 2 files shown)


src/pubsub/src/subscriber/session.rs
```diff
@@ -202,13 +202,13 @@ impl Session {
 
 #[cfg(test)]
 mod tests {
+    use super::super::client::Subscriber;
```
coryan, line 205: At some point `crate::` is easier to spell 🤷

src/pubsub/src/subscriber/session.rs
```diff
@@ -286,15 +292,15 @@ mod tests {
...
-        let builder = StreamingPull::new(
-            Arc::new(transport),
-            "projects/p/subscriptions/s".to_string(),
-        )
-        .set_ack_deadline_seconds(20)
-        .set_max_outstanding_messages(2000)
-        .set_max_outstanding_bytes(200 * MIB);
-        let _ = Session::new(builder).await;
+        let client = test_client(endpoint).await?;
+        let _ = client
```
coryan, line 296: You don't want to say something about the type of error?

dbolduc: Not in this test, which is narrowly focused on the contents of the initial request.

We have an `error_starting_stream()` test which verifies the error:

https://github.com/googleapis/google-cloud-rust/blob/8e8de8c167d92a2fd6abe7f9a44a6cfdf0c6f760/src/pubsub/src/subscriber/session.rs#L253-L259

src/pubsub/src/subscriber/session.rs
dbolduc: Ack, lol.

PR #4242, opened 2026-01-11
impl(pubsub): identify subscribers

### LINKED ISSUE #4222: Subscriber needs to set client_id field

dbolduc:
Fixes #4222 

This identifier helps the server recover state on retries and make routing decisions.

### DIFF
--- a/src/pubsub/src/subscriber/session.rs
+++ b/src/pubsub/src/subscriber/session.rs
@@ -102,6 +102,7 @@ impl Session {
             stream_ack_deadline_seconds: builder.ack_deadline_seconds,
             max_outstanding_messages: builder.max_outstanding_messages,
             max_outstanding_bytes: builder.max_outstanding_bytes,
+            client_id: builder.client_id,
             ..Default::default()
         };
         let (stream, request_tx) = open_stream(inner, initial_req).await?;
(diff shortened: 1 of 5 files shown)


src/pubsub/src/subscriber/session.rs
```diff
@@ -713,4 +715,70 @@ mod tests {
...
+            .expect("should receive a request")?;
+        let _ = c1
+            .streaming_pull("projects/p/subscriptions/s")
+            .start()
+            .await;
+        let req2 = recover_writes_rx
+            .recv()
+            .await
+            .expect("should receive a request")?;
+        assert_eq!(req1.client_id, req2.client_id);
```
coryan, line 767: All our tests require a lot of boilerplate for just one or two assertion lines.  Sigh.

dbolduc: Yeah... :shrug:

The tests all set up a server and at least one client. There is a way to make that shorter.

Some streams succeed, some fail. Sometimes I care about which acks were sent. Other times I care about which `StreamingPullRequest`s are sent.

I am not sure trying to unify every case is the answer. I'll think about it, though.

Issue #4211, opened 2026-01-09
Remove expired messages from lease management in subscriber


dbolduc:
When a message is expired, we should remove it from lease management.

We should have a maximum deadline extension, where when this is reached, we nack the message.

This will take a refactor, as we only pass around the ack ID. We will also need to keep an `Instant` of when we received the message.

(Or you launch a task to nack each message after the deadline when they are received, and its a no-op if they have already been handled. I think I like this less.)

PR #4243, opened 2026-01-11
refactor(pubsub): prepare for lease expirations

dbolduc:
Part of the work for #4211

Keep track of when a message was accepted under lease management.

The next PR will use this information to decide if a message should be dropped from lease management, because it has been held for too long.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -14,6 +14,7 @@
 
 use super::leaser::Leaser;
 use std::collections::HashMap;
+// Use a `tokio::time::Instant` to facilitate time-based unit testing.
 use tokio::time::{Duration, Instant, Interval, interval_at};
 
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -49,8 +49,8 @@ pub(super) struct LeaseState<L>
 where
     L: Leaser,
 {
-    // TODO(#3957) - support message expiry
-    under_lease: HashSet<String>,
+    // A map of ack IDs to the time they were first received.
+    under_lease: HashMap<String, Instant>,
```
coryan, line 54: Note to self: this is a `tokio::time::Instant`, so the tests are  deterministic (or more so?).

coryan: Maybe add a comment for "future us"?

dbolduc: Done.

coryan (approved):

PR #4245, opened 2026-01-11
docs: Samples for simple RPCs including get and delete

amanda-tarafa:
This includes the baseline sample for all simple RPCs (non-LRO and non-paginated) plus specific samples for get and non-LRO deletes.
Its latest counterpart is https://github.com/googleapis/librarian/pull/3519

Best to review one commit at a time.

### DIFF
--- a/src/generated/cloud/secretmanager/v1/src/client.rs
+++ b/src/generated/cloud/secretmanager/v1/src/client.rs
@@ -146,9 +146,10 @@ impl SecretManagerService {
     /// # Example
     /// ```
     /// # use google_cloud_secretmanager_v1::client::SecretManagerService;
+    /// use google_cloud_secretmanager_v1::Result;
     /// async fn sample(
     ///    client: &SecretManagerService
-    /// ) -> gax::Result<()> {
+    /// ) -> Result<()> {
     ///     let response = client
     ///         .create_secret()
     ///         /* set fields */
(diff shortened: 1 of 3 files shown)


amanda-tarafa:
This is ready for review, one commit at a time is best.

src/generated/cloud/secretmanager/v1/src/client.rs
```diff
@@ -152,13 +169,48 @@ impl SecretManagerService {
...
+    /// ```
+    /// # use google_cloud_secretmanager_v1::client::SecretManagerService;
+    /// use google_cloud_secretmanager_v1::Result;
+    /// async fn sample(
+    ///    client: &SecretManagerService,
+    ///    resource_name: &str
+    /// ) -> Result<()> {
+    ///     let response = client
+    ///         .get_secret()
+    ///         .set_name(resource_name)
```
coryan, line 207: Closer. I look forward to some information about `resource_name`'s format.

dbolduc: That would be nice.

As-is, if the customer tries something incorrect, they will get an informative binding error. That is not a bad worst case.

amanda-tarafa: Yes, we are tracking that separately and canbe done for all sample types at once, at the end.

PR #4248, opened 2026-01-12
impl(pubsub): message expiration

### LINKED ISSUE #4211: Remove expired messages from lease management in subscriber

dbolduc:
Fixes #4211 

Add an internal max_lease_extension setting, defaulted to 10 minutes. When a message has been held for this long, we stop extending its lease.

The math is a little loose, but given that this is an internal setting, I will not be upset if we hold something for 10 minutes and 10 seconds.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -32,6 +32,9 @@ pub(super) struct LeaseOptions {
     pub(super) extend_period: Duration,
     /// How long we wait for the initial extensions
     pub(super) extend_start: Duration,
+    /// How long messages can be kept under lease. A message's lease can be
+    /// extended as long as `max_lease_extension` has not elapsed.
+    pub(super) max_lease_extension: Duration,
 }
 
 impl Default for LeaseOptions {
@@ -41,6 +44,7 @@ impl Default for LeaseOptions {
             flush_start: Duration::from_secs(1),
             extend_period: Duration::from_secs(3),
             extend_start: Duration::from_millis(500),
+            max_lease_extension: Duration::from_secs(600),
         }
     }
 }
@@ -61,6 +65,8 @@ where
     flush_interval: Interval,
     // A timer for extending leases
     extend_interval: Interval,
+    // How long messages can be kept under lease
+    max_lease_extension: Duration,
 }
 
 /// Actions taken by the `LeaseState` in the lease loop.
@@ -88,6 +94,7 @@ where
             leaser,
             flush_interval,
             extend_interval,
+            max_lease_extension: options.max_lease_extension,
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -152,11 +159,31 @@ where
     ///
     /// Drops messages whose lease deadline cannot be extended any further.
     pub(super) async fn extend(&mut self) {
```
dbolduc, line 161: Note to reviewer: I also considered this:

```rs
    pub(super) async fn extend(&mut self) {
        let now = Instant::now();
        let mut to_extend = Vec::new();
        self.under_lease.retain(|ack_id, receive_time| {
            // Note that using `HashMap::retain` allows us to iterate over the
            // map and conditionally drop elements in one pass.

            if *receive_time + self.max_lease_extension < now {
                // Drop messages that have been held for too long.
                false
            } else {
                // Extend leases for all other messages.
                to_extend.push(ack_id.clone());
                true
            }
        });

        // Break the list of ack IDs into chunks.
        let mut iter = to_extend.into_iter();
        loop {
            // Note that using `slice::chunks()` would force us to clone the
            // strings twice, as it returns reference slices (`&[String]`). That
            // is why we choose to write the chunking algorithm ourselves.
            let ack_ids: Vec<String> = iter.by_ref().take(ACK_IDS_PER_RPC).collect();
            if ack_ids.is_empty() {
                break;
            }
            // TODO(#3975) - send RPCs concurrently
            self.leaser.extend(ack_ids).await;
        }
    }
```

But I think what I have will evolve better. I think when #3975 is done, `batches` goes away.

coryan: I am just going to leave this here:

https://stackoverflow.com/questions/74920317/how-can-i-drain-a-vector-in-chunks

dbolduc: I saw [`itertools::Itertools::chunks`](https://docs.rs/itertools/latest/itertools/trait.Itertools.html#method.chunks) but that looked like it was going to do an extra clone of the keys.

One funny thing the bot suggested was `futures::stream::iter(v)` + `futures::StreamExt::chunks`, then just iterate over the chunks (asyncly, but they are available immediately). I balked at doing anything `async`, even if everything would be ready now...

Lemme merge, because I am not sure what the answer looks like and this is the last thing I really wanted in the release.

PR #4249, opened 2026-01-12
impl(pubsub): cloneable `Leaser`

dbolduc:
Part of the work for #3975 

To run lease operations in a background task, we will need to clone the leaser into the background task.

We do not have `trait Leaser : Clone`, because it complicates mocking. Instead, our tests are written in terms of an `Arc<MockLeaser>`.

Note that the production code (`DefaultLeaser`) does not create an additional `Arc`. It does hold its client internally as an `Arc`, though.


dbolduc:
Had to rebase, which was kinda lucky, because if not, I would have merged broken code.

PR #4256, opened 2026-01-12
feat(auth): add retry to mds idtoken provider

alvarowolfx:
Towards #3592


src/auth/src/credentials/idtoken/mds.rs
```diff
@@ -319,6 +400,70 @@ mod tests {
 
     type TestResult = anyhow::Result<()>;
 
+    #[tokio::test]
+    #[parallel]
```
dbolduc: nit: parallel is the default right? we don't need to specify it?

src/auth/src/retry.rs
```diff
@@ -40,6 +41,18 @@ pub(crate) struct Builder {
...
 
+/// This is necessary because adding [Builder] to existing, released auth features
+/// caused a breaking change. The containing structs would lose their automatically derived
+/// `RefUnwindSafe` implementation because the dynamic trait objects (like `dyn RetryPolicy`)
+/// are not `RefUnwindSafe` nor `UnwindSafe` by default.
+///
+/// This solves that by manually implementing `RefUnwindSafe` and `UnwindSafe`, allowing us to add
+/// retry functionality without triggering a semver-check failure. This is safe because
+/// we control the implementation of the inner types and can ensure that they are
+/// `RefUnwindSafe` and `UnwindSafe`.
```
coryan: I think this is an implementation detail comment, and not something that needs to go to the public API docs?

```suggestion
// This is necessary because adding [Builder] to existing, released auth features
// caused a breaking change. The containing structs would lose their automatically derived
// `RefUnwindSafe` implementation because the dynamic trait objects (like `dyn RetryPolicy`)
// are not `RefUnwindSafe` nor `UnwindSafe` by default.
//
// This solves that by manually implementing `RefUnwindSafe` and `UnwindSafe`, allowing us to add
// retry functionality without triggering a semver-check failure.
//
// This is safe because we control the implementation of the inner types and can ensure that they are
// `RefUnwindSafe` and `UnwindSafe`.
```

coryan: This part is true-ish today, but may not be in the future.  We are planning to make the retry policy traits public and implemented by customers.

I think more convincing is:

> In the builder, we never call or use the retry policies. We just hold them until we create the client. There is no opportunity for a panic to leave them in an inconsistent state.

PR #4257, opened 2026-01-12
impl(auth): add retry to impersonated idtoken provider

alvarowolfx:
Towards #3592


alvarowolfx:
I don't know why the diff is looking weird here on Github. But I removed the `test_impersonated_id_token_fail` test as with retries, is not gonna hard fail in one call.

dbolduc (approved):

PR #4258, opened 2026-01-12
chore: update googleapis SHA circa 2026-01-12

dbolduc:


librarian.yaml
```diff
@@ -123,13 +123,13 @@ libraries:
...
       modules:
         - disabled_rustdoc_warnings: []
+          generate_setter_samples: "false"
           module_path: crate::generated
           module_roots:
             project-root: .
           output: src/wkt/tests/common/src/generated
           source: src/wkt/tests/protos
           template: mod
-          generate_setter_samples: "false"
```
dbolduc, line 132: wasn't me, btw.

coryan: The new structure is sorting the fields. I guess librarian sorts the fields, but our formatter does not, come to think about it, this file is skipped by the formatter. @amanda-tarafa edited the file by hand (as she should have), but not in the same order that `librarian` uses.

Maybe @zhumin8 @JoeWang1127 or @ldetmer have suggestions on how to avoid these problems in the future.

coryan: Let's continue the discussion in #4259

dbolduc: Aside: we also have a `Cargo.lock` problem.

coryan: > Aside: we also have a `Cargo.lock` problem.

That may precede `librarian`.  It only breaks when a GAPIC gets new dependencies (e.g. gaining a LRO).  We should add a `cargo update --workspace` to the commands that librarian runs (or to our instructions).

PR #4274, opened 2026-01-13
chore: add rust storage control

JoeWang1127:
After resolving https://github.com/googleapis/librarian/issues/3532, add the config to generate rust storage control client.

With the updated librarian.yaml, to regenerate rust storage control client:
```
librarian generate google-cloud-storage
```

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -1434,6 +1434,8 @@ libraries:
         - language: rust_storage
           output: src/storage/src/control/generated
           specification_format: none
+          source: ""
+          template: ""
     veneer: true
   - name: google-cloud-storagebatchoperations-v1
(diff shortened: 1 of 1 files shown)


src/storage/src/control/generated/client.rs
```diff
@@ -19,11 +19,11 @@ use crate::control::client::*;
 impl StorageControl {
     /// Returns a builder for [StorageControl].
     ///
-    /// ```no_run
-    /// # tokio_test::block_on(async {
+    /// ```
+    /// # async fn sample() -> anyhow::Result<()> {
```
coryan, line 23: Note to self: we missed at least one update, but it is not a big deal.

PR #4278, opened 2026-01-14
chore: add protojson-conformance

JoeWang1127:
After resolving https://github.com/googleapis/librarian/issues/3533, add the config to generate  protojson-conformance client.

With the updated librarian.yaml, to regenerate protojson-conformance client:
```
librarian generate protojson-conformance
```


librarian.yaml
```diff
@@ -1566,6 +1566,36 @@ libraries:
...
+        - generate_setter_samples: "false"
+          include_list: conformance.proto
+          module_path: crate::generated::gapic::model
+          output: tests/protojson-conformance/src/generated/gapic
+          source: conformance
+          template: grpc-client
+        - include_list: conformance.proto
+          module_path: crate::generated::gapic::model
+          output: tests/protojson-conformance/src/generated/protos
+          root_name: conformance-root
```
coryan, line 1588: It is a bit odd that only this thing requires `root_name`.  I assume there are good reasons 🤷

JoeWang1127: The root name is hard coded as `googleapis-root`:
https://github.com/googleapis/librarian/blob/d798500afbbd6a351683346431d518e32a782f96/internal/sidekick/rust_prost/codec.go#L43

I have to use this field to override the default value.

coryan: Yes, I am not surprised you need to override the default value.  I am surprised that only one module needs the override.  Why don't we need that for the other modules here?

amanda-tarafa: +1 To @coryan's comment. I just had to manually copy `wkt/src/tests/protos` into the my local `googleapis@...` folder for getting generation to work. I think it's the same thing as this.

Issue #4282, opened 2026-01-15
Simplify `google-cloud-pubsub` client creation


suzmue:
To simplify client creation for the publisher, we want to allow users to be able to create a Publisher directly without creating a Client first. 

```rust
let publisher = pubsub::client::Publisher::builder(topic_name).build().await?;
// ready to start publishing
```

We will accomplish this by
1. Renaming Client -> BasePublisher
2. Renaming PublisherBuilder -> PublisherPartialBuilder
3. Adding a new builder() to Publisher that configures client and batching configuration.
4. Moving BasePublisher to a publisher module as an advanced use case.

suzmue:
This is complete, but we have decided not to do step 4 and will leave the BasePublisher in crate::client::*.

PR #4283, opened 2026-01-15
refactor(pubsub)!: Rename Client to BasePublisher

suzmue:
This is the first step in simplifying the Publisher client creation and renames the Client and associated builders.

For #4282


coryan (dismissed):
SGTM. I think we can bikeshed the names once we are closer to "not preview".

suzmue:
> `refactor(pubsub)!: ...` ?
> 
> I think we would include this change in our release notes.

Right it is released. Updated.

PR #4296, opened 2026-01-16
fix: force http1 only

PhongChuong:
Force reqwest to use http1 only as we currently do not support HTTP2 in our clients.

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -58,6 +58,7 @@ impl ReqwestClient {
         let cred = Self::make_credentials(&config).await?;
         let mut builder = reqwest::Client::builder();
         // Force http1 as http2 with not currently supported.
+        // TODO(#4298): Remove after adding HTTP2 support.
         builder = builder.http1_only();
         if config.disable_automatic_decompression {
             builder = builder.no_gzip().no_brotli().no_deflate();


src/gax-internal/src/http.rs
```diff
@@ -57,6 +57,8 @@ impl ReqwestClient {
     ) -> gax::client_builder::Result<Self> {
         let cred = Self::make_credentials(&config).await?;
         let mut builder = reqwest::Client::builder();
+        // Force http1 as http2 with not currently supported.
```
coryan, line 60: Please add a feature request to support HTTP/2. And consider referencing that issue here.

PhongChuong: Done. PTAL.

PhongChuong:
Note: I was able to manually test this with @dbolduc repro instructions. However, I couldn't find a good place to automate the testing since it requires adding the "http2" feature to reqwest in the cargo.toml. Any pointer in that regard is welcomed.

coryan:
> Note: I was able to manually test this with @dbolduc repro instructions. However, I couldn't find a good place to automate the testing since it requires adding the "http2" feature to reqwest in the cargo.toml. Any pointer in that regard is welcomed.

I think you would need a new integration test crate that enables that feature, and you would need to run that with `cargo test --package new-crate-name`.  Make sure the new crate is **not** in the `default-members` list.  If you try to add this to an existing crate all the tests in that crate would gain the feature. And if it is in the default-members list then any `cargo test` would include them and enable the feature for everybody else too.

PR #4299, opened 2026-01-16
test: iterate on crypto provider tests

coryan:
This brings the tests closer to my prototype that uses `reqwest-0.13` and can prune the crypto provider.

It makes two related changes:

1. Bring the `test-*` helper crates to the workspace, easier development and better guarantees on using the right versions.
1. Use `cargo tree` to discover what dependencies are definitely used. I no longer think it is possible to do this with `cargo-metadata`.

more work for #4170

### DIFF
--- a/tests/crypto-providers/gaxi-with-default/Cargo.toml
+++ b/tests/crypto-providers/gaxi-with-default/Cargo.toml
@@ -32,5 +32,5 @@ rust-version = "1.92.0"
 [dependencies]
 anyhow        = { default-features = false, version = "1", features = ["std"] }
 tokio         = { default-features = false, version = "1", features = ["test-util"] }
-test-gaxi     = { path = "../test-gaxi" }
+test-gaxi     = { path = "../test-gaxi", features = ["with-default"] }
 test-metadata = { path = "../test-metadata" }
--- a/tests/crypto-providers/test-gaxi/Cargo.toml
+++ b/tests/crypto-providers/test-gaxi/Cargo.toml
@@ -42,7 +42,14 @@ path             = "../../../src/gax"
 [dependencies.google-cloud-gax-internal]
 default-features = false
 path             = "../../../src/gax-internal"
-features         = ["_internal-http-client"]
+# Activate the minimal features to make it possible to use the HTTP client.
+# Soon, this client will not have a default crypto provider for TLS, and the
+# tests will detect that change.
+#
+# Separately, we have chosen to not test the gRPC client in gax-internal in
+# isolation. That would require a lot more infrastructure. The storage-with-*
+# tests will have to suffice.
+features = ["_internal-http-client"]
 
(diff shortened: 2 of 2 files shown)


tests/crypto-providers/gaxi-with-default/src/main.rs
```diff
@@ -14,6 +14,6 @@
 
 #[tokio::main]
 async fn main() -> anyhow::Result<()> {
-    test_metadata::has_default_crypto_provider()?;
+    test_metadata::has_default_crypto_provider(env!("CARGO"), env!("CARGO_MANIFEST_DIR"))?;
```
amanda-tarafa, line 17: I don't understand how this works, but it obviously does. In my mind there shouldn't be any default here because none of the internal gaxi features are being used.

tests/crypto-providers/gaxi-with-default/Cargo.toml
```diff
@@ -34,9 +34,3 @@ anyhow        = { default-features = false, version = "1", features = ["std"] }
 tokio         = { default-features = false, version = "1", features = ["test-util"] }
 test-gaxi     = { path = "../test-gaxi" }
```
amanda-tarafa: Should this one activate "with_default", I see is empty below, with a TODO, so it should work now.

tests/crypto-providers/gaxi-with-default/src/main.rs
amanda-tarafa: Ah, just saw test-gaxi enabling the internal HTTP feature.

amanda-tarafa (dismissed):

tests/crypto-providers/gaxi-with-default/src/main.rs
coryan: Included some comments in that file to explain what is going on.

tests/crypto-providers/gaxi-with-default/Cargo.toml
coryan: Done.

tests/crypto-providers/test-metadata/src/lib.rs
```diff
@@ -86,16 +125,34 @@ fn metadata() -> anyhow::Result<Metadata> {
     Ok(metadata)
 }
 
+fn reqwest_version(_metadata: &Metadata) -> Version {
+    // TODO(4170) - we would like to read this version from the top-level
```
dbolduc, line 129: super nit: `TODO(#4170)`

dbolduc: Seems like it will be short lived though. Feel free to merge as is

dbolduc (approved):

amanda-tarafa (approved):
Thanks!

PR #4312, opened 2026-01-20
docs(pubsub): subscriber observability note

dbolduc:


src/pubsub/src/subscriber/client.rs
```diff
@@ -59,8 +59,15 @@ use std::sync::Arc;
 /// an [Rc](std::rc::Rc) or [Arc] to reuse it, because it already uses an `Arc`
 /// internally.
 ///
+/// # Troubleshooting
+///
+/// The Subscriber is intentionally opaque. The best view into its performance
```
coryan: Hmmm... I am not sure "intentionally opaque" is what we want to go with here.  Maybe "At the moment, the subscriber is opaque" or "it is not possible to locally examine the performance of the subscriber.".

dbolduc: Done. Went with both of your suggestions. Thanks.

PR #4313, opened 2026-01-20
refactor(storage): use `gaxi::http::ReqwestClient`

### LINKED ISSUE #4308: Refactor `google-cloud-storage` to use `gaxi::ReqwestClient`

coryan:
Fixes #4308

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -61,7 +61,7 @@ impl ReqwestClient {
         // TODO(#4298): Remove after adding HTTP2 support.
         builder = builder.http1_only();
         if config.disable_automatic_decompression {
-            builder = builder.no_gzip().no_brotli().no_deflate();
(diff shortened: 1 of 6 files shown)


dbolduc (dismissed):
Nice

dbolduc:
Might want to call this a `feat` because it adds a lot of the features built into the HTTP client to GCS, e.g. respecting application's user agent. I think.

coryan:
> Might want to call this a `feat` because it adds a lot of the features built into the HTTP client to GCS, e.g. respecting application's user agent. I think.

I don't think those features get used here.  In this PR the storage client calls `ReqwestClient::builder*()` which gets back a `reqwest::RequestBuilder` and then the storage client adds headers and calls `send()`.  All the features (like user agent) are implemented in `execute()`.  We need some additional refactoring for them to have any effect.

dbolduc:
Ah thanks

dbolduc (approved):

PR #4314, opened 2026-01-20
test(storage): skip signed_urls test if creds do not support signing

alvarowolfx:


tests/integration/tests/driver.rs
```diff
@@ -218,13 +218,31 @@ mod driver {
...
+            Ok(_) => {}
+            Err(err) if err.is_not_supported() => {
+                tracing::warn!("skipping run_storage_signed_urls: {err:?}");
+                return Ok(());
+            }
+            Err(err) => {
+                anyhow::bail!("error creating signer: {err:?}")
+            }
+        };
+        let signer = signer?;
```
dbolduc: nit:

```rs
        let signer = match signer {
            Ok(s) => s,
            Err(err) if err.is_not_supported() => {
                tracing::warn!("skipping run_storage_signed_urls: {err:?}");
                return Ok(());
            }
            Err(err) => {
                anyhow::bail!("error creating signer: {err:?}")
            }
        };
```

coryan: Consider:

```suggestion
        let signer = match signer {
            Ok(s) => s,
            Err(err) if err.is_not_supported() => {
                tracing::warn!("skipping run_storage_signed_urls because the default credentials do not support signing: {err:?}");
                return Ok(());
            }
            Err(err) => return Err(err),
        };
```

coryan (dismissed):

dbolduc (dismissed):

tests/integration/tests/driver.rs
alvarowolfx: oh, I thought `return Ok(())` was gonna return the value for the `match` statement. I though initially on doing this and ended up not even trying.

coryan (approved):

PR #4322, opened 2026-01-21
chore: update rust config

JoeWang1127:
After resolving https://github.com/googleapis/librarian/issues/3620, update librarian version and librarian.yaml


coryan:
Should this wait for #4321 ?

julieqiu:
Yes - it actually will be on a later version this commit was merged afterwards.

julieqiu:
Actually it looks like the version is updated to `v0.8.1-0.20260121165046-0665a32ba0ec` in this PR (I didn't look closely before), so it should be fine as is.

PR #4323, opened 2026-01-21
ci: move integration builds to private pool

coryan:


.gcb/integration.yaml
```diff
@@ -88,3 +88,5 @@ substitutions:
   _RUST_VERSION: '1.92'
   _SCCACHE_VERSION: 'v0.12.0'
   _SCCACHE_SHA256: 'b0e89ead6899224a4ba2b90e9073bf1ce036d95bab30f3dc33c1e1468bc4ad44'
+  _POOL_REGION: us-central1
+  _POOL_ID: rust-sdk-pool
```
dbolduc, line 92: Neat. Should I expect a `terraform` PR soon?

dbolduc (approved):

.gcb/integration.yaml
coryan: I was not planning to. AFAICT, we don't set these in any of the other terraform scripts or triggers, they are in the yaml files purely in case you want to call them with `gcloud builds submit`.

dbolduc: How did you create the pool?

coryan: Ah, I am using the pool I created for the other builds / triggers.  I guess we should add a dependency or something, yuck.

dbolduc: https://registry.terraform.io/providers/hashicorp/google/latest/docs/resources/cloudbuild_worker_pool

coryan: Yes, the pool is already created by terraform:

https://github.com/googleapis/google-cloud-rust/blob/b6edf25e99c74133963f465a34d7e684eb361c39/.gcb/builds/resources/main.tf#L86-L94

And the dependency was already there (I got lucky):

https://github.com/googleapis/google-cloud-rust/blob/b6edf25e99c74133963f465a34d7e684eb361c39/.gcb/builds/main.tf#L54

Recall that all triggers are created by that module, including the integration build triggers:

https://github.com/googleapis/google-cloud-rust/blob/b6edf25e99c74133963f465a34d7e684eb361c39/.gcb/builds/triggers/main.tf#L70-L72

dbolduc: >Yes, the pool is already created by terraform:

:monkey: I `git grep rust-sdk-pool` from within `/src/pubsub/` and saw nothing. Sorry.

PR #4330, opened 2026-01-21
docs: update version bump instructions

### LINKED ISSUE #4326: docs: simplify release step
I think you can simplify this further because `bump --all` already calls `cargo fmt`:

```suggestion
go run github.com/googleapis/librarian/cmd/librarian@${V} bump --all
git commit -m"chore: bump version numbers circa $(date +%Y-%m-%d)" .
```

_Originally posted by @coryan in https://github.com/googleapis/google-cloud-rust/pull/4319#discussion_r2713438304_

coryan:
Fixes #4326

### DIFF
--- a/doc/contributor/howto-guide-generated-code-maintenance.md
+++ b/doc/contributor/howto-guide-generated-code-maintenance.md
@@ -93,15 +93,27 @@ accept your certs.
 
 ### Update top-level Cargo.toml
 
-If the version bumps include any of:
+Sometimes the top-level `Cargo.toml` file may need manual updates. Changes to
+GAPICs never trigger the need for such updates, or at least have never triggered
+such need. But changes to core crates, including `google-cloud-auth`,
+`google-cloud-gax`, `google-cloud-lro`, and `google-cloud-gax-internal` may
+trigger the need for manual updates.
 
-- `src/auth/Cargo.toml`
-- `src/gax/Cargo.toml`
-- `src/gax-internal/Cargo.toml`
-- `src/lro/Cargo.toml`
+You can rely on the CI build to detect when such an update is needed, or you
+could run some local tests first:
 
-You need to manually bump the corresponding versions in the top-level
-`Cargo.toml` file.
+```bash
+cargo install --locked cargo-minimal-versions
+# Make sure a veneer works.
+cargo minimal-versions check --package google-cloud-storage
+# Make sure a GAPIC (with the locations and iam mixins) works:
+cargo minimal-versions check --package google-cloud-secretmanager-v1
+# Make sure a GAPIC (with the longrunning mixin) works:
+cargo minimal-versions check --package google-cloud-workflows-v1
+```
+
+Alternatively, consider
+[this feature request](https://github.com/googleapis/librarian/issues/3720).
 
 ## Refreshing the code
 


doc/contributor/howto-guide-generated-code-maintenance.md
```diff
@@ -74,27 +74,33 @@ once. Note that this includes `showcase` and `protojson-conformance`, though.
...
 git commit -m"chore: bump version numbers circa $(date +%Y-%m-%d)"
 ```
 
 When running on Cloudtop, you might need to set
 `CARGO_HTTP_CAINFO=/etc/ssl/certs/ca-certificates.crt` in order for crates.io to
 accept your certs.
 
+### Update top-level Cargo.toml
+
+If the version bumps include any of:
```
dbolduc: Q: can't this happen with any non-leaf crate? e.g. `google-cloud-longrunning` recently required a top level `Cargo.toml` update

The version bump tool does not handle that, right?

doc/contributor/howto-guide-generated-code-maintenance.md
```diff
@@ -74,27 +74,33 @@ once. Note that this includes `showcase` and `protojson-conformance`, though.
...
+
+- `src/auth/Cargo.toml`
+- `src/gax/Cargo.toml`
+- `src/gax-internal/Cargo.toml`
+- `src/lro/Cargo.toml`
+
+You need to manually bump the corresponding versions in the top-level
```
dbolduc: nit: "You probably need to ..."

If no downstream crates use new types, we wouldn't need to bump the version in the top-level `Cargo.toml`.

In practice, I think I would send a PR and see if the minimal versions check fails. Then react in the `Cargo.toml`. :shrug:

dbolduc: Should we suggest running the minimal versions check locally? Or build it into `librarian`?

doc/contributor/howto-guide-generated-code-maintenance.md
coryan: > Q: can't this happen with any non-leaf crate? e.g. `google-cloud-longrunning` recently required a top level `Cargo.toml` update

Yes, any non-leaf crate may need a new version. But I am not sure I can write general instructions (for humans, far less for computers) as to when those changes would require a top-level Cargo.toml version bump. Why?

Most changes (new fields in messages, new RPCS) do not affect the downstream deps.  But a new message that was *used* by the downstream dep would require a bump.

> The version bump tool does not handle that, right?

Correct, it does not handle that, but I am not sure it needs to.

doc/contributor/howto-guide-generated-code-maintenance.md
coryan: Maybe we could simplify our life, and `librarian` could bump all the top-level version to match the actual version.  That seems safer, easier to implement, and easier to reason about: https://github.com/googleapis/librarian/issues/3720

Meanwhile, I am going to follow your suggestions.

doc/contributor/howto-guide-generated-code-maintenance.md
```diff
@@ -87,15 +82,39 @@ git checkout -b chore-bump-version-numbers-circa-$(date +%Y-%m-%d)
 V=$(cat .librarian-version.txt)
 go run github.com/googleapis/librarian/cmd/librarian@${V} bump --all
 go run github.com/googleapis/librarian/cmd/librarian@${V} generate --all
-git add Cargo.lock librarian.yaml '*Cargo.toml' '*README.md'
-git restore . # Effectively a `cargo fmt`, but much faster.
-git commit -m"chore: bump version numbers circa $(date +%Y-%m-%d)"
+# It is safe to commit everything because `bump` stops you from updating a
```
zhumin8, line 85: I don't think it's safe to commit everything?  because `librarian generate` is run, if source were updated since the last re-generate, then there might be code unwanted changes?

coryan: > I don't think it's safe to commit everything? because `librarian generate` is run, if source were updated since the last re-generate, then there might be code unwanted changes?

What source do you have in mind?  The `googleapis` and `discovery` SHAs are pinned, and do not change in this step. One needs to run `librarian update` for any SHA to change.

doc/contributor/howto-guide-generated-code-maintenance.md
dbolduc: > In practice, I think I would send a PR and see if the minimal versions check fails.

Oops, I didn't realize our minimal versions check only runs on a sample of the crates, not all of the crates.

> That seems safer, easier to implement, and easier to reason about

Now I agree with you.

doc/contributor/howto-guide-generated-code-maintenance.md
zhumin8: from offline: there is this build check that makes sure of consistency between the sources on record and generated code.  https://github.com/googleapis/google-cloud-rust/blob/c7e3488885ffd5224cb7d7577457dc311010c8e8/.gcb/scripts/regenerate.sh#L37-L44

Issue #4334, opened 2026-01-22
refactor(auth): interactions with MDS should live into a centralized client.


alvarowolfx:
So far MDS interactions were too scattered through the auth crate. For #4186 more MDS interactions are gonna be needed, so make sense to introduce a MDS client/mod for now on the auth crate. Later we can plan a MDS client that customers can use, similar to what we have in other SDKs.

* https://github.com/googleapis/google-cloud-go/tree/main/compute/metadata
* https://github.com/googleapis/google-cloud-node-core/tree/main/packages/gcp-metadata

coryan:
*terrible-ideas-discuss:* why not move this MDS client to a separate crate?

If we learnt something from the idtoken dependency tree problems is that deps are easier to manage if crates do fewer things.

alvarowolfx:
> _terrible-ideas-discuss:_ why not move this MDS client to a separate crate?
> 
> If we learnt something from the idtoken dependency tree problems is that deps are easier to manage if crates do fewer things.

later when making this client something that customer can use I agree that it needs to live into a separated crate. But right now I'm basically only gonna move some code that is interacting with MDS in the auth crate into a module. Even in that case you still feel like a separated crate can bring any advantage ? I'm not gonna be adding any new dep or something like that (like in the idtoken story).

PR #4336, opened 2026-01-22
refactor(auth): move signer to use MDS client

alvarowolfx:
Part of work for https://github.com/googleapis/google-cloud-rust/issues/4334


src/auth/src/signer/mds.rs
```diff
@@ -75,19 +73,7 @@ impl SigningProvider for MDSSigner {
...
-            .header(
-                METADATA_FLAVOR,
-                HeaderValue::from_static(METADATA_FLAVOR_VALUE),
-            );
-
-        let response = request.send().await.map_err(SigningError::transport)?;
-        let email = response.text().await.map_err(SigningError::transport)?;
-
-        Ok(email)
+        self.client.email().await.map_err(SigningError::transport)
```
coryan, line 76: This is going to cache the HTTP connection for as long as the `client` lives.... But the connection is only used rarely (once?) so you are wasting resources.

I believe the previous code created a client and then discarded the client and its resources.  That was probably more memory efficient.

alvarowolfx: Good call, I changed both PRs to stop caching the `reqwuest::Client`.

PR #4337, opened 2026-01-22
refactor(auth): move mds idtoken cred to use MDS client

alvarowolfx:
Part of work for https://github.com/googleapis/google-cloud-rust/issues/4334

### DIFF
--- a/src/auth/src/credentials/mds.rs
+++ b/src/auth/src/credentials/mds.rs
@@ -78,6 +78,7 @@ use crate::credentials::dynamic::{AccessTokenCredentialsProvider, CredentialsPro
 use crate::credentials::{AccessToken, AccessTokenCredentials, CacheableResource, Credentials};
 use crate::errors::CredentialsError;
 use crate::headers_util::build_cacheable_headers;
+use crate::mds::client::Client as MDSClient;
 use crate::mds::{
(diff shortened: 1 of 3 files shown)


src/auth/src/credentials/idtoken/mds.rs
```diff
@@ -257,9 +241,9 @@ impl Builder {
     }
 }
 
-#[derive(Debug, Clone, Default)]
+#[derive(Debug, Clone)]
 struct MDSTokenProvider {
-    endpoint: String,
+    client: MDSClient,
```
coryan, line 246: As in the other PR: this is keeping a cache of connections for longer than before, and therefore using more memory. 

Moreover, since the tokens are cached, the underlying client is used every hour or so. The cache probably saves no time because the connection is (likely) dropped by the server before the hour goes by.

I think you should consider just keeping the endpoint as before, and creating the MDSClient when needed.

alvarowolfx: I changed to stop caching the `reqwest::Client` and I still think worth keeping the MDSClient instead of the endpoint because the MDS Client is now handling the endpoint resolution (if is provided by the customer via props, env var or should use default value)

coryan: SG, but this needs a rebase.

PR #4338, opened 2026-01-22
feat(pubsub): add ability to resume publishing for Publisher

PhongChuong:
Fixes: #4015

### DIFF
--- a/src/pubsub/src/publisher/publisher.rs
+++ b/src/pubsub/src/publisher/publisher.rs
@@ -141,14 +141,7 @@ impl Publisher {
     /// # }
     /// ```
     pub fn resume_publish<T: std::convert::Into<std::string::String>>(&self, ordering_key: T) {
-        let (tx, _) = oneshot::channel();
-        if self
-            .tx
-            .send(ToWorker::ResumePublish(tx, ordering_key.into()))
-            .is_err()
-        {
-            // `tx` is dropped here if the send errors.
-        }
+        let _ = self.tx.send(ToWorker::ResumePublish(ordering_key.into()));
     }
 }
 
--- a/src/pubsub/src/publisher/worker.rs
+++ b/src/pubsub/src/publisher/worker.rs
@@ -26,7 +26,7 @@ pub(crate) enum ToWorker {
     /// A request to flush all outstanding messages.
     Flush(oneshot::Sender<()>),
     /// A request to resume publishing on an ordering key.
-    ResumePublish(oneshot::Sender<()>, String),
+    ResumePublish(String),
 }
 
 /// A command sent from the `Worker` to the background `batch` worker.
@@ -36,7 +36,7 @@ pub(crate) enum ToBatchWorker {
     /// A request to flush all outstanding messages.
     Flush(oneshot::Sender<()>),
     /// A request to resume publishing.
-    ResumePublish(oneshot::Sender<()>),
+    ResumePublish(),
 }
 
 /// Object that is passed to the worker task over the
@@ -154,15 +154,13 @@ impl Worker {
                             flush_set.join_all().await;
                             let _ = tx.send(());
                         },
-                        Some(ToWorker::ResumePublish(tx, ordering_key)) => {
+                        Some(ToWorker::ResumePublish(ordering_key)) => {
                             if let Some(batch_worker) = batch_workers.get_mut(&ordering_key) {
                                 // Send down the same tx for the BatchWorker to directly signal completion
                                 // instead of spawning a new task.
                                 batch_worker
-                                    .send(ToBatchWorker::ResumePublish(tx))
+                                    .send(ToBatchWorker::ResumePublish())
                                     .expect(batch_worker_error_msg);
-                            } else {
-                                let _ = tx.send(());
                             }
                         }
                         None => {
@@ -298,9 +296,8 @@ impl BatchWorker {
                             inflight = JoinSet::new();
                             let _ = tx.send(());
                         },
-                        Some(ToBatchWorker::ResumePublish(tx)) => {
+                        Some(ToBatchWorker::ResumePublish()) => {
                             // Nothing to resume as we do not pause without ordering key.
-                            let _ = tx.send(());
                         }
                         None => {
                             // TODO(#4012): Add shutdown procedure for BatchWorker.
@@ -346,9 +343,8 @@ impl BatchWorker {
                         // it was already handled when this was paused.
                         let _ = tx.send(());
                     }
-                    Some(ToBatchWorker::ResumePublish(tx)) => {
+                    Some(ToBatchWorker::ResumePublish()) => {
                         self.paused = false;
-                        let _ = tx.send(());
                     }
                     None => {
                         // TODO(#4012): Add shutdown procedure for BatchWorker.
@@ -386,8 +382,8 @@ impl BatchWorker {
                             }
                             let _ = tx.send(());
                         },
-                        Some(ToBatchWorker::ResumePublish(tx)) => {
-                            let _ = tx.send(());
+                        Some(ToBatchWorker::ResumePublish()) => {
+                            // Nothing to resume as we are not paused.
                         },
                         None => {
                             // TODO(#4012): Add shutdown procedure for BatchWorker.


src/pubsub/src/publisher/publisher.rs
```diff
@@ -118,6 +122,38 @@ impl Publisher {
         rx.await
             .expect("the client library should not release the sender");
     }
+
+    /// Resume accepting publish for a paused ordering key.
+    ///
+    /// If there is no ordering key or the ordering key is not currently paused, then this function is no-op.
+    ///
```
suzmue, line 130: nit: rewording. I think "if there is no ordering key" is slightly confusing to me.

```suggestion
    /// Resume accepting publish for a paused ordering key.
    ///
    /// Publishing using an ordering key might be paused if an error is encountered while publishing, to prevent messages from being published out of order.
    /// If the ordering key is not currently paused, this function is a no-op.
    ///
```

src/pubsub/src/publisher/publisher.rs
```diff
@@ -118,6 +122,38 @@ impl Publisher {
...
+    /// # Example
+    ///
+    /// ```
+    /// # use google_cloud_pubsub::model::PubsubMessage;
+    /// # async fn sample(publisher: google_cloud_pubsub::client::Publisher) -> anyhow::Result<()> {
+    /// if let Err(_) = publisher.publish(PubsubMessage::new().set_data("foo").set_ordering_key("bar")).await {
+    ///     // Handle the error.
```
suzmue: This comment looks like its referring to the next line of code but it isn't. Maybe like from https://docs.cloud.google.com/pubsub/docs/samples/pubsub-resume-publish-with-ordering-keys#pubsub_resume_publish_with_ordering_keys-go and do. Or just remove the comment might also be fine.

```suggestion
    ///     // Error handling code can go here.
```

src/pubsub/src/publisher/publisher.rs
```diff
@@ -118,6 +122,38 @@ impl Publisher {
...
+    ///     publisher.resume_publish("bar");
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn resume_publish<T: std::convert::Into<std::string::String>>(
```
suzmue: This should be a synchronous operation. We don't need for the resume to complete. Any messages that are put into the channel afterward will reach the unpaused worker.

src/pubsub/src/publisher/publisher.rs
```diff
@@ -1244,6 +1280,376 @@ mod tests {
...
+            .publish(PubsubMessage::new().set_ordering_key("").set_data("msg 1"))
+            .await;
+        assert_eq!(got.expect("expected message id"), "msg 1");
+
+        // Test resume and publish before the BatchWorker has been created.
+        publisher.resume_publish("ordering key without error").await;
+        let got = publisher
+            .publish(
+                PubsubMessage::new()
+                    .set_ordering_key("ordering key with error")
```
suzmue: Should this be the same as the line 1318? without error? I'm not sure the name is that much more descriptive than just doing key1 and key2 if we want a second one.

src/pubsub/src/publisher/publisher.rs
```diff
@@ -1244,6 +1280,376 @@ mod tests {
...
+
+        // Cause "ordering key with error" ordering key to pause.
+        let handle = publisher.publish(
+            PubsubMessage::new()
+                .set_ordering_key("ordering key with error")
+                .set_data("msg 0"),
+        );
+        publisher.flush().await;
+        // Assert the error is caused by the Publish send operation.
+        let mut got_err = handle.await.unwrap_err();
```
suzmue: I think we can drop the extra flush() when we use start_paused, because it will cause the timer to fire when we call handle.awai.

```suggestion
        // Assert the error is caused by the Publish send operation.
        let mut got_err = handle.await.unwrap_err();
```

src/pubsub/src/publisher/publisher.rs
```diff
@@ -1244,6 +1280,376 @@ mod tests {
...
+        let source = got_err
+            .source()
+            .and_then(|e| e.downcast_ref::<crate::error::PublishError>());
+        assert!(
+            matches!(
+                source,
+                Some(crate::error::PublishError::OrderingKeyPaused(()))
+            ),
+            "{got_err:?}"
+        );
```
suzmue, line 1580: nit: Maybe this could be in a helper, its kind of verbose and these tests become hard to skim.

src/pubsub/src/publisher/publisher.rs
PhongChuong: Good catch.
Updated to use a variable instead.
I'll follow up this PR with a cleanup of the tests.

src/pubsub/src/publisher/publisher.rs
PhongChuong: Acknowledge.
I'll follow up this PR with a cleanup of the tests.

src/pubsub/src/publisher/publisher.rs
PhongChuong: While messages that are put into the channel after will reach the unpaused worker, I think there are cases where the application would want a signal that the worker is now unpaused and publishing is back to the normal behavior.

This is achievable with async or potentially a long blocking sync. My preference is for it to be async and the application choose to .await on it if needed.

suzmue: I'm not sure I understand the use case. Could you explain it a bit more? If you call a synchronous (fast) resume_publish(), all publishes that occur after it will be batched and sent (until the next error of course). Why do we need to wait for the signal to reach the background worker? What additional benefit is there?

PhongChuong: Image if generating a message is expensive with a short TTL. In this case, the application may choose to await until the worker is ready again instead of having the messages potentially expire due to some other on going processes (i.e., flush).

Of course, this use case is completely imaginary but I also think that giving the application the option is beneficial. Is there a reason why sync fast resume is preferred over async version?

suzmue: > In this case, the application may choose to await until the worker is ready again instead of having the messages potentially expire due to some other on going processes (i.e., flush).
This is interesting, but I think maybe they should be awaiting on the flush not the resume_publish in this case.

One reason is to keep resume_publish cheap, its basically flipping a bit and we can immediately know that future publishes will get batched so there is just no need to delay in saying that its done.

Also from discussing this with Alex early in the design process he thought this should be a synchronous operation (if I am remembering correctly).

PhongChuong: As discussed offline, the issue with sync is that the worker may be blocked on a long flush operation causing a large delay in the resume_publish being processed due to it awaiting on other ordering keys. The ideal case would be for resume_publish to be sync and for flush not block other operations from being processed.
I'll update resume_publish to be sync and we can update flush in a later PR.

src/pubsub/src/publisher/publisher.rs
```diff
@@ -118,6 +122,34 @@ impl Publisher {
...
+    /// # async fn sample(publisher: google_cloud_pubsub::client::Publisher) -> anyhow::Result<()> {
+    /// if let Err(_) = publisher.publish(PubsubMessage::new().set_data("foo").set_ordering_key("bar")).await {
+    ///     // Error handling code can go here.
+    ///     publisher.resume_publish("bar");
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn resume_publish<T: std::convert::Into<std::string::String>>(&self, ordering_key: T) {
+        let (tx, _) = oneshot::channel();
```
suzmue: We can delete this channel and the comment below now.

src/pubsub/src/publisher/publisher.rs
```diff
@@ -118,6 +122,34 @@ impl Publisher {
...
+        if self
+            .tx
+            .send(ToWorker::ResumePublish(tx, ordering_key.into()))
+            .is_err()
+        {
+            // `tx` is dropped here if the send errors.
```
suzmue: delete, we are getting rid of the oneshot

src/pubsub/src/publisher/worker.rs
```diff
@@ -25,8 +25,8 @@ pub(crate) enum ToWorker {
     Publish(BundledMessage),
     /// A request to flush all outstanding messages.
     Flush(oneshot::Sender<()>),
-    // TODO(#4015): Add a resume function to allow resume Publishing on a ordering key after a
-    // failure.
+    /// A request to resume publishing on an ordering key.
+    ResumePublish(oneshot::Sender<()>, String),
```
suzmue: can get rid of this oneshot.

src/pubsub/src/publisher/worker.rs
```diff
@@ -35,8 +35,8 @@ pub(crate) enum ToBatchWorker {
     Publish(BundledMessage),
     /// A request to flush all outstanding messages.
     Flush(oneshot::Sender<()>),
-    // TODO(#4015): Add a resume function to allow resume Publishing on a ordering key after a
-    // failure.
+    /// A request to resume publishing.
+    ResumePublish(oneshot::Sender<()>),
```
suzmue: Remove oneshot

src/pubsub/src/publisher/worker.rs
```diff
@@ -367,6 +386,9 @@ impl BatchWorker {
                             }
                             let _ = tx.send(());
                         },
+                        Some(ToBatchWorker::ResumePublish(tx)) => {
+                            let _ = tx.send(());
```
suzmue: nit: Maybe a comment that the worker is not paused, resume is a no-op could be useful here.

src/pubsub/src/publisher/worker.rs
```diff
@@ -35,8 +35,8 @@ pub(crate) enum ToBatchWorker {
     Publish(BundledMessage),
     /// A request to flush all outstanding messages.
     Flush(oneshot::Sender<()>),
-    // TODO(#4015): Add a resume function to allow resume Publishing on a ordering key after a
-    // failure.
+    /// A request to resume publishing.
+    ResumePublish(),
```
suzmue, line 39: note: I'm only now noticing that ToWorker and ToBatchWorker are basically the same thing. I don't know if we can save anything performance wise from using the same type, but could be a place to consider refactoring in the future.

PR #4340, opened 2026-01-22
docs: Generate undelete RPC samples

amanda-tarafa:


src/generated/cloud/resourcemanager/v3/src/client.rs
coryan, line 1: I don't see anything here that looks like an undelete?

amanda-tarafa: Oh sorry, the undeletes in Resource Manager are long running, and that's what I have lined up after this one. I'm dropping enabling Resource Manager from this PR as it adds nothing here. It will be on the next one.

Here, the one from IAM Admin is the significant one.

coryan: SG

PR #4341, opened 2026-01-22
impl(dns/v1): generate library

coryan:
Mostly as a test of `librarian add` and to document what I had to manually edit.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2521,7 +2521,7 @@ dependencies = [
 
 [[package]]
 name = "google-cloud-dns-v1"
-version = "0.1.0"
+version = "0.0.0"
 dependencies = [
  "async-trait",
  "bytes",


librarian.yaml
```diff
@@ -642,6 +642,14 @@ libraries:
     copyright_year: "2025"
     rust:
       per_service_features: true
+  - name: google-cloud-dns-v1
+    copyright_year: "2026"
+    skip_publish: true
```
dbolduc, line 647: just curious: under what condition will we publish this crate?

Also, feels like an `impl` if we are not publishing the crate. Is this PR just to win the `repo-metadata.json` wars?

coryan: This has some LRO-looking things, I want to make sure we add the hand-crafted code before declaring success.

I also have some integration tests pending.

PR #4345, opened 2026-01-23
docs(auth): how-to migrate from 1.4.0 features

coryan:
In 1.5.0 we are introducing new default features, which may require some applications to change how they depend on and initialize their client.

Part of the work for #4170 , probably should wait until #4220 is merged.

[google_cloud_auth__manual___migrate_from_1_4_x_features - Rust.pdf](https://github.com/user-attachments/files/24824688/google_cloud_auth__manual___migrate_from_1_4_x_features.-.Rust.pdf)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -85,6 +85,28 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "c08606f8c3cbf4ce6ec8e28fb0014a2c086708fe954eaa885384a6165172e7e8"
 
 [[package]]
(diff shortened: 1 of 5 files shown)


coryan:
> Do you think we should move the compute v2 migration guide into the rustdocs?

I guess we could, as we now have a mechanism to add hand-crafted modules to generated libraries.

The question I have been asking myself is whether we should put this in docs.cloud.google.com

coryan:
Closing, we are going to use a different approach.

Issue #4350, opened 2026-01-23
Flake in secret manager integration tests


coryan:
The integration test just failed with a flake:

```
Testing update_secret_by_project_and_location_and_secret()
  |  
  |  
  | ERROR the service reports an error with code UNAVAILABLE described as: Visibility check was unavailable. Please retry the request and contact support if the problem persists
  |  
  |  
```
https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/18753ddc-71b0-443b-b2f4-b58f12189689?project=rust-sdk-testing&e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=monitoring_api_staging

I suspect the `update` test does not set the idempotency flag (or preconditions) correctly.

PR #4354, opened 2026-01-23
test: set idempotency in secret manager integration tests

suzmue:
Set `idempotency=true` for requests that are POST to make the integration tests more resistant to failures. These requests can safely be retried in our tests. 

We also deletion retries during cleanup.

For #4350


tests/integration/src/secret_manager/openapi_locational.rs
```diff
@@ -67,6 +69,7 @@ pub async fn run() -> Result<()> {
         .set_secret(&secret_id)
         .set_update_mask(wkt::FieldMask::default().set_paths(["labels"]))
         .set_body(smo::model::Secret::new().set_labels(new_labels))
+        .with_idempotency(true)
```
coryan: Consider making the idempotency thing true:

```suggestion
.set_body(smo::model::Secret::new().set_labels(new_labels).set_etag(&get.etag))
        .with_idempotency(true)
```

tests/integration/src/secret_manager/openapi_locational.rs
```diff
@@ -125,6 +129,7 @@ async fn run_iam(
             smo::model::TestIamPermissionsRequest::new()
                 .set_permissions(["secretmanager.versions.access"]),
         )
+        .with_idempotency(true)
```
coryan, line 136: Weird that this is a POST, this is a read-only operation 🤷

tests/integration/src/secret_manager/openapi_locational.rs
```diff
@@ -160,6 +165,7 @@ async fn run_iam(
                 .set_update_mask(wkt::FieldMask::default().set_paths(["bindings"]))
                 .set_policy(new_policy),
         )
+        .with_idempotency(true)
```
coryan, line 172: Using an etag would be the update truly idempotent.

suzmue: Since new_policy is created via policy.clone(), the etag should be included already.

tests/integration/src/secret_manager/openapi_locational.rs
suzmue: 🤷‍♀️  yeah strange, I did double check this one though.

tests/integration/src/secret_manager/openapi_locational.rs
suzmue: Done thank you!

PR #4356, opened 2026-01-23
feat(apiregistry/v1): generate library

### LINKED ISSUE #4352: Generate `cloudapiregistry.googleapis.com`
Upstream repo: https://github.com/googleapis/googleapis/tree/master/google/cloud/apiregistry/v1

Instructions: https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-generated-code-maintenance.md

coryan:
Fixes #4352

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1235,6 +1235,7 @@ dependencies = [
  "bytes",
  "google-cloud-gax",
  "google-cloud-gax-internal",
+ "google-cloud-location",
  "google-cloud-wkt",
  "lazy_static",
  "reqwest",


Cargo.lock
```diff
@@ -1235,6 +1235,7 @@ dependencies = [
  "bytes",
  "google-cloud-gax",
  "google-cloud-gax-internal",
+ "google-cloud-location",
```
zhumin8, line 1238: note to self: This had to be added manually via `cargo check` because `librarian generate` was run a second time after fixing and updating librarian version (not triggering the new lib generation flow, thus missing validation steps). If code deleted and run `librarian generate`, this Cargo.lock change would be included.

Issue #4097, opened 2025-12-18
Resumes for subscriber stream transient failures


dbolduc:
The subscriber `StreamingPull` RPC needs a resume loop.

We have one bidi stream (in `storage`), which required a custom resume loop, so we wrote it in the `storage` crate.

The plan of record is to write the StreamingPull resume loop in the `pubsub` crate.

This issue is to track the implementation of the resume loop, with some reasonable resume policy. A non-goal is to define the resume policy APIs and make it configurable.

coryan:
Does this overlap with #3020 ?

dbolduc:
> Does this overlap with https://github.com/googleapis/google-cloud-rust/issues/3020 ?

Yes it overlaps, but they are not necessarily the same.

I updated the issue description to say exactly what I intend this issue to cover. Sorry about leaving out context.

dbolduc:
TODO : We need to retry (at least) this RST_STREAM error, which I caught in the wild:

```
tonic::Status {
  code: Unknown,
  message: "error reading a body from connection",
  source: Some(hyper::Error(
    Body,
    h2::Error {
      kind: Reset(StreamId(1), INTERNAL_ERROR, Remote)
    }
  ))
}
```

@coryan thinks we should maybe retry all errors from tonic that contain a `hyper::Error`?

I see that we already classify errors as `Transport` errors when `tonic::Status::source().is_some() == true`.

https://github.com/googleapis/google-cloud-rust/blob/5e47416854b15b4777651efae00edd6bfc9651a0/src/gax-internal/src/grpc/from_status.rs#L38-L40

The thing giving me pause is this comment about regional endpoints.

https://github.com/googleapis/google-cloud-rust/blob/5e47416854b15b4777651efae00edd6bfc9651a0/src/gax/src/error/core_error.rs#L657-L664

Using the wrong endpoint should be a permanent error. If the subscriber retries it indefinitely, the application never gets to troubleshoot the real problem. Yuck. I haven't had time to test if it is a real problem.

PR #4357, opened 2026-01-23
impl(pubsub): streaming pull retry policy

dbolduc:
Part of the work for #4097 

Add a retry policy for streaming pull RPCs.

In future PRs, we will use this policy to retry opening a stream, and to determine if a mid-stream error is transient.

My plan is to keep this policy internal until further notice.

### DIFF
--- a/src/pubsub/src/subscriber/retry_policy.rs
+++ b/src/pubsub/src/subscriber/retry_policy.rs
@@ -19,6 +19,8 @@ use gax::retry_result::RetryResult;
 use gax::retry_state::RetryState;
 
 /// The subscriber's retry policy, specifically for StreamingPull RPCs.
+///
+/// This same policy applies to both starting and resuming a stream.
 #[derive(Debug)]
 pub(super) struct StreamRetryPolicy;
 


dbolduc:
Let me add some easy tests for better code coverage. 1 sec on the review...

src/pubsub/src/subscriber/retry_policy.rs
```diff
@@ -0,0 +1,231 @@
...
+//     }
+//   ))
+// }
+// ```
+//
+// [RST_STREAM]: https://datatracker.ietf.org/doc/html/rfc7540#section-6.4
+fn is_rst_stream(s: &Error) -> bool {
+    // Sadly, there is no unit test coverage for this variant as
+    // `h2::Error::is_reset()` is too difficult to trigger/fake.
+    as_inner::<h2::Error, Error>(s).is_some_and(|e| e.is_reset())
```
coryan: Isn't any `hyper::Error` retryable? All of them are basically "something bad in the connection".  Only definite gRPC errors with a proper `tonic::Status` are non-retryable, no?

Cargo.toml
```diff
@@ -315,6 +315,7 @@ crates_io_api         = { default-features = false, version = "0.12" }
 clap                  = { default-features = false, version = "4" }
 crc32c                = { default-features = false, version = "0.6.8" }
 futures               = { default-features = false, version = "0.3" }
+h2                    = { default-features = false, version = "0.4.13" }
```
coryan: A difficulty here is that you must link the same (major) version of `h2` that tonic is using, otherwise the `downcast_ref<>` fails.  `hyper` is at 1.x, which at least makes it more likely to be the same vs. a `0.x` crate like `h2`.

src/pubsub/src/subscriber/retry_policy.rs
coryan: Also, storage has similar code, maybe time to refactor this to `google_cloud_gax_internal::grpc` ?

dbolduc: > Only definite gRPC errors with a proper tonic::Status are non-retryable, no?

Sorry, I don't know, but I believe you. I still have a bad feel for the layers under `tonic` / `reqwest`.

From glancing at the tonic implementation, it looks like they construct their own errors when they don't get a definitive status from the server. And those are either `INTERNAL` or `UNKNOWN`.

> with a proper `tonic::Status`

Just to clarify, are you suggesting a "proper `tonic::Status`" is one where `err.source().is_none() == true` ?

If there is a source (and if the source is a `hyper::Error`), then we call the thing retryable?

dbolduc: > Also, storage has similar code, maybe time to refactor this to `google_cloud_gax_internal::grpc` ?

There are already (private) copies of `as_inner` in both `gaxi::grpc` and `gaxi::http`. :sweat_smile: 

I hesitated to add public APIs to gaxi, but I doubt this one will change. So sure, I'll consolidate these.

Cargo.toml
dbolduc: Ack. It would be doable, but super easy to mess up. I agree we should avoid it.

src/pubsub/src/subscriber/retry_policy.rs
coryan: > > with a proper `tonic::Status`
> 
> Just to clarify, are you suggesting a "proper `tonic::Status`" is one where `err.source().is_none() == true` ?

Yes. All such errors are created on the client-side and are (maybe excluding timeouts) retryable.

> If there is a source (and if the source is a `hyper::Error`), then we call the thing retryable?

If the `source()` is a `hyper::Error` then it is definitely related to I/O, and retryable.  I don't think we will get into too much trouble if even the errors with `source().is_some()` are treated as retryable, even if they are not a `hyper::Error`.

dbolduc:
> I think a more relaxed error would work better. Thoughts?

I think you are right. I appreciate the input, because this area is a blind spot for me.

I reverted the retries for the specific RST_STREAM, while I figure out what implementation we want. I want to get the `StreamRetryPolicy` interface in, so I can start sending other PRs that build on top of it. (The other PRs shouldn't be blocked on this implementation detail)

PTAL

src/pubsub/src/subscriber/retry_policy.rs
dbolduc: I captured this discussion + some extra thoughts as a comment in the parent issue: https://github.com/googleapis/google-cloud-rust/issues/4097#issuecomment-3793304044

Thanks.

dbolduc:
> A followup question: do you need a different policy to _resume_ a stream (when the stream succeeded at least once) vs. _retry_ the start?

It is the same policy for both. I asked the Pub/Sub folks.

Added a comment to the struct saying as much.

Also, had to rebase, so PTAL.

PR #4361, opened 2026-01-23
cleanup(pubsub): cleanup Publisher unit tests

PhongChuong:
This PR does the following:
- Move "my-topic" to global const
- Remove macro vec! when array is sufficient
- Add helper functions for mocking publish return for both error and non error cases
- Add helper function to validate publish when there are no errors
- Add helper function to validate publish when there is a paused error

This PR also removes some assertions that are deemed unnecessary or are already tested elsewhere.
E.g., asserting Publish request message size.

### DIFF
--- a/doc/contributor/howto-guide-generated-code-maintenance.md
+++ b/doc/contributor/howto-guide-generated-code-maintenance.md
@@ -22,8 +22,22 @@ instructions in [Set Up Development Environment].
 
 ## Generate new library
 
-First define the library's name. Note this should match the directory path where
-the code lives delimited by "-" e.g. google-cloud-kms-v1:
+### Verify `librarian` knows about the library
+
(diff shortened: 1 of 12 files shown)


src/pubsub/src/publisher/publisher.rs
```diff
@@ -303,28 +306,84 @@ mod tests {
...
+                PubsubMessage::new()
+                    .set_ordering_key(ordering_key)
+                    .set_data(generate_random_data()),
+            )
+            .await
+            .unwrap_err();
+        let source = got_err
+            .source()
+            .and_then(|e| e.downcast_ref::<crate::error::PublishError>());
+        assert!(
```
dbolduc, line 367: If we factor out an `assert!` like this, and it fails, the line number tells us nothing. We don't actually know which call to `assert_publishing_is_paused` triggered it.

PR #4371, opened 2026-01-24
cleanup: disable default features in local crates

### LINKED ISSUE #4238: Change top-level `Cargo.toml` file to use `default-features = false` on `google-cloud-*` crates
We want the top-level `Cargo.toml` file to *not* enable the default features in our own crates.  That way if `google-cloud-{foo}` depends on `google-cloud-longrunning`, `google-cloud-iam-v1`, `google-cloud-location` or any other dependency it will **not** enable the default features on those deps.

The crates will still enable the default features in `google-cloud-gax-internal`, making the crate usable by default, but also configurable to prune dependencies.

coryan:
Leaf crates (think `google-cloud-pubsub`) should depend on the
intermediate crates (think `google-cloud-gax-internal`) without enabling
any default dependencies of the intermediate crate.

With that setup the leaf crates can be consumed without any default
features throughout the dependency graph. That is, applications can pull
`google-cloud-pubsub` without default features and then all the upstream
dependencies also have no default features.

Fixes #4238

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -444,7 +444,7 @@ google-cloud-secretmanager-v1   = { default-features = false, version = "1", pat
 secretmanager                   = { default-features = false, version = "1", path = "src/generated/cloud/secretmanager/v1", package = "google-cloud-secretmanager-v1" }
 google-cloud-aiplatform-v1      = { default-features = false, version = "1", path = "src/generated/cloud/aiplatform/v1" }
 google-cloud-compute-v1         = { default-features = false, version = "2", path = "src/generated/cloud/compute/v1" }
-google-cloud-storage            = { default-features = false, version = "1", path = "src/storage" }
+google-cloud-storage            = { default-features = false, version = "1.7", path = "src/storage" }
 google-cloud-pubsub             = { default-features = false, path = "src/pubsub" }
 google-cloud-bigquery-v2        = { default-features = false, version = "1", path = "src/generated/cloud/bigquery/v2" }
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -397,61 +397,61 @@ tokio-stream      = { default-features = false, version = "0.1" }
...
-iamcredentials                  = { version = "1", path = "src/generated/iam/credentials/v1", package = "google-cloud-iam-credentials-v1" }
-google-cloud-language-v2        = { version = "1", path = "src/generated/cloud/language/v2" }
-language                        = { version = "1", path = "src/generated/cloud/language/v2", package = "google-cloud-language-v2" }
+google-cloud-speech-v2          = { default-features = false, version = "1", path = "src/generated/cloud/speech/v2" }
+google-cloud-secretmanager-v1   = { default-features = false, version = "1", path = "src/generated/cloud/secretmanager/v1" }
+secretmanager                   = { default-features = false, version = "1", path = "src/generated/cloud/secretmanager/v1", package = "google-cloud-secretmanager-v1" }
+google-cloud-aiplatform-v1      = { default-features = false, version = "1", path = "src/generated/cloud/aiplatform/v1" }
+google-cloud-compute-v1         = { default-features = false, version = "2", path = "src/generated/cloud/compute/v1" }
+google-cloud-storage            = { default-features = false, version = "1.7", path = "src/storage" }
```
dbolduc, line 447: Ack.

PR #4374, opened 2026-01-24
refactor: move `RandomChars` to test-utils

coryan:
Motivated by #4355

### DIFF
--- a/src/test-utils/src/random_chars.rs
+++ b/src/test-utils/src/random_chars.rs
@@ -1,0 +1,79 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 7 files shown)


src/test-utils/src/random_chars.rs
```diff
@@ -0,0 +1,79 @@
...
+impl RandomChars {
+    /// Initializes a new generator of random characters.
+    pub fn new(chars: &str) -> Self {
+        Self {
+            chars: chars.chars().collect(),
+        }
+    }
+
+    /// Create a string with `n` characters from the character set.
+    pub fn sample(&self, n: usize) -> String {
```
dbolduc, line 45: comment: we are covering the trait function... but if the compiler doesn't complain, neither will I.

`sample()` is a good name.

coryan: Ack.

PR #4376, opened 2026-01-24
feat(dns/v1): add LRO helpers

coryan:
Cloud DNS has a couple of LRO-type functions. Enabling the LRO wrappers
requires a few lines of hand-crafted code.

Note that the `librarian.yaml` file refers to the symbols without a package
name. That is needed because discovery-based services do not have a package
name. If you think: "But what about the Compute package further up in the
file?": for Compute we "infer" a package name from the service config yaml
file, but we have no such file for Cloud DNS.

With these changes I think the client library is ready to publish.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2540,7 +2540,7 @@ dependencies = [
 
 [[package]]
 name = "google-cloud-dns-v1"
-version = "0.0.0"
+version = "1.0.0"
 dependencies = [
  "async-trait",
  "bytes",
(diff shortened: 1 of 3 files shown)


tests/dns/src/lib.rs
```diff
@@ -52,6 +53,34 @@ pub async fn run_all() -> anyhow::Result<()> {
...
+        .set_body(update.clone())
+        .poller()
+        .until_done()
+        .await?;
+    tracing::info!("successfully updated zone: {updated:?}");
+    assert_eq!(
+        updated
+            .zone_context
+            .as_ref()
+            .and_then(|c| c.old_value.clone()),
```
dbolduc, line 73: optional nit: could take a reference instead of cloning

coryan: Ack.

PR #4377, opened 2026-01-24
feat(pubsub): retries starting subscribe stream

dbolduc:
Part of the work for #4097 

Retry attempts to start a stream using default policies.

I do not have plans to let applications set custom policies, but if we need to, we can add support for custom policies later.

### DIFF
--- a/src/pubsub/src/subscriber/session.rs
+++ b/src/pubsub/src/subscriber/session.rs
@@ -240,6 +240,7 @@ mod tests {
     use super::super::client::Subscriber;
     use super::super::keepalive::KEEPALIVE_PERIOD;
     use super::super::lease_state::tests::{test_id, test_ids};
+    use super::super::stream::{INITIAL_DELAY, MAXIMUM_DELAY};
     use super::*;
     use auth::credentials::anonymous::Builder as Anonymous;
     use pubsub_grpc_mock::google::pubsub::v1;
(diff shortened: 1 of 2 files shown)


src/pubsub/src/subscriber/session.rs
```diff
@@ -823,4 +824,50 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test(start_paused = true)]
+    async fn retry_transient_when_starting_stream() -> anyhow::Result<()> {
+        const NUM_RETRIES: u32 = 20;
```
coryan, line 830: nit: why 20? Seems like trying with 2 or three would be enough?

src/pubsub/src/subscriber/stream.rs
```diff
@@ -92,14 +125,63 @@ where
...
+    use gax::error::rpc::{Code, Status};
+    use gax::retry_state::RetryState;
+
+    mockall::mock! {
+        #[derive(Debug)]
+        BackoffPolicy {}
+        impl BackoffPolicy for BackoffPolicy {
+            fn on_failure(&self, state: &RetryState) -> Duration;
+        }
+    }
```
coryan, line 168: nit: we have like 5 of these by now. Time to refactor to `google-cloud-test-utils` ?  Hmmm... Dependencies may be more complicated than we want.

src/pubsub/src/subscriber/stream.rs
```diff
@@ -207,4 +289,95 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test]
+    async fn retry_then_success() -> anyhow::Result<()> {
+        let mut seq = mockall::Sequence::new();
+        let mut mock_stub = MockStub::new();
+        let mut mock_backoff = MockBackoffPolicy::new();
+        for attempt in 1..20 {
```
coryan, line 298: Want to add a constant for that 20 ? Is that supposed to be 2 * 10?

src/pubsub/src/subscriber/stream.rs
```diff
@@ -207,4 +289,95 @@ mod tests {
...
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn retry_then_permanent_failure() -> anyhow::Result<()> {
+        let mut seq = mockall::Sequence::new();
+        let mut mock_stub = MockStub::new();
+        let mut mock_backoff = MockBackoffPolicy::new();
+        for attempt in 1..20 {
```
coryan, line 340: Ditto.

src/pubsub/src/subscriber/session.rs
dbolduc: 20 isn't special. It is just supposed to represent "no attempt limit" aka higher than I can count.

I wanted to at least go above 10, because that is what we use in our default GAPIC policies.

https://github.com/googleapis/google-cloud-rust/blob/6124f3a210378ab6b645f3683d8bc7819c53885a/src/gax-internal/src/http.rs#L88

src/pubsub/src/subscriber/stream.rs
dbolduc: > Time to refactor to google-cloud-test-utils ?

Probably.

src/pubsub/src/subscriber/stream.rs
dbolduc: It is arbitrary. I can add a comment, but in a follow up PR.

src/pubsub/src/subscriber/stream.rs
dbolduc: Ditto.

src/pubsub/src/subscriber/session.rs
dbolduc: #4380

src/pubsub/src/subscriber/stream.rs
dbolduc: #4380

PR #4381, opened 2026-01-25
feat(pubsub): subscriber stream resumes

dbolduc:
Most of the work left for #4097 

Resume streams that were once successful and fail with a transient error.

### DIFF
--- a/src/pubsub/src/subscriber/session.rs
+++ b/src/pubsub/src/subscriber/session.rs
@@ -162,7 +162,7 @@ impl Session {
             // likely populate the message pool.
             //
             // Note that a successful read does not necessarily mean there is a
-            // message in the pool. The server occassionally sends heartbeats
+            // message in the pool. The server occasionally sends heartbeats
             // (responses with an empty message list). Hence the loop.
             if let Err(e) = self.read_from_stream().await? {
                 // Handle errors opening or reading from the stream.


src/pubsub/src/subscriber/session.rs
```diff
@@ -157,8 +159,17 @@ impl Session {
             }
             // Otherwise, read more messages from the stream.
             if let Err(e) = self.stream_next().await? {
-                // TODO(#4097) - support stream resumes.
-                return Some(Err(e));
+                match StreamRetryPolicy::is_transient(e) {
+                    RetryResult::Continue(_) => {
+                        // The stream failed with a transient error. Reset the stream.
```
coryan: So if `self.stream_next().await` returns `Some(Ok(...))` we do nothing? That seems odd and maybe should get a comment?

dbolduc: I think `self.stream_next()` was a really bad name. Changed to `self.read_from_stream()` so hopefully it's more clear.

And documented the member function for `read_from_stream()` to say that it populates the message pool.

dbolduc:
> I think the code is unclear. Maybe a comment would explain things. But happy to hear that I am wrong.

Totally fair. I think the diff also didn't help show context. I added a comment.

Also, you should know that I considered this way of writing the logic:

```rs
    pub async fn next(&mut self) -> Option<Result<(PubsubMessage, Handler)>> {
        while self.pool.is_empty() {
            // We do not have a message to serve. Keep reading responses from
            // the stream until we do.
            //
            // Note that a successful read does not necessarily mean there is a
            // message in the pool. The server occasionally sends heartbeats
            // (responses with an empty message list). Hence the loop.
            if let Err(e) = self.read_from_stream().await? {
                // Handle errors opening or reading from the stream.
                match StreamRetryPolicy::is_transient(e) {
                    RetryResult::Continue(_) => {
                        // The stream failed with a transient error. Reset the stream.
                        self.stream = None;
                        continue;
                    }
                    RetryResult::Permanent(e) | RetryResult::Exhausted(e) => {
                        // The stream failed with a permanent error. Return the error.
                        return Some(Err(e));
                    }
                }
            }
        }
        Some(Ok(self
            .pool
            .pop_front()
            .expect("I literally just told you the pool is not empty")))
    }
```

It is the code I would write in C++. The `while` makes it totally clear under what condition we stop reading from the stream. But the fact that I have to `unwrap()` from the front of the deque, when I know the deque is not empty just felt not Rust-y.

Let me know if you prefer this spelling of the code. My preference is not strong at all.

src/pubsub/src/subscriber/session.rs
dbolduc: > returns `Some(Ok(...))` we do nothing?

It is `Some(Ok(()))`, there is nothing to do.

I think your point is that from reading just this part of the code, the `...` type is not obvious. I am open to suggestions here if you have any.

src/pubsub/src/subscriber/session.rs
```diff
@@ -178,7 +196,13 @@ impl Session {
...
     }
 
-    async fn stream_next(&mut self) -> Option<Result<()>> {
+    /// Reads the next response from the stream.
+    ///
+    /// If we receive a response, we store the messages in `self.pool` and
+    /// forward the ack IDs to the lease management task.
+    ///
+    /// If we receive an error reading from the stream, we return it.
+    async fn read_from_stream(&mut self) -> Option<Result<()>> {
```
coryan, line 205: Maybe this should be called `populate_message_pool()`?

dbolduc:
> What about:
> ```rs
>     pub async fn next(&mut self) -> Option<Result<(PubsubMessage, Handler)>> {
>        let Some((message, handler)) = self.pool.pop_front() else {
>            return self.populate_pool_with_retry().await;
>        };
>        Some(Ok((message, handler)))
>    }
> ```

But then we would just say...

```rs
     pub async fn next(&mut self) -> Option<Result<(PubsubMessage, Handler)>> {
        self.populate_pool_with_retry().await
    }
```

and then at that point, why even have `populate_pool_with_retry()`?

src/pubsub/src/subscriber/session.rs
dbolduc: Hm. `read_from_stream` is more "what" we are doing, `populate_message_pool()` is more "why" we are doing it. Answering "why" seems better than answering "what"...

but we don't always receive a message... so it does not always populate the message pool... so I think I still prefer `read_from_stream`

coryan:
> and then at that point, why even have `populate_pool_with_retry()`?

Fair.

PR #4385, opened 2026-01-25
refactor: move crypto provider tests to workspace

### LINKED ISSUE #4246: Refactor `tests/cryto-provider` tests to use the workspace
Once all the dependencies are under control I believe we can add these tests to the workspace and still get the minimal set of deps using:

```
cargo run --package storage-with-aws-lc-rs
```

Having the tests in the workspace will make it easier to keep the dependencies (such as `anyhow` and `reqwest`) up to date. It also makes it easy to `cargo fmt` and `cargo clippy` all the things.

There may be problems with  `cargo build --workspace --all-features` because we have have impossible to build `jsonwebtoken` in that case.

Even getting "most" of these tests into the workspace would be useful though.

coryan:
Making the tests part of the workspace speeds up the builds and make it easier to maintain them.  I had to change the idtoken tests because the tricks with metadata do not work. Using `cargo tree` is more reliable anyways.

I also inspected `cargo tree` for each directory.

Fixes #4246


coryan:
> PR description might be stale. I didnt see auth changes. Maybe check before merging

Argh... you meant the `idtoken` tests?  The description was wrong, but not about `auth`.  Ooops.

PR #4388, opened 2026-01-25
ci: test all features in `minimal-versions` build

coryan:


coryan:
Do you mind approving #4387 first? I am afraid of archeologists, with or without a fedora.

PR #4396, opened 2026-01-26
cleanup(pubsub): use macros instead of functions for test assert helpers

PhongChuong:
By using macro instead of functions for test helpers with asserts, we preserves the line number during a test failure.


PhongChuong:
This is to address the concern in this [comment](https://github.com/googleapis/google-cloud-rust/pull/4361#discussion_r2723147079). @dbolduc , thanks for bringing up the issue.

PR #4399, opened 2026-01-26
fix(storage): `sign_with` is Send

msdrigg:
This PR updates signed_url.rs to ensure that when canonical_query is created (using a Serializer which is non-send), we drop the Serializer at that point so it doesn't remain alive until the end of the scope.

Currently even though the canonical_query variable is shadowed, rust does not automatically drop the shadowed variable until the end of scope, and `.finish()` only takes an `&mut Serializer`. So the original non-send `canonical_query` lives for the whole scope across an await point and therefore makes `sign_with` non-send.

### DIFF
--- a/src/storage/src/storage/signed_url.rs
+++ b/src/storage/src/storage/signed_url.rs
@@ -528,8 +528,9 @@ impl SignedUrlBuilder {
             for (k, v) in &query_parameters {
                 canonical_query.append_pair(k, v);
             }
-    
-            canonical_query.finish()
+
+            canonical_query
+                .finish()
                 .replace("%7E", "~") // rollback to ~
                 .replace("+", "%20") // missing %20 in +
         };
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/signed_url.rs
```diff
@@ -523,15 +523,16 @@ impl SignedUrlBuilder {
...
-            .replace("+", "%20"); // missing %20 in +
+        let canonical_query = {
+            let mut canonical_query = url::form_urlencoded::Serializer::new("".to_string());
+            for (k, v) in &query_parameters {
+                canonical_query.append_pair(k, v);
+            }
+    
+            canonical_query.finish()
+                .replace("%7E", "~") // rollback to ~
+                .replace("+", "%20") // missing %20 in +
```
dbolduc: `cargo fmt -p google-cloud-storage` wants the following:

```suggestion
            canonical_query
                .finish()
                .replace("%7E", "~") // rollback to ~
                .replace("+", "%20") // missing %20 in +
```

msdrigg:
Give me a minute on the test, just fixed the suggested lint issue

dbolduc:
> Give me a minute on the test, just fixed the suggested lint issue

No worries on the test. I can always add it in a follow up PR. Your PR is good to go.

Also thanks for doing the CLA business

msdrigg:
Sorry just saw your comment here -- added the test and ensured it passed, and ran cargo fmt again on the whole file

dbolduc:
> Sorry just saw your comment here -- added the test and ensured it passed, and ran cargo fmt again on the whole file

Even better. Thanks!

msdrigg:
Let me know if you need anything else like rebase etc.

PR #4403, opened 2026-01-26
impl(auth): move away from jsonwebtoken

alvarowolfx:
Towards #3947

### DIFF
--- a/src/auth/Cargo.toml
+++ b/src/auth/Cargo.toml
@@ -46,6 +46,7 @@ serde_json.workspace  = true
 # TODO: move new deps to workspace
 pkcs1                 = { version = "0.7", optional = true, features = ["alloc"] }
 spki                  = { version = "0.7", optional = true, features = ["alloc"] }
+p256                  = { workspace = true, optional = true, features = ["ecdsa"] }
 thiserror.workspace   = true
 time                  = { workspace = true, features = ["serde"] }
 tokio                 = { workspace = true, features = ["fs", "process"] }
@@ -72,7 +73,7 @@ mutants.workspace     = true
 default = ["default-idtoken-backend", "default-rustls-provider"]
 # The `idtoken` feature enables support to create and validate OIDC ID Tokens.
 # See the create top-level documentation for more information.
-idtoken = ["dep:pkcs1", "dep:spki", "jsonwebtoken"]
+idtoken = ["dep:pkcs1", "dep:spki", "dep:p256", "jsonwebtoken"]
 # By default this crate enables the `aws_lc_rs` backend. Applications can
 # link `google-cloud-auth` with `default-features = false, features = ["idtoken"]
 # and then directly configure the `jsonwebtoken` features to select the
--- a/src/auth/src/credentials/idtoken/verifier.rs
+++ b/src/auth/src/credentials/idtoken/verifier.rs
@@ -39,6 +39,7 @@
 use crate::credentials::internal::jwk_client::JwkClient;
 use crate::credentials::service_account::jws::JwsHeader;
 use base64::Engine;
+use p256::ecdsa::Signature;
 use rustls::crypto::CryptoProvider;
 /// Represents the claims in an ID token.
 pub use serde_json::Map;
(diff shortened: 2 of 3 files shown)


src/auth/Cargo.toml
```diff
@@ -72,20 +71,20 @@ mutants.workspace     = true
 default = ["default-idtoken-backend", "default-rustls-provider"]
 # The `idtoken` feature enables support to create and validate OIDC ID Tokens.
 # See the create top-level documentation for more information.
-idtoken = ["dep:jsonwebtoken", "jsonwebtoken"]
+idtoken = ["dep:rsa"]
```
coryan: `cargo semver-checks` will not like it if you remove the `jsonwebtoken` feature here.

src/auth/Cargo.toml
```diff
@@ -46,7 +46,7 @@ serde_json.workspace  = true
 thiserror.workspace   = true
 time                  = { workspace = true, features = ["serde"] }
 tokio                 = { workspace = true, features = ["fs", "process"] }
-jsonwebtoken          = { workspace = true, optional = true }
+rsa                   = { workspace = true, optional = true, features = ["pem", "sha2"] }
```
coryan: Is rsa FIPS certified? Some customers prefer `aws-lc-rs` because it is ... I don't mean that we *must* use `aws-lc-rs`, just that whatever we use (or allow customers to use) must have that property.

coryan: I though the current version of rsa (0.9.x) has some CVE related to timing attacks. It was fine to use as a dev-dependency, but maybe not as a normal dep?

alvarowolfx: yeah, rsa still has the CVE, but `jsonwebtoken` depend on it anyway. When customers enable the `idtoken` feature, it still has the rust-sec warning.

alvarowolfx: will check that in depth, but so far it looks like it's not FIPS certified.

src/auth/Cargo.toml
alvarowolfx: good call

src/auth/Cargo.toml
alvarowolfx: depends on it, with the `rust_crypto` feature***. Sorry, forgot that we are defaulting to `aws-lc-rs`

alvarowolfx:
@coryan another take on this, now trying to move most of the crypto related stuff to rustls, but to read public certificates as DER I had to add others crate lower level crates - `spki` and `pkcs1`, which were already part of the dependency tree IIUC.

coryan:
> @coryan another take on this, now trying to move most of the crypto related stuff to rustls, but to read public certificates as DER I had to add others crate lower level crates - `spki` and `pkcs1`, which were already part of the dependency tree IIUC.

I am warming up to this idea.  Can you explore if `rustls-pki-types` has the features you need? That is also in our dependency tree, and it is at 1.x:

https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.SubjectPublicKeyInfoDer.html

alvarowolfx:
> > @coryan another take on this, now trying to move most of the crypto related stuff to rustls, but to read public certificates as DER I had to add others crate lower level crates - `spki` and `pkcs1`, which were already part of the dependency tree IIUC.
> 
> 
> 
> I am warming up to this idea.  Can you explore if `rustls-pki-types` has the features you need? That is also in our dependency tree, and it is at 1.x:
> 
> 
> 
> https://docs.rs/rustls-pki-types/latest/rustls_pki_types/struct.SubjectPublicKeyInfoDer.html
> 
> 
> 
> 

I tried that too, but JWK key sets comes as a pair of prime numbers that can be used to derive the certificate key. Looks like the rustls-pki-types expect the cert to be already in DER/bytes format. But will try a bit more here.

src/auth/Cargo.toml
```diff
@@ -71,7 +73,7 @@ mutants.workspace     = true
 default = ["default-idtoken-backend", "default-rustls-provider"]
 # The `idtoken` feature enables support to create and validate OIDC ID Tokens.
 # See the create top-level documentation for more information.
```
coryan, line 75: Since you are here, I just noticed a typo "create" -> "crate's"

alvarowolfx:
temporarily close this, will revisit later

PR #4407, opened 2026-01-26
refactor: another common test function

coryan:


src/test-utils/src/resource_names.rs
```diff
@@ -12,13 +12,26 @@
...
+//! Helper functions and types to generate random resource names.
 
 use rand::{
     Rng,
     distr::{Distribution, Uniform},
 };
 
+/// A common prefix for resource ids.
+///
+/// Where possible, we use this prefix for randomly generate resource ids.
```
dbolduc: nit: s/generate/generated/

Issue #4400, opened 2026-01-26
chore: update ci to check direct minimal versions only


suzmue:
We are currently checking that our builds succeed with the minimal versions we declare for our direct dependencies, as well as the indirect dependencies. Since we don't have control over the indirect dependencies, this is too strict (those problems should be fixed in our dependencies not google-cloud-rust). 

https://doc.rust-lang.org/nightly/cargo/reference/unstable.html#direct-minimal-versions

cargo-minimal-versions support: https://crates.io/crates/cargo-minimal-versions#--direct--z-direct-minimal-versions

suzmue:
Starting to run these tests has revealed a number of errors in our minimal version declarations (especially wrt versions not having the feature that we want to use). https://crates.io/crates/cargo-minimal-versions#--direct--z-direct-minimal-versions has a disclaimer that minimal-versions and direct-minimal-versions may reveal different bugs, which is turning out to be true for us.

<img width="618" height="228" alt="Image" src="https://github.com/user-attachments/assets/0cab13f0-e7de-4a02-8d1e-2fe98d993951" />

We will probably want to run both instead of switching.

suzmue:
We now also have tests for direct minimal versions. Follow up work in #4430

PR #4410, opened 2026-01-26
test: add direct minimal versions check to ci

suzmue:
Copy the build script to run minimal versions using the direct flag. We want to run both since they catch different issues. In particular, the --direct flag catches cases where we declare a version that does not have the feature flag that we use in our code.

Additionally, `cargo minimal-versions --direct` is unable to select the minimal version when the version we declare is below the version our dependencies declare. Update the versions in our Cargo.toml to be aligned with those in our dependencies.

For #4400

### DIFF
--- a/.gcb/scripts/minimal-versions-direct.sh
+++ b/.gcb/scripts/minimal-versions-direct.sh
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-# Copyright 2025 Google LLC
+# Copyright 2026 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.


suzmue:
Temporarily added the build trigger for this PR to check that --direct passes: https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/51b997ac-c659-423a-9808-23d74e3c0549?project=rust-sdk-testing&e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=monitoring_api_staging

coryan:
> Temporarily added the build trigger 

Any reason why `gcloud builds submit` is not good enough to test a new build?

suzmue:
> > Temporarily added the build trigger
> 
> Any reason why `gcloud builds submit` is not good enough to test a new build?

Only that I did not know about it.

suzmue:
[Build log](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/12340b50-6ca2-4c35-acf4-31ad3d50f05f?project=rust-sdk-testing&e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=monitoring_api_staging)
```sh
gcloud builds submit https://github.com/suzmue/google-cloud-rust --git-source-revision=dep-resolution --config=.gcb/complex.yaml --substitutions=_SCRIPT=minimal-versions-direct --region=us-central1
```

.gcb/scripts/minimal-versions-direct.sh
```diff
@@ -0,0 +1,46 @@
+#!/usr/bin/env bash
```
dbolduc, line 1: Should we consolidate this and `minimal-versions.sh`? The only difference is the extra `--direct` flag.

I think it can be done with a build substitution, or by exporting a variable in this script, and then invoking the other script.

Like

```sh
export MIN_VERSION_ARGS=("--direct")
source .gcb/scripts/minimal-versions.sh
```

and the other one looks like:

```sh
cargo minimal-versions "${MIN_VERSION_ARGS[@]}" test --all-features --package google-cloud-storage
```

^ untested, but something like that.

suzmue: agreed. created https://github.com/googleapis/google-cloud-rust/issues/4430

PR #4425, opened 2026-01-27
refactor(pubsub): move shutdown logic into BatchWorker

PhongChuong:
Towards: #4012


src/pubsub/src/publisher/worker.rs
```diff
@@ -247,6 +237,29 @@ impl BatchWorker {
         }
     }
 
+    // Flush the pending batch and pending messages by sending and awaiting Publish
+    // sequentially.
+    async fn flush_sequential(&mut self, mut inflight: JoinSet<Result<(), gax::error::Error>>) {
```
suzmue, line 241: This refactor makes me happy 😄

src/pubsub/src/publisher/worker.rs
```diff
@@ -247,6 +237,29 @@ impl BatchWorker {
...
+        while !self.pending_batch.is_empty() || !self.pending_msgs.is_empty() {
+            self.move_to_batch();
+            self.pending_batch
+                .flush(self.client.clone(), self.topic.clone(), &mut inflight);
+            self.handle_inflight_join(inflight.join_next().await);
+        }
+    }
+
+    // Flush the pending batch and pending messages by sending Publish
+    // concurrent.
```
suzmue: nit:
```suggestion
    // Flush the pending batch and pending messages by sending remaining
    // messages in concurrent batches.
```

src/pubsub/src/publisher/worker.rs
```diff
@@ -247,6 +237,29 @@ impl BatchWorker {
         }
     }
 
+    // Flush the pending batch and pending messages by sending and awaiting Publish
+    // sequentially.
```
suzmue: nit: something like
```suggestion
    // Flush the pending messages by sending the messages in sequential batches.
```

src/pubsub/src/publisher/worker.rs
```diff
@@ -373,20 +385,17 @@ impl BatchWorker {
...
                         Some(ToBatchWorker::Flush(tx)) => {
-                            // Send batches sequentially.
-                            self.handle_inflight_join(inflight.join_next().await);
-                            while !self.pending_batch.is_empty() || !self.pending_msgs.is_empty() {
-                                self.move_to_batch();
-                                self.pending_batch.flush(self.client.clone(), self.topic.clone(), &mut inflight);
-                                self.handle_inflight_join(inflight.join_next().await);
-                            }
+                            self.flush_sequential(inflight).await;
+                            inflight = JoinSet::new();
```
suzmue, line 388: It feels a little off that we need to do this reset outside of flush_sequential. If its necessary, can we do it inside?

PhongChuong: IMO, it's slightly better to have it outside of the flush functions. The function consumes the JoinSet similar to tokio JoinSet join_all() so the compiler will catch cases where accidentally forget to reset the inflight. In the other case where it's not needed such as shutdown, it will be fine and saves a very minor initialization.

PR #4433, opened 2026-01-28
fix(secret_manager): garbage collect locational secrets in integration tests

### LINKED ISSUE #4349: Integration tests are not garbage collecting locational secrets
Our integration tests create locational secrets with dummy data, but we don't seem to garbage collect them correctly:

Image

Probably some silly mistake in the garbage collection loops?

haphungw:
Fixes #4349

The stale secrets were not correctly removed since wrong arguments were passed into `delete_secret_by_project_and_location_and_secret`. This fix parses the secret name to properly delete stale secrets.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5831,7 +5831,6 @@ dependencies = [
  "google-cloud-telcoautomation-v1",
  "google-cloud-test-utils",
  "google-cloud-wkt",
- "google-cloud-workflows-executions-v1",
  "google-cloud-workflows-v1",
  "http",
  "httptest",
@@ -5886,6 +5885,23 @@ dependencies = [
  "google-cloud-lro",
  "google-cloud-test-utils",
  "rand 0.9.2",
+ "tokio",
+ "tracing",
+ "tracing-subscriber",
+]
+
+[[package]]
+name = "integration-tests-enums-query-parameters"
+version = "0.0.0"
+dependencies = [
+ "anyhow",
+ "chrono",
+ "futures",
+ "google-cloud-gax",
+ "google-cloud-lro",
+ "google-cloud-test-utils",
+ "google-cloud-workflows-executions-v1",
+ "google-cloud-workflows-v1",
  "tokio",
  "tracing",
  "tracing-subscriber",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -292,6 +292,7 @@ members = [
   "tests/crypto-providers/test-storage",
   "tests/dns",
   "tests/endurance",
+  "tests/enums-query-parameters",
   "tests/integration",
   "tests/integration-auth",
   "tests/o11y",
@@ -472,6 +473,9 @@ language                        = { default-features = false, version = "1", pat
 google-cloud-dns-v1             = { default-features = false, path = "src/generated/cloud/dns/v1" }
 google-cloud-showcase-v1beta1   = { default-features = false, path = "src/generated/showcase" }
 google-cloud-trace-v1           = { default-features = false, path = "src/generated/devtools/cloudtrace/v1" }
+# The names on these are too long, I do not want to reformat the whole file for them.
+google-cloud-workflows-v1            = { default-features = false, path = "src/generated/cloud/workflows/v1" }
+google-cloud-workflows-executions-v1 = { default-features = false, path = "src/generated/cloud/workflows/executions/v1" }
 
 # Local test packages (never add a version)
 google-cloud-test-utils = { default-features = false, path = "src/test-utils" }
--- a/src/test-utils/src/resource_names.rs
+++ b/src/test-utils/src/resource_names.rs
@@ -16,7 +16,7 @@
 
 use rand::{
     Rng,
-    distr::{Distribution, Uniform},
+    distr::{Alphanumeric, Distribution, Uniform},
 };
 
 /// A common prefix for resource ids.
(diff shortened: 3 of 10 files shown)


tests/integration/src/secret_manager/openapi_locational.rs
```diff
@@ -399,12 +399,18 @@ async fn cleanup_stale_secrets(
 
     let pending = stale_secrets
         .iter()
-        .map(|secret_id| {
+        .map(|name| {
+            // format: projects/{project}/locations/{location}/secrets/{secret}
+            let parts: Vec<&str> = name.split('/').collect();
+            let project = parts[1];
+            let location = parts[3];
+            let secret = parts[5];
```
coryan: This will panic if the name is not in the right format:

https://doc.rust-lang.org/std/vec/struct.Vec.html#indexing

You may try something like:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=c1164c67bd2b89c1fcde61bcd31ed939

Which validates all the stuff and never panics.

tests/integration/src/secret_manager/openapi_locational.rs
```diff
@@ -399,12 +399,18 @@ async fn cleanup_stale_secrets(
 
     let pending = stale_secrets
         .iter()
-        .map(|secret_id| {
+        .map(|name| {
+            // format: projects/{project}/locations/{location}/secrets/{secret}
+            let parts: Vec<&str> = name.split('/').collect();
```
coryan: Materializing the vector always makes me sad (an allocation that strictly we do not need).  But for test code, this is fine.

tests/integration/src/secret_manager/openapi_locational.rs
```diff
@@ -419,3 +421,44 @@ async fn cleanup_stale_secrets(
 
     Ok(())
 }
+
+fn extract(name: &str) -> Result<(&str, &str, &str)> {
```
coryan: I was maybe too succinct in my playground example: 

```suggestion
fn parse_secret_name(name: &str) -> Result<(&str, &str, &str)> {
```

tests/integration/src/secret_manager/openapi_locational.rs
```diff
@@ -419,3 +421,44 @@ async fn cleanup_stale_secrets(
...
+    #[test]
+    fn test_extract_invalid_name() {
+        let name = "invalid/format";
+        assert!(extract(name).is_err());
+    }
+
+    #[test]
+    fn test_extract_missing_parts() {
+        assert!(extract("projects/only-one").is_err());
+    }
```
coryan: Consider:


```suggestion
    #[test_case("invalid/format")]
    #[test_case("projects/only-one")]
    #[test_case("projects/p/locations/l/bad/s")]
    #[test_case("projects/p/bad/l/secrets/s")]
    #[test_case("bad/p/locations/l/secrets/s")]
    fn test_extract_invalid_name(input: &str) {
        let result = parse_secret_name(input);
        assert!(result.is_err(), "{result:?});
    }
```

tests/integration/src/secret_manager/openapi_locational.rs
```diff
@@ -419,3 +421,44 @@ async fn cleanup_stale_secrets(
...
+            "secrets",
+            s,
+        ] => Ok((p, l, s)),
+        _ => Err(anyhow!("invalid secret name: {}", name)),
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
```
coryan, line 435:
```suggestion
    use super::*;
    use test_case::test_case;
```

to use `#[test_case]` below.

tests/integration/src/secret_manager/openapi_locational.rs
```diff
@@ -399,12 +400,13 @@ async fn cleanup_stale_secrets(
 
     let pending = stale_secrets
         .iter()
-        .map(|secret_id| {
+        .map(|name| {
+            let (project, location, secret) = extract(name).unwrap();
```
coryan: This still panics, no? Maybe something like (but move the function definition to the right place):

```rust
async fn delete_by_name(client: smo::client::SecretManagerService, &name: &str) -> anyhow::Result<()> {
    let (project, location, secret) = parse_secret_name(name)?;
    client.delete_secret_by_project_and_location_and_secret()
        .set_project(project)
        .set_location(location)
        .set_secret(secret)
        .with_idempotency(true)
        .send()
        .await?;    
    Ok(())
}

let pending = stale_secrets.iter().map(|name| delete_by_name(name));
```

coryan: Hmmm... now that I think about it, we don't need a function, I think an expression like this should work to:

```rust
let pending = stale_secrets.iter().map(|name| -> anyhow::Result<()> {
    let (project, location, secret) = parse_secret_name(name)?;
    client.delete_secret_by_project_and_location_and_secret()
        .set_project(project)
        .set_location(location)
        .set_secret(secret)
        .with_idempotency(true)
        .send()
        .await?;    
    Ok(())    
});
```

Your call.

haphungw: keeping it inline makes sense since it's only used once here. 
I wrapped it in an `async move` block to make the `.await` work inside the closure.

coryan (approved):

PR #4439, opened 2026-01-28
chore(deps): bump gax-internal major

dbolduc:
We missed this in a previous PR. Better late than never.

We could probably get away with not bumping the major, but if a GAPIC ever uses `gaxi::http::to_http_error`, we will have broken semver with the gaxi `0.7.x` series.


dbolduc:
We decided that once we miss a major bump it is already too late. We should only bump the major if there is another break.

PR #4440, opened 2026-01-28
cleanup(storage): use re-exports from `gaxi`

coryan:


src/storage/Cargo.toml
```diff
@@ -60,17 +59,23 @@ hex                        = { workspace = true, features = ["std"] }
...
+  "_internal-common",
+  "_internal-grpc-client",
+  "_internal-http-multipart",
+  "_internal-http-stream",
+] }
 
 [features]
 default         = ["default-rustls-provider"]
-unstable-stream = ["reqwest/stream"]
+unstable-stream = []
```
dbolduc, line 80: Doesn't this have to turn on the gaxi stream feature?

src/gax-internal/Cargo.toml
```diff
@@ -72,6 +72,8 @@ _internal-grpc-client = [
   "dep:tracing",
   "dep:wkt",
 ]
+_internal-http-multipart = ["reqwest?/multipart"]
+_internal-http-stream = ["reqwest?/stream"]
```
dbolduc: https://github.com/googleapis/google-cloud-rust/blob/d1003c28cffbefe87ec960b95baeeca850f4ad3a/.gcb/scripts/features.sh#L48

src/storage/src/storage/perform_upload/buffered/progress.rs
```diff
@@ -180,7 +181,7 @@ impl InProgressUpload {
             }
             None
         });
-        reqwest::Body::wrap_stream(stream)
+        ReqwestBody::wrap_stream(stream)
```
dbolduc: This needs to be gated by the `_internal-http-stream` feature

src/storage/Cargo.toml
dbolduc: Which doesn't actually gate any code... That seems wrong.

dbolduc (changes requested):
I don't think the `_internal-http-stream` feature is correct.

And should I expect changes to the features build?

src/gax-internal/Cargo.toml
coryan: Fixed.

src/storage/src/storage/perform_upload/buffered/progress.rs
coryan: I assume you mean that `ReqwestBody` should be re-exported only if the feature is enabled. That is done, but I am not sure it was needed. `ReqwestBody` is always available, `ReqwestBody::wrap_stream` is enabled by the streaming feature, but that is always turned on in this code.

If you mean the code in `src/storage` then I disagree. That code always turns on the streaming feature, it does not need to gate anything.

src/storage/Cargo.toml
coryan: It does gate some code:

https://github.com/googleapis/google-cloud-rust/blob/8c79e7d80411d49bfa71b0404b5af3a1d3fe8ea1/src/storage/src/read_object.rs#L110-L113

> Doesn't this have to turn on the gaxi stream feature?

No, that feature is always turned on because it is needed to implement the uploads (which stream the ` contents).  Why turn it on again?

src/storage/src/storage/perform_upload/buffered/progress.rs
dbolduc: > `reqwest::Body::wrap_stream` is enabled by the streaming feature

Ah. I see that now. Thanks. https://docs.rs/reqwest/0.13.1/reqwest/struct.Body.html#method.wrap_stream

Sorry, I saw the new feature `_internal-http-stream` in gaxi and didn't understand what it did, because there was no code in our repo behind it. Thanks for being patient while I catch up.

src/gax-internal/src/http.rs
```diff
@@ -30,12 +30,23 @@ use gax::retry_throttler::SharedRetryThrottler;
...
 };
-use http::{Extensions, Method};
+use http::Extensions;
 use std::sync::Arc;
 use std::time::Duration;
 #[cfg(google_cloud_unstable_tracing)]
 use tracing::Instrument;
 
+#[cfg(feature = "_internal-http-stream")]
+pub use reqwest::Body as ReqwestBody;
```
dbolduc: > [gating `ReqwestBody`] is done, but I am not sure it was needed.

Yeah, feel free to revert.

src/gax-internal/Cargo.toml
dbolduc: Can you add a comment for those in the back of the class:

"The `stream` feature in `reqwest` exposes the `Body::wrap_stream` API, which is needed by `google-cloud-storage`."

And maybe also: "We expose the `reqwest` feature in `gaxi` so that `google-cloud-storage` does not take a direct dependency on `reqwest`. This way we can ensure only one version of `reqwest` is used by the crate."

.gcb/scripts/features.sh
```diff
@@ -51,6 +51,8 @@ for sub in test doc; do
   cargo "${sub}" --package google-cloud-gax-internal --no-default-features --features _internal-common
   cargo "${sub}" --package google-cloud-gax-internal --no-default-features --features _internal-http-client
   cargo "${sub}" --package google-cloud-gax-internal --no-default-features --features _internal-grpc-client
+  cargo "${sub}" --package google-cloud-gax-internal --no-default-features --features _internal-http-client,_internal-http-multipart
+  cargo "${sub}" --package google-cloud-gax-internal --no-default-features --features _internal-http-client,_internal-http-stream
```
dbolduc: Q: Hmm should the new `_internal-http-*` features should enable the existing `_internal-http-client` feature?

I think it works either way. Just wondering if you considered it.

dbolduc (approved):
Looks good. I did not understand things in my initial review. Consider adding a comment to mitigate future confusion.

src/gax-internal/Cargo.toml
coryan: Done.

src/gax-internal/src/http.rs
coryan: Reverted.

.gcb/scripts/features.sh
coryan: I did, was ambivalent, changed.

src/gax-internal/src/http/reqwest.rs
```diff
@@ -0,0 +1,27 @@
+// Copyright 2025 Google LLC
```
dbolduc: nit

src/gax-internal/src/http/reqwest.rs
```diff
@@ -0,0 +1,27 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Re-export symbols from the [reqwest] crate.
+//!
+//! [reqwest]: [::reqwest]
```
dbolduc, line 17: TIL Rust has fully qualified names.

coryan:
Already approved, I had this in draft, saving for future me:

> 
> I am not too worried about the names. I have different naming standards for types that only we use vs. types that we expect customers to use.

I worry about naming conflicts and ambiguity, but not really about "pretty" for these.

> But that idea is much nicer if `reqwest` (the crate) is banned from `google-cloud-storage`'s direct dependencies.

Not banned at the moment. It is removed.  I cannot get `cargo deny` to even run.

> Up to you though.

I did the `reqwest` sub-module thing.

PR #4441, opened 2026-01-29
refactor(pubsub): reorganize publisher modules

suzmue:
The publisher module was getting a bit large. This change splits out building the publisher from creating the publisher and renames it to client to get rid of the publisher::publisher path.  PublisherBuilder and PublisherPartialBuilder were copied directly (minus a single fix to a typo in the the throttler documentation).

### DIFF
--- a/src/pubsub/src/publisher.rs
+++ b/src/pubsub/src/publisher.rs
@@ -12,12 +12,12 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-pub mod base_publisher;
+pub(crate) mod base_publisher;
 pub(crate) mod batch;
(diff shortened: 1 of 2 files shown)


src/pubsub/src/publisher.rs
```diff
@@ -14,9 +14,10 @@
 
 pub mod base_publisher;
 pub(crate) mod batch;
+pub mod builder;
+pub mod client;
```
dbolduc: can these be `pub(crate)` ?

and same for `mod base_publisher`

suzmue: I think they all can be.

PR #4447, opened 2026-01-29
impl(auth): make TokenProviderWithRetry RefUnwindSafe

alvarowolfx:
The `TokenProviderWithRetry` struct was not implementing `RefUnwindSafe` because it held `Arc<dyn RetryPolicy>` and `Arc<dyn BackoffPolicy>`, which are trait objects that do not automatically implement `RefUnwindSafe`.

When  `TokenProviderWithRetry` was embedded into other public structs (like `IDTokenCredentials`), those structs also lost their `RefUnwindSafe` implementation. This caused a semantic versioning check failure as `RefUnwindSafe` is part of the public API promise.
 
Towards #3592

### DIFF
--- a/src/auth/src/retry.rs
+++ b/src/auth/src/retry.rs
@@ -23,85 +23,43 @@ use gax::retry_loop_internal::retry_loop;
 use gax::retry_policy::{AlwaysRetry, RetryPolicy, RetryPolicyArg, RetryPolicyExt};
 use gax::retry_throttler::{AdaptiveThrottler, RetryThrottlerArg, SharedRetryThrottler};
 use std::error::Error;
-use std::panic::RefUnwindSafe;
+use std::panic::{RefUnwindSafe, UnwindSafe};
 use std::sync::{Arc, Mutex};
-
-/// A wrapper to assert `RefUnwindSafe` on the inner type.
-///
-/// This is necessary because adding [TokenProviderWithRetry] to existing, released auth features
-/// caused a breaking change. The containing structs would lose their automatically derived
-/// `RefUnwindSafe` implementation because the dynamic trait objects (like `dyn RetryPolicy`)
-/// are not `RefUnwindSafe` by default.
-///
-/// This wrapper solves that by manually implementing `RefUnwindSafe`, allowing us to add
-/// retry functionality without triggering a semver-check failure. This is safe because
-/// we control the implementation of the inner types and can ensure that they are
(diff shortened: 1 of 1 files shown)


alvarowolfx:
unblocks https://github.com/googleapis/google-cloud-rust/pull/4260, https://github.com/googleapis/google-cloud-rust/pull/4257 and https://github.com/googleapis/google-cloud-rust/pull/4256

src/auth/src/retry.rs
```diff
@@ -23,31 +23,85 @@ use gax::retry_loop_internal::retry_loop;
...
+/// are not `RefUnwindSafe` by default.
+///
+/// This wrapper solves that by manually implementing `RefUnwindSafe`, allowing us to add
+/// retry functionality without triggering a semver-check failure. This is safe because
+/// we control the implementation of the inner types and can ensure that they are
+/// `RefUnwindSafe`.
+#[derive(Debug)]
+struct UnwindSafeAdapter<T>(T);
+
+impl<T> RefUnwindSafe for UnwindSafeAdapter<T> {}
```
coryan: If we are going to do this, I propose we simply implement the trait for `TokenProviderWithRetry` (or the builder for it).

alvarowolfx: It's an option too. I did this because only `retry_policy` and `backoff_policy` are not UnwindSafe, but you're right, in the end we are adding a `impl RefUnwindSafe` thing anyway.

alvarowolfx: hmm, weird. I had to add both `RefUnwindSafe` and `UnwindSafe` explicitly to make it work:
```
impl RefUnwindSafe for Builder {}
impl UnwindSafe for Builder {}
```

coryan: Yeah, weird. I don't see the changes here, I assume you are experimenting locally or something. Let me know when this is ready for review again.

alvarowolfx: yeah, I'm testing in another branch, so I can run `semver-check` to make sure it looks good. The latest changes are on this one: https://github.com/googleapis/google-cloud-rust/pull/4256/changes#diff-43612f24f6413f8e5304d6af3e912553d7913ceea55c203c549db570b52d0d30

If I don't add both `impl`, the check fails.

alvarowolfx:
@coryan  I've updated this PR based on your comments on https://github.com/googleapis/google-cloud-rust/pull/4256

Issue #2023, opened 2025-05-07
Use generated code for LRO integration tests


coryan:
The integration tests for `src/lro` use a hand-crafted client that "simulates" the generated code. We should use a normal, generated client. This would require moving the test to `src/integration-tests`, but that is not too much of a burden.

PR #4449, opened 2026-01-29
refactor: move LRO integration test

coryan:
Move the test from src/lro/tests to tests/lro. Then we can start using a generated library instead of hand-crafted code to complete the test.

Part of the work for #2023


tests/lro/README.md
```diff
@@ -0,0 +1,5 @@
+# Integration tests for LROs
```
dbolduc, line 1: Does this belong under `src/lro/tests` ?

The fake client means it is not much of an integration test. (i.e. nothing is leaving the machine running the tests, there are no pre-requisites on the environment). And we have all the workflows integration tests that do real LROs...

---

If you respond "dude, who cares?", I will approve.

coryan: I am moving this *from* `src/lro/tests`. The next step is to replace all the code in `fake/client.rs` and `fake/model.rs` with a real generated client, at which point it cannot be there.  I will also move the integration tests using workflows here.

dbolduc: hmm ok

PR #4450, opened 2026-01-29
docs(contributing): update commit message guidelines

suzmue:
The commit message guidelines in CONTRIBUTING.md have been updated to be more detailed and align with Cloud SDK best practices established in github.com/googleapis/librarian.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -24,7 +24,9 @@ information on using pull requests.
 
 ## Commit Messages
 
-Commit messages for `google-cloud-rust` follow the conventions below.
+Commit messages for `google-cloud-rust` follow the conventions below. Note that
+your PR title defaults to the first commit in your branch, and that the merge
+commit is composed of your PR title and PR description by default.
 
 Here is an example:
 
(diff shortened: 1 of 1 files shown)


suzmue:
I dropped the following line from the librarian docs since I wasn't sure if we wanted to adopt it. I think we do tend to use minimal markdown.
> Don't use HTML, Markdown, or any other markup language.

CONTRIBUTING.md
```diff
@@ -24,10 +24,77 @@ information on using pull requests.
...
+
+#### scope
+
+The name of the crate affected by the change, which should be provided in
+parentheses before the colon. Please omit the `google-cloud-` prefix (e.g., use
+`storage` instead of `google-cloud-storage`).
+
+#### description
+
+A short one-line summary of the change, that it should be written so to complete
```
dbolduc:
```suggestion
A short one-line summary of the change. It should complete
```

CONTRIBUTING.md
```diff
@@ -24,10 +24,77 @@ information on using pull requests.
...
+the sentence "This change modifies the crate to ..." That means it does not
```
dbolduc: nit: I think we want a period after the quote too.

and maybe just: `the crate to ...". It does not start with`

CONTRIBUTING.md
```diff
@@ -24,10 +24,77 @@ information on using pull requests.
...
+start with a capital letter, is not a complete sentence, and actually summarizes
+the result of the change. Note that the verb after the colon is lowercase, and
+there is no trailing period
```
dbolduc: ok, but this sentence needs a trailing period. 😄

CONTRIBUTING.md
```diff
@@ -24,10 +24,77 @@ information on using pull requests.
...
+The first line of the change description is a short one-line summary of the
+change, following the structure `<type>(<scope>): <description>`:
+
+#### type
+
+A structural element defined by the conventions at
+[https://www.conventionalcommits.org/en/v1.0.0/#summary](https://www.conventionalcommits.org/en/v1.0.0/#summary).
+
+Conventional commits are parsed by our release tooling to generate release
+notes. See [Guidelines for Commit Types](#guidelines-for-commit-types) for more
```
dbolduc, line 58: I think we should inline this. Why have two separate sections on types?

CONTRIBUTING.md
dbolduc: Apparently we should not put an extra period after the quote. I retained none of the MLA formatting lessons from high school.

CONTRIBUTING.md
suzmue: Done.

CONTRIBUTING.md
suzmue: Woah I also would have though there should be a period TBF.

CONTRIBUTING.md
suzmue: Done.

It seemed kind of long to inline but its probably fine.

CONTRIBUTING.md
suzmue: Done.

PR #4457, opened 2026-01-30
refactor(integration-tests): split protobuf tests

coryan:


tests/protobuf/src/mocking.rs
```diff
@@ -40,8 +47,8 @@ mod mocking {
             .await
     }
 
-    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
PhongChuong, line 43: Not sure if removing `(flavor = "multi_thread", worker_threads = 1)` was intentional.

tests/protobuf/src/pagination.rs
```diff
@@ -286,8 +290,8 @@ mod mocking {
         Ok(())
     }
 
-    #[tokio::test(flavor = "multi_thread", worker_threads = 1)]
```
PhongChuong, line 289: ditto.

tests/protobuf/src/mocking.rs
coryan: It was, for a mocking test those are useless. Arguably harmful because they may mask some errors.

tests/protobuf/src/pagination.rs
coryan: Same.

PR #4461, opened 2026-01-30
ci: speed up integration tests

coryan:
I used `--workspace` to make it easier to add integration tests. That slows down the compilation too much. This is slightly harder, we need to add the integration tests in two places, but it may reduce the compilation time.


coryan:
11m and 14m vs. 30m, I'll take that.

PR #4474, opened 2026-01-30
chore: make `cargo update` respect the MSRV

### LINKED ISSUE #4428: `time` requires Rust 1.88 and we require Rust 1.85
It seems that the (mostly optional) `time` dependency now requires a more recent version of Rust than we do. Until then we cannot automatically update the `Cargo.lock` file, which is annoying. 

We could:

- Wait until we move to Rust 1.88: by our current policy (which we can change!) that is until 2026-07 or so.
- Stop updating the `Cargo.lock` file manually.
- Investigate if any `cargo` tool, or feature from nightly solves the problem.

Also see: https://github.com/rust-lang/cargo/issues/9930

coryan:
TIL, one can configure the resolver to [respect the MSRV]. It requires rust >= 1.84 and the 2024 edition. Neither of those is a problem for us.

[respect the MSRV]:
https://doc.rust-lang.org/cargo/reference/resolver.html#resolver-versions

Fixes #4428

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 [workspace]
-resolver = "2"
+resolver = "3"
 
 default-members = [
   "guide/samples",


dbolduc (approved):
Nice.

PR #4475, opened 2026-01-31
impl(pubsub): retry acks on connection closed

### LINKED ISSUE #4471: Retry at-least-once lease operations that fail when a connection is closed
See: https://github.com/googleapis/google-cloud-rust/issues/4275#issuecomment-3820223699

And then a passing thought: I am using the transport stub, which is built atop the gax retry loop. The retry loop clones the request on each attempt. That is wasteful if we are never retrying. This is moot now though, because I actually want to retry things every once in a while.

dbolduc:
Fixes #4471 

I have been running a benchmark for >1 hour with these changes, and the client reports no failed acks thus far.

### DIFF
--- a/src/pubsub/src/subscriber/retry_policy.rs
+++ b/src/pubsub/src/subscriber/retry_policy.rs
@@ -71,9 +71,10 @@ impl RetryPolicy for AtLeastOnceRetryPolicy {
             // can race with accepting one of our requests.
             //
             // It is safe to retry these requests immediately.
-            return RetryResult::Continue(error);
+            RetryResult::Continue(error)
+        } else {
+            RetryResult::Permanent(error)
         }
-        return RetryResult::Permanent(error);
     }
 }
 


src/pubsub/src/subscriber/retry_policy.rs
```diff
@@ -55,6 +57,39 @@ impl RetryPolicy for StreamRetryPolicy {
...
 
+/// The retry policy for lease management RPCs in at-least-once delivery.
+///
+/// Specifically, these are the `Acknowledge` and `ModifyAckDeadline` RPCs.
+#[derive(Debug)]
+pub(super) struct AtLeastOnceRetryPolicy;
+
+impl RetryPolicy for AtLeastOnceRetryPolicy {
+    fn on_error(&self, state: &RetryState, error: Error) -> RetryResult {
+        if state.attempt_count == 1 && error.is_transport() {
```
coryan, line 68: Fine, but a `OnlyTransportErrors.with_attempt_limit(1)` would give you more reusable components.

Do you want to retry auth errors too?

src/pubsub/src/subscriber/retry_policy.rs
```diff
@@ -55,6 +57,39 @@ impl RetryPolicy for StreamRetryPolicy {
...
+
+/// The backoff policy for lease management RPCs in at-least-once delivery.
+///
+/// Specifically, these are the `Acknowledge` and `ModifyAckDeadline` RPCs.
+#[derive(Debug)]
+pub(super) struct AtLeastOnceBackoffPolicy;
+
+impl BackoffPolicy for AtLeastOnceBackoffPolicy {
+    fn on_failure(&self, _state: &RetryState) -> Duration {
+        Duration::ZERO
```
coryan, line 89: I am not sure why, but this gives me the heebie-jeebies.  Works, but only because of that attempt limit.

src/pubsub/src/subscriber/retry_policy.rs
dbolduc: No, because auth errors would warrant some kind of backoff.

The problem I am trying to solve is losing Acks when a channel closes, which reliably happens every hour.

From testing, this change reduces the lost acks, but not entirely. I missed that the same Ack requests could get retried on a different channel that is also closed, and receive another transport error.

So maybe I should retry these immediately up to N times where N is the number of channels.

Or maybe I should manually reset the channels every 55 minutes, or something.

coryan: > No, because auth errors would warrant some kind of backoff.
> 
> The problem I am trying to solve is losing Acks when a channel closes, which reliably happens every hour.

Right. I am arguing that maybe you should should retry any request that never hit the network. 🤷 

> Or maybe I should manually reset the channels every 55 minutes, or something.

Maybe there is a way to keep the channels active. I think gRPC for C++ had some configs to send keep alive messages every X seconds.

coryan: Maybe this:

https://docs.rs/tonic/0.14.2/tonic/transport/channel/struct.Endpoint.html#method.http2_keep_alive_interval

dbolduc: Hm, I didn't look there. I'll see if it helps. Thanks.

dbolduc:
> Fine, but a `OnlyTransportErrors.with_attempt_limit(1)` would give you more reusable components.

Ack. Thanks for the review. I am just going to start over.

PR #4476, opened 2026-01-31
refactor: split storage integration tests

coryan:
Move the integration tests for storage to its own crate. In a future refactor we may move these to a subdirectory of `src/storage/`.

### DIFF
--- a/guide/samples/tests/storage.rs
+++ b/guide/samples/tests/storage.rs
@@ -28,6 +28,7 @@ pub mod storage {
     #[cfg(all(test, feature = "run-integration-tests"))]
     mod driver {
         use super::*;
+        use storage_samples::create_test_hns_bucket;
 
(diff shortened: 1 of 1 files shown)


tests/storage/tests/driver.rs
```diff
@@ -0,0 +1,140 @@
...
+            "default-endpoint",
+        )
+        .await;
+        if let Err(e) = storage_samples::cleanup_bucket(control, bucket.name.clone()).await {
+            tracing::error!("error cleaning up test bucket {}: {e:?}", bucket.name);
+        };
+        result
+    }
+
+    #[test_case(Storage::builder(); "default")]
```
dbolduc, line 93: Unrelated nit: just noticed these single test case tests

coryan: Ack.

PR #4478, opened 2026-01-31
chore: prepare for `cargo deny`

coryan:
It seems that `cargo deny` does not work when we alias the crate in the
`[dev-dependencies]` section. In addition, a number of "private" crates
lacked a license entry, and we need some allow lists here and there.

There is room to tighten the `cargo deny` configuration, but that can
wait.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -61,7 +61,8 @@ registries = []
 
 [bans]
 # Allow multiple versions, there are too many of them via indirect dependencies.
-# Note that `
+# Note that `bans.workspace-dependencies` can make sure we avoid dups in our
+# direct dependencies.
 multiple-versions = "allow"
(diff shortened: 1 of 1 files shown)


deny.toml
```diff
@@ -0,0 +1,92 @@
...
+license-files = []
+
+# Even private crates must have a license.
+[licenses.private]
+ignore     = false
+registries = []
+
+[bans]
+# Allow multiple versions, there are too many of them via indirect dependencies.
+# Note that `
```
dbolduc: ...

dbolduc (approved):

deny.toml
coryan: Fixed.

dbolduc (approved):

PR #4493, opened 2026-02-02
chore: bump version numbers circa 2026-02-02

coryan:


coryan:
> why no gaxi bump?

Already bumped in #4440

dbolduc:
> > why no gaxi bump?
> 
> Already bumped in #4440 

Oops, right.

There is a version of gaxi in librarian.yaml that still says 0.7.8. That confused me. I assume we should update that too.

coryan:
> 
> There is a version of gaxi in librarian.yaml that still says 0.7.8. That confused me. I assume we should update that too.

Good catch:  #4497

PR #4496, opened 2026-02-02
cleanup: prevent direct uses of `tonic` with exemptions

### LINKED ISSUE #4485: Stop direct use of the `reqwest`, `tonic` and similar deps
We can configure `cargo deny` to avoid direct dependencies on `reqwest` and `tonic`.  Without a direct dependency our code cannot use `tonic::*` or `reqwest::*` names. Only any re-exports via `google-cloud-gax-internal`.

We can make a few exceptions for tests and `gax-internal` itself:

https://embarkstudios.github.io/cargo-deny/checks/bans/cfg.html#the-wrappers-field-optional

coryan:
Fixes #4485

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -98,8 +98,8 @@ wrappers = [
 [[bans.deny]]
 crate = "tonic@0.14"
 wrappers = [
-  # Uses `reqwest` in its "public" API. The types are re-exported. This is
-  # the main motivation for denying reqwest in general: we want to avoid
+  # Uses `tonic` in its "public" API. The types are re-exported. This is
+  # the main motivation for denying `tonic` in general: we want to avoid
   # accidental use of the crate in other crates.
   "google-cloud-gax-internal",
   # Use in tests is fine.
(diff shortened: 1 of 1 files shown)


deny.toml
```diff
@@ -95,6 +95,38 @@ wrappers = [
   "user-guide-samples",
 ]
 
+[[bans.deny]]
+crate = "tonic@0.14"
+wrappers = [
+  # Uses `reqwest` in its "public" API. The types are re-exported. This is
```
dbolduc: nit: tonic

deny.toml
```diff
@@ -95,6 +95,38 @@ wrappers = [
...
+  # Indirect dependencies are fine.
+  "tonic-prost",
+  "opentelemetry-otlp",
+  "opentelemetry-proto",
+]
+
+[[bans.deny]]
+crate = "tonic-prost@0.14"
+wrappers = [
+  # Uses `reqwest` in its "public" API. The types are re-exported. This is
```
dbolduc: same

deny.toml
coryan: Fixed, thanks.

deny.toml
coryan: Changed to say fewer lies.

dbolduc (approved):

PR #4506, opened 2026-02-02
cleanup: use full name for `google-cloud-lro` 

coryan:
Using the full name makes for clearer examples, easier to grok code when customers read the source in docs.rs, and avoids duplicate dependencies.


coryan:
/FYI: @amanda-tarafa after this PR it is possible to `google_cloud_lro::` in any sample code.

PR #4512, opened 2026-02-02
refactor(pubsub): move error messages to constants

PhongChuong:
This PR also updates the error message to be slightly more specific.


src/pubsub/src/publisher/model_ext.rs
```diff
@@ -17,6 +17,8 @@ use std::pin::Pin;
 use std::task::{Context, Poll, ready};
 use tokio::sync::oneshot;
```
dbolduc, line 19: nit: why the blank line?

src/pubsub/src/publisher/model_ext.rs
```diff
@@ -58,7 +60,7 @@ impl Future for PublishHandle {
         // which would be a bug.
         Poll::Ready(
             result
-                .expect("the client library should not release the sender")
+                .expect(constants::PUBLISH_HANDLE_SENDER_ERROR_MSG)
```
dbolduc: Q: are we going to reuse this error message? If not, I would just spell out the error here.

dbolduc (approved):

src/pubsub/src/publisher/model_ext.rs
PhongChuong: It was a mistake.

PR #4522, opened 2026-02-03
cleanup(integration-tests): prefer full dep names

coryan:
We can use the dependencies from the workspace and avoid dups.

### DIFF
--- a/tests/integration/Cargo.toml
+++ b/tests/integration/Cargo.toml
@@ -27,41 +27,26 @@ log-integration-tests = ["google-cloud-test-utils/log-integration-tests"]
 run-integration-tests = []
 
 [dependencies]
-anyhow.workspace        = true
-bytes.workspace         = true
-futures.workspace       = true
-google-cloud-auth       = { workspace = true, features = ["default"] }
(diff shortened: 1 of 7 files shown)


tests/integration/Cargo.toml
```diff
@@ -27,41 +27,26 @@ log-integration-tests = ["google-cloud-test-utils/log-integration-tests"]
...
-
-[dependencies.wf]
-package = "google-cloud-workflows-v1"
-path    = "../../src/generated/cloud/workflows/v1"
+anyhow.workspace                = true
+bytes.workspace                 = true
+futures.workspace               = true
+google-cloud-auth               = { workspace = true, features = ["default"] }
+google-cloud-gax                = { workspace = true, features = ["unstable-stream"] }
+gaxi                            = { workspace = true, features = ["_internal-http-client"] }
```
dbolduc, line 35: ah, the one crate that will never appear in our docs

coryan: FWIW, I may want to use the full name for `gaxi` at some point. I think it will require more changes to the generator. The name appears in the `*.go` (as opposed to just `*.mustache`) files.  Ditto for `wkt`.

Issue #4370, opened 2026-01-24
Cannot run storage and tracing tests with user credentials


coryan:
Without a quota project the telemetry integration tests fail with:

```
test telemetry::test_telemetry_e2e ... FAILED

failures:

---- telemetry::test_telemetry_e2e stdout ----
View generated trace in Console: https://console.cloud.google.com/traces/explorer;traceId=b781fd6bb6b7160712ec4d83e53b3fae?project=coryan-test
[...]

PR #4525, opened 2026-02-03
docs(storage): add samples for generate signed urls

### LINKED ISSUE #3081: storage_generate_signed_url_v4

### LINKED ISSUE #3082: storage_generate_upload_signed_url_v4

joshuatants:
Fixes #3081
Fixes #3082

### DIFF
--- a/src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
+++ b/src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
@@ -18,32 +18,21 @@ use google_cloud_storage::http::Method;
 use std::time::Duration;
 
 pub async fn sample(bucket_name: &str, object_name: &str) -> anyhow::Result<()> {
-    let signer = google_cloud_auth::credentials::Builder::default().build_signer();
-    let signer = match signer {
-        Ok(s) => s,
-        Err(err) if err.is_not_supported() => {
-            tracing::warn!("skipping run_storage_signed_urls: {err:?}");
-            return Ok(());
-        }
-        Err(err) => {
-            anyhow::bail!("error creating signer: {err:?}")
-        }
-    };
+    let signer = google_cloud_auth::credentials::Builder::default().build_signer()?;
 
-    let signed_url = SignedUrlBuilder::for_object(bucket_name, object_name)
-        .with_method(Method::PUT)
-        .with_expiration(Duration::from_secs(15 * 60)) // 15 minutes
-        .with_header("content-type", "application/octet-stream")
-        .sign_with(&signer)
-        .await
-        .map_err(anyhow::Error::new)?;
+    let signed_url =
+        SignedUrlBuilder::for_object(format!("projects/_/buckets/{bucket_name}"), object_name)
+            .with_method(Method::PUT)
+            .with_expiration(Duration::from_secs(15 * 60)) // 15 minutes
+            .with_header("content-type", "application/octet-stream")
+            .sign_with(&signer)
+            .await?;
 
     println!("Generated PUT signed URL:");
-    println!("{}", signed_url);
+    println!("{signed_url}");
     println!("You can use this URL with any user agent, for example:");
     println!(
-        "curl -X PUT -H 'Content-Type: application/octet-stream' --upload-file my-file '{}'",
-        signed_url
+        "curl -X PUT -H 'Content-Type: application/octet-stream' --upload-file my-file '{signed_url}'",
     );
 
     Ok(())
--- a/src/storage/examples/src/objects/generate_signed_url_v4.rs
+++ b/src/storage/examples/src/objects/generate_signed_url_v4.rs
@@ -18,29 +18,19 @@ use google_cloud_storage::http::Method;
 use std::time::Duration;
 
 pub async fn sample(bucket_name: &str, object_name: &str) -> anyhow::Result<()> {
-    let signer = google_cloud_auth::credentials::Builder::default().build_signer();
-    let signer = match signer {
-        Ok(s) => s,
-        Err(err) if err.is_not_supported() => {
-            tracing::warn!("skipping run_storage_signed_urls: {err:?}");
-            return Ok(());
-        }
-        Err(err) => {
-            anyhow::bail!("error creating signer: {err:?}")
-        }
-    };
+    let signer = google_cloud_auth::credentials::Builder::default().build_signer()?;
 
-    let signed_url = SignedUrlBuilder::for_object(bucket_name, object_name)
-        .with_method(Method::GET)
-        .with_expiration(Duration::from_secs(15 * 60)) // 15 minutes
-        .sign_with(&signer)
-        .await
-        .map_err(anyhow::Error::new)?;
+    let signed_url =
+        SignedUrlBuilder::for_object(format!("projects/_/buckets/{bucket_name}"), object_name)
+            .with_method(Method::GET)
+            .with_expiration(Duration::from_secs(15 * 60)) // 15 minutes
+            .sign_with(&signer)
+            .await?;
 
     println!("Generated GET signed URL:");
-    println!("{}", signed_url);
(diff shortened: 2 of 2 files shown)


src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
```diff
@@ -0,0 +1,51 @@
...
+    let signer = match signer {
+        Ok(s) => s,
+        Err(err) if err.is_not_supported() => {
+            tracing::warn!("skipping run_storage_signed_urls: {err:?}");
+            return Ok(());
+        }
+        Err(err) => {
+            anyhow::bail!("error creating signer: {err:?}")
+        }
+    };
```
coryan: This kind of error handling is confusing for an example.  Keep it focused on the signing, we can have a separate example or user guide on the errors:


```suggestion
    let signer = google_cloud_auth::credentials::Builder::default().build_signer()?;
```

src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
```diff
@@ -0,0 +1,51 @@
...
+
+    let signed_url = SignedUrlBuilder::for_object(bucket_name, object_name)
```
coryan: This needs to be:

```suggestion
    let signed_url = SignedUrlBuilder::for_object(format!("projects/_/buckets/{bucket_name}"), object_name)
```

The Rust storage client uses bucket names in proto format everywhere.

src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
```diff
@@ -0,0 +1,51 @@
...
+        .with_method(Method::PUT)
+        .with_expiration(Duration::from_secs(15 * 60)) // 15 minutes
+        .with_header("content-type", "application/octet-stream")
+        .sign_with(&signer)
+        .await
+        .map_err(anyhow::Error::new)?;
+
+    println!("Generated PUT signed URL:");
+    println!("{}", signed_url);
```
coryan: nit: 
```suggestion
    println!("{signed_url}");
```

src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
```diff
@@ -0,0 +1,51 @@
...
+    println!("You can use this URL with any user agent, for example:");
+    println!(
+        "curl -X PUT -H 'Content-Type: application/octet-stream' --upload-file my-file '{}'",
+        signed_url
+    );
```
coryan, line 36: nit: 
```suggestion
    println!(
        "curl -X PUT -H 'Content-Type: application/octet-stream' --upload-file my-file '{signed_url}'"
    );
```

src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
```diff
@@ -0,0 +1,51 @@
...
+        }
+    };
+
+    let signed_url = SignedUrlBuilder::for_object(bucket_name, object_name)
+        .with_method(Method::PUT)
+        .with_expiration(Duration::from_secs(15 * 60)) // 15 minutes
+        .with_header("content-type", "application/octet-stream")
+        .sign_with(&signer)
+        .await
+        .map_err(anyhow::Error::new)?;
```
coryan: No need, you should be able to say:


```suggestion
        .await?;
```

src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
coryan: BTW, this is why the integration tests are failing

src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
joshuatants: Got it. I was following https://github.com/googleapis/google-cloud-rust/blob/b255800a76f8c57be3b91d5031b50b50b272c935/tests/storage/tests/driver.rs#L65 . Updated.

src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
joshuatants: Thanks Carlos.

Regarding the integration tests, is there a way I can get access to `rust-sdk-testing` so that I can see the build/test logs? It seems like I have to request access manually: https://screenshot.googleplex.com/9cwhiTTEzJQ5tZd

src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
joshuatants: Done.

joshuatants:
Sorry Carlos some more newbie questions:

1. Is there a way to use Storage Testbench? In C++ this could be set with various environment variables like `GOOGLE_CLOUD_CPP_STORAGE_TEST_JSON_ENDPOINT` or programmatically using RestEndpointOption or EndpointOption. Searching through the Rust repo turns up https://github.com/googleapis/google-cloud-rust/blob/b255800a76f8c57be3b91d5031b50b50b272c935/src/storage/src/storage/signed_url.rs#L822 and https://github.com/googleapis/google-cloud-rust/blob/b255800a76f8c57be3b91d5031b50b50b272c935/src/gax-internal/src/host.rs#L87 but these don't appear to be options to use the Testbench.
2. When runnig the integration tests locally they fail with the message
```Error: the service reports an error with code INVALID_ARGUMENT described as: This operation does not support custom billing projects at this time.

Caused by:
    code: 'Client specified an invalid argument', message: "This operation does not support custom billing projects at this time."
```

(For context, the tests that currently fail are for the samples `set_public_access_prevention_unspecified`, `change_file_storage_class`, and `create bucket for object ACL, retention`.)

I suspect this is because I'm currently using the project I set up [here](https://critique.corp.google.com/cl/855050833/depot/google3/configs/cloud/gong/org_hierarchy/google.com/billing_accounts.dir/Tech_Infra_Cloud/Storage/billing_account.Cloud_Storage/billing_associations.yaml). I'm creating a new project but am unsure of what Billing Account and Location to select ([screenshot](https://screenshot.googleplex.com/5pVTJBroGV5cunQ)). Is there a sample project I can follow?

Thank you.

coryan:
> Sorry Carlos some more newbie questions:
> 
> 1. Is there a way to use Storage Testbench?

Maybe. But I think the testbench is missing some headers that we depend on.  But why? The motivations for the testbench are: (1) to avoid flakes because you are using production, (2) to create specific errors.

For (2) we decided to launch HTTP (or gRPC) servers from within Rust. It is easy to do, and you can control sequence of messages much better than with the testbench (including timeouts).

For (1) same thing, we try to run as few tests as possible against production.  I guess the examples could be excluded, but wouldn't you want at least one run against production for the examples? And we only have one-ish test.

> 2. When runnig the integration tests locally they fail with the message
> 
> ```
> 
> Caused by:
>     code: 'Client specified an invalid argument', message: "This operation does not support custom billing projects at this time."
> ```

You are running into https://github.com/googleapis/google-cloud-rust/issues/4370 maybe it would be nice to automatically disable the tracing test if the billing project is not enabled, or disable the gRPC tests if it was.

src/storage/examples/src/objects/generate_upload_signed_url_v4.rs
coryan: Yes, I am moving the detailed discussion to an internal forum.

coryan (approved):

coryan:
> > Caused by:
> >     code: 'Client specified an invalid argument', message: "This operation does not support custom billing projects at this time."
> > ```
> 
> You are running into #4370 maybe it would be nice to automatically disable the tracing test if the billing project is not enabled, or disable the gRPC tests if it was.

Fixed in https://github.com/googleapis/google-cloud-rust/pull/4584

PR #4535, opened 2026-02-03
ci: split features build

coryan:


.gcb/scripts/features-gax-internal.sh
```diff
@@ -0,0 +1,37 @@
...
+set -ev
+
+rustup component add clippy
+cargo version
+rustup show active-toolchain -v
+
+cargo clean
+echo "==== google-cloud-gax-internal ===="
+cargo test --package google-cloud-gax-internal --no-default-features
+cargo doc  --package google-cloud-gax-internal --no-default-features --no-deps
```
PhongChuong, line 25: I'm assuming that `--no-deps` for `cargo doc ...` should have been added in the previous version of the script.

coryan: Yeah, it is a drive-by improvement: we don't need to generate the documents for all the deps, we just need to verify the crate's document work.

PR #4538, opened 2026-02-03
cleanup: full name for `google-cloud-recommender-v1`

coryan:


Cargo.toml
```diff
@@ -471,8 +471,9 @@ osconfig_v1               = { default-features = false, version = "1", path = "s
 kms                       = { default-features = false, version = "1", path = "src/generated/cloud/kms/v1", package = "google-cloud-kms-v1" }
 oslogin_common            = { default-features = false, version = "1", path = "src/generated/oslogin/common", package = "google-cloud-oslogin-common" }
 iam_v2                    = { default-features = false, version = "1", path = "src/generated/iam/v2", package = "google-cloud-iam-v2" }
-recommender               = { default-features = false, version = "1", path = "src/generated/cloud/recommender/v1", package = "google-cloud-recommender-v1" }
 accesscontextmanager_type = { default-features = false, version = "1", path = "src/generated/identity/accesscontextmanager/type", package = "google-cloud-identity-accesscontextmanager-type" }
+# Used by google-cloud-recommender-logging-v1
```
dbolduc, line 475: consider one comment for `# Used as a dependent by GAPICs` or `# Not a leaf GAPIC`

just cause nothing will keep these comments up to date, and not sure they add anything when the information is encoded in `librarian.yaml`.

coryan: Will do. Partly this is to keep the reformatting churn lower.

coryan:
> FWIW, I would prefer to review one PR with all the easy changes (like librarian changes + regeneration only).

If I knew what those are, maybe. But I am finding things to change as I go along.

PR #4547, opened 2026-02-04
cleanup: better messages in test assertions

coryan:
Add messages to things like `assert!(...is_ok());` and `assert!(...is_err());`

### DIFF
--- a/src/storage/benchmarks/random/src/args.rs
+++ b/src/storage/benchmarks/random/src/args.rs
@@ -113,7 +113,7 @@ mod tests {
     fn validate_success() -> anyhow::Result<()> {
         let args = Args::try_parse_from(["program", "--bucket-name=bucket", "json"])?;
         assert_eq!(args.bucket_name, "bucket");
-        let _ = args.validate()?;
+        args.validate()?;
         Ok(())
     }
 
--- a/src/storage/tests/scenarios/src/args.rs
+++ b/src/storage/tests/scenarios/src/args.rs
@@ -85,7 +85,7 @@ mod tests {
     fn validate_success() -> anyhow::Result<()> {
         let args = Args::try_parse_from(["program", "--bucket-name=bucket"])?;
         assert_eq!(args.bucket_name, "bucket");
-        let _ = args.validate()?;
+        args.validate()?;
         Ok(())
     }
 


src/gax-internal/src/http.rs
```diff
@@ -778,7 +778,14 @@ mod tests {
         let result = client
             .execute_streaming_once(builder, options, None, 0)
             .await;
-        assert!(result.is_err());
+        assert!(
+            result.is_err(),
+            "expected error, got successful stream: {:?}",
+            {
+                let (parts, _body) = result.unwrap().into_parts();
```
coryan, line 782: `result` is not `Debug`, so this is the best we can do.

src/auth/src/headers_util.rs
```diff
@@ -179,7 +179,7 @@ mod tests {
 
         let result = AuthHeadersBuilder::new(&cacheable_token).build();
 
-        assert!(result.is_ok());
+        assert!(result.is_ok(), "{result:?}");
         let cached_headers = result.unwrap();
```
dbolduc:
```rs
        let cached_headers = AuthHeadersBuilder::new(&cacheable_token).build()?;
```

dbolduc: and return an `anyhow::Result<()>` from the test

src/auth/src/headers_util.rs
```diff
@@ -201,7 +201,7 @@ mod tests {
 
         let result = AuthHeadersBuilder::new(&cacheable_token).build();
 
-        assert!(result.is_ok());
+        assert!(result.is_ok(), "{result:?}");
```
dbolduc: here and below.

src/auth/src/credentials/idtoken/verifier.rs
```diff
@@ -434,7 +434,7 @@ pub(crate) mod tests {
             .build();
 
         let result = verifier.verify(token).await;
-        assert!(result.is_ok());
+        assert!(result.is_ok(), "{result:?}");
         let claims = result.unwrap();
```
dbolduc:
```rs
let claims = verifier.verify(token).await?;
```

src/auth/src/credentials/idtoken/verifier.rs
```diff
@@ -546,7 +546,7 @@ pub(crate) mod tests {
             .build();
 
         let result = verifier.verify(token).await;
-        assert!(result.is_ok());
+        assert!(result.is_ok(), "{result:?}");
```
dbolduc:
```rs
let _ = verifier.verify(token).await?;
```

src/auth/src/credentials/service_account.rs
```diff
@@ -1002,7 +1003,7 @@ mod tests {
 
         let result = signer.sign(b"test").await;
 
-        assert!(result.is_ok());
+        assert!(result.is_ok(), "{result:?}");
```
dbolduc: or use `?`

dbolduc (approved):
Thanks, I think some of the `assert!(foo.is_ok()); let bar = foo.unwrap();` could just be `let bar = foo?;`

Your call on doing another pass.

src/auth/src/credentials/idtoken/verifier.rs
coryan: Fixed.

src/auth/src/credentials/idtoken/verifier.rs
coryan: Fixed.

src/auth/src/credentials/service_account.rs
coryan: Fixed.

src/auth/src/headers_util.rs
coryan: Fixed.

src/auth/src/headers_util.rs
coryan: Done, also simplified some of the assertions about `cached_headers.

dbolduc (approved):

PR #4548, opened 2026-02-04
fix(pubsub): nack messages when handler is dropped

### LINKED ISSUE #4514: Pub/Sub - Nack messages on drop of handler
The `Handler` has a `fn ack(self)` and a `fn nack(self)`. If the application drops the handler without calling either, we continue to hold the message in lease management, and continue to extend its lease, even though it can never be acknowledged by the application.

We should nack messages on drop. And we should consider just getting rid of `h.nack()`, when we can `drop(h)` instead.

---

To implement this, we will want to hold an inner state in an option[^1]. Something like:

```rs
struct AtLeastOnce {
  inner: Option<AtLeastOnceImpl>;
}

impl AtLeastOnce {
  fn ack(self) {
    if let Some(inner) = std::mem::take(self.inner) {
      inner.ack(); // or whatever.
    }
  }
}

impl Drop for AtLeastOnce {
  fn drop(&mut self) {
    if let Some(inner) = std::mem::take(self.inner) {
      inner.nack(); // or whatever.
    }
  }
}
```

[^1]: We cannot consume `self` for structs that have a manual drop. Unless we are willing to be `unsafe`.

haphungw:
Implement `Drop` for `AtLeastOnce` to provide a safety net `nack`.

Previously, if a handler went out of scope without being processed, the message would remain leased. Now, they are automatically rejected if the application fails to explicitly acknowledge or reject them. 

Note: The next PR will refactor by removing the explicit `nack()` in favor of this automatic behavior.

Fix #4514

### DIFF
--- a/src/pubsub/src/subscriber/handler.rs
+++ b/src/pubsub/src/subscriber/handler.rs
@@ -111,11 +111,10 @@ impl AtLeastOnce {
 }
 
 impl Drop for AtLeastOnce {
-    /// Automatically rejects the message if it hasn't been acknowledged yet.
+    /// Rejects the message associated with this handler.
     ///
-    /// This ensures that messages are redelivered even if the application
-    /// forgets to call `ack()`, preventing leaked messages in the lease
-    /// management system.
+    /// The message will be removed from this `Subscriber`'s lease management.
+    /// The service will redeliver this message, possibly to another client.
     fn drop(&mut self) {
         if let Some(inner) = self.inner.take() {
             inner.nack();


src/pubsub/src/subscriber/handler.rs
```diff
@@ -52,32 +52,74 @@ impl Handler {
...
+    pub(crate) fn ack_id(&self) -> &str {
+        self.inner
+            .as_ref()
+            .map(|i| i.ack_id.as_str())
+            .unwrap_or_default()
+    }
+}
+
+impl Drop for AtLeastOnce {
+    /// Automatically rejects the message if it hasn't been acknowledged yet.
```
dbolduc: nit: I think we should copy the documentation from the existing `nack()` here, especially if we are going to get rid of `fn nack()` altogether (which I think we should do).

src/pubsub/src/subscriber/handler.rs
```diff
@@ -52,32 +52,74 @@ impl Handler {
...
     /// The service will redeliver this message, possibly to another client.
-    pub fn nack(self) {
-        let _ = self.ack_tx.send(AckResult::Nack(self.ack_id));
+    pub fn nack(mut self) {
+        if let Some(inner) = self.inner.take() {
+            inner.nack();
+        }
+    }
+
+    #[cfg(test)]
```
dbolduc, line 104: Good thinking.

dbolduc (approved):

PR #4549, opened 2026-02-04
cleanup(pubsub)!: remove explicit nack

### LINKED ISSUE #4514: Pub/Sub - Nack messages on drop of handler

haphungw:
Simplify the pubsub subscriber API by removing the explicit `nack` method and replacing it with an idiomatic `Drop` implementation. Update unit tests and documentation accordingly.

Fixes #4514

### DIFF
--- a/src/pubsub/src/subscriber/handler.rs
+++ b/src/pubsub/src/subscriber/handler.rs
@@ -136,19 +136,6 @@ mod tests {
     }
 
     #[test]
-    fn handler_implicit_drop() -> anyhow::Result<()> {
-        let (ack_tx, mut ack_rx) = unbounded_channel();
-        {
-            let _h = Handler::AtLeastOnce(AtLeastOnce::new(test_id(1), ack_tx));
-            assert_eq!(ack_rx.try_recv(), Err(TryRecvError::Empty));
-        }
-        let ack = ack_rx.try_recv()?;
-        assert_eq!(ack, AckResult::Nack(test_id(1)));
-
-        Ok(())
-    }
-
-    #[test]
     fn at_least_once_ack() -> anyhow::Result<()> {
         let (ack_tx, mut ack_rx) = unbounded_channel();
         let h = AtLeastOnce::new(test_id(1), ack_tx);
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/handler.rs
```diff
@@ -148,7 +128,20 @@ mod tests {
...
-        h.nack();
+        drop(h);
+        let ack = ack_rx.try_recv()?;
+        assert_eq!(ack, AckResult::Nack(test_id(1)));
+
+        Ok(())
+    }
+
+    #[test]
+    fn handler_implicit_drop() -> anyhow::Result<()> {
```
dbolduc: nit: I would remove this test, given that we have the `handler_nack` test. It is more testing the Rust language than our code.

src/pubsub/src/subscriber/handler.rs
```diff
@@ -174,7 +167,20 @@ mod tests {
...
-        h.nack();
+        drop(h);
+        let ack = ack_rx.try_recv()?;
+        assert_eq!(ack, AckResult::Nack(test_id(1)));
+
+        Ok(())
+    }
+
+    #[test]
+    fn at_least_once_implicit_drop() -> anyhow::Result<()> {
```
dbolduc: ditto

src/pubsub/src/subscriber/handler.rs
haphungw: done!

src/pubsub/src/subscriber/handler.rs
haphungw: done!

dbolduc:
FYI: I changed the PR title to include a `!` which signals a breaking change, as we are removing a method from the API. Note that this is a preview version, so breaking changes are allowed / expected.

Issue #4541, opened 2026-02-03
google-cloud-auth: permanent credential failure after transient network failures


johnsonj:
## Summary

The `google-cloud-auth` library can get into a state where authentication credentials are permanently marked as failed after transient token refresh failures (for example, DNS fails for a period of time). Once in this state, the client never recovers, even after the underlying network issue is resolved.

The actual authentication used to create the client is still valid, so it would be fine to retry, but the only workaround due to this behavior is to recreate the entire client.

This error shows up as: `cannot create the authentication headers All retry attempts to fetch the token were exhausted. Subsequent calls with this credential will also fail. and future attempts will not succeed`

## Reproduction

This repro uses a mock GCE Metadata Service (MDS) server that proxies to real ADC credentials but with a shorter expiration time. Initially, it returns fine, but for a period of time it immediately closes the TCP connection.

[gist here](https://gist.github.com/johnsonj/f28c5d67957b06353a66ecb1657a1c64)

```
[0-4s]  ✓ SUCCESS - Real GCS requests work
[5-19s] ✗ OUTAGE - Mock server drops connections
[20s+]  ✗ RECOVERY - Server works, but client never asks for tokens!

^^^ BUG: Client is permanently poisoned ^^^
```

## Root Cause

In [`src/auth/src/retry.rs`](https://github.com/googleapis/google-cloud-rust/blob/5bde8954738919bb47213a2ea1fc261a066c3c44/src/auth/src/retry.rs#L119-L132):

```rust
// After exhausting retries, the error is marked as non-transient
let transient = remaining > 0;
return Err(CredentialsError::new(transient, source));
```

The problem: once `transient = false`, this flag is permanent. The cached credential state is never cleared, so the library never attempts to refresh credentials again.

## Workaround

Detect permanent auth failures and recreate the GCS client:

```rust
if is_permanent_auth_failure(&error) {
     client = Storage::builder().build().await?; // Recreate!
}

...
fn is_permanent_auth_failure(error: &GcsError) -> bool {
    use std::error::Error;

    if !error.is_authentication() {
        return false;
    }

    let mut source = error.source();
    while let Some(err) = source {
        if let Some(cred_err) = err.downcast_ref::<CredentialsError>() {
            return !cred_err.is_transient();
        }
        source = err.source();
    }

    false
}

```

## Environment

- `google-cloud-auth`: 1.x
- `google-cloud-gax`: 1.4.x
- `google-cloud-storage`: 1.x
[...]

PR #4551, opened 2026-02-04
fix: token cache exits after exhaustive retries

alvarowolfx:
Towards #4541

### DIFF
--- a/src/auth/src/retry.rs
+++ b/src/auth/src/retry.rs
@@ -294,7 +294,10 @@ mod tests {
 
         let error = provider.token().await.unwrap_err();
         assert!(error.is_transient(), "{error:?}");
-        assert!(error.to_string().contains(constants::RETRY_EXHAUSTED_ERROR), "{error:?}");
+        assert!(
+            error.to_string().contains(constants::RETRY_EXHAUSTED_ERROR),
+            "{error:?}"
+        );
         let original_error = find_source_error::<CredentialsError>(&error).unwrap();
         assert!(original_error.is_transient(), "{original_error:?}");
     }


src/auth/src/credentials/idtoken/user_account.rs
```diff
@@ -422,7 +422,7 @@ mod tests {
             .build()?;
 
         let err = credentials.id_token().await.unwrap_err();
-        assert!(!err.is_transient());
+        assert!(err.is_transient());
```
coryan:
```suggestion
        assert!(err.is_transient(), "{err:?}");
```

src/auth/src/credentials/external_account.rs
```diff
@@ -1959,7 +1959,7 @@ mod tests {
             .unwrap();
 
         let err = creds.headers(Extensions::new()).await.unwrap_err();
-        assert!(!err.is_transient());
+        assert!(err.is_transient());
```
coryan: Ditto

src/auth/src/credentials/external_account.rs
```diff
@@ -2084,7 +2084,7 @@ mod tests {
...
```
coryan: Ditto.

src/auth/src/credentials/mds.rs
```diff
@@ -483,7 +483,7 @@ mod tests {
             .build_token_provider();
 
         let err = provider.token().await.unwrap_err();
-        assert!(!err.is_transient());
+        assert!(err.is_transient());
```
coryan: And so on...

src/auth/src/retry.rs
```diff
@@ -493,6 +494,26 @@ mod tests {
...
+        let error = CredentialsError::from_msg(transient, message);
+        let error = gax::error::Error::authentication(error);
+        let error_string = error.to_string();
+
+        // 2. Call the function under test.
+        let credentials_error =
+            TokenProviderWithRetry::<MockTokenProvider>::map_retry_error(error);
+
+        // 3. Assert that the resulting error is transient or not like the original error and wraps the original error.
+        assert_eq!(credentials_error.is_transient(), transient);
```
coryan:
```suggestion
        assert_eq!(credentials_error.is_transient(), transient, "{credentials_error:?}");
```

src/auth/src/token_cache.rs
```diff
@@ -129,15 +129,20 @@ async fn refresh_task<T>(
...
                 break;
             }
-            None => {
+            Err(err) => {
                 // The retry policy has been used already by the inner token provider.
-                // If it ended in an error, just quit the background task.
-                break;
+                // If it ended in an error, the background task will wait for a while
+                // and try again. This allows the task to eventually recover if the
+                // error was transient but exhausted all the retries.
```
coryan: This comment does not read right.

```suggestion
            Err(err) => {
                // On transient errors, even if the retry policy is exhausted,
                // we want to continue running this retry loop.
                // This loop cannot stop because that may leave the
                // credentials in an unrecoverable state (see #...).
                // We considered using a notification to wake up the next time
                // a caller wants to retrieve a token, but that seemed prone to
                // deadlocks. We may implement this as an improvement (#....).
                // On permanent errors, then there is really no point in trying
                // again, by definition of "permanent".  If the error was misclassified
                // as permanent, that is a bug in the retry policy and better fixed
                // there than implemented as a workaround here.
```

With prompts the question: are we certain our default policy does not misclassify some errors as permanent? Can you check?

src/auth/src/retry.rs
```diff
@@ -133,14 +133,15 @@ where
...
-        let msg = match e
+        match e
             .source()
             .and_then(|s| s.downcast_ref::<CredentialsError>())
         {
-            Some(cred_error) if cred_error.is_transient() => constants::RETRY_EXHAUSTED_ERROR,
-            _ => constants::TOKEN_FETCH_FAILED_ERROR,
-        };
-        CredentialsError::new(false, msg, e)
+            Some(cred_error) if cred_error.is_transient() => {
```
coryan, line 142: unrelated... I think our retry loop is missing a trick. It should not wrap the `CredentialsError`, it should operate on `gax::error::Error` and `CredentialsError` using some trait with an associated type.

coryan (changes requested):

src/auth/src/token_cache.rs
alvarowolfx: The retry loop always map errors to gax:authentication errors, so I think in that part is correct, so I don't  any other gax error being raised. If was not that I would be worried because  https://github.com/googleapis/google-cloud-rust/pull/4551/changes#diff-43612f24f6413f8e5304d6af3e912553d7913ceea55c203c549db570b52d0d30L133 treat all non auth errors as permanent. 

In the auth side, I was doing some audit on every place that creates `CredentialsError::from_msg(false, ...)`, `CredentialsError::new(false,...)`, `CredentialsError::from_source(`  or `CredentialsError::from_http_response` and they are mostly consider non transient errors the ones related to transport and decoding errors when fetching data from remote services, so I think that's correct. 

I could not find yet any case of misclassification, but you raised a valid point

alvarowolfx: I updated the comment and created an issue to track future improvement on the retry loop.

src/auth/src/constants.rs
```diff
@@ -29,7 +29,7 @@ pub(crate) const JWT_TOKEN_TYPE: &str = "urn:ietf:params:oauth:token-type:jwt";
 /// SAML2 Token OAuth Token Type
 pub(crate) const SAML2_TOKEN_TYPE: &str = "urn:ietf:params:oauth:token-type:saml2";
 
-pub(crate) const RETRY_EXHAUSTED_ERROR: &str = "All retry attempts to fetch the token were exhausted. Subsequent calls with this credential will also fail.";
+pub(crate) const RETRY_EXHAUSTED_ERROR: &str = "All retry attempts to fetch the token were exhausted with transient errors. Subsequent calls with this credential might succeed.";
```
coryan: I find this confusing, but I am tired. Maybe another team member can help you wordsmith this a bit?

src/auth/src/token_cache.rs
coryan: I am not sure the parsing errors are permanent.  We sent a request, got a bad response (sometimes the load balancers like to return silly things). The next attempt may succeed. OTOH, maybe the endpoint is configured wrong and we got a nice cocktail recipe 🤷 

The error classification is too close to the source, we should let the retry policy decide what errors are transient vs. permanent, instead of hard-coding that decision in each function.

src/auth/src/constants.rs
alvarowolfx: @dbolduc or @sai-sunder-s can help a bit here ?

dbolduc: It is hard to explain the double retry loop in auth. Do we need to add extra context to the error?

I think this retry exhausted context was to explain why we stopped retrying a transient error. Now that we keep retrying transient errors, I think it is unnecessary.

src/auth/src/token_cache.rs
```diff
@@ -101,16 +101,16 @@ async fn refresh_task<T>(
 {
     loop {
         let token_result = token_provider.token().await;
-        let expiry = token_result.as_ref().ok().map(|t| t.expires_at);
-        let tagged = token_result.map(|token| {
+        let expiry = token_result.as_ref().map(|t| t.expires_at);
+        let tagged = token_result.clone().map(|token| {
```
dbolduc, line 105: silly nit: you can avoid the clone.

```rs
enum Refresh {
  Deadline(Instant),
  Never,
  Soon,
}

let refresh = match token_result.as_ref() {
  Ok(t) => {
    match t.expires_at {
      Some(d) => Refresh::Deadline(d),
      None => Refresh::Never,
    }
  },
  Err(e) if e.is_transient() => Refresh::Soon,
  Err(_) => Refresh::Never,
};
```

But maybe you think saving one clone is silly for a token that is cloned on every request anyway

dbolduc (commented):
(not important. I just like doing puzzles)

src/auth/src/constants.rs
sai-sunder-s: How about "Fetching the token resulted in transient errors. Subsequent calls with this credential might succeed."

No need to mention that we retried already and we will retry more.

src/auth/src/retry.rs
```diff
@@ -292,7 +293,7 @@ mod tests {
             .build(mock_provider);
 
         let error = provider.token().await.unwrap_err();
-        assert!(!error.is_transient());
+        assert!(error.is_transient());
```
coryan: Please make a pass fixing these assertions:


```suggestion
        assert!(error.is_transient(), "{error:?}");
```

src/auth/src/constants.rs
```diff
@@ -29,7 +29,7 @@ pub(crate) const JWT_TOKEN_TYPE: &str = "urn:ietf:params:oauth:token-type:jwt";
 /// SAML2 Token OAuth Token Type
 pub(crate) const SAML2_TOKEN_TYPE: &str = "urn:ietf:params:oauth:token-type:saml2";
 
-pub(crate) const RETRY_EXHAUSTED_ERROR: &str = "All retry attempts to fetch the token were exhausted. Subsequent calls with this credential will also fail.";
+pub(crate) const RETRY_EXHAUSTED_ERROR: &str = "Request to fetch the token failed after multiple retries with transient errors. Subsequent calls with this credential might succeed.";
```
coryan: I liked @sai-sunder-s 's suggestion.

src/auth/src/token_cache.rs
coryan: Can you open a bug to track these things? It may be too late to change, but we should think it through.

PR #4557, opened 2026-02-04
refactor(pubsub): simplify ConcurrentBatchActor

PhongChuong:
This PR simplify ConcurrentBatchActor by removing pause and the pending_msgs queue. A followup PR will cleanup SequentialBatchActor.

### DIFF
--- a/src/pubsub/src/publisher/actor.rs
+++ b/src/pubsub/src/publisher/actor.rs
@@ -550,7 +550,7 @@ mod tests {
             msg: PubsubMessage::new().set_data("hello"),
             tx: publish_tx,
         };
-        let _ = actor_tx.send(ToBatchActor::Publish(bundle))?;
+        actor_tx.send(ToBatchActor::Publish(bundle))?;
         let res = publish_rx.await;
         assert!(matches!(res, Ok(Ok(ref msg)) if msg == "hello"), "{res:?}");
         Ok(())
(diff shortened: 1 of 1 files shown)


src/pubsub/src/publisher/actor.rs
```diff
@@ -490,14 +496,28 @@ mod tests {
         }
     }
 
+    fn publish_ok(
+        req: crate::model::PublishRequest,
+        _options: gax::options::RequestOptions,
+    ) -> gax::Result<gax::response::Response<crate::model::PublishResponse>> {
+        assert_eq!(req.topic, TOPIC);
```
dbolduc: You can use `.withf(|r, _o| r.topic == TOPIC)` on the mock expectation and avoid the assertion in a common function.

src/pubsub/src/publisher/actor.rs
```diff
@@ -507,4 +527,90 @@ mod tests {
...
+            )
+            .run(),
+        );
+
+        let (publish_tx, publish_rx) = tokio::sync::oneshot::channel();
+        let bundle = BundledMessage {
+            msg: PubsubMessage::new().set_data("hello"),
+            tx: publish_tx,
+        };
+        let res = actor_tx.send(ToBatchActor::Publish(bundle));
```
dbolduc: nit: use `?`

```rs
let _ = actor_tx.send(ToBatchActor::Publish(bundle))?;
```

src/pubsub/src/publisher/actor.rs
```diff
@@ -507,4 +527,90 @@ mod tests {
...
+            )
+            .run(),
+        );
+
+        // Flush on empty.
+        let (flush_tx, flush_rx) = tokio::sync::oneshot::channel();
+        let res = actor_tx.send(ToBatchActor::Flush(flush_tx));
+        assert!(res.is_ok(), "{res:?}");
+        let res = flush_rx.await;
+        assert!(res.is_ok(), "{res:?}");
```
dbolduc: same on preferring `?`.

dbolduc: There are more below. You can find them.

src/pubsub/src/publisher/actor.rs
PhongChuong: Thanks. Good to know.

Issue #3017, opened 2025-08-24
Write the critical hand-crafted examples


coryan:
Manually write the framework for the examples, and a few critical examples.

Ideally we can use the gemini-cli trick to write most hand-crafted examples once those are done.

suzmue:
Full list of pubsub samples in google-cloud-cpp

<details>
<summary>
<code>$ sed "s/START //" <(git -C ../google-cloud-cpp/ grep -h -o 'START pubsub_[a-z0-9_]*' -- *.cc | sort -u)</code>
</summary>
<code>pubsub_commit_avro_schema
pubsub_commit_proto_schema
pubsub_create_avro_schema
pubsub_create_bigquery_subscription
pubsub_create_cloud_storage_subscription
pubsub_create_proto_schema
pubsub_create_pull_subscription
pubsub_create_push_subscription
pubsub_create_subscription_with_exactly_once_delivery
pubsub_create_subscription_with_filter
pubsub_create_topic
pubsub_create_topic_with_aws_msk_ingestion
pubsub_create_topic_with_azure_event_hubs_ingestion
pubsub_create_topic_with_cloud_storage_ingestion
pubsub_create_topic_with_confluent_cloud_ingestion
pubsub_create_topic_with_kinesis_ingestion
pubsub_create_topic_with_schema
pubsub_create_topic_with_schema_revisions
pubsub_create_unwrapped_push_subscription
pubsub_dead_letter_create_subscription
pubsub_dead_letter_delivery_attempt
pubsub_dead_letter_remove
pubsub_dead_letter_update_subscription
pubsub_delete_schema
pubsub_delete_schema_revision
pubsub_delete_subscription
pubsub_delete_topic
pubsub_detach_subscription
pubsub_enable_subscription_ordering
pubsub_get_schema
pubsub_get_schema_revision
pubsub_get_subscription_policy
pubsub_get_topic_policy
pubsub_list_schema_revisions
pubsub_list_schemas
pubsub_list_subscriptions
pubsub_list_topic_subscriptions
pubsub_list_topics
pubsub_optimistic_subscribe
pubsub_publish
pubsub_publish_avro_records
pubsub_publish_custom_attributes
pubsub_publish_messages_error_handler
pubsub_publish_proto_messages
pubsub_publish_with_error_handler
pubsub_publish_with_ordering_keys
pubsub_publisher_batch_settings
pubsub_publisher_concurrency_control
pubsub_publisher_flow_control
[...]

PR #4572, opened 2026-02-05
test(pubsub): add `quickstart_subscriber` sample

haphungw:
Implement `quickstart_subscriber` and include an integration test.

For #3017

### DIFF
--- a/src/pubsub/examples/tests/integration.rs
+++ b/src/pubsub/examples/tests/integration.rs
@@ -54,10 +54,7 @@ mod tests {
         let (subscription_admin, subscription) =
             pubsub_samples::create_test_subscription(topic.name.clone()).await?;
 
-        let topic_id = topic.name.split('/').last().unwrap();
         let subscription_id = subscription.name.split('/').last().unwrap();
-
-        quickstart_publisher::sample(&project_id, topic_id).await?;
 
         let result = quickstart_subscriber::sample(&project_id, subscription_id).await;
 


src/pubsub/examples/src/quickstart_subscriber.rs
```diff
@@ -0,0 +1,42 @@
...
+    // In production, applications typically process messages indefinitely in a long-running loop.
+    let deadline = tokio::time::Instant::now() + tokio::time::Duration::from_secs(10);
+
+    while let Ok(Some(item)) = tokio::time::timeout_at(deadline, session.next()).await {
+        let (message, handler) = item?;
+
+        let data = String::from_utf8_lossy(&message.data);
+        println!("received message: {data}");
+
+        // Acknowledge the message so it isn't redelivered.
```
dbolduc: nit:
```suggestion
        println!("Received message: {message:?}");
```

- print the whole thing because the message has other fields like `ordering_key` and `attributes` that might be interesting to the application
- I am not sure this is the place to explain `ack`ing. I peeked at the other samples, and they don't bother.

src/pubsub/examples/src/quickstart_subscriber.rs
```diff
@@ -0,0 +1,42 @@
...
+    let mut session = client.streaming_pull(subscription_name).start();
+
+    println!("listening for messages...");
+
+    // For demonstration purposes, this example terminates after 10 seconds.
+    // In production, applications typically process messages indefinitely in a long-running loop.
+    let deadline = tokio::time::Instant::now() + tokio::time::Duration::from_secs(10);
+
+    while let Ok(Some(item)) = tokio::time::timeout_at(deadline, session.next()).await {
+        let (message, handler) = item?;
```
dbolduc: comment: This is a little verbose, but it might be the best we can do with the current API.

Which reminds me, we should do: #4574

src/pubsub/examples/src/quickstart_subscriber.rs
```diff
@@ -0,0 +1,42 @@
...
+
+// [START pubsub_quickstart_subscriber]
+use google_cloud_pubsub::client::Subscriber;
+
+pub async fn sample(project_id: &str, subscription_id: &str) -> anyhow::Result<()> {
+    let subscription_name = format!("projects/{project_id}/subscriptions/{subscription_id}");
+    let client = Subscriber::builder().build().await?;
+    let mut session = client.streaming_pull(subscription_name).start();
+
+    println!("listening for messages...");
```
dbolduc, line 24: nit: 
```suggestion
    println!("Listening for messages...");
```

src/pubsub/examples/tests/integration.rs
```diff
@@ -45,4 +45,28 @@ mod tests {
...
         result
     }
+
+    #[tokio::test]
+    async fn quickstart_subscriber() -> anyhow::Result<()> {
+        let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
+
+        let (topic_admin, topic) = pubsub_samples::create_test_topic().await?;
+        let (subscription_admin, subscription) =
+            pubsub_samples::create_test_subscription(topic.name.clone()).await?;
```
dbolduc, line 55: comment: No action needed now, but an FYI: we might want to combine this with the existing `quickstart_publisher` you added.

Sometimes there are quotas around creating GCP resources, so it's nice to do it once, and share them across tests.

src/pubsub/examples/tests/integration.rs
```diff
@@ -45,4 +45,28 @@ mod tests {
...
+
+        let topic_id = topic.name.split('/').last().unwrap();
+        let subscription_id = subscription.name.split('/').last().unwrap();
+
+        quickstart_publisher::sample(&project_id, topic_id).await?;
```
dbolduc: nit: I don't think we need to run the publisher quickstart.

The subscriber does not crash and burn when there are no messages. Or at least I hope it does not.

src/pubsub/examples/src/quickstart_subscriber.rs
```diff
@@ -0,0 +1,42 @@
...
+        let (message, handler) = item?;
+
+        let data = String::from_utf8_lossy(&message.data);
+        println!("received message: {data}");
+
+        // Acknowledge the message so it isn't redelivered.
+        handler.ack();
+    }
+
+    println!("done listening for messages");
```
dbolduc, line 36: nit: I'd start with a capital.

src/pubsub/examples/src/quickstart_subscriber.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub async fn sample(project_id: &str, subscription_id: &str) -> anyhow::Result<()> {
+    let subscription_name = format!("projects/{project_id}/subscriptions/{subscription_id}");
+    let client = Subscriber::builder().build().await?;
+    let mut session = client.streaming_pull(subscription_name).start();
+
+    println!("listening for messages...");
+
+    // For demonstration purposes, this example terminates after 10 seconds.
+    // In production, applications typically process messages indefinitely in a long-running loop.
```
dbolduc: nit: at least drop "In production"

```suggestion
    // For demonstration purposes, this example terminates after 10 seconds.
    // Applications typically process messages indefinitely in a long-running loop.
```

or maybe be even more direct?

```suggestion
    // Terminate the example after 10 seconds. Applications typically process
    // messages indefinitely in a long-running loop.
```

I am fine with either.

src/pubsub/examples/src/quickstart_subscriber.rs
```diff
@@ -0,0 +1,42 @@
...
+    let deadline = tokio::time::Instant::now() + tokio::time::Duration::from_secs(10);
```
dbolduc: We should use `std::time::Instant`, `std::time::Duration` in the example. Probably with a `use std::time::{Duration, Instant};` up top.

Internally, we like `tokio::time::Instant` in most places because it lets us control the clock in our unit tests.

src/pubsub/examples/src/quickstart_subscriber.rs
haphungw: agreed, `ack` should be self-explanatory.

src/pubsub/examples/src/quickstart_subscriber.rs
haphungw: this seems pretty inconsistent across our samples.

src/pubsub/examples/src/quickstart_subscriber.rs
haphungw: done!

PR #4573, opened 2026-02-05
impl(pubsub): add options for lease RPCs

dbolduc:
Part of the work for #4471

Introduce retry options for lease operations that can withstand an hourly connection reset.

Note that the options are not in use yet. The next PR will hook them up to the client.

### DIFF
--- a/src/pubsub/src/subscriber/retry_policy.rs
+++ b/src/pubsub/src/subscriber/retry_policy.rs
@@ -13,10 +13,13 @@
 // limitations under the License.
 
 use crate::Error;
+use gax::backoff_policy::BackoffPolicy;
 use gax::error::rpc::Code;
-use gax::retry_policy::RetryPolicy;
+use gax::options::RequestOptions;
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/retry_policy.rs
```diff
@@ -175,4 +222,113 @@ mod tests {
...
+    }
+
+    #[test]
+    fn at_least_once_options() {
+        let o = super::at_least_once_options(42);
+        verify_policies(o, 42);
+    }
+
+    #[track_caller]
+    pub(in super::super) fn verify_policies(o: RequestOptions, grpc_subchannel_count: u32) {
```
dbolduc, line 254: Note to the reviewer:

TIL: `#[track_caller]` which will display the caller's line number on failed asserts in this function. It does not display the line number that fails in this file, which is why I went for very descriptive error messages.

This is factored out because I will reuse it from `leaser.rs` in my next PR.

cc: @PhongChuong because you might get a kick out of this.

PR #4575, opened 2026-02-05
fix(pubsub): retry acks when channel closes

### LINKED ISSUE #4471: Retry at-least-once lease operations that fail when a connection is closed

dbolduc:
Fixes #4471 

Use the at least once policies in the default leaser. The policies depend on the gRPC subchannel count, so plumb that through all the layers.

The effect is that we retry acks that fail with transport errors when a channel closes, and the application avoids frozen messages every hour or so.

### DIFF
--- a/src/pubsub/src/subscriber/builder.rs
+++ b/src/pubsub/src/subscriber/builder.rs
@@ -23,17 +23,24 @@ pub struct StreamingPull {
     pub(super) inner: Arc<Transport>,
     pub(super) subscription: String,
     pub(super) client_id: String,
+    pub(super) grpc_subchannel_count: usize,
     pub(super) ack_deadline_seconds: i32,
     pub(super) max_outstanding_messages: i64,
     pub(super) max_outstanding_bytes: i64,
 }
 
 impl StreamingPull {
-    pub(super) fn new(inner: Arc<Transport>, subscription: String, client_id: String) -> Self {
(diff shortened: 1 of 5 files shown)


src/pubsub/src/subscriber/builder.rs
```diff
@@ -23,17 +23,24 @@ pub struct StreamingPull {
...
     pub(super) max_outstanding_bytes: i64,
 }
 
 impl StreamingPull {
-    pub(super) fn new(inner: Arc<Transport>, subscription: String, client_id: String) -> Self {
+    pub(super) fn new(
+        inner: Arc<Transport>,
+        subscription: String,
+        client_id: String,
+        grpc_subchannel_count: usize,
```
coryan, line 37: These three fields repeat in a number of places.  Time to refactor to a struct?

coryan (approved):

src/pubsub/src/subscriber/builder.rs
dbolduc: Pass. Maybe next time.

PR #4577, opened 2026-02-05
test(storage): conditionally skip `rename_folder()` tests

coryan:
The test will fail if the credentials are using a custom billing project, so just skip the test in that case.

Part of the work for #4370


coryan:
PTAL, I also needed to fix the examples.

PR #4579, opened 2026-02-05
tests(storage): skip signed URLs examples when needed

coryan:
Signed URLs do not work with user credentials, it makes no sense to run them when using that credential type.

### DIFF
--- a/src/storage/examples/src/lib.rs
+++ b/src/storage/examples/src/lib.rs
@@ -511,11 +511,25 @@ pub async fn run_object_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
     objects::remove_file_owner::sample(&control, &id, &service_account).await?;
     tracing::info!("running set_object_retention_policy example");
     objects::set_object_retention_policy::sample(&control, &id).await?;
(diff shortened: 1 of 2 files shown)


src/storage/examples/src/lib.rs
```diff
@@ -511,11 +511,25 @@ pub async fn run_object_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
...
+            return Ok(());
+        }
+        Err(e) => return Err(e.into()),
+        Ok(_) => {}
+    };
+    let id = random_bucket_id();
     tracing::info!("running generate_signed_url_v4 example");
     objects::generate_signed_url_v4::sample(&id, "object-to-read").await?;
     tracing::info!("running generate_upload_signed_url_v4 example");
     objects::generate_upload_signed_url_v4::sample(&id, "object-to-upload").await?;
```
dbolduc, line 532: Q: does it matter that these objects don't exist?

Guess not.

coryan: Correct, it does not matter.

PR #4581, opened 2026-02-05
test(pubsub): add more publisher actor tests

PhongChuong:


src/pubsub/src/publisher/actor.rs
```diff
@@ -527,70 +618,192 @@ mod tests {
...
+        for (k, v) in publish_rxs {
+            let res = v.await;
+            assert!(
+                matches!(res, Ok(Ok(ref msg)) if *msg == k),
+                "got {res:?}, expected {k:?}"
+            );
+        }
+        assert_eq!(
+            start.elapsed(),
+            Duration::from_millis(50), // 10msg / 2 msg/batch * 10ms/batch = 50ms
```
dbolduc: nit: I am a computer and need parentheses
```suggestion
            Duration::from_millis(50), // 10msg / (2 msg/batch) * 10ms/batch = 50ms
```

or just

```suggestion
            Duration::from_millis(50), // 5 batches * 10ms/batch = 50ms
```

or just make the math obvious with constants for:

```rs
const EXPECTED_BATCHES: u32 = 5;
const MESSAGES_PER_BATCH: u32 = 2;
const TIME_PER_BATCH: Duration = Duration::from_secs(10);
```

src/pubsub/src/publisher/actor.rs
```diff
@@ -527,70 +618,192 @@ mod tests {
...
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn sequential_actor_flush() -> anyhow::Result<()> {
+        let mut mock = MockGapicPublisherWithFuture::new();
+        mock.expect_publish()
+            .withf(|req, _o| req.topic == TOPIC)
+            .times(5)
+            .returning({
+                |r, o| {
```
dbolduc, line 761: optional: This test doesn't really verify that the order the messages go in is the order they show up in the requests.

You could send the requests out on an `unbounded_channel`...

https://github.com/googleapis/google-cloud-rust/blob/76753ba168b8cb2de397addaed3b2ab8aee6bbf6/src/pubsub/src/subscriber/session.rs#L395-L397

and verify the requests contained `[m1, m2], [m3, m4], ... , [m9, m10]` at the end of the test:

https://github.com/googleapis/google-cloud-rust/blob/76753ba168b8cb2de397addaed3b2ab8aee6bbf6/src/pubsub/src/subscriber/session.rs#L422-L425

dbolduc (approved):

src/pubsub/src/publisher/actor.rs
PhongChuong: Thanks. Let me think about how to do this cleanly for all sequential actor tests and add it in a future PR.

PR #4585, opened 2026-02-06
feat(pubsub): include subscription name in StreamingPull header

### LINKED ISSUE #4470: Set routing header for `StreamingPull` RPC
The routing header (`x-goog-request-params`) should include the requests `subscription` field. i.e. the value should be `subscription=projects/my-project/subscriptions/my-subscription`.

https://github.com/googleapis/google-cloud-rust/blob/aa9f9f74f27f26f3c230ff0f3b29eae563306e82/src/pubsub/src/subscriber/transport.rs#L70

It will sadly take a mini refactor to plumb this, as the interface only accepts a channel.

My first instinct is to just add a `subscription: &str` to the interface (assuming lifetimes aren't an issue).

haphungw:
Fix #4470

### DIFF
--- a/src/pubsub/src/subscriber/session.rs
+++ b/src/pubsub/src/subscriber/session.rs
@@ -1046,8 +1046,6 @@ mod tests {
     async fn routing_header() -> anyhow::Result<()> {
         let mut mock = MockSubscriber::new();
 
-        let subscription = "projects/p/subscriptions/s";
-
         mock.expect_streaming_pull().return_once(move |request| {
             let metadata = request.metadata();
             assert_eq!(
@@ -1062,7 +1060,11 @@ mod tests {
         let (endpoint, _server) = start("0.0.0.0:0", mock).await?;
         let client = test_client(endpoint).await?;
 
-        let _ = client.streaming_pull(subscription).start().next().await;
+        let _ = client
+            .streaming_pull("projects/p/subscriptions/s")
+            .start()
+            .next()
+            .await;
 
         Ok(())
     }


src/pubsub/src/subscriber/stub.rs
```diff
@@ -68,6 +69,7 @@ pub(super) mod tests {
             type Stream = MockStream;
             async fn streaming_pull(
                 &self,
+                subscription: &str,
```
dbolduc: comment: This works, and I am ok with it.

Some background: this is a handwritten interface for a bidirectional gRPC stream added onto a fully generated struct.

We have not come up with a generic solution for bidirectional streams yet that we can generate.

So `subscription` works but it won't help us solve the general problem. (Which might look like accepting an initial request? (although I don't think in general that is even known. Ugh. It is a hard problem that probably @coryan will have to tackle next quarter)).

dbolduc: Consider accepting the whole `x_goog_request_params`

haphungw: done!

Cargo.lock
```diff
@@ -34,9 +34,9 @@ checksum = "5192cca8006f1fd4f7237516f40fa183bb07f8fbdfedaa0036de5ea9b0b45e78"
 
 [[package]]
```
dbolduc, line 35: Can you revert these changes? We typically do a `cargo update` in its own PR just to keep the commits focused.

src/pubsub/src/subscriber/session.rs
```diff
@@ -1041,4 +1041,31 @@ mod tests {
...
+        mock.expect_streaming_pull().return_once(move |request| {
+            let metadata = request.metadata();
+            assert_eq!(
+                metadata
+                    .get("x-goog-request-params")
+                    .expect("routing header missing"),
+                &format!("subscription={subscription}")
+            );
+            Err(TonicStatus::failed_precondition(
+                "header verification failed",
```
dbolduc: nit: if we get here, the verification succeeded

```suggestion
                "ignored",
```

src/pubsub/src/subscriber/session.rs
```diff
@@ -1041,4 +1041,31 @@ mod tests {
...
+
+        let subscription = "projects/p/subscriptions/s";
+
+        mock.expect_streaming_pull().return_once(move |request| {
+            let metadata = request.metadata();
+            assert_eq!(
+                metadata
+                    .get("x-goog-request-params")
+                    .expect("routing header missing"),
+                &format!("subscription={subscription}")
```
dbolduc: nit: I prefer to spell these out. Less logic in the test.

```suggestion
                "subscription=projects/p/subscriptions/s"
```

src/pubsub/src/subscriber/stream.rs
```diff
@@ -217,7 +219,7 @@ mod tests {
         let mut mock = MockStub::new();
         mock.expect_streaming_pull()
             .times(1)
```
dbolduc, line 222: can we add a `.withf(|s, _, _| s == "subscription=blah/blah")` ?

(The unit test in `session.rs` is already giving us coverage, but it's still a good idea to cover the interfaces individually)

src/pubsub/src/subscriber/transport.rs
```diff
@@ -137,6 +138,7 @@ pub(super) mod tests {
         let transport = test_transport(endpoint).await?;
         let mut stream = Stub::streaming_pull(
             &transport,
+            "projects/p/subscriptions/s",
```
dbolduc: nit: subscription=... is more accurate

dbolduc: And we should probably copy the `x-goog-request-params` verification in this test so that this interface has its own coverage

src/pubsub/src/subscriber/session.rs
```diff
@@ -1041,4 +1041,29 @@ mod tests {
...
+                    .expect("routing header missing"),
+                "subscription=projects/p/subscriptions/s"
+            );
+            Err(TonicStatus::failed_precondition("ignored"))
+        });
+
+        let (endpoint, _server) = start("0.0.0.0:0", mock).await?;
+        let client = test_client(endpoint).await?;
+
+        let _ = client.streaming_pull(subscription).start().next().await;
```
dbolduc: nit: just spell out the subscription here

dbolduc (approved):

PR #4588, opened 2026-02-06
test(pubsub): validate sequential actor message sequencing

PhongChuong:


src/pubsub/src/publisher/actor.rs
```diff
@@ -530,6 +531,17 @@ mod tests {
         ))
     }
 
+    fn track_publish_msg_seq(
+        req: &crate::model::PublishRequest,
+        msg_seq_tx: UnboundedSender<PubsubMessage>,
+    ) {
+        req.messages.iter().for_each(|m| {
```
dbolduc, line 538: comment: good idea to send messages individually

src/pubsub/src/publisher/actor.rs
```diff
@@ -781,25 +822,10 @@ mod tests {
...
-            let msg = generate_random_data();
-            let bundle = BundledMessage {
-                msg: PubsubMessage::new().set_data(msg.clone()),
-                tx: publish_tx,
-            };
-            actor_tx.send(ToBatchActor::Publish(bundle))?;
-            publish_rxs.insert(msg, publish_rx);
-        }
+        let mut publish_rxs = VecDeque::new();
+        publish_random_data!(publish_rxs, actor_tx, 10);
```
dbolduc, line 830: comment: hmm my instinct is not to use macros here. But I will say the resulting test code reads nicely. It is self-documenting.

src/pubsub/src/publisher/actor.rs
```diff
@@ -561,15 +572,50 @@ mod tests {
...
+                assert!(matches!(res, Ok(Ok(ref res_msg)) if *res_msg == msg), "got {res:?}, expected {msg:?}");
+                // Assert that publish message matches the expected message sequence.
+                let expected_msg = $expected_msg_seq_rx.try_recv()?.data;
+                assert_eq!(msg, expected_msg, "message sequence mismatch, got {msg:?}, expected {expected_msg:?}");
+            }
+        };
+    }
+
+    // Send ToBatchActor::Publish with random data n times then await and assert the result.
+    macro_rules! assert_publish_is_ok {
```
dbolduc, line 610: Just curious, did you consider a fn with a `#[track_caller]` annotation?

dbolduc: Answering my own question: `#[track_caller]` does not work for `async fn`s. TIL.

https://github.com/rust-lang/rust/issues/104588

src/pubsub/src/publisher/actor.rs
```diff
@@ -561,15 +572,50 @@ mod tests {
...
+        };
+    }
+
+    // Verify that the publish data in publish_rxs matches the resolved message.
+    macro_rules! assert_publish_data {
+        ($publish_rxs:ident) => {
+            for (msg, publish_rx) in $publish_rxs {
+                let res = publish_rx.await;
+                // Assert that input publish message matches resolved publish message.
+                assert!(matches!(res, Ok(Ok(ref res_msg)) if *res_msg == msg), "got {res:?}, expected {msg:?}");
```
dbolduc: Q: can this just be:

```suggestion
                assert_eq!(publish_rx.await??, msg, "unexpected message for given handler");
```

src/pubsub/src/publisher/actor.rs
```diff
@@ -561,15 +572,50 @@ mod tests {
...
+    // Verify that the publish data in publish_rxs the resolved message and the expected sequence.
+    macro_rules! assert_publish_data_with_seq {
+        ($publish_rxs:ident, $expected_msg_seq_rx:ident) => {
+            for (msg, publish_rx) in $publish_rxs {
+                let res = publish_rx.await;
+                // Assert that input publish message matches resolved publish message.
+                assert!(matches!(res, Ok(Ok(ref res_msg)) if *res_msg == msg), "got {res:?}, expected {msg:?}");
+                // Assert that publish message matches the expected message sequence.
+                let expected_msg = $expected_msg_seq_rx.try_recv()?.data;
+                assert_eq!(msg, expected_msg, "message sequence mismatch, got {msg:?}, expected {expected_msg:?}");
```
dbolduc: Same, maybe just:

```suggestion
                assert_eq!(publish_rx.await??, msg, "unexpected message for given handler");
                // Assert that publish message matches the expected message sequence.
                let expected_msg = $expected_msg_seq_rx.try_recv()?.data;
                assert_eq!(msg, expected_msg, "message published out of order");
```

(`assert_eq!` already prints the left and right, so we don't need to repeat them in the error string)

PhongChuong:
@dbolduc , thanks for the review and code suggestions.
I've also added seq. testing for the resume case. PTAL.

src/pubsub/src/publisher/actor.rs
```diff
@@ -876,7 +917,7 @@ mod tests {
 
         // Resume then validate that the actor is no longer paused.
         actor_tx.send(ToBatchActor::ResumePublish())?;
-        assert_publish_is_ok!(actor_tx, 5);
+        assert_publish_is_ok!(actor_tx, msg_seq_rx, EXPECTED_BATCHES);
```
dbolduc, line 920: overloading in Rust? tsk tsk. 😆

dbolduc (approved):

PR #4590, opened 2026-02-06
feat(pubsub): implement `into_stream` for a `Session`

### LINKED ISSUE #4574: Session should be usable as a `futures::Stream`
Most of our types like the `Paginator`, `Poller`, `storage::ReadObjectResponse` are compatible with `futures::Stream`.

We offer an `unstable-stream` feature on these crates, and a conversion function to change the time into something that `impl futures::Stream`.

(This stuff is feature gated because the `futures` crate is not generally available (GA), as its version is `0.x.y`. We cannot have unstable types in our stable public API surface.)

We should do the same thing for a `pubsub::subscriber::Session`.

---

As  followup, we can simplify the examples using streaming primitives, like timing out a stream with `StreamExt::take_until`:

```rs
use futures::StreamExt;

let deadline = tokio::time::sleep_until(Duration::from_secs(10));
let stream = session.into_stream().take_until(deadline);
while let Some(m, h) = stream.next().transpose()? { ... }
```

https://docs.rs/futures/latest/futures/stream/trait.StreamExt.html#method.take_until

haphungw:
Allow `Session` to be used as a `futures::Stream`. This enables the use of streaming primitives for message processing.
This feature is gated behind `unstable-stream` to maintain API stability.

Fixes #4574

### DIFF
--- a/src/pubsub/src/lib.rs
+++ b/src/pubsub/src/lib.rs
@@ -56,7 +56,7 @@
 //! [gcloud-pubsub]: https://crates.io/crates/gcloud-pubsub
 //! [ring]: https://crates.io/crates/ring
 
-#[cfg_attr(docsrs, feature(doc_cfg))]
+#![cfg_attr(docsrs, feature(doc_cfg))]
 #[allow(rustdoc::broken_intra_doc_links)]
 pub(crate) mod generated;
 


src/pubsub/Cargo.toml
```diff
@@ -55,6 +55,7 @@ default = ["default-rustls-provider"]
 # cryptography (such as exclusively using the [ring] crate) should disable this
 # default and call `rustls::CryptoProvider::install_default()`.
 default-rustls-provider = ["gaxi/_default-rustls-provider"]
+unstable-stream         = []
```
dbolduc, line 61: We will want to exercise this in the CI, e.g. how storage does:

https://github.com/googleapis/google-cloud-rust/blob/9038e2ceb8e48971a9996ff86921f66fef6a5e9e/.gcb/scripts/features.sh#L55-L64

dbolduc: cc: @coryan who is very sensitive to build times

dbolduc: Also, can you add a stanza like in #4594, so the new API is included in the docs

src/pubsub/src/subscriber/session.rs
```diff
@@ -1068,4 +1085,33 @@ mod tests {
...
+
+        let mut mock = MockSubscriber::new();
+        mock.expect_streaming_pull()
+            .return_once(|_| Ok(TonicResponse::from(response_rx)));
+
+        let (endpoint, _server) = start("0.0.0.0:0", mock).await?;
+        let client = test_client(endpoint).await?;
+
+        let session = client.streaming_pull("projects/p/subscriptions/s").start();
+        let mut stream = session.into_stream();
```
dbolduc: style nit: I'd one line this.

(which, ironically, the formatter might split into three lines)

src/pubsub/src/subscriber/session.rs
```diff
@@ -1068,4 +1085,33 @@ mod tests {
...
+
+        response_tx.send(Ok(test_response(1..3))).await?;
+        drop(response_tx);
+
+        let mut count = 0;
+        while let Some((m, _h)) = stream.next().await.transpose()? {
+            assert_eq!(m.data, test_data(count + 1));
+            count += 1;
+        }
+        assert_eq!(count, 2);
```
dbolduc: Hm, what about using a fancy new helper from `futures::StreamExt` or `futures::TryStreamExt`:

https://docs.rs/futures/latest/futures/stream/trait.TryStreamExt.html#method.try_collect

```rs
        let got: Vec<_> = stream.map_ok(|(m, h)| {
            h.ack();
            m.data
        }).try_collect().await?;
        assert_eq!(got, vec![test_data(1), test_data(2)]);
```

src/pubsub/src/subscriber/session.rs
```diff
@@ -1068,4 +1085,33 @@ mod tests {
 
         Ok(())
     }
+
+    #[cfg(feature = "unstable-stream")]
+    #[tokio::test]
```
dbolduc: We should `#[tokio::test(start_paused = true)]` because this class has timers in it. It will make the behavior more reliable and less flaky.

(We also don't want the nacks going out to race with the program exiting. Note that it would kill our mock server which is not expecting them).

src/pubsub/src/subscriber/session.rs
```diff
@@ -1068,4 +1085,33 @@ mod tests {
...
+    async fn into_stream() -> anyhow::Result<()> {
+        use futures::StreamExt;
+        let (response_tx, response_rx) = channel(10);
+
+        let mut mock = MockSubscriber::new();
+        mock.expect_streaming_pull()
+            .return_once(|_| Ok(TonicResponse::from(response_rx)));
```
dbolduc, line 1098: optional: I feel like ideally we would ack the messages, and make sure the acknowledgements go out.

The annoying thing is that we lose `session.close().await` (which lets our test wait for everything to send before exiting. drat.)

So maybe we add like:

https://github.com/haphungw/google-cloud-rust/blob/5af3de93926ab4a59d06300a6226262dc6b8784e/src/pubsub/src/subscriber/session.rs#L411-L416

plus

https://github.com/haphungw/google-cloud-rust/blob/5af3de93926ab4a59d06300a6226262dc6b8784e/src/pubsub/src/subscriber/session.rs#L439-L442

except with `.recv().await` instead of `.try_recv()` ?

src/pubsub/src/subscriber/session.rs
```diff
@@ -182,6 +182,23 @@ impl Session {
         }
     }
 
+    #[cfg(feature = "unstable-stream")]
+    #[cfg_attr(docsrs, doc(cfg(feature = "unstable-stream")))]
```
dbolduc, line 186: Oh yeah, we will need the changes from #4594 in pubsub too.

src/pubsub/src/subscriber/session.rs
dbolduc: Feel free to merge the PR without doing this bit. It's not necessary.

dbolduc (commented):
The code all looks good.

- we definitely need to exercise the new feature in our CI
- we should document the new feature
- the tests could be cleaned up a bit

src/pubsub/src/lib.rs
```diff
@@ -56,6 +56,7 @@
 //! [gcloud-pubsub]: https://crates.io/crates/gcloud-pubsub
 //! [ring]: https://crates.io/crates/ring
 
+#[cfg_attr(docsrs, feature(doc_cfg))]
```
dbolduc: This works?? I think it should be:

```suggestion
#![cfg_attr(docsrs, feature(doc_cfg))]

```

basically `#[...]` applies to the thing immediately below it, but `#![...]` applies to the module in scope, which in `lib.rs`, is the whole crate.

dbolduc: Ahh lol, we only try to build 20 crates on PRs (probably for timing reasons). I guess pubsub is not one of them.

https://github.com/googleapis/google-cloud-rust/blob/6e053dd1e3905ffec78adec8f3432d5526fc0f86/.gcb/scripts/docs-rs.sh#L25-L28

I think the post-merge build would fail.

haphungw: makes sense

src/pubsub/src/lib.rs
```diff
@@ -56,6 +56,7 @@
 //! [gcloud-pubsub]: https://crates.io/crates/gcloud-pubsub
 //! [ring]: https://crates.io/crates/ring
 
+#![cfg_attr(docsrs, feature(doc_cfg))]
 #[allow(rustdoc::broken_intra_doc_links)]
```
dbolduc, line 60: super nit: I'd put a newline between these so the `#![...]` stands out.

dbolduc: I know you have another PR pending. Feel free to merge this as-is, and make the fix in that PR

Issue #4186, opened 2026-01-08
auth: Trust Boundaries support


alvarowolfx:
The purpose of trust boundaries is to restrict the ability for these credentials to be used globally if they are only intended to be used in a subset of GCP regions. 

Using these trust boundaries ensures that users or workloads attempting to access resources in a region are authorized to do so.

Internal doc: go/trust-boundaries-auth-sdk-v2

PR #4591, opened 2026-02-06
impl(auth): add initial access boundary module

alvarowolfx:
Towards #4186

### DIFF
--- a/src/auth/src/access_boundary.rs
+++ b/src/auth/src/access_boundary.rs
@@ -16,9 +16,6 @@ use crate::credentials::CacheableResource;
 use crate::errors::CredentialsError;
 use crate::headers_util::AuthHeadersBuilder;
 use crate::token::CachedTokenProvider;
-use google_cloud_gax::backoff_policy::BackoffPolicy;
-use google_cloud_gax::exponential_backoff::ExponentialBackoffBuilder;
-use google_cloud_gax::retry_state::RetryState;
 use http::Extensions;
 use reqwest::Client;
 use std::clone::Clone;
@@ -36,8 +33,6 @@ const DEFAULT_TTL: Duration = Duration::from_secs(6 * 60 * 60);
 const REFRESH_INTERVAL: Duration = Duration::from_secs(60 * 60);
 // Period to wait after an error: 15 minutes
 const COOLDOWN_INTERVAL: Duration = Duration::from_secs(15 * 60);
-// Max period to wait after an error: 2 hours
-const MAX_COOLDOWN_INTERVAL: Duration = Duration::from_secs(2 * 60 * 60);
 
 #[derive(Debug)]
 pub(crate) struct AccessBoundary {
@@ -49,21 +44,20 @@ impl AccessBoundary {
     where
         T: CachedTokenProvider + 'static,
     {
-        let enabled = Self::is_regional_access_boundaries_enabled();
         let (tx_header, rx_header) = watch::channel(None);
-        let provider = IAMAccessBoundaryProvider {
-            token_provider,
-            url,
-        };
 
-        if enabled {
+        if Self::is_enabled() {
+            let provider = IAMAccessBoundaryProvider {
+                token_provider,
+                url,
+            };
             tokio::spawn(refresh_task(Arc::new(provider), tx_header));
         }
 
         Self { rx_header }
     }
 
-    fn is_regional_access_boundaries_enabled() -> bool {
+    fn is_enabled() -> bool {
         std::env::var(REGIONAL_ACCESS_BOUNDARIES_ENV_VAR)
             .map(|v| v.to_lowercase())
             .map(|v| v == "true" || v == "1")
@@ -73,10 +67,8 @@ impl AccessBoundary {
     pub(crate) fn header_value(&self) -> Option<String> {
         // TODO(#4186): handle expiration. Each new entry should have TTL per design doc.
         let val = self.rx_header.borrow().clone();
-        if let Some(ref v) = val {
-            if v == NO_OP_ENCODED_LOCATIONS {
-                return None;
-            }
+        if val.as_ref().is_some_and(|v| v == NO_OP_ENCODED_LOCATIONS) {
+            return None;
         }
         val
     }
@@ -167,26 +159,14 @@ async fn refresh_task<T>(provider: Arc<T>, tx_header: watch::Sender<Option<Strin
 where
     T: AccessBoundaryProvider,
 {
-    let backoff = ExponentialBackoffBuilder::new()
-        .with_initial_delay(COOLDOWN_INTERVAL)
-        .with_maximum_delay(MAX_COOLDOWN_INTERVAL)
-        .with_scaling(2.0)
-        .build()
-        .expect("static cooldown settings should be valid");
-
-    let mut state = RetryState::new(true);
-
     loop {
         match provider.fetch_access_boundary().await {
             Ok(val) => {
                 let _ = tx_header.send(val);
-                state = RetryState::new(true); // reset backoff on success
                 sleep(REFRESH_INTERVAL).await;
             }
             Err(_e) => {
-                let delay = backoff.on_failure(&state);
-                sleep(delay).await;
-                state = state.clone().set_attempt_count(state.attempt_count + 1);
+                sleep(COOLDOWN_INTERVAL).await;
             }
         }
     }
@@ -194,8 +174,7 @@ where
 
 pub(crate) fn service_account_lookup_url(email: &str) -> String {
     format!(
-        "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{}/allowedLocations",
-        email
+        "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{email}/allowedLocations"
     )
 }
 
@@ -210,6 +189,7 @@ mod tests {
     use serde_json::json;
     use serial_test::{parallel, serial};
     use std::sync::Arc;
+    use test_case::test_case;
 
     type TestResult = anyhow::Result<()>;
 
@@ -261,7 +241,7 @@ mod tests {
         let url = server.url("/allowedLocations").to_string();
 
         let result = fetch_access_boundary(&token_provider, &url).await?;
-        assert_eq!(result.as_deref(), Some("0x123"));
+        assert_eq!(result.as_deref(), Some("0x123"), "{result:?}");
 
         Ok(())
     }
@@ -292,8 +272,8 @@ mod tests {
         let token_provider = TokenCache::new(mock);
         let url = server.url("/allowedLocations").to_string();
 
-        let result = fetch_access_boundary(&token_provider, &url).await?;
-        assert_eq!(result, None);
+        let val = fetch_access_boundary(&token_provider, &url).await?;
+        assert!(val.is_none(), "{val:?}");
 
         Ok(())
     }
@@ -322,7 +302,7 @@ mod tests {
 
         let result = fetch_access_boundary(&token_provider, &url).await;
         let err = result.unwrap_err();
-        assert!(err.is_transient());
+        assert!(err.is_transient(), "{err:?}");
     }
 
     #[tokio::test]
@@ -339,7 +319,7 @@ mod tests {
         let token_provider = TokenCache::new(mock);
         let result = fetch_access_boundary(&token_provider, "http://localhost").await;
         let err = result.unwrap_err();
-        assert!(!err.is_transient());
+        assert!(!err.is_transient(), "{err:?}");
     }
 
     #[tokio::test]
@@ -360,7 +340,8 @@ mod tests {
         let token_provider = TokenCache::new(mock);
         let access_boundary = AccessBoundary::new(token_provider, "http://localhost".to_string());
 
-        assert_eq!(access_boundary.header_value(), None);
+        let val = access_boundary.header_value();
+        assert!(val.is_none(), "{val:?}");
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+use std::fmt::Debug;
+use tokio::sync::watch;
+use tokio::time::{Duration, sleep};
+
+const REGIONAL_ACCESS_BOUNDARIES_ENV_VAR: &str = "GOOGLE_AUTH_ENABLE_TRUST_BOUNDARIES";
+const NO_OP_ENCODED_LOCATIONS: &str = "0x0";
+
+// TTL: 6 hours
+const DEFAULT_TTL: Duration = Duration::from_secs(6 * 60 * 60);
+// Refresh interval: 1 hour of TTL
```
coryan: "of TTL" ?

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+const REFRESH_INTERVAL: Duration = Duration::from_secs(60 * 60);
+// Period to wait after err: 15 minutes
```
coryan:
```suggestion
// Period to wait after an error: 15 minutes
```

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+const COOLDOWN_INTERVAL: Duration = Duration::from_secs(15 * 60);
+// Max period to wait after err: 6 hours
```
coryan:
```suggestion
// Max period to wait after an error: 6 hours
```

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+
+    fn is_regional_access_boundaries_enabled() -> bool {
+        std::env::var(REGIONAL_ACCESS_BOUNDARIES_ENV_VAR)
+            .map(|v| v.to_lowercase())
+            .map(|v| v == "true" || v == "1")
+            .unwrap_or(false)
+    }
+
+    pub(crate) fn header_value(&self) -> Option<String> {
+        // TODO(#4186): handle expiration. Each new entry should have TTL per design doc.
```
coryan, line 68: Consider an expiration time (computed when the thing is fetched), so we can use the same ideas from tokens.

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+const NO_OP_ENCODED_LOCATIONS: &str = "0x0";
+
+// TTL: 6 hours
+const DEFAULT_TTL: Duration = Duration::from_secs(6 * 60 * 60);
+// Refresh interval: 1 hour of TTL
+const REFRESH_INTERVAL: Duration = Duration::from_secs(60 * 60);
+// Period to wait after err: 15 minutes
+const COOLDOWN_INTERVAL: Duration = Duration::from_secs(15 * 60);
+// Max period to wait after err: 6 hours
+const MAX_COOLDOWN_INTERVAL: Duration = Duration::from_secs(6 * 60 * 60);
```
coryan: That seems entirely too long. The backoff policy waiting for 6h means you may retry with backoffs of 15m, 30m, 1h, 2h, 4h, and 6h.... Somewhere in that sequence you are waiting after the thing as long expired.

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+            )
+            .await;
+        });
+
+        // allow task to start and fail the first request
+        tokio::time::advance(Duration::from_secs(2)).await;
+        tokio::task::yield_now().await;
+
+        let val = rx.borrow().clone();
+        assert!(val.is_none(), "should be None on startup/error");
```
coryan: You just got an error in the coverage build without enough details:

```suggestion
        assert!(val.is_none(), "should be None on startup/error: {val:?}");
```

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+            })
+        });
+
+        let token_provider = TokenCache::new(mock);
+        let url = server.url("/allowedLocations").to_string();
+
+        let result = fetch_access_boundary(&token_provider, &url).await;
+
+        assert!(result.is_ok());
+        assert_eq!(result.unwrap(), Some("0x123".to_string()));
```
coryan: You can simplify this to:


```suggestion
        let respoinse = fetch_access_boundary(&token_provider, &url).await?;
        assert_eq!(response.as_deref(), Some("0x123"));
```

Which has the advantage of printing the error if it fails too. Please make sure all your assertions have meaningful errors.

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+        let url = server.url("/allowedLocations").to_string();
+
+        let result = fetch_access_boundary(&token_provider, &url).await;
+
+        assert!(result.is_err());
+        assert!(
+            result
+                .unwrap_err()
+                .to_string()
+                .contains("Failed to fetch access boundary")
```
coryan: We should check errors by type, not string values.

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+                true,
+                "invalid creds".to_string(),
+            ))
+        });
+
+        let token_provider = TokenCache::new(mock);
+        let result = fetch_access_boundary(&token_provider, "http://localhost").await;
+
+        assert!(result.is_err());
+        assert!(result.unwrap_err().to_string().contains("invalid creds"));
```
coryan: Ditto.

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+    #[serial]
+    fn test_is_regional_access_boundaries_enabled() {
+        let cases = [
+            (Some("true"), true),
+            (Some("TRUE"), true),
+            (Some("1"), true),
+            (Some("false"), false),
+            (Some("0"), false),
+            (None, false),
+        ];
```
coryan: Can we use `#[test_case()]` for these?

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,443 @@
...
+        assert!(val.is_none(), "should still be None after second error");
+
+        // advance 30 minutes, third call succeeds
+        tokio::time::advance(2 * COOLDOWN_INTERVAL).await;
+        for _ in 0..1000 {
+            tokio::task::yield_now().await;
+        }
+
+        let val = rx.borrow().clone();
+        assert_eq!(val, Some("0x123".to_string()));
```
coryan: nit:

```suggestion
        assert_eq!(val.as_deref(), Some("0x123"));
```

src/auth/src/access_boundary.rs
alvarowolfx: classic copy/pasta issue, I was reusing the same const as default TTL and refresh interval and left that here 😮‍💨

src/auth/src/access_boundary.rs
alvarowolfx: yeah, I will do that, but I wanted to make the initial less convoluted.

src/auth/src/access_boundary.rs
alvarowolfx: `test_case` and `serial_test` macros don't play well, that's why I ended using just an array here.

alvarowolfx (commented):
ack on the comments, will address the suggestion here in a minute

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+const REGIONAL_ACCESS_BOUNDARIES_ENV_VAR: &str = "GOOGLE_AUTH_ENABLE_TRUST_BOUNDARIES";
+const NO_OP_ENCODED_LOCATIONS: &str = "0x0";
+
+// TTL: 6 hours
+const DEFAULT_TTL: Duration = Duration::from_secs(6 * 60 * 60);
+// Refresh interval: every hour
+const REFRESH_INTERVAL: Duration = Duration::from_secs(60 * 60);
+// Period to wait after an error: 15 minutes
+const COOLDOWN_INTERVAL: Duration = Duration::from_secs(15 * 60);
+// Max period to wait after an error: 2 hours
```
coryan: Even that seems too long.  This may be a case where min==max at 15m is good enough 🤷

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+    token_provider: T,
+    url: String,
+}
+
+#[async_trait::async_trait]
+impl<T> AccessBoundaryProvider for IAMAccessBoundaryProvider<T>
+where
+    T: CachedTokenProvider + 'static,
+{
+    async fn fetch_access_boundary(&self) -> Result<Option<String>, CredentialsError> {
```
coryan, line 100: codecov says this is not touched, is that right?

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+        let enabled = Self::is_regional_access_boundaries_enabled();
+        let (tx_header, rx_header) = watch::channel(None);
+        let provider = IAMAccessBoundaryProvider {
+            token_provider,
+            url,
+        };
+
+        if enabled {
+            tokio::spawn(refresh_task(Arc::new(provider), tx_header));
+        }
```
coryan: Consider:

```suggestion
        let (tx_header, rx_header) = watch::channel(None);
        let provider = IAMAccessBoundaryProvider {
            token_provider,
            url,
        };

        if Self::is_regional_access_boundaries_enabled() {
            tokio::spawn(refresh_task(Arc::new(provider), tx_header));
        }
```

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+    }
+
+    pub(crate) fn header_value(&self) -> Option<String> {
+        // TODO(#4186): handle expiration. Each new entry should have TTL per design doc.
+        let val = self.rx_header.borrow().clone();
+        if let Some(ref v) = val {
+            if v == NO_OP_ENCODED_LOCATIONS {
+                return None;
+            }
+        }
```
coryan:
```suggestion
        if val.is_some_and(|v| v == NO_OP_ENCODED_LOCATIONS) {
                return None;
            }
```

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+    }
+}
+
+pub(crate) fn service_account_lookup_url(email: &str) -> String {
+    format!(
+        "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{}/allowedLocations",
+        email
+    )
```
coryan, line 178:
```suggestion
    format!(
        "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{email}/allowedLocations",
    )
```

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+                metadata: None,
+            })
+        });
+
+        let token_provider = TokenCache::new(mock);
+        let url = server.url("/allowedLocations").to_string();
+
+        let result = fetch_access_boundary(&token_provider, &url).await;
+        let err = result.unwrap_err();
+        assert!(err.is_transient());
```
coryan:
```suggestion
        assert!(err.is_transient(), "{err:?}");
```

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+            Err(CredentialsError::from_msg(
+                false,
+                "invalid creds".to_string(),
+            ))
+        });
+
+        let token_provider = TokenCache::new(mock);
+        let result = fetch_access_boundary(&token_provider, "http://localhost").await;
+        let err = result.unwrap_err();
+        assert!(!err.is_transient());
```
coryan: Ditto.

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+                token_type: "Bearer".into(),
+                expires_at: None,
+                metadata: None,
+            })
+        });
+
+        let token_provider = TokenCache::new(mock);
+        let access_boundary = AccessBoundary::new(token_provider, "http://localhost".to_string());
+
+        assert_eq!(access_boundary.header_value(), None);
```
coryan: I think clippy will tell you to use:


```suggestion
        assert!(access_boundary.header_value().is_none(), "{header_boundary:?}");
```

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+    #[parallel]
+    fn test_access_boundary_header_value_no_op() {
+        let (tx, rx_header) = watch::channel(None);
+        let access_boundary = AccessBoundary { rx_header };
+
+        let _ = tx.send(Some("0x123".to_string()));
+        assert_eq!(access_boundary.header_value().as_deref(), Some("0x123"));
+
+        let _ = tx.send(Some(NO_OP_ENCODED_LOCATIONS.to_string()));
+        assert_eq!(access_boundary.header_value(), None);
```
coryan: Ditto.

src/auth/src/access_boundary.rs
coryan: Now I remember, please add a comment.

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+        };
+
+        if enabled {
+            tokio::spawn(refresh_task(Arc::new(provider), tx_header));
+        }
+
+        Self { rx_header }
+    }
+
+    fn is_regional_access_boundaries_enabled() -> bool {
```
coryan: I don't think this is an awesome name btw.  `is_enabled()` would be fine.  Or `regional_access_boundaries_are_enabled()`

src/auth/src/access_boundary.rs
```diff
@@ -0,0 +1,455 @@
...
+                expected,
+                "Failed on case: env_val={:?}, expected={}",
+                env_val,
+                expected
+            );
+        }
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn test_refresh_task_backoff() {
```
coryan, line 394: Does this need a `#[parallel]` to avoid flakes? Is that possible for async functions?

src/auth/src/access_boundary.rs
dbolduc: Doubt:

https://github.com/googleapis/google-cloud-rust/blob/e624867463b1a49cc411532f0387dfa0756e72db/src/auth/src/credentials/external_account_sources/executable_sourced.rs#L405-L437

I remember something like `#[serial]` needs to come last. Did you try that?

src/auth/src/access_boundary.rs
alvarowolfx: It's being called at https://github.com/googleapis/google-cloud-rust/pull/4591/changes#diff-cb16fd6bb22d77cff0e35c4feeff34806a015cb1e5d49dfaf3be4900ad1bc4ffR180 🤷

src/auth/src/access_boundary.rs
alvarowolfx: ugh, I tried before but I haven't used the right incantation. Thanks for the tip @dbolduc

PR #4594, opened 2026-02-06
docs(storage): include into_stream adapter

dbolduc:
We are missing the `into_stream()` feature in the docs:

https://docs.rs/google-cloud-storage/1.7.0/google_cloud_storage/read_object/struct.ReadObjectResponse.html

We want something like `gax` has:

https://docs.rs/google-cloud-gax/latest/google_cloud_gax/paginator/trait.Paginator.html#tymethod.into_stream


dbolduc:
Sorry I missed a bit of set up to turn on the docs rs feature. PTAL

coryan (approved):
I wonder if we can set this option globally, somehow <waves hands>

PR #4600, opened 2026-02-08
refactor: use librarian.yaml as single version source

julieqiu:
Delete .librarian-version.txt and use the version field in librarian.yaml as the single source of truth for the expected librarian binary version.


coryan:
The pending jobs:

image

Are finished:

image

This is a bug in Cloud Build, where the notification to GitHub is lost.  I will restart the builds.

PR #4601, opened 2026-02-09
feat(storage): add with_user_agent to the builder.

joshuatants:
As requested in #3615

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -832,14 +832,14 @@ mod tests {
 
         // Test with client agent only
         let builder = client.builder(Method::GET, "foo".to_string());
-        let options = gax::options::RequestOptions::default();
+        let options = RequestOptions::default();
         let _ = client
             .execute_streaming_once(builder, options, None, 0)
             .await?;
 
         // Test that request agent overrides client agent
         let builder = client.builder(Method::GET, "foo".to_string());
-        let mut options = gax::options::RequestOptions::default();
+        let mut options = RequestOptions::default();
         options.set_user_agent(request_agent);
         let _ = client
             .execute_streaming_once(builder, options, None, 0)
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -538,7 +538,6 @@ mod tests {
 #[cfg(test)]
 mod headers_tests {
     use super::*;
-    use gax::options::RequestOptions;
 
     #[tokio::test]
     async fn test_user_agent_headers() {


src/gax/src/client_builder.rs
```diff
@@ -199,6 +199,25 @@ impl<F, Cr> ClientBuilder<F, Cr> {
...
+    /// # use google_cloud_gax::client_builder::examples;
+    /// # use google_cloud_gax::client_builder::Result;
+    /// # tokio_test::block_on(async {
+    /// use examples::Client; // Placeholder for examples
+    /// let client = Client::builder()
+    ///     .with_user_agent("my-app/1.0.0")
+    ///     .build().await?;
+    /// # Result::<()>::Ok(()) });
+    /// ```
+    pub fn with_user_agent<V: Into<String>>(mut self, v: V) -> Self {
```
coryan, line 216: This would require a minor version bump to `google-cloud-gax`.  Fortunately the bump already was made in `src/gax/Cargo.toml` and `librarian.yaml`, but has not been made in `/Cargo.toml`.

src/gax-internal/src/http.rs
```diff
@@ -79,6 +79,9 @@ impl ReqwestClient {
         if config.disable_follow_redirects {
             builder = builder.redirect(::reqwest::redirect::Policy::none());
         }
+        if let Some(ref user_agent) = config.user_agent {
+            builder = builder.user_agent(user_agent);
+        }
```
coryan: You would also need to make a similar change to `grpc.rs`. That matters for most gRPC requests.

And you need to decide how this feature (a client option to set the user agent) interacts with the existing feature to add user agents per request:

https://github.com/googleapis/google-cloud-rust/blob/6230db392c5e9668bbc143df817092c9be65b9d0/src/gax-internal/src/http.rs#L193-L199

and

https://github.com/googleapis/google-cloud-rust/blob/6230db392c5e9668bbc143df817092c9be65b9d0/src/gax-internal/src/grpc.rs#L415-L420

Should the per-request user agent override the client? Or append to it?

src/gax-internal/src/http.rs
```diff
@@ -786,4 +789,32 @@ mod tests {
         assert_eq!(result.err().unwrap().http_status_code(), Some(308));
         Ok(())
     }
+    #[tokio::test]
```
coryan, line 798: nit:


```suggestion

    #[tokio::test]
```

joshuatants: Done

src/gax/src/client_builder.rs
joshuatants: Bumped to 1.7.0 in /Cargo.toml

src/gax-internal/src/http.rs
joshuatants: Updated `grpc.rs`.

Updated the logic in both `grpc.rs` and `http.rs` so that the request user agent overrides the client user agent.

joshuatants:
Is there a way to run the GCB checks locally before creating or updating a PR? I looked at https://github.com/googleapis/google-cloud-rust/tree/main/.gcb/ , but the scripts look like they are designed for a disposable CI environment - they modify Cargo.toml and also attempt to globally install/update cargo tools.

coryan:
> As requested in #3615

Hm... Maybe not? That was asking for parity with the other clients. This is adding functionality to the storage and other clients.  Maybe good functionality, but more than the other clients.  Do I miss something?

src/storage/src/storage/client.rs
```diff
@@ -421,6 +421,25 @@ impl ClientBuilder {
         self
     }
 
+    /// Sets the user-agent.
```
coryan, line 424: To achieve parity with the other client libraries you also need to add `with_user_agent` to the request builders, including:

https://github.com/googleapis/google-cloud-rust/blob/c3d99d7238541c05b37ff06ad7f021d4d1dbe405/src/storage/src/storage/read_object.rs#L90-L93

https://github.com/googleapis/google-cloud-rust/blob/c3d99d7238541c05b37ff06ad7f021d4d1dbe405/src/storage/src/storage/write_object.rs#L85-L87

and

https://github.com/googleapis/google-cloud-rust/blob/c3d99d7238541c05b37ff06ad7f021d4d1dbe405/src/storage/src/storage/open_object.rs#L49-L51

Cargo.toml
```diff
@@ -451,8 +451,8 @@ tokio-test        = { default-features = false, version = "0.4" }
 
 # Local packages used as dependencies.
 google-cloud-auth        = { default-features = false, version = "1.5", path = "src/auth" }
-gax                      = { default-features = false, version = "1.6", path = "src/gax", package = "google-cloud-gax" }
-google-cloud-gax         = { default-features = false, version = "1.6", path = "src/gax" }
+gax                      = { default-features = false, version = "1.7", path = "src/gax", package = "google-cloud-gax" }
+google-cloud-gax         = { default-features = false, version = "1.7", path = "src/gax" }
```
coryan, line 455:
```suggestion
google-cloud-gax         = { default-features = false, version = "1.7", path = "src/gax" }
```

joshuatants:
> > As requested in #3615
> 
> Hm... Maybe not? That was asking for parity with the other clients. This is adding functionality to the storage and other clients. Maybe good functionality, but more than the other clients. Do I miss something?

This is in the context of https://screenshot.googleplex.com/5PWF6ftwUYATr48 . I couldn't quote the doc, so I quoted the FR mentioned there. Should I create a new issue just for this and link it to #3615?

joshuatants:
> I think you want to break down this PR. The changes to `gax-internal` and `gax` should be on their own PR. They are effectively adding functionality to all the client libraries, not just storage.
> 
> You also need to change the request builders for storage. I would also do those in a separate PR.
> 
> Thoughts?

Will split the PR. I'll ask my teammate sharonlynn@ to make the changes to `gax-internal` and `gax` so that she can get familiar with working on the Rust repo. 

Did you manage to see my question on whether there's a way to run the GCB checks non-destructively locally? Right now I'm creating a draft PR just to run the checks, since they check things that aren't available via `cargo test`.

coryan:
Closing. You have other PRs in progress for the same thing that look in better shape. You can reopen if needed.

PR #4617, opened 2026-02-09
docs(pubsub): match the Publisher to other client comment style

### LINKED ISSUE #4616: Publisher doc comment does not match style of other clients in crate::client::*

suzmue:
Fixes #4616

### DIFF
--- a/src/pubsub/src/publisher/client.rs
+++ b/src/pubsub/src/publisher/client.rs
@@ -167,7 +167,7 @@ mod tests {
         model::{PublishResponse, PubsubMessage},
     };
     use mockall::Sequence;
-    use rand::{Rng, distr::Alphanumeric};
+    use rand::{RngExt, distr::Alphanumeric};
     use std::error::Error;
     use std::time::Duration;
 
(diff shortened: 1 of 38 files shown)


src/pubsub/src/publisher/client.rs
```diff
@@ -28,9 +30,11 @@ use tokio::sync::oneshot;
 /// # use google_cloud_pubsub::client::Publisher;
 /// # use model::PubsubMessage;
 /// let publisher = Publisher::builder("projects/my-project/topics/my-topic").build().await?;
-/// let message_id = publisher.publish(PubsubMessage::new().set_data("Hello, World"));
+/// let message_id_future = publisher.publish(PubsubMessage::new().set_data("Hello, World"));
```
dbolduc, line 33: I see that #4634 will conflict with this. I will wait my turn.

dbolduc (approved):

src/pubsub/src/publisher/client.rs
suzmue: TY!

PR #4620, opened 2026-02-09
fix(pubsub): change error return type for publish to Arc

### LINKED ISSUE #4619: Return Arc<Error> directly for publish results
The current implementation wraps all errors returned from a publish call into an Error::io. This obscures the actual error returned from the service. We want to resolve this by returning Arc<Error> directly.

suzmue:
To allow sharing the same crate::Error for multiple messages bundled in the same batch, we wrap the error in an Arc. This allows us to remove the code that wraps the originally Error in an Error::io.

The tests also needed to be updates since they were actually checking for the io error type.

Fixes #4619

### DIFF
--- a/src/pubsub/src/publisher/client.rs
+++ b/src/pubsub/src/publisher/client.rs
@@ -219,16 +219,14 @@ mod tests {
         ))
     }
 
-    macro_rules! assert_publish_err {
-        ($got_err:expr) => {
-            let got_err = $got_err;
-            assert!(got_err.status().is_some(), "{got_err:?}");
-            assert_eq!(
-                got_err.status().unwrap().code,
-                google_cloud_gax::error::rpc::Code::Unknown,
-                "{got_err:?}"
-            );
-        };
+    #[track_caller]
+    fn assert_publish_err(got_err: std::sync::Arc<google_cloud_gax::error::Error>) {
+        assert!(got_err.status().is_some(), "{got_err:?}");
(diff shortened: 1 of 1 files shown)


src/pubsub/src/publisher/client.rs
```diff
@@ -187,34 +187,43 @@ mod tests {
...
+        _options: gax::options::RequestOptions,
+    ) -> gax::Result<gax::response::Response<crate::model::PublishResponse>> {
+        Err(gax::error::Error::service(
+            gax::error::rpc::Status::default()
+                .set_code(gax::error::rpc::Code::Unknown)
                 .set_message("unknown error has occurred"),
         ))
     }
 
+    fn assert_publish_err(got_err: std::sync::Arc<gax::error::Error>) {
```
PhongChuong: nit. using fn to assert will not give the line number of test failure at the calling site.
I ended up using macro_rules! instead for test assert helpers.

dbolduc: There is also `#[track_caller]` for sync functions

https://github.com/googleapis/google-cloud-rust/blob/a00cd10164706eb8b8f4a8207409ff2bb31a156c/src/pubsub/src/subscriber/retry_policy.rs#L246-L253

suzmue: Oh nice! Thanks those are great workarounds.

I'll use track_caller here to keep things simple.

PR #4622, opened 2026-02-09
fix(pubsub): handle shutdown behavior for publisher

### LINKED ISSUE #4483: Gracefully handle forced shutdown for Publish background tasks
If the background task that contains the Publish actors gets forcefully stopped (not from dropping the Publisher), the oneshot channels that it holds don't get correctly resolved and we get the unexpected error when polling the channel.

We can resolve this in two ways:
1. Handle the closed channel on the PublishHandle side
2. impl Drop for the BatchActors and resolve all PublishHandle's it knows about

1 will likely be easier to implement.

suzmue:
During the normal flow of using a publisher, the background task itself should not be dropped. When the task containing the Publisher is shutting down however, the background task may be forcefully aborted before the task that is processing the results of the publish. This will result in any resources owned by the background task to be dropped, including the channel for processing incoming publish requests and any oneshot channels to be resolved for publish requests. This means that the `PublishFuture` must be able to handle the case where these resources are dropped without panicking.

Fixes #4483

### DIFF
--- a/src/pubsub/src/publisher/model_ext.rs
+++ b/src/pubsub/src/publisher/model_ext.rs
@@ -77,12 +77,10 @@ fn convert_error(e: crate::error::PublishError) -> Arc<crate::Error> {
 
 #[cfg(test)]
 mod tests {
-    use crate::generated::gapic_dataplane::builder::publisher::Publish;
-
     use super::*;
 
     #[tokio::test]
-    async fn resolve_publish_handle_success() {
+    async fn resolve_publish_future_success() {
         let (tx, rx) = oneshot::channel();
         let handle = PublishFuture { rx };
         let _ = tx.send(Ok("message_id".to_string()));
@@ -91,7 +89,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn resolve_publish_handle_error() {
+    async fn resolve_publish_future_error() {
         use std::error::Error as _;
 
         let (tx, rx) = oneshot::channel();
@@ -112,7 +110,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn resolve_publish_handle_error_send_error() {
+    async fn resolve_publish_future_error_send_error() {
         let (tx, rx) = oneshot::channel();
         let fut = PublishFuture { rx };
         drop(tx);


src/pubsub/src/publisher/client.rs
```diff
@@ -308,6 +309,34 @@ mod tests {
         }
     }
 
+    #[tokio::test]
```
dbolduc, line 312: nit: `start_paused = true`

We don't want to race with the batch timer, however unlikely that is to fire first

src/pubsub/src/publisher/client.rs
```diff
@@ -308,6 +309,34 @@ mod tests {
...
+            PubsubMessage::new().set_data("hello"),
+            PubsubMessage::new().set_data("world"),
+        ];
+        let mut handles = Vec::new();
+        for msg in messages {
+            let handle = publisher.publish(msg);
+            handles.push(handle);
+        }
+
+        background_task_handle.abort();
```
dbolduc, line 332: good idea to test this sort of thing

src/pubsub/src/publisher/model_ext.rs
```diff
@@ -54,12 +54,13 @@ impl Future for PublishFuture {
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let result = ready!(Pin::new(&mut self.rx).poll(cx));
         // An error will only occur if the sender of the self.rx was dropped,
-        // which would be a bug.
-        Poll::Ready(
-            result
-                .expect("publisher should not close the sender for PublishFuture")
-                .map_err(convert_error),
-        )
+        // which can happen when the background worker is dropped.
```
PhongChuong, line 57:
```suggestion
        // which can happen when the dispatcher is dropped.
```

suzmue:
Going to address the comments in a follow up PR, because I'd like this fix to land in the release if possible.

src/pubsub/src/publisher/client.rs
suzmue: addressed in https://github.com/googleapis/google-cloud-rust/pull/4647

src/pubsub/src/publisher/model_ext.rs
suzmue: addressed in https://github.com/googleapis/google-cloud-rust/pull/4647

PR #4623, opened 2026-02-09
chore: update librarian, googleapis and discovery-artifact-manager, and regenerate

miguelvelezsa:
Update librarian version to @main (v0.8.1-0.20260209175721-adbf644f2605).


src/storage/src/generated/gapic_control/builder.rs
```diff
@@ -698,6 +698,174 @@ pub mod storage_control {
...
+
+        /// Creates a [Poller][google_cloud_lro::Poller] to work with `delete_folder_recursive`.
+        pub fn poller(
+            self,
+        ) -> impl google_cloud_lro::Poller<(), crate::model::DeleteFolderRecursiveMetadata>
+        {
+            type Operation = google_cloud_lro::internal::Operation<
+                wkt::Empty,
+                crate::model::DeleteFolderRecursiveMetadata,
+            >;
```
coryan, line 765: I think this is the first time we have a LRO returning `Empty` in `google-cloud-storage`. We need to do some tweaking in the hand-crafted code.

Please add the magic flag to **not** update `google-cloud-storage`. We will take care of updating that library.

librarian.yaml
```diff
@@ -13,17 +13,17 @@
 # limitations under the License.
 
 language: rust
-version: v0.8.1-0.20260208220643-e77cccf75c05
+version: v0.8.1-0.20260209175721-adbf644f2605
```
coryan, line 16: I would prefer separate PRs for librarian vs. the sources SHAs. It is easier to reason about what is changing and why.

src/storage/src/generated/gapic_control/builder.rs
dbolduc: d'oh

coryan:
Please rebase. Or start over from an up to date fork / branch.

librarian.yaml
coryan: I am now going to require that you make this change separately.

I caught https://github.com/googleapis/librarian/issues/3973 because I could tell what changes were expected from a librarian update. I would have completely missed the bug in a PR combining librarian and SHA changes.

coryan (changes requested):

miguelvelezsa:
Closing this, librarian config shouldn't be modified. Also should regenerate after fix https://github.com/googleapis/librarian/issues/3973

FYI @coryan.

PR #4637, opened 2026-02-10
cleanup(pubsub)!: move the BasePublisher out of crate::client::*

suzmue:
Resource sharing between publishers is an advanced use case. Moving the BasePublisher out of the main entry point (crate::client::*) simplifies the discoverability of the Publisher for most of our users while still leaving it accessible. It is likely we will also want to export the Publisher from this module as well (and do the same with the Subscriber).

For #4282

### DIFF
--- a/src/pubsub/src/publisher/implementation.rs
+++ b/src/pubsub/src/publisher/implementation.rs
@@ -16,6 +16,7 @@ use super::options::BatchingOptions;
 use crate::publisher::actor::BundledMessage;
 use crate::publisher::actor::ToDispatcher;
 use crate::publisher::builder::PublisherBuilder;
+
 use tokio::sync::mpsc::UnboundedSender;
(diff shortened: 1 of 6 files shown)


src/pubsub/src/publisher/implementation.rs
```diff
@@ -16,6 +16,7 @@ use super::options::BatchingOptions;
 use crate::publisher::actor::BundledMessage;
 use crate::publisher::actor::ToDispatcher;
 use crate::publisher::builder::PublisherBuilder;
+
```
coryan, line 19: remove the blank line? `cargo fmt` works better without them.

suzmue: Done.

PhongChuong (approved):

PhongChuong (approved):

Issue #4508, opened 2026-02-02
Pub/Sub Subscriber API review


dbolduc:
Everything should have examples. Common use cases should be documented. etc.

There will be a design review. We will collect findings in this issue, and resolve them.

dbolduc:
We did most of the reviews.

I will leave this open, in case we do a review of the APIs related to shutdown.

dbolduc:
> I will leave this open, in case we do a review of the APIs related to shutdown.

#5257 included a review of the shutdown APIs. Closing.

PR #4639, opened 2026-02-10
docs(pubsub): handler examples

dbolduc:
Part of the work for #4508

I am too lazy to play the doc pdf game. Here is a screenshot that googlers can see: https://screenshot.googleplex.com/7tNufCHHTjudwDy

### DIFF
--- a/src/pubsub/src/subscriber/handler.rs
+++ b/src/pubsub/src/subscriber/handler.rs
@@ -12,6 +12,34 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Handlers for acknowledging or rejecting messages.
+//!
+//! To acknowledge (ack) a message, you call [`Handler::ack()`].
+//!
+//! To reject (nack) a message, you [`drop()`][Drop::drop] the handler. The
+//! message will be redelivered.
+//!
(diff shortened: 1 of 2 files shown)


src/pubsub/src/subscriber/handler.rs
```diff
@@ -12,6 +12,34 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+//! Handlers for acknowledging or rejecting messages.
+//!
+//! To acknowledge (ack) a message, you call [`Handler::ack()`].
```
coryan, line 17: I think you are going to prove me wrong, but I expect we use `[Handler::ack()]` (without backticks inside the brackets) more often.

dbolduc: Regardless, I think code should look like `code`.

---

But yes, I do think we omit the backticks more often than not. e.g. when linking builders back to the client methods.

dbolduc: https://doc.rust-lang.org/rustdoc/write-documentation/linking-to-items-by-name.html

Most of the examples in here stylize code links with the backticks

PR #4641, opened 2026-02-10
docs: Generate sample for List RPCs

amanda-tarafa:
Note that for list like operations (e.g. search_folders in resourcemanager v3) we also generate the sample stub, we just don't know (yet) which fields to initialize.

### DIFF
--- a/src/generated/cloud/speech/v2/src/client.rs
+++ b/src/generated/cloud/speech/v2/src/client.rs
@@ -155,6 +155,26 @@ impl Speech {
     }
 
     /// Lists Recognizers.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_speech_v2::client::Speech;
+    /// use google_cloud_gax::paginator::ItemPaginator as _;
+    /// use google_cloud_speech_v2::Result;
+    /// async fn sample(
+    ///    client: &Speech,
+    ///    parent: &str
+    /// ) -> Result<()> {
+    ///     let mut list = client
+    ///         .list_recognizers()
+    ///         .set_parent(parent)
+    ///         .by_item();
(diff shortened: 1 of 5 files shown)


src/generated/cloud/speech/v2/src/client.rs
```diff
@@ -619,6 +659,26 @@ impl Speech {
...
+    /// ) -> Result<()> {
+    ///     let mut list = client
+    ///         .list_phrase_sets()
+    ///         .set_parent(parent)
+    ///         .by_item();
+    ///     while let Some(item) = list.next().await.transpose()? {
+    ///         println!("{:?}", item);
+    ///     }
+    ///     Ok(())
+    /// }
```
coryan, line 680: These samples are starting to consume a lot of vertical space.  They will be hard to use in IDE tooltips, we should try to make them look more like this:


```suggestion
    /// # use google_cloud_speech_v2::client::Speech;
    /// use google_cloud_gax::paginator::ItemPaginator as _;
    /// use google_cloud_speech_v2::Result;
    /// async fn sample(client: &Speech, parent: &str) -> Result<()> {
    ///     let mut list = client.list_phrase_sets().set_parent(parent).by_item();
    ///     while let Some(item) = list.next().await.transpose()? {
    ///         println!("{:?}", item);
    ///     }
    ///     Ok(())
    /// }
```

zashed (approved):

src/generated/cloud/speech/v2/src/client.rs
amanda-tarafa: Acked, will do, in upcoming PRs.

amanda-tarafa:
This is now ready for review.

PR #4647, opened 2026-02-10
cleanup(pubsub): fix broken tests

suzmue:
The PR that added this test was incorrectly merged and used the stale name for Message. This also fixes a potential source of flakiness where the timer for a batch fires before the task is aborted.

### DIFF
--- a/src/pubsub/src/publisher/client.rs
+++ b/src/pubsub/src/publisher/client.rs
@@ -309,7 +309,7 @@ mod tests {
         }
     }
 
-    #[tokio::test]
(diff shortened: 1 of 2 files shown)


suzmue:
> nit: PR title? maybe: `cleanup(pubsub): avoid timing in test`

Yeah I had something similar to that at first but it also importantly uses the correct name for Message now so went more general.

Thank you for quick review!

dbolduc:
> uses the correct name for Message

Oh duh. You're right. The original title was good. I somehow missed that.

Issue #3970, opened 2025-12-02
Determine if special handling is needed for ordered delivery in subscriber


dbolduc:
Pub/Sub supports ordered delivery for subscriptions.

In callback-based APIs, the client libraries guarantee that a callback for a message with a given ordering key completes before any subsequent callbacks on the same ordering key start. So says the Pub/Sub folks.

Note that applications can acknowledge messages outside of these callbacks. i.e. serving subsequent messages for a given ordering key is not blocked on receiving acknowledgement of previous messages.

In a streaming API, each call to `next()` of the stream is essentially a callback starting and finishing. So I do not think we need to do any special handling to make the same guarantee.

This issue is to confirm that my understanding is correct. We can close it if it is.

If an application acks messages out of order for a given ordering key, that is their prerogative. If the Ack IDs make it into the same ack batch, they will be acked. If not, they won't be, and there will be redelivery of some of the messages.

We do not want to prevent this at the cost of letting applications intake more messages into an (ordered) queue before acking.

dbolduc:
I will close this when we have an integration test.

coryan:
> I will close this when we have an integration test.

An integration test can tell you if something is wrong, but not if it is correct.  And that is not super comforting for things like this, where (in my experience) the problems only manifest at scale and under heavy load.

dbolduc:
> > I will close this when we have an integration test.
> 
> An integration test can tell you if something is wrong, but not if it is correct. And that is not super comforting for things like this, where (in my experience) the problems only manifest at scale and under heavy load.

What would give you confidence?

coryan:
> 
> What would give you confidence?

A code review from somebody who understands how these clients should work (not me at the moment, I am sleep deprived).

A unit test where a fake service delivers messages in correct, but challenging orders: maybe reconnects where the same messages are resent? Can that happen? What about gaps, simulating another client receiving messages? Maybe batches with messages interleaved from two different keys?  Can the application ack many messages from one key and starve the other key?


A TLA+ specification, but you would need somebody smarter than me to read it.

A Jepsen report 😜

coryan:
Also: an endurance test would help: K clients, receiving N ordering keys, the publishers send many messages for hours.

dbolduc:
I think we are good. Maybe (exactly-once + ordering) needs a test, but if so, we track that in another issue.

PR #4648, opened 2026-02-10
refactor(pubsub): prepare for ordered delivery test

dbolduc:
Some refactors around passing `String` / `&str` + generalizing the `create_subscription` helper. We will create an ordered subscription and use it in a follow up PR.

Related to #3970

### DIFF
--- a/src/pubsub/examples/src/lib.rs
+++ b/src/pubsub/examples/src/lib.rs
@@ -137,16 +137,13 @@ pub async fn cleanup_test_subscription(
 pub async fn create_test_subscription(
     topic_name: &str,
 ) -> anyhow::Result<(SubscriptionAdmin, Subscription)> {
-    create_test_subscription_impl(topic_name, |s| s).await
+    create_test_subscription_with_request(topic_name, Subscription::new()).await
 }
 
-async fn create_test_subscription_impl<F>(
+async fn create_test_subscription_with_request(
     topic_name: &str,
-    modifier: F,
-) -> anyhow::Result<(SubscriptionAdmin, Subscription)>
-where
-    F: Fn(Subscription) -> Subscription,
-{
+    request: Subscription,
+) -> anyhow::Result<(SubscriptionAdmin, Subscription)> {
     let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
     let client = SubscriptionAdmin::builder().with_tracing().build().await?;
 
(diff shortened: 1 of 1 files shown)


src/pubsub/examples/src/lib.rs
```diff
@@ -135,8 +135,18 @@ pub async fn cleanup_test_subscription(
...
+    create_test_subscription_impl(topic_name, |s| s).await
+}
+
+async fn create_test_subscription_impl<F>(
+    topic_name: &str,
+    modifier: F,
+) -> anyhow::Result<(SubscriptionAdmin, Subscription)>
+where
+    F: Fn(Subscription) -> Subscription,
+{
```
coryan: I think the Rusty way is to create a function to initialize the subscription parameters and another to create them:

```rust
fn new_subscription_request(topic_name: &str) -> Result<Subscription> { ...}

async fn create_test_subscription_with_request(subscription: Subscription) -> Result<(SubscriptionAdmin, Subscription)> { ...}

pub async fn create_test_subscription(topic_name: &str) -> Result<(SubscriptionAdmin, Subscription)> {
    let sub = new_subscription_request(topic_name)?;
    create_test_subscription_with_request(sub).await
}
```

And later you just modify the `sub` between the two calls.

coryan (changes requested):
I think you should change this, a function argument for this is gross.

src/pubsub/examples/src/lib.rs
dbolduc: I went for passing in a prototype of a `Subscription`, to avoid plumbing the `project_id` from the env var.

src/pubsub/examples/src/lib.rs
```diff
@@ -135,7 +135,14 @@ pub async fn cleanup_test_subscription(
...
+    topic_name: &str,
+) -> anyhow::Result<(SubscriptionAdmin, Subscription)> {
+    create_test_subscription_with_request(topic_name, Subscription::new()).await
+}
+
+async fn create_test_subscription_with_request(
+    topic_name: &str,
+    request: Subscription,
 ) -> anyhow::Result<(SubscriptionAdmin, Subscription)> {
     let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
```
coryan, line 147: BTW, I think there is a helper for this in `google-cloud-test-utils`...

coryan (approved):

src/pubsub/examples/src/lib.rs
dbolduc: Good call. I see a bunch of similar code that could use the helper. I'll merge this and send a follow up to update them all.

PR #4651, opened 2026-02-10
fix(pubsub): Publisher byte threshold off by one error

PhongChuong:
This PR fix the issue where the Publisher configured byte threshold is not respected. In the bug, the Publisher will add the next message to the pending batch until the batch size is greater then the byte threshold. The correct behavior is to batch the messages such that each batch size is less or equal to the configured byte threshold.

In the fix, we add the next message to the batch only if the batch will stay within the configured byte threshold. This is done lazily and the pending batch is not Send until we know that the next message will not fit.

### DIFF
--- a/src/pubsub/src/publisher/actor.rs
+++ b/src/pubsub/src/publisher/actor.rs
@@ -198,7 +198,7 @@ pub(crate) struct BatchActorContext {
     topic: String,
     client: GapicPublisher,
     rx: mpsc::UnboundedReceiver<ToBatchActor>,
-    pending_batch: Batch, // TOD(NOW): Not needed anymore.
+    batching_options: BatchingOptions,
 }
 
 impl BatchActorContext {
@@ -209,10 +209,10 @@ impl BatchActorContext {
         rx: mpsc::UnboundedReceiver<ToBatchActor>,
     ) -> Self {
         BatchActorContext {
-            pending_batch: Batch::new(topic.len() as u32, batching_options),
             topic,
             client,
             rx,
+            batching_options,
         }
     }
 }
@@ -253,6 +253,10 @@ impl ConcurrentBatchActor {
     pub(crate) async fn run(mut self) {
         // We have multiple inflight batches concurrently.
         let mut inflight = JoinSet::new();
+        let mut batch = Batch::new(
+            self.context.topic.len() as u32,
+            self.context.batching_options.clone(),
+        );
         loop {
             tokio::select! {
                 // Remove completed inflight batches.
@@ -262,10 +266,10 @@ impl ConcurrentBatchActor {
                 msg = self.context.rx.recv() => {
                     match msg {
                         Some(ToBatchActor::Publish(msg)) => {
-                            self.move_to_batch_and_flush_at_threshold(&mut inflight, msg);
+                            self.add_msg_and_flush(&mut inflight, &mut batch, msg);
                         },
                         Some(ToBatchActor::Flush(tx)) => {
-                            self.flush(&mut inflight);
+                            self.flush(&mut inflight, &mut batch);
                             inflight.join_all().await;
                             inflight = JoinSet::new();
                             let _ = tx.send(());
@@ -276,7 +280,7 @@ impl ConcurrentBatchActor {
                         None => {
                             // This isn't guaranteed to execute if a user does not .await on the
                             // corresponding PublishFutures.
-                            self.flush(&mut inflight);
+                            self.flush(&mut inflight, &mut batch);
                             inflight.join_all().await;
                             break;
                         }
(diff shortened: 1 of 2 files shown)


src/pubsub/src/publisher/actor.rs
```diff
@@ -197,9 +197,8 @@ impl Dispatcher {
 pub(crate) struct BatchActorContext {
     topic: String,
     client: GapicPublisher,
-    batching_options: BatchingOptions,
     rx: mpsc::UnboundedReceiver<ToBatchActor>,
-    pending_batch: Batch,
+    pending_batch: Batch, // TOD(NOW): Not needed anymore.
```
suzmue: ?

src/pubsub/src/publisher/actor.rs
```diff
@@ -428,24 +420,49 @@ impl SequentialBatchActor {
...
         }
+        self.handle_inflight_join(inflight.join_next().await);
     }
 
-    // Move pending messages to the pending batch respecting batch thresholds.
-    pub(crate) fn move_to_batch(&mut self) {
-        while let Some(publish) = self.pending_msgs.pop_front() {
+    // Move message to the pending batch respecting batch thresholds
+    // and flush the batch if it is full.
+    pub(crate) fn move_to_batch_and_flush_at_threshold(
```
suzmue: nit: naming is super wordy.

Maybe just `move_msg_and_flush`

src/pubsub/src/publisher/batch.rs
```diff
@@ -63,6 +66,16 @@ impl Batch {
             })
     }
 
+    pub(crate) fn at_threshold(&mut self) -> bool {
+        self.len() as u32 >= self.batching_options.message_count_threshold
+            || self.size() >= self.batching_options.byte_threshold
+    }
+
+    // Return true if adding the next message is within the byte threshold.
+    pub(crate) fn within_byte_threshold(&mut self, next: &BundledMessage) -> bool {
```
suzmue: nit: might make code more clear to call this `can_add_to_pending` or something.

src/pubsub/src/publisher/actor.rs
```diff
@@ -428,24 +420,49 @@ impl SequentialBatchActor {
...
-            if self.context.at_batch_threshold() {
+            if self.context.pending_batch.at_threshold() {
+                self.context.pending_batch.flush(
+                    self.context.client.clone(),
+                    self.context.topic.clone(),
+                    inflight,
+                );
                 break;
             }
         }
```
suzmue, line 471: I think this might be easier to parse if its split into two phases, making clear that we can only have a single flush
```suggestion
        let mut should_flush = false;
        while let Some(next) = self.pending_msgs.front() {
            if !self.context.pending_batch.within_byte_threshold(next) {
               /// Flush the previous batch since the batch would overflow.
                should_flush = true
                break;
            }
            let publish = self
                .pending_msgs
                .pop_front()
                .expect("front should contain an element");
           self.context.pending_batch.push(publish);
            if self.context.pending_batch.at_threshold() {
                should_flush = true
                break;
             }
          }
          
          if should_flush {
             self.context.pending_batch.flush(
                      self.context.client.clone(),
                      self.context.topic.clone(),
                      inflight,
                  );
          }
```

src/pubsub/src/publisher/actor.rs
suzmue: Also maybe the functions could use different names for Sequential vs Not, since they have different behaviors.

src/pubsub/src/publisher/actor.rs
```diff
@@ -310,9 +297,20 @@ impl ConcurrentBatchActor {
...
-    // Move msg to the pending batch.
-    pub(crate) fn move_to_batch(&mut self, msg: BundledMessage) {
+    // Move message to the pending batch respecting batch thresholds
+    // and flush the batch if it is full.
+    pub(crate) fn move_to_batch_and_flush_at_threshold(
+        &mut self,
+        inflight: &mut JoinSet<crate::Result<()>>,
+        msg: BundledMessage,
+    ) {
+        if !self.context.pending_batch.within_byte_threshold(&msg) {
```
suzmue: nit: can we pull out self.context.pending_batch to variable? I haven't checked that the types line up but it makes every line so long to have self.context.pending_batch everywhere. Something like:

let mut batch = &self.context.pending_batch;

src/pubsub/src/publisher/actor.rs
suzmue: nit: naming is super wordy.

Maybe just `add_msg_and_flush`

src/pubsub/src/publisher/actor.rs
```diff
@@ -921,4 +940,131 @@ mod tests {
...
+                    .set_message_count_threshold(MAX_MESSAGES)
+                    .set_byte_threshold(25_u32), // The current test generates 24 byte single message batches.
+                actor_rx,
+            )
+            .run(),
+        );
+
+        let mut publish_rxs = VecDeque::new();
+        publish_random_data!(publish_rxs, actor_tx, 1);
+        // We flush here otherwise the last message will await forever since it never exceed the byte threshold.
```
suzmue, line 1075: aside: I think actually with start_paused = true, the testing framework will fire the timer if its the only thing that can happen, so this might not be necessary.

src/pubsub/src/publisher/actor.rs
PhongChuong: I wanted to move the batch out of the context and added a TODO to fix before sending out the PR but couldn't find it since there was a typo in the "TOD". It's fixed now.

src/pubsub/src/publisher/actor.rs
PhongChuong: The flush is needed here because the timer is in the Dispatcher.

PhongChuong:
@suzmue , thanks for the review. PTAL.

Issue #4519, opened 2026-02-03
Pub/Sub - send receipt ModAck from the subscriber


dbolduc:
We need to send a receipt ModAck upon receiving a message. So says the Pub/Sub folks. I will describe why we should / should not do this.

The server defaults the ack deadline to 10 minutes for all messages, regardless of the subscription setting. It is only upon receiving the initial ModAck that they set the server-side deadline (to 10s by default).

So if there is some transient error acking a message, and we never sent a receipt ModAck, the message gets stuck in a frozen state, where the server holds it for 10 minutes ([which I have seen](https://github.com/googleapis/google-cloud-rust/issues/4275#issuecomment-3825782229)).

Customers who configure a 10s ack deadline will be confused as to why it takes 10 minutes to redeliver the message. This is very convincing.

Also the Pub/Sub server teams expect a receipt ModAck for reasons (maybe to gauge stream health?, and who knows what features they will come up with in the future.). It is also a headache if the Pub/Sub customer support teams cannot assume the client is sending receipt ModAcks. This is somewhat convincing.

---

Still, sending receipt ModAcks means extra RPCs, that are not always necessary. Depending on implementation, it might even be an extra RPC *per  `StreamingPullResponse`*.

Also, these receipt ModAcks can fail for all the same reasons as an Ack RPC can fail.

So is "RPCs can fail" really a reason to send receipt ModAcks, or is it more an argument to retry acknowledgements that fail with transient errors?

(There are other failure modes, like an application crashing. I am not sure that convinces me we benefit from the immediate ModAck, though. Maybe you lose less messages, but the receipt ModAck won't necessarily go out in a crash. 🤷 ).

---

Ultimately, the onus will be on us to provide evidence that there is a reason NOT to send a receipt ModAck. (e.g. if the throughput gets cut in half, or something).

And even if we have this evidence, I still think we need to send receipt ModAcks by default. I think we could justify an option to turn off receipt ModAcks. (Note that we would still send lease extension ModAcks, it just may be after 5s instead of immediately).

dbolduc:
We send a receipt modack within some period. Good enough for me.

PR #4660, opened 2026-02-11
test(pubsub): use mockall withf to validate request parameters

PhongChuong:


src/pubsub/src/publisher/client.rs
```diff
@@ -443,38 +435,26 @@ mod tests {
...
-                    vec![
-                        Message::new().set_data("hello"),
-                        Message::new().set_data("world")
-                    ]
-                );
-                publish_ok(r, o)
-            }
-        });
+        mock.expect_publish()
+            .withf(|r, _| r.messages.len() == 2 && r.messages[0].data == r.messages[1].data)
```
dbolduc: this test must have a problem if it still passes.

`r.messages[0].data` used to be "hello" and `r.messages[1].data` used to be "world". They are not equal?

dbolduc: Mocks that get moved to background task are kinda unreliable btw. The mock can go out of scope, panic because the expectation is not met, but we never notice in the main thread.

https://github.com/asomers/mockall/discussions/341

If I run with `--no-capture` I see the panic:

```sh
$ cargo test -p google-cloud-pubsub dropping_handles_does_not -- --no-capture

thread 'publisher::client::tests::dropping_handles_does_not_prevent_publishing' (2437862) panicked at src/pubsub/src/publisher/client.rs:177:5:
MockGapicPublisher::publish(PublishRequest { topic: "my-topic", messages: [PubsubMessage { data: b"hello", attributes: {}, message_id: "", publish_time: None, ordering_key: "" }, PubsubMessage { data: b"world", attributes: {}, message_id: "", publish_time: None, ordering_key: "" }] }, RequestOptions { idempotent: None, user_agent: None, attempt_timeout: None, retry_policy: None, backoff_policy: None, retry_throttler: None, polling_error_policy: None, polling_backoff_policy: None, path_template: None }): No matching expectation found
test publisher::client::tests::dropping_handles_does_not_prevent_publishing ... ok
```

dbolduc:
> You should think about how to write these tests so that they fail on incorrect expectations.

And if you figure out a good solution, let me know, because I have the same problem in the subscriber. :confounded:. It makes me hesitant to do #4519 and #3975

src/pubsub/src/publisher/client.rs
PhongChuong: Good catch.
I fixed the validations.
Let me sit on this overnight and fix the panic capture in a later PR.

PR #4661, opened 2026-02-11
test(pubsub): interleaved ordering

dbolduc:
For #3970, #4012

Interleave publishing ordering and force multiple batches for extra-correctness guarantees.

I might also use multiple sessions, with some sort of lock mechanism around `expected_index` in a follow up.

### DIFF
--- a/tests/pubsub/src/ordering.rs
+++ b/tests/pubsub/src/ordering.rs
@@ -15,6 +15,7 @@
 use futures::future::join_all;
 use google_cloud_pubsub::client::{Publisher, Subscriber};
 use google_cloud_pubsub::model::Message;
+use std::collections::HashMap;
 
 pub async fn roundtrip(topic_name: &str, subscription_name: &str) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


tests/pubsub/src/ordering.rs
```diff
@@ -17,45 +17,54 @@ use google_cloud_pubsub::client::{Publisher, Subscriber};
...
+        while expected_index.iter().all(|&v| v == MESSAGES_PER_KEY) {
             let Some((m, h)) = session.next().await.transpose()? else {
                 anyhow::bail!("Stream somehow ended.")
             };
-            if m.ordering_key != ORDERING_KEY {
+            let Some(key) = m.ordering_key.strip_prefix(ORDERING_PREFIX) else {
                 continue;
-            }
+            };
+            let key = key.parse::<usize>()?;
```
coryan: I think a hashmap would save you a lot of trouble 🤷

dbolduc: Done. That was my instinct, but I gave up too early.

PR #4663, opened 2026-02-11
cleanup(pubsub): fix background panic in test

dbolduc:
The client sends lease extensions in the `ModifyAckDeadline` rpc. Our background task was silently panicking on exit, because the mock did not expect these calls.

Found with:

```
cargo test -p google-cloud-pubsub subscriber:: -- --no-capture
```

Note that handling lease extensions is not the point of the test. It verified the thing it needed to verify. The background panic was just messy.


dbolduc:
cc: @PhongChuong - I had one of these too. :confounded:

PhongChuong (approved):

PR #4665, opened 2026-02-11
impl(auth): handle expiration for trust boundary value

alvarowolfx:
Towards #4186

### DIFF
--- a/src/auth/src/access_boundary.rs
+++ b/src/auth/src/access_boundary.rs
@@ -36,11 +36,19 @@ const COOLDOWN_INTERVAL: Duration = Duration::from_secs(15 * 60);
 
 #[derive(Debug)]
 pub(crate) struct AccessBoundary {
+    /// A channel to keep track of the boundary state.
+    /// - `None`: We haven't fetched anything yet (uninitialized).
+    /// - `Some(...)`: We successfully talked to the IAM service or have a customer provided override.
+    ///   These values come with a TTL so we know how long to keep them around.
     rx_header: watch::Receiver<Option<BoundaryValue>>,
 }
 
 #[derive(Debug, Clone)]
 struct BoundaryValue {
+    /// This is an `Option` because the IAM service can signal that the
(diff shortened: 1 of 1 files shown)


src/auth/src/access_boundary.rs
```diff
@@ -22,21 +22,36 @@ use std::clone::Clone;
...
 
 #[derive(Debug)]
 pub(crate) struct AccessBoundary {
-    rx_header: watch::Receiver<Option<String>>,
+    rx_header: watch::Receiver<Option<BoundaryValue>>,
+}
+
+#[derive(Debug, Clone)]
+struct BoundaryValue {
+    value: Option<String>,
```
coryan, line 52: Maybe explain what `Some(BoundaryValue::new(None)` means because it is not obvious to me.

src/auth/src/access_boundary.rs
```diff
@@ -431,6 +456,70 @@ mod tests {
...
+            value: Some("old-value".to_string()),
+            expires_at,
+        }));
+
+        // value is valid
+        let val = access_boundary.header_value();
+        assert_eq!(val.as_deref(), Some("old-value"), "{val:?}");
+
+        // advance time to expire the value
+        tokio::time::advance(Duration::from_secs(11)).await;
```
coryan: Maybe use a constant (shared with line 469) because it was not obvious why 11 was important.

src/auth/src/access_boundary.rs
alvarowolfx: that's a good point and I was going back and forth on this because of the double Option thing. 

the channel has an `Option<BoundaryValue>` to differentiate if is not initialized (None) or if we already have some information that came from the IAM service or from a customer provides overrides (the later two have TTL). The service can also signal that the given credential has no access boundary, so we save it as None, but with a TTL (so we don't fetch more often).

coryan: Makes sense, please included that as a comment

zashed (approved):

PR #4669, opened 2026-02-12
test(integration-tests-o11): prepare for `showcase` bump

coryan:
We don't want these tests to fail when `showcase` gets a version bump.


coryan:
/cc: @westarle in case other languages have the same problems in their tests.

westarle:
> /cc: @westarle in case other languages have the same problems in their tests.

Thank you!  We should be OK.

PR #4675, opened 2026-02-12
feat: add `google_cloud_unstable_resource_name` feature flag

haphungw:
This is a part of the upcoming OpenTelemetry tracing work and  allows us to use `#[cfg(google_cloud_unstable_resource_name)`] without triggering lint errors.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -512,6 +512,7 @@ storage-samples         = { path = "src/storage/examples" }
 
 [workspace.lints.rust]
 unexpected_cfgs = { level = "deny", check-cfg = [
+  'cfg(google_cloud_unstable_resource_name)',
   'cfg(google_cloud_unstable_storage_bidi)',
   'cfg(google_cloud_unstable_tracing)',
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -512,6 +512,7 @@ storage-samples         = { path = "src/storage/examples" }
 
 [workspace.lints.rust]
 unexpected_cfgs = { level = "deny", check-cfg = [
+  'cfg(google_cloud_unstable_resource_name)',
```
dbolduc, line 515: Why not use `cfg(google_cloud_unstable_tracing)` ?

haphungw: Wasn't `unstable_tracing` reserved for the propagation work? I can definitely use it here if we want to fold them together.

dbolduc: `unstable_tracing` is for all things tracing. e.g. capturing compute resource names in traces.

haphungw: sounds good, closing this PR

PR #4679, opened 2026-02-13
ci: add a gcb trigger to run tokio unstable tests

PhongChuong:
Add a new gcb trigger to run tests with RUSTFLAGS="--cfg tokio_unstable". The new trigger will be applied post merge.

Within the tests, we use tokio-unstable shutdown_runtime to propagate mockall panic to the test results.
Running the tests with:
- cargo test: runs the tests normally without tokio-unstable
- RUSTFLAGS="--cfg tokio_unstable" cargo test: runs the tests with unhandled_panic = "shutdown_runtime"

Followup PRs will enable unhandled_panic = "shutdown_runtime" in more tests.

The new terraform plan will be applied post review/merge.

For #4733

### DIFF
--- a/.gcb/builds/triggers/main.tf
+++ b/.gcb/builds/triggers/main.tf
@@ -40,6 +40,8 @@ locals {
     "--cfg google_cloud_unstable_storage_bidi"
   ])
 
+  tokio_unstable_flags = "--cfg tokio_unstable"
+
   # These builds appear in both the PR (Pull Request) triggers and the
   # PM (Post Merge) triggers. See below for builds that only appear in one.
   common_builds = {
@@ -125,6 +127,11 @@ locals {
     test-current = {
       config = "complex.yaml"
       flags  = local.unstable_flags
+      script = "test"
+    }
+    test-tokio-unstable = {
+      config = "complex.yaml"
+      flags  = local.tokio_unstable_flags
       script = "test"
     }
     test-msrv = {


src/pubsub/src/publisher/client.rs
```diff
@@ -459,8 +459,16 @@ mod tests {
...
 
+    #[cfg_attr(
+        tokio_unstable,
+        tokio::test(
+            start_paused = true,
+            flavor = "current_thread",
+            unhandled_panic = "shutdown_runtime"
+        )
+    )]
+    #[cfg_attr(not(tokio_unstable), tokio::test(start_paused = true))]
```
coryan, line 470: I think I like this better:

```suggestion
    #[cfg(tokio_unstable)]
    #[tokio::test(start_paused = true, flavor = "current_thread", unhandled_panic = "shutdown_runtime")]
    async fn no_task_leaks() -> anyhow::Result<()> {
          dropping_handles_does_not_prevent_publishing().await
    }
    
    #[tokio::test(start_paused = true)]
```

Thoughts?

PhongChuong: This runs into the following error:
`Cannot start a runtime from within a runtime. This happens because a function (like `block_on`) attempted to block the current thread while the thread is being used to drive asynchronous tasks.` because the tokio_unstable #[tokio::test...] creates a runtime which conflict with the runtime created by the embedded test dropping_handles_does_not_prevent_publishing #[tokio::test...].

IIRC, I previously played around with embedding the tests by having 2 sets of test "drivers" but decided against it. E.g.,
```
    async fn all_tests() -> anyhow::Result<()> {
          dropping_handles_does_not_prevent_publishing().await
    }

    #[cfg(tokio_unstable)]
    #[tokio::test(start_paused = true, flavor = "current_thread", unhandled_panic = "shutdown_runtime")]
    async fn no_task_leaks() -> anyhow::Result<()> {
          dropping_handles_does_not_prevent_publishing().await
    }
```

I didn't like the idea of having to maintain 2 lists of test + compressing the test results when running `cargo test`.

coryan: TIL: thanks.

Issue #4650, opened 2026-02-10
Pubsub should return error when Publish message size is larger than batch threshold


PhongChuong:
The PubSub Publisher should return an error to the application when attempt to publish a message (including topic + ordering key) that has byte size greater than the configured batch size.

We should decide if we should also pause publishing when this error occurs (leaning no).

suzmue:
> should return an error

Let's check with the Pub/Sub team if this is expected. Since we do batching, we are responsible for not creating batches that are too large, but I'm not sure this applies to messages that are too large. In general client libraries are not supposed to do client side verification.

> We should decide if we should also pause publishing when this error occurs (leaning no).

I think we will want to pause the ordering key. This would count as a permanent failure and would break guarantees if we allowed publishing to continue.

suzmue:
Update: Looks like the Go client library does check this limit in the client library: https://github.com/googleapis/google-cloud-go/blob/b232bb1265bbfa484be7db2c9ae8cf3984871c95/pubsublite/internal/wire/publish_batcher.go#L144-L145

suzmue:
Also see this integration test: https://github.com/googleapis/google-cloud-go/blob/b232bb1265bbfa484be7db2c9ae8cf3984871c95/pubsub/v2/integration_test.go#L543

suzmue:
@PhongChuong  that setting I was looking at was for flow control NOT for the batch settings, so it seems we should not be returning an error for this.

We still want to respect the service limits when batching, but when there is a message that is too large, we should send it anyway and just return the service error

cc @hongalex

PR #4681, opened 2026-02-13
fix(pubsub)!: return PublishError from Publish instead of an Arc<gax::error>

PhongChuong:
Switch Publisher Publish to return PublishError instead of Arc.

For SendError, the application can inspect the source of the send error. The source is wrapped in Arc as an error in a Batch send can cause multiple Publish errors.

The tests are also updated to check for the new error type.

For #4650

### DIFF
--- a/src/pubsub/src/publisher/model_ext.rs
+++ b/src/pubsub/src/publisher/model_ext.rs
@@ -61,17 +61,6 @@ impl Future for PublishFuture {
     }
 }
 
-// fn convert_error(e: crate::error::PublishError) -> Arc<crate::Error> {
-//     // TODO(#3689): The error type for these are not ideal, we will need will
-//     // need to handle error propagation better.
-//     match e {
-//         crate::error::PublishError::SendError(s) => s,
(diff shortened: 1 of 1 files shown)


src/pubsub/src/publisher/client.rs
```diff
@@ -897,14 +898,8 @@ mod tests {
...
         got_err = msg_1_handle.await.unwrap_err();
-        let source = got_err
-            .source()
-            .and_then(|e| e.downcast_ref::<crate::error::PublishError>());
         assert!(
-            matches!(
-                source,
-                Some(crate::error::PublishError::OrderingKeyPaused(()))
-            ),
+            matches!(got_err, crate::error::PublishError::OrderingKeyPaused(())),
```
suzmue, line 902: In a possible future PR, is it possible to clean these up so that we don't have to have a unit type value at all?

suzmue (approved):
PR Title nit: reword to finish the sentence "This change modifies the crate to ...". For example "return a PublishError from publish instead of a Arc<gax::error>"

PR description nit: no colon needed after "For"

Issue #3178, opened 2025-08-29
Tracing for `Storage` client


dbolduc:
- [ ] Need to create a span for each ranged read after `open_object()`.
- [ ] The storage HTTP requests need all the annotations, including resource name (the bucket) and the URL template.
- [x] #4907

----
We should ideally have spans for each client operation that is a parent of the RPCs going out.

We should see if there is a performance hit to always instrumenting the internals

PR #4682, opened 2026-02-13
refactor(integration-tests-o11y): e2e test into smaller functions

coryan:
Breaking the test into smaller functions makes it easier to reuse it later.

Motivated by #3178

### DIFF
--- a/tests/o11y/src/e2e.rs
+++ b/tests/o11y/src/e2e.rs
@@ -20,15 +20,15 @@ use google_cloud_gax::error::rpc::Code;
 use google_cloud_trace_v1::client::TraceService;
 use google_cloud_trace_v1::model::Trace;
 use opentelemetry_sdk::trace::SdkTracerProvider;
-use tracing::subscriber::DefaultGuard;
+use tokio::sync::OnceCell;
 use tracing_subscriber::layer::SubscriberExt;
-use tracing_subscriber::util::SubscriberInitExt;
 
 pub const SERVICE_NAME: &str = "e2e-telemetry-test";
+static PROVIDER: OnceCell<anyhow::Result<SdkTracerProvider>> = OnceCell::const_new();
 
 /// Waits for a trace to appear in Cloud Trace.
 ///
-/// Traces may take a few seconds to propagate from the collector endpoints to
+/// Traces may take a few minutes to propagate from the collector endpoints to
 /// the service. This function retrieves the trace, polling if the trace is
 /// not found.
 pub async fn wait_for_trace(project_id: &str, trace_id: &str) -> anyhow::Result<Trace> {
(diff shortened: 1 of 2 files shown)


tests/o11y/src/e2e.rs
```diff
@@ -12,98 +12,26 @@
...
-        trace_id, project_id
-    );
-
-    // 4. Force flush to ensure spans are sent.
-    provider.force_flush()?;
-
-    // 5. Verify (Poll Cloud Trace API)
+/// Waits for a trace to appear in Cloud Trace.
+///
+/// Traces may take a few seconds to propagate from the collector endpoints to
```
westarle: Nit:

```suggestion
/// Traces may take a few minutes to propagate from the collector endpoints to
```

Above the gnarly retry loop we say "// Traces can take several minutes to propagate after they have been written."

tests/o11y/src/e2e.rs
```diff
@@ -142,20 +70,39 @@ pub async fn run() -> anyhow::Result<()> {
...
 
     let trace = trace.ok_or_else(|| anyhow::anyhow!("Timed out waiting for trace"))?;
+    Ok(trace)
+}
 
-    // 6. Assertions
-    // Check for root span
-    let root_found = trace.spans.iter().any(|s| s.name == span_name);
-    assert!(root_found, "Root span '{}' not found in trace", span_name);
+/// Sets up a OTLP provider tracing provider to use with Cloud Trace.
```
westarle: nit
```suggestion
/// Sets up a OTLP tracing provider to use with Cloud Trace.
```

tests/o11y/src/e2e.rs
```diff
@@ -142,20 +70,39 @@ pub async fn run() -> anyhow::Result<()> {
...
+///
+/// This function configures a OpenTelemetry provider that sends traces to Google
+/// Trace via the OTLP endpoint (telemetry.googleapis.com).
+///
+/// This uses ADC, and configure a quota project for user credentials because
+/// telemetry endpoint rejects user credentials without the quota user project.
+///
+/// Note that some other services reject requests *with* a quota user project, so
+/// we cannot assume it is set.
+pub async fn set_up_otel_provider(
```
westarle: nit

Returns a configured provider and a guard.  While the guard is not dropped, the provider will be configured as this thread's default subscriber for tracing.

```rust
{
    let (provider, _guard) = set_up_otel_provider(project_id)?;
    // my_span sent to Cloud Trace.
    tracing::info_span!("my_span", "otel.name" = "my_span");
    provider.force_flush();
}
```

tests/o11y/src/e2e.rs
```diff
@@ -142,20 +70,39 @@ pub async fn run() -> anyhow::Result<()> {
...
+}
 
-    // 6. Assertions
-    // Check for root span
-    let root_found = trace.spans.iter().any(|s| s.name == span_name);
-    assert!(root_found, "Root span '{}' not found in trace", span_name);
+/// Sets up a OTLP provider tracing provider to use with Cloud Trace.
+///
+/// This function configures a OpenTelemetry provider that sends traces to Google
+/// Trace via the OTLP endpoint (telemetry.googleapis.com).
```
westarle:
```suggestion
/// This function configures a OpenTelemetry provider that sends traces to Cloud
/// Trace via the OTLP endpoint (telemetry.googleapis.com).
```

tests/o11y/src/e2e.rs
```diff
@@ -142,20 +70,39 @@ pub async fn run() -> anyhow::Result<()> {
...
-    // Check for showcase client span
-    let client_span_name = "google-cloud-showcase-v1beta1::client::Echo::echo";
-    let client_found = trace.spans.iter().any(|s| s.name == client_span_name);
-    assert!(
-        client_found,
-        "Client library span '{}' not found in trace",
-        client_span_name
-    );
+    // Install subscriber
+    let guard = tracing_subscriber::Registry::default()
```
westarle: I'm a bit worried about this causing flakes.  We might want to move to this if the tests get more complicated and get moved between threads:

```rust
set_up_otel_provider_once(...) -> ... {
    ... only once ...
    tracing::subscriber::set_global_default(subscriber);
}
```

tests/o11y/src/e2e.rs
coryan: Done.

tests/o11y/src/e2e.rs
coryan: Fixed.

tests/o11y/src/e2e.rs
coryan: Fixed.

tests/o11y/src/e2e.rs
coryan: I think this no longer applies.

tests/o11y/src/e2e.rs
coryan: Hmmm... Yuck, I hate globals.  I tried using a single threaded tokio runtime (and enforcing that it was), but that breaks with `httptest` for reasons I cannot troubleshoot at the moment.

Using a `tokio::sync::OnceCell` seems like the best path forward.

PR #4702, opened 2026-02-16
chore: update librarian circa 2026-02-16

coryan:
This new version requires using `skip_release` instead of `skip_publish`, and includes some new samples.


coryan:
@amanda-tarafa this includes your changes, I was too lazy to separate the `skip_publish` changes from yours.

Issue #3019, opened 2025-08-24
Define default retry policy for subscriber and publisher


coryan:
These are different from Aip194, and should define good timeouts, and maximum number of retries (or maximum time retrying).

dbolduc:
I think we can close this. @suzmue added this for the publisher and I am not exposing any retry policies for the subscriber unary RPCs

PR #4711, opened 2026-02-17
feat(pubsub): define default retry policy for publisher

suzmue:
This default retry policy was determined by looking at the reference docs and comparing with the Java and Go implementations. Both of these implementations include Unknown as a retryable grpc code. 10 minutes was chosen as the timeout to align with the Java implementation. This should be matched with a more aggressive backoff policy.

Go also has a special check to avoid retrying Internal errors that are invalid UTF-8. Since strings are valid UTF-8, this problem should not need to be handled in our Rust client library.

For #3019

### DIFF
--- a/src/pubsub/src/publisher/retry_policy.rs
+++ b/src/pubsub/src/publisher/retry_policy.rs
@@ -78,11 +78,8 @@ impl RetryPolicy for RetryableErrors {
         // - 429: Resource Exhausted
         // - 499: Cancelled Request
         // - 5xx: Internal Server Error, Bad Gateway, etc.
-        if let Some(code) = error.http_status_code() {
-            match code {
-                408 | 429 | 499 | 500..600 => return RetryResult::Continue(error),
-                _ => (),
-            }
+        if let Some(408 | 429 | 499 | 500..600) = error.http_status_code() {
+            return RetryResult::Continue(error);
         }
 
         if let Some(status) = error.status() {


src/pubsub/src/publisher/retry_policy.rs
```diff
@@ -0,0 +1,219 @@
...
+        if let Some(status) = error.status() {
+            use google_cloud_gax::error::rpc::Code;
+            return match status.code {
+                Code::Aborted
+                | Code::Cancelled
+                | Code::DeadlineExceeded
+                | Code::Internal
+                | Code::ResourceExhausted
+                | Code::Unavailable
+                | Code::Unknown => RetryResult::Continue(error),
```
coryan, line 94: This is strangely similar to the storage policy. Maybe we should refactor them?

The case of `Unknown`: is that to handle the `GoAway` errors?  If so, I think we should convert those to `Error::transport()` when converting from `tonic::Status` to `gax::error::Error`. That is a generic problem.

suzmue: Including Unknown here is mostly to be aligned with the Java and Go clients, which both retry all Unknown errors.

I am actually currently running my test without Unknown handling because I thought the cases like GoAway errors would have been included in one of the previous checks so wanted to see if there are any other Unknown errors that come up.

coryan (approved):

PR #4717, opened 2026-02-18
refactor(gax-internal): more reusable host module

coryan:
I will need to use the functions in this module in a context where I must return `gax::error::Error`. A custom time with some helpers is easier to reuse. Took the opportunity to also remove some complexity.

Motivated by #3178

### DIFF
--- a/src/gax-internal/src/host.rs
+++ b/src/gax-internal/src/host.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use google_cloud_gax::client_builder::Error as BuilderError;
+#[cfg(test)]
 use google_cloud_gax::error::Error;
 use http::Uri;
 use std::str::FromStr;
@@ -21,6 +22,7 @@ use std::str::FromStr;
 ///
 /// Notably, locational and regional endpoints are detected and used as the
 /// host. For VIPs and private networks, we need to use the default host.
+#[cfg(any(test, feature = "_internal-http-client"))]
 pub(crate) fn header(endpoint: Option<&str>, default_endpoint: &str) -> Result<String, HostError> {
     origin_and_header(endpoint, default_endpoint).map(|(_, header)| header)
 }
@@ -31,6 +33,7 @@ pub(crate) fn header(endpoint: Option<&str>, default_endpoint: &str) -> Result<S
 /// host. For VIPs and private networks, we need to use the default host.
 ///
 /// Tonic consumes the authority as a [http::Uri].
+#[cfg(any(test, feature = "_internal-grpc-client"))]
 pub(crate) fn origin(endpoint: Option<&str>, default_endpoint: &str) -> Result<Uri, HostError> {
     origin_and_header(endpoint, default_endpoint).map(|(origin, _)| origin)
 }
@@ -96,7 +99,7 @@ impl HostError {
         }
     }
 
-    #[allow(dead_code)]
+    #[cfg(test)]
     pub fn gax(self) -> Error {
         match self {
             Self::Uri(e) => Error::io(e),


src/gax-internal/src/host.rs
```diff
@@ -12,116 +12,225 @@
...
+use google_cloud_gax::error::Error;
 use http::Uri;
 use std::str::FromStr;
 
-/// Calculate the host based on the configured endpoint and default endpoint.
+/// Calculate the host header given the endpoint and default endpoint.
 ///
 /// Notably, locational and regional endpoints are detected and used as the
 /// host. For VIPs and private networks, we need to use the default host.
+pub(crate) fn header(endpoint: Option<&str>, default_endpoint: &str) -> Result<String, HostError> {
```
dbolduc, line 26:
```sh
$ cargo clippy -p google-cloud-gax-internal --features _internal-grpc-client

warning: function `header` is never used
  --> src/gax-internal/src/host.rs:24:15
   |
24 | pub(crate) fn header(endpoint: Option<&str>, default_endpoint: &str) -> Result<String, HostError> {
   |               ^^^^^^
   |
   = note: `#[warn(dead_code)]` (part of `#[warn(unused)]`) on by default

warning: `google-cloud-gax-internal` (lib) generated 1 warning
```

do you care?

src/gax-internal/src/host.rs
```diff
@@ -12,116 +12,225 @@
...
+/// Notably, locational and regional endpoints are detected and used as the
+/// host. For VIPs and private networks, we need to use the default host.
 ///
-/// We could have returned a boolean that says whether to use the default
-/// endpoint vs. the custom endpoint. Note, however, that:
-///   (1) this function already computes the origin and host.
-///   (2) consumers of this function would have to recompute the values.
-pub(crate) fn from_endpoint<T>(
+/// Tonic consumes the authority as a [http::Uri].
+pub(crate) fn origin(endpoint: Option<&str>, default_endpoint: &str) -> Result<Uri, HostError> {
```
dbolduc, line 37: same but for http

src/gax-internal/src/host.rs
```diff
@@ -12,116 +12,225 @@
...
+    match &parts[..] {
+        // This is a regional endpoint. It should be used as the host.
+        // `{service}.{region}.rep.googleapis.com`
+        [s, _, "rep"] if *s == service => Ok((custom_origin, custom_host)),
+        // This is a locational endpoint. It should be used as the host.
+        // `{region}-{service}.googleapis.com`
+        [location]
+            if location
+                .strip_suffix(service)
+                .and_then(|s| s.strip_suffix("-"))
```
dbolduc, line 77: comment: ah, looks odd to do this in two parts, but saves an allocation.

src/gax-internal/src/host.rs
```diff
@@ -12,116 +12,225 @@
...
+
+impl HostError {
+    pub fn client_builder(self) -> BuilderError {
+        match self {
+            Self::Uri(e) => BuilderError::transport(e),
+            Self::MissingAuthority(e) => BuilderError::transport(Self::error_message(e)),
+        }
+    }
+
+    #[allow(dead_code)]
```
dbolduc: nit: it might be slightly nicer to use `#[cfg(test)]` for these interim implementations. When we actually use these things in production code, it will force us to remove the attribute. Whereas with `#[allow(dead_code)]`, nothing prompts us to remove the attribute; we have to remember.

src/gax-internal/src/host.rs
coryan: Fixed.

src/gax-internal/src/host.rs
coryan: Fixed.

src/gax-internal/src/host.rs
coryan: yup.

src/gax-internal/src/host.rs
coryan: Good idea. Done.

PR #4718, opened 2026-02-18
cleanup(showcase): simpler tracing code

coryan:
Use a newer version of `librarian` to simplify the generated code. There will be more of these, but this change allows us to remove some functions.

Motivated by #3178

### DIFF
--- a/tests/o11y/src/e2e/showcase.rs
+++ b/tests/o11y/src/e2e/showcase.rs
@@ -87,7 +87,7 @@ pub async fn run() -> anyhow::Result<()> {
         .collect::<BTreeSet<_>>();
     let required = BTreeSet::from_iter([
         ROOT_SPAN_NAME,
-        "google-cloud-showcase-v1beta1::client::Echo::echo",
+        "google_cloud_showcase_v1beta1::client::Echo::echo",
     ]);
     let missing = required.difference(&span_names).collect::<Vec<_>>();
     assert!(missing.is_empty(), "missing={missing:?}\n\n{trace:?}",);


coryan:
PTAL, I needed to fix one more test.

librarian.yaml
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 language: rust
-version: v0.8.4-0.20260217201626-1015eadde3c6
+version: v0.8.4-0.20260218181222-507378f20682
```
miguelvelezsa, line 16: would you mind getting latest version again? right now is v0.8.4-0.20260218182016-be3e0d8993f7 I need a version with this change https://github.com/googleapis/librarian/pull/4088, using versions without it are making integration tests to fail.

miguelvelezsa: probably better to just have a PR changing the librarian version?

coryan: Yes, I would mind.  That would mix two changes with different motivations: one to get new generated code (which needs matching changes to the tests), and one to update the format of `librarian.yaml`.

We should not be mixing changes if we can avoid them.

coryan: > probably better to just have a PR changing the librarian version?

You can't.  The CI requires that the generated code matches the librarian version and requires that `librarian tidy` leaves the librarian.yaml file unchanged.

coryan: > > probably better to just have a PR changing the librarian version?
> 
> You can't. The CI requires that the generated code matches the librarian version and requires that `librarian tidy` leaves the librarian.yaml file unchanged.

And I should add: we want those properties. That is how we prevent drift from the generated code and the generator, and how we can go back to an old release branch and generate the code again if needed.

Issue #4018, opened 2025-12-07
Generate more setter and RPC samples for Pub/Sub once the surface is more stable.


amanda-tarafa:
We are skipping setter sample generation for `src/pubsub/src/generated/gapic_dataplane/.sidekick.toml`. We should look at that once the surface of Pub/Sub is closer to stable. At the moment, the problem is that most types in `gapic_dataplane` are internal, and some generated setter samples reference those types that cargotest cannot see.

amanda-tarafa:
Note we also need to enable RPC sample generation for Pub/Sub.

PR #4720, opened 2026-02-18
docs: Generate RPC samples for all APIs

amanda-tarafa:


amanda-tarafa:
The `gcb-pr-showcase` check is green, and it's really running the following code (i.e. testing all doc samples):

```
set -ev

# We use `--all-features` which triggers the Tonic+Prost code generation.
echo "==== Install protoc ===="
curl -fsSL --retry 5 --retry-delay 15 -o /tmp/protoc.zip https://github.com/protocolbuffers/protobuf/releases/download/v28.3/protoc-28.3-linux-x86_64.zip
sha256sum -c <(echo 0ad949f04a6a174da83cdcbdb36dee0a4925272a5b6d83f79a6bf9852076d53f  /tmp/protoc.zip)
env -C /usr/local unzip -x /tmp/protoc.zip
protoc --version

rustup component add clippy
cargo version
rustup show active-toolchain -v

echo "RUSTFLAGS in test-ignored-samples: ${RUSTFLAGS:-}"
echo "RUSTDOCFLAGS in test-ignored-samples: ${RUSTDOCFLAGS:-}"

cargo test --workspace --doc --all-features

echo "==== DONE ===="

/workspace/.bin/sccache --show-stats
```

All other checks are green. I'll clean up the PR and push. It's still possible that the actual showcase check fails at that point, in which case we'll temporarily skip RPC sample generation for showcase.

amanda-tarafa:
This is ready for review. I've created #4729 for tracking the fixes for the few APIs that we are skipping here. And we also have #4018 for Pub/Sub.

See https://github.com/googleapis/google-cloud-rust/pull/4720#issuecomment-3924533374 for how these were tested.

src/generated/showcase/src/client.rs
```diff
@@ -600,89 +1158,311 @@ impl Identity {
...
+    /// ```
+    /// # use google_cloud_showcase_v1beta1::client::Identity;
+    /// use google_cloud_gax::paginator::ItemPaginator as _;
+    /// use google_cloud_showcase_v1beta1::Result;
+    /// async fn sample(
+    ///    client: &Identity
+    /// ) -> Result<()> {
+    ///     let mut list = client.list_users()
+    ///         /* set fields */
+    ///         .by_item();
```
coryan, line 1257: FYI. Several cases in showcase have these empty list of fields for `list_*` operations.

src/generated/cloud/workflows/v1/src/client.rs
```diff
@@ -163,6 +216,21 @@ impl Workflows {
...
     /// [user guide]: https://googleapis.github.io/google-cloud-rust/
     /// [working with long-running operations]: https://googleapis.github.io/google-cloud-rust/working_with_long_running_operations.html
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_workflows_v1::client::Workflows;
+    /// use google_cloud_lro::Poller;
+    /// use google_cloud_workflows_v1::Result;
+    /// async fn sample(
+    ///    client: &Workflows, resource_name: &str
```
coryan, line 226: nit: `resource_name` is not any more helpful than `name` and probably gets people wondering, "what is a resource".

src/generated/cloud/workflows/v1/src/client.rs
```diff
@@ -146,6 +179,26 @@ impl Workflows {
...
+    /// use google_cloud_workflows_v1::Result;
+    /// async fn sample(
+    ///    client: &Workflows, parent: &str
+    /// ) -> Result<()> {
+    ///     let response = client.create_workflow()
+    ///         .set_parent(parent).set_workflow_id("workflow_id_value")
+    ///         .set_workflow(
+    ///             Workflow::new()/* set fields */
+    ///         )
+    ///         .poller().until_done().await?;
```
coryan, line 197: FYI, this is close to ideal...

coryan (approved):
I left a few comments as FYIs, things to consider in the future.

An interesting question: how many RPCs follow the AIPs vs. how many don't?  You could modify `librarian` to log "AIP: YES" and "AIP: NO" and then count the results.

Issue #4013, opened 2025-12-05
Pubsub: Support Publisher message ordering retry


PhongChuong:
See [retry ordering](https://docs.cloud.google.com/pubsub/docs/retry-requests#retry_ordering).

suzmue:
We will keep the same retry policy for ordered messages as unordered messages by default.

PR #4722, opened 2026-02-18
feat(pubsub): retry batches for ordering key messages forever

suzmue:
The Pub/Sub client libraries support ordered publishing for messages with ordering keys. Publishing for an ordering key is stopped when there is a permanent error. Since the cost of failing on trayient libraries will retry transient errors *indefinitely*.

This change overrides the retry policy for batches with ordering keys. There is currently no way to configure this retry policy for the clients, we may wish to add one in the future.

For #4013

### DIFF
--- a/src/pubsub/src/publisher/actor.rs
+++ b/src/pubsub/src/publisher/actor.rs
@@ -1172,7 +1172,7 @@ mod tests {
         mock.expect_publish()
             .times(1)
             .in_sequence(&mut seq)
-            .withf(|_, o| o.retry_policy().is_none ())
+            .withf(|_, o| o.retry_policy().is_none())
             .returning(|r, o| Box::pin(async { publish_ok(r, o) }));
 
         let (actor_tx, actor_rx) = tokio::sync::mpsc::unbounded_channel();


src/pubsub/src/publisher/batch.rs
```diff
@@ -34,7 +35,21 @@ impl Batch {
             initial_size,
             messages_byte_size: initial_size,
             batching_options,
-            ..Batch::default()
+            has_ordering_key: false,
+            messages: Vec::new(),
+        }
+    }
+
+    pub(crate) fn new_with_ordering_key(
```
PhongChuong, line 43: Why not add has_ordering_key to the existing constructor as a parameter instead of a new constructor?

src/pubsub/src/publisher/batch.rs
```diff
@@ -219,6 +240,71 @@ mod tests {
...
+        let mut inflight = JoinSet::new();
+        batch.flush(client, "topic".to_string(), &mut inflight);
+        assert_eq!(batch.len(), 0);
+    }
+
+    #[tokio::test]
+    async fn send_uses_default_retry() {
+        let mut mock = MockGapicPublisher::new();
+        mock.expect_publish()
+            .withf(|r, o| r.topic == "topic" && r.messages.len() == 3 && o.retry_policy().is_some())
```
PhongChuong:
```suggestion
            .withf(|r, o| r.topic == "topic" && r.messages.len() == 3 && o.retry_policy().is_none())
```

This may be the same issue with mockall that we've discussed previously.
Try running `cargo test -- --nocapture`

PhongChuong: I poked around a bit and this is a slightly different issue. The reason is because we didn't await on the inflight and so it never Send.
https://github.com/googleapis/google-cloud-rust/pull/4735 should fix this.

src/pubsub/src/publisher/batch.rs
suzmue: I had that originally but I think it makes the code slightly more readable and a little harder to misuse with a different constructor but don't have strong preference.

src/pubsub/src/publisher/batch.rs
suzmue: TY!

src/pubsub/src/publisher/batch.rs
PhongChuong: Sg. We can keep it for now and if it becomes cumbersome when we need more constructors, we can always fix it as it's internal.

coryan:
"forever" scares me.... does that literally mean that the client library will retry without any stop condition? Regardless of the number of attempts or the time trying? Is that a default? Can applications override it?

suzmue:
> "forever" scares me.... does that literally mean that the client library will retry without any stop condition? Regardless of the number of attempts or the time trying?  Can applications override it?

Yes, forever means forever. 

> Is that a default?

Its the default for messages with ordering keys

> Can applications override it?

Not currently.

This is the guidance for how to implement ordering in the publisher (see [go/ordering-keys-library](http://go/ordering-keys-library)).

If we are scared of actual forever, I think some of the other high level client libraries use a time limit that is Duration::MAX, but I think that is a product of how their retries are implemented, and it is effectively the same.

PR #4725, opened 2026-02-18
fix(pubsub): handle batch actor shutdown gracefully

### LINKED ISSUE #4716: Publisher should not panic when the batch actors close
```
thread 'tokio-runtime-worker' (255914) panicked at src/pubsub/src/publisher/actor.rs:154:34:
batch actors should not close the receiver from the Dispatcher: SendError { .. }
```

This is likely during shutdown when tasks may be killed in arbitrary order. Similar to https://github.com/googleapis/google-cloud-rust/issues/4483

suzmue:
During shutdown, tasks may be aborted in any order. This means that we cannot expect that any particular task is alive at a given time. To handle shutdown of the tasks that handle batching, we shut the publisher down completely. Any dropped receiver handles for messages that were not yet sent will receive a ShutdownError when the sender half is dropped.

Shutting down the Dispatcher completely is the cleanest way to handle this and we only expect this to come up when the whole program is shutting down or there is another serious error and we want to stop publishing to indicate that things are not in a good state to continue publishing.

Fixes #4716

### DIFF
--- a/src/pubsub/src/publisher/actor.rs
+++ b/src/pubsub/src/publisher/actor.rs
@@ -15,7 +15,6 @@
 use super::options::BatchingOptions;
 use crate::generated::gapic_dataplane::client::Publisher as GapicPublisher;
 use crate::publisher::batch::Batch;
-use crate::publisher::constants;
 use std::collections::{HashMap, VecDeque};
 use tokio::sync::{mpsc, oneshot};
 use tokio::task::JoinSet;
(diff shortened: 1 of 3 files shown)


suzmue:
> Out of curiosity, is there any good way to test this? A quick search did not yield any useful result for testing.

Yeah I was thinking about it and I think if we mock or make an alternative implementation for the BatchActors that just gives up that would work. Will look at adding in a follow up.

Issue #4730, opened 2026-02-19
Concurrency bug in striped_downloads.md guide


fellhorn:
## Description

The guide on how to implement striped downloads uses `tokio::fs::File` in a non-thread-safe way by concurrently seeking the file cursor across the different tasks. The downloaded file will usually be messed up as the chunks are written with the wrong cursor position.

## Fix

Instead of using `seek` & `write` one should use `pwrite`, e.g. with the synchronous `write_all_at` from `std::os::unix::fs::FileExt`, wrapped in a `tokio::task::spawn_blocking`.
Asynchronous `write_at` is not available in tokio [yet](https://github.com/tokio-rs/tokio/issues/1529).

## Test

A simple comparison of a file downloaded in a single stripe or 32 stripes usually already shows the issue.
Add this test to [striped.rs](https://github.com/googleapis/google-cloud-rust/blob/8e336e1c781c21206f6757183830e1f128da09e2/guide/samples/tests/storage/striped.rs) and change the bucket:

```rs
#[cfg(test)]
mod tests {
    use super::*;

    #[tokio::test]
    async fn test_striped_download() -> anyhow::Result<()> {
        const MB: usize = 1024 * 1024;
        let bucket_name = "projects/_/buckets/my_bucket";
        let destination_single = "/tmp/striped_download_single.txt";
        let destination_striped = "/tmp/striped_download_striped.txt";

        let client = Storage::builder().build().await?;
        let control = StorageControl::builder().build().await?;

        seed(client.clone(), control.clone(), bucket_name).await?;

        // Download with large stripe size (single stripe)
        println!("\n--- Download with 32 MiB stripe (single stripe) ---");
        download(
            client.clone(),
            control.clone(),
            bucket_name,
            "32MiB.txt",
            32 * MB,
            destination_single,
        )
        .await?;

        // Download with small stripe size (32 stripes)
        println!("\n--- Download with 1 MiB stripe (32 stripes) ---");
        download(
            client.clone(),
            control.clone(),
            bucket_name,
            "32MiB.txt",
            MB,
            destination_striped,
        )
        .await?;

        // Compute and compare hashes
        println!("\n--- Computing hashes ---");
        let hash_single = compute_sha256(destination_single).await?;
        let hash_striped = compute_sha256(destination_striped).await?;

        println!("Hash (single stripe):  {}", hash_single);
        println!("Hash (32 stripes):     {}", hash_striped);

        assert_eq!(
[...]

PR #4731, opened 2026-02-19
fix(doc): Concurrency bug in striped_downloads.md guide

fellhorn:
The guide on how to implement striped downloads uses `tokio::fs::File` in a non-thread-safe way by concurrently seeking the file cursor across the different tasks. The downloaded file will usually be messed up as the chunks are written with the wrong cursor position.

This PR adds a reference implementation using `pwrite` for Unix systems.

Towards #4730

### DIFF
--- a/guide/src/storage/striped_downloads.md
+++ b/guide/src/storage/striped_downloads.md
@@ -122,7 +122,8 @@ futures:
 
 Once they complete, the file is downloaded.
 
-Now you should complete writing the `write_stripe()` function. First, start a download from Cloud Storage:
+Now you should complete writing the `write_stripe()` function. First, start a
+download from Cloud Storage:
 
 ```rust,ignore,noplayground
 {{#rustdoc_include ../../samples/tests/storage/striped.rs:write-stripe-reader}}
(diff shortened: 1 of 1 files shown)


fellhorn:
I don't have access to any Windows machines to verify that `seek_write` from `std::os::windows::fs::FileExt` would work and is more or less performant.

coryan:
> I don't have access to any Windows machines to verify that `seek_write` from `std::os::windows::fs::FileExt` would work and is more or less performant.

First, thanks for the bug report and PR. I am quite embarrassed you had to find this problem yourself.

Second, I think we do need the example to work on Windows, even if we must caveat the code with "beware, this may be slower".  Do you want to take a stab at making the Windows changes too?  If you prefer, we can merge the code with a `panic!()` on Windows referencing the bug you filed.

coryan:
Some formatting problems, can you use mdformat on this file:

```

diff --git a/guide/src/storage/striped_downloads.md b/guide/src/storage/striped_downloads.md
--
  | index e8293a1..c471416 100644
  | --- a/guide/src/storage/striped_downloads.md
  | +++ b/guide/src/storage/striped_downloads.md
  | @@ -122,7 +122,8 @@ futures:
  |  
  | Once they complete, the file is downloaded.
  |  
  | -Now you should complete writing the `write_stripe()` function. First, start a download from Cloud Storage:
  | +Now you should complete writing the `write_stripe()` function. First, start a
  | +download from Cloud Storage:
  |  
  | ```rust,ignore,noplayground
  | {{#rustdoc_include ../../samples/tests/storage/striped.rs:write-stripe-reader}}
  | @@ -143,10 +144,11 @@ get inconsistent reads:
  | {{#rustdoc_include ../../samples/tests/storage/striped.rs:write-stripe-reader-generation}}
  | ```
  |  
  | -Then you read the data and write it to the local file. This example uses the Unix
  | -specific `write_all_at` to write the chunk with a specific cursor offset in a thread-safe way.
  | -On Windows use `seek_write` from `std::os::windows::fs::FileExt` instead.
  | -To not block within the future, we run the closure on a dedicated thread:
  | +Then you read the data and write it to the local file. This example uses the
  | +Unix specific `write_all_at` to write the chunk with a specific cursor offset in
  | +a thread-safe way. On Windows use `seek_write` from
  | +`std::os::windows::fs::FileExt` instead. To not block within the future, we run
  | +the closure on a dedicated thread:


```

coryan (approved):
I will make the fixes for Windows in a second pass.

coryan:
Thanks again for the PR! I made some formatting changes and will fix the code on Windows shortly.

fellhorn:
Sorry, I was busy with other things yesterday. Thanks for taking over the Windows part!

PR #4732, opened 2026-02-19
fix(pubsub): return  error when publishing messages that are too large

### LINKED ISSUE #4650: Pubsub should return error when Publish message size is larger than batch threshold

PhongChuong:
Publisher returns ExceededByteThresholdError when the application publish a message that is larger (with topic and ordering key) than the configured batch threshold.

For messages without ordering key, the Publisher returns an ExceededByteThresholdError for that message and continue as normal.
For messages with an ordering key, the Publisher returns an ExceededByteThresholdError for that message and pause publishing for the ordering key.

Fixes #4650

### DIFF
--- a/src/pubsub/src/publisher/batch.rs
+++ b/src/pubsub/src/publisher/batch.rs
@@ -23,7 +23,7 @@ use std::sync::Arc;
 #[derive(Debug, Default)]
 pub(crate) struct Batch {
     messages: Vec<BundledMessage>,
-    initial_size: u32, // TODO(NOW): Change this to topic instead.
+    initial_size: u32,
     messages_byte_size: u32,
     batching_options: BatchingOptions,
 }


src/pubsub/src/publisher/batch.rs
```diff
@@ -23,7 +23,7 @@ use std::sync::Arc;
 #[derive(Debug, Default)]
 pub(crate) struct Batch {
     messages: Vec<BundledMessage>,
-    initial_size: u32,
+    initial_size: u32, // TODO(NOW): Change this to topic instead.
```
dbolduc: ?

src/pubsub/src/publisher/implementation.rs
```diff
@@ -313,6 +313,67 @@ mod tests {
...
+        let key = "key";
+        let byte_threshold = TOPIC.len() + "hello".len() + key.len() + 1;
+        let publisher = PublisherPartialBuilder::new(client, TOPIC.to_string())
+            .set_byte_threshold(byte_threshold as u32)
+            .build();
+
+        // Validate without ordering key.
+        let handle = publisher.publish(Message::new().set_data("hello"));
+        assert_eq!(handle.await?, "hello");
+        let handle = publisher.publish(Message::new().set_data("too loooooooong"));
```
dbolduc, line 334: lol

src/pubsub/src/publisher/batch.rs
PhongChuong: Removed. Sorry I sometimes leave TODOs for revisit for the PR. I forgot to remove this one.

Issue #4733, opened 2026-02-19
Tokio test does not assert mockall expectations


PhongChuong:
When running tests that spawns tasks with #[tokio::test], if the task panics such as from unmet mockall expectations, the tests do not fail as the panic is not propagated back to the test result. This can be resolved by running the test with:
`RUSTFLAGS="--cfg tokio_unstable" cargo test`
and
unhandled_panic = "shutdown_runtime" in the tests.

PR #4734, opened 2026-02-19
test(pubsub): run publisher mockall tests with tokio unstable

PhongChuong:
For #4733

### DIFF
--- a/src/pubsub/src/publisher/actor.rs
+++ b/src/pubsub/src/publisher/actor.rs
@@ -700,7 +700,15 @@ mod tests {
         Ok(())
     }
 
-    #[tokio::test(start_paused = true)]
+    #[cfg_attr(
+        tokio_unstable,
+        tokio::test(
+            start_paused = true,
+            flavor = "current_thread",
+            unhandled_panic = "shutdown_runtime"
+        )
+    )]
+    #[cfg_attr(not(tokio_unstable), tokio::test(start_paused = true))]
     async fn concurrent_actor_publish() -> anyhow::Result<()> {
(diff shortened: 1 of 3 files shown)


PhongChuong:
@dbolduc , using this instead of #[tokio::test] in the subscriber tests should detect the failed expectations. You would need to run the tests with: `RUSTFLAGS="--cfg tokio_unstable` or wait for the tokio-unstable CI tests.

src/pubsub/src/publisher/actor.rs
```diff
@@ -821,7 +845,15 @@ mod tests {
         Ok(())
     }
 
-    #[tokio::test(start_paused = true)]
+    #[cfg_attr(
+        tokio_unstable,
+        tokio::test(
+            start_paused = true,
+            flavor = "current_thread",
```
dbolduc, line 852: Q: is it necessary to specify the `flavor`? I would have guessed it defaults to the value we want.

dbolduc:
> @dbolduc , using this instead of #[tokio::test] in the subscriber tests should detect the failed expectations. You would need to run the tests with: `RUSTFLAGS="--cfg tokio_unstable` or wait for the tokio-unstable CI tests.

Ack, and thanks. I plan to copy your work for the relevant subscriber tests.

src/pubsub/src/publisher/actor.rs
PhongChuong: From [this](https://docs.rs/tokio/latest/tokio/attr.test.html#configure-unhandled-panic-behavior):
`This option is only compatible with the current_thread runtime.`

So I think it might be better to be explicit here.

PR #4346, opened 2026-01-23
impl(auth): trust boundaries module

alvarowolfx:
Towards #4186

PR #4736, opened 2026-02-19
impl(auth): add access boundary credential decorator

alvarowolfx:
Towards #4186


alvarowolfx:
PR #4346 is up to date with how the decorator is going to be used

PR #4739, opened 2026-02-19
chore: update googleapis and discovery-artifact-manager, and regenerate

miguelvelezsa:

### DIFF
--- a/guide/samples/tests/storage/striped.rs
+++ b/guide/samples/tests/storage/striped.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 // ANCHOR: all
+use std::sync::Arc;
 
 use google_cloud_gax::options::RequestOptionsBuilder;
 // ANCHOR: seed-function
@@ -128,7 +129,8 @@ async fn download(
     // ANCHOR_END: get-metadata
 
     // ANCHOR: create-destination
-    let file = tokio::fs::File::create(destination).await?;
+    let file = Arc::new(std::fs::File::create(destination)?);
+    file.set_len(metadata.size as u64)?;
     // ANCHOR_END: create-destination
     let start = std::time::Instant::now();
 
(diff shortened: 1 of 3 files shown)


src/generated/showcase/src/tracing.rs
```diff
@@ -37,576 +37,480 @@ impl<T> super::stub::Compliance for Compliance<T>
...
-        self.inner
-            .repeat_data_body(req, options)
-            .instrument(span.clone())
-            .await
-            .record_in_span(&span)
-    }
-
-    #[cfg(not(google_cloud_unstable_tracing))]
-    #[tracing::instrument(ret)]
+    #[tracing::instrument(level = tracing::Level::DEBUG, ret)]
```
coryan, line 40: Seems like you are using a version of `librarian` that is different from the version configured in the `librarian.yaml` file.  Please regenerate the code with that version.

Separately,   #4762 will update librarian and these files.

miguelvelezsa: Yes. I am trying to use the new librarianops upgrade command to update this repo to the latest version. Any concern?

what is #4762?

coryan: Do'h #4726

coryan: > I am trying to use the new librarianops upgrade command to update this repo to the latest version. Any concern?

I don't know how librarianops works, but in this repo we should:

1. Only upgrade librarian or update the googleapis / discovery-blah-blah SHAs in a PR, not both [^1].
2. When updating the SHAs you should use the version of librarian configured in the librarian.yaml file: this PR is obviously using a different version of librarian.
3. When upgrading librarian you should run `generate --all` **with the new version**.


[^1]: the original title of this PR suggested otherwise, you fixed the title and the PR only upgrades the SHAs. All good.

miguelvelezsa: Yep, if that is the case I can use the command (after I fix it :-P cause I am seeing some issues), the command does 1- Get the latest version of librarian, 2- Update librarian.yaml, 3- Run librarian generate (https://github.com/googleapis/librarian/issues/3911).

coryan: Consider rebasing. That may fix the regenerate problems.

PR #4747, opened 2026-02-20
feat(gax-internal): cleanup full URL query parameters

coryan:
The OpenTelemetry semantic conventions [require redacting] some query parameter values. I am adding `upload_id` to that list because we should.

[require redacting]: https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/#url-full

Motivated by #3178

### DIFF
--- a/src/gax-internal/src/observability/http_tracing.rs
+++ b/src/gax-internal/src/observability/http_tracing.rs
@@ -27,7 +27,7 @@ use tracing::{Span, field};
 lazy_static::lazy_static! {
     pub(crate) static ref REDACTED_QUERY_PARAMETERS: HashSet<&'static str> = {
         HashSet::from_iter([
-            // Required by OpenTelementr semantic conventions:
+            // Required by OpenTelemetry semantic conventions:
             //     https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/#url-full
             "AWSAccessKeyId", "Signature", "sig", "X-Goog-Signature",
             // Google uses this as a key in resumable uploads.


src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -21,8 +21,21 @@ use google_cloud_gax::options::{
     internal::{get_path_template, get_resource_name},
 };
 use opentelemetry_semantic_conventions::{attribute as otel_attr, trace as otel_trace};
+use std::collections::HashSet;
 use tracing::{Span, field};
 
+lazy_static::lazy_static! {
+    pub(crate) static ref REDACTED_QUERY_PARAMETERS: HashSet<&'static str> = {
+        HashSet::from_iter([
+            // Required by OpenTelementr semantic conventions:
```
dbolduc: s/OpenTelementr/OpenTelemetry/

src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -21,8 +21,21 @@ use google_cloud_gax::options::{
...
+            //     https://opentelemetry.io/docs/specs/semconv/registry/attributes/url/#url-full
+            "AWSAccessKeyId", "Signature", "sig", "X-Goog-Signature",
+            // Google uses this as a key in resumable uploads.
+            "upload_id",
```
dbolduc, line 34: err... Can we configure this per-client instead of having a global list?

src/gax-internal/src/observability/http_tracing.rs
coryan: Fixed.

src/gax-internal/src/observability/http_tracing.rs
coryan: Maybe, I do not think it is worth it.  Consider: (a) if there was an `upload_id` field in any client when used as a query parameter it would be called `uploadId`, we use JSON encoding for query parameters, and (b) there are no `upload_id` field anywhere else in the GAPICs.

westarle: I hope this doesn't come up enough to need specific filters for specific clients.  If it happens, we likely need to deal with it in some way that isn't a low-level URL attribute filter.

westarle (approved):
This LGTM.  A couple (future) concerns:

1. The "entire" upload URL is defined as a credential
2. This fixes `url.full` but isn't a systemic change that prevents the url from being logged

I think we're good for tracing attributes, but we should look carefully at this before including it in logs & traces that include error information that comes from `reqwest` directly.

PR #4754, opened 2026-02-20
impl(auth): add access boundary support for MDS

alvarowolfx:
Towards #4186


alvarowolfx:
> LGTM. You need to rebase.

@coryan  was in the process of doing it. Now it's up to date.

PR #4755, opened 2026-02-21
feat(gax-internal): builder for raw HTTP requests

coryan:
Most client libraries use `ReqwestClient` to make HTTP+JSON RPCs. In the Storage client (and maybe in others) we need to make raw HTTP requests.

These are different in seveal dimensions:

- The retry loop is typically implemented outside the `ReqwestClient`, so calling `execute()` is a problem.
- The URL may target an endpoint that is different from the `ReqwestClient`'s endpoint, so the host header needs adjustment.
- The request or response may be streamed instead of parsed from/to a JSON value.

It was getting too hard to keep these things in mind, adding a new type makes it harder to use these incorrectly.

I also refactored some of the tests to remove duplicated code.

Towards #3178

### DIFF
--- a/src/gax-internal/src/http/http_request_builder.rs
+++ b/src/gax-internal/src/http/http_request_builder.rs
@@ -33,6 +33,7 @@ use crate::attempt_info::AttemptInfo;
 /// # use google_cloud_gax_internal::http::ReqwestClient;
 /// use google_cloud_gax_internal::http::reqwest::Method;
 /// use google_cloud_gax::options::RequestOptions;
+/// use google_cloud_gax_internal::attempt_info::AttemptInfo;
 /// async fn sample(client: &ReqwestClient, options: RequestOptions) -> anyhow::Result<()> {
 ///     let builder = client.http_builder_with_url(
 ///         Method::GET,
@@ -41,7 +42,7 @@ use crate::attempt_info::AttemptInfo;
 ///     )?
 ///     .query("alt", "media")
 ///     .header("x-goog-api-client", "client/1.2.3");
-///     let response = builder.send(options, None, 0).await?;
+///     let response = builder.send(options, AttemptInfo::new(0)).await?;
 ///     println!("status={:?}", response.status());
 ///     Ok(())
 /// }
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/http/http_request_builder.rs
```diff
@@ -0,0 +1,207 @@
...
+
+    /// Adds the body to the request.
+    pub fn body<B>(self, body: B) -> Self
+    where
+        B: Into<Body>,
+    {
+        Self::new(self.client, self.builder.body(body))
+    }
+
+    /// The storage client tests verify the right headers and query parameters are set.
```
PhongChuong, line 122: Is it requirement to set [body, query] for all clients? If so, maybe using a typestate pattern to force compile time check that the parameters are set might be useful.

src/gax-internal/src/http/http_request_builder.rs
```diff
@@ -0,0 +1,207 @@
...
+    ///     }
+    ///     println!("exhausted: {err:?}");
+    ///     Ok(())
+    /// }
+    /// ```
+    pub async fn send(
+        self,
+        options: RequestOptions,
+        remaining_time: Option<Duration>,
+        attempt_count: u32,
```
dbolduc: Maybe this should be a struct so it can grow.

I know we can always plumb things via "internal" fields in the `RequestOptions`.... but I don't like that

src/gax-internal/src/http/http_request_builder.rs
coryan: I briefly prototype that idea.... intriguing, but seems like overkill for an internal-only type.

Also, while `GET` cannot have a body, it is optional for `PUT` and `POST`....  And all the other fields are optional in practice.

coryan:
@dbolduc do you want to be the second (or third) pair of eyes on this?

src/gax-internal/src/http/http_request_builder.rs
```diff
@@ -0,0 +1,207 @@
...
+    /// Adds the body to the request.
+    pub fn body<B>(self, body: B) -> Self
+    where
+        B: Into<Body>,
+    {
+        Self::new(self.client, self.builder.body(body))
+    }
+
+    /// The storage client tests verify the right headers and query parameters are set.
+    pub async fn build_for_tests(self) -> Result<Request> {
```
dbolduc, line 123: I assume this is `pub` because storage will use it?

coryan: Correct, I could not figure out another way to keep the storage tests as they were. We would need to create `httptest` servers otherwise.

src/gax-internal/src/http/http_request_builder.rs
coryan: Done.

PR #4764, opened 2026-02-22
feat(integration-tests-o11y): add resource detector

coryan:
This will just make any demos a little nicer. And we can incubate the resource detector here before contributing it to a different project.

Motivated by #3178 but not required.

### DIFF
--- a/tests/o11y/src/detector.rs
+++ b/tests/o11y/src/detector.rs
@@ -132,7 +132,7 @@ impl GoogleCloudResourceDetectorBuilder {
             builder = builder.with_attribute(KeyValue::new("k8s.cluster.name", v));
         }
 
-        Ok(Self::attributes(
+        Ok(Self::attributes_from_env(
             builder,
             &[
                 ("POD_NAME", "k8s.pod.name"),
@@ -177,7 +177,7 @@ impl GoogleCloudResourceDetectorBuilder {
     }
 
     fn gae_resource(&self, builder: ResourceBuilder) -> ResourceBuilder {
-        Self::attributes(
+        Self::attributes_from_env(
             builder,
             &[
                 ("GAE_SERVICE", "gae.service"),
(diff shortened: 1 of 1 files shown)


tests/o11y/src/detector.rs
```diff
@@ -0,0 +1,690 @@
...
+            builder,
+            &[
+                ("K_SERVICE", "gcr.service"),
+                ("K_REVISION", "gcr.revision"),
+                ("K_CONFIGURATION", "gcr.configuration"),
+            ],
+        )
+    }
+
+    fn attributes(
```
alvarowolfx: nit: maybe make this more explicit that is going to derive attribute from env. I was a bit confused when seeing it being used in the gae_resource and gcr_resource. `Self:atrributes_from_env` would be more direct.

tests/o11y/src/detector.rs
```diff
@@ -0,0 +1,690 @@
...
+        list.iter().fold(builder, |builder, (var, name)| {
+            if let Ok(value) = std::env::var(var) {
+                builder.with_attribute(KeyValue::new(*name, value))
+            } else {
+                builder
+            }
+        })
+    }
+
+    async fn fetch_instance_metadata(&self) -> Result<String, Error> {
```
alvarowolfx, line 214: why not use  gax internal retry_loop and our existing retry policies ?

coryan: That I can answer ... there is a good chance we will add this to an external crate, like https://github.com/open-telemetry/opentelemetry-rust-contrib/tree/main/opentelemetry-resource-detectors

alvarowolfx: oh, make sense

tests/o11y/src/detector.rs
coryan: Done.

PR #4777, opened 2026-02-23
impl(auth): add access boundary support for external accounts

alvarowolfx:
Towards #4186


alvarowolfx:
After #4790, we can test in our 3P infrastructure:

Code:
```rust
use google_cloud_auth::credentials::{Builder, CacheableResource};
use httptest::http::Extensions;
use scoped_env::ScopedEnv;

#[tokio::main]
async fn main() -> anyhow::Result<()> {
    // Enable trust boundaries
    let _env = ScopedEnv::set("GOOGLE_AUTH_ENABLE_TRUST_BOUNDARIES", "true");

    println!("Building Credentials...");
    // Will use ADC with the environment variable set
    let creds = Builder::default().build()?;
    println!("Credentials built: {:?}", creds);

    for _ in 0..10 {
        let cached_headers = creds.headers(Extensions::new()).await?;
        let headers = match cached_headers {
            CacheableResource::New { data, .. } => data,
            CacheableResource::NotModified => unreachable!("should always get new headers"),
        };
        let token = headers.get("Authorization");
        println!("Token: {:?}", token);
        let locations = headers.get("x-goog-allowed-locations");
        println!("Locations: {:?}", locations);
        if locations.is_some() {
            println!("Locations found, breaking");
            break;
        }
        tokio::time::sleep(tokio::time::Duration::from_secs(5)).await;
    }
 }
 ```


output:
```
[REDACTEDrust]$ export GOOGLE_APPLICATION_CREDENTIALS=/home/ec2-user/aws-credentials.json
[REDACTED rust]$ ./trust_boundaries 
Building Credentials...
Credentials built: Credentials { inner: AccessTokenCredentials { inner: CredentialsWithAccessBoundary { credentials: ExternalAccountCredentials { token_provider: TokenCache { rx_token: Receiver { shared: Shared { value: RwLock(PhantomData<std::sync::poison::rwlock::RwLock<core::option::Option<core::result::Result<(google_cloud_auth::token::Token, google_cloud_auth::credentials::EntityTag), google_cloud_gax::error::credentials::CredentialsError>>>>, RwLock { data: None }), version: Version(0), is_closed: false, ref_count_rx: 1 }, version: Version(0) } }, quota_project_id: None }, access_boundary: AccessBoundary { rx_header: Receiver { shared: Shared { value: RwLock(PhantomData<std::sync::poison::rwlock::RwLock<core::option::Option<google_cloud_auth::access_boundary::BoundaryValue>>>, RwLock { data: None }), version: Version(0), is_closed: false, ref_count_rx: 1 }, version: Version(0) } } } } }
Token: Some(Sensitive)
Locations: None
Token: Some(Sensitive)
Locations: Some("0x80000000000")
Locations found, breaking
```

File: src/auth/src/access_boundary.rs [hunk 1]
```diff
@@ -337,6 +341,44 @@ pub(crate) fn service_account_lookup_url(
     format!("{iam_endpoint}/v1/projects/-/serviceAccounts/{email}/allowedLocations")
 }
 
+static WORKLOAD_PATTERN: OnceLock<Regex> = OnceLock::new();
+static WORKFORCE_PATTERN: OnceLock<Regex> = OnceLock::new();
```
coryan: Can we do this parsing without regular expressions?  These are a source of denial of service attacks, and ~I think the `(?P` patterns are particularly prone to quadratic searches~.  That was wrong, `(?P` is fine, but the point stands I think


Could a recursive descent parser do the job?

alvarowolfx: removed the regex crate usage with some of the tricks that you pointed out on the other PR to do a split and match on the given format.

coryan (approved):

PR #4778, opened 2026-02-23
cleanup(pubsub): move PublishFuture to publisher module

suzmue:


src/pubsub/src/publisher.rs
```diff
@@ -24,6 +24,8 @@ pub(crate) mod model_ext;
 pub(crate) mod options;
 pub(crate) mod retry_policy;
 
+pub use model_ext::PublishFuture;
```
coryan, line 27: Seems sad that this lives in a `model_ext` module, maybe it should be in a `publish_future` or `future` module?

suzmue: True, will update in the other PR that is having the multi-approvals issues and will just merge this.

Issue #4772, opened 2026-02-23
Stabilize the APIs for observability


coryan:
We used `--cfg google_cloud_unstable_tracing` to explore the APIs for observability. It is time to decide which APIs get to stay, and which change.  So far I have identified:

PR #4781, opened 2026-02-23
cleanup(showcase): use `RequestOptionsExt`

coryan:
Part of the work for https://github.com/googleapis/google-cloud-rust/issues/4772

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 language: rust
-version: v0.8.4-0.20260219173508-c7c499a09294
+version: v0.8.4-0.20260224172937-39ac98873227
 sources:
   conformance:
     commit: b407e8416e3893036aee5af9a12bd9b6a0e2b2e6


src/generated/showcase/src/transport.rs
```diff
@@ -952,17 +976,15 @@ impl super::stub::Compliance for Compliance {
...
-            if let Some(rn) = resource_name {
-                let full_resource_name = format!("//localhost:7469/{}", rn);
-                google_cloud_gax::options::internal::set_resource_name(options, full_resource_name)
-            } else {
-                options
-            }
+            use google_cloud_gax::options::internal::ResourceName;
+            use google_cloud_gax::options::internal::{PathTemplate, RequestOptionsExt};
+            let options = Option::<&String>::None
+                .or(Some(&req.resource))
```
PhongChuong, line 983: I know this is generated but isn't `Option::<&String>::None.or(Some(&req.resource))` just `Some(&req.resource)`?

dbolduc: yes, but there is not always 1 candidate resource. This approach works when there are N candidate resources that might not have presence.

coryan: yes, and I would be happy to approve a PR that optimized this.  We use the trick in other places too, and it is annoying.  I don't know how to easily write "do this for the first element and this other thing for the other elements" in go + mustache templates without a lot of boilerplate.

PR #4782, opened 2026-02-23
cleanup(pubsub): update PublishError variants

suzmue:
This change rename PublishError::SendError to PublishError::Rpc and PublishError::ShutdownError to PublishError::Shutdown and cleans up the documentation. The types are also improved to remove the unit type.

This change does not touch the ExceededByteThresholdError since that error needs to be removed.

### DIFF
--- a/src/pubsub/src/error.rs
+++ b/src/pubsub/src/error.rs
@@ -20,8 +20,8 @@
 pub enum PublishError {
     /// The underlying RPC call failed.
     ///
-    /// The inner error is wrapped in an [`Arc`](std::sync::Arc) to allow this error to be cloned
-    /// and returned for each message in the batch.
+    /// The inner error is wrapped in an [`Arc`](std::sync::Arc) because the same error
+    /// may affect multiple [`publish()`](crate::client::Publisher::publish) calls.
     #[error("the publish operation was interrupted by an error: {0}")]
     Rpc(#[source] std::sync::Arc<crate::Error>),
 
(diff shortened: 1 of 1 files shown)


src/pubsub/src/error.rs
```diff
@@ -13,30 +13,34 @@
...
 
 /// Represents an error that can occur when publishing a message.
 #[derive(thiserror::Error, Debug)]
 #[non_exhaustive]
 pub enum PublishError {
-    /// Publish operation failed sending the RPC.
+    /// The underlying RPC call failed.
+    ///
+    /// The inner error is wrapped in an [`Arc`](std::sync::Arc) to allow this error to be cloned
+    /// and returned for each message in the batch.
```
coryan: Maybe write this from the perspective of the user:


```suggestion
    /// The inner error is wrapped in an [`Arc`](std::sync::Arc) because the same error
    /// may affect multiple `publish()` calls.
```

src/pubsub/src/error.rs
```diff
@@ -13,30 +13,34 @@
...
+    OrderingKeyPaused,
 
-    /// The publisher is shutdown and is no longer running.
-    #[error("publisher is shutdown")]
-    ShutdownError(()),
+    /// The operation failed because the [`Publisher`](crate::client::Publisher) has
+    /// been shut down.
+    ///
+    /// This may occur when the runtime has dropped the background tasks that handle
+    /// message publishing.
```
coryan: Maybe:


```suggestion
    /// Typically this can happen when the application is shutting down. Some background
    /// tasks in the client library may be terminated before they can send all the
    /// pending messages.
```

suzmue: TY, these were great suggestions.

src/pubsub/src/error.rs
suzmue: Done, thanks for the great suggestion.

PR #4783, opened 2026-02-23
feat(gax): retry policy decorator for 429 errors

coryan:
Someday we will document how to use the `RetryPolicy` trait in applications and add your own policies. This may become duplicative (but still convenient) at that time.

Likewise, we may add a generic decorator that ignores specific HTTP status codes or gRPC status codes. But the API for that is unclear at the moment. If that happens this will still be convenient, but would duplicate some code.

At the moment, this seems like useful thing for customers to have.  The most common case would be `Aip194Strict.continue_on_too_many_requests().with_...`. That is a reasonable policy for services that use `ResourceExhausted` to signal that a rate-based quota limit has been breached.

### DIFF
--- a/src/gax/src/retry_policy.rs
+++ b/src/gax/src/retry_policy.rs
@@ -196,17 +196,17 @@ pub trait RetryPolicyExt: RetryPolicy + Sized {
     /// status code "429 - TOO_MANY_REQUESTS" **or** where the service returns
     /// an error with code [ResourceExhausted].
     ///
-    /// For other errors it returns thesame value as the inner policy.
+    /// For other errors it returns the same value as the inner policy.
     ///
     /// Note that [ResourceExhausted] is ambiguous and may cause problems with
-    /// some services. The code is used for both "too many requests" or for
-    /// and for "quota exceeded" problems. If the quota in question is some
-    /// kind of rate limit, then using this policy may be helpful. If the quota
-    /// is not a rate limit, then this retry policy may needlessly send the
-    /// same RPC multiple times.
+    /// some services. The code is used for both "too many requests"  and for
+    /// "quota exceeded" problems. If the quota in question is some kind of rate
+    /// limit, then using this policy may be helpful. If the quota is not a rate
+    /// limit, then this retry policy may needlessly send the same RPC multiple
+    /// times.
     ///
-    /// You **MUST** consult the documentation for the service and RPC in
-    /// question before using this policy.
+    /// You should consult the documentation for the service and RPC in question
+    /// before using this policy.
     ///
     /// # Example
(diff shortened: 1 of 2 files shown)


src/gax/src/retry_policy.rs
```diff
@@ -185,6 +189,42 @@ pub trait RetryPolicyExt: RetryPolicy + Sized {
...
         LimitedAttemptCount::custom(self, maximum_attempts)
     }
+
+    /// Decorate a [RetryPolicy] to continue on certain status codes.
+    ///
+    /// This policy decorates an inner policy and retries any errors with HTTP
+    /// status code "429 - TOO_MANY_REQUESTS" **or** where the service returns
+    /// an error with code [ResourceExhausted].
+    ///
+    /// For other errors it returns thesame value as the inner policy.
```
dbolduc: s/thesame/the same/

src/gax/src/retry_policy.rs
```diff
@@ -185,6 +189,42 @@ pub trait RetryPolicyExt: RetryPolicy + Sized {
...
+    ///
+    /// Note that [ResourceExhausted] is ambiguous and may cause problems with
+    /// some services. The code is used for both "too many requests" or for
```
dbolduc: s/or for//

src/gax/src/retry_policy.rs
```diff
@@ -185,6 +189,42 @@ pub trait RetryPolicyExt: RetryPolicy + Sized {
...
+    /// and for "quota exceeded" problems. If the quota in question is some
+    /// kind of rate limit, then using this policy may be helpful. If the quota
+    /// is not a rate limit, then this retry policy may needlessly send the
+    /// same RPC multiple times.
+    ///
+    /// You **MUST** consult the documentation for the service and RPC in
```
dbolduc: A little aggressive. I would think "should", not **MUST**

src/gax/src/retry_policy/too_many_requests.rs
```diff
@@ -0,0 +1,165 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::{Error, RetryPolicy, RetryResult, RetryState, ThrottleResult};
+use crate::error::rpc::Code;
+use std::time::Duration;
+
+/// A retry policy decorator continues on `ResourceExhausted` or `TOO_MANY_REQUESTS`.
```
dbolduc: s/continues/that continues/

Issue #4595, opened 2026-02-07
Storage `open_object` samples


dbolduc:
These region tags can be written now:

- [x] `storage_open_multiple_objects_ranged_read`
- [x] `storage_open_object_multiple_ranged_read`
- [x] `storage_open_object_read_full_object`
- [x] `storage_open_object_single_ranged_read`

Issue #2856, opened 2025-08-10
Helper functions to collect N bytes from an object download


coryan:
When downloading objects it might be convenient to download N bytes into a buffer.

We used to have a function to download *all* the bytes, but this can be dangerous if the application does not check the size of the object.

Separately: consider if we want a function to collect up to N bytes into a `bytes::Buf` but without making any data copies.

PR #4785, opened 2026-02-24
docs(storage): add sample for storage_open_object_multiple_ranged_read

vsharonlynn:
Add one of the samples requested in #4595.

### DIFF
--- a/src/storage/examples/src/lib.rs
+++ b/src/storage/examples/src/lib.rs
@@ -409,15 +409,7 @@ pub async fn run_object_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
     .await?;
 
     tracing::info!("running open_object_multiple_ranged_read example");
-    let downloaded_file = tempfile::NamedTempFile::new()?;
-    let downloaded_file_prefix = downloaded_file.path().to_str().unwrap();
-    objects::open_object_multiple_ranged_read::sample(
-        &client,
-        &id,
-        object_name,
-        downloaded_file_name,
-    )
-    .await?;
+    objects::open_object_multiple_ranged_read::sample(&client, &id, object_name).await?;
 
     tracing::info!("running open_object_read_full_object example");
     let downloaded_file = tempfile::NamedTempFile::new()?;
--- a/src/storage/examples/src/objects/open_object_multiple_ranged_read.rs
+++ b/src/storage/examples/src/objects/open_object_multiple_ranged_read.rs
@@ -15,41 +15,36 @@
 // [START storage_open_object_multiple_ranged_read]
 use google_cloud_storage::client::Storage;
 use google_cloud_storage::model_ext::ReadRange;
-use tokio::fs::File;
-use tokio::io::AsyncWriteExt; // Required for write_all
 
-pub async fn sample(
-    client: &Storage,
-    bucket: &str,
-    object: &str,
-    destination_file_name: &str,
-) -> anyhow::Result<()> {
+pub async fn sample(client: &Storage, bucket: &str, object: &str) -> anyhow::Result<()> {
     let bucket_name = format!("projects/_/buckets/{bucket}");
 
-    // 1. Open the connection and request the first 5 bytes.
+    // 1. Open the connection and read the first range.
     let (descriptor, mut reader) = client
         .open_object(bucket_name, object)
         .send_and_read(ReadRange::head(5))
         .await?;
 
-    let file_path = format!("{}_part1", destination_file_name);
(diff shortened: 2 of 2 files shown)


src/storage/examples/src/objects/open_object_multiple_ranged_read.rs
```diff
@@ -0,0 +1,61 @@
...
+    bucket: &str,
+    object: &str,
+    destination_file_name: &str,
+) -> anyhow::Result<()> {
+    let bucket_name = format!("projects/_/buckets/{bucket}");
+
+    // 1. Open the connection and request the first byte range. Read 1,024 bytes starting from offset 0.
+    let (descriptor, mut reader1) = client
+        .open_object(bucket_name, object)
+        .send_and_read(ReadRange::segment(0, 1024))
```
joshuatants: How about using ReadRange::head?

vsharonlynn: Updated. Thanks!
I also:
1. Used variable shadowing, to prevent accidental reuse of the first read from the second read and make the code look cleaner.
2. Used ReadRange::tail to read the last bytes to make this code sample simulate reading a header and footer scenario.

joshuatants (approved):
LGTM. Waiting for @coryan for final approval.

src/storage/examples/src/objects/open_object_multiple_ranged_read.rs
```diff
@@ -0,0 +1,57 @@
...
+        .open_object(bucket_name, object)
+        .send_and_read(ReadRange::head(5))
+        .await?;
+
+    let file_path = format!("{}_part1", destination_file_name);
+    let mut file = File::create(&file_path).await?;
+    while let Some(data) = reader.next().await.transpose()? {
+        file.write_all(&data).await?;
+    }
+    file.sync_all().await?;
```
coryan: Why download to a file?  That seems more complicated than downloading to a buffer (specially for 5 bytes!) and distracts from the main topic of the example.

src/storage/examples/src/objects/open_object_multiple_ranged_read.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    // 2. Request the last 2 bytes using the existing `descriptor`.
+    let mut reader = descriptor.read_range(ReadRange::tail(2)).await;
+
+    let file_path = format!("{}_part2", destination_file_name);
+    let mut file = File::create(&file_path).await?;
+    while let Some(data) = reader.next().await.transpose()? {
+        file.write_all(&data).await?;
+    }
+    file.sync_all().await?;
```
coryan: Ditto.

src/storage/examples/src/objects/open_object_multiple_ranged_read.rs
```diff
@@ -0,0 +1,57 @@
...
+    let (descriptor, mut reader) = client
+        .open_object(bucket_name, object)
+        .send_and_read(ReadRange::head(5))
+        .await?;
+
+    let file_path = format!("{}_part1", destination_file_name);
+    let mut file = File::create(&file_path).await?;
+    while let Some(data) = reader.next().await.transpose()? {
+        file.write_all(&data).await?;
+    }
```
coryan, line 31: Aside, it would be nice to offer a helper function to collect all this data into a buffer: #2856

src/storage/examples/src/lib.rs
```diff
@@ -408,6 +408,17 @@ pub async fn run_object_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
     )
     .await?;
 
+    tracing::info!("running open_object_multiple_ranged_read example");
+    let downloaded_file = tempfile::NamedTempFile::new()?;
+    let downloaded_file_name = downloaded_file.path().to_str().unwrap();
```
coryan: nit: here and below

```suggestion
    let downloaded_file_prefix = downloaded_file.path().to_str().unwrap();
```

src/storage/examples/src/objects/open_object_multiple_ranged_read.rs
vsharonlynn: Thanks for your inputs! I agree with your concerns.

Here are my changes to simplify the logic to focus on the main topic of the example:
* Write to file --> Download to a buffer and printing
* Downloading 5 and 2 bytes --> Download 5 bytes for both ranges
* Clarified print statements to say "Read first range[...]" and "Read second range[...]" and print the content.

Please let me know if this version is clearer.

coryan (approved):

Issue #3964, opened 2025-12-02
Support exactly-once subscribe


dbolduc:
Our initial implementation will not support exactly once subscribe. We will need to:

- [ ] Add exactly once acking to the `Leaser` and `DefaultLeaser`
- [ ] Add an async `ack` to the handler, plumbing results back to the handler.

dbolduc:
Handing this off... the remaining work looks something like this:

https://github.com/googleapis/google-cloud-rust/commit/bafa17f526e7895a9a399113a4c11139b3cc2a61

PhongChuong:
@dbolduc , thanks again for you implementation in https://github.com/googleapis/google-cloud-rust/commit/bafa17f526e7895a9a399113a4c11139b3cc2a61.

PR #4786, opened 2026-02-24
refactor(pubsub): prepare for exactly-once leasing

dbolduc:
Part of the work for #3964. A pure refactor.

Introduce `LeaseInfo`. Plumb it from the stream to the lease loop, to the lease state.

This will grow to include an `ExactlyOnce` branch, which holds a `oneshot::Sender` for returning the final result of the ack.

### DIFF
--- a/src/pubsub/src/subscriber/lease_loop.rs
+++ b/src/pubsub/src/subscriber/lease_loop.rs
@@ -98,7 +98,7 @@ mod tests {
 
     fn test_message(id: i32) -> NewMessage {
         NewMessage {
-            ack_id: test_id(i),
+            ack_id: test_id(id),
             lease_info: test_info(),
         }
     }


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -49,13 +49,30 @@ impl Default for LeaseOptions {
     }
 }
 
+pub(super) type NewMessage = (String, LeaseInfo);
```
coryan: A tuple seems a bit hacky.  Why not use a `struct` ?

dbolduc: It's a `struct` now.

> A tuple seems a bit hacky

It was the (k, v) of the hashmap, so not completely random.

PR #4788, opened 2026-02-24
feat(gax-internal): add traces to raw HTTP requests

coryan:
Some amount of refactoring to make the code easier to grok.

Part of the work for #3178

### DIFF
--- a/tests/o11y/src/e2e/metrics.rs
+++ b/tests/o11y/src/e2e/metrics.rs
@@ -43,7 +43,9 @@ pub async fn run() -> anyhow::Result<()> {
             metric.record(d.as_secs_f64(), &[KeyValue::new("testId", id.clone())]);
         }
         // Ignore errors because it may have flushed recently.
-        let _ = provider.force_flush();
+        if let Err(e) = provider.force_flush() {
+            tracing::warn!("error flushing provider: {e:}");
+        }
         found = try_get_metric(&client, &project_id, METRIC_NAME, ("testId", id.as_str())).await?;
         if found.is_some() {
             break;
--- a/tests/o11y/src/e2e/showcase.rs
+++ b/tests/o11y/src/e2e/showcase.rs
@@ -74,7 +74,9 @@ pub async fn run() -> anyhow::Result<()> {
     );
 
     // 4. Force flush to ensure spans are sent.
-    provider.force_flush()?;
+    if let Err(e) = provider.force_flush() {
+        tracing::error!("error flushing provider: {e:}");
+    }
 
     // 5. Verify (Poll Cloud Trace API)
     let trace = wait_for_trace(&project_id, &trace_id).await?;
--- a/tests/o11y/src/e2e/storage.rs
+++ b/tests/o11y/src/e2e/storage.rs
@@ -80,7 +80,9 @@ async fn send_trace(project_id: &str) -> anyhow::Result<String> {
     );
 
     // 4. Force flush to ensure spans are sent.
-    provider.force_flush()?;
+    if let Err(e) = provider.force_flush() {
+        tracing::error!("error flushing provider: {e:}");
+    }
     Ok(trace_id)
 }
 
--- a/tests/o11y/tests/metrics.rs
+++ b/tests/o11y/tests/metrics.rs
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
+// Copyright 2026 Google LLC
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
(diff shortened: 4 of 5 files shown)


src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -104,60 +106,65 @@ pub(crate) fn create_http_attempt_span(
...
-    result: Result<&reqwest::Response, &google_cloud_gax::error::Error>,
+pub fn record_intermediate_client_request(
+    result: &Result<reqwest::Response>,
     prior_attempt_count: u32,
     method: &http::Method,
     request_url: &reqwest::Url,
 ) {
     let span = Span::current();
+    // Only enrich spans that are explicitly marked as GAX client spans.
+    // This prevents accidental enrichment of user-provided spans that happen to have the same fields.
```
dbolduc: Doesn't this comment belong inside the `if` block that checks for the presence of "gax.client.span" attribute?

src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -395,7 +402,7 @@ mod tests {
     #[test_case(StatusCode::OK; "OK")]
     #[test_case(StatusCode::CREATED; "Created")]
     #[tokio::test]
-    async fn test_record_response_attributes_ok(status_code: StatusCode) {
+    async fn record_http_ok(status_code: StatusCode) {
```
dbolduc: nit, predating this PR: this should probably return `-> anyhow::Result<()>` and then we replace some `unwrap()`s with `?`.

src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -643,19 +648,23 @@ mod tests {
...
+        let t4 = tracing::info_span!("t4");
+        let _ = Ok(response).record_http(&t4, 1, Method::GET, url);
 
         let captured = TestLayer::capture(&guard);
-        assert_eq!(captured.len(), 1, "captured spans: {:?}", captured);
+        let _ = captured
+            .iter()
+            .find(|s| s.name == "client_request")
+            .unwrap_or_else(|| panic!("cannot find `client_request` span: {captured:?}"));
         let attributes = &captured[0].attributes;
```
dbolduc:
```suggestion
        let attributes = captured
            .iter()
            .find(|s| s.name == "client_request")
            .unwrap_or_else(|| panic!("cannot find `client_request` span: {captured:?}"))
            .attributes;
```

dbolduc: or do this in two lines, like below.

src/gax-internal/src/http.rs
```diff
@@ -249,6 +247,37 @@ impl ReqwestClient {
...
+        use crate::observability::HttpResultExt;
+        let span = create_http_attempt_span(
+            &request,
+            &options,
+            self.instrumentation,
+            attempt_info.attempt_count,
+        );
+        let (method, url) = (request.method().clone(), request.url().clone());
+        self.execute_http_untraced(request)
+            .instrument(span.clone())
```
dbolduc: nit on the name: Consider `execute_http_inner` instead of `execute_http_untraced`

just cause `*_untraced(...).instrument(...)` looks goofy

src/gax-internal/src/http.rs
```diff
@@ -368,45 +397,39 @@ impl ReqwestClient {
...
+    async fn request_attempt_traced(
+        &self,
+        request: reqwest::Request,
+        options: &RequestOptions,
+        attempt_count: u32,
+    ) -> Result<reqwest::Response> {
+        use crate::observability::HttpResultExt;
+        let span = create_http_attempt_span(&request, options, self.instrumentation, attempt_count);
+        let (method, url) = (request.method().clone(), request.url().clone());
+        self.request_attempt_untraced(request)
```
dbolduc: same

src/gax-internal/src/observability/http_tracing.rs
coryan: Fixed.

src/gax-internal/src/observability/http_tracing.rs
coryan: Thanks, done (option 2, option 1 does not work).

src/gax-internal/src/http.rs
coryan: Fixed.

src/gax-internal/src/http.rs
coryan: Fixed too.

src/gax-internal/src/observability/http_tracing.rs
coryan: Ack. I was tempted, but some of these have a single `.unwrap()`.  Seemed like a bad tradeoff except for the surprise factor.

PR #4793, opened 2026-02-24
cleanup(gax): remove or deprecate unused o11y functions

coryan:
Part of the work for #4772

### DIFF
--- a/src/gax/src/options.rs
+++ b/src/gax/src/options.rs
@@ -262,7 +262,9 @@ pub mod internal {
     #[derive(Debug, Clone, Default, PartialEq)]
     pub struct ResourceName(pub String);
 
-    // TODO(#4772) - stop using these functions and use `TracingExtension` directly
+    // Cannot remove this function, as that would break any client libraries
+    // that are released and use this function.
+    #[deprecated]
     pub fn set_path_template(
         options: RequestOptions,
         path_template: &'static str,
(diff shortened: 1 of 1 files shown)


src/gax/src/options.rs
```diff
@@ -262,32 +262,22 @@ pub mod internal {
...
 
-    // TODO(#4772) - stop using these functions and use `TracingExtension` directly
+    // Cannot remove this function, as that would break any client libraries
+    // that are released and use this function.
+    #[deprecated]
     pub fn set_path_template(
         options: RequestOptions,
         path_template: &'static str,
     ) -> RequestOptions {
         options.insert_extension(PathTemplate(path_template))
```
dbolduc, line 272: Q: should it do nothing? i.e. just give back the `options`

src/gax/src/options.rs
```diff
@@ -262,32 +262,22 @@ pub mod internal {
...
     }
 
-    // TODO(#4772) - stop using these functions and use `TracingExtension` directly
+    // Cannot remove this function, as that would break any client libraries
+    // that are released and use this function.
+    #[deprecated]
     pub fn get_path_template(options: &RequestOptions) -> Option<&'static str> {
         options.get_extension::<PathTemplate>().map(|e| e.0)
```
dbolduc, line 279: Same Q. I am not sure preserving the behavior buys us anything (although it is easy to do).

src/gax/src/options.rs
coryan: If an existing library out there (say `google-cloud-storage`) picked out a new version of `*-gax` and this function did not work that would be sad, no?  I think it is unlikely, but breaking behavior is about as bad as breaking interface, maybe even worse?

src/gax/src/options.rs
coryan: In practice, I think nothing is using it, but given the cost I would rather keep the behavior.  If we are going to break the behavior then `unreachable!()` may be the way to go.

src/gax/src/options.rs
dbolduc: The only time we `get_path_template` is behind a `#[cfg(google_cloud_unstable_tracing)]` in gaxi...

So that is the only thing that could break.

> maybe even worse

It might be worse to accidentally depend on the behavior in the future. :shrug:

(But none of this is important. I just like arguing).

PR #4798, opened 2026-02-24
impl(pubsub): basic exactly-once handler

dbolduc:
Part of the work for #3964 

Start the implementation for the `ExactlyOnce` handler. This adds the common ack/nack functions.

A follow up will add a function for confirmed acks. Another follow up will add an `ExactlyOnce` branch to the `Handler`. etc.

### DIFF
--- a/src/pubsub/src/subscriber/handler.rs
+++ b/src/pubsub/src/subscriber/handler.rs
@@ -167,6 +167,76 @@ impl Drop for AtLeastOnce {
     }
 }
 
+#[cfg(test)] // TODO(#3964): implementation in progress...
+/// A handler for exactly-once delivery.
+#[derive(Debug)]
+pub struct ExactlyOnce {
+    inner: Option<ExactlyOnceImpl>,
+}
+
+#[cfg(test)] // TODO(#3964): implementation in progress...
+impl ExactlyOnce {
+    pub(super) fn new(
+        ack_id: String,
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/handler.rs
```diff
@@ -225,4 +295,30 @@ mod tests {
...
+        Ok(())
+    }
+
+    #[test]
+    fn exactly_once_nack() -> anyhow::Result<()> {
+        let (ack_tx, mut ack_rx) = unbounded_channel();
+        let h = ExactlyOnce::new(test_id(1), ack_tx);
+        assert_eq!(ack_rx.try_recv(), Err(TryRecvError::Empty));
+
+        drop(h);
```
suzmue, line 318: nit: worth testing an explicit .nack() as well?

suzmue (approved):

src/pubsub/src/subscriber/handler.rs
dbolduc: No such thing. (Well, there is `ExactlyOnceImpl::nack`, but I am not testing that struct, because it's not public).

dbolduc: @haphungw and I decided we don't need a `h.nack()` when `drop(h)` is the same exact thing. https://github.com/googleapis/google-cloud-rust/issues/4514

PR #4808, opened 2026-02-25
refactor(guide): prepare `update_resource` for devsite

coryan:


guide/samples/src/update_resource.rs
```diff
@@ -12,30 +12,30 @@
...
+use google_cloud_wkt::FieldMask;
+use std::collections::HashMap;
 
-// ANCHOR: update-field
 pub async fn update_field(project_id: &str) -> anyhow::Result<()> {
-    use google_cloud_secretmanager_v1::client::SecretManagerService;
-    use std::collections::HashMap;
-
-    // ANCHOR: create
+    // [START rust_create] ANCHOR: create
```
dbolduc: Q: are these global region tags? We might want to make them more particular like `rust_update_resource_create`

coryan: Fixed.

Issue #4809, opened 2026-02-25
Stabilize `AckError`


dbolduc:
We need an `AckError` type to represent the result of a confirmed ack in exactly-once delivery.

Golang uses an enum + a generic error:

https://github.com/googleapis/google-cloud-go/blob/6ddbda6521b6a7f70bd3e05a21c1d64b80e25847/internal/pubsub/message.go#L116-L117

https://github.com/googleapis/google-cloud-go/blob/6ddbda6521b6a7f70bd3e05a21c1d64b80e25847/pubsub/message.go#L116-L127

They map to the `FailedPrecondition` and `PermissionDenied` branches based on the gRPC status code. All other gRPC status codes are mapped to `Other`. I can't say I like this.

My plan is to have:

```rs
/// Represents an error that can occur when acknowledging a message.
#[derive(thiserror::Error, Debug)]
#[non_exhaustive]
pub enum AckError {
    /// The message expired before the client could attempt an acknowledgement.
    ///
    /// The message has not been acknowledged, and will be redelivered.
    #[error(
        "the message has already expired. It has not been acknowledged, and will be redelivered."
    )]
    MessageExpired,

    /// The underlying RPC failed.
    #[error("the acknowledgement failed{}. RPC error: {source}",
        .details.as_ref().map(|m| format!(", with the server reporting: {m}")).unwrap_or_default())]
    Rpc {
        /// Extra details from the server on why the acknowledgement failed.
        details: Option<String>,

        /// The error returned by the service for the acknowledge request.
        #[source]
        source: Arc<Error>,
    },
}
```

Where the `details` field represents any metadata related to the message in question. I plan to just copy whatever the service tells us, such as `PERMANENT_FAILURE_INVALID_ACK_ID`. And gRPC errors without `ErrorInfo` for the particular ack ID get mapped to `None`.

dbolduc:
The documentation for `Shutdown` could use some work. It will be easier to write when there is a shutdown API to reference.

https://github.com/googleapis/google-cloud-rust/pull/4867#pullrequestreview-3879395076

PR #4810, opened 2026-02-25
impl(pubsub): introduce AckError

dbolduc:
Part of the work for #3964 

This is a soon-to-be public error type which we will use for exactly-once confirmed acks.

I have other PRs blocked on an `AckError` existing, but not the particular representation. We do not need to get it perfect in this PR. I opened #4809 to track stabilizing the interface.

With that being said, feedback on this interface is appreciated. I think the `details` field deserves scrutiny. I mostly did not want to lose context we could potentially forward to the application.

### DIFF
--- a/src/pubsub/src/error.rs
+++ b/src/pubsub/src/error.rs
@@ -56,13 +56,14 @@ pub enum PublishError {
 #[derive(thiserror::Error, Debug)]
 #[non_exhaustive]
 pub enum AckError {
-    /// The message expired before the client could attempt an acknowledgement.
+    /// The message's lease expired before the client could acknowledge it.
     ///
-    /// The message has not been acknowledged, and will be redelivered.
+    /// The message has not been acknowledged, and will be redelivered, maybe to
+    /// another client.
     #[error(
-        "the message has already expired. It has not been acknowledged, and will be redelivered."
+        "the message's lease has already expired. It was not acknowledged, and will be redelivered."
     )]
-    MessageExpired,
+    LeaseExpired,
 
     /// The underlying RPC failed.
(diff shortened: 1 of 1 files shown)


src/pubsub/src/error.rs
```diff
@@ -47,3 +50,60 @@ pub enum PublishError {
     #[error("message size exceeded configured byte threshold")]
     ExceededByteThresholdError(()),
 }
+
+#[cfg(test)] // TODO(#3964): implementation in progress...
+/// Represents an error that can occur when acknowledging a message.
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+pub enum AckError {
+    /// The message expired before the client could attempt an acknowledgement.
```
coryan:
```suggestion
    /// The message expired before the client could acknowledge the message.
```

Maybe the client attempted early but there traffic jams in the interweb pipes...

src/pubsub/src/error.rs
```diff
@@ -47,3 +50,60 @@ pub enum PublishError {
...
+    ///
+    /// The message has not been acknowledged, and will be redelivered.
```
coryan:
```suggestion
    /// The message has not been acknowledged, and will be redelivered, maybe to a different client.
```

src/pubsub/src/error.rs
dbolduc: Done. Also changed this branch to be `LeaseExpired`, not `MessageExpired`.

src/pubsub/src/error.rs
dbolduc: Done

PR #4812, opened 2026-02-25
feat(gax): add with_user_agent to ClientBuilder

joshuatants:
Splitting the changes to `gax-internal` and `gax` into its own PR as requested in https://github.com/googleapis/google-cloud-rust/pull/4601#pullrequestreview-3777432667 .

### DIFF
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -69,6 +69,7 @@ pub struct Client {
     retry_throttler: SharedRetryThrottler,
     polling_error_policy: Arc<dyn PollingErrorPolicy>,
     polling_backoff_policy: Arc<dyn PollingBackoffPolicy>,
+    user_agent: Option<String>,
 }
 
 impl Client {
@@ -136,6 +137,7 @@ impl Client {
             polling_backoff_policy: config
                 .polling_backoff_policy
                 .unwrap_or_else(|| Arc::new(ExponentialBackoff::default())),
+            user_agent: config.user_agent,
         })
     }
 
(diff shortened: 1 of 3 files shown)


src/gax/src/client_builder.rs
```diff
@@ -380,6 +380,25 @@ impl<F, Cr> ClientBuilder<F, Cr> {
         self.config.polling_backoff_policy = Some(v.into().0);
         self
     }
+
+    /// Sets the user-agent.
+    ///
+    /// The user-agent header is set in all requests made by the client.
```
coryan, line 386:
```suggestion
    /// Replaces the default value for the `User-Agent` header.
```

Which I guess makes me wonder: "how do I find out the default value"? and "how do I prepend an element and leave the existing values untouched"? (the latter is the most common usage).

src/gax-internal/src/grpc.rs
```diff
@@ -498,12 +509,84 @@ where
...
 #[cfg(test)]
-#[cfg(google_cloud_unstable_tracing)]
 mod tests {
     use super::Client;
+    #[cfg(google_cloud_unstable_tracing)]
     use crate::options::InstrumentationClientInfo;
+    use google_cloud_gax::options::RequestOptions;
+
+    #[tokio::test]
+    async fn test_make_headers_user_agent_none() {
```
coryan, line 519: Thanks for the thorough tests!

coryan:
Let me know if you still want to proceed with these changes, otherwise we should close this PR.

rfz-com:
I would very much like to be able to explicitly dictate the user-agent string to be able to differentiate between apps and app-versions and app-versions-on-whateverOS as that kind of information is sort of standard to be able to do from a client.
Please.

joshuatants:
Closing this now that https://github.com/googleapis/google-cloud-rust/pull/4839 is merged.

PR #4813, opened 2026-02-25
refactor(guide): prepare `retry_policies.rs` for devsite

coryan:


guide/samples/src/retry_policies.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright 2025 Google LLC
```
dbolduc, line 1: revert?

coryan: Fixed, and resolved conflicts.

PR #4816, opened 2026-02-25
docs(storage): add sample for storage_open_multiple_objects_ranged_read

vsharonlynn:
Add one of the samples requested in #4595.

### DIFF
--- a/src/storage/examples/src/objects/open_multiple_objects_ranged_read.rs
+++ b/src/storage/examples/src/objects/open_multiple_objects_ranged_read.rs
@@ -55,8 +55,8 @@ pub async fn sample(client: &Storage, bucket: &str, object_names: &[&str]) -> an
         println!("The footer for object {object_name} is {footer:?}");
         descriptors.insert(object_name, (descriptor, footer));
     }
-    
-    // You can use the descriptors in the map to read more ranges.
+
+    // Use the descriptors in the map to read more ranges.
 
     Ok(())
 }


src/storage/examples/src/objects/open_multiple_objects_ranged_read.rs
```diff
@@ -0,0 +1,54 @@
...
+
+            Ok::<(String, Vec<u8>), anyhow::Error>((object, content))
+        });
+    }
+
+    while let Some(result) = set.join_next().await {
+        let (object, content) = result??;
+        println!(
+            "Downloaded first 10 bytes of object {object} in bucket {bucket}. Content: {:?}",
+            String::from_utf8_lossy(&content)
```
coryan: Why convert to a string when you are saying they are bytes and are treated as such?

```suggestion
            "Downloaded first 10 bytes of object {object} in bucket {bucket}. Content: {:?}",
            content
```

src/storage/examples/src/objects/open_multiple_objects_ranged_read.rs
```diff
@@ -0,0 +1,54 @@
...
+        );
+    }
+
```
coryan: This is slightly odd: why use `open_object()` to read a single range?  You could use `read_object()` for that, and it may even be more efficient (depending on the size of the range).

If the example must use `open_object()` then consider putting the descriptors in a hash map and showing how this could be reused...  Maybe something like:




```suggestion
    let mut set = JoinSet::new();
    let bucket_name = format!("projects/_/buckets/{bucket}");
    async fn  read_footer(client: &Storage, bucket: &str, object: &str)
        -> anyhow::Result<(ObjectDescriptor, Vec<u8>)>
    {
        // Reading the last 8 bytes is common in Parquet and other column-oriented formats
        // with a footer.
        let (descriptor, mut reader) = client.open_object(bucket_name.clone(), object)
            .send_and_read(ReadRange::tail(8))
            .await?;
        let mut footer = Vec::new();
        while let Some(data) = reader.next().await.transpose()? {
            content.extend_from_slice(&data);
        }
        Ok((descriptor, footer, object.to_string()))
    };
    
    object_names.iter().for_each(|o| set.spawn(read_footer(client, &bucket_name, o.as_str())));
    let mut descriptors = HashMap::new();
    while let Some(result) = set.join_next().await {
        let (descriptor, footer, name) = result??;
        println!("The footer for {name} is {footer:?}");
        descriptors.insert(name, (descriptor, footer));
    }
```

Then you can add a comment telling the customer that they have descriptors and footers ready:

```
 // You can use the descriptors in the map to read more ranges.
```

src/storage/examples/src/objects/open_multiple_objects_ranged_read.rs
coryan: Also the text is wrong.

vsharonlynn: Updated with the other comment. Thanks for the catch and sorry about that!

vsharonlynn:
@coryan I've pushed an update addressing your feedback. The sample now:
* Reads the last 8 bytes (simulating a Parquet footer)
* Handles the data as raw bytes
* Stores the ObjectDescriptors in a HashMap to demonstrate caching them for future reads
* Has an code comment to explain the descriptor map

Thank you!

src/storage/examples/src/objects/open_multiple_objects_ranged_read.rs
```diff
@@ -0,0 +1,63 @@
...
+            .open_object(bucket_name, object_name)
+            .send_and_read(ReadRange::tail(8))
+            .await?;
+
+        let mut footer = Vec::new();
+        while let Some(data) = reader.next().await.transpose()? {
+            footer.extend_from_slice(&data);
+        }
+        Ok((descriptor, footer, object_name.to_string()))
+    }
```
coryan, line 42: You could move this outside (or below) the body of the `sample()` function. Frankly, I am not sure what is best here, you folks will need to develop a style for the samples.  Looks good enough as-is.

vsharonlynn: Thanks, Carlos.

I considered moving this out and same like you - I'm not sure whether it's better to move it out or keep it as an inner function in the context of a sample. My thought is to keep a highly specific, single-use helper function to where it is used and not expose it to the module level. At least for now.

+1 on developing a style and standard for the samples.

FYI @joshuatants

joshuatants: Ack, let's discuss this further. 

LGTM.

PR #4822, opened 2026-02-25
fix(pubsub): publish messages larger than the configured batch threshold

PhongChuong:
This PR removes the ExceededByteThresholdError. The Publisher now sends the messages that are larger than the configured batch threshold in its own batch. For messages that are larger than the server message size limit, the Publisher will send the message and return the server side error message to the application. Similar to other server errors, if and only if the message have an ordering key, then we pause Publishing for that key.

This also remove the publisher_publish_msg_too_big test for Publisher as the case is handled the same way as any other server side error.

Fixes https://github.com/googleapis/google-cloud-rust/issues/4650

### DIFF
--- a/src/pubsub/src/publisher/actor.rs
+++ b/src/pubsub/src/publisher/actor.rs
@@ -454,7 +454,7 @@ impl SequentialBatchActor {
     ) {
         let mut should_flush = false;
         while let Some(next) = self.pending_msgs.front() {
-            if !batch.can_add(next) {
+            if !batch.can_add(next) && !batch.is_empty() {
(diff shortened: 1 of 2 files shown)


suzmue (changes requested):
Can you please add a test that covers this case?

Reading the code it doesn't seem like this will actually add the too big message to a batch. Doesn't can_add always return false so it just tries flush the empty batch? Even if so, I'd like to see a test to check the behavior.

PhongChuong:
@suzmue , added the tests. Good catch on the sequential actor empty message issue. PTAL.

PR #4823, opened 2026-02-25
fix(pubsub): update Flush to not block the Publisher

### LINKED ISSUE #4505: Pubsub Publisher flush operation blocks until completion
The Publisher flush operation awaits until all child actors flush to complete. We could instead tokio::spawn this as well so the publisher can keep working on additional messages. The Dispatcher would also need to keep track of any pending flushes, and make sure all of those resolve as well.

PhongChuong:
The Publisher currently blocks on a Flush operation to complete across all ordering keys before additional operations can be processed. For example during a Flush operation, when ordering key A Flush is complete and ordering key B is still pending, the Publisher will not process new operations for A until after B's Flush is complete.

This PR updates the Publisher to spawn a task to await on all the Flush operations instead of blocking the main Dispatcher loop. This allows the Publisher to dispatch new operations to the ordering key actors without awaiting for all keys Flush to complete. In the example above, new operations for A will make process. We are guarantee correct ordering because the ToBatchActor::Flush has been sent to each batch actors.

Fixes #4505

### DIFF
--- a/src/pubsub/src/publisher/actor.rs
+++ b/src/pubsub/src/publisher/actor.rs
@@ -161,13 +161,11 @@ impl Dispatcher {
                                 }
                                 flush_set.spawn(rx);
                             }
-                            // Wait on all the tasks that exist right now.
-                            // TODO(#4505): We could instead tokio::spawn this as well so the
-                            // publisher can keep working on additional messages. The Dispatcher
-                            // would also need to keep track of any pending flushes, and make sure
-                            // all of those resolve as well.
-                            flush_set.join_all().await;
-                            let _ = tx.send(());
+                            tokio::spawn(async move {
+                                // Wait on all the flush operations.
+                                flush_set.join_all().await;
+                                let _ = tx.send(());
+                            });
                         },
                         Some(ToDispatcher::ResumePublish(ordering_key)) => {
                             if let Some(batch_actor) = batch_actors.get_mut(&ordering_key) {


PhongChuong:
@suzmue , I've updated the PR description. See if that explains scenario better. If not, can you describe the scenario that you have in mind?

suzmue:
>  I've updated the PR description. See if that explains scenario better. If not, can you describe the scenario that you have in mind?

Ok sounds good. 

The scenario I am thinking of is when we are doing concurrent publishes (no ordering keys), the batcher is blocked waiting for those flush operations to complete:

https://github.com/googleapis/google-cloud-rust/blob/36a50c605c65585c8082b67e394d8d530f8f9d87/src/pubsub/src/publisher/actor.rs#L270-L275

suzmue (approved):

PhongChuong:
For your described scenario, we first need to discuss and clarify the behavior for Flush as defined [here] (https://github.com/googleapis/google-cloud-rust/blob/aaa316fc4f11f66e6878b1da705cf9d9622587c0/src/pubsub/src/publisher/client.rs#L93-L108). It is not clear to me when Flush completes for the following scenario (no ordering key):
1. let publish_1 = Publish(msg_1);
2. let flush_1 = Flush();
3. let publish_2 = Publish(msg_2);
4. let flush_2 = Flush();

Does flush_2 returning guarantee that all previous publish are complete (both publish_[1, 2])? Or just up to the previous Flush (just publish_2)?

I'll add a tracking issue and/or fix it in a following PR after the discussion.

PR #4824, opened 2026-02-25
impl(pubsub): confirmed acks in `Leaser`

dbolduc:
Part of the work for #3964 

Teach the leaser how to send exactly-once acks.

Reporting the results is a bit strange. We use a channel, as I explain in the comments.

The other thing to note is that the lease loop holds the only `&mut` to the `LeaseState`. So if we want to modify the `LeaseState` (i.e. to satisfy the promise for an exactly-once ack on a message, and drop the message from lease management), that needs to be triggered in the `LeaseLoop`. That is where the channel is ultimately sending its `AckResult`s to. 

I am just going to use `#[allow(dead_code)]` in one place. Annotating every single symbol with `#[cfg(test)]` was a nightmare.

### DIFF
--- a/src/pubsub/src/subscriber/handler.rs
+++ b/src/pubsub/src/subscriber/handler.rs
@@ -234,7 +234,7 @@ impl ExactlyOnceImpl {
 }
 
 /// The result of a confirmed acknowledgement.
-pub type AckResult = std::result::Result<(), AckError>;
+pub(crate) type AckResult = std::result::Result<(), AckError>;
 
 #[cfg(test)]
 mod tests {
--- a/src/pubsub/src/subscriber/leaser.rs
+++ b/src/pubsub/src/subscriber/leaser.rs
@@ -129,6 +129,12 @@ where
 
     /// The exactly-once ack retry loop.
     ///
+    /// The request has N ack IDs. The server can tell us the result of
+    /// individual acks in the response metadata.
+    ///
+    /// If the result for an ack ID is a success or permanent error, we can
+    /// report it, and remove that ack ID from subsequent attempts of the RPC.
+    ///
     /// Results are reported via the channel, as they are known. This lets us
     /// keep the retry logic in the leaser, while allowing for partial results
     /// to be reported before the entire operation completes.
--- a/src/pubsub/src/error.rs
+++ b/src/pubsub/src/error.rs
@@ -51,10 +51,11 @@ pub enum PublishError {
     ExceededByteThresholdError(()),
 }
 
+#[allow(dead_code)] // TODO(#3964): implementation in progress...
 /// Represents an error that can occur when acknowledging a message.
 #[derive(thiserror::Error, Debug)]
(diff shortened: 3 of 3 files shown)


src/pubsub/src/subscriber/leaser.rs
```diff
@@ -29,8 +33,25 @@ pub(super) trait Leaser {
...
+    /// This method handles retries. Results are reported via the channel, as
+    /// they are known. This lets us keep the retry logic in the leaser, while
+    /// allowing for partial results to be reported before the entire operation
+    /// completes.
+    ///
+    /// The caller should spawn a task for this operation.
+    async fn confirmed_ack(
+        &self,
+        ack_ids: Vec<String>,
+        confirmed_tx: UnboundedSender<ConfirmedAcks>,
```
dbolduc: This is incorrect / convoluted. The sender should be a field of the `DefaultLeaser`.

dbolduc: Fixed. Now the `LeaseState` will not need to hold the `Sender`.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -107,17 +126,47 @@ where
             .modify_ack_deadline(req, self.options.clone())
             .await;
     }
+
+    /// The exactly-once ack retry loop.
+    ///
+    /// Results are reported via the channel, as they are known. This lets us
+    /// keep the retry logic in the leaser, while allowing for partial results
```
suzmue, line 139: What partial results might we get?

src/pubsub/src/subscriber/handler.rs
```diff
@@ -167,14 +168,12 @@ impl Drop for AtLeastOnce {
     }
 }
 
-#[cfg(test)] // TODO(#3964): implementation in progress...
 /// A handler for exactly-once delivery.
 #[derive(Debug)]
 pub struct ExactlyOnce {
```
suzmue, line 173: this is actually pub now, is that fine?

src/pubsub/src/subscriber/leaser.rs
suzmue, line 139: what kind of partial results would come here?

dbolduc: The request has N ack IDs. The server can tell us the result of individual acks in the response metadata.

If they are final (success or permanent error), we can report them, and remove those ack IDs from subsequent attempts of the RPC.

^ Added this to the comment

src/pubsub/src/subscriber/handler.rs
dbolduc: Fine, but maybe better as `pub(crate)` for now in case there are any surprise releases. Fixed, thanks.

src/pubsub/src/subscriber/leaser.rs
dbolduc: see above

PR #4826, opened 2026-02-25
feat(compute): generate resource_name with sidekick

haphungw:
This PR is a demo of the generated Rust code for resource_name within `google-cloud-compute-v1`.


coryan (commented):
nit: the PR title is wrong, and you may want to try DNS first, it would be easier to review.

PR #4827, opened 2026-02-26
docs: Slight change in generated sample formatting

amanda-tarafa:
This is a slight regression in formatting , due to a simplification of templates which is peferrable. We'll deal with formatting as we add mire fields to samples.


amanda-tarafa:
Ready for review.

coryan (approved):

PR #4828, opened 2026-02-26
chore(spanner): add generated data client

olavloite:
Adds a generated client for google/spanner/v1.
This is the data API for Spanner.

This change was created by:
1. Adding configuration for generating google-cloud-spanner to librarian.yaml
2. Running the code generator

The code generator must be executed from source, as the latest release of librarian does not yet allow-list google/spanner/v1.

### DIFF
--- a/.gcb/scripts/docs-rs.sh
+++ b/.gcb/scripts/docs-rs.sh
@@ -15,6 +15,10 @@
 
 set -ev
 
+# This is the only case where using `nightly` for CI is acceptable. Or maybe
+# "we must do, with regrets". The rationale is that docs.rs uses nightly to
+# generate the documentation, our releases would break if this test does not
+# pass.
 rustup toolchain install nightly
 cargo install --locked cargo-docs-rs
 cargo install --locked cargo-workspaces
--- a/.gcb/scripts/lint-unstable.sh
+++ b/.gcb/scripts/lint-unstable.sh
@@ -15,8 +15,6 @@
 
 set -ev
 
-rustup toolchain install nightly
-rustup default nightly
 rustup component add clippy
 cargo version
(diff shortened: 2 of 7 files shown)


src/spanner/src/lib.rs
```diff
@@ -19,7 +19,41 @@
 //! **not** recommend that you use this crate in production. We welcome feedback
 //! about the APIs, documentation, missing features, bugs, etc.
 //!
-//! This crate contains traits, types, and functions to interact with
-//! [Spanner].
-//!
-//! [spanner]: https://cloud.google.com/spanner
+pub use google_cloud_gax::Result;
```
coryan, line 23: nit: while line before the first declaration, the `//! ` block is comments about the module, we don't want them to be associated with the `Error`.

src/spanner/src/status.rs
```diff
@@ -0,0 +1,47 @@
...
+        status = status.set_message(self.message);
+        status = status.set_details(
+            self.details
+                .into_iter()
+                .filter_map(gaxi::grpc::status::any_from_prost)
+                .collect::<Vec<wkt::Any>>(),
+        );
+        Ok(status)
+    }
+}
```
coryan, line 48: We will want unit tests for these, look at the `status.rs` idea in Storage for inspiration.

Or maybe we should refactor the code to a common place.  Maybe `gaxi`?  Feel free to merge as-is and create an issue for a follow up PR.

olavloite: Added tests for `status.rs` and created #4832 for combining the two files in a common place.

src/spanner/src/lib.rs
olavloite: Added empty line.

PR #4829, opened 2026-02-26
chore(guide): update gemini models

coryan:


alvarowolfx:
Here flash 2.0 is also mentioned: https://github.com/coryan/google-cloud-rust/blob/main/guide/src/generate_text_using_the_vertex_ai_gemini_api.md#send-a-prompt-to-the-vertex-ai-gemini-api

coryan:
> Here flash 2.0 is also mentioned: https://github.com/coryan/google-cloud-rust/blob/main/guide/src/generate_text_using_the_vertex_ai_gemini_api.md#send-a-prompt-to-the-vertex-ai-gemini-api

Thanks. Fixed that too

PR #4834, opened 2026-02-26
feat(gax-internal): server-streaming RPC support

olavloite:
Adds support for server-streaming RPC calls to gax-internal. This will be needed for the Spanner data-client, that uses server-streaming for ExecuteStreamingSql, StreamingRead, and BatchWrite.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -510,8 +510,8 @@ storage-samples          = { path = "src/storage/examples" }
 
 [workspace.lints.rust]
 unexpected_cfgs = { level = "deny", check-cfg = [
-  'cfg(google_cloud_unstable_storage_bidi)',
   'cfg(google_cloud_unstable_grpc_server_streaming)',
+  'cfg(google_cloud_unstable_storage_bidi)',
   'cfg(google_cloud_unstable_tracing)',
   'cfg(google_cloud_unstable_trusted_boundaries)',
   # Enables re-generation of protos. Only needed on major version updates to


src/gax-internal/src/grpc.rs
```diff
@@ -215,6 +215,62 @@ impl Client {
             .await)
     }
 
+    /// Opens a server stream.
+    pub async fn server_streaming<Request, Response>(
```
coryan, line 220: Is there any chance you may need to change this method (or `server_streaming_with_status`) before a release of the spanner client?

If so, we should hide it under a `#[cfg(google_cloud_unstable_*)]` guard so we can make changes without introducing breaks into `google-cloud-gax-internal`.

src/gax-internal/src/grpc.rs
```diff
@@ -215,6 +215,62 @@ impl Client {
...
+            options,
+            api_client_header,
+            request_params,
+        )
+        .await?
+        .map_err(to_gax_error)
+    }
+
+    /// Opens a server stream with detailed status.
+    pub async fn server_streaming_with_status<Request, Response>(
```
coryan: Do you need this method for spanner?  If not, we can add it as `pub(crate) async fn` or just `async fn` at first, and make it public only when needed.

src/gax-internal/src/grpc.rs
olavloite: I've added a `google_cloud_unstable_server_streaming` guard to be sure.

src/gax-internal/src/grpc.rs
olavloite: Changed to `pub(crate)` for now

src/gax-internal/src/grpc.rs
coryan: Thanks.

src/gax-internal/src/grpc.rs
```diff
@@ -215,6 +215,64 @@ impl Client {
...
+            api_client_header,
+            request_params,
+        )
+        .await?
+        .map_err(to_gax_error)
+    }
+
+    /// Opens a server stream with detailed status.
+    #[cfg(google_cloud_unstable_grpc_server_streaming)]
+    pub(crate) async fn server_streaming_with_status<Request, Response>(
```
coryan, line 247: I don't think you need both, but does not hurt either.

olavloite: (I'll consider removing the annotation in a later follow-up)

PR #4838, opened 2026-02-27
feat(dns): generate resource naming for google-cloud-dns-v1

haphungw:
This PR is a demo of the generated Rust code for resource_name within `google-cloud-dns-v1`. `dns-v1` is also another Discovery API.

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -561,7 +561,6 @@ libraries:
       - googleapis
     specification_format: discovery
     rust:
-      resource_name: true
       package_dependencies:
         - name: google-cloud-lro
           package: google-cloud-lro
@@ -581,6 +580,7 @@ libraries:
         - projects
       disabled_clippy_warnings:
         - doc_lazy_continuation
+      resource_name: true
       discovery:
         operation_id: .google.cloud.compute.v1.Operation
         pollers:
@@ -750,11 +750,11 @@ libraries:
       - googleapis
     specification_format: discovery
     rust:
-      resource_name: true
       package_dependencies:
         - name: google-cloud-lro
           package: google-cloud-lro
           force_used: true
+      resource_name: true
       discovery:
         operation_id: ..Operation
         pollers:


src/generated/cloud/dns/v1/src/transport.rs
```diff
@@ -91,6 +91,25 @@ impl super::stub::Changes for Changes {
...
+                Some(&req)
+                    .map(|m| &m.project)
+                    .map(|s| s.as_str())
+                    .filter(|s| !s.is_empty())?,
+                Some(&req)
+                    .map(|m| &m.managed_zone)
+                    .map(|s| s.as_str())
+                    .filter(|s| !s.is_empty())?,
+            ))
+        })() {
```
coryan: Now that I see all this code put together: note that the same computation was done in lines 55-65.  We could return `(builder, method, path)` from that expression and save ourselves a lot of code.

haphungw: The `path` computed up on 55-65 is the literal HTTP routing path (`/dns/v1/projects/{}/managedZones/{}/changes`). 

We cannot reuse it down here because the "canonical relative resource name" needed for tracing excluded API versions and method suffixes (`//dns.googleapis.com/projects/{}/managedZones/{}`).

---

I also tried deduplicating the inner accessors (extracting `Some(&req).map(|m| &m.project)...` into local variables that both blocks could share). However, that seems to require the generator to dynamically escape Rust keywords, and just seems _hard_ in general.

haphungw: [https://github.com/googleapis/google-cloud-rust/compare/main...coryan:google-cloud-rust:prototype-dns-v1-resource-name](url)

Now I understand your suggestion. Splitting the HTTP paths into prefix, resource, and suffix makes sense. I'll look into it more.

src/generated/cloud/dns/v1/src/transport.rs
```diff
@@ -64,12 +64,24 @@ impl super::stub::Changes for Changes {
...
+                    try_match(
+                        Some(&req).map(|m| &m.project).map(|s| s.as_str()),
+                        &[Segment::SingleWildcard]
+                    )?,
+                    try_match(
+                        Some(&req).map(|m| &m.managed_zone).map(|s| s.as_str()),
+                        &[Segment::SingleWildcard]
+                    )?,
+                );
+
```
coryan, line 72: This is closer, but we are still doing the `try_match` twice, and the second time they are redundant [^1].  Can we get the compiler to generate:


```suggestion
        let project_id = try_match(
                        Some(&req).map(|m| &m.project).map(|s| s.as_str()),
                        &[Segment::SingleWildcard]
                    )?;
       let id_managed_zone_id =                     try_match(
                        Some(&req).map(|m| &m.managed_zone).map(|s| s.as_str()),
                        &[Segment::SingleWildcard]
                    )?,
       let path = format!(
            "/dns/v1/projects/{project_id}/managedZones/{managed_zone_id}/changes");
       let resource_name = format!(
           "//dns.googleapis.com/projects/{project_id}/managedZones/{managed_zone_id}");
```

If we always name the variables `${segment}_id` (or `id_${segment}`) then they never conflict with `path` and `resource_name`.

I guess we have to worry about the variables conflicting with each other.

I don't think it matters if the variables are inside the format string or as arguments.

[^1]: you could also convince me that the compiler elides the second `try_match()` in release mode.  I could be convinced by a godbolt.org output or something similar.

haphungw: Thanks for the suggestion. I'll continue working on the template. This is to confirm that the algorithm is working as expected.

src/generated/cloud/dns/v1/src/transport.rs
```diff
@@ -2222,24 +2035,25 @@ impl super::stub::ResourceRecordSets for ResourceRecordSets {
...
+                )?;
+                let managed_zone = try_match(
+                    Some(&req).map(|m| &m.managed_zone).map(|s| s.as_str()),
+                    &[Segment::SingleWildcard],
+                )?;
+                let name = try_match(
+                    Some(&req).map(|m| &m.name).map(|s| s.as_str()),
+                    &[Segment::SingleWildcard],
+                )?;
+                let r#type = try_match(
```
coryan: nice.

src/generated/cloud/dns/v1/src/transport.rs
```diff
@@ -113,30 +123,36 @@ impl super::stub::Changes for Changes {
...
-                    )?,
-                    try_match(
-                        Some(&req).map(|m| &m.change_id).map(|s| s.as_str()),
-                        &[Segment::SingleWildcard]
-                    )?,
+                    project, managed_zone, change_id,
+                );
+
+                let _resource_name = format!(
+                    "//dns.googleapis.com/projects/{}/managedZones/{}/changes/{}",
```
westarle, line 148: these resource names lgtm

librarian.yaml
```diff
@@ -581,6 +580,7 @@ libraries:
         - projects
       disabled_clippy_warnings:
         - doc_lazy_continuation
+      resource_name: true
```
coryan, line 583: Could we start with just `google-cloud-dns-v1` and (optionally) `google-cloud-showcase-v1beta1` ?

Compute is too large to effectively review.

haphungw: sounds good, i'll go with these 2 from now on.

PR #4839, opened 2026-02-27
feat(storage): support setting a custom user-agent on the client

joshuatants:
As requested in https://github.com/googleapis/google-cloud-rust/issues/3615

### DIFF
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -681,25 +681,6 @@ impl ClientBuilder {
         self
     }
 
-    /// Configures the user-agent header.
-    ///
-    /// Sets the `User-Agent` header for all requests.
-    ///
-    /// # Example
-    /// ```
-    /// # use google_cloud_storage::client::Storage;
-    /// # async fn sample() -> anyhow::Result<()> {
-    /// let client = Storage::builder()
-    ///     .with_user_agent("my-application/1.0.0")
-    ///     .build()
-    ///     .await?;
-    /// # Ok(()) }
-    /// ```
-    pub fn with_user_agent(mut self, user_agent: impl Into<String>) -> Self {
-        self.common_options.user_agent = Some(user_agent.into());
-        self
-    }
-
     pub(crate) fn apply_default_credentials(&mut self) -> BuilderResult<()> {
         if self.config.cred.is_some() {
             return Ok(());
@@ -818,20 +799,6 @@ pub(crate) mod tests {
         assert!(
             config.grpc_subchannel_count.is_some_and(|v| v == 42),
             "{config:?}"
-        );
-    }
-
-    #[test]
-    fn user_agent() {
-        let user_agent = "my-application/1.0.0";
-        let builder = ClientBuilder::new()
-            .with_credentials(Anonymous::new().build())
-            .with_user_agent(user_agent);
-        let (_, options) = builder.into_parts().expect("should get parts");
-        assert_eq!(
-            options.common_options.user_agent.as_deref(),
-            Some(user_agent),
-            "{options:?}",
         );
     }
 
--- a/src/storage/src/storage/read_object.rs
+++ b/src/storage/src/storage/read_object.rs
@@ -407,6 +407,25 @@ where
         self
     }
 
+    /// Sets the `User-Agent` header for this request.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_storage::client::Storage;
+    /// # async fn sample(client: &Storage) -> anyhow::Result<()> {
+    /// let response = client
+    ///     .read_object("projects/_/buckets/my-bucket", "my-object")
+    ///     .with_user_agent("my-app/1.0.0")
+    ///     .send()
+    ///     .await?;
+    /// println!("response details={response:?}");
+    /// # Ok(()) }
+    /// ```
+    pub fn with_user_agent(mut self, user_agent: impl Into<String>) -> Self {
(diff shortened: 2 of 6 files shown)


joshuatants:
I'm wondering if it's better to add `user_agent` agent directly to `storage::RequestOptions` instead of to `CommonOptions`. WDYT?

src/storage/src/storage/client.rs
```diff
@@ -676,6 +676,25 @@ impl ClientBuilder {
...
+    ///     .with_user_agent("my-application/1.0.0")
+    ///     .build()
+    ///     .await?;
+    /// # Ok(()) }
+    /// ```
+    pub fn with_user_agent(mut self, user_agent: impl Into<String>) -> Self {
+        self.common_options.user_agent = Some(user_agent.into());
+        self
+    }
+
```
coryan: There is no requirement to add this function.  If we want the storage client to work just like the other clients we can limit the new public API to the request builders, that is in these structs:

https://github.com/googleapis/google-cloud-rust/blob/3f3d6de83c34903c40eb402989a6c955f7ff4520/src/storage/src/storage/open_object.rs#L106


https://github.com/googleapis/google-cloud-rust/blob/3f3d6de83c34903c40eb402989a6c955f7ff4520/src/storage/src/storage/write_object.rs#L85-L88

https://github.com/googleapis/google-cloud-rust/blob/3f3d6de83c34903c40eb402989a6c955f7ff4520/src/storage/src/storage/read_object.rs#L89-L92

joshuatants: Got it. Updated the PR.

src/gax-internal/src/http/reqwest.rs
```diff
@@ -22,6 +22,6 @@ pub use reqwest::Request;
 pub use reqwest::RequestBuilder;
 pub use reqwest::Response;
 pub use reqwest::StatusCode;
-pub use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
+pub use reqwest::header::{HeaderMap, HeaderName, HeaderValue, USER_AGENT};
```
coryan: This change seems unrelated.

src/gax-internal/src/http.rs
```diff
@@ -322,7 +322,7 @@ impl ReqwestClient {
     ) -> Result<reqwest::RequestBuilder> {
         if let Some(user_agent) = options.user_agent() {
             builder = builder.header(
-                ::reqwest::header::USER_AGENT,
+                reqwest::USER_AGENT,
```
coryan: Ditto.

rfz-com:
So the below should be possible with this change?

`    let latest_builder: ReadObject =
        gcs_client.read_object(format!("projects/_/buckets/{bucket}"), version_marker);

    //  Looks like this is addressed by issue: https://github.com/googleapis/google-cloud-rust/issues/3615
    //        .with_user_agent(format!("{app_name}/{user_agent_version} ({os_arch})"));
`

coryan:
> So the below should be possible with this change?
> 

yes, that is the intent.

src/gax-internal/src/http/reqwest.rs
joshuatants: I wanted to use it in the read_object test. I'll open a comment in read_object so there's more context. For now I'll revert this change so that the PR can be merged.

src/gax-internal/src/http.rs
joshuatants: Reverted.

src/storage/src/storage/read_object.rs
```diff
@@ -969,6 +988,45 @@ mod tests {
...
+    async fn read_object_with_user_agent() -> Result {
+        use gaxi::http::reqwest::USER_AGENT;
+
+        let user_agent = "quick_fox_lazy_dog/1.2.3";
+        let server = Server::run();
+        server.expect(
+            Expectation::matching(all_of![
+                request::method_path("GET", "/storage/v1/b/test-bucket/o/test-object"),
+                request::headers(contains(("accept-encoding", "gzip"))),
+                request::headers(contains((USER_AGENT.as_str(), user_agent))),
```
joshuatants: @coryan I wanted to use USER_AGENT here without having to have the storage crate depend direct on `reqwest`. I thought it'd be safer to use the constant from `reqwest`, though this is very minor since `user-agent` is practically a universal string. If I do want to use the constant from `reqwest`, is there a better way than making it visible through `gaxi`?

joshuatants:
And sorry I realise I amended my commit instead of creating a new one, thereby losing your comments from before...

coryan:
> And sorry I realise I amended my commit instead of creating a new one, thereby losing your comments from before...

No worries.

src/storage/src/storage/read_object.rs
coryan: I would not export an extra name from `google-cloud-gax-internal` for a test like this.  You can use `http` which also provides the constants and it is 1.0:

https://docs.rs/http/latest/http/header/constant.USER_AGENT.html

and since you are using `.as_str()` there is no type compatibility to worry about.  And you have to use `.as_str()`, because this is going into `httptest` that does not use `HeaderName` for matching, just string slices.

I am not sure if those are "better", but the exporting the constant is not helping you that much anyway.

coryan (approved):

src/storage/src/storage/read_object.rs
joshuatants: Got it thanks. If I decide it's worth it I'll submit the change in a separate PR.

PR #4846, opened 2026-03-01
refactor(pubsub): prepare to hold EO leases separately

dbolduc:
Part of the work for #3964 

From prototyping, I learned it is best to track the leases for at-least-once and exactly-once messages separately. (vs. in one shared data structure with an enum for each branch).

So simplify `under_lease`, as it now only holds leases for messages with at-least-once delivery semantics.

While we're here, bolster some of the internal documentation.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -60,24 +60,23 @@ pub(super) enum LeaseInfo {
     // TODO(#3964) - support exactly once delivery
 }
 
-impl LeaseInfo {
-    fn receive_time(&self) -> Instant {
-        match self {
-            LeaseInfo::AtLeastOnce(t) => *t,
-        }
-    }
-}
-
 #[derive(Debug)]
 pub(super) struct LeaseState<L>
 where
     L: Leaser + Clone,
 {
-    // A map of ack IDs to the time they were first received.
-    under_lease: HashMap<String, LeaseInfo>,
+    // Ack IDs with at-least-once semantics that are under lease management. The
+    // `Instant` denotes the time they were received.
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -60,24 +60,23 @@ pub(super) enum LeaseInfo {
...
 pub(super) struct LeaseState<L>
 where
     L: Leaser + Clone,
 {
-    // A map of ack IDs to the time they were first received.
-    under_lease: HashMap<String, LeaseInfo>,
+    // Ack IDs with at-least-once semantics that are under lease management. The
+    // `Instant` denotes the time they were received.
+    under_lease: HashMap<String, Instant>,
+    // Ack IDs we need to acknowledge with at-least-once semantics.
```
coryan, line 71: I think you are missing a trick here where you store more information in the `LeaseInfo` type. Or maybe the `String` should be the `enum Action` you define elsewhere and then you can just call methods on that enum.

In any case, this is all a feeling, no reason to block these changes.

dbolduc: I will do that for exactly-once leases. They will have some pending flag in the lease info that we set when we attempt to send an ack rpc. That way we can continue to extend leases while the retry loop is executing.

Extending leases while acks pend didn't seem necessary for at-least-once leases, so I just move those out when the application acks them. We could revisit this decision. At best it is a micro-optimization

dbolduc: The micro-optimization is processing the lease management before acks are sent, instead of after they return, for at-least-once delivery.


But yeah, then the internals are slightly different, for the two delivery types

PR #4849, opened 2026-03-01
fix(pubsub): permanent errors end stream

### LINKED ISSUE #4848: Subscriber streams should not be retried after permanent error
If a stream ends in a permanent error, and the application continues to call `stream.next()`, we will continue to retry the stream.

Our library is more predictable if it returns a single error, then `None` for any subsequent calls to `stream.next()`.

Applications should be able to write code like this where the loop does not continuously error:

```rs
while let Some(r) = stream.next().await {
  match (r) {
    Err(_) => println!("D'oh, forgot to gcloud auth application-default login"); // for example
    Ok(m) => { ... }
  }
}
```

dbolduc:
Fixes #4848 

Track whether the stream has failed with a permanent error. If so, stop trying to recreate the stream, if the application asks for more messages.

This is as Rust-y as I could get it, but I still needed an `unreachable!()` after the state update. :shrug: 

`StreamState` may grow to include `Shutdown` in the future. (Or we may just use `Failed` to mean `Shutdown`, kinda how we use `Unstarted` to also mean `ShouldRetry`). Open to naming suggestions here.

### DIFF
--- a/src/pubsub/src/subscriber/message_stream.rs
+++ b/src/pubsub/src/subscriber/message_stream.rs
@@ -21,7 +21,7 @@ use super::retry_policy::StreamRetryPolicy;
 use super::stream::Stream;
 use super::stub::TonicStreaming as _;
 use super::transport::Transport;
-use crate::google::pubsub::v1::StreamingPullRequest;
+use crate::google::pubsub::v1::{StreamingPullRequest, StreamingPullResponse};
 use crate::model::Message;
 use crate::{Error, Result};
 use gaxi::grpc::from_status::to_gax_error;
@@ -67,7 +67,7 @@ pub struct MessageStream {
     /// of `MessageStream` is blocked on the first message being available.
     ///
     /// [^1]: <https://github.com/hyperium/tonic/issues/515>
-    stream: StreamState,
+    stream: Option<StreamState>,
 
     /// Applications ask for messages one at a time. Individual stream responses
     /// can contain multiple messages. We use `pool` to hold the extra messages
@@ -97,8 +97,9 @@ pub struct MessageStream {
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug)]
 enum StreamState {
-    Unstarted,
+    /// The stream failed with a permanent error. It should not be re-opened.
     Failed,
+    /// The stream is active.
     Active(Stream<Transport>),
 }
 
@@ -136,7 +137,7 @@ impl MessageStream {
         Self {
             inner,
             initial_req,
-            stream: StreamState::Unstarted,
+            stream: None,
             pool: VecDeque::new(),
             message_tx,
             ack_tx,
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -92,6 +92,16 @@ pub struct MessageStream {
...
 
+// We would rather always allocate enough space to hold the stream on the stack
+// than add a layer of indirection by `Box`ing it.
+#[allow(clippy::large_enum_variant)]
+#[derive(Debug)]
+enum StreamState {
+    Unstarted,
+    Failed,
+    Active(Stream<Transport>),
+}
```
coryan, line 104: This could also be a `Option<Option<T>>` or `Option<StreamState>` where StreamState only has `Active(s)` and `Failed` states.

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -203,16 +214,23 @@ impl MessageStream {
     /// Returns a mutable reference to the underlying stream.
     ///
     /// If a stream is not yet open, this method opens the stream.
-    async fn mut_stream(&mut self) -> Result<&mut Stream<Transport>> {
-        if self.stream.is_none() {
+    async fn mut_stream(&mut self) -> Option<Result<&mut Stream<Transport>>> {
```
coryan: Not pretty, I think returning `&mut Stream<Transport>` is causing you grief, maybe you should just have a function that returns `Option<Result< Message Type >`. Or maybe your StreamState should implement `next_message(start: AsyncFn() -> Result<Stream::<Transport>>` and you just call that function on it with:

```
self.stream_state.next_message(async || { Stream::<Transport>::new(self.inner.clone(), self.initial_req.clone()).await  })
```

The lifetimes could get tricky on that, maybe.

src/pubsub/src/subscriber/message_stream.rs
dbolduc: I went with `Option<StreamState>`, because `None` is a perfectly good way to say "there is no stream". Thanks.

It also lets us use a cheeky `?` here: https://github.com/dbolduc/google-cloud-rust/blob/4d618d6f6f19d293d0f94734e8c8c2fa1ace46b9/src/pubsub/src/subscriber/message_stream.rs#L226

The `None` branch of that is unreachable.... but mehhhh I am fine with `?` over a `match` + `unreachable!()`.

src/pubsub/src/subscriber/message_stream.rs
dbolduc: With apologies to the archaeologists, I did some refactoring in this PR.

I took your suggestions and split out (1) opening the stream and (2) receiving the next response from the stream. And also your suggestion on naming from: https://github.com/googleapis/google-cloud-rust/pull/4381#pullrequestreview-3704241449

I think the code reads a lot nicer, so thanks.

PR #4850, opened 2026-03-01
feat(spanner): add type and value definitions

olavloite:
- Adds implementation for Type, TypeCode and Value for Spanner. Value is a transparent wrapper around a protobuf Value. This allows us to replace the underlying protobuf implementation in the future, while retaining the same public API.
- Adds FromValue and ToValue traits for Value.
- Adds standard implementations for FromValue and ToValue for standard Rust types.


src/spanner/Cargo.toml
```diff
@@ -48,4 +51,5 @@ wkt.workspace         = true
 anyhow.workspace            = true
 google-cloud-auth.workspace = true
 spanner-grpc-mock           = { path = "grpc-mock" }
+static_assertions           = "1.1.0"
```
coryan:
```suggestion
static_assertions           = { workspace = true }
```

coryan: Also, this is duplicated in the non-dev dependencies, I missed that in the previous review.

src/spanner/Cargo.toml
```diff
@@ -37,6 +39,7 @@ lazy_static.workspace = true
 mockall               = "0.14.0"
```
coryan: That should be a workspace dependency too.

src/spanner/Cargo.toml
```diff
@@ -37,6 +39,7 @@ lazy_static.workspace = true
...
 prost.workspace       = true
 prost-types.workspace = true
+rust_decimal          = { version = "1.33.1", default-features = false, features = ["std"] }
```
coryan: In general, please add dependencies to the workspace file.  This should be:

```suggestion
rust_decimal          = { workspace = true, default-features = false, features = ["std"] }
```

And we should define the version in the top-level `Cargo.toml` file.  It is easier to upgrade things if we list them in a single place.

src/spanner/src/value.rs
```diff
@@ -0,0 +1,263 @@
...
+
+impl Value {
+    /// Safely reinterprets a reference to the inner protobuf value as a reference to Value.
+    /// Logical safety is guaranteed by #[repr(transparent)].
+    pub(crate) fn from_ref(v: &ProtoValue) -> &Self {
+        // Safety: Value is #[repr(transparent)] wrapper around ProtoValue.
+        // This structure guarantees that Value has the exact same memory layout as ProtoValue.
+        // This is the standard Rust pattern for safe zero-cost newtype references.
+        unsafe { &*(v as *const ProtoValue as *const Value) }
+    }
```
coryan, line 48: Ugh.... I was "this" close from adding a rule to prohibit `unsafe` in our code.

- A move is cheap for strings, numbers, bools, etc. Is the `unsafe` justified vs. moving the values out?

- What will happen when we use Google Protobuf ? The mapping will be something like:

https://docs.rs/protobuf-well-known-types/latest/protobuf_well_known_types/struct.Value.html

Can you still avoid the copying / moves with that API?

If the answers are "no" or even "not sure". I think you should avoid this approach.

src/spanner/src/types.rs
```diff
@@ -0,0 +1,390 @@
...
+/// Returns a `Type` representing `ARRAY<t>` (GoogleSQL) or `t[]` (PostgreSQL).
+pub fn array(element_type: Type) -> Type {
+    let mut t = create_type(TypeCode::Array);
+    t.0.array_element_type = Some(Box::new(element_type.0));
+    t
+}
+
+/// Returns a `Type` representing `UUID` (GoogleSQL) or `uuid` (PostgreSQL).
+pub fn uuid() -> Type {
+    static TYPE_UUID: LazyLock<Type> = LazyLock::new(|| create_type(TypeCode::Uuid));
```
coryan: Why is this one defined inside the function?

src/spanner/src/types.rs
```diff
@@ -0,0 +1,390 @@
...
+    String = 6,
+    Bytes = 7,
+    Array = 8,
+    Struct = 9,
+    Numeric = 10,
+    Json = 11,
+    Proto = 13,
+    Enum = 14,
+    Interval = 16,
+    Uuid = 17,
```
coryan, line 93: This will break if anybody uses an old version of the client library with a database using new datatypes. We designed the enums in sidekick to support unknown types, you may want to use the same techniques, even if you cannot use the type directly.

src/spanner/src/to_value.rs
```diff
@@ -0,0 +1,272 @@
...
+
+impl ToValue for chrono::DateTime<chrono::Utc> {
+    fn to_value(&self) -> Value {
+        Value(ProtoValue {
+            kind: Some(prost_types::value::Kind::StringValue(
+                self.format("%Y-%m-%dT%H:%M:%S.%9fZ").to_string(),
+            )),
+        })
+    }
+}
```
coryan: chrono is at 0.4 and this: https://github.com/chronotope/chrono/issues/1768 

I think you should make `chrono` an optional feature call it: `unstable-chrono`.

src/spanner/src/to_value.rs
```diff
@@ -0,0 +1,272 @@
...
+// limitations under the License.
+
+pub use crate::value::Value;
+use base64::Engine;
+use base64::prelude::BASE64_STANDARD;
+use prost_types::Value as ProtoValue;
+use rust_decimal::Decimal;
+use std::time::SystemTime;
+
+/// A trait for converting a Rust type to a Spanner Value.
```
coryan:
```suggestion
/// Converts Rust types to Spanner [Value].
///
/// # Example
/// ```
/// use google_cloud_spanner::value::Value;
/// use google_cloud_spanner::to_value::ToValue as _;
/// let value: Value = 123_i32.to_value();
/// ```
```

src/spanner/src/from_value.rs
```diff
@@ -0,0 +1,594 @@
...
+    T: FromValue,
+{
+    fn from_value(value: &Value, type_: &Type) -> crate::Result<Self> {
+        match &value.0.kind {
+            Some(prost_types::value::Kind::NullValue(_)) => Ok(None),
+            None => Err(crate::Error::service(
+                Status::default()
+                    .set_code(Code::Internal)
+                    .set_message("unexpected missing value kind"),
+            )),
```
coryan: This is not a service error.... Arguably this is a deserialization error and you could use `Error::deser()`.  Another possibility is to use your own error type, defined in this crate, and return that.

src/spanner/src/from_value.rs
```diff
@@ -0,0 +1,594 @@
...
+use crate::value::Value;
+use base64::Engine;
+use base64::prelude::BASE64_STANDARD;
+use google_cloud_gax::error::rpc::{Code, Status};
+use rust_decimal::Decimal;
+use std::time::SystemTime;
+
+/// A trait for converting a Spanner Value to a Rust type.
+pub trait FromValue: Sized {
+    fn from_value(value: &Value, type_: &Type) -> crate::Result<Self>;
```
coryan: In general, we use `google_cloud_gax::error::Error` for errors involving RPCs, and use dedicated error types for conversion errors with local types.  I think you should use an specific error here, a good starting point  might be:

https://github.com/googleapis/google-cloud-rust/blob/df77ebc841b24f23dc52cd747e07bc82e61975ec/src/wkt/src/timestamp.rs#L86-L96

src/spanner/src/from_value.rs
```diff
@@ -0,0 +1,594 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+pub use crate::types::{Type, TypeCode};
+use crate::value::Value;
+use base64::Engine;
+use base64::prelude::BASE64_STANDARD;
+use google_cloud_gax::error::rpc::{Code, Status};
+use rust_decimal::Decimal;
```
coryan, line 20: You should re-export this type somewhere, e.g. in `types.rs` you would say `pub use rust_decimal::Decimal;`.  That way users of your crate can use the same version of `Decimal` as you are.

src/spanner/src/from_value.rs
```diff
@@ -0,0 +1,594 @@
...
+                    .set_code(Code::InvalidArgument)
+                    .set_message(format!(
+                        "expected TIMESTAMP type code, got {:?}",
+                        type_.code()
+                    )),
+            ));
+        }
+        match &value.0.kind {
+            Some(prost_types::value::Kind::StringValue(s)) => {
+                let dt = chrono::DateTime::parse_from_rfc3339(s).map_err(|e| {
```
coryan: In passing: using `chrono` in the implementation, as you do here, is fine.

src/spanner/src/from_value.rs
```diff
@@ -0,0 +1,594 @@
...
+            _ => Err(crate::Error::service(
+                Status::default()
+                    .set_code(Code::InvalidArgument)
+                    .set_message(format!("expected String (TIMESTAMP), got {:?}", value)),
+            )),
+        }
+    }
+}
+
+impl FromValue for chrono::DateTime<chrono::Utc> {
```
coryan: See my comment in `to_value.rs` about chrono's stability and long-time prospects.

src/spanner/src/from_value.rs
```diff
@@ -0,0 +1,594 @@
...
+            )),
+            _ => Err(crate::Error::service(
+                Status::default()
+                    .set_code(Code::InvalidArgument)
+                    .set_message(format!("expected String (base64), got {:?}", value)),
+            )),
+        }
+    }
+}
+
```
coryan, line 264: When we were discussing escape hatches I thought you would have a helper here, like:


```rust
pub struct RawStringValue {
    pub value: String,
    pub kind: Kind,
}

impl FromValue for RawStringValue {
    ....
}
```

That way applications can get to the raw string if they want, without base64 decoding, or when new types are introduced. And that works whatever the underlying representation happens to be.

src/spanner/src/from_value.rs
```diff
@@ -0,0 +1,594 @@
...
+pub use crate::types::{Type, TypeCode};
+use crate::value::Value;
+use base64::Engine;
+use base64::prelude::BASE64_STANDARD;
+use google_cloud_gax::error::rpc::{Code, Status};
+use rust_decimal::Decimal;
+use std::time::SystemTime;
+
+/// A trait for converting a Spanner Value to a Rust type.
```
coryan:
```suggestion
/// Converts Spanner [Value] to Rust types.
///
/// # Example
/// ```
/// # use google_cloud_spanner::from_value::FromValue as _;
/// use google_cloud_spanner::to_value::ToValue as _;
/// let input = 42_i32;
/// let value = input.to_value();
/// let got = value.from_value::<std::time::SystemTime>();
/// assert!(got.is_err(), "{got:?}");
/// let got: i32 = value.from_value()?;
/// assert_eq!(got, input);
///```
```

src/spanner/src/types.rs
```diff
@@ -0,0 +1,384 @@
...
+static TYPE_PG_OID: LazyLock<Type> = LazyLock::new(|| {
+    let mut t = create_type(TypeCode::Int64);
+    t.0.type_annotation = TypeAnnotationCode::PgOid;
+    t
+});
+
+/// Returns a `Type` representing `INT64` (GoogleSQL) or `bigint` (PostgreSQL).
+pub fn int64() -> Type {
+    TYPE_INT64.clone()
+}
```
coryan, line 164: These standalone functions smell funny.  maybe I do not understand how they will be used. 🤷  Are they intended for customers to use in application development? If so, how?  If not, then these should not be `pub fn`, maybe `pub(crate) fn`?

src/spanner/src/value.rs
olavloite: This implementation was actually created exactly with the knowledge that we might change the internal implementation. This allows us to:
1. Expose our own `Value` struct with a user-friendly API (e.g. with methods like `as_string()` and `try_get_string()`).
2. Define the `ToValue` and `FromValue` traits for our own struct.
3. Very low cost creation. We've seen in other languages that one of the biggest contributors to CPU usage of the client library is iterating through large query results, partly due to a lot of copying and creating of new instances of hand-written wrapper classes/structs.
4. Keep the same public API when we change to Google Protobuf.

Regarding the last point, which I believe is your biggest concern. The answer is 'definitively yes'. It would look (approx) like this:

```rust
use crate::value::Kind;
use protobuf::well_known_types::struct_::Value as ProtoValue;

/// A wrapper around the Google Protobuf Value type.
#[repr(transparent)]
#[derive(Clone, Debug, PartialEq, Default)]
pub struct Value(pub(crate) ProtoValue);

impl Value {
    /// Safely reinterprets a reference to the inner protobuf value as a reference to Value.
    /// Logical safety is guaranteed by #[repr(transparent)].
    pub(crate) fn from_ref(v: &ProtoValue) -> &Self {
        // Safety: Value is #[repr(transparent)] wrapper around ProtoValue.
        // This structure guarantees that Value has the exact same memory layout as ProtoValue.
        // This is the standard Rust pattern for safe zero-cost newtype references.
        unsafe { &*(v as *const ProtoValue as *const Value) }
    }

    /// Returns the kind of the value.
    pub fn kind(&self) -> Kind {
        match &self.0.kind {
            Some(protobuf::well_known_types::struct_::value::Kind::NullValue(_)) => Kind::Null,
            Some(protobuf::well_known_types::struct_::value::Kind::NumberValue(_)) => Kind::Number,
            Some(protobuf::well_known_types::struct_::value::Kind::StringValue(_)) => Kind::String,
            Some(protobuf::well_known_types::struct_::value::Kind::BoolValue(_)) => Kind::Bool,
            Some(protobuf::well_known_types::struct_::value::Kind::StructValue(_)) => Kind::Struct,
            Some(protobuf::well_known_types::struct_::value::Kind::ListValue(_)) => Kind::List,
            _ => Kind::Null,
        }
    }

    /// Returns the underlying string value if the kind is String.
    pub fn try_as_string(&self) -> Option<&str> {
        match &self.0.kind {
            Some(protobuf::well_known_types::struct_::value::Kind::StringValue(s)) => Some(s),
            _ => None,
        }
    }

    /// Returns the underlying string value. Panics if the kind is not String.
    pub fn as_string(&self) -> &str {
        self.try_as_string().expect("value is not a String")
    }

    /// Returns the underlying bool value if the kind is Bool.
    pub fn try_as_bool(&self) -> Option<bool> {
        match &self.0.kind {
            Some(protobuf::well_known_types::struct_::value::Kind::BoolValue(b)) => Some(*b),
            _ => None,
        }
    }

    /// Returns the underlying bool value. Panics if the kind is not Bool.
    pub fn as_bool(&self) -> bool {
        self.try_as_bool().expect("value is not a Bool")
    }

    /// Returns the underlying number value if the kind is Number.
    pub fn try_as_number(&self) -> Option<f64> {
        match &self.0.kind {
            Some(protobuf::well_known_types::struct_::value::Kind::NumberValue(n)) => Some(*n),
            _ => None,
        }
    }

    /// Returns the underlying number value. Panics if the kind is not Number.
    pub fn as_number(&self) -> f64 {
        self.try_as_number().expect("value is not a Number")
    }

    /// Returns the underlying struct value if the kind is Struct.
    pub fn try_as_struct(&self) -> Option<&Struct> {
        match &self.0.kind {
            Some(protobuf::well_known_types::struct_::value::Kind::StructValue(s)) => {
                Some(Struct::from_ref(s))
            }
            _ => None,
        }
    }

    /// Returns the underlying struct value. Panics if the kind is not Struct.
    pub fn as_struct(&self) -> &Struct {
        self.try_as_struct().expect("value is not a Struct")
    }

    /// Returns the underlying list value if the kind is List.
    pub fn try_as_list(&self) -> Option<&ListValue> {
        match &self.0.kind {
            Some(protobuf::well_known_types::struct_::value::Kind::ListValue(l)) => {
                Some(ListValue::from_ref(l))
            }
            _ => None,
        }
    }

    /// Returns the underlying list value. Panics if the kind is not List.
    pub fn as_list(&self) -> &ListValue {
        self.try_as_list().expect("value is not a List")
    }

    /// Returns true if the value is Null.
    pub fn is_null(&self) -> bool {
        matches!(self.kind(), Kind::Null)
    }
}

use protobuf::well_known_types::struct_::Struct as ProtoStruct;

#[repr(transparent)]
#[derive(Clone, Debug, PartialEq, Default)]
pub struct Struct(pub(crate) ProtoStruct);

impl Struct {
    pub(crate) fn from_ref(v: &ProtoStruct) -> &Self {
        unsafe { &*(v as *const ProtoStruct as *const Struct) }
    }

    pub fn get(&self, key: &str) -> Option<&Value> {
        self.0.fields.get(key).map(Value::from_ref)
    }
}

use protobuf::well_known_types::struct_::ListValue as ProtoListValue;

#[repr(transparent)]
#[derive(Clone, Debug, PartialEq, Default)]
pub struct ListValue(pub(crate) ProtoListValue);

impl ListValue {
    pub(crate) fn from_ref(v: &ProtoListValue) -> &Self {
        unsafe { &*(v as *const ProtoListValue as *const ListValue) }
    }

    pub fn values(&self) -> impl Iterator<Item = &Value> {
        self.0.values.iter().map(Value::from_ref)
    }
}

```

src/spanner/src/types.rs
olavloite: I actually don't think it will. This is not the full list of type codes that the library defines. This list is used in the macro here: https://github.com/olavloite/google-cloud-rust/blob/b91110b26ba3c938c54f481db3af470c9f44a4b0/src/spanner/src/types.rs#L24

```rust
macro_rules! define_type_code {
    ($($variant:ident = $val:expr),* $(,)?) => {
        /// Spanner type code.
        #[derive(Clone, Debug, PartialEq, Copy, Default)]
        #[repr(i32)]
        pub enum TypeCode {
            #[default]
            Unspecified = 0,
            $($variant = $val),*,
            Unknown(i32),
        }
```

The static list of type codes will be augmented with `Unspecified` and `Unknown`. The latter also stores the original (unknown) type code.

I added an additional test case for this:

```rust
    #[test]
    fn test_unknown_type_code_from_generated() {
        use crate::generated::gapic_dataplane::model::type_code::UnknownValue;
        use wkt::internal::UnknownEnumValue;

        let i = 999;
        let unknown = UnknownValue(UnknownEnumValue::Integer(i));
        let generated = crate::generated::gapic_dataplane::model::TypeCode::UnknownValue(unknown);
        let code: TypeCode = generated.into();
        assert_eq!(code, TypeCode::Unknown(i));
    }
```

src/spanner/src/types.rs
olavloite: No good reason, fixed.

src/spanner/Cargo.toml
olavloite: Done

src/spanner/Cargo.toml
olavloite: Done

src/spanner/Cargo.toml
olavloite: `static_assertions` is now `static_assertions.workspace = true` and only present in `dev-dependencies`.

src/spanner/src/to_value.rs
olavloite: Thanks. I've removed all references to chrono from the public API and replaced it with `std::time`. We can always consider adding it as an optional feature in the future.

src/spanner/src/to_value.rs
olavloite: Fixed

src/spanner/src/from_value.rs
olavloite: I've added a new `ConvertError` for this and refactored `FromValue` to use this.

src/spanner/src/from_value.rs
olavloite: Thanks for the pointer. I've added a new `ConvertError` implementation along the same lines, and refactored `FromValue` to use `ConvertError` instead.

src/spanner/src/from_value.rs
olavloite: Done

src/spanner/src/from_value.rs
olavloite: (not relevant in this PR anymore, as I've removed chrono based on the comments above)

src/spanner/src/from_value.rs
olavloite: ack

src/spanner/src/from_value.rs
olavloite: See my comment above on the current `Value` struct. Getting the underlying value with the current implementation would look (approx) like this:

```rust
let value: Value = result_set.get("my_column");
// If I know that the underlying value is a non-null string
let s = value.as_string();
// If the underlying value could be a string, null or even something else
let opt_s = value.try_as_string();
```

src/spanner/src/from_value.rs
olavloite: Fixed

src/spanner/src/types.rs
olavloite: They are intended for customers to use in application development. Spanner supports query parameters, and a query parameter can optionally include an explicit type definition for the parameter. In most cases, an application developer should just include a parameter value and let Spanner infer the type from the SQL string. However, in some cases Spanner is unable to infer the type, and the application developer must explicitly include the type of the query parameter.

Examples:

```sql
create table my_table (id int64 primary key, value string(max));

-- The following statement does not require the application to include
-- the type of each parameter, as Spanner can infer it from the SQL string.
insert into my_table (id, value) values (@id, @value);

-- This statement requires the application to include the type of the
-- parameter, as Spanner cannot infer it from the SQL string.
select @greeting;
```

Side note: We could also infer the parameter type from the Rust type that the customer uses in the application. This is a strategy that is (partly) used in some other client libraries. That however has proven to have some unwanted consequences, which has forced us to introduce workarounds for it:
1. Spanner is 'picky' about the type if you explicitly set it. That is: If a parameter includes a type code like `STRING`, but the parameter is bound to a column that has type `BYTES`, then Spanner will reject the statement. However, if the same query parameter is just sent with a string value without an explicit type code, then Spanner will accept it, as long as the string value is a valid base64 string. If the client automatically sets the type code to `STRING` when the application uses for example a `String` for a query parameter, it gets hard for the application to just use an existing base64 string for a query parameter that is bound to a `BYTES` column.
2. The introduction of new types in Spanner can break existing type mappings: When Spanner introduced support for the data type `UUID`, it caused problems in client libraries that already accepted `Uuid` instances/structs as a query parameter value and explicitly set the type code to `STRING` for these query parameters.

We would therefore like to be explicit about query parameter types in the Rust client:
1. The default is to send query parameters 'untyped' and let Spanner infer the type.
2. The client will only include a type with a query parameter if the application explicitly sets it. For this, these helper methods offer a simple way of getting an instance of a well-known Spanner type.

olavloite:
> Thanks for the well thought out answers. Sorry you had to write them.

Nothing to say 'sorry' for. Please keep challenging me.

PR #4851, opened 2026-03-01
feat(spanner): add DatabaseClient

olavloite:
Adds a DatabaseClient struct and a builder to create it.

A DatabaseClient creates a single multiplexed session on Spanner and uses this session for all data operations.

### DIFF
--- a/src/spanner/src/client.rs
+++ b/src/spanner/src/client.rs
@@ -63,11 +63,7 @@ impl Spanner {
     /// Returns a new [DatabaseClientBuilder](crate::database_client::DatabaseClientBuilder) for
     /// interacting with a specific database.
     ///
-    /// The returned `DatabaseClient` is intended to be a long-lived object and should be reused
-    /// for all operations on the database.
-    ///
     /// # Example
-    ///
     /// ```
     /// # use google_cloud_spanner::client::Spanner;
     /// # async fn sample() -> anyhow::Result<()> {
@@ -76,9 +72,12 @@ impl Spanner {
     ///         .database_client("projects/my-project/instances/my-instance/databases/my-db")
     ///         .build()
     ///         .await?;
-    ///     Ok(())
+    ///     # Ok(())
     /// # }
     /// ```
+    ///
+    /// The returned `DatabaseClient` is intended to be a long-lived object and should be reused
+    /// for all operations on the database.
     pub fn database_client(
         &self,
         database: impl Into<String>,
--- a/src/spanner/src/database_client.rs
+++ b/src/spanner/src/database_client.rs
@@ -19,16 +19,7 @@ use std::sync::Arc;
 /// A client for interacting with a specific Spanner database.
 ///
 /// `DatabaseClient` provides methods to execute transactions and queries.
-/// It holds a single multiplexed session for the database.
-///
-/// A `DatabaseClient` is intended to be a long-lived object, and normally an
-/// application will have a single `DatabaseClient` per database. The client is
-/// thread-safe and should be reused for all operations on the database.
-///
-/// Cloning a `DatabaseClient` is cheap, as it shares the underlying session and channel.
-///
 /// # Example
-///
 /// ```
 /// # use google_cloud_spanner::client::Spanner;
 /// # async fn sample() -> anyhow::Result<()> {
@@ -37,9 +28,18 @@ use std::sync::Arc;
 ///         .database_client("projects/my-project/instances/my-instance/databases/my-db")
 ///         .build()
 ///         .await?;
-///     Ok(())
+///     # Ok(())
 /// # }
 /// ```
+///
+/// `DatabaseClient` provides methods to execute transactions and queries.
+/// It holds a single multiplexed session for the database.
+///
+/// A `DatabaseClient` is intended to be a long-lived object, and normally an
+/// application will have a single `DatabaseClient` per database. The client is
+/// thread-safe and should be reused for all operations on the database.
+///
+/// Cloning a `DatabaseClient` is cheap, as it shares the underlying session and channel.
 #[derive(Clone, Debug)]
 pub struct DatabaseClient {
     #[allow(dead_code)]
@@ -68,12 +68,7 @@ impl DatabaseClientBuilder {
 
     /// Sets the database role for the client.
     ///
-    /// Database roles are used for Fine-Grained Access Control (FGAC).
-    /// You can assign a database role to a session, and that role determines the permissions for that session.
-    /// For more information, see [Access with FGAC](https://docs.cloud.google.com/spanner/docs/access-with-fgac).
-    ///
     /// # Example
-    ///
     /// ```
     /// # use google_cloud_spanner::client::Spanner;
     /// # async fn sample() -> anyhow::Result<()> {
@@ -83,9 +78,13 @@ impl DatabaseClientBuilder {
     ///         .with_database_role("my-role")
     ///         .build()
     ///         .await?;
-    ///     Ok(())
+    ///     # Ok(())
     /// # }
     /// ```
+    ///
+    /// Database roles are used for Fine-Grained Access Control (FGAC).
+    /// You can assign a database role to a session, and that role determines the permissions for that session.
+    /// For more information, see [Access with FGAC](https://docs.cloud.google.com/spanner/docs/access-with-fgac).
     pub fn with_database_role(mut self, role: impl Into<String>) -> Self {
         self.database_role = Some(role.into());
         self
@@ -95,7 +94,6 @@ impl DatabaseClientBuilder {
     /// session for the client.
     ///
     /// # Example
-    ///
     /// ```
     /// # use google_cloud_spanner::client::Spanner;
     /// # use google_cloud_gax::options::RequestOptions;
@@ -109,7 +107,7 @@ impl DatabaseClientBuilder {
     ///         .with_request_options(options)
     ///         .build()
     ///         .await?;
-    ///     Ok(())
+    ///     # Ok(())
     /// # }
     /// ```
     pub fn with_request_options(mut self, options: crate::RequestOptions) -> Self {


src/spanner/src/client.rs
```diff
@@ -161,6 +161,18 @@ impl Spanner {
             .send()
             .await
     }
+
+    /// Returns a new [DatabaseClientBuilder](crate::database_client::DatabaseClientBuilder) for
+    /// interacting with a specific database.
+    ///
```
coryan:
```suggestion
    ///
    /// # Example
    /// ```
    /// # use google_cloud_spanner::client::Client;
    /// async fn sample(client: &Client, project_id: &str, instance_id: &str, id: &str) -> anyhow::Result<()> {
    ///     let database = client.database_client(
    ///         format!("projects/{project_id}/instances/{instance_id}/databases/{id}").send().await?;
    ///     // use `database` to execute queries, mutations, DML statements, etc. 
    /// }
    /// ```
```

src/spanner/src/database_client.rs
```diff
@@ -0,0 +1,169 @@
...
+impl DatabaseClientBuilder {
+    pub(crate) fn new(spanner: Spanner, database_name: String) -> Self {
+        Self {
+            spanner,
+            database_name,
+            database_role: None,
+        }
+    }
+
+    /// Sets the database role for the client.
```
coryan, line 69: Needs an example.

Also, should link some documentation in docs.cloud.google.com about what is a database role. Ideally something that tells me why I would want to set one.

src/spanner/src/database_client.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub fn database_role(mut self, role: impl Into<String>) -> Self {
+        self.database_role = Some(role.into());
+        self
+    }
+
+    /// Builds the [DatabaseClient] and creates a multiplexed session on Spanner.
```
coryan: I am not sure customers know, or care, about what is a "multiplexed session on Spanner".

src/spanner/src/database_client.rs
```diff
@@ -0,0 +1,169 @@
...
+    pub async fn build(self) -> crate::Result<DatabaseClient> {
+        let mut request = crate::model::CreateSessionRequest::new();
+        request.database = self.database_name;
+
+        let mut session_template = crate::model::Session::new();
+        session_template.multiplexed = true;
+        if let Some(role) = &self.database_role {
+            session_template.creator_role = role.clone();
+        }
+        request.session = Some(session_template);
```
coryan: I think this should work?

```suggestion
        let request = crate::model::CreateSessionRequest::new()
            .set_database(self.database_name)
            .set_session(crate::model::Session::new()
                .set_multiplexed(true)
                .set_or_clear_database_role(self.database_role)
            );
```

src/spanner/src/database_client.rs
```diff
@@ -0,0 +1,169 @@
...
+
+        let session = self
+            .spanner
+            .create_session(request, crate::RequestOptions::default())
```
coryan: Eventually you would want to initialize the request options in `DatabaseClientBuilder`?

src/spanner/src/database_client.rs
```diff
@@ -0,0 +1,169 @@
...
+                spanner_grpc_mock::google::spanner::v1::Session {
+                    name: "projects/test-project/instances/test-instance/databases/test-db/sessions/123".to_string(),
+                    multiplexed: true,
+                    creator_role: "test-role".to_string(),
+                    ..Default::default()
+                },
+            ))
+        });
+
+        let (address, _server) = start("0.0.0.0:0", mock)
```
coryan, line 172: These tests are good, but soon you will want to create integration tests against production?  We should talk about where those live, what project they use, and how we can keep them from being flaky.

src/spanner/src/lib.rs
```diff
@@ -30,6 +30,7 @@ pub(crate) use google_cloud_gax::response::Response;
 
 pub mod client;
 pub mod builder {}
+pub mod database_client;
```
coryan: You are adding a lot of public modules. Maybe consider narrowing the public APIs if you can. For example, you could make this module private, and re-export only the types you want in the modules you want them, e.g. `DatabaseClient` could be part of `client` and `DatabaseClientBuilder` part of `builder`.

src/spanner/src/database_client.rs
```diff
@@ -0,0 +1,169 @@
...
+        request.session = Some(session_template);
+
+        let session = self
+            .spanner
+            .create_session(request, crate::RequestOptions::default())
+            .await?;
+
+        Ok(DatabaseClient {
+            spanner: self.spanner,
+            session: std::sync::Arc::new(session),
```
coryan: Over time we have learnt that it is easier to use say `use std::sync::Arc` at the top-level, even if the name is used a couple of times.

src/spanner/src/database_client.rs
```diff
@@ -0,0 +1,169 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::client::Spanner;
+use crate::model::Session;
+
+/// A client for interacting with a specific Spanner database.
+///
```
coryan, line 20: If you can, add an example.   It forces you to think about the public API.

src/spanner/src/database_client.rs
olavloite: If possible, I would prefer to start with adding integration tests against the [Spanner emulator](https://docs.cloud.google.com/spanner/docs/emulator). The emulator is a lot faster for typical test workloads. That is, this type of workloads:

1. Create a database with one or more tables
2. Execute one or a few queries/transactions on that database
3. Throw away and repeat for a different use case

Especially creating a database and tables can be a relatively expensive operation on Spanner.

The Emulator uses the same gRPC API as real Spanner. The main difference from a user perspective is that the Emulator uses plain text communication and no credentials.

Then later modify the test setup to run the same tests on both the Emulator and a real Spanner instance.

coryan: > If possible, I would prefer to start with adding integration tests against the [Spanner emulator](https://docs.cloud.google.com/spanner/docs/emulator). 

SGTM, with some conditions to keep the impact on everybody else's workflows small.  Something to discuss over a document or GVC. 

> Then later modify the test setup to run the same tests on both the Emulator and a real Spanner instance.

We did that for C++ too, for similar reasons.  We ran one build against production, and used multiple instances (picking one at random) to minimize flakes on concurrent builds.

src/spanner/src/client.rs

src/spanner/src/database_client.rs
olavloite: Done

src/spanner/src/database_client.rs
olavloite: It depends on whether it is a completely new customer, or one who is already using Spanner with a different client.

Session management has long been a complex topic for Spanner, as Spanner for a long time had a model with 'regular' sessions, where one session could execute at most one transaction at a time. This meant that our client libraries contained a session pool that needed to be configured by the customer application, and that the customer application had to make sure that they always returned sessions to the pool (in most cases, this means closing some type of resource, or similar, depending on the programming language). Failing to do so, would cause a session to be leaked, which again could eventually lead to the pool not having any sessions available anymore. We have recently moved to 'multiplexed sessions', where one session can execute any number of transactions in parallel. This has completely removed the need for a session pool and simplified the clients that use this significantly.

We therefore explicitly call this out to clarify that:
1. There is no need to configure a session pool.
2. There is no risk of session leaks.

(I've reworded the comment a bit, though)

src/spanner/src/database_client.rs
olavloite: Done (or at least something very similar)

src/spanner/src/client.rs
```diff
@@ -58,6 +60,32 @@ impl Spanner {
...
+    /// ```
+    /// # use google_cloud_spanner::client::Spanner;
+    /// # async fn sample() -> anyhow::Result<()> {
+    ///     let spanner = Spanner::builder().build().await?;
+    ///     let database_client = spanner
+    ///         .database_client("projects/my-project/instances/my-instance/databases/my-db")
+    ///         .build()
+    ///         .await?;
+    ///     Ok(())
```
coryan: nit: hide the `Ok(())` to match the hidden `# async fn sample()`

```suggestion
    /// #   Ok(())
```

src/spanner/src/database_client.rs
```diff
@@ -0,0 +1,203 @@
...
+/// A `DatabaseClient` is intended to be a long-lived object, and normally an
+/// application will have a single `DatabaseClient` per database. The client is
+/// thread-safe and should be reused for all operations on the database.
+///
+/// Cloning a `DatabaseClient` is cheap, as it shares the underlying session and channel.
+///
+/// # Example
+///
+/// ```
+/// # use google_cloud_spanner::client::Spanner;
```
coryan, line 24: nit: we prefer to put the example block right after the one-line description. When using IDEs, the pop-ups / hovers for each function type have limited space and it is better to make the example more visible.

src/spanner/src/database_client.rs
olavloite: Yeah, that is a good point. I added an option to `DatabaseClientBuilder` for that now.

src/spanner/src/lib.rs
olavloite: Done

src/spanner/src/database_client.rs
olavloite: Done

src/spanner/src/database_client.rs
olavloite: Done

src/spanner/src/client.rs

src/spanner/src/database_client.rs
olavloite: Done

PR #4856, opened 2026-03-02
feat(spanner): add mappings for server-streaming RPCs

olavloite:
Adds mappings and implementations for the server-streaming RPCs ExecuteStreamingSql, StreamingRead and BatchWrite of Spanner.

### DIFF
--- a/src/spanner/src/client.rs
+++ b/src/spanner/src/client.rs
@@ -16,6 +16,8 @@ use crate::generated::gapic_dataplane::client::Spanner as GapicSpanner;
 use crate::server_streaming::builder;
 use gaxi::options::{ClientConfig, Credentials};
 
+pub use crate::database_client::DatabaseClient;
+
 /// A client for the [Spanner] API.
 ///
 /// Use this client to interact with the Spanner service.
@@ -59,6 +61,31 @@ pub type ClientBuilder = google_cloud_gax::client_builder::ClientBuilder<Factory
 impl Spanner {
     pub fn builder() -> ClientBuilder {
         google_cloud_gax::client_builder::internal::new_builder(Factory)
+    }
+
+    /// Returns a new [DatabaseClientBuilder](crate::database_client::DatabaseClientBuilder) for
+    /// interacting with a specific database.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_spanner::client::Spanner;
+    /// # async fn sample() -> anyhow::Result<()> {
+    ///     let spanner = Spanner::builder().build().await?;
+    ///     let database_client = spanner
+    ///         .database_client("projects/my-project/instances/my-instance/databases/my-db")
+    ///         .build()
+    ///         .await?;
+    ///     # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// The returned `DatabaseClient` is intended to be a long-lived object and should be reused
+    /// for all operations on the database.
+    pub fn database_client(
+        &self,
+        database: impl Into<String>,
+    ) -> crate::builder::DatabaseClientBuilder {
+        crate::builder::DatabaseClientBuilder::new(self.clone(), database.into())
     }
 
     /// Creates a new client from the provided stub.
--- a/src/spanner/src/database_client.rs
+++ b/src/spanner/src/database_client.rs
@@ -1,0 +1,280 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::client::Spanner;
+use crate::model::Session;
+use std::sync::Arc;
+
+/// A client for interacting with a specific Spanner database.
+///
+/// `DatabaseClient` provides methods to execute transactions and queries.
+/// # Example
+/// ```
+/// # use google_cloud_spanner::client::Spanner;
+/// # async fn sample() -> anyhow::Result<()> {
+///     let spanner = Spanner::builder().build().await?;
+///     let database_client = spanner
+///         .database_client("projects/my-project/instances/my-instance/databases/my-db")
+///         .build()
+///         .await?;
+///     # Ok(())
+/// # }
+/// ```
+///
+/// `DatabaseClient` provides methods to execute transactions and queries.
+/// It holds a single multiplexed session for the database.
+///
+/// A `DatabaseClient` is intended to be a long-lived object, and normally an
+/// application will have a single `DatabaseClient` per database. The client is
+/// thread-safe and should be reused for all operations on the database.
+///
+/// Cloning a `DatabaseClient` is cheap, as it shares the underlying session and channel.
+#[derive(Clone, Debug)]
+pub struct DatabaseClient {
+    #[allow(dead_code)]
+    pub(crate) spanner: Spanner,
+    #[allow(dead_code)]
+    pub(crate) session: Arc<Session>,
+}
+
+/// A builder for [DatabaseClient].
+pub struct DatabaseClientBuilder {
+    spanner: Spanner,
+    database_name: String,
+    database_role: Option<String>,
+    options: Option<crate::RequestOptions>,
+}
+
+impl DatabaseClientBuilder {
+    pub(crate) fn new(spanner: Spanner, database_name: String) -> Self {
+        Self {
+            spanner,
+            database_name,
+            database_role: None,
+            options: None,
+        }
+    }
+
+    /// Sets the database role for the client.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_spanner::client::Spanner;
+    /// # async fn sample() -> anyhow::Result<()> {
+    ///     let spanner = Spanner::builder().build().await?;
+    ///     let database_client = spanner
+    ///         .database_client("projects/my-project/instances/my-instance/databases/my-db")
+    ///         .with_database_role("my-role")
+    ///         .build()
+    ///         .await?;
+    ///     # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// Database roles are used for Fine-Grained Access Control (FGAC).
+    /// You can assign a database role to a session, and that role determines the permissions for that session.
+    /// For more information, see [Access with FGAC](https://docs.cloud.google.com/spanner/docs/access-with-fgac).
+    pub fn with_database_role(mut self, role: impl Into<String>) -> Self {
+        self.database_role = Some(role.into());
+        self
+    }
+
+    /// Sets the request options that will be used when creating the multiplexed
+    /// session for the client.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_spanner::client::Spanner;
+    /// # use google_cloud_gax::options::RequestOptions;
+    /// # use std::time::Duration;
+    /// # async fn sample() -> anyhow::Result<()> {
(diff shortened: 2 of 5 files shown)


src/spanner/Cargo.toml
```diff
@@ -28,24 +28,32 @@ rust-version.workspace = true
...
+static_assertions = "1.1.0"
+tracing.workspace = true
+wkt.workspace = true
 
 [dev-dependencies]
-anyhow.workspace            = true
+anyhow.workspace = true
+gaxi = { package = "google-cloud-gax-internal", workspace = true, features = [
+  "google_cloud_unstable_grpc_server_streaming",
+] }
```
coryan: You shouldn't dup a dev dependency if it is already a regular dep.

src/spanner/Cargo.toml
```diff
@@ -28,24 +28,32 @@ rust-version.workspace = true
...
+  "_internal-grpc-client",
+  "google_cloud_unstable_grpc_server_streaming",
+] }
+google-cloud-gax = { workspace = true }
+google-cloud-rpc = { workspace = true }
+http.workspace = true
 lazy_static.workspace = true
-mockall               = "0.14.0"
-prost.workspace       = true
+mockall = "0.14.0"
```
coryan: This should be a dev dependency *only* and should use `mockall.workspace = true` or `mockall = {workspace = true}`

src/spanner/Cargo.toml
```diff
@@ -28,24 +28,32 @@ rust-version.workspace = true
...
-serde.workspace       = true
-serde_json.workspace  = true
-serde_with.workspace  = true
-static_assertions     = "1.1.0"
-tracing.workspace     = true
-wkt.workspace         = true
+serde.workspace = true
+serde_json.workspace = true
+serde_with.workspace = true
+static_assertions = "1.1.0"
```
coryan: as with `lazy_static`

src/spanner/src/server_streaming.rs
```diff
@@ -0,0 +1,2 @@
+pub mod builder;
```
coryan: Even for two lines, you need to copyright boilerplate.

src/spanner/src/client.rs
```diff
@@ -22,7 +25,8 @@ use gaxi::options::{ClientConfig, Credentials};
 /// [Spanner]: https://docs.cloud.google.com/spanner/docs
 #[derive(Clone, Debug)]
 pub struct Spanner {
-    inner: GapicSpanner,
+    inner: Arc<GapicSpanner>,
```
coryan: YOu do not need an `Arc<>` for this, cloning a `GapicSpanner` is as cheap as cloning an `Arc<>`.

src/spanner/src/client.rs
```diff
@@ -12,8 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use crate::server_streaming::builder;
+
```
coryan: nit: remove blank line.

src/spanner/src/server_streaming/stream.rs
```diff
@@ -0,0 +1,77 @@
...
+        self.inner
+            .message()
+            .await
+            .map_err(|e| {
+                crate::Error::service(
+                    google_cloud_gax::error::rpc::Status::default()
+                        .set_code(e.code() as i32)
+                        .set_message(e.message()),
+                )
+            })
```
coryan: You should  not assume the error is a service error.  Tonic can return errors that are about failure to connect.  There are helper functions to convert `tonic::Status` to the closest gax error:

https://github.com/googleapis/google-cloud-rust/blob/724c7eba9ccde3ed7e00c1e1a62022b11c8dd84b/src/gax-internal/src/grpc/from_status.rs#L30

src/spanner/src/server_streaming/stream.rs
```diff
@@ -0,0 +1,77 @@
...
+    /// Fetches the next `PartialResultSet` from the stream.
+    ///
+    /// Returns `Some(Ok(PartialResultSet))` when a message is successfully received,
+    /// `None` when the stream concludes naturally, or `Some(Err(_))` on RPC errors.
+    pub async fn next_message(&mut self) -> Option<crate::Result<PartialResultSet>> {
+        self.inner
+            .message()
+            .await
+            .map_err(|e| {
+                crate::Error::service(
```
coryan: Ditto: https://github.com/googleapis/google-cloud-rust/blob/724c7eba9ccde3ed7e00c1e1a62022b11c8dd84b/src/gax-internal/src/grpc/from_status.rs#L30

src/gax-internal/src/grpc.rs
```diff
@@ -243,7 +243,7 @@ impl Client {
     }
 
     /// Opens a server stream with detailed status.
-    #[cfg(google_cloud_unstable_grpc_server_streaming)]
+    #[cfg(feature = "google_cloud_unstable_grpc_server_streaming")]
```
coryan: This you can simply remove as the function is `pub(crate)`...

src/gax-internal/src/grpc.rs
```diff
@@ -216,7 +216,7 @@ impl Client {
     }
 
     /// Opens a server stream.
-    #[cfg(google_cloud_unstable_grpc_server_streaming)]
+    #[cfg(feature = "google_cloud_unstable_grpc_server_streaming")]
```
coryan: Applications can also turn on features...  I think it is fine, but let me bring in @dbolduc for a second opinion.  Darren, if we add this in `gax-internal`, and it is only used in Spanner, and Spanner is not released, and we remember to bump versions on breaking changes to gaxi we are fine, right?

src/gax-internal/Cargo.toml
```diff
@@ -96,6 +96,8 @@ _internal-http-stream = ["_internal-http-client", "reqwest?/stream"]
 _internal-common = ["dep:google-cloud-auth", "dep:google-cloud-gax", "dep:percent-encoding", "dep:thiserror"]
 # Enables the default rustls crypto provider for whatever client(s) are enabled.
 _default-rustls-provider = ["google-cloud-auth?/default-rustls-provider", "tonic?/tls-aws-lc"]
+# Enables the unstable server streaming functionality in the gRPC client.
+google_cloud_unstable_grpc_server_streaming = []
```
coryan: rename to `_internal-grpc_server_streaming` the leading underscore makes the feature undocumented (and thus not supposed to be used by application developers).

src/spanner/src/server_streaming/builder.rs
```diff
@@ -0,0 +1,225 @@
...
+    let path = http::uri::PathAndQuery::from_static(path_str);
+    let x_goog_request_params = [session.map(|v| format!("session={v}"))]
+        .into_iter()
+        .flatten()
+        .fold(String::new(), |b, p| b + "&" + &p);
+
+    let ac = gaxi::api_header::XGoogApiClient {
+        name: env!("CARGO_PKG_NAME"),
+        version: env!("CARGO_PKG_VERSION"),
+        library_type: gaxi::api_header::GAPIC,
```
coryan: This should be GCCL because this is a hand-written client:

https://github.com/googleapis/google-cloud-rust/blob/724c7eba9ccde3ed7e00c1e1a62022b11c8dd84b/src/gax-internal/src/api_header.rs#L28

You may also consider putting this in a lazy initialized global.

src/spanner/src/server_streaming/builder.rs
```diff
@@ -0,0 +1,225 @@
...
+    let extensions = {
+        let mut e = Extensions::new();
+        e.insert(GrpcMethod::new("google.spanner.v1.Spanner", method_name));
+        e
+    };
+    let path = http::uri::PathAndQuery::from_static(path_str);
+    let x_goog_request_params = [session.map(|v| format!("session={v}"))]
+        .into_iter()
+        .flatten()
+        .fold(String::new(), |b, p| b + "&" + &p);
```
coryan: Since there is only one element you can simplify this to:


```suggestion
    let x_goog_request_params = session.map(|v| format!("session={v}")).unwrap_or_default();
```

Aside: without sessions this will have no `x-goog-request-params` header, which seems wrong because those are used for all kinds of routing in the CFE.    But I see no new annotations in the proto:

https://github.com/googleapis/googleapis/blob/c83d354f79b02e3dc98f22248e22851fcf9b961d/google/spanner/v1/spanner.proto#L143-L148

🤷

src/spanner/src/server_streaming/builder.rs
```diff
@@ -1,3 +1,4 @@
+#![allow(dead_code)]
```
coryan: We would typically put these after the boiler plate.

src/spanner/src/server_streaming/builder.rs
olavloite: Removed the mapping and flattening, and made session a required argument. `session` is indeed a required field for all these RPCs.

src/spanner/src/server_streaming/builder.rs
olavloite: Changed to "gccl" and added a lazy initialized field for the header.

src/gax-internal/Cargo.toml
olavloite: Renamed

src/gax-internal/src/grpc.rs
olavloite: Removed

src/spanner/src/server_streaming/stream.rs
olavloite: Thanks, adjusted error mapping to use the `to_gax_error` method.

src/spanner/src/server_streaming/stream.rs
olavloite: Thanks, adjusted error mapping to use the to_gax_error method.

src/spanner/src/client.rs

src/spanner/src/client.rs
olavloite: Done (sorry, was brought back in when porting code from the prototype into this PR)

src/spanner/src/server_streaming.rs
olavloite: Added. Also made these modules `pub(crate)`. They are not intended for public usage.

src/gax-internal/src/grpc.rs
dbolduc: Yes, we would be fine.

But why break from the existing `--cfg` flag pattern? For convenience?

We also need to remember to exercise this stuff in the CI, regardless of the approach.

coryan: I think the `--cfg` approach would quickly become a hassle in Spanner, almost every method would call one of these.... 

I am pretty sure the API is not going to change. I would like to see an integration test making a server-side streaming request against to be sure, and it seems like we are not too far away from that.

coryan (approved):
A small nit, please fix and then merge. @dbolduc is right, we should fix the CI here:

https://github.com/googleapis/google-cloud-rust/blob/a50ea91dd5cab4cf1b2f1a3216c38bbd0cc51edc/.gcb/scripts/features-gax-internal.sh#L26

That can wait for a follow up PR if you want.

src/spanner/src/server_streaming/builder.rs
olavloite: Done

src/gax-internal/src/grpc.rs
olavloite: FWIW: I did try this end-to-end with a real Spanner database with the prototype that we created earlier, and that used the same API. So I feel confident that the API works, at least in the happy flow.

I agree that we should quickly work towards a few integration test cases that shows that things really work end-to-end.

.gcb/scripts/features-gax-internal.sh
```diff
@@ -23,7 +23,7 @@ cargo clean
 echo "==== google-cloud-gax-internal ===="
 cargo test --package google-cloud-gax-internal --no-default-features
 cargo doc  --package google-cloud-gax-internal --no-default-features --no-deps
-for feature in _internal-common _internal-http-client _internal-grpc-client _internal-http-multipart _internal-http-stream; do
```
coryan, line 26: Thanks

PR #4864, opened 2026-03-02
cleanup(pubsub): simplify handler matching in tests

dbolduc:
Motivated by #3964 

When we add an exactly-once branch to the handler enum, these tests will need an update. So do that now.

Also, :monkey:. You cannot format in an `expect(...)`. So it would have been printing `"expected message {i}/6"` instead of `"expected message 1/6"`. Fix that too.

### DIFF
--- a/src/pubsub/src/subscriber/handler.rs
+++ b/src/pubsub/src/subscriber/handler.rs
@@ -107,6 +107,13 @@ impl Handler {
             Handler::AtLeastOnce(h) => h.ack(),
         }
     }
+
+    #[cfg(test)]
+    pub(crate) fn ack_id(&self) -> &str {
+        match self {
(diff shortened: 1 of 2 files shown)


src/pubsub/src/subscriber/handler.rs
```diff
@@ -107,6 +107,13 @@ impl Handler {
             Handler::AtLeastOnce(h) => h.ack(),
         }
     }
+
+    #[cfg(test)]
+    pub(crate) fn ack_id(&self) -> &str {
```
coryan, line 112: Wouldn't `pub(crate)` be enough, ah... the dead code warnings?

dbolduc: Exactly.

PR #4866, opened 2026-03-02
impl(pubsub): add exactly-once handler branch

dbolduc:
Part of the work for #3964 

Add a branch for exactly-once delivery to the handler.

Currently, nothing creates exactly-once handlers outside of unit tests.

### DIFF
--- a/src/pubsub/src/subscriber/handler.rs
+++ b/src/pubsub/src/subscriber/handler.rs
@@ -78,7 +78,7 @@ pub(super) enum Action {
 /// To acknowledge (ack) a message, you call [`Handler::ack()`].
 ///
 /// To reject (nack) a message, you [`drop()`][Drop::drop] the handler. The
-/// message will be redelivered.
+/// service will redeliver the message.
 ///
 /// ## Exactly-once delivery
 ///


src/pubsub/src/subscriber/handler.rs
```diff
@@ -79,11 +74,40 @@ pub(super) enum Action {
...
+///
+/// If your subscription has [exactly-once delivery] enabled, you need to
+/// destructure this enum into its [`Handler::ExactlyOnce`] branch.
+///
+/// Only when `ExactlyOnce::confirmed_ack()` returns `Ok` can you be certain
+/// that the message will not be redelivered.
+///
+/// [exactly-once delivery]: https://docs.cloud.google.com/pubsub/docs/exactly-once-delivery
+///
+/// ```no_rust
```
dbolduc, line 93: FYI: The next PR adds `ExactlyOnce::confirmed_ack()` and removes this `no_rust` tag on the code block.

src/pubsub/src/subscriber/handler.rs
```diff
@@ -79,11 +74,40 @@ pub(super) enum Action {
 ///   # panic!()
 /// }
 /// ```
+///
+/// To acknowledge (ack) a message, you call [`Handler::ack()`].
+///
+/// To reject (nack) a message, you [`drop()`][Drop::drop] the handler. The
+/// message will be redelivered.
```
coryan:
```suggestion
/// To reject (nack) a message, you [`drop()`][Drop::drop] the handler. The
/// service will redeliver the message.
```

src/pubsub/src/subscriber/handler.rs
```diff
@@ -79,11 +74,40 @@ pub(super) enum Action {
...
+///
+/// ## Exactly-once delivery
+///
+/// If your subscription has [exactly-once delivery] enabled, you need to
+/// destructure this enum into its [`Handler::ExactlyOnce`] branch.
```
coryan, line 86: This is a bit yuck, but we have tried several approaches.

dbolduc: Yes, but there are shorthands for this in Rust, and exactly-once subscriptions are the uncommon case. So we are okay with it taking more words to type.

More discussion in: go/cloud-rust:subscriber-ack-api, as you know.

But yeah, maybe too hard to do the "right thing" for exactly-once subscriptions.

PR #4867, opened 2026-03-02
impl(pubsub): confirmed ack on handlers

dbolduc:
Part of the work for #3964 

Add ability to await the result of a confirmed acks on the exactly-once handler.

`ShutdownBeforeAck` and `Shutdown` have different redelivery guarantees, so they should be different error branches. If we are returning `Shutdown` something has gone horribly wrong, likely a process exiting early.

Note that nothing creates an exactly-once handler at the moment. Eventually, the `MessageStream` will, and the `oneshot::Sender` for this channel will be held in lease management (in the `LeaseState`).

### DIFF
--- a/src/pubsub/src/error.rs
+++ b/src/pubsub/src/error.rs
@@ -51,7 +51,7 @@ pub enum PublishError {
 /// Represents an error that can occur when acknowledging a message.
 #[derive(thiserror::Error, Debug)]
 #[non_exhaustive]
-pub(crate) enum AckError {
+pub enum AckError {
     /// The message's lease expired before the client could acknowledge it.
     ///
     /// The message has not been acknowledged, and will be redelivered, maybe to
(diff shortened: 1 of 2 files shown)


src/pubsub/src/error.rs
```diff
@@ -72,6 +72,23 @@ pub(crate) enum AckError {
         #[source]
         source: Arc<Error>,
     },
+
+    /// Lease management shutdown before the client could acknowledge the
+    /// message.
+    ///
+    /// The client did not acknowledge the message. The service will redeliver
+    /// message.
```
coryan, line 80:
```suggestion
    /// This is returned if the background tasks acking messages shutdown
    /// before the application called `ack()`.
    ///
    /// # Troubleshooting
    ///
    /// The application shuts down the background task when ... Consider
    /// extending the $blah until all messages are acked.
    ///
    /// This error indicates the client library did not send the acknowledgement to the service. The service will redeliver
    /// message.
```

src/pubsub/src/error.rs
```diff
@@ -72,6 +72,23 @@ pub(crate) enum AckError {
...
+    #[error(
+        "shutdown before attempting the acknowledgement. The message was not acknowledged, and will be redelivered."
+    )]
+    ShutdownBeforeAck,
+
+    /// Error during shutdown.
+    ///
+    /// The result of the acknowledgement is unknown. The service may or may not
+    /// redeliver the message.
```
coryan, line 89: Also expand so people understand "shutdown of what?" and "why did that happen"?

dbolduc:
> Feel free to do in a follow up PR.

Made a note in #4809 for this.

PR #4871, opened 2026-03-03
refactor(gax-internal): consolidate configure_builder in request

joshuatants:
Refactor ReqwestClient so that only `request` calls `configure_builder`. This removes duplication across `execute`, `execute_http`, and `execute_streaming_once`, and ensures that the builder is configured uniformly just before the request is sent.

I thought this would make it easier to ensure that `configure_builder` (which sets the `user_agent`) is always called  no matter which `execute*` method is called.  It also makes the code more readable.

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -312,20 +312,6 @@ impl ReqwestClient {
         ))
     }
 
-    fn configure_builder(
-        &self,
-        mut builder: reqwest::RequestBuilder,
-        options: &RequestOptions,
-    ) -> Result<reqwest::RequestBuilder> {
-        if let Some(user_agent) = options.user_agent() {
-            builder = builder.header(
-                reqwest::USER_AGENT,
-                reqwest::HeaderValue::from_str(user_agent).map_err(Error::ser)?,
-            );
-        }
-        Ok(builder)
-    }
-
     async fn make_credentials(
         config: &crate::options::ClientConfig,
     ) -> ClientBuilderResult<Credentials> {
@@ -373,7 +359,14 @@ impl ReqwestClient {
         options: &RequestOptions,
         remaining_time: Option<std::time::Duration>,
     ) -> Result<reqwest::Request> {
-        builder = self.configure_builder(builder, options)?;
+        builder = if let Some(user_agent) = options.user_agent() {
+            builder.header(
+                reqwest::USER_AGENT,
+                reqwest::HeaderValue::from_str(user_agent).map_err(Error::ser)?,
+            )
+        } else {
+            builder
+        };
 
         builder = effective_timeout(options, remaining_time)
             .into_iter()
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/http.rs
```diff
@@ -376,6 +373,8 @@ impl ReqwestClient {
         options: &RequestOptions,
         remaining_time: Option<std::time::Duration>,
     ) -> Result<reqwest::Request> {
+        builder = self.configure_builder(builder, options)?;
+
```
coryan: Can you just inline the function here and remove it?  If it is called from a single place we no longer need it.

src/gax-internal/src/http.rs
```diff
@@ -912,4 +911,180 @@ mod tests {
         assert_eq!(result.err().unwrap().http_status_code(), Some(308));
         Ok(())
     }
+    #[tokio::test]
```
coryan: nit: missing blank line

src/gax-internal/src/http.rs
```diff
@@ -912,4 +911,180 @@ mod tests {
...
+    #[allow(deprecated)]
+    async fn execute_streaming_success_with_user_agent() -> TestResult {
```
coryan: I wouldn't add new tests for a function that is deprecated and not used. The only reason it is not removed is to avoid a breaking change in `google-cloud-gax-internal` (we will, on the next real breaking change).

src/gax-internal/src/http.rs
```diff
@@ -912,4 +911,180 @@ mod tests {
...
+        config.cred = Some(Anonymous::new().build());
+        let client = ReqwestClient::new(config, &server.url_str("/")).await?;
+        let builder = client.builder(Method::POST, "post".to_string());
+
+        let options = RequestOptions::default();
+
+        #[derive(serde::Deserialize, serde::Serialize, PartialEq, Debug, Default)]
+        struct ResponseBody {
+            message: String,
+        }
```
coryan: You can use `serde_json::Value` and save yourself some code.

src/gax-internal/src/http.rs
```diff
@@ -912,4 +911,180 @@ mod tests {
...
+        let client = ReqwestClient::new(config, &server.url_str("/")).await?;
+        let builder = client.builder(Method::POST, "post".to_string());
+
+        let mut options = RequestOptions::default();
+        options.set_user_agent(user_agent);
+
+        #[derive(serde::Deserialize, serde::Serialize, PartialEq, Debug, Default)]
+        struct ResponseBody {
+            message: String,
+        }
```
coryan: Ditto.

src/gax-internal/src/http.rs
```diff
@@ -912,4 +911,180 @@ mod tests {
...
+            .map_ok(|b| b.to_vec())
+            .try_concat()
+            .await?;
+        assert_eq!(body_bytes, b"hello world");
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn execute_success() -> TestResult {
+        let server = httptest::Server::run();
```
coryan: We have tests for this, just not unit tests:

https://github.com/googleapis/google-cloud-rust/blob/ec6a27757a0f6a9b17e5e5f90e5eff2dfb4313a1/src/gax-internal/tests/http_user_agent.rs#L27-L28

https://github.com/googleapis/google-cloud-rust/blob/ec6a27757a0f6a9b17e5e5f90e5eff2dfb4313a1/src/gax-internal/tests/http_request_builder.rs#L78-L84

I have no strong preference for unit vs. integration tests for this feature, but I think we should only have one of them.

src/gax-internal/src/http/reqwest.rs
```diff
@@ -22,6 +22,6 @@ pub use reqwest::Request;
 pub use reqwest::RequestBuilder;
 pub use reqwest::Response;
 pub use reqwest::StatusCode;
-pub use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
+pub use reqwest::header::{HeaderMap, HeaderName, HeaderValue, USER_AGENT};
```
coryan: This exposes `USER_AGENT` as a public API of the crate.  We should be trying to minimize the number of public APIs. Instead, consider:


```suggestion
pub use reqwest::header::{HeaderMap, HeaderName, HeaderValue};
pub(crate) use reqwest::header::USER_AGENT;
```

Or just continue to use `::reqwest::header::USER_AGENT` in the code.

coryan (changes requested):

coryan:
This is a good cleanup btw, thanks

src/gax-internal/src/http.rs
joshuatants: Ack. For posterity, here's an example. https://github.com/googleapis/google-cloud-rust/blob/ec6a27757a0f6a9b17e5e5f90e5eff2dfb4313a1/src/gax-internal/tests/http_user_agent.rs#L37C8-L40C26 .

src/gax-internal/src/http.rs
joshuatants: Ack.

src/gax-internal/src/http.rs
joshuatants: Removed.

src/gax-internal/src/http/reqwest.rs
joshuatants: Switched to pub(crate).

src/gax-internal/src/http.rs
joshuatants: Let's go with the integration tests. Removing these.

src/gax-internal/src/http.rs
joshuatants: Done.

Issue #4804, opened 2026-02-25
Improve Subscriber exactly-once reliability


dbolduc:
A follow up to #3964

The full exactly once implementation requires retrying lease extensions, retrying acks, and tweaking some settings.

We need to:

- [x] Add exactly once lease extensions to the `Leaser` and `DefaultLeaser`
- [x] Add logic to identify `TRANSIENT_FAILURE_`s
- [x] Implement exactly once retry loop for confirmed acks
- [ ] Implement exactly once retry loop for lease extensions

dbolduc:
We might also want to stop retrying if the lease has reached its max extension already. https://github.com/googleapis/google-cloud-rust/pull/4874#discussion_r2879361130

dbolduc:
We caught an error in the wild. They look like this:

```
[...]

PR #4874, opened 2026-03-03
impl(pubsub): exactly-once leases add/ack/nack

dbolduc:
Part of the work for #3964 

Start to build out the `exactly_once::Leases` data struct, which we will eventually use in the `LeaseState`.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state/at_least_once.rs
+++ b/src/pubsub/src/subscriber/lease_state/at_least_once.rs
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::ACK_IDS_PER_RPC;
 use std::collections::HashMap;
 // Use a `tokio::time::Instant` to facilitate time-based unit testing.
-use super::ACK_IDS_PER_RPC;
 use tokio::time::{Duration, Instant};
 
 /// Leases for messages with at-least-once delivery semantics.
--- a/src/pubsub/src/subscriber/lease_state/exactly_once.rs
+++ b/src/pubsub/src/subscriber/lease_state/exactly_once.rs
@@ -1,0 +1,267 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::super::handler::AckResult;
+use std::collections::HashMap;
+use tokio::sync::oneshot::Sender;
+// Use a `tokio::time::Instant` to facilitate time-based unit testing.
+use tokio::time::Instant;
+
+#[derive(Debug)]
+pub(super) struct ExactlyOnceInfo {
+    receive_time: Instant,
+    result_tx: Sender<AckResult>,
+    // If true, we are currently trying to ack this message.
+    //
+    // We need to continue to extend these leases because the exactly-once
+    // confirmed ack retry loop can take arbitrarily long.
+    //
+    // The client will not expire leases in this state. The server will
+    // report if a lease has expired. We do not want to mask a success with
+    // a `LeaseExpired` error.
+    pending: bool,
+}
+
+/// Leases for messages with exactly-once delivery semantics.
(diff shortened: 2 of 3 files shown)


src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -0,0 +1,267 @@
...
+use super::super::handler::AckResult;
+use std::collections::HashMap;
+use tokio::sync::oneshot::Sender;
+// Use a `tokio::time::Instant` to facilitate time-based unit testing.
+use tokio::time::Instant;
+
+#[derive(Debug)]
+pub(super) struct ExactlyOnceInfo {
+    receive_time: Instant,
+    result_tx: Sender<AckResult>,
```
dbolduc, line 24: These fields are unused at the moment, but I have some PRs queued up that use them eventually. (for expiration, and returning results). Happy to remove the fields if asked.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
```diff
@@ -12,9 +12,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::ACK_IDS_PER_RPC;
```
coryan, line 15: Is that a `MAX_ACK_IDS_PER_RPC` maybe ?

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -0,0 +1,267 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::super::handler::AckResult;
+use std::collections::HashMap;
+use tokio::sync::oneshot::Sender;
+// Use a `tokio::time::Instant` to facilitate time-based unit testing.
+use tokio::time::Instant;
+
+#[derive(Debug)]
```
coryan, line 21: I assume the lack of comments indicates this is an internal-only type?

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -0,0 +1,267 @@
...
+pub(super) struct ExactlyOnceInfo {
+    receive_time: Instant,
+    result_tx: Sender<AckResult>,
+    // If true, we are currently trying to ack this message.
+    //
+    // We need to continue to extend these leases because the exactly-once
+    // confirmed ack retry loop can take arbitrarily long.
```
coryan, line 28: So the scenario is that the acks are stuck in a retry loop, but the lease extensions are happily going through? That seems unlikely, but maybe not harmful...

I think in C++ we had something about not retrying the acks if the lease extension was expired.  There was no point. Maybe consider that for a future iteration.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -0,0 +1,267 @@
...
+
+/// Leases for messages with exactly-once delivery semantics.
+#[derive(Debug, Default)]
+pub struct Leases {
+    /// Ack IDs that are under lease management.
+    under_lease: HashMap<String, ExactlyOnceInfo>,
+    /// Ack IDs we need to acknowledge.
+    to_ack: Vec<String>,
+    /// Ack IDs we need to nack.
+    to_nack: Vec<String>,
```
coryan, line 44: Should these be `HashSet<String>`? You are treating them as such, and you would save yourself some time when removing elements.

dbolduc: > You are treating them as such

Only in the unit tests. But for the production code, I only ever push to these things. Computing a hash on insert would be wasteful.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
dbolduc: Yep.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
dbolduc: > That seems unlikely, but maybe not harmful...

Good point. I _think_ it is not just that the Acknowledge RPC can fail with some transient error (in which case we expect any ModifyAckDeadline RPCs to fail with similar errors), it is also that some internal server mechanism to confirm the acknowledgement can fail with a transient error. Maybe.

> not retrying the acks if the lease extension was expired

Tagged #4804 for when we get to the retry loop side of things.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
dbolduc: You are right. I will fix it in the next PR.

dbolduc: #4877

PR #4875, opened 2026-03-03
refactor(pubsub): introduce BatchActorHandle to manage batch actors

PhongChuong:
This PR is in preparation for deleting batch actors that are no longer in use.

BatchActorHandle currently only contains the sender to the batch actors but additional fields will be added.

Towards #4012

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -388,7 +388,7 @@ thiserror             = { default-features = false, version = "2.0.12" }
 time                  = { default-features = false, version = "0.3.45" }
 tokio                 = { default-features = false, version = "1.45.1" }
 tokio-metrics         = { default-features = false, version = "0.4.9" }
-tokio-util            = { default-features = false, version = "0.7.18", features = ["join-map", "rt"] }
+tokio-util            = { default-features = false, version = "0.7.18" }
 toml_edit             = { default-features = false, version = "0.25" }
 tracing-log           = { default-features = false, version = "0.2" }
 tracing-opentelemetry = { default-features = false, version = "0.32" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -388,7 +388,7 @@ thiserror             = { default-features = false, version = "2.0.12" }
 time                  = { default-features = false, version = "0.3.45" }
 tokio                 = { default-features = false, version = "1.45.1" }
 tokio-metrics         = { default-features = false, version = "0.4.9" }
-tokio-util            = { default-features = false, version = "0.7.9", features = ["join-map"] }
+tokio-util            = { default-features = false, version = "0.7.9", features = ["join-map", "rt"] }
 toml_edit             = { default-features = false, version = "0.25" }
 tracing-log           = { default-features = false, version = "0.2" }
 tracing-opentelemetry = { default-features = false, version = "0.32" }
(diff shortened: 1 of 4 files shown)


src/pubsub/src/publisher/actor.rs
coryan: Maybe no longer applicable?

src/pubsub/src/publisher/actor.rs
```diff
@@ -74,6 +75,41 @@ impl Dispatcher {
         }
     }
 
+    pub(crate) fn spawn_actor(
+        &mut self,
+        key: String,
+        tasks: &mut JoinMap<String, ()>,
+    ) -> BatchActorHandle {
+        let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
```
coryan, line 82: Bounded channels have more predictable memory usage and can one side from running wild. Would that be more appropriate here?

src/pubsub/src/publisher/actor.rs
```diff
@@ -74,6 +75,41 @@ impl Dispatcher {
...
+        match key.as_str() {
```
coryan: I love `match`, but this may be a case where `if key.is_empty()` is good enough ?

Cargo.toml
```diff
@@ -388,7 +388,7 @@ thiserror             = { default-features = false, version = "2.0.12" }
 time                  = { default-features = false, version = "0.3.45" }
 tokio                 = { default-features = false, version = "1.45.1" }
 tokio-metrics         = { default-features = false, version = "0.4.9" }
-tokio-util            = { default-features = false, version = "0.7.9" }
+tokio-util            = { default-features = false, version = "0.7.18", features = ["join-map", "rt"] }
```
coryan: I think we should add the features only where needed, so only the libraries that use the features pay for them.

src/pubsub/src/publisher/actor.rs
coryan: Ah, I see this is just a refactor, maybe as a follow up PR then.

coryan (approved):
A number of comments / questions, nothing blocking.

src/pubsub/src/publisher/actor.rs
dbolduc: nit: if `key.is_empty()`

src/pubsub/src/publisher/actor.rs
```diff
@@ -171,15 +184,21 @@ impl Dispatcher {
...
                             // Gracefully shutdown since the Publisher has dropped the Sender.
-                            // By dropping the batch actor Senders, they will individually handle the
+                            // By dropping the BatchActorHandles, they will individually handle the
                             // shutdown procedures.
+                            drop(batch_actors);
+                            // When we drop actor_tasks, some batch actors may not have started execution
+                            // so the batch actor is aborted with messages in its receiving channel.
+                            // We wait for the batch actors instead of aborting so that it can
+                            // gracefully shutdown.
+                            while (actor_tasks.join_next().await).is_some() {};
```
dbolduc: nit: unnecessary parentheses

dbolduc (approved):
optional nits

PhongChuong:
Thanks for the review. PTAL.

coryan (approved):

PR #4878, opened 2026-03-03
impl(pubsub): exactly-once leases flush

dbolduc:
Part of the work for #3964

### DIFF
--- a/src/pubsub/src/subscriber/lease_state/exactly_once.rs
+++ b/src/pubsub/src/subscriber/lease_state/exactly_once.rs
@@ -77,8 +77,8 @@ impl Leases {
         self.to_ack.len() >= MAX_IDS_PER_RPC || self.to_nack.len() >= MAX_IDS_PER_RPC
     }
 
-    /// Returns a pair of pending (acks, nacks) to flush.
-    pub fn flush(&mut self) -> (Vec<String>, Vec<String>) {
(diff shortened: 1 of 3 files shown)


src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -68,6 +69,21 @@ impl Leases {
...
+
+    /// If true, an ack or nack batch is full. We need to flush it.
+    pub fn needs_flush(&self) -> bool {
+        // This is an OR because `Acknowledge` and `ModifyAckDeadline` are
+        // separate RPCs, with separate limits.
+        self.to_ack.len() >= MAX_IDS_PER_RPC || self.to_nack.len() >= MAX_IDS_PER_RPC
+    }
+
+    /// Returns a pair of pending (acks, nacks) to flush.
+    pub fn flush(&mut self) -> (Vec<String>, Vec<String>) {
```
coryan: So maybe call the function: `drain()` or `extract_ids()` ?

coryan: Or `contents_for_flush()`...

dbolduc: `drain()` is best. Thanks.

Also, updated the equivalent at-least-once version in this PR, because I am lazy.

PR #4880, opened 2026-03-03
impl(pubsub): exactly-once leases retain

dbolduc:
Part of the work for #3964 

Implement message expiration and lease extension batching for exactly-once leases.

The logic is slightly different than its at-least-once counterpart, as we need to report the error on the result channel. There can be some extra look ups when leases expire, but not a big deal.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state/exactly_once.rs
+++ b/src/pubsub/src/subscriber/lease_state/exactly_once.rs
@@ -14,10 +14,11 @@
 
 use super::super::handler::AckResult;
 use super::MAX_IDS_PER_RPC;
+use crate::error::AckError;
 use std::collections::HashMap;
 use tokio::sync::oneshot::Sender;
 // Use a `tokio::time::Instant` to facilitate time-based unit testing.
-use tokio::time::Instant;
+use tokio::time::{Duration, Instant};
 
 #[derive(Debug)]
 pub(super) struct ExactlyOnceInfo {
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -84,6 +85,38 @@ impl Leases {
...
+
+    /// Returns batches of ack IDs to extend.
+    ///
+    /// Drops messages whose lease deadline cannot be extended any further.
+    pub fn retain(&mut self, max_lease_extension: Duration) -> Vec<Vec<String>> {
+        let now = Instant::now();
+
+        // We cannot satisfy the `result_tx` if we use `HashMap::retain()`. We
+        // prefer to just look up expired ack IDs again in the map, over holding
+        // the `result_tx` in an `Option<>`.
```
coryan, line 97: I think I know what you are saying, but maybe:

1) You should be able to use a `Vec<&str>` instead of a `Vec<String> and save some allocations.
2) Maybe `HashMap::extract_if()` helps?

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -84,6 +85,38 @@ impl Leases {
...
+        }
+
+        self.under_lease
+            .keys()
+            .cloned()
+            .collect::<Vec<_>>()
+            .chunks(MAX_IDS_PER_RPC)
+            .map(|chunk| chunk.to_vec())
+            .collect()
+    }
```
coryan, line 119: Maybe something like:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=0b61e6c9b82a6719e76cda273f6d2a38

???

I also think it is sad to allocate vectors of vectors... Can you return a single vector and let the caller operate on the slices from `.chunks()` themselves?

src/pubsub/src/subscriber/lease_state/exactly_once.rs
dbolduc: 2. `extract_if()` would help, but was added in 1.88

src/pubsub/src/subscriber/lease_state/exactly_once.rs
coryan: Blegh on 1.88 and extract_if().  There is this:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=1c5defa6a0a5d2d256d7f61af6b01d5a

coryan: Forget that, you want the keys that remain...  Yeah, what you have is as good as it gets I think.

dbolduc: > what you have is as good as it gets

I definitely would not go that far. It iterates over the hashmap twice, and does extra look ups on expiration. I'd say "it is fine".

> it is sad to allocate vectors of vectors

Hmm.. I think I see what you're saying with a big flat `Vec<&str>`. Like you should be able to iterate over chunks of that and use it to populate requests without requiring intermediate `Vec`s.

I think it doesn't save us any allocations, given that I wrote this Leaser interface expecting a Vec<String> for each request.

https://github.com/googleapis/google-cloud-rust/blob/42ac50f91d6ab330beb37c78f199b2b5a79d41c4/src/pubsub/src/subscriber/leaser.rs#L35

coryan: ack.

coryan (approved):

Issue #4760, opened 2026-02-22
Add per-request duration metrics to the GAPICs


coryan:
One of the golden signals is the duration of each RPC, with many attributes to identify the client library, the service, the RPC, whether it was successful, etc.

- [x] `sidekick`: create the "duration histogram" metric per-client
- [ ] `sidekick`: generate the code updating the "duration histogram" after each request
- [ ] Storage: update the "duration histogram" for hand-crafted methods
- [ ] Write an end-to-end integration test for duration metrics

coryan:
This is done, we just need to enable the generation for all GAPICs.

PR #4881, opened 2026-03-03
impl(gax-internal): per-request metric helpers

coryan:
Define helpers to record per-request duration metrics in the client. The metrics are histograms, in seconds, with attributes to separate by method, service, status, client library version, etc.

The helpers in this PR create the metric and then record the duration with all the right attributes. In a future PR, some macros and trait extensions will simplify using these things even further.

Towards #4760 , to see a larger picture (with some tests missing and other problems) look at #4879

### DIFF
--- a/src/gax-internal/src/observability/client_signals/duration_metric.rs
+++ b/src/gax-internal/src/observability/client_signals/duration_metric.rs
@@ -87,7 +87,7 @@ impl DurationMetric {
         // Use a `Vec` to omit HTTP_RESPONSE_STATUS_CODE. This extra allocation
         // occurs only on error paths, which should be rare.
         let mut attributes = Vec::from_iter([
-            KeyValue::new("rpc.system.name", RPC_SYSTEM_HTTP),
+            KeyValue::new(RPC_SYSTEM_NAME, RPC_SYSTEM_HTTP),
             KeyValue::new(attribute::URL_DOMAIN, start.info().default_host),
             KeyValue::new(attribute::URL_TEMPLATE, start.url_template()),
             KeyValue::new(attribute::RPC_METHOD, start.method()),


src/gax-internal/src/observability/client_signals/duration_metric.rs
```diff
@@ -0,0 +1,287 @@
...
+    }
+
+    #[track_caller]
+    fn check_data(
+        metrics: &Vec<ResourceMetrics>,
+        want_attributes: &[(&'static str, &'static str)],
+    ) {
+        let mut iter = metrics
+            .iter()
+            .flat_map(|s| s.scope_metrics())
```
dbolduc, line 262: TIL: `flat_map`

src/gax-internal/src/observability/client_signals/duration_metric.rs
```diff
@@ -0,0 +1,287 @@
...
+                ("rpc.method", "test-method"),
+            ]
+            .iter()
+            .chain(want_attributes)
+            .map(|(k, v)| (*k, v.to_string())),
+        );
+        let diff = attr.symmetric_difference(&want).collect::<Vec<_>>();
+        assert_eq!(attr, want, "diff={diff:?}");
+
+        // The sample was for 123.456 seconds.
```
dbolduc: .5 seconds

src/gax-internal/src/observability/client_signals/duration_metric.rs
```diff
@@ -0,0 +1,287 @@
...
+        let bucket = point
+            .bucket_counts()
+            .zip(point.bounds())
+            .find(|(count, _bound)| *count >= 1_u64);
+        assert!(
+            bucket.is_some_and(|(c, b)| c == 1 && b >= 0.5),
```
dbolduc: why is it `b >= .5`?

Maybe you have the tests use `Duration::from_millis(999)` and check that the `b == 1` instead of picking a data point on a bucket boundary?

src/gax-internal/src/observability/client_signals/duration_metric.rs
```diff
@@ -0,0 +1,287 @@
...
+        let elapsed = start.elapsed();
+        self.0.record(
+            elapsed.as_secs_f64(),
+            &[
+                KeyValue::new("rpc.system.name", RPC_SYSTEM_HTTP),
+                KeyValue::new(attribute::URL_DOMAIN, start.info().default_host),
+                KeyValue::new(attribute::URL_TEMPLATE, start.url_template()),
+                KeyValue::new(attribute::RPC_METHOD, start.method()),
+                KeyValue::new(OTEL_STATUS_CODE, otel_status_codes::OK),
+                KeyValue::new("rpc.response.status_code", Code::Ok.name()),
```
westarle: Yikes the semconv crate is a few versions out of date 

https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-semantic-conventions/CHANGELOG.md

I wonder if we're better off hardcoding more of these.

westarle: Please add this to `attributes::keys`.

src/gax-internal/src/observability/client_signals/duration_metric.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub(crate) fn record_ok(&self, start: RequestStart) {
+        let elapsed = start.elapsed();
+        self.0.record(
+            elapsed.as_secs_f64(),
+            &[
+                KeyValue::new("rpc.system.name", RPC_SYSTEM_HTTP),
+                KeyValue::new(attribute::URL_DOMAIN, start.info().default_host),
+                KeyValue::new(attribute::URL_TEMPLATE, start.url_template()),
+                KeyValue::new(attribute::RPC_METHOD, start.method()),
+                KeyValue::new(OTEL_STATUS_CODE, otel_status_codes::OK),
```
westarle: I don't think this is needed or desirable on metrics.  The [`otel.x` attributes](https://docs.rs/tracing-opentelemetry/latest/tracing_opentelemetry/index.html#special-fields) were set on spans to cue tracing_opentelemetry crate to do special things with OTel API.

src/gax-internal/src/observability/client_signals/duration_metric.rs
```diff
@@ -0,0 +1,287 @@
...
+        Self(histogram)
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn record_ok(&self, start: RequestStart) {
+        let elapsed = start.elapsed();
+        self.0.record(
+            elapsed.as_secs_f64(),
+            &[
+                KeyValue::new("rpc.system.name", RPC_SYSTEM_HTTP),
```
westarle: Please add this to `attributes::keys`.

westarle: Could this duration be used on gRPC clients?  If so, we should pass in the rpc system value (to new_with_provider?)

src/gax-internal/src/observability/client_signals/duration_metric.rs
```diff
@@ -0,0 +1,287 @@
...
+                KeyValue::new(
+                    "rpc.response.status_code",
+                    error
+                        .status()
+                        .map(|s| s.code.name())
+                        .unwrap_or(Code::Unknown.name()),
+                ),
+                KeyValue::new(
+                    attribute::HTTP_RESPONSE_STATUS_CODE,
+                    error.http_status_code().unwrap_or_default() as i64,
```
westarle: I think we should skip this value if None (instead of emitting 0).

src/gax-internal/src/observability/client_signals/duration_metric.rs
```diff
@@ -0,0 +1,287 @@
...
+    pub(crate) fn record_error(&self, start: RequestStart, error: &Error) {
+        let elapsed = start.elapsed();
+        self.0.record(
+            elapsed.as_secs_f64(),
+            &[
+                KeyValue::new("rpc.system.name", RPC_SYSTEM_HTTP),
+                KeyValue::new(attribute::URL_DOMAIN, start.info().default_host),
+                KeyValue::new(attribute::URL_TEMPLATE, start.url_template()),
+                KeyValue::new(attribute::RPC_METHOD, start.method()),
+                KeyValue::new(OTEL_STATUS_CODE, otel_status_codes::ERROR),
```
westarle: Rather than this, we should (eventually) have `error.type` with values like `RATE_LIMIT_EXCEEDED` from the error reason.

westarle:
This is looking really great!  Just a few nits to fix.

src/gax-internal/src/observability/client_signals/duration_metric.rs
coryan: Fixed.

src/gax-internal/src/observability/client_signals/duration_metric.rs
coryan: Thanks. Having a value exactly at a boundary was hiding a test bug.  The counts in the histogram start with "the bucket for all the values before the first boundary", and I was not taking that into account. It is a bit more complicated now, but it works.

src/gax-internal/src/observability/client_signals/duration_metric.rs
coryan: Cannot be in the metric, we will have one metric per client library.  Punting to #4891

src/gax-internal/src/observability/client_signals/duration_metric.rs
coryan: I had not paid any attention to `error.type`: I am not sure it is a good idea. It seems like a recipe for bug reports: "the service returned ResourceExhausted because we were out of quota and you said RATE_LIMIT_EXCEEEDED in the metrics". Or "the service was missing the error details so we put a weird error type".

I think we should stick to status codes.

coryan: I don't think you are asking me to change this right now, so I won't.  We can talk later and (most likely) you can change my mind.

src/gax-internal/src/observability/client_signals/duration_metric.rs
coryan: Done.

src/gax-internal/src/observability/client_signals/duration_metric.rs
coryan: Removed.

src/gax-internal/src/observability/client_signals/duration_metric.rs
coryan: Frankly, I find the constants less readable.  I have to constantly look them up to find out what is the right constant for "url.domain"....  I understand the desired to avoid typos though.

coryan: Done.

src/gax-internal/src/observability/client_signals/duration_metric.rs
coryan: Done.  I remember something about adding or removing attributes to a metric being a breaking change.  I take it that this is fine?

src/gax-internal/src/observability/client_signals/duration_metric.rs
```diff
@@ -0,0 +1,325 @@
...
+            ],
+        );
+    }
+
+    #[allow(dead_code)]
+    pub(crate) fn record_error(&self, start: RequestStart, error: &Error) {
+        let elapsed = start.elapsed();
+        // We want to omit the HTTP_RESPONSE_STATUS_CODE attribute if the error
+        // is missing the HTTP status code. We could try to avoid the memory
+        // allocation required for the `Vec`. I (coryan@) do not think it is
```
westarle: super-nit: I think this is generally true.

"Use a Vec to omit HTTP_RESPONSE_STATUS_CODE.  This extra allocation occurs only on error paths, which should be rare."

coryan: Done.

PR #4882, opened 2026-03-03
impl(auth): add retry to access boundary rpc

alvarowolfx:
Towards #4186

### DIFF
--- a/src/auth/src/access_boundary.rs
+++ b/src/auth/src/access_boundary.rs
@@ -322,7 +322,7 @@ where
     let retry_throttler: SharedRetryThrottler = retry_throttler.into();
 
     retry_loop(
-        async move |d| {        
+        async move |d| {
             let headers = credentials
                 .headers(Extensions::new())
                 .await
@@ -330,15 +330,12 @@ where
 
             let attempt = fetch_access_boundary_call(&client, &url, headers);
             match d {
-                Some(timeout) => {
-                    match tokio::time::timeout(timeout, attempt).await {
-                        Ok(r) => r,
-                        Err(e) => Err(GaxError::timeout(e)),
-                    }
-                }
-                None => attempt.await
-
-            }  
+                Some(timeout) => match tokio::time::timeout(timeout, attempt).await {
+                    Ok(r) => r,
+                    Err(e) => Err(GaxError::timeout(e)),
+                },
+                None => attempt.await,
+            }
         },
         sleep,
         true, // fetch access boundary is idempotent


src/auth/src/access_boundary.rs
```diff
@@ -274,45 +284,91 @@ struct AllowedLocationsResponse {
...
+    url: String,
+) -> GaxResult<AllowedLocationsResponse>
 where
     T: dynamic::AccessTokenCredentialsProvider + 'static,
 {
-    let headers = credentials.headers(Extensions::new()).await?;
+    let client = Client::new();
+    let sleep = async |d| tokio::time::sleep(d).await;
+
+    let retry_policy: RetryPolicyArg = Aip194Strict.with_attempt_limit(3).into();
```
coryan: This is super conservative. For an idempotent operation I would consider `AlwaysRetry.with_attempt_limit(3)` kind of fine, or `Aip194Strict.continue_on_too_many_requests().with_attempt_limit(3)`.

Separately, I am not sure why `with_attempt_limit(3)` has gained such cache amongst software engineers.  Saying things three times does not make them true, and retrying 3 times is not perfect either.

Consider `.with_time_limit(Duration::from_secs(60))` which at least tells you how long it will run (modulo some nits).

src/auth/src/access_boundary.rs
```diff
@@ -274,45 +284,91 @@ struct AllowedLocationsResponse {
...
+
+    retry_loop(
+        async move |_| {
+            let headers = credentials
+                .headers(Extensions::new())
+                .await
+                .map_err(GaxError::authentication)?;
+
+            fetch_access_boundary_call(&client, &url, headers).await
+        },
```
coryan: Consider adding a timeout to these things:


```suggestion
        async move |d /*: Option<Duration>*/| {
            let headers = credentials
                .headers(Extensions::new())
                .await
                .map_err(GaxError::authentication)?;

            fetch_access_boundary_call(&client, &url, headers, d).await
        },
```

And then you can limit each attempt to some timeout value.

src/auth/src/access_boundary.rs
```diff
@@ -822,4 +879,48 @@ pub(crate) mod tests {
...
+            .version(http::Version::HTTP_3) // unsupported version
+            .status(204)
+            .body(Vec::new())
+            .unwrap();
+
+        server.expect(
+            Expectation::matching(request::method_path("GET", "/allowedLocations"))
+                .times(3)
+                .respond_with(cycle![
+                    invalid_res, // forces i/o error
```
coryan, line 905: Interesting idea... though I guess could fail in the future. We can worry about it then.

alvarowolfx: that's the only way that I found to simulate I/O errors on `reqwest` calls. It's janky, but works for now 🤷

src/auth/src/access_boundary.rs
alvarowolfx: I added a timeout to the whole `fetch_access_boundary_call`, I thought about using `timeout` in the `reqwest` builder since is the slowest operation, but then I would have to add some match to detect the diff between I/O and Timeout errors. Also I wasn't sure if adding a timeout to the credentials part would be good, since this starts together with the token fetch mechanism and it's not the core of the attempt.

src/auth/src/access_boundary.rs
alvarowolfx: I moved to `Aip194Strict.with_time_limit(Duration::from_secs(60))` since the Access Boundary design asks to retry on 503 errors, with exp backoff and for a minute. Was a oversight from my end.

coryan: Aip194Strict does **not** retry on 503s... It only retries on `Unavailable`..

PR #4885, opened 2026-03-03
impl(pubsub): exactly-once leases evict

dbolduc:
Part of the work for #3964 

Support the `NackImmediatly` shutdown behavior for exactly-once leases.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state/exactly_once.rs
+++ b/src/pubsub/src/subscriber/lease_state/exactly_once.rs
@@ -21,7 +21,8 @@ use tokio::sync::oneshot::Sender;
 use tokio::time::{Duration, Instant};
 
 // TODO(#4868) - mention alternative shutdown options, when implemented.
-const NACK_SHUTDOWN_ERROR: &str = "subscriber is configured to nack all pending messages on shutdown.";
+const NACK_SHUTDOWN_ERROR: &str =
+    "subscriber is configured to nack all pending messages on shutdown.";
 
 #[derive(Debug)]
 pub(super) struct ExactlyOnceInfo {


src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -20,6 +20,8 @@ use tokio::sync::oneshot::Sender;
 // Use a `tokio::time::Instant` to facilitate time-based unit testing.
 use tokio::time::{Duration, Instant};
 
+const NACK_SHUTDOWN_ERROR: &str = "subscriber was configured to `NackImmediately`. If this is surprising, you should configure it to `WaitForProcessing` instead.";
```
dbolduc: This could be a bad idea. We have not implemented `NackImmediately` or `WaitForProcessing` yet. And when we do implement these, it may be with different names.

Maybe something like:

"the subscriber nacks messages immediately when it shuts down. We are working on supporting other behaviors, such as waiting for the application to process its messages. See #XXXX."

🤷

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -117,6 +119,20 @@ impl Leases {
             .map(|chunk| chunk.to_vec())
             .collect()
     }
+
+    /// Nacks all messages under lease management that have not been acked by
+    /// the application.
+    ///
+    /// Called during shutdown, if configured to `NackImmediately`.
```
dbolduc, line 138: same, but this is not user-facing, so less of a big deal.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
coryan: Maybe just say:


```suggestion
const NACK_SHUTDOWN_ERROR: &str = "subscriber was configured to `NackImmediately`. During shutdown, any pending messages were nacked and calls to `ack() generate this error. To avoid this error, use `blah.blah(WaitForProcessing)`.";
```

dbolduc: Sorry, I meant to call this out. I decided to not mention `NackImmediately` and `WaitForProcessing` because they don't exist yet. I tagged an issue to update the error message when we offer an alternative shutdown behavior.

PR #4890, opened 2026-03-04
test(storage): integration test for user agent in open_object

joshuatants:
Adds a test to ensure the custom user agent is correctly passed to the server during open_object requests.

Additionally, refactors the open_object tests to use shared constants (BUCKET_NAME, OBJECT_NAME, USER_AGENT, ENDPOINT), removes redundant clone calls, and consolidates duplicate imports.

### DIFF
--- a/src/storage/src/storage/open_object.rs
+++ b/src/storage/src/storage/open_object.rs
@@ -435,7 +435,7 @@ mod tests {
     const BUCKET_NAME: &str = "projects/_/buckets/test-bucket";
     const OBJECT_NAME: &str = "test-object";
     const USER_AGENT: &str = "quick_foxes_lazy_dogs/1.2.3";
-    const ENDPOINT: &str = "0.0.0.0:0";
+    const BIND_ADDRESS: &str = "0.0.0.0:0";
 
     // Verify `open_object()` meets normal Send, Sync, requirements.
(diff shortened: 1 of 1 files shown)


joshuatants:
Will look into adding integration tests for read_object and write_object later.

src/storage/src/storage/open_object.rs
```diff
@@ -432,11 +432,15 @@ mod tests {
     };
     use storage_grpc_mock::{MockStorage, start};
 
+    const BUCKET_NAME: &str = "projects/_/buckets/test-bucket";
+    const OBJECT_NAME: &str = "test-object";
+    const USER_AGENT: &str = "quick_foxes_lazy_dogs/1.2.3";
+    const ENDPOINT: &str = "0.0.0.0:0";
```
coryan: nit: when jumping to the middle of the code I found `ENDPOINT` surprising.  At first I thought it was a fixed endpoint, which made no sense. But I don't have a better suggestion.

joshuatants: How about BIND_ADDRESS? Merging first, happy to open another PR if you think of a better suggestion.

coryan: Great idea, thanks.

PR #4897, opened 2026-03-04
WIP: fix(pubsub): remove actors that are no longer processing messages

### LINKED ISSUE #4012: Pubsub: Support message ordering in Publisher

PhongChuong:
The Dispatcher now removes batch_actors entries that are no longer processing messages.

A batch actor will close its channel and begin shutdown when it detects that it has no pending messages. Any remaining messages in the channel will be handled by the run loop.

When an actor shutdown, the Dispatcher joins its task and removes the entry from batch_actors. Since sequential actors requires batches to complete before the next batch can be sent, we keep a buffer of messages that arrives after the channel is closed. This buffer is passed onto the next batch actor is needed.

This PR updates 2 tests to reflect the new behavior:
1. client::batch_sends_on_byte_threshold: don't check the number of times the rpc is performed since the actor shutdown can change how batching is performed.
2. actor::sequential_actor_flush: don't check empty flush since it causes the actor to shutdown. Plus, the case is already covered by the test client::flush_with_no_messages_is_noop

Fixes #4012

### DIFF
--- a/src/pubsub/src/publisher/actor.rs
+++ b/src/pubsub/src/publisher/actor.rs
@@ -16,6 +16,7 @@ use super::options::BatchingOptions;
 use crate::generated::gapic_dataplane::client::Publisher as GapicPublisher;
 use crate::publisher::batch::Batch;
 use std::collections::{HashMap, VecDeque};
+use tokio::sync::mpsc::error::SendError;
 use tokio::sync::{mpsc, oneshot};
 use tokio::task::JoinSet;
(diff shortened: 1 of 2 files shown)


coryan:
Is this still needed?

Issue #4795, opened 2026-02-24
Add error logs to generated clients


coryan:
All the generated clients need:

- [ ] A configuration to enable error logging.
- [ ] When enabled, log errors returned to the application, in the T3 span.
- [ ] When enabled, log errors in the HTTP or gRPC low-level (T4) spans.
- [ ] The error messages should have the same attributes as the corresponding trace.
- [ ] The error message should include whatever attribute or context to correlate them to the span
- [ ] In logs, the `rpc.method` attribute should be the name of the gRPC method.
- [ ] In storage logs, the `rpc.method` attribute should be the name of the storage JSON v1 method
- [ ] We need to decide if we use the same `target:` and `name:` attribute for all the `tracing::error!()` calls or these are different for each library
- [ ] If we use a different attribute per library then the macro invocation needs to happen in client library code.

coryan:
We will fix the fields in separate bugs.

PR #4898, opened 2026-03-04
impl(gax-internal): future with client-request signals

coryan:
Implements a `std::future::Future` with request duration metrics and error logging.

Part of the work for #4795 and for #4760

### DIFF
--- a/src/gax-internal/src/observability/client_signals/with_client_signals.rs
+++ b/src/gax-internal/src/observability/client_signals/with_client_signals.rs
@@ -76,7 +76,9 @@ const TARGET: &str = "experimental.client.request";
 pub struct WithClientSignals<F> {
     #[pin]
     inner: F,
+    #[pin]
     metric: DurationMetric,
+    #[pin]
     start: RequestStart,
     span: Span,
 }
@@ -104,14 +106,12 @@ where
     type Output = <F as Future>::Output;
     fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
         let span = self.span.clone();
-        let start = self.start;
-        let metric = self.metric.clone();
         let this = self.project();
         let output = futures::ready!(this.inner.poll(cx));
         // Record the metric and log the value in the context of the span.
         let _enter = span.entered();
         match &output {
-            Ok(_) => metric.record_ok(start),
+            Ok(_) => this.metric.record_ok(&this.start),
             Err(error) => {
                 let rpc_status_code = error
                     .status()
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/observability/client_signals/with_client_signals.rs
```diff
@@ -0,0 +1,393 @@
...
+
+impl<F, R> Future for WithClientSignals<Instrumented<F>>
+where
+    F: Future<Output = Result<R, Error>>,
+{
+    type Output = <F as Future>::Output;
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let span = self.span.clone();
+        let start = self.start;
+        let metric = self.metric.clone();
```
westarle: Do you need these, or can you use `this.metric` directly after projecting?

If you need these cloned, could we move them below the `futures::ready!` to save cost of arc for `metric`?

NB: This is at the edge of my Rust knowledge.

coryan: > Do you need these, or can you use `this.metric` directly after projecting?

We need the span, it has be cloned to call `entered` anyway.  The others we can make part of the projection.

> If you need these cloned, could we move them below the `futures::ready!` to save cost of arc for `metric`?
>

They cannot be accessed after projecting the Pin.
 
> NB: This is at the edge of my Rust knowledge.

I have read the `std::pin` documentation like 12 times.  I know what they are trying to do, but barely understand how to use a `std::pin::Ping<>` and have very little clue of what `pin_project!` does (shame, shame, shame).

src/gax-internal/src/observability/client_signals/with_client_signals.rs
```diff
@@ -0,0 +1,393 @@
...
+/// callsite.
+///
+/// The final code also uses an extension trait to simplify the call to decorate
+/// the `self.inner.echo()` future.
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+#[pin_project]
+pub struct WithClientSignals<F> {
+    #[pin]
+    inner: F,
+    #[pin]
```
westarle, line 79: I'm not sure you need pin here, since you're not polling it.  And project should just give you a standard mut & ref to it.

src/gax-internal/src/observability/client_signals/with_client_signals.rs
```diff
@@ -0,0 +1,393 @@
...
+    }
+}
+
+impl<F, R> Future for WithClientSignals<Instrumented<F>>
+where
+    F: Future<Output = Result<R, Error>>,
+{
+    type Output = <F as Future>::Output;
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let span = self.span.clone();
```
westarle, line 108: maybe a comment to say a span is a handle and so clone is cheap.

src/gax-internal/src/observability/client_signals/with_client_signals.rs
```diff
@@ -0,0 +1,393 @@
...
+        let this = self.project();
+        let output = futures::ready!(this.inner.poll(cx));
```
westarle, line 110: and a comment to say this returns internally if inner is not ready.

src/gax-internal/src/observability/client_signals/with_client_signals.rs
coryan: The `Pin<>` applies to the whole structure, the `#[pin]` directive is for `pin_project!`:

https://docs.rs/pin-project/latest/pin_project/attr.pin_project.html#:~:text=An%20attribute%20that,to%20the%20field.

src/gax-internal/src/observability/client_signals/with_client_signals.rs
coryan: Pass. I will assume people working in this code know about `Span`.

src/gax-internal/src/observability/client_signals/with_client_signals.rs
coryan: Ditto.

PR #4904, opened 2026-03-05
cleanup(guide): use `random_secret_id()`

coryan:
Refactor the code to use the common function to generate secret ids.

### DIFF
--- a/guide/samples/tests/driver.rs
+++ b/guide/samples/tests/driver.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// #[cfg(all(test, feature = "run-integration-tests"))]
+#[cfg(all(test, feature = "run-integration-tests"))]
 mod driver {
     use google_cloud_gax::error::rpc::{Code, StatusDetails};
     use google_cloud_secretmanager_v1::client::SecretManagerService;


guide/samples/tests/driver.rs
```diff
@@ -12,13 +12,11 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-#[cfg(all(test, feature = "run-integration-tests"))]
+// #[cfg(all(test, feature = "run-integration-tests"))]
```
PhongChuong: uncomment?

guide/samples/tests/driver.rs
```diff
@@ -186,11 +180,7 @@ mod driver {
         use std::time::Duration;
 
         let project_id = std::env::var("GOOGLE_CLOUD_PROJECT").unwrap();
-        let secret_id: String = rand::rng()
-            .sample_iter(&Alphanumeric)
-            .take(SECRET_ID_LENGTH)
-            .map(char::from)
-            .collect();
+        let secret_id = random_secret_id();
```
PhongChuong, line 183: Not in this PR but would it be useful to pass in a test name prefix (in addition to the "rust-sdk-testing-" prefix) to easily map the resource to the test?

Another option to get the test name using `std::thread::current().name` within the random function.

guide/samples/tests/driver.rs
coryan: Fixed, thanks.

guide/samples/tests/driver.rs
coryan: It would make troubleshooting resource leaks a bit easier. I would approve something automated like what you describe. Keep in mind the resource id length limits.

PR #4911, opened 2026-03-05
impl(auth): handle etags for boundaries

alvarowolfx:
Towards #4186

### DIFF
--- a/src/auth/src/access_boundary.rs
+++ b/src/auth/src/access_boundary.rs
@@ -35,7 +35,7 @@ use std::clone::Clone;
 use std::error::Error;
 use std::fmt::Debug;
 use std::sync::{Arc, OnceLock};
-use tokio::sync::watch;
+use tokio::sync::{Mutex, watch};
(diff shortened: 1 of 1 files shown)


src/auth/src/access_boundary.rs
```diff
@@ -114,6 +115,77 @@ where
...
+    cache: Arc<EntityTagCache<Option<String>>>,
+}
+
+/// A cache designed to track cache validity for composite resources.
+/// It maps a dynamically requested "outer" `EntityTag` to the "inner" `EntityTag`
+/// provided by the underlying credentials layer, alongside a value that represents
+/// the current properties injected by the composite layer (e.g. an access boundary).
+#[derive(Debug)]
+struct EntityTagCache<T> {
+    state: std::sync::RwLock<Option<EntityTagState<T>>>,
```
coryan: Under low contention, Mutex should be faster than RwLock, even if most operations are "reads".  Are you sure we need the extra complexity?

PR #4919, opened 2026-03-06
test(integration-tests-o11y): fix flakes

westarle:
#### Using `enter` across `await`s is a problem, because the span is stored in TLS.

See https://docs.rs/tracing/latest/tracing/span/struct.Span.html#in-asynchronous-code

#### Remove debug span list_buckets from expectations.

This span is consistently dropped, likely because it is large (~130k of OTLP data).  It (an event attached to it) contains the full response data from the listbuckets request which can be quite large.

A sufficient e2e test for debug level spans (those that include self and ret) would be to enable tracing, set RUSTLOGS=debug and make sure they are visible in console-printed tracing output.

Ideally we update the generated clients with our new Client Request traces.  I think the next best thing is to remove our expectation from tests to tighten them up.

### DIFF
--- a/tests/o11y/src/e2e/showcase.rs
+++ b/tests/o11y/src/e2e/showcase.rs
@@ -84,7 +84,7 @@ pub async fn run() -> anyhow::Result<()> {
         ROOT_SPAN_NAME,
         "google_cloud_showcase_v1beta1::client::Echo::echo",
     ]);
-    let trace = wait_for_trace(&project_id, &trace_id, required.len()).await?;
+    let trace = wait_for_trace(&project_id, &trace_id, &required).await?;
 
     // Verify the expected spans appear in the trace:
     let span_names = trace
--- a/tests/o11y/src/e2e.rs
+++ b/tests/o11y/src/e2e.rs
@@ -46,7 +46,7 @@ static CREDENTIALS: OnceCell<anyhow::Result<Credentials>> = OnceCell::const_new(
 pub async fn wait_for_trace(
     project_id: &str,
     trace_id: &str,
-    required_spans: usize,
+    required_spans: &std::collections::BTreeSet<&str>,
 ) -> anyhow::Result<Trace> {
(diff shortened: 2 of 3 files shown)


tests/o11y/src/e2e/showcase.rs
```diff
@@ -43,27 +43,31 @@ pub async fn run() -> anyhow::Result<()> {
...
-        let _ = client.echo().set_content("test").send().await?;
+            // Make the API call
+            // This will generate child spans within the library
+            let _ = client.echo().set_content("test").send().await?;
 
-        trace_id
+            Ok::<_, anyhow::Error>(trace_id)
+        }
+        .instrument(root_span.clone())
+        .await?
```
coryan: Maybe just instrument the future making the call?  Thought 

```suggestion
            // Initialize showcase client pointing to local mock server
            let client = Echo::builder()
                .with_endpoint(format!("http://{}", echo_server.addr()))
                .with_credentials(Anonymous::new().build())
                .with_tracing()
                .build()
                .await?;

            // Make the API call
            // This will generate child spans within the library
            let _ = client.echo().set_content("test").send().await.instrument(root_span.clone()?;
```

But I am not sure how you get the trace_id?

coryan:
Maybe we are hitting some limit, like the total number of spans + events per trace?  Consider disabling the events for `read_object()`, or maybe just generate them on errors and EOF?

westarle:
> Maybe we are hitting some limit, like the total number of spans + events per trace? Consider disabling the events for `read_object()`, or maybe just generate them on errors and EOF?

westarle:
> Maybe we are hitting some limit, like the total number of spans + events per trace? Consider disabling the events for `read_object()`, or maybe just generate them on errors and EOF?

I think you're right, the size of the list_buckets is consistently really large.  Refocused the PR on removing that expectation (see the PR description)

coryan (approved):

PR #4925, opened 2026-03-06
impl(gax-internal): a macro for the client tracing layer

coryan:
Introduce a macro to simplify the client tracing layer. With this macro and a few changes in the generator we can have spans, duration metrics, and logs in all the GAPICs. The attributes will need some work.

Towards #4795 and #4760 

Sorry, some large moves of helper functions for testing.

### DIFF
--- a/src/gax-internal/src/observability/client_signals.rs
+++ b/src/gax-internal/src/observability/client_signals.rs
@@ -75,7 +75,7 @@ pub use with_client_signals::WithClientSignals;
 /// [Span]: [tracing::Span]
 #[macro_export]
 macro_rules! client_request_signals {
-    ($info:expr, $options:expr, $client:expr, $method:expr, $rpc_method:expr) => {{
+    ($info:expr, $options:expr, $client:literal, $method:literal, $rpc_method:expr) => {{
         use $crate::observability::attributes::keys::*;
         use $crate::observability::attributes::{
             GCP_CLIENT_LANGUAGE_RUST, GCP_CLIENT_REPO_GOOGLEAPIS, OTEL_KIND_INTERNAL,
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/observability/client_signals.rs
```diff
@@ -22,6 +22,101 @@ pub use duration_metric::DurationMetric;
...
+        };
+        let start = $crate::observability::RequestStart::new(
+            $info,
+            $options,
+            concat!(env!("CARGO_CRATE_NAME"), "::", $client, "::", $method),
+        );
+        let span = tracing::info_span!(
+            "client_request",
+            "gax.client.span" = true, // Marker field
+            { OTEL_NAME } = concat!(env!("CARGO_CRATE_NAME"), "::", $client, "::", $method),
```
westarle, line 92: nit: since concat takes literals, you might be able to change the types to $client:literal / $method:literal and maybe improve the error.

src/gax-internal/src/observability/client_signals.rs
```diff
@@ -22,6 +22,101 @@ pub use duration_metric::DurationMetric;
...
+        use $crate::observability::attributes::{
+            GCP_CLIENT_LANGUAGE_RUST, GCP_CLIENT_REPO_GOOGLEAPIS, OTEL_KIND_INTERNAL,
+            RPC_SYSTEM_HTTP, otel_status_codes::UNSET,
+        };
+        let start = $crate::observability::RequestStart::new(
+            $info,
+            $options,
+            concat!(env!("CARGO_CRATE_NAME"), "::", $client, "::", $method),
+        );
+        let span = tracing::info_span!(
```
westarle: nit
```suggestion
        let span = ::tracing::info_span!(
```

src/gax-internal/src/observability/client_signals.rs
coryan: Done.

src/gax-internal/src/observability/client_signals.rs
coryan: Done.

PR #4927, opened 2026-03-06
chore: bump librarian version circa 2026-03-06

coryan:


src/generated/cloud/secretmanager/v1/src/transport.rs
```diff
@@ -321,44 +337,44 @@ impl super::stub::SecretManagerService for SecretManagerService {
...
+                        Segment::SingleWildcard,
+                        Segment::Literal("/locations/"),
+                        Segment::SingleWildcard,
+                        Segment::Literal("/secrets/"),
+                        Segment::SingleWildcard,
+                    ],
+                )?;
+                let path = format!("/v1/{}", var_name,);
+
+                let _resource_name = format!("//secretmanager.googleapis.com/v1/{}", var_name,);
```
coryan: @haphungw @westarle I think I saw some discussion about the correct format for these. That `v1/` looks spurious. Thoughts?

FWIW, the `_resource_name` is unused at the moment [^1], so we can ignore the problem.

[^1]: It is used in this function, so one thing is to rename it, but once stored in `RequestOptionsExtension` it goes unused by any of the downstream code.

src/generated/devtools/cloudbuild/v1/src/transport.rs
```diff
@@ -50,39 +50,40 @@ impl super::stub::CloudBuild for CloudBuild {
...
-                        &[Segment::SingleWildcard]
-                    )?,
-                );
+                let var_project_id = try_match(
+                    Some(&req).map(|m| &m.project_id).map(|s| s.as_str()),
+                    &[Segment::SingleWildcard],
+                )?;
+                let path = format!("/v1/projects/{}/builds", var_project_id,);
+
+                let _resource_name = format!("",);
```
coryan: It seems like the resource should be the project?


```suggestion
                let path = format!("/v1/projects/{}/builds", var_project_id,);

                let _resource_name = format!("projects/{}", var_project_id);
```

???

@haphungw @westarle

src/generated/cloud/secretmanager/v1/src/transport.rs
haphungw: yes you're right, `v1` is not supposed to be here. i'll look into it.

src/generated/devtools/cloudbuild/v1/src/transport.rs
haphungw: i've only been looking at the results for `compute` and `dns`. will investigate further.

PR #4930, opened 2026-03-06
docs(pubsub): improve flush documentation

suzmue:
This rework of the documentation attempts to make the wording and the example more concise. It also now tries to guide users to await on the PublishFutures instead.

### DIFF
--- a/src/pubsub/src/publisher/client.rs
+++ b/src/pubsub/src/publisher/client.rs
@@ -108,11 +108,12 @@ impl Publisher {
     ///
     /// ### Recommendations
     ///
-    /// *   **Individual Confirmation:** For most use cases, it is better to `.await`
+    /// *   For most use cases, we recommend you `.await`
     ///     the [`PublishFuture`][crate::publisher::PublishFuture] returned by
-    ///     [`publish`][Self::publish] to retrieve message IDs and handle specific errors.
-    /// *   **Collective Flush:** Use `flush()` for collective confirmation or as a
-    ///     convenience during application shutdown to ensure all outstanding data is sent.
+    ///     [`publish`][Self::publish] to retrieve message IDs and handle
+    ///     specific errors.
+    /// *   Use `flush()` as a convenience during application shutdown to
+    ///     ensure the client attempts to send all outstanding data.
     pub async fn flush(&self) {
         let (tx, rx) = oneshot::channel();
         if self.tx.send(ToDispatcher::Flush(tx)).is_ok() {


src/pubsub/src/publisher/client.rs
```diff
@@ -90,42 +90,29 @@ impl Publisher {
...
-    /// // The results for handle1 and handle2 are available.
-    /// let id1 = handle1.await?;
-    /// let id2 = handle2.await?;
     /// # Ok(())
     /// # }
     /// ```
+    ///
+    /// This method bypasses configured batching delays and returns only after all
+    /// messages buffered at the time of the call have reached a terminal state
+    /// (success or permanent failure).
```
coryan, line 107: Optional:

```suggestion
    /// messages buffered at the time of the call have been successfully delivered,
    /// the retry policy determines the error is permanent, or the retry policy has
    /// been exhausted.
```

src/pubsub/src/publisher/client.rs
```diff
@@ -90,42 +90,29 @@ impl Publisher {
...
+    ///
+    /// ### Recommendations
+    ///
+    /// *   **Individual Confirmation:** For most use cases, it is better to `.await`
+    ///     the [`PublishFuture`][crate::publisher::PublishFuture] returned by
+    ///     [`publish`][Self::publish] to retrieve message IDs and handle specific errors.
+    /// *   **Collective Flush:** Use `flush()` for collective confirmation or as a
+    ///     convenience during application shutdown to ensure all outstanding data is sent.
```
coryan: I am not sure what "collective flush" or "collective confirmation" are [^1].

```suggestion
    /// *   For most use cases, we recommend you `.await`
    ///     the [`PublishFuture`][crate::publisher::PublishFuture] returned by
    ///     [`publish`][Self::publish] to retrieve message IDs and handle specific errors.
    /// *   Use `flush()` as a convenience during application shutdown to ensure all
    ///      outstanding data is sent.
```



[^1]: Sounds like something the communist cadres would ask you to commit to, but maybe I have been reading too much 20th-century history.

dbolduc: nit: maybe "to ensure the client attempts to send all outstanding data." ?

`flush()` can't tell you if the data was actually sent or not.

src/pubsub/src/publisher/client.rs
suzmue: I skipped this suggestion for now.

src/pubsub/src/publisher/client.rs
suzmue: done.

Issue #4924, opened 2026-03-06
Consistent definitions and for attributes across signals


coryan:
I have introduced some inconsistency in attribute names and values for the different signals (more or less intentionally, I wanted to see all the signals working, even if not "perfect").  Including:

- The `rpc.method` should be the name of the gRPC method, as in `google.cloud.secretmanager.v1.SecretManager/CreateSecret`
- Errors should include the `error.type` field

westarle:
I updated this issue with some additional updates/gaps in traces, metrics.

haphungw:
#4937: Include error.type field on errors.
#5199: Add the gcp.client.* keys to metrics and handle the Schema URL requirement.
#5200: Replaced `rpc.system` with `rpc.system.name` and `gcp.resource.name` with `gcp.resource.destination.id`.
#5208: Remove `rpc.service` and ensure `rpc.method` carries the standardized format.

Closing this out.

PR #4937, opened 2026-03-07
impl(gax-internal): set the client span error attributes

coryan:
Part of the work for #4924 adds some missing fields to the client request spans.

### DIFF
--- a/src/gax-internal/src/observability/client_signals/with_client_signals.rs
+++ b/src/gax-internal/src/observability/client_signals/with_client_signals.rs
@@ -19,10 +19,17 @@
 use super::DurationMetric;
 use super::RequestStart;
 use crate::observability::attributes::RPC_SYSTEM_HTTP;
-use crate::observability::attributes::keys::{RPC_RESPONSE_STATUS_CODE, RPC_SYSTEM_NAME};
+use crate::observability::attributes::keys::OTEL_STATUS_DESCRIPTION;
+use crate::observability::attributes::keys::{
+    OTEL_STATUS_CODE, RPC_RESPONSE_STATUS_CODE, RPC_SYSTEM_NAME,
+};
+use crate::observability::attributes::otel_status_codes;
+use crate::observability::errors::ErrorType;
 use google_cloud_gax::error::Error;
 use google_cloud_gax::error::rpc::Code;
-use opentelemetry_semantic_conventions::attribute;
+use opentelemetry_semantic_conventions::attribute::{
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/observability/client_signals/with_client_signals.rs
```diff
@@ -108,10 +115,20 @@ where
         let this = self.project();
         let output = futures::ready!(this.inner.poll(cx));
         // Record the metric and log the value in the context of the span.
-        let _enter = span.entered();
+        let span = span.entered();
         match &output {
-            Ok(_) => this.metric.record_ok(&this.start),
+            Ok(_) => {
+                span.record(OTEL_STATUS_CODE, otel_status_codes::OK);
```
westarle, line 121: please leave `UNSET`

Maybe we should delete this status code.

coryan: I don't think that applies here. First, because if we don't set it, nothing will (and nothing does). The span is gone, we created it, we own it, we are the last one with access.

And second because I don't think that definition applies:

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/overview.md#instrumentation-libraries

https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/glossary.md#instrumentation-library

This code is not the instrumentation library, it is the instrumented library... That rule would apply to something like `tracing-opentelemetry` or `opentelemetry-otlp`.

westarle: I also found this confusing.  I think it's fair to say the app can't change this.

See https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status with a similar decision (that motivated me to raise this).

The examples of instrumentation library are things like attachable gRPC or mongodb instrumentation.  I think this statement describes our client libs, so I think when we look at the spec we should think of ourselves as Instrumentation Library.

> Instrumented Library and Instrumentation Library may be the same library if it has built-in OpenTelemetry instrumentation.

The definition of "Ok":

> Ok
> The operation has been validated by an **Application developer or Operator** to have completed successfully... When span status is set to Ok it SHOULD be considered final

I think most spans are 'unset' (treated as success downstream) and 'ok' is just used to override 'error'.

I don't really see it as a big deal outside of interpreting the spec.  The only 2 practical problems I can imagine are:
* some downstream processor extension (within tracing-opentelemetry or later) might find it more difficult to label this span an "error" since OK is "final".
* some downstream dashboard might put some green check mark next to these that you won't find next to the 'unset' HTTP spans

coryan: I think you convinced me.  I opened #4938 and will commit this to unblock other work.

PR #4951, opened 2026-03-09
cleanup(pubsub): add must_use attribute to publish

### LINKED ISSUE #4950: Mark publish() as must_use
Ignoring the result of a publish operation is an antipattern in Pub/Sub. Mark the future as "must_use" so that it needs to be explicitly ignored.

suzmue:
Fixes #4950

### DIFF
--- a/src/pubsub/src/publisher/client.rs
+++ b/src/pubsub/src/publisher/client.rs
@@ -492,7 +492,8 @@ mod tests {
             Message::new().set_data("world"),
         ];
         for msg in messages {
-            let _ = publisher.publish(msg.clone());
+            let handle = publisher.publish(msg.clone());
+            drop(handle);
         }
 
         publisher.flush().await;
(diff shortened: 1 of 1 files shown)


src/pubsub/src/publisher/client.rs
```diff
@@ -491,7 +492,8 @@ mod tests {
             Message::new().set_data("world"),
         ];
         for msg in messages {
-            publisher.publish(msg.clone());
+            let handle = publisher.publish(msg.clone());
+            drop(handle);
```
dbolduc, line 496: will clippy let us say:

```suggestion
            let _ = publisher.publish(msg.clone());
```

dbolduc (approved):

src/pubsub/src/publisher/client.rs
suzmue: No. We can say `let _handle = publisher.publish(msg.clone());`

suzmue: This is because of a clippy that only applies to futures: https://github.com/googleapis/google-cloud-rust/issues/4950#issuecomment-4026721023

dbolduc: Ah TIL

PR #4952, opened 2026-03-09
feat(spanner): add execute_query method

olavloite:
- Adds a method for creating a single-use read-only transaction and for executing a query using this transaction.
- Adds an integration test using the Spanner emulator to execute a query to show that the client works end-to-end.

### DIFF
--- a/src/spanner/src/client.rs
+++ b/src/spanner/src/client.rs
@@ -17,6 +17,7 @@ use crate::server_streaming::builder;
 use gaxi::options::{ClientConfig, Credentials};
 
 pub use crate::database_client::DatabaseClient;
+pub use crate::error::SpannerInternalError;
 pub use crate::from_value::{ConvertError, FromValue};
 pub use crate::read_only_transaction::SingleUseReadOnlyTransaction;
 pub use crate::read_only_transaction::SingleUseReadOnlyTransactionBuilder;
--- a/src/spanner/src/read_only_transaction.rs
+++ b/src/spanner/src/read_only_transaction.rs
@@ -153,7 +153,7 @@ impl SingleUseReadOnlyTransaction {
         let stream = self
             .client
             .spanner
-            // TODO: make request options configurable
+            // TODO(#4972): make request options configurable
             .execute_streaming_sql(request, crate::RequestOptions::default())
             .send()
             .await?;
--- a/src/spanner/src/result_set.rs
+++ b/src/spanner/src/result_set.rs
@@ -148,7 +148,7 @@ impl ResultSet {
 }
 
 fn internal_error(message: &str) -> crate::Error {
-    crate::Error::deser(message)
+    crate::Error::deser(crate::error::SpannerInternalError::new(message))
 }
 
 /// Merges two values from successive `PartialResultSet`s into a single value.
--- a/src/spanner/src/server_streaming/stream.rs
+++ b/src/spanner/src/server_streaming/stream.rs
@@ -12,6 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+// TODO(#4969)
 #![allow(dead_code)]
 
 use crate::google::spanner::v1::BatchWriteResponse;
--- a/src/spanner/src/statement.rs
+++ b/src/spanner/src/statement.rs
@@ -35,7 +35,7 @@
 #[derive(Clone, Debug, PartialEq)]
 pub struct Statement {
     pub sql: String,
-    // TODO: Add support for query parameters.
+    // TODO(#4970): Add support for query parameters.
 }
 
 impl Statement {
--- a/src/spanner/src/timestamp_bound.rs
+++ b/src/spanner/src/timestamp_bound.rs
@@ -39,21 +39,27 @@ impl TimestampBound {
     /// Returns a strong timestamp bound. Strong reads are guaranteed to see the
     /// effects of all transactions that have committed before the start of the read.
     ///
-    /// See <https://docs.cloud.google.com/spanner/docs/timestamp-bounds#strong> for more information.
+    /// See [timestamp_bound_strong] for more information.
+    ///
+    /// [timestamp_bound_strong]: https://docs.cloud.google.com/spanner/docs/timestamp-bounds#strong
     pub fn strong() -> Self {
         Self(ReadOnlyTimestampBound::Strong(true))
     }
 
     /// Returns a timestamp bound for an exact timestamp. The data will be read as it was at the given timestamp.
     ///
-    /// See <https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness> for more information.
+    /// See [timestamp_bound_exact_staleness] for more information.
+    ///
+    /// [timestamp_bound_exact_staleness]: https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness
     pub fn read_timestamp(timestamp: OffsetDateTime) -> Self {
         Self::try_read_timestamp(timestamp).expect("timestamp out of range")
     }
 
     /// Returns a timestamp bound for an exact timestamp, returning an error if the timestamp is out of range.
     ///
-    /// See <https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness> for more information.
+    /// See [timestamp_bound_exact_staleness] for more information.
+    ///
+    /// [timestamp_bound_exact_staleness]: https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness
     pub fn try_read_timestamp(
         timestamp: OffsetDateTime,
     ) -> Result<Self, crate::client::TimestampError> {
@@ -66,14 +72,18 @@ impl TimestampBound {
     /// Returns a timestamp bound for a minimum read timestamp. The data will be read as it was at the
     /// given timestamp or later.
     ///
-    /// See <https://docs.cloud.google.com/spanner/docs/timestamp-bounds#bounded_staleness> for more information.
+    /// See [timestamp_bound_bounded_staleness] for more information.
+    ///
+    /// [timestamp_bound_bounded_staleness]: https://docs.cloud.google.com/spanner/docs/timestamp-bounds#bounded_staleness
     pub fn min_read_timestamp(timestamp: OffsetDateTime) -> Self {
         Self::try_min_read_timestamp(timestamp).expect("timestamp out of range")
     }
 
     /// Returns a timestamp bound for a minimum read timestamp, returning an error if the timestamp is out of range.
     ///
-    /// See <https://docs.cloud.google.com/spanner/docs/timestamp-bounds#bounded_staleness> for more information.
+    /// See [timestamp_bound_bounded_staleness] for more information.
+    ///
+    /// [timestamp_bound_bounded_staleness]: https://docs.cloud.google.com/spanner/docs/timestamp-bounds#bounded_staleness
     pub fn try_min_read_timestamp(
         timestamp: OffsetDateTime,
     ) -> Result<Self, crate::client::TimestampError> {
@@ -86,14 +96,18 @@ impl TimestampBound {
     /// Returns a timestamp bound for an exact staleness. The data will be read as it was at the given timestamp
     /// calculated by the current server time minus the given duration.
     ///
-    /// See <https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness> for more information.
+    /// See [timestamp_bound_exact_staleness] for more information.
+    ///
+    /// [timestamp_bound_exact_staleness]: https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness
     pub fn exact_staleness(duration: Duration) -> Self {
         Self::try_exact_staleness(duration).expect("duration out of range")
     }
 
     /// Returns a timestamp bound for an exact staleness, returning an error if the duration is out of range.
     ///
-    /// See <https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness> for more information.
+    /// See [timestamp_bound_exact_staleness] for more information.
+    ///
+    /// [timestamp_bound_exact_staleness]: https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness
     pub fn try_exact_staleness(duration: Duration) -> Result<Self, crate::client::DurationError> {
         let duration = wkt::Duration::try_from(duration)?;
         Ok(Self(ReadOnlyTimestampBound::ExactStaleness(Box::new(
@@ -104,14 +118,18 @@ impl TimestampBound {
     /// Returns a timestamp bound for a maximum staleness. The data will be read as it was at the
     /// current server time minus the given duration or later.
     ///
-    /// See <https://docs.cloud.google.com/spanner/docs/timestamp-bounds#bounded_staleness> for more information.
+    /// See [timestamp_bound_bounded_staleness] for more information.
+    ///
+    /// [timestamp_bound_bounded_staleness]: https://docs.cloud.google.com/spanner/docs/timestamp-bounds#bounded_staleness
     pub fn max_staleness(duration: Duration) -> Self {
         Self::try_max_staleness(duration).expect("duration out of range")
     }
 
     /// Returns a timestamp bound for a maximum staleness, returning an error if the duration is out of range.
     ///
-    /// See <https://docs.cloud.google.com/spanner/docs/timestamp-bounds#bounded_staleness> for more information.
+    /// See [timestamp_bound_bounded_staleness] for more information.
+    ///
+    /// [timestamp_bound_bounded_staleness]: https://docs.cloud.google.com/spanner/docs/timestamp-bounds#bounded_staleness
     pub fn try_max_staleness(duration: Duration) -> Result<Self, crate::client::DurationError> {
         let duration = wkt::Duration::try_from(duration)?;
         Ok(Self(ReadOnlyTimestampBound::MaxStaleness(Box::new(
--- a/tests/spanner/tests/driver.rs
+++ b/tests/spanner/tests/driver.rs
@@ -44,7 +44,7 @@ mod spanner {
     // Provisions the Spanner Emulator with a test instance and database.
     // ALREADY_EXISTS errors that are returned when creating an instance or database are ignored.
     async fn provision_emulator(endpoint: &str) {
-        // TODO: Re-write this to use the admin clients once those also support the Emulator.
+        // TODO(#4973): Re-write this to use the admin clients once those also support the Emulator.
         let rest_endpoint = endpoint.replace("9010", "9020");
         let rest_endpoint =
             if rest_endpoint.starts_with("http://") || rest_endpoint.starts_with("https://") {
--- a/src/spanner/src/error.rs
+++ b/src/spanner/src/error.rs
@@ -1,0 +1,36 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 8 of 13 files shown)


src/spanner/src/client.rs
```diff
@@ -64,7 +70,24 @@ pub type ClientBuilder = google_cloud_gax::client_builder::ClientBuilder<Factory
 #[allow(dead_code)]
 impl Spanner {
     pub fn builder() -> ClientBuilder {
-        google_cloud_gax::client_builder::internal::new_builder(Factory)
+        let mut builder = google_cloud_gax::client_builder::internal::new_builder(Factory);
+        // The Spanner client should automatically use the Spanner emulator if the
+        // SPANNER_EMULATOR_HOST environment variable is set.
+        if let Ok(endpoint) = std::env::var("SPANNER_EMULATOR_HOST") {
```
coryan: Consider:


```suggestion
        let Ok(endpoint) = std::env::var("SPANNER_EMULATOR_HOST") else {
            return builder;
        };
```

Saves you some indentation.

src/spanner/src/client.rs
```diff
@@ -64,7 +70,24 @@ pub type ClientBuilder = google_cloud_gax::client_builder::ClientBuilder<Factory
...
+            if !endpoint.is_empty() {
+                // Determine if we need to prefix the endpoint with a scheme
+                let full_endpoint = if endpoint.starts_with("http") {
```
coryan: Hmmm... What about "http_carlos_is_evil.com" ? If you need to parse the endpoint, consider the `url` crate, which I think we already link.

src/spanner/src/client.rs
```diff
@@ -64,7 +70,24 @@ pub type ClientBuilder = google_cloud_gax::client_builder::ClientBuilder<Factory
...
+                    endpoint
+                } else {
+                    format!("http://{}", endpoint)
+                };
+
+                builder = builder.with_endpoint(full_endpoint).with_credentials(
+                    google_cloud_auth::credentials::anonymous::Builder::new().build(),
+                );
```
coryan: An alternative would be:


```suggestion
    let anonymous = || google_cloud_auth::credentials::anonymous::Builder::new().build();
    builder = match endpoint {
        e if e.starts_with("https://) => builder.with_endpoint(e).with_credentials(anonymous()),
        e if e.starts_with("http://) => builder.with_endpoint(e).with_credentials(anonymous()),
        e if !e.is_empty() => builder.with_endpoint(format!("http://{e}").with_credentials(anonymous()),
        _ => builder,
    }
```

.gcb/integration.yaml
```diff
@@ -37,8 +37,13 @@ steps:
     env:
       - _SCCACHE_VERSION=${_SCCACHE_VERSION}
       - _SCCACHE_SHA256=${_SCCACHE_SHA256}
+  - id: 'Start Spanner Emulator'
+    name: 'gcr.io/cloud-spanner-emulator/emulator'
+    waitFor: ['-']
```
coryan, line 43: This is fine for CI, but now I need to install the spanner emulator to do `cargo test --all-features`, event if I am working on something unrelated.

src/spanner/src/read_only_transaction.rs
```diff
@@ -0,0 +1,320 @@
...
+use crate::model::transaction_options::ReadOnly;
+use crate::model::transaction_options::read_only::TimestampBound as ReadOnlyTimestampBound;
+use crate::result_set::ResultSet;
+use crate::statement::Statement;
+use crate::timestamp_bound::TimestampBound;
+
+/// A builder for [SingleUseReadOnlyTransaction].
+///
+/// # Example
+/// ```rust,no_run
```
coryan:
```suggestion
/// ```
```

That should be enough as the example only defines a function, does not invoke it.

src/spanner/src/read_only_transaction.rs
```diff
@@ -0,0 +1,320 @@
...
+        Self {
+            client,
+            timestamp_bound: None,
+        }
+    }
+
+    /// Sets the timestamp bound for the read-only transaction.
+    ///
+    /// # Example
+    /// ```rust,no_run
```
coryan: Here and below:


```suggestion
    /// ```
```

src/spanner/src/result_set.rs
```diff
@@ -0,0 +1,565 @@
...
+    }
+
+    /// Fetches the next row from the result set.
+    ///
+    /// # Example
+    /// ```rust,no_run
+    /// # use google_cloud_spanner::client::{ResultSet, Row};
+    /// # async fn fetch_next(mut rs: ResultSet) -> Result<(), google_cloud_spanner::Error> {
+    /// if let Some(row) = rs.next().await {
+    ///     let row: Row = row?;
```
coryan: You can also write:

```suggestion
    /// if let Some(row) = rs.next().await.transpose()? {
```

src/spanner/src/result_set.rs
```diff
@@ -0,0 +1,565 @@
...
+    ///     // Process the row
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// Returns `None` when all rows have been returned.
```
coryan: nit: I think this reads better.


```suggestion
    /// Returns `None` when all rows have been retrieved.
```

src/spanner/src/result_set.rs
```diff
@@ -0,0 +1,565 @@
...
+                        .as_ref()
+                        .map(|rt| rt.fields.len())
+                        .unwrap_or(0);
+                    self.metadata = Some(metadata);
+                } else {
+                    return Some(Err(internal_error(
+                        "First PartialResultSet did not contain metadata",
+                    )));
+                }
+            }
```
coryan: Consider a match, which forces you to think about one extra case:

```suggestion
            match (self.metadata.as_ref(), prs.metadata) {
                (Some(_), None)) => {},
                (None, None) => return Some(Err(internal_error("First PartialResultSet did not contain metadata")),
                (Some(_), Some(m)) => return Some(Err(internal_err("Additional metadata after first result set"))),
                (None, Some(m)) => {
                    self.column_count = m
                        .row_type
                        .as_ref()
                        .map(|rt| rt.fields.len())
                        .unwrap_or_default();
                    self.metadata = Some(metadata);
                },
```

src/spanner/src/result_set.rs
```diff
@@ -0,0 +1,565 @@
...
+        None
+    }
+}
+
+fn internal_error(message: &str) -> crate::Error {
+    crate::Error::service(
+        google_cloud_gax::error::rpc::Status::default()
+            .set_code(google_cloud_gax::error::rpc::Code::Internal as i32)
+            .set_message(message.to_string()),
+    )
```
coryan: I don't think these are service errors as we define them (an error *reported* by the service), I think you want `Error::deser()` for these: the server sent you data that you cannot deserialize.

coryan: Or you can define your own error type, that may be the better option.

src/spanner/src/timestamp_bound.rs
```diff
@@ -0,0 +1,210 @@
...
+        Self::try_read_timestamp(timestamp).expect("timestamp out of range")
+    }
+
+    /// Returns a timestamp bound for an exact timestamp, returning an error if the timestamp is out of range.
+    ///
+    /// See https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness for more information.
+    pub fn try_read_timestamp(timestamp: OffsetDateTime) -> crate::Result<Self> {
+        let seconds = timestamp.unix_timestamp();
+        let nanos = timestamp.nanosecond();
+        let timestamp = wkt::Timestamp::new(seconds, nanos as i32)
```
coryan: FYI, we regret using the `wkt::` alias.  It is short, but it makes for confusing examples and links to source in the docs.rs documentation.  We will remove them one of these days.

src/spanner/src/timestamp_bound.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// See https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness for more information.
+    pub fn read_timestamp(timestamp: OffsetDateTime) -> Self {
+        Self::try_read_timestamp(timestamp).expect("timestamp out of range")
+    }
+
+    /// Returns a timestamp bound for an exact timestamp, returning an error if the timestamp is out of range.
+    ///
+    /// See https://docs.cloud.google.com/spanner/docs/timestamp-bounds#exact_staleness for more information.
+    pub fn try_read_timestamp(timestamp: OffsetDateTime) -> crate::Result<Self> {
+        let seconds = timestamp.unix_timestamp();
```
coryan: Consider this:

https://docs.rs/google-cloud-wkt/latest/google_cloud_wkt/struct.Timestamp.html#impl-TryFrom%3COffsetDateTime%3E-for-Timestamp

You can add `google-cloud-wkt = { workspace = true, features = ["time"] }` to your `Cargo.toml` file and get the time helpers.

src/spanner/src/timestamp_bound.rs
coryan: Consider this:

https://docs.rs/google-cloud-wkt/latest/google_cloud_wkt/struct.Timestamp.html#impl-TryFrom%3COffsetDateTime%3E-for-Timestamp

You could re-export the error type too and return a non-stringly-typed error from this function.

src/spanner/src/timestamp_bound.rs
```diff
@@ -0,0 +1,210 @@
...
+        let nanos = timestamp.nanosecond();
+        let timestamp = wkt::Timestamp::new(seconds, nanos as i32)
+            .map_err(|e| crate::Error::binding(format!("timestamp out of range: {}", e)))?;
```
coryan: `Error::binding()` is intended for an RPC that is missing the inputs for a HTTP request.  In general, google_cloud_gax::error::Error is intended for RPC errors. Please avoid using it elsewhere.

tests/spanner/tests/driver.rs
```diff
@@ -0,0 +1,340 @@
...
+    // ALREADY_EXISTS errors that are returned when creating an instance or database are ignored.
+    async fn provision_emulator(endpoint: &str) {
+        // TODO: Re-write this to use the admin clients once those also support the Emulator.
+        let rest_endpoint = endpoint.replace("9010", "9020");
+        let rest_endpoint =
+            if rest_endpoint.starts_with("http://") || rest_endpoint.starts_with("https://") {
+                rest_endpoint
+            } else {
+                format!("http://{}", rest_endpoint)
+            };
```
coryan: Consider using `url::Url` to make this parsing and modifications to the Url.

.gcb/integration.yaml
coryan: Oh I see, you skip the tests if the environment variable is not set. SGTM, thanks.

tests/spanner/tests/driver.rs
```diff
@@ -0,0 +1,340 @@
...
+        }
+        if !connected {
+            panic!("Failed to connect to emulator at {}", endpoint);
+        }
+    }
+
+    // Provisions the Spanner Emulator with a test instance and database.
+    // ALREADY_EXISTS errors that are returned when creating an instance or database are ignored.
+    async fn provision_emulator(endpoint: &str) {
+        // TODO: Re-write this to use the admin clients once those also support the Emulator.
```
coryan: You can certainly override the endpoint and the authentication for them.

tests/spanner/tests/driver.rs
```diff
@@ -0,0 +1,340 @@
...
+"#;
+
+        let stmt = Statement::new(sql);
+        let mut rs = rot
+            .execute_query(stmt)
+            .await
+            .expect("Failed to execute query");
+
+        let mut count = 0;
+        while let Some(row_result) = rs.next().await {
```
coryan: Consider:


```suggestion
        let mut rows = Vec::new();
        while let Some(row) = rs.next().await.transpose()? {
            rows.push_back(row)
        }
        let (row1, row2, row3) = match &rows[..] {
            [r1, r2, r3] => (r1, r2, r3),
            _ => panic!("unexpected number of rows, got={}, want=3\n{rows:?}", rows.len());
        };
```

And then you can make assertions (maybe in helper functions?) about row1, row2 and row3.  You could also capture the results if that is useful.

tests/spanner/tests/driver.rs
```diff
@@ -0,0 +1,340 @@
...
+                1 => {
+                    // Assert NULL row
+                    // Spanner sorts NULLs first.
+                    assert!(
+                        raw_values.iter().all(|v| v.kind() == Kind::Null),
+                        "Expected all columns to be NULL"
+                    );
+                }
+                2 => {
+                    // Assert row 1
```
coryan: Ugh. I wished there was a better way to write these...

src/spanner/src/client.rs

src/spanner/src/client.rs
olavloite: I've used the `url` crate to parse and check the URL, and only add a scheme if there is no scheme.

src/spanner/src/client.rs
olavloite: This is solved in a slightly different way with the `url` crate.

src/spanner/src/read_only_transaction.rs
olavloite: Removed here and elsewhere

src/spanner/src/read_only_transaction.rs
olavloite: Removed

src/spanner/src/result_set.rs

src/spanner/src/result_set.rs

src/spanner/src/result_set.rs
olavloite: Good tip, thanks. Changed.

src/spanner/src/result_set.rs
olavloite: I've changed this to use `Error::deser()`. I don't think we need our own error type for this. These are conditions that should never happen.

src/spanner/src/timestamp_bound.rs
olavloite: Ack

src/spanner/src/timestamp_bound.rs
olavloite: Changed this and the duration parsing to use the `TryFrom` implementations.

src/spanner/src/timestamp_bound.rs
olavloite: Done

src/spanner/src/timestamp_bound.rs
olavloite: Changed. I'm now re-exporting the error from the try_from implementation.

tests/spanner/tests/driver.rs
olavloite: Done.

src/spanner/src/server_streaming/stream.rs
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+#![allow(dead_code)]
```
coryan, line 16: nit: these are easy to forget, consider opening an issue to track them / remind yourself.

FWIW, we use `TODO(#NNNN)` where NNNN is the GitHub issue number.

tests/spanner/tests/driver.rs
olavloite: Ack. In this case I meant that we should also have those clients automatically pick up the SPANNER_EMULATOR_HOST environment variable. That is the standard that we use in all the client libraries: https://docs.cloud.google.com/spanner/docs/emulator#client-libraries

tests/spanner/tests/driver.rs
olavloite: I've refactored this along these lines, including adding some helper methods to verify the row contents.

tests/spanner/tests/driver.rs
olavloite: Moved to helper methods.

src/spanner/src/statement.rs
```diff
@@ -0,0 +1,56 @@
...
+///     let row = row?;
+///     // process row
+/// }
+/// # Ok(())
+/// # }
+/// ```
+#[derive(Clone, Debug, PartialEq)]
+pub struct Statement {
+    pub sql: String,
+    // TODO: Add support for query parameters.
```
coryan: nit: `TODO(#KKKK)` would be more in the style of this repo.

src/spanner/src/result_set.rs
```diff
@@ -0,0 +1,563 @@
...
+            return Some(Err(internal_error("Stream ended with chunked_value=true")));
+        }
+
+        None
+    }
+}
+
+fn internal_error(message: &str) -> crate::Error {
+    crate::Error::deser(message)
+}
```
coryan, line 152: Food for thought. You may consider something like:


```suggestion
fn internal_error(e: MyErrorType) -> crate::Error {
    crate::Error::deser(e)
}
```

And then application developers can use [Error::source()](https://doc.rust-lang.org/std/error/trait.Error.html#method.source) to get more details about the problem. 

In this case it is unclear how that would be better than an error message, as the application can do very little with such an error other than logging.

Oh, you may want to include something in these messages about where to file bugs. That was either a bug in the library or the service, the customer may need some hints on how to proceed.

src/spanner/src/result_set.rs
```diff
@@ -0,0 +1,563 @@
...
+///
+/// Cloud Spanner can return a single logical row or column value split across multiple
+/// `PartialResultSet` messages. This occurs when a value (especially large strings or
+/// arrays) exceeds the message size limits of the underlying stream. In these cases,
+/// the `chunked_value` flag is set on the first `PartialResultSet`, indicating that the
+/// final value in the message's `values` array is incomplete and must be combined with
+/// the first value in the `values` array of the subsequent `PartialResultSet`.
+///
+/// This function handles the concatenation of split `StringValue` and `ListValue` types.
+fn merge_values(target: &mut prost_types::Value, source: prost_types::Value) -> crate::Result<()> {
```
coryan, line 164: As a reminder: you are the Spanner expert, I am not an effective reviewer as to the correctness of this code. If you want an (informed) second opinion maybe ask one of your colleagues to do a drive-by.

src/spanner/src/timestamp_bound.rs
```diff
@@ -0,0 +1,206 @@
...
+///
+/// See <https://cloud.google.com/spanner/docs/timestamp-bounds> for more information.
+#[derive(Clone, Debug)]
+pub struct TimestampBound(pub(crate) ReadOnlyTimestampBound);
+
+impl TimestampBound {
+    /// Returns a strong timestamp bound. Strong reads are guaranteed to see the
+    /// effects of all transactions that have committed before the start of the read.
+    ///
+    /// See <https://docs.cloud.google.com/spanner/docs/timestamp-bounds#strong> for more information.
```
coryan: You can also write:


```suggestion
    /// See the [Timestamp bound types] for more information.
    ///
    /// [Timestamp bound types]: https://docs.cloud.google.com/spanner/docs/timestamp-bounds#strong
```

Which may read better in the rendered documentation (try `cargo doc --no-deps -p google-cloud-spanner` locally).

tests/spanner/tests/driver.rs
coryan: Yes.  That would probably require a change in the generator (nothing too bad I think).

I was suggesting that you can still use the admin libraries today, albeit you would need to manually override the endpoint and credentials until they suppose that environment variable.

src/spanner/src/server_streaming/stream.rs
olavloite: Opened an issue and added a TODO referencing it here and elsewhere.

src/spanner/src/result_set.rs
olavloite: I added an `SpannerInternalError` enum following the example from the Storage and PubSub clients, and added a hint to the documentation regarding filing an issue (example also taken from the Storage client). The internal error is then wrapped in `crate::Error::deser`.

src/spanner/src/result_set.rs
olavloite: I've asked a colleague from the Spanner team to take a look.

PR #4953, opened 2026-03-09
docs(pubsub): subscriber throughput example

dbolduc:
Part of the work for #4508 

Add an example demonstrating how to use the subscriber client effectively.

This probably belongs in a user guide, but until we have that set up, let's carve out a section for it in the class documentation.


dbolduc:
This belongs somewhere, but not here. Closing.

PR #4955, opened 2026-03-09
test(integration-tests-o11y): test for all signals

coryan:
Introduce a test to validate all the signals (traces, logs and metrics) using a single client. This uses an `EventLogger` to format the logs, when we write an OTLP exporter for the logs we will need to change the test too.

Part of the work for #4760 and #4795

### DIFF
--- a/tests/o11y/src/e2e/signals.rs
+++ b/tests/o11y/src/e2e/signals.rs
@@ -300,16 +300,8 @@ async fn check_metrics(
             .await
             .inspect_err(|e| tracing::info!("error fetching metric: {e:?}"));
 
-        let Some(response) = response.ok() else {
-            continue;
-        };
-        let contains_resource = response.time_series.iter().any(|ts| {
-            ts.resource
-                .as_ref()
-                .is_some_and(|r| r.labels.get("node_id") == Some(&node_id))
-        });
-        if contains_resource {
-            found = Some(response);
+        found = response.ok();
+        if found.is_some() {
             break;
         }
     }
@@ -318,11 +310,6 @@ async fn check_metrics(
     let metrics = found
         .time_series
         .iter()
-        .filter(|ts| {
-            ts.resource
-                .as_ref()
-                .is_some_and(|r| r.labels.get("node_id") == Some(&node_id))
-        })
         .filter_map(|ts| ts.metric.as_ref())
         .collect::<Vec<_>>();
 


tests/o11y/src/otlp/logs.rs
```diff
@@ -0,0 +1,175 @@
...
+use std::fmt::Result as FmtResult;
+use tracing::{Event, Subscriber};
+use tracing_opentelemetry::OtelData;
+use tracing_serde::AsSerde;
+use tracing_serde::fields::AsMap;
+use tracing_subscriber::fmt::format::{FormatEvent, Writer};
+use tracing_subscriber::fmt::{FmtContext, FormatFields};
+use tracing_subscriber::registry::LookupSpan;
+
+/// Format [tracing] events into the format consumed by the Google Cloud [Ops Agent].
```
dbolduc, line 28: comment: this thing seems useful to our customers too. Not just us.

tests/o11y/src/e2e/signals.rs
```diff
@@ -0,0 +1,376 @@
...
+    opentelemetry::global::set_meter_provider(meter_provider.clone());
+    Ok((tracer_provider, meter_provider, buffer))
+}
+
+fn configure_mock_server() -> Server {
+    let echo_server = Server::run();
+    let respond = || {
+        const DELAY: Duration = Duration::from_millis(50);
+        match rand::rng().random_range(0..1000) {
+            n if n == 37 * 31 => delay_and_then(DELAY, status_code(404)),
```
dbolduc: 37 * 31 = 1147 > 1000. 😬 

This never gets hit.

tests/o11y/src/e2e/signals.rs
```diff
@@ -0,0 +1,376 @@
...
+            .await
+            .inspect_err(|e| tracing::info!("error fetching metric: {e:?}"));
+
+        let Some(response) = response.ok() else {
+            continue;
+        };
+        let contains_resource = response.time_series.iter().any(|ts| {
+            ts.resource
+                .as_ref()
+                .is_some_and(|r| r.labels.get("node_id") == Some(&node_id))
```
dbolduc: comment: ah, this is how we filter on metrics produced by the current build.

Q: can't we add this condition to the `set_filter` statement? If so, should be cleaner.

tests/o11y/src/e2e/signals.rs
```diff
@@ -0,0 +1,376 @@
...
+        })
+        .filter_map(|ts| ts.metric.as_ref())
+        .collect::<Vec<_>>();
+
+    let _code_404 = metrics
+        .iter()
+        .find(|m| {
+            m.labels
+                .get("http.response.status_code")
```
dbolduc, line 320: ... unless these don't have the `node_id` label.

If so, they'd get filtered out, and my above suggestion is bad.

dbolduc (approved):
fizzbuzz claims another victim 😛

tests/o11y/src/e2e/signals.rs
```diff
@@ -0,0 +1,376 @@
...
+    // Just want to make sure it exists and is a string.
+    assert!(
+        fields
+            .remove("message")
+            .is_some_and(|v| v.as_str().is_some()),
+        "{value:?}"
+    );
+    let want = serde_json::json!({
+        "http.response.status_code": 404,
+        "rpc.method": "google_cloud_showcase_v1beta1::client::Echo::echo",
```
westarle, line 252: nit: can this be "google.showcase.v1beta1.Echo/Echo"?

(when we do low-level gRPC/HTTP logs)

tests/o11y/src/otlp/logs.rs
```diff
@@ -0,0 +1,175 @@
...
+            serializer.serialize_entry("severity", &meta.level().as_serde())?;
+            serializer.serialize_entry("fields", &event.field_map())?;
+            serializer.serialize_entry("target", meta.target())?;
+            match self.trace_info(ctx, event) {
+                (Some((tid, sid)), sampled) => {
+                    serializer.serialize_entry(
+                        "logging.googleapis.com/trace",
+                        &format!("projects/{}/traces/{tid}", self.project_id),
+                    )?;
+                    serializer.serialize_entry("logging.googleapis.com/span", &sid.to_string())?;
```
westarle:
```suggestion
                    serializer.serialize_entry("logging.googleapis.com/spanId", &sid.to_string())?;
```

NB: `logging.googleapis.com/trace` is correct above.

https://docs.cloud.google.com/logging/docs/structured-logging#structured_logging_special_fields

tests/o11y/src/otlp/logs.rs
coryan: Yes. But I am not sure where it should go. @westarle this may also belong in https://github.com/open-telemetry/opentelemetry-rust-contrib, along with the auth layers for the exporters.

tests/o11y/src/otlp/logs.rs
coryan: Fixed.

tests/o11y/src/e2e/signals.rs
coryan: Good catch, fixed.

tests/o11y/src/e2e/signals.rs
```diff
@@ -0,0 +1,376 @@
...
+        ])
+        .times(1..)
+        .respond_with(respond),
+    );
+    echo_server
+}
+
+async fn run_test_iterations(project_id: &str, client: Echo) -> anyhow::Result<TraceId> {
+    let mut trace_id = None;
+    for _ in 0..1_000 {
```
coryan: I increased the time to catch more errors.

tests/o11y/src/e2e/signals.rs
coryan: Already part of #4924 I think.

tests/o11y/src/e2e/signals.rs
coryan: Yes, we can! I had tried before but could not figure out the right expression. I just did.

tests/o11y/src/e2e/signals.rs
```diff
@@ -0,0 +1,374 @@
...
+            )
+            .set_order_by("timestamp desc")
+            .send()
+            .await
+            .inspect_err(|e| tracing::info!("error fetching metric: {e:?}"));
+
+        let Some(response) = response.ok() else {
+            continue;
+        };
+        let contains_resource = response.time_series.iter().any(|ts| {
```
dbolduc: hopefully this is not needed

tests/o11y/src/e2e/signals.rs
```diff
@@ -0,0 +1,374 @@
...
+            found = Some(response);
+            break;
+        }
+    }
+
+    let found = found.expect("metric not exported after timeout");
+    let metrics = found
+        .time_series
+        .iter()
+        .filter(|ts| {
```
dbolduc: Same

tests/o11y/src/e2e/signals.rs
coryan: Fixed.

tests/o11y/src/e2e/signals.rs
coryan: Removed.

tests/o11y/src/e2e/signals.rs
coryan: I think you are fine, we only get things with a node_id label out of the loop.

Issue #4938, opened 2026-03-08
Keep the span status as `UNSET` even on success


coryan:
I also found this confusing.  I think it's fair to say the app can't change this.

See https://opentelemetry.io/docs/specs/semconv/http/http-spans/#status with a similar decision (that motivated me to raise this).

The examples of instrumentation library are things like attachable gRPC or mongodb instrumentation.  I think this statement describes our client libs, so I think when we look at the spec we should think of ourselves as Instrumentation Library.

> Instrumented Library and Instrumentation Library may be the same library if it has built-in OpenTelemetry instrumentation.

The definition of "Ok":

> Ok
> The operation has been validated by an **Application developer or Operator** to have completed successfully... When span status is set to Ok it SHOULD be considered final

I think most spans are 'unset' (treated as success downstream) and 'ok' is just used to override 'error'.

I don't really see it as a big deal outside of interpreting the spec.  The only 2 practical problems I can imagine are:
* some downstream processor extension (within tracing-opentelemetry or later) might find it more difficult to label this span an "error" since OK is "final".
* some downstream dashboard might put some green check mark next to these that you won't find next to the 'unset' HTTP spans

_Originally posted by @westarle in https://github.com/googleapis/google-cloud-rust/pull/4937#discussion_r2901184258_

PR #4967, opened 2026-03-10
fix(gax-internal): keep traces as `Unset` on success

coryan:
The convention / rules for OpenTelemetry are to keep traces as `Unset` on success.

Fixes https://github.com/googleapis/google-cloud-rust/issues/4938. See that bug for background and justification, because we all think it is not intuitive.

### DIFF
--- a/src/gax-internal/src/observability/attributes.rs
+++ b/src/gax-internal/src/observability/attributes.rs
@@ -126,10 +126,9 @@ pub mod error_type_values {
     pub const UNKNOWN: &str = "UNKNOWN";
 }
 
+// There is no OK constant because in practice an instrumented library should never use it.
 /// Values for the OpenTelemetry `otel.status_code` attribute.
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/observability/client_tracing.rs
```diff
@@ -50,12 +50,12 @@ impl<T> sealed::ResultExt for std::result::Result<T, Error> {}
 impl<T> ResultExt for std::result::Result<T, Error> {
     fn record_in_span(self, span: &Span) -> Self {
         match &self {
-            Ok(_) => span.record(OTEL_STATUS_CODE, otel_status_codes::OK),
```
westarle, line 53: nit: we could remove `otel_status_codes::OK` I'm not sure we ever get to use it.

coryan: Done.

PR #4979, opened 2026-03-10
cleanup(pubsub): future-proof `AckError`

### LINKED ISSUE #4809: Stabilize `AckError`

dbolduc:
Fixes #4809, related to #4508

We discussed whether the RPC branch should parse the string the server returns into an enum. My official stance on the matter is "why....".

We remove the `details` field for now, and mark the branch as non-exhaustive.

Non breaking, because `AckError` was never released.

### DIFF
--- a/src/pubsub/src/error.rs
+++ b/src/pubsub/src/error.rs
@@ -62,12 +62,9 @@ pub enum AckError {
     LeaseExpired,
 
     /// The underlying RPC failed.
-    #[error("the acknowledgement failed{}. RPC error: {source}",
-        .details.as_ref().map(|m| format!(", with the server reporting: {m}")).unwrap_or_default())]
+    #[non_exhaustive]
+    #[error("the acknowledgement failed. RPC error: {source}")]
     Rpc {
-        /// Extra details from the server on why the acknowledgement failed.
-        details: Option<String>,
-
         /// The error returned by the service for the acknowledge request.
         #[source]
         source: Arc<Error>,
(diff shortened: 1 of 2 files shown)


src/pubsub/src/error.rs
```diff
@@ -62,12 +62,9 @@ pub enum AckError {
     LeaseExpired,
 
     /// The underlying RPC failed.
-    #[error("the acknowledgement failed{}. RPC error: {source}",
-        .details.as_ref().map(|m| format!(", with the server reporting: {m}")).unwrap_or_default())]
+    #[non_exhaustive]
```
coryan, line 65: Terrible-ideas-discuss: we can always add `RpcWithDetails` later if needed, and basically never use `Rpc` once that is introduced.

dbolduc: Errr, but then applications that matched on `Rpc` change behavior silently when they update their `google-cloud-pubsub`. That seems avoidable.

PR #4990, opened 2026-03-11
impl(pubsub): add confirmed ack for lease state

PhongChuong:


PhongChuong:
This is largely based on @dbolduc 's implementation in https://github.com/googleapis/google-cloud-rust/commit/bafa17f526e7895a9a399113a4c11139b3cc2a61.

For the tests: [message_expiration, limit_size_of_extends], I manually verified it for exactly once leases. I'll follow up this PR with a test PR that would parameterize these tests as this PR is already quite large.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -120,7 +138,9 @@ where
     /// hold one mutable reference to `LeaseState` within its `select!`
     /// statement.
     pub(super) async fn next_event(&mut self) -> LeaseEvent {
-        if self.leases.needs_flush() {
+        if self.leases.needs_flush() || self.eo_leases.needs_flush() {
+            // This flushes both leases and eo_leases. This can be updated so
+            // that the [leases, eo_leases] are flushed independently.
```
dbolduc: comment: 99.999% of the time one of `leases` or `eo_leases` will be empty. For them both to be non-empty, the subscription's delivery would have to change midstream. That would be a weird app.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -177,17 +219,27 @@ where
...
         if !to_ack.is_empty() {
             self.leaser.ack(to_ack).await;
         }
+        if !to_nack.is_empty() {
+            self.leaser.nack(to_nack).await;
+        }
 
-        // TODO(#4847) - this nack needs to be broken into batches.
+        self.eo_leases.evict();
+        // TODO(#4869) - default shutdown behavior needs to be more clearly defined.
```
dbolduc, line 233: True, thanks for writing this up.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -177,17 +219,27 @@ where
...
+        // Currently, evict returns NACK_SHUTDOWN_ERROR for all exactly once
+        // leases. This includes the to_ack leases. Specifically,
+        // the leases that has been acknowledged by the application but not yet
```
dbolduc: nit:

```suggestion
        // the leases that have been acknowledged by the application but not yet
```

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -177,17 +219,27 @@ where
...
         }
+        if !to_nack.is_empty() {
+            self.leaser.nack(to_nack).await;
+        }
 
-        // TODO(#4847) - this nack needs to be broken into batches.
+        self.eo_leases.evict();
+        // TODO(#4869) - default shutdown behavior needs to be more clearly defined.
+        // Currently, evict returns NACK_SHUTDOWN_ERROR for all exactly once
+        // leases. This includes the to_ack leases. Specifically,
```
dbolduc, line 235: Aside, no action required: this behavior is dumb. We could see if they are pending and set `Shutdown` and for the non-pending ones do `ShutdownBeforeAck`.

But I don't think we should take any action now. I think doing anything is good enough for now. And then we get everything right as part of the shutdown work.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -439,33 +643,112 @@ pub(super) mod tests {
...
+        }
+        for i in 40..50 {
+            state.process(ExactlyOnceNack(test_id(i)));
+        }
         state.shutdown().await;
     }
 
     #[tokio::test(start_paused = true)]
-    async fn ack_out_of_lease_included() {
+    async fn ack_out_of_lease() {
```
dbolduc: optional: I think I would just remove this test. It is already covered in `at_least_once::Leases` and `exactly_once::Leases`.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -500,8 +802,17 @@ pub(super) mod tests {
             },
             state.leases
         );
+        assert_eq!(
```
dbolduc: Same with the nack variety. Up to you though.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -558,12 +877,19 @@ pub(super) mod tests {
         let start = Instant::now();
 
         const FLUSH_START: Duration = Duration::from_secs(1);
+        const CONFIRM_ACK_OFFSET: i32 = MAX_IDS_PER_RPC * 2;
```
dbolduc: Please move this into a separate test for exactly-once. It will be a lot easier to follow.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -597,11 +937,16 @@ pub(super) mod tests {
...
```
dbolduc: same

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -636,6 +995,7 @@ pub(super) mod tests {
...
```
dbolduc: same.

I know the `||` is over the at-least-once leases and exactly-once leases, but testing how they interact together just isn't that important. I'd rather have tests that are more readable for each.

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: Ack.

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: Ack.

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: Done.

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: Done.

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: 👍

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: Done.

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: Done.

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: Done.

PhongChuong:
Thanks for the review. PTAL.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -574,26 +807,63 @@ pub(super) mod tests {
...
+            state.add(test_id(i), at_least_once_info());
             state.process(Ack(test_id(i)));
         }
         assert_eq!(state.next_event().await, LeaseEvent::Flush);
         assert_eq!(start.elapsed(), FLUSH_START);
     }
 
     #[tokio::test(start_paused = true)]
-    async fn limit_size_of_nack_batch() {
+    async fn limit_size_of_ack_batch_exactly_once() {
```
dbolduc, line 828: much clearer, thanks.

dbolduc (approved):

PR #4992, opened 2026-03-11
test(pubsub): parameterize lease state test

PhongChuong:
This PR also add tests for exactly once lease for the tests message_expiration and limit_size_of_extends.

Towards #3964

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -248,10 +248,10 @@ pub(super) mod tests {
     use super::super::leaser::tests::MockLeaser;
     use super::Action::{Ack, ExactlyOnceAck, ExactlyOnceNack, Nack};
     use super::*;
-    use parameterized::parameterized;
     use std::collections::HashMap;
     use std::collections::HashSet;
     use std::sync::Arc;
+    use test_case::test_case;
     use tokio::sync::mpsc::unbounded_channel;
     use tokio::sync::oneshot::channel;
     use tokio::time::interval;
@@ -1004,11 +1004,9 @@ pub(super) mod tests {
         assert_eq!(start.elapsed(), FLUSH_START);
     }
 
-    #[parameterized(lease_info_factory = {
-        super::at_least_once_info,
-        super::exactly_once_info,
-    })]
-    #[parameterized_macro(tokio::test(start_paused = true))]
+    #[test_case(super::at_least_once_info)]
+    #[test_case(super::exactly_once_info)]
+    #[tokio::test(start_paused = true)]
     async fn limit_size_of_extends(lease_info_factory: fn() -> LeaseInfo) -> anyhow::Result<()> {
         const NUM_BATCHES: i32 = 5;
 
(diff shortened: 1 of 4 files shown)


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -1003,9 +1004,12 @@ pub(super) mod tests {
...
     }
 
-    #[tokio::test(start_paused = true)]
-    async fn limit_size_of_extends() -> anyhow::Result<()> {
-        // TODO(#3964) - Parameterize this test.
+    #[parameterized(lease_info_factory = {
+        super::at_least_once_info,
+        super::exactly_once_info,
+    })]
+    #[parameterized_macro(tokio::test(start_paused = true))]
```
dbolduc: TIL: `parameterized`

Does `test_case` work? That is what we typically use.

```suggestion
#[test_case(super::at_least_once_info)]
#[test_case(super::exactly_once_info)]
#[tokio::test(start_paused = true)]
```

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -1050,9 +1054,12 @@ pub(super) mod tests {
         Ok(())
     }
 
-    #[tokio::test(start_paused = true)]
-    async fn message_expiration() -> anyhow::Result<()> {
-        // TODO(#3964) - Parameterize this test.
+    #[parameterized(lease_info_factory = {
```
dbolduc: same

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -1023,7 +1027,7 @@ pub(super) mod tests {
 
         let mut want = HashSet::new();
         for i in 0..NUM_BATCHES * MAX_IDS_PER_RPC {
-            state.add(test_id(i), at_least_once_info());
+            state.add(test_id(i), lease_info_factory());
```
dbolduc, line 1028: ah, clever way to keep the diff in check

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: `test_case` does not work with `tokio::test`

dbolduc: this seems to work:

https://github.com/googleapis/google-cloud-rust/blob/925a0279553ea9f5d1dd467b4983f87c0a812fcb/src/storage/src/storage/bidi/worker.rs#L280-L282

PhongChuong: Weird. There was a reason why I switched `parameterized` instead.
Let me see if I can find it again.

dbolduc: The order matters with this stuff. Like this doesn't work, if I remember correctly:

```rs
[tokio::test] // bad, has to come after test_case, or serial
[test_case(...)]
```

PhongChuong: The suggestion worked and I can't recall why I switched to `parameterized`.
Anyway, thanks for the suggestion. I've updated to remove the parameterized crate.

Issue #4285, opened 2026-01-15
Add option to disable checksums on downloads


coryan:
We want an option to disable CRC32C computations on downloads. The option probably should be automatically enabled when downloading a range.

vsharonlynn:
@coryan I'm thinking of making a public method to `read_object` builder to set the RequestOption checksum crc32c to `None`.

The call will look like this:
```
client
    .read_object("projects/_/buckets/test-bucket", "test-object")
    .disable_crc32c_checksum(true)
    .send()
```
What do you think?

About the second sentence:

> The option probably should be automatically enabled when downloading a range.

Should this be enabled or disabled by default?
The option should likely be **disabled** by default when downloading a range.
According to [Cloud Storage documentation on Client-side validation](https://docs.cloud.google.com/storage/docs/data-validation#client-validation), the Cloud Storage server stores the checksum for the complete object, and the checksum can't be used to validate partial data.

Thanks!

coryan:
> [@coryan](https://github.com/coryan) I'm thinking of making a public method to `read_object` builder to set the RequestOption checksum crc32c to `None`.
> 
> The call will look like this:
> 
> ```
> client
>     .read_object("projects/_/buckets/test-bucket", "test-object")
>     .disable_crc32c_checksum(true)

Maybe `.enable_crc32c_checksum(false)`: it is easier to reason about positive "turn things on" options.

>     .send()
> ```
> 
> 
> > The option probably should be automatically enabled when downloading a range.
> 
> Should this be enabled or disabled by default? The option should likely be **disabled** by default when downloading a range. 

You are right, though I could argue that depends on what the option does (enable or disable things).  Let's just say "**checksums** should be disabled by default when downloading a range".

> and the checksum can't be used to validate partial data.

You probably know this, but that is not true for gRPC (including `open_object()` calls).  Also, you should ask your colleagues about go/gcs-e2e-checksumming and what parts of it are implemented or in progress.

vsharonlynn:
> Maybe .enable_crc32c_checksum(false): it is easier to reason about positive "turn things on" options.

Good idea.

> You are right, though I could argue that depends on what the option does (enable or disable things). Let's just say "checksums should be disabled by default when downloading a range".

Got it. Thanks for the clarification.

> You probably know this, but that is not true for gRPC (including open_object() calls). Also, you should ask your colleagues about go/gcs-e2e-checksumming and what parts of it are implemented or in progress.

Will check this out.

[...]

PR #4993, opened 2026-03-12
feat(storage): Add method to disable checksums on read object

vsharonlynn:
For issue #4285.

### DIFF
--- a/src/storage/src/storage/read_object.rs
+++ b/src/storage/src/storage/read_object.rs
@@ -168,8 +168,8 @@ where
     /// println!("object contents={:?}", contents);
     /// # Ok(()) }
     /// ```
-    pub fn compute_crc32c(mut self, v: bool) -> Self {
-        if v {
+    pub fn compute_crc32c(mut self, enable: bool) -> Self {
+        if enable {
             self.options
                 .checksum
                 .crc32c


src/storage/src/storage/read_object.rs
```diff
@@ -144,6 +144,38 @@ where
...
+    ///     .await?;
+    /// let mut contents = Vec::new();
+    /// while let Some(chunk) = reader.next().await.transpose()? {
+    ///     contents.extend_from_slice(&chunk);
+    /// }
+    /// println!("object contents={:?}", contents);
+    /// # Ok(()) }
+    /// ```
+    pub fn enable_crc32c_checksum(self, v: bool) -> Self {
+        let mut this = self;
```
coryan: You can just say:


```suggestion
    pub fn enable_crc32c_checksum(mut self, v: bool) -> Self {
```

src/storage/src/storage/read_object.rs
```diff
@@ -144,6 +144,38 @@ where
         this
     }
 
+    /// Enables computation of CRC32C checksums.
+    ///
+    /// CRC32C checksums are checked by default.
```
coryan:
```suggestion
    /// Enables computation of CRC32C checksums.
    ///
    /// Note that the library computes and verifies (if avaiable) CRC32C checksums at the end of
    /// the download. Use `enable_crc32c_checksum(false)` to disable the computation, but note
    /// that this reduces the data integrity guarantees. Data *can* be corrupted even when
    /// downloaded over HTTPS or other encrypted channels.
```

src/storage/src/storage/read_object.rs
```diff
@@ -935,6 +967,45 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn read_object_disabled_crc32c_check() -> Result {
+        // Calculate and serialize the incorrect crc32c checksum
+        let u = crc32c::crc32c("goodbye world".as_bytes());
+        let value = base64::prelude::BASE64_STANDARD.encode(u.to_be_bytes());
```
coryan: This is the test I would have written too. But note that what we want is to verify the CPU cycles are saved, not just that the checksums are ignored.

For a future improvement: I wonder if we could do that somehow? Maybe some counters / metrics?  Maybe a micro-benchmark that compares the CPU usage with and without checksums?

src/storage/src/storage/read_object.rs
```diff
@@ -144,6 +144,38 @@ where
...
+    pub fn enable_crc32c_checksum(self, v: bool) -> Self {
+        let mut this = self;
+        if !v {
+            this.options.checksum.crc32c = None;
+        } else if this.options.checksum.crc32c.is_none() {
+            this.options.checksum.crc32c =
+                Some(crate::storage::checksum::details::Crc32c::default());
+        }
+        this
+    }
```
joshuatants, line 181: How about: 

pub fn enable_crc32c_checksum(mut self, v: bool) -> Self {
    if v {
        self.options.checksum.crc32c.get_or_insert_with(Default::default);
    } else {
        self.options.checksum.crc32c = None;
    }
    self
}

Reasoning about a condition is easier than a "not" condition i.e. !v .

The other changes are using some Rust stuff that makes it more readable.

joshuatants (changes requested):

src/storage/src/storage/read_object.rs
vsharonlynn: Done. Thanks!

src/storage/src/storage/read_object.rs
```diff
@@ -144,6 +144,42 @@ where
...
+    ///     .send()
+    ///     .await?;
+    /// let mut contents = Vec::new();
+    /// while let Some(chunk) = reader.next().await.transpose()? {
+    ///     contents.extend_from_slice(&chunk);
+    /// }
+    /// println!("object contents={:?}", contents);
+    /// # Ok(()) }
+    /// ```
+    pub fn compute_crc32c_checksum(mut self, v: bool) -> Self {
```
coryan: There is an inconsistency in naming here vs. `compute_md5()`.  Consider naming this `comptue_crc32c()`.

src/storage/src/storage/read_object.rs
```diff
@@ -144,6 +144,42 @@ where
...
+        if v {
+            self.options
+                .checksum
+                .crc32c
+                .get_or_insert_with(Default::default);
```
coryan, line 176: This code path is untested.

vsharonlynn: Added a unit test to cover these lines. Thanks!

src/storage/src/storage/read_object.rs
vsharonlynn: Fixed, thanks! Sorry for not being careful with the naming.

src/storage/src/storage/read_object.rs
vsharonlynn: Thanks for the suggestion!

I've added a feature request to verify CPU cycles in the benchmark.

For the unit tests, I am checking the configs and behaviours of read object when crc32c is disabled:
1. Request option's crc32c checksum is None.
2. It will not validate the crc32c checksum returned from the server.

PR #4994, opened 2026-03-12
feat(spanner): query parameters

### LINKED ISSUE #4970: spanner: implement query parameters
Add support for query parameters in the Spanner client.

olavloite:
Adds support for adding query parameters for Spanner queries. Query parameters can be both typed and untyped:

1. Untyped: Only the value is sent to Spanner. Spanner will infer the data type from the SQL string.
2. Types: Both a value and data type is sent to Spanner. If the data type that is included with the query parameter does not correspond with the data type that is inferred from the SQL expression, then Spanner will reject the query.

Fixes #4970

### DIFF
--- a/src/spanner/src/statement.rs
+++ b/src/spanner/src/statement.rs
@@ -120,11 +120,12 @@ impl Statement {
         if self.params.is_empty() {
             None
         } else {
-            let mut map = serde_json::Map::new();
-            for (k, v) in &self.params {
-                map.insert(k.clone(), v.clone().into());
-            }
-            Some(map)
+            Some(
+                self.params
+                    .iter()
+                    .map(|(k, v)| (k.clone(), v.clone().into_serde_value()))
+                    .collect(),
+            )
         }
     }
 
--- a/src/spanner/src/value.rs
+++ b/src/spanner/src/value.rs
@@ -126,12 +126,12 @@ impl Value {
     }
 }
 
-/// Converts a `prost_types::Value` to a `serde_json::Value`.
-/// This is needed because the generated gapic client uses `serde_json::Value` instead of `prost_types::Value`.
-/// It is converted back from `serde_json::Value` to `prost_types::Value` before hitting the wire.
-impl From<Value> for serde_json::Value {
-    fn from(val: Value) -> Self {
-        match val.0.kind {
+impl Value {
+    /// Converts a `prost_types::Value` to a `serde_json::Value`.
+    /// This is needed because the generated gapic client uses `serde_json::Value` instead of `prost_types::Value`.
+    /// It is converted back from `serde_json::Value` to `prost_types::Value` before hitting the wire.
+    pub(crate) fn into_serde_value(self) -> serde_json::Value {
+        match self.0.kind {
             Some(prost_types::value::Kind::NullValue(_)) => serde_json::Value::Null,
             Some(prost_types::value::Kind::NumberValue(n)) => {
                 if let Some(num) = serde_json::Number::from_f64(n) {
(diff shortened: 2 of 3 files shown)


src/spanner/src/statement.rs
```diff
@@ -35,13 +106,51 @@
...
+    pub(crate) params: BTreeMap<String, Value>,
+    pub(crate) param_types: BTreeMap<String, Type>,
 }
 
 impl Statement {
-    /// Creates a new statement.
-    pub fn new(sql: impl Into<String>) -> Self {
-        Self { sql: sql.into() }
+    /// Creates a new statement builder.
+    pub fn builder(sql: impl Into<String>) -> StatementBuilder {
```
coryan, line 115: Have you considered using `&'static str` for the SQL statements?  That will prevent people from using `format!("select * from table where name = '{}', sql_injection)` as the sql statement and steer them towards safe parameters.

src/spanner/src/statement.rs
```diff
@@ -35,13 +106,51 @@
...
+    }
+
+    pub(crate) fn get_params(&self) -> Option<wkt::Struct> {
+        if self.params.is_empty() {
+            None
+        } else {
+            let mut map = serde_json::Map::new();
+            for (k, v) in &self.params {
+                map.insert(k.clone(), v.clone().into());
+            }
```
coryan: In think you could simply write:


```suggestion
            Some(self.params.iter().map(|(k, v)| (k.clone(), v.clone().into())).collect::<serde_json::Map>())
```

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=4910cc3e3b6b8f90fbfa40c75bca5ecb

You may even be able to skip the turbofish? Yup:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=d0c5f124f868b34bccda3048574b88db

src/spanner/src/value.rs
```diff
@@ -126,6 +126,41 @@ impl Value {
...
+                }
+            }
+            Some(prost_types::value::Kind::StringValue(s)) => serde_json::Value::String(s),
+            Some(prost_types::value::Kind::BoolValue(b)) => serde_json::Value::Bool(b),
+            Some(prost_types::value::Kind::StructValue(s)) => {
+                let mut map = serde_json::Map::with_capacity(s.fields.len());
+                for (k, v) in s.fields {
+                    map.insert(k, Value(v).into());
+                }
+                serde_json::Value::Object(map)
```
coryan: As above, I believe these transformations are more idiomatic as `s.fields.iter().map().collect()`

tests/spanner/tests/driver.rs
```diff
@@ -224,6 +235,41 @@ ORDER BY col_int64
         Ok(())
     }
 
+    #[tokio::test]
+    async fn test_query_with_parameters() -> Result<(), Box<dyn std::error::Error>> {
```
coryan: Generally we use `anyhow::Result<()>` in our tests. Saves a lot of typing.

tests/spanner/tests/driver.rs
```diff
@@ -126,6 +127,16 @@ mod spanner {
             .await
             .expect("Failed to build database client");
 
+        Some(db_client)
+    }
+
+    #[tokio::test]
+    async fn test_simple_query() -> Result<(), Box<dyn std::error::Error>> {
```
coryan: Same comment about `anyhow::Result<()>` for tests.  Also, Rust tests do not need to start with `test_`, in fact, it is idiomatic to just call them `simple_query()`.  They are a test by virtue of the `#[tokio::test]` or `#[test]` attribute.

src/spanner/src/value.rs
```diff
@@ -126,6 +126,41 @@ impl Value {
     }
 }
 
+/// Converts a `prost_types::Value` to a `serde_json::Value`.
+/// This is needed because the generated gapic client uses `serde_json::Value` instead of `prost_types::Value`.
+/// It is converted back from `serde_json::Value` to `prost_types::Value` before hitting the wire.
+impl From<Value> for serde_json::Value {
```
coryan: This makes the trait implementation public, do you want customers to be able to do this?  If not, consider a function called `pub(crate) fn from_serde_value()`.  At the very least, consider:


```suggestion
#[doc(hidden)]
impl From<Value> for serde_json::Value {
```

To avoid advertising the trait in the public documentation.

src/spanner/src/statement.rs
olavloite: That is an interesting suggestion, that I honestly had not considered (also because I did not know the option existed). However, I believe that this would be too restrictive, because:
1. It would also prevent users from easily build SQL strings that use dynamic identifiers, and identifiers cannot be bound to query parameters. E.g. a query like `select * from some_tenant_schema.fixed_table_name`, where the schema name is dynamically determined from for example the logged in user, would then be a lot harder to create.
2. Also other types of query generation would then get a lot harder, e.g. queries that are dynamically built from a user interface, and where `WHERE` clauses are conditionally added to the query based on what the user has chosen in the UI.

I've also checked with other database drivers (e.g. PostgreSQL). They also do not limit the query string to `&'static str`, even though they also strongly encourage users to use query parameters.

src/spanner/src/statement.rs
olavloite: Done

src/spanner/src/value.rs
olavloite: Thanks, this was indeed not intended for public use. I created a `pub(crate)` method for it.

src/spanner/src/value.rs
olavloite: This was actually deliberate, in order to be able to call `with_capacity`. But it turns out that the `collect()` method actually also does that (or something very similar). So fixed.

tests/spanner/tests/driver.rs
olavloite: Done

tests/spanner/tests/driver.rs
olavloite: Done

PR #4995, opened 2026-03-12
chore: update googleapis and discovery-artifact-manager and regenerate

jskeet:


jskeet:
(This was created when testing `librarianops --docker` with Rust.)

PR #4998, opened 2026-03-12
feat(spanner): add get(..) and try_get(..) methods for ResultSet

### LINKED ISSUE #4971: spanner: implement methods for getting column values by name and index
Get a column value from a ResultSet by name and index.

olavloite:
- Adds `get(..)` and `try_get(..)` methods for getting the column values of a `ResultSet`.
- Also adds a `ResultSetMetadata` struct that holds the column names and types of a `ResultSet`.

Fixes #4971

### DIFF
--- a/src/spanner/src/result_set_metadata.rs
+++ b/src/spanner/src/result_set_metadata.rs
@@ -47,14 +47,14 @@ impl ResultSetMetadata {
         let mut column_names = Vec::new();
         let mut column_types = Vec::new();
 
-        if let Some(m) = &metadata {
-            if let Some(row_type) = &m.row_type {
-                for field in row_type.fields.iter() {
-                    column_names.push(field.name.clone());
-                    let column_type = field.r#type.clone().map(Into::into).unwrap_or_default();
-                    column_types.push(column_type);
-                }
-            }
+        let fields = metadata
+            .and_then(|m| m.row_type)
+            .into_iter()
+            .flat_map(|r| r.fields.into_iter());
+        for field in fields {
+            column_names.push(field.name);
+            let column_type = field.r#type.map(Into::into).unwrap_or_default();
+            column_types.push(column_type);
         }
 
         Self {
--- a/src/spanner/src/row.rs
+++ b/src/spanner/src/row.rs
@@ -55,10 +55,7 @@ impl ColumnIndex for &str {
 
 impl ColumnIndex for String {
     fn index(&self, row: &Row) -> Option<usize> {
-        row.metadata
-            .column_names
-            .iter()
-            .position(|name| name == self)
+        self.as_str().index(row)
     }
 }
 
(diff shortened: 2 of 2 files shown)


src/spanner/src/result_set_metadata.rs
```diff
@@ -0,0 +1,134 @@
...
+
+        if let Some(m) = &metadata {
+            if let Some(row_type) = &m.row_type {
+                for field in row_type.fields.iter() {
+                    column_names.push(field.name.clone());
+                    let column_type = field.r#type.clone().map(Into::into).unwrap_or_default();
+                    column_types.push(column_type);
+                }
+            }
+        }
```
coryan: Check my math, but I think you can just write....

```suggestion
        let fields = metadata.and_then(|m| m.row_type).flat_map(|r| r.fields.into_iter());
        for field in  fields {
                    column_names.push(field.name);
                    let column_type = field.r#type.map(Into::into).unwrap_or_default();
                    column_types.push(column_type);
        }
```

Which looks simpler to me.  Note how this  avoids the clone because you are effectively moving the data out of `metadata`. 

You could try to do `.map(|f| (f.name, f.r#type.map(crate::types::Type::from).unwrap_or_default())).unzip()` but even I think that is getting too clever.

src/spanner/src/row.rs
```diff
@@ -12,12 +12,66 @@
...
+            .position(|name| name == *self)
+    }
+}
+
+impl ColumnIndex for String {
+    fn index(&self, row: &Row) -> Option<usize> {
+        row.metadata
+            .column_names
+            .iter()
+            .position(|name| name == self)
```
coryan: Maybe:


```suggestion
        self.as_str().index(row)
```

???

src/spanner/src/row.rs
```diff
@@ -26,15 +80,212 @@ impl Row {
...
+        let idx = index
+            .index(self)
+            .ok_or_else(|| crate::Error::deser(RowError::ColumnNotFound(format!("{:?}", index))))?;
+        let value = self.values.get(idx).ok_or_else(|| {
+            crate::Error::deser(RowError::IndexOutOfRange {
+                index: idx,
+                len: self.values.len(),
+            })
+        })?;
+        let type_ = self.metadata.column_types.get(idx).ok_or_else(|| {
```
coryan: In Rust, it is customary to use `r#type` to escape reserved identifiers.

src/spanner/src/result_set_metadata.rs
olavloite: Done

src/spanner/src/row.rs
olavloite: Done

src/spanner/src/row.rs
olavloite: Done

PR #5003, opened 2026-03-12
feat(spanner): multi-use read-only transactions

olavloite:
Adds support for multi-use read-only transactions. The current implementation always uses a BeginTransaction RPC to start the transaction. In a follow-up pull request, we will also add support for including the BeginTransaction option with the first query in the transaction.

### DIFF
--- a/src/spanner/src/database_client.rs
+++ b/src/spanner/src/database_client.rs
@@ -67,6 +67,10 @@ impl DatabaseClient {
     /// # Ok(())
     /// # }
     /// ```
+    ///
+    /// A single-use read-only transaction is optimized for the case where only a single
+    /// read or query is needed. This is more efficient than using a read-only transaction
+    /// for a single read or query.
     pub fn single_use(&self) -> SingleUseReadOnlyTransactionBuilder {
         SingleUseReadOnlyTransactionBuilder::new(self.clone())
     }
@@ -86,6 +90,10 @@ impl DatabaseClient {
     /// # Ok(())
     /// # }
     /// ```
+    ///
+    /// A read-only transaction can be used to execute multiple reads or queries.
+    /// These transactions guarantee data consistency across multiple read operations,
+    /// but don't permit data modifications. Read-only transactions do not take locks.
     pub fn read_only_transaction(&self) -> MultiUseReadOnlyTransactionBuilder {
         MultiUseReadOnlyTransactionBuilder::new(self.clone())
     }
--- a/src/spanner/src/read_only_transaction.rs
+++ b/src/spanner/src/read_only_transaction.rs
@@ -15,7 +15,6 @@
 use crate::database_client::DatabaseClient;
 use crate::model::TransactionOptions;
 use crate::model::transaction_options::ReadOnly;
-use crate::model::transaction_options::read_only::TimestampBound as ReadOnlyTimestampBound;
 use crate::result_set::ResultSet;
 use crate::statement::Statement;
 use crate::timestamp_bound::TimestampBound;
@@ -81,26 +80,12 @@ impl SingleUseReadOnlyTransactionBuilder {
     /// # }
     /// ```
     pub fn build(self) -> SingleUseReadOnlyTransaction {
-        let mut read_only = ReadOnly::default();
-        if let Some(bound) = self.timestamp_bound {
-            read_only.timestamp_bound = Some(bound.0);
-        } else {
-            read_only.timestamp_bound = Some(ReadOnlyTimestampBound::Strong(true));
-        }
-
-        let transaction_options = TransactionOptions {
-            mode: Some(crate::model::transaction_options::Mode::ReadOnly(Box::new(
-                read_only,
-            ))),
-            ..Default::default()
+        let read_only = match self.timestamp_bound {
+            Some(b) => ReadOnly::default().set_timestamp_bound(b.0),
+            None => ReadOnly::default().set_strong(true),
         };
-
-        let transaction_selector = crate::model::TransactionSelector {
-            selector: Some(crate::model::transaction_selector::Selector::SingleUse(
-                Box::new(transaction_options),
-            )),
-            ..Default::default()
-        };
+        let transaction_selector = crate::model::TransactionSelector::default()
+            .set_single_use(TransactionOptions::default().set_read_only(read_only));
 
         SingleUseReadOnlyTransaction {
             context: ReadContext::new(self.client, transaction_selector),
@@ -125,6 +110,7 @@ impl SingleUseReadOnlyTransactionBuilder {
 /// # Ok(())
 /// # }
 /// ```
+#[derive(Debug)]
 pub struct SingleUseReadOnlyTransaction {
     context: ReadContext,
 }
@@ -150,7 +136,10 @@ impl SingleUseReadOnlyTransaction {
     /// # Ok(())
     /// # }
     /// ```
-    pub async fn execute_query(&self, statement: impl Into<Statement>) -> crate::Result<ResultSet> {
+    pub async fn execute_query<T: Into<Statement>>(
+        &self,
+        statement: T,
+    ) -> crate::Result<ResultSet> {
         self.context.execute_query(statement).await
     }
 }
@@ -213,26 +202,14 @@ impl MultiUseReadOnlyTransactionBuilder {
     /// # }
     /// ```
     pub async fn build(self) -> crate::Result<MultiUseReadOnlyTransaction> {
-        let mut read_only = ReadOnly::default();
-        if let Some(bound) = self.timestamp_bound {
-            read_only.timestamp_bound = Some(bound.0);
-        } else {
-            read_only.timestamp_bound = Some(ReadOnlyTimestampBound::Strong(true));
-        }
-        read_only.return_read_timestamp = true;
-
-        let transaction_options = TransactionOptions {
-            mode: Some(crate::model::transaction_options::Mode::ReadOnly(Box::new(
-                read_only,
-            ))),
-            ..Default::default()
+        let read_only = ReadOnly::default().set_return_read_timestamp(true);
+        let read_only = match self.timestamp_bound {
+            Some(b) => read_only.set_timestamp_bound(b.0),
+            None => read_only.set_strong(true),
         };
-
-        let request = crate::model::BeginTransactionRequest {
-            session: self.client.session.name.clone(),
-            options: Some(transaction_options),
-            ..Default::default()
-        };
+        let request = crate::model::BeginTransactionRequest::default()
+            .set_session(self.client.session.name.clone())
+            .set_options(TransactionOptions::default().set_read_only(read_only));
 
         // TODO(#4972): make request options configurable
         let response = self
@@ -241,18 +218,10 @@ impl MultiUseReadOnlyTransactionBuilder {
             .begin_transaction(request, crate::RequestOptions::default())
             .await?;
 
-        let transaction_selector = crate::model::TransactionSelector {
-            selector: Some(crate::model::transaction_selector::Selector::Id(
-                response.id,
-            )),
-            ..Default::default()
-        };
-
+        let transaction_selector = crate::model::TransactionSelector::default().set_id(response.id);
         Ok(MultiUseReadOnlyTransaction {
             context: ReadContext::new(self.client, transaction_selector),
-            read_timestamp: response
-                .read_timestamp
-                .and_then(|ts| std::convert::TryInto::<time::OffsetDateTime>::try_into(ts).ok()),
+            read_timestamp: response.read_timestamp,
         })
     }
 }
@@ -278,14 +247,15 @@ impl MultiUseReadOnlyTransactionBuilder {
 /// # Ok(())
 /// # }
 /// ```
+#[derive(Debug)]
 pub struct MultiUseReadOnlyTransaction {
     context: ReadContext,
-    pub(crate) read_timestamp: Option<time::OffsetDateTime>,
+    pub(crate) read_timestamp: Option<wkt::Timestamp>,
 }
 
 impl MultiUseReadOnlyTransaction {
     /// Returns the read timestamp chosen for the transaction.
-    pub fn read_timestamp(&self) -> Option<time::OffsetDateTime> {
+    pub fn read_timestamp(&self) -> Option<wkt::Timestamp> {
         self.read_timestamp
     }
 
@@ -309,12 +279,15 @@ impl MultiUseReadOnlyTransaction {
     /// # Ok(())
     /// # }
     /// ```
-    pub async fn execute_query(&self, statement: impl Into<Statement>) -> crate::Result<ResultSet> {
+    pub async fn execute_query<T: Into<Statement>>(
+        &self,
+        statement: T,
+    ) -> crate::Result<ResultSet> {
         self.context.execute_query(statement).await
     }
 }
 
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub(crate) struct ReadContext {
     client: DatabaseClient,
     transaction_selector: crate::model::TransactionSelector,
@@ -331,20 +304,18 @@ impl ReadContext {
         }
     }
 
-    pub(crate) async fn execute_query(
+    pub(crate) async fn execute_query<T: Into<Statement>>(
         &self,
-        statement: impl Into<Statement>,
+        statement: T,
     ) -> crate::Result<ResultSet> {
         let statement = statement.into();
 
-        let request = crate::model::ExecuteSqlRequest {
-            session: self.client.session.name.clone(),
-            transaction: Some(self.transaction_selector.clone()),
-            params: statement.get_params(),
-            param_types: statement.get_param_types(),
-            sql: statement.sql,
-            ..Default::default()
-        };
+        let mut request = crate::model::ExecuteSqlRequest::default()
+            .set_session(self.client.session.name.clone())
+            .set_transaction(self.transaction_selector.clone());
+        request.params = statement.get_params();
+        request.param_types = statement.get_param_types();
+        request = request.set_sql(statement.sql);
 
         let stream = self
             .client
@@ -365,10 +336,10 @@ mod tests {
     #[test]
     fn auto_traits() {
         static_assertions::assert_impl_all!(SingleUseReadOnlyTransactionBuilder: Send, Sync);
-        static_assertions::assert_impl_all!(SingleUseReadOnlyTransaction: Send, Sync);
+        static_assertions::assert_impl_all!(SingleUseReadOnlyTransaction: Send, Sync, std::fmt::Debug);
         static_assertions::assert_impl_all!(MultiUseReadOnlyTransactionBuilder: Send, Sync);
-        static_assertions::assert_impl_all!(MultiUseReadOnlyTransaction: Send, Sync);
-        static_assertions::assert_impl_all!(ReadContext: Send, Sync);
+        static_assertions::assert_impl_all!(MultiUseReadOnlyTransaction: Send, Sync, std::fmt::Debug);
+        static_assertions::assert_impl_all!(ReadContext: Send, Sync, std::fmt::Debug);
     }
 
     fn create_session_mock() -> spanner_grpc_mock::MockSpanner {
@@ -388,23 +359,15 @@ mod tests {
         spanner_grpc_mock::google::spanner::v1::PartialResultSet {
             metadata: Some(spanner_grpc_mock::google::spanner::v1::ResultSetMetadata {
                 row_type: Some(spanner_grpc_mock::google::spanner::v1::StructType {
-                    fields: vec![spanner_grpc_mock::google::spanner::v1::struct_type::Field {
-                        name: "".to_string(),
-                        r#type: None,
-                    }],
+                    fields: vec![Default::default()],
                 }),
-                transaction: None,
-                undeclared_parameters: None,
+                ..Default::default()
             }),
             values: vec![prost_types::Value {
                 kind: Some(prost_types::value::Kind::StringValue("1".to_string())),
             }],
-            chunked_value: false,
-            resume_token: vec![],
-            stats: None,
-            precommit_token: None,
-            cache_update: None,
             last: true,
+            ..Default::default()
         }
     }
 
@@ -432,18 +395,6 @@ mod tests {
         (db_client, server)
     }
 
-    async fn assert_select1_result_set(mut rs: crate::result_set::ResultSet) {
-        let row = rs.next().await.expect("has row").expect("has valid row");
-        assert_eq!(row.raw_values().len(), 1);
-        if let Some(prost_types::value::Kind::StringValue(ref s)) = row.raw_values()[0].0.kind {
-            assert_eq!(s, "1");
-        } else {
-            panic!("Expected StringValue");
-        }
-
-        assert!(rs.next().await.is_none());
-    }
-
     #[tokio::test]
     async fn single_use_builder() {
         let mock = create_session_mock();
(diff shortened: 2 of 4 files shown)


src/spanner/src/database_client.rs
```diff
@@ -68,6 +70,25 @@ impl DatabaseClient {
...
+
+    /// Returns a builder for a multi-use read-only transaction.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_spanner::client::{Spanner, Statement};
+    /// # async fn run(spanner: Spanner) -> Result<(), google_cloud_spanner::Error> {
+    /// let db_client = spanner.database_client("projects/p/instances/i/databases/d").build().await?;
+    /// let tx = db_client.read_only_transaction().build().await?;
+    /// let stmt = Statement::builder("SELECT * FROM users WHERE id = @id")
```
dbolduc, line 86: > stmt

nit: what do you have against vowels? I think we tend to spell these out.

src/spanner/src/database_client.rs
```diff
@@ -68,6 +70,25 @@ impl DatabaseClient {
...
+    ///     .add_param("id", &42)
+    ///     .build();
+    /// let mut rs = tx.execute_query(stmt).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn read_only_transaction(&self) -> MultiUseReadOnlyTransactionBuilder {
```
dbolduc, line 97: comment: this name is not self-consistent with `single_use`.

Also, as someone new to spanner, which one do I want to use? Can you explain, or link to something that explains the difference?

src/spanner/src/read_only_transaction.rs
```diff
@@ -144,16 +151,195 @@ impl SingleUseReadOnlyTransaction {
...
+    /// ```
+    /// # use google_cloud_spanner::client::Spanner;
+    /// # use google_cloud_spanner::client::TimestampBound;
+    /// # async fn set_bound(spanner: Spanner) -> Result<(), google_cloud_spanner::Error> {
+    /// let db_client = spanner.database_client("projects/p/instances/i/databases/d").build().await?;
+    /// let builder = db_client.read_only_transaction().with_timestamp_bound(TimestampBound::strong());
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub fn with_timestamp_bound(mut self, bound: TimestampBound) -> Self {
```
dbolduc, line 187: Q: why do we wrap the generated type for `TimestampBound` in a handwritten type?

src/spanner/src/read_only_transaction.rs
```diff
@@ -144,16 +151,195 @@ impl SingleUseReadOnlyTransaction {
...
+                read_only,
+            ))),
+            ..Default::default()
+        };
+
+        let request = crate::model::BeginTransactionRequest {
+            session: self.client.session.name.clone(),
+            options: Some(transaction_options),
+            ..Default::default()
+        };
```
dbolduc: the generated model types all have fluent setters to make this sort of thing more pleasant to write. e.g. You can set oneof fields directly, or supply any type that can be converted to the expected type.


```suggestion
        let request = crate::model::BeginTransactionRequest::default()
            .set_session(self.client.session.name.clone())
            .set_options(TransactionOptions::default().set_read_only(read_only));
```

src/spanner/src/read_only_transaction.rs
```diff
@@ -144,16 +151,195 @@ impl SingleUseReadOnlyTransaction {
...
+    /// # }
+    /// ```
+    pub async fn build(self) -> crate::Result<MultiUseReadOnlyTransaction> {
+        let mut read_only = ReadOnly::default();
+        if let Some(bound) = self.timestamp_bound {
+            read_only.timestamp_bound = Some(bound.0);
+        } else {
+            read_only.timestamp_bound = Some(ReadOnlyTimestampBound::Strong(true));
+        }
+        read_only.return_read_timestamp = true;
```
dbolduc:
```suggestion
        let read_only = ReadOnly::default().set_return_read_timestamp(true);
        let read_only = match self.timestamp_bound {
            Some(b) => read_only.set_timestamp_bound(b.0),
            None => read_only.set_strong(true),
        };
```

or if you'd rather one line:

```rs
        let read_only = ReadOnly::default()
            .set_timestamp_bound(
                self.timestamp_bound
                    .map_or_else(|| ReadOnlyTimestampBound::Strong(true), |b| b.0),
            )
            .set_return_read_timestamp(true);
```

src/spanner/src/read_only_transaction.rs
```diff
@@ -144,16 +151,195 @@ impl SingleUseReadOnlyTransaction {
...
+            .spanner
+            .begin_transaction(request, crate::RequestOptions::default())
+            .await?;
+
+        let transaction_selector = crate::model::TransactionSelector {
+            selector: Some(crate::model::transaction_selector::Selector::Id(
+                response.id,
+            )),
+            ..Default::default()
+        };
```
dbolduc:
```suggestion
        let transaction_selector = crate::model::TransactionSelector::default().set_id(response.id);
```

and consider `use crate::model::TransactionSelector;` and all the other model types you use.

src/spanner/src/read_only_transaction.rs
```diff
@@ -144,16 +151,195 @@ impl SingleUseReadOnlyTransaction {
...
+
+        Ok(MultiUseReadOnlyTransaction {
+            context: ReadContext::new(self.client, transaction_selector),
+            read_timestamp: response
+                .read_timestamp
+                .and_then(|ts| std::convert::TryInto::<time::OffsetDateTime>::try_into(ts).ok()),
```
dbolduc:
```suggestion
            read_timestamp: response.read_timestamp.and_then(|ts| ts.try_into().ok()),
```

Aside: it is ok to `.unwrap()`/`.expect()` when you are 100% sure something doesn't error. e.g. if you know the timestamp will be in range, and the conversion will work.

Using `.ok()` is totally cool with me though.

src/spanner/src/read_only_transaction.rs
```diff
@@ -144,16 +151,195 @@ impl SingleUseReadOnlyTransaction {
...
+///     .build();
+/// let mut rs1 = tx.execute_query(stmt1).await?;
+///
+/// let stmt2 = Statement::builder("SELECT * FROM other_table")
+///     .build();
+/// let mut rs2 = tx.execute_query(stmt2).await?;
+/// # Ok(())
+/// # }
+/// ```
+pub struct MultiUseReadOnlyTransaction {
```
dbolduc, line 251: public types should `#[derive(Debug)` at least.

src/spanner/src/read_only_transaction.rs
```diff
@@ -144,16 +151,195 @@ impl SingleUseReadOnlyTransaction {
...
+    context: ReadContext,
+    pub(crate) read_timestamp: Option<time::OffsetDateTime>,
+}
+
+impl MultiUseReadOnlyTransaction {
+    /// Returns the read timestamp chosen for the transaction.
+    pub fn read_timestamp(&self) -> Option<time::OffsetDateTime> {
```
dbolduc: We can't use the `time` crate in our public API. It is version `0.x` which is not stable.

Why don't we do everything in terms of `wkt::Timestamp`, and let applications convert it?

src/spanner/src/read_only_transaction.rs
```diff
@@ -144,16 +151,195 @@ impl SingleUseReadOnlyTransaction {
...
+    ///     .build();
+    /// let mut rs = tx.execute_query(stmt).await?;
+    /// while let Some(row) = rs.next().await {
+    ///     let _row = row?;
+    ///     // process row
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn execute_query(&self, statement: impl Into<Statement>) -> crate::Result<ResultSet> {
```
dbolduc: super minor consistency nit:

```suggestion
    pub async fn execute_query<T: Into<Statement>>(&self, statement: T) -> crate::Result<ResultSet> {
```

(These two things are exactly the same... but they show up differently in the documentation for some reason.)

src/spanner/src/read_only_transaction.rs
```diff
@@ -88,14 +88,22 @@ impl SingleUseReadOnlyTransactionBuilder {
             read_only.timestamp_bound = Some(ReadOnlyTimestampBound::Strong(true));
         }
 
+        let transaction_options = TransactionOptions {
```
dbolduc: you should clean up these too btw.

src/spanner/src/read_only_transaction.rs
```diff
@@ -198,8 +381,39 @@ mod tests {
                 },
             ))
         });
+        mock
+    }
+
+    fn setup_select1() -> spanner_grpc_mock::google::spanner::v1::PartialResultSet {
+        spanner_grpc_mock::google::spanner::v1::PartialResultSet {
```
dbolduc, line 359: comment: sadly there are no fluent setters available for the tonic-generated types :(

src/spanner/src/read_only_transaction.rs
```diff
@@ -198,8 +381,39 @@ mod tests {
...
+            metadata: Some(spanner_grpc_mock::google::spanner::v1::ResultSetMetadata {
+                row_type: Some(spanner_grpc_mock::google::spanner::v1::StructType {
+                    fields: vec![spanner_grpc_mock::google::spanner::v1::struct_type::Field {
+                        name: "".to_string(),
+                        r#type: None,
```
dbolduc: nit: there should be a `Field::default()`. No need to set fields to their default values. That is just clutter IMO.

Here and below.

src/spanner/src/read_only_transaction.rs
```diff
@@ -215,10 +429,36 @@ mod tests {
             .await
             .expect("Failed to create DatabaseClient");
 
+        (db_client, server)
+    }
+
+    async fn assert_select1_result_set(mut rs: crate::result_set::ResultSet) {
```
dbolduc: If you factor this out and a test fails, we will see the line number in this function. That is unfortunate. We would rather see the calling line fail.

src/spanner/src/read_only_transaction.rs
```diff
@@ -215,10 +429,36 @@ mod tests {
...
+        let row = rs.next().await.expect("has row").expect("has valid row");
+        assert_eq!(row.raw_values().len(), 1);
+        if let Some(prost_types::value::Kind::StringValue(ref s)) = row.raw_values()[0].0.kind {
+            assert_eq!(s, "1");
+        } else {
+            panic!("Expected StringValue");
+        }
+
+        assert!(rs.next().await.is_none());
```
dbolduc:
```suggestion
        let result = rs.next().await;
        assert!(result.is_none(), "expected None, got {result:?}");
```

src/spanner/src/read_only_transaction.rs
```diff
@@ -215,10 +429,36 @@ mod tests {
             .await
             .expect("Failed to create DatabaseClient");
 
+        (db_client, server)
+    }
+
+    async fn assert_select1_result_set(mut rs: crate::result_set::ResultSet) {
+        let row = rs.next().await.expect("has row").expect("has valid row");
+        assert_eq!(row.raw_values().len(), 1);
+        if let Some(prost_types::value::Kind::StringValue(ref s)) = row.raw_values()[0].0.kind {
```
dbolduc: Q: Hm, does this have to be a `prost_types::value` vs. a `wkt::Value`?

It feels like we should generally be working in terms of our own types and convert them to/from prost/tonic types only when we have to send them. But maybe it is saving conversions? Or maybe the conversions are impossible for some reason.

This is non-blocking, but something we should think about it.

src/spanner/src/read_only_transaction.rs
```diff
@@ -215,10 +429,36 @@ mod tests {
...
+            assert_eq!(s, "1");
+        } else {
+            panic!("Expected StringValue");
+        }
```
dbolduc:
```suggestion
        use crate::value::Value;
        use super::super::result_set::tests::string_val;
        assert_eq!(row.raw_values(), [Value(string_val("1"))]);
```

you could make the `result_set.rs`'s `pub(crate) mod tests`, and also make it `pub(crate) fn string_val` and reuse it here.

Aside, for another PR: The `result_set.rs` tests could benefit from comparing the containers for equality, instead of comparing length then each element individually.

src/spanner/src/read_only_transaction.rs
```diff
@@ -215,10 +429,36 @@ mod tests {
...
-            _ => panic!("Expected ReadOnly mode"),
+        let ro = match tx.context.transaction_selector.selector {
+            Some(crate::model::transaction_selector::Selector::SingleUse(opts)) => {
+                match opts.mode {
+                    Some(crate::model::transaction_options::Mode::ReadOnly(ro)) => ro,
+                    _ => panic!("Expected ReadOnly mode"),
+                }
+            }
+            _ => panic!("Expected SingleUse selector"),
         };
```
dbolduc:
```suggestion
        let ro = tx
            .context
            .transaction_selector
            .single_use()
            .expect("Expected SingleUse selector")
            .read_only()
            .expect("Expected ReadOnly mode");
```

oneofs also have convenient getters

src/spanner/src/read_only_transaction.rs
```diff
@@ -231,9 +471,14 @@ mod tests {
                 std::time::Duration::from_secs(10),
             ))
             .build();
-        let ro2 = match tx2.transaction_options.mode {
-            Some(crate::model::transaction_options::Mode::ReadOnly(ro)) => ro,
-            _ => panic!("Expected ReadOnly mode"),
+        let ro2 = match tx2.context.transaction_selector.selector {
```
dbolduc: and here

tests/spanner/tests/driver.rs
```diff
@@ -393,4 +393,40 @@ ORDER BY col_int64
...
+
+        // Execute the first query.
+        let mut rs1 = tx
+            .execute_query(Statement::builder("SELECT 1 AS col_int").build())
+            .await?;
+        let row1 = rs1.next().await.transpose()?.unwrap();
+        let val1 = row1.raw_values()[0].as_string();
+        assert_eq!(val1, "1");
+        let next1 = rs1.next().await.transpose()?;
+        assert!(next1.is_none());
```
dbolduc:
```suggestion
        assert!(next1.is_none(), "{next1:?}");
```

tests/spanner/tests/driver.rs
```diff
@@ -393,4 +393,40 @@ ORDER BY col_int64
...
+
+        // Execute the second query reusing the same transaction.
+        let mut rs2 = tx
+            .execute_query(Statement::builder("SELECT 2 AS col_int").build())
+            .await?;
+        let row2 = rs2.next().await.transpose()?.unwrap();
+        let val2 = row2.raw_values()[0].as_string();
+        assert_eq!(val2, "2");
+        let next2 = rs2.next().await.transpose()?;
+        assert!(next2.is_none());
```
dbolduc: same

src/spanner/src/read_only_transaction.rs
dbolduc: For synchronous fns we use `#[track_caller]` to achieve this. But it doesn't work for `async fn`.

We have written macros for this sort of thing... but I don't love that.

My default would probably be to spell this thing out in each of the individual tests. But you have a better sense for how convenient this helper is.

src/spanner/src/database_client.rs
olavloite: To be honest: I really don't care, as long as the answer will be consistent.

The current choice is based on what is normal for the PostgreSQL driver (tokio-postgres). If this repository prefers spelled out variable names, then I'll make that change in a follow-up PR for all the samples.

src/spanner/src/database_client.rs
olavloite: I've added some more commentary on when to use which.

The naming of the methods are consistent with the other Spanner clients (to a certain degree as there is some variance between the different clients. E.g. the Java client defines both `singleUse()` and `singleUseReadOnlyTransaction()`, where only the latter will return the read timestamp, but other than that they are completely equal)

src/spanner/src/read_only_transaction.rs
olavloite: I think that I might have misunderstood something regarding generated types / protobuf types in the current Rust ecosystem.

1. My preference would be to use 'generated types' whenever possible.
2. I have referred to this as 'I would prefer to use the raw protobuf types as much as possible', which got pushback.
3. I now realize that 'raw protobuf types' was probably interpreted as 'prost / tonic'. I initially did not realize that there are two sets of generated types: model and tonic/prost.

So that indeed means that we should replace this with just using the model type (I'll do this in a follow-up PR).

src/spanner/src/read_only_transaction.rs
olavloite: Done

src/spanner/src/read_only_transaction.rs
olavloite: There's some more background info in this conversation: https://github.com/googleapis/google-cloud-rust/pull/4850/changes#r2899107755

The reason that this is `prost_types::value` is that it comes from a server-streaming RPC, which is currently not covered by the gapic generator. That means that the values that are returned by those calls to the hand-written client are of type `prost_types::...`.

But in some ways it is also a happy coincidence, because this something that we have seen in other clients that eat a large part of the total CPU consumption of the client; converting from the underlying protobuf classes/structs to some higher-level intermediate representation that is used in the client, which again is converted to an idiomatic language type when the application gets the value. The fact that we use `prost_types::value` directly in the hand-written client means that we skip the intermediate conversion.

src/spanner/src/read_only_transaction.rs
olavloite: Done

src/spanner/src/read_only_transaction.rs
olavloite: Done

src/spanner/src/read_only_transaction.rs
olavloite: Not relevant anymore, as the return type has been changed to `wkt::Timestamp`.

src/spanner/src/read_only_transaction.rs
olavloite: Done (see also my other comment regarding generated types)

src/spanner/src/read_only_transaction.rs
olavloite: Done

src/spanner/src/read_only_transaction.rs
olavloite: Done

src/spanner/src/read_only_transaction.rs
olavloite: Will do in a separate PR for all usages

src/spanner/src/read_only_transaction.rs
olavloite: ack

src/spanner/src/read_only_transaction.rs
olavloite: Done

src/spanner/src/read_only_transaction.rs
olavloite: I've moved the code to the two tests that used it (for now). The helper is purely for reducing code duplication. For two tests, duplicating seems OK.

src/spanner/src/read_only_transaction.rs
olavloite: Done

src/spanner/src/read_only_transaction.rs
olavloite: Done

src/spanner/src/read_only_transaction.rs
olavloite: Done

src/spanner/src/read_only_transaction.rs
olavloite: Done

tests/spanner/tests/driver.rs
olavloite: Done

tests/spanner/tests/driver.rs
olavloite: done

src/spanner/src/database_client.rs
dbolduc: > To be honest: I really don't care

+1. This is unimportant. It is just a style nit, which I have been trained to notice.

> If this repository

I think Rust generally follows the C++ style guide, which says "Do not abbreviate by deleting letters within a word." go/cstyle#General_Naming_Rules

So I think either `s` or `statement` is more adherent to the scripture. But no need to change anything in this PR.

src/spanner/src/database_client.rs
dbolduc: > I've added some more commentary on when to use which.

Thanks!

> are consistent with the other Spanner clients

Feel free to improve on the names in the Rust client. :)

src/spanner/src/read_only_transaction.rs
dbolduc: > two sets of generated types: model and tonic/prost.

Ahh, yeah I think only Rust has that. Glad you figured it out. The plan SGTM.

src/spanner/src/read_only_transaction.rs
```diff
@@ -318,13 +467,75 @@ mod tests {
...
+            .returning(|req| {
+                let req = req.into_inner();
+                assert_eq!(
+                    req.session,
+                    "projects/p/instances/i/databases/d/sessions/123"
+                );
+                assert_eq!(
+                    req.transaction.unwrap().selector.unwrap(),
+                    mock_v1::transaction_selector::Selector::Id(vec![1, 2, 3])
+                );
```
dbolduc, line 515: nit: Using `unwrap()` works but there is typically a Rust-y way around it. I tend to avoid `unwrap()`s, or at least prefer `.expect("nicer error message")` in tests.

Maybe:

```suggestion
                assert_eq!(
                    req.transaction.map(|t| t.selector),
                    Some(Some(mock_v1::transaction_selector::Selector::Id(vec![1, 2, 3])))
                );
```

tests/spanner/tests/driver.rs
```diff
@@ -393,4 +393,40 @@ ORDER BY col_int64
...
+        let tx = db_client.read_only_transaction().build().await?;
+
+        // Expect a read timestamp to have been chosen.
+        assert!(tx.read_timestamp().is_some());
+
+        // Execute the first query.
+        let mut rs1 = tx
+            .execute_query(Statement::builder("SELECT 1 AS col_int").build())
+            .await?;
+        let row1 = rs1.next().await.transpose()?.unwrap();
```
dbolduc: Like here I would probably say: `.expect("should yield a row")` instead of `.unwrap()`.

tests/spanner/tests/driver.rs
```diff
@@ -393,4 +393,40 @@ ORDER BY col_int64
...
+        let val1 = row1.raw_values()[0].as_string();
+        assert_eq!(val1, "1");
+        let next1 = rs1.next().await.transpose()?;
+        assert!(next1.is_none(), "{next1:?}");
+
+        // Execute the second query reusing the same transaction.
+        let mut rs2 = tx
+            .execute_query(Statement::builder("SELECT 2 AS col_int").build())
+            .await?;
+        let row2 = rs2.next().await.transpose()?.unwrap();
```
dbolduc: same

src/spanner/src/read_only_transaction.rs
dbolduc: > There's some more background info in this conversation: https://github.com/googleapis/google-cloud-rust/pull/4850/changes#r2899107755

Thanks, this is exactly what I was looking for. I probably wouldn't have thought to do it, but it makes sense.

src/spanner/src/read_only_transaction.rs
olavloite: done

tests/spanner/tests/driver.rs
olavloite: done

tests/spanner/tests/driver.rs
olavloite: done

PR #5005, opened 2026-03-12
feat(pubsub): configurable max lease

### LINKED ISSUE #4947: Configurable subscriber lease deadline
We have this thing hardcoded to 10 minutes. Some applications take longer to process a message. So this needs to be configurable as a builder option.

I would call it:

```rs
impl Subscribe {
  /// Sets the maximum lease deadline.
  ///
  /// This represents how long the application has to ack or nack a message.
  /// If the client holds a message for this long, it will consider the
  /// message expired, and nack it.
  ///
  /// The default value is 10 minutes. If it takes your application longer
  /// than 10 minutes to process a message, you should increase this value.
  pub fn set_max_deadline(mut self, v: Duration) -> Self;
}
```

dbolduc:
Fixes #4947

Let applications configure the max lease time - how long the client extends a message lease for before giving up.

nit: I am not overly concerned, about the hold time falling in the range of `[max_lease, max_lease + max_lease_extension]`.

### DIFF
--- a/src/pubsub/src/subscriber/builder.rs
+++ b/src/pubsub/src/subscriber/builder.rs
@@ -28,6 +28,7 @@ pub struct Subscribe {
     pub(super) client_id: String,
     pub(super) grpc_subchannel_count: usize,
     pub(super) ack_deadline_seconds: i32,
+    pub(super) max_lease: Duration,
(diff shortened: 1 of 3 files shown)


dbolduc:
welp, I guess I'll just try again

PR #5010, opened 2026-03-13
impl(pubsub): add exactly once processing to MessageStream

PhongChuong:
Process exactly once subscription in MessageStream.

A followup PR will update subsriber/client.rs documentation and add integration test.

Towards #3964

### DIFF
--- a/src/pubsub/src/subscriber/lease_loop.rs
+++ b/src/pubsub/src/subscriber/lease_loop.rs
@@ -14,7 +14,7 @@
 
 use super::handler::Action;
 use super::lease_state::{LeaseEvent, LeaseOptions, LeaseState, NewMessage};
-use super::leaser::Leaser;
+use super::leaser::{ConfirmedAcks, Leaser};
 use tokio::sync::mpsc::{UnboundedReceiver, UnboundedSender, unbounded_channel};
 use tokio::task::JoinHandle;
 
@@ -29,7 +29,11 @@ pub(super) struct LeaseLoop {
 }
 
 impl LeaseLoop {
-    pub(super) fn new<L>(leaser: L, options: LeaseOptions) -> Self
+    pub(super) fn new<L>(
+        leaser: L,
+        mut confirmed_rx: UnboundedReceiver<ConfirmedAcks>,
+        options: LeaseOptions,
+    ) -> Self
     where
         L: Leaser + Clone + Send + 'static,
     {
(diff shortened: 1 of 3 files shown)


PhongChuong:
This PR is largely based off @dbolduc implementation in https://github.com/googleapis/google-cloud-rust/compare/main...dbolduc:google-cloud-rust:impl-pubsub-exactly-once-lease-state-the-rest

Sorry for the large PR but it's mostly adding and updating tests.

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -270,6 +270,10 @@ impl MessageStream {
             Err(e) => return Some(Err(e)),
         };
 
+        let exactly_once = resp
+            .subscription_properties
+            .is_some_and(|m| m.exactly_once_delivery_enabled);
```
dbolduc, line 275: Please revert this piece of the PR and add it back in a follow up. I am not going to look at the message_stream tests.

src/pubsub/src/subscriber/lease_loop.rs
```diff
@@ -58,8 +62,14 @@ impl LeaseLoop {
                             None => break,
                             Some(Action::Ack(ack_id)) => state.process(Action::Ack(ack_id)),
                             Some(Action::Nack(ack_id)) => state.process(Action::Nack(ack_id)),
-                            // TODO(#3964) - process exactly-once acks/nacks in the lease state
-                            _ => unreachable!("we do not return exactly-once handlers yet."),
+                            Some(Action::ExactlyOnceAck(ack_id)) => state.process(Action::ExactlyOnceAck(ack_id)),
+                            Some(Action::ExactlyOnceNack(ack_id)) => state.process(Action::ExactlyOnceNack(ack_id)),
```
dbolduc, line 66:
```suggestion
                            Some(action) => state.process(action),
```

Then we don't need to duplicate all of the tests that distinguish between exactly-once and at-least-once.

src/pubsub/src/subscriber/lease_loop.rs
```diff
@@ -206,26 +247,127 @@ mod tests {
     }
 
     #[tokio_test_no_panics(start_paused = true)]
-    async fn deadline_interval() -> anyhow::Result<()> {
+    async fn flush_acks_nacks_on_interval_exactly_once() -> anyhow::Result<()> {
```
dbolduc, line 250: Duplicating these tests is not helpful IMO.

I think you should just have a `basic_exactly_once` test that gets sent `Action::ExactlyOnceAck` and `Action::ExactlyOnceNack`, and we verify that the `leaser.confirmed_ack()` gets called and the `leaser.nack()` gets called.

dbolduc: As far as testing the lease loop as an interface, we added two things.

1. forwarding actions from the application
2. processing results of confirmed acks

(Doing two things at once is already too many. I would send one PR for part 1 and another for part 2)

To test 1, we do the `basic_exactly_once` test I described above.

Then to test 2, we can just add an exactly once message, report a result on the confirmed ack channel and see that it shows up in the message's receiver.

src/pubsub/src/subscriber/lease_loop.rs
```diff
@@ -91,47 +104,75 @@ where
...
+        Action::Nack(test_id(id))
+    }
+
+    fn test_exactly_once_ack(id: i32) -> Action {
+        Action::ExactlyOnceAck(test_id(id))
+    }
+
+    fn test_exactly_once_nack(id: i32) -> Action {
+        Action::ExactlyOnceNack(test_id(id))
+    }
```
dbolduc, line 147: unhelpful, just spell these things out. And if you want to save characters you can

```rs
use Action::{Ack, Nack, ExactlyOnceAck, ExactlyOnceNack};
```

PhongChuong:
Will break this into smaller PRs with the suggestions.

PR #5018, opened 2026-03-16
feat(o11y): implement trace context propagation for HTTP and gRPC

haphungw:
Implement W3C Trace Context propagation (`traceparent` and `tracestate`) across both `reqwest` and `tonic` transports.

### DIFF
--- a/src/gax-internal/tests/http_observability.rs
+++ b/src/gax-internal/tests/http_observability.rs
@@ -524,7 +524,7 @@ mod tests {
 
         assert_eq!(got, want);
     }
-    
+
     #[tokio::test]
     async fn propagate_trace_context() {
         let server = Server::run();


src/gax-internal/tests/http_observability.rs
```diff
@@ -524,4 +524,41 @@ mod tests {
 
         assert_eq!(got, want);
     }
+    #[tokio::test]
```
coryan, line 528: nit: blank line before the test.

src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -220,9 +221,9 @@ impl<S> NoTracingTowerService<S> {
     }
 }
 
-impl<S, Req, ResBody> Service<Req> for NoTracingTowerService<S>
+impl<S, B, ResBody> Service<http::Request<B>> for NoTracingTowerService<S>
```
coryan, line 224: I think this means we need to bump the patch version of `google-cloud-gax-internal` and use the new version in `Cargo.toml`. Better to make those changes as part of this PR.

coryan: Also, run a `cargo semver-checks --all-features -p google-cloud-gax-internal` to see if we need to bump the "major" version too.

src/gax-internal/src/http.rs
```diff
@@ -275,7 +275,13 @@ impl ReqwestClient {
             .record_http(&span, attempt_info.attempt_count, method, url)
     }
 
-    async fn execute_http_inner(&self, request: reqwest::Request) -> Result<reqwest::Response> {
+    #[allow(unused_mut)]
```
coryan: I think you want to use `#[cfg_attr(not(google_cloud_unstable_tracing), allow(unused_mut))]` that way we know to remove it when `google_cloud_unstable_tracing` goes away.

coryan (changes requested):

src/gax-internal/tests/http_observability.rs
haphungw: done

src/gax-internal/src/observability/grpc_tracing.rs
haphungw: i bumped the patch, no need to update major version.

PR #5021, opened 2026-03-16
impl(pubsub): process confirmed ack in LeaseLoop

PhongChuong:
Towards #3964

### DIFF
--- a/src/pubsub/src/subscriber/lease_loop.rs
+++ b/src/pubsub/src/subscriber/lease_loop.rs
@@ -208,7 +208,7 @@ mod tests {
             .await
             .expect_confirmed_ack()
             .times(1)
-            .withf(|v| v.len() == 1 && v[0] == test_id(0))
+            .withf(|v| *v == vec![test_id(0)])
             .returning(|_| ());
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_loop.rs
```diff
@@ -162,21 +176,67 @@ mod tests {
...
+        lease_loop.ack_tx.send(Action::ExactlyOnceAck(test_id(0)))?;
+        let mut ack_results = HashMap::new();
+        ack_results.insert(test_id(0), Ok(()));
+        confirmed_tx.send(ack_results)?;
+
+        mock.lock()
+            .await
+            .expect_confirmed_ack()
+            .times(1)
+            .withf(|v| sorted(v) == test_ids(0..1))
```
dbolduc: nit: no need to sort a thing with one element

```suggestion
            .withf(|v| v == test_ids(0..1))
```

dbolduc: d'oh, never trust my code. I should have said `*v == ...`. :monkey:

PR #5026, opened 2026-03-16
feat(spanner): add key implementation

olavloite:
Adds an implementation for Key, KeyRange, and KeySet for Spanner. Keys are used for Read and Delete operations. The actual implementation of those operations will be added in follow-up pull requests.

#### Why Wrappers?

We add wrappers for Keys and related structs because:
1. It allows us to create a slightly more user-friendly API for application developers (e.g. the API prohibits the user from specifying non-sense, like a KeySet that contains both all keys and a specific key range)
2. Keys contain values, and if we were to use the generated types directly, the application developer would also need to know how to encode the various Spanner data types. E.g. the `INT64` data type is encoded as a String value. The same is also true for the `BYTES` data, which requires the data to be sent from the client to Spanner as a base64-encoded string.

### DIFF
--- a/src/spanner/src/key.rs
+++ b/src/spanner/src/key.rs
@@ -17,7 +17,7 @@ use crate::value::Value;
 /// Represents a primary key or index key for Cloud Spanner.
 ///
 /// # Example
-/// ```rust
+/// ```
 /// use google_cloud_spanner::key;
 ///
 /// let key = key![1_i64, "Alice"];
@@ -43,6 +43,13 @@ impl Key {
     pub fn new(values: Vec<Value>) -> Self {
         Key { values }
     }
+
+    pub(crate) fn into_values(self) -> Vec<serde_json::Value> {
+        self.values
+            .into_iter()
+            .map(|v| v.into_serde_value())
+            .collect()
+    }
 }
 
 impl Default for Key {
@@ -51,26 +58,17 @@ impl Default for Key {
     }
 }
 
-impl From<Key> for Vec<serde_json::Value> {
-    fn from(key: Key) -> Self {
-        key.values
-            .into_iter()
-            .map(|v| v.into_serde_value())
-            .collect()
-    }
-}
-
 /// Defines whether a boundary of a key range is open (exclusive) or closed (inclusive).
-#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[derive(Clone, Debug, PartialEq)]
 pub(crate) enum Endpoint {
-    Closed,
-    Open,
+    Closed(Key),
+    Open(Key),
 }
 
 /// Represents a contiguous range of keys.
 ///
 /// # Example
-/// ```rust
+/// ```
 /// use google_cloud_spanner::client::KeyRange;
 /// use google_cloud_spanner::key;
 ///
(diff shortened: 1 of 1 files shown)


src/spanner/src/key.rs
```diff
@@ -0,0 +1,447 @@
...
+    }
+}
+
+impl Default for Key {
+    fn default() -> Self {
+        Self::new(vec![])
+    }
+}
+
+impl From<Key> for Vec<serde_json::Value> {
```
coryan: Remember that this is part of the public API, is that desired?

src/spanner/src/key.rs
```diff
@@ -0,0 +1,447 @@
...
+/// use google_cloud_spanner::key;
+///
+/// let range = KeyRange::closed_open(key![1_i64], key![10_i64]);
+/// ```
+#[derive(Clone, Debug, PartialEq)]
+pub struct KeyRange {
+    pub(crate) start: Key,
+    pub(crate) start_endpoint: Endpoint,
+    pub(crate) end: Key,
+    pub(crate) end_endpoint: Endpoint,
```
coryan: I think I would have defined `Endpoint` as:

```rust
pub(crate) enum Endpoint {
  Closed(Key),
  Open(Key),
}
```

and then this as 

```suggestion
    pub(crate) start: Endpoint,
    pub(crate) end: Endpoint,
```

src/spanner/src/key.rs
```diff
@@ -0,0 +1,447 @@
...
+    /// # Example
+    /// ```rust
+    /// use google_cloud_spanner::client::KeyRange;
+    /// use google_cloud_spanner::key;
+    ///
+    /// // Creates a key for the range (1, 10)
+    /// let range = KeyRange::open_open(key![1_i64], key![10_i64]);
+    /// ```
+    ///
+    /// Creates a key for the range (start, end).
```
coryan: I find these more descriptive than "Constructs an open-open `KeyRange`" 🤷

src/spanner/src/key.rs
```diff
@@ -0,0 +1,447 @@
...
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum Endpoint {
+    Closed,
+    Open,
+}
+
+/// Represents a contiguous range of keys.
+///
+/// # Example
+/// ```rust
```
coryan: nit: you can omit `rust` in these, that is the default.

src/spanner/src/key.rs
olavloite: Thanks. No, that is not desired, so modified to make it private.

src/spanner/src/key.rs
olavloite: Thanks, done.

src/spanner/src/key.rs
olavloite: Yeah, I think I agree with that. I removed the 'open-open'-style comments and move the (start, end) style comments to the top of each method.

src/spanner/src/key.rs
olavloite: Ack. Done (in this file)

PR #5030, opened 2026-03-16
chore(gax-internal): inject diagnostic headers conditionally for integration tests

haphungw:
Add a test-only environment variable (`GOOGLE_CLOUD_TEST_EXTRA_HEADERS`) to generically inject diagnostic headers during execution.


haphungw:
included in #5031.

PR #5032, opened 2026-03-16
feat(pubsub): add exactly once delivery in Subscriber

PhongChuong:
Towards #3964

### DIFF
--- a/src/pubsub/src/subscriber/client.rs
+++ b/src/pubsub/src/subscriber/client.rs
@@ -57,7 +57,7 @@ use std::sync::Arc;
 ///
 /// If you enable exactly-once delivery for a subscription, your application
 /// can be opinionated about the delivery type, by destructuring the handler
-/// into its [`crate::subscriber::handler::ExactlyOnce`] branch.
+/// into its [`Handler::ExactlyOnce`][eo-branch] branch.
 ///
 /// ```
 /// use google_cloud_pubsub::subscriber::MessageStream;
@@ -105,6 +105,7 @@ use std::sync::Arc;
 ///
 /// [application default credentials]: https://cloud.google.com/docs/authentication#adc
 /// [cloud pub/sub]: https://docs.cloud.google.com/pubsub/docs/overview
+/// [eo-branch]: crate::subscriber::handler::Handler::ExactlyOnce
 /// [private google access with vpc service controls]: https://cloud.google.com/vpc-service-controls/docs/private-connectivity
 /// [pull subscription]: https://docs.cloud.google.com/pubsub/docs/pull
 /// [with_endpoint()]: ClientBuilder::with_endpoint
--- a/tests/pubsub/src/exactly_once.rs
+++ b/tests/pubsub/src/exactly_once.rs
@@ -29,9 +29,11 @@ pub async fn roundtrip(topic_name: &str, subscription_name: &str) -> anyhow::Res
     let subscribe: tokio::task::JoinHandle<Result<(), Error>> = tokio::spawn(async move {
         let mut acks = JoinSet::new();
         while acks.len() < MESSAGE_COUNT {
-            if let Some((_, Handler::ExactlyOnce(h))) = stream.next().await.transpose()? {
-                acks.spawn(h.confirmed_ack());
-            }
+            let h = match stream.next().await.transpose()? {
+                Some((_, Handler::ExactlyOnce(h))) => h,
+                _ => unreachable!("exactly-once delivery is enabled, and the stream stays open."),
+            };
+            acks.spawn(h.confirmed_ack());
         }
         while let Some(r) = acks.join_next().await {
             r?.inspect_err(|e| tracing::info!("received unexpected confirm_ack error: {e:?}"))?;


tests/pubsub/src/exactly_once.rs
```diff
@@ -0,0 +1,57 @@
...
+        let mut handles = JoinSet::new();
+        let mut confirms = HashSet::new();
+        while confirms.len() < MESSAGE_COUNT
+            && let Some((m, Handler::ExactlyOnce(h))) = stream.next().await.transpose()?
+        {
+            if confirms.insert(m.data) {
+                handles.spawn(h.confirmed_ack());
+            }
+            // Duplicate messages are ignored.
+        }
```
dbolduc: I think a duplicate means the server broke its contract. I think if N exactly-once acks succeed, then all N messages were unique.

So I think we can just say:

```suggestion
        let mut confirms = 0_usize;
        while confirms < MESSAGE_COUNT
            && let Some((m, Handler::ExactlyOnce(h))) = stream.next().await.transpose()?
        {
            handles.spawn(h.confirmed_ack());
            confirms += 1;
        }
        assert_eq!(confirms, MESSAGE_COUNT);
```

tests/pubsub/src/exactly_once.rs
```diff
@@ -0,0 +1,57 @@
...
+        if handles.join_all().await.into_iter().any(|r| r.is_err()) {
+            anyhow::bail!("received unexpected confirm_ack error")
+        }
```
dbolduc: I think if there is an error, we will want to know what it is. So maybe like:

```rs
while let Some(r) = handles.join_next().await {
    r??;
}
```

Aside: I think just using `?` will be good enough, but we could also add more color to the error message like:

```rs
while let Some(r) = handles.join_next().await {
    r?.inspect_err(|e| "received unexpected confirm_ack error: {e:?}")?;
}
```

tests/pubsub/src/exactly_once.rs
```diff
@@ -0,0 +1,57 @@
...
+        Ok(())
+    });
+
+    let mut publish = JoinSet::new();
+    for i in 0..MESSAGE_COUNT {
+        publish.spawn(publisher.publish(Message::new().set_data(format!("{i}"))));
+    }
+    publish.join_all().await;
+    tracing::info!("successfully published messages");
```
dbolduc, line 51: errr... did we? Don't we need to make sure the Results are all `Ok`? I think if any publishes fail, the subscribe while loop doesn't terminate.

```rs
while let Some(r) = publish.join_next().await {
    r??;
}
```

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -511,6 +550,64 @@ mod tests {
...
+        let client = test_client(endpoint).await?;
+        let mut stream = client.subscribe("projects/p/subscriptions/s").build();
+
+        response_tx
+            .send(Ok(test_exactly_once_response(0..2)))
+            .await?;
+        response_tx
+            .send(Ok(test_exactly_once_response(2..4)))
+            .await?;
+        // We use MAX_IDS_PER_RPC total ids to indirectly force a flush in the subscriber.
```
dbolduc, line 577: 👍

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -511,6 +550,64 @@ mod tests {
...
+        let mut ack_join_set = JoinSet::new();
+        for i in 0..MAX_IDS_PER_RPC {
+            let Some((m, Handler::ExactlyOnce(h))) = stream.next().await.transpose()? else {
+                anyhow::bail!("expected message {i}/{MAX_IDS_PER_RPC}")
+            };
+            assert_eq!(m.data, test_data(i));
+            assert_eq!(h.ack_id(), test_id(i));
+            ack_join_set.spawn(h.confirmed_ack());
+        }
+        ack_join_set.join_all().await;
```
dbolduc: Can we verify these are successful?

PhongChuong:
@dbolduc , thanks for the review. PTAL.

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -511,6 +550,64 @@ mod tests {
...
+        let mut ack_join_set = JoinSet::new();
+        for i in 0..MAX_IDS_PER_RPC {
+            let Some((m, Handler::ExactlyOnce(h))) = stream.next().await.transpose()? else {
+                anyhow::bail!("expected message {i}/{MAX_IDS_PER_RPC}")
+            };
+            assert_eq!(m.data, test_data(i));
+            assert_eq!(h.ack_id(), test_id(i));
+            ack_join_set.spawn(h.confirmed_ack());
+        }
+        assert!(ack_join_set.join_all().await.into_iter().all(|r| r.is_ok()));
```
dbolduc: If this is false, I'd like to see the error.

So probably this again:

```rs
       while let Some(r) = ack_join_set.join_next().await {
            r?.inspect_err(|e| tracing::info!("received unexpected confirm_ack error: {e:?}"))?;
        }
```

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -511,6 +550,64 @@ mod tests {
...
+            .send(Ok(test_exactly_once_response(2..4)))
+            .await?;
+        // We use MAX_IDS_PER_RPC total ids to indirectly force a flush in the subscriber.
+        // This is needed for exactly once because the current shutdown behavior does not
+        // send a final AcknowledgeRequest for application acknowledged ids.
+        response_tx
+            .send(Ok(test_exactly_once_response(4..MAX_IDS_PER_RPC)))
+            .await?;
+
+        let mut ack_join_set = JoinSet::new();
```
dbolduc: nit: I think I would call this `acks` or `results`.

This isn't a hard and fast rule, but we don't say `foos_vec` or `optional_foo` (unless there is some other representation of `foo` we want to distinguish from).

tests/pubsub/src/exactly_once.rs
```diff
@@ -0,0 +1,56 @@
...
+    const MESSAGE_COUNT: usize = 1_000;
+
+    tracing::info!("testing exactly once subscription");
+    let publisher = Publisher::builder(topic_name).build().await?;
+    let subscriber = Subscriber::builder().build().await?;
+    let mut stream = subscriber.subscribe(subscription_name).build();
+
+    let subscribe: tokio::task::JoinHandle<Result<(), Error>> = tokio::spawn(async move {
+        let mut handles = JoinSet::new();
+        while handles.len() < MESSAGE_COUNT
```
dbolduc: comment: ah, this is cleaner than what I suggested.

dbolduc: ah, this is cleaner than what I was suggesting.

tests/pubsub/src/exactly_once.rs
```diff
@@ -0,0 +1,56 @@
...
+pub async fn roundtrip(topic_name: &str, subscription_name: &str) -> anyhow::Result<()> {
+    const MESSAGE_COUNT: usize = 1_000;
+
+    tracing::info!("testing exactly once subscription");
+    let publisher = Publisher::builder(topic_name).build().await?;
+    let subscriber = Subscriber::builder().build().await?;
+    let mut stream = subscriber.subscribe(subscription_name).build();
+
+    let subscribe: tokio::task::JoinHandle<Result<(), Error>> = tokio::spawn(async move {
+        let mut handles = JoinSet::new();
```
dbolduc: nit: this is the exact same thing as the `ack_join_set` from the other fixture, but it has a different name. I think we should use identical names to reduce mental load.

tests/pubsub/src/exactly_once.rs
dbolduc: Ugh, although I see that is too slick for our MSRV.

maybe like:

```rs
while handles.len() < MESSAGE_COUNT {
  let h = match stream.next().await.transpose()? {
    Some((_, Handler::ExactlyOnce(h)) => h,
    _ => unreachable!("exactly-once delivery is enabled, and the stream stays open."),
  };
  handles.spawn(h.confirmed_ack());
}
```

and if we do that then the `assert_eq!(handles.len(), MESSAGE_COUNT);` is redundant.

dbolduc (approved):
LGTM 🚀

dbolduc:
Is the only thing left to remove the `#[allow(dead_code)] // TODO(#3964) ...` stuff?

tests/pubsub/src/exactly_once.rs
```diff
@@ -0,0 +1,55 @@
...
+    let publisher = Publisher::builder(topic_name).build().await?;
+    let subscriber = Subscriber::builder().build().await?;
+    let mut stream = subscriber.subscribe(subscription_name).build();
+
+    let subscribe: tokio::task::JoinHandle<Result<(), Error>> = tokio::spawn(async move {
+        let mut acks = JoinSet::new();
+        while acks.len() < MESSAGE_COUNT {
+            if let Some((_, Handler::ExactlyOnce(h))) = stream.next().await.transpose()? {
+                acks.spawn(h.confirmed_ack());
+            }
```
dbolduc: optional: I prefer the `unreachable!("...")` way to do things so we get a clear panic instead of an infinite while loop if something was misconfigured. 🤷

src/pubsub/src/subscriber/client.rs
```diff
@@ -53,15 +53,30 @@ use std::sync::Arc;
...
-/// The subscriber does **not** support [exactly-once] delivery yet.
+/// The subscriber supports [exactly-once] delivery.
 ///
-/// If you subscribe to a subscription with exactly-once delivery enabled, the
-/// subscriber will deliver you messages with at-least-once semantics. There is
-/// no way for you to confirm the acknowledgements from the server. Messages may
-/// get redelivered.
+/// If you enable exactly-once delivery for a subscription, your application
+/// can be opinionated about the delivery type, by destructuring the handler
+/// into its [`crate::subscriber::handler::ExactlyOnce`] branch.
```
dbolduc: Maybe:

```
/// ... [`Handler::ExactlyOnce`][eo-branch] ...
///
/// [eo-branch]: crate::subscriber::handler::Handler::ExactlyOnce
```

(Note that this links the branch of the enum, not the type inside the branch)

PhongChuong:
@dbolduc , thanks for the review. PTAL.

> Is the only thing left to remove the `#[allow(dead_code)] // TODO(#3964) ...` stuff?

Yep. I believe so.

src/pubsub/src/subscriber/message_stream.rs
PhongChuong: Thanks for the tip.

PR #5035, opened 2026-03-16
test(pubsub): add resume publish with ordering key sample

### LINKED ISSUE #5034: `pubsub_resume_publish_with_ordering_keys`

suzmue:
This is a very simple example for how to resume_publish with an ordering key. A more complex example may provide better guidance for how to recover from this use case, but this is about the depth the other samples go into for other languages.

I am working on writing a more interesting example, but I think it would benefit in review to have the simple example to compare to so we can see if that is an actual improvement in clarity or not and this example provides some benefit over it not existing at all.

Fixes #5034

### DIFF
--- a/src/pubsub/examples/src/publisher/resume_publish_with_ordering_keys.rs
+++ b/src/pubsub/examples/src/publisher/resume_publish_with_ordering_keys.rs
@@ -20,7 +20,7 @@ pub async fn sample(project: &str, topic_id: &str) -> anyhow::Result<()> {
     let publisher = Publisher::builder(format!("projects/{project}/topics/{topic_id}"))
         // Pub/Sub's ordered delivery guarantee only applies when publishes for
         // an ordering key are in the same region.
-        .with_endpoint("us-east1-pubsub.googleapis.com")
+        .with_endpoint("https://us-east1-pubsub.googleapis.com")
         .build()
         .await?;
 


src/pubsub/examples/src/publisher/resume_publish_with_ordering_keys.rs
```diff
@@ -0,0 +1,51 @@
...
+
+// [START pubsub_resume_publish_with_ordering_keys]
+use google_cloud_pubsub::client::Publisher;
+use google_cloud_pubsub::model::Message;
+
+pub async fn sample(project: &str, topic_id: &str) -> anyhow::Result<()> {
+    let publisher = Publisher::builder(format!("projects/{project}/topics/{topic_id}"))
+        // Pub/Sub's ordered delivery guarantee only applies when publishes for
+        // an ordering key are in the same region.
+        .with_endpoint("us-east1-pubsub.googleapis.com")
```
dbolduc: Needs `https://`

suzmue: done.

src/pubsub/examples/src/publisher/resume_publish_with_ordering_keys.rs
```diff
@@ -0,0 +1,51 @@
...
+            .set_ordering_key(ordering_key),
+    );
+
+    match result.await {
+        Ok(message_id) => println!("published message with ID: {message_id}"),
+        Err(e) => {
+            // Fix internal state to make sure publishes with errors are not
+            // published out of order. This might mean moving messages to a queue
+            // and retrying those messages before publishing subsequent messages.
+            println!("error publishing message: {e}");
```
dbolduc, line 40: `{e:?}`

dbolduc (approved):

src/pubsub/examples/src/publisher/resume_publish_with_ordering_keys.rs
suzmue: I think using the Display implementation probably makes sense so will leave it.

dbolduc: It loses a lot of detail

suzmue: but I might want to see the error that is designed for me as a person to read when running a sample and not have output be cluttered with a bunch of details (I also don't feel strongly either way feel free to change it).

Issue #5033, opened 2026-03-16
Align default flow control settings to canonical pub/sub client library values


suzmue:
There are two values that are slightly off:
1. max_outstanding_bytes: other libraries are 100 MB not 100 MiB
2. max lease extension: other libraries are 60 minutes, we have 10 minutes.

dbolduc:
> max lease extension

I messed up the names for these initially. If the value is how long we hold a message for, that is the "max lease" after #5005. The only problem is that I couldn't merge that PR to save my life.

suzmue:
It actually looks like more libraries are aligned on 100MiB than 100 MB, so I think we can leave that as is.

PR #5036, opened 2026-03-16
cleanup(pubsub): increase max_lease to 60 minutes

suzmue:
Update the max lease time to match other client library default of 60 minutes (increased from 10 minutes).

For #5033

### DIFF
--- a/src/pubsub/src/subscriber/builder.rs
+++ b/src/pubsub/src/subscriber/builder.rs
@@ -45,7 +45,7 @@ impl Subscribe {
             subscription,
             client_id,
             grpc_subchannel_count,
-            ack_deadline_seconds: 10,
+            ack_deadline_seconds: 60,
             max_lease: Duration::from_secs(60 * 60),
             max_outstanding_messages: 1000,
             max_outstanding_bytes: 100 * MIB,


src/pubsub/src/subscriber/builder.rs
```diff
@@ -46,7 +46,7 @@ impl Subscribe {
             client_id,
             grpc_subchannel_count,
             ack_deadline_seconds: 10,
-            max_lease: Duration::from_secs(600),
+            max_lease: Duration::from_mins(60),
```
dbolduc: hilariously, TIL: `Duration::from_mins`. I would have been using it.

coryan: Sadly a 1.91.0 feature:

https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_mins

image

dbolduc:
> Need to update this too:
> 
> https://github.com/googleapis/google-cloud-rust/blob/0a6a4efa303d0d225de82cfcdd0b16cef6830ab0/src/pubsub/src/subscriber/builder.rs#L94-L95

^ bump

src/pubsub/src/subscriber/builder.rs
suzmue: Unfortunate, fixed.

suzmue:
> > Need to update this too:
> > https://github.com/googleapis/google-cloud-rust/blob/0a6a4efa303d0d225de82cfcdd0b16cef6830ab0/src/pubsub/src/subscriber/builder.rs#L94-L95
> 
> ^ bump

Done

dbolduc (approved):

PR #5038, opened 2026-03-17
feat(spanner): read/write transaction foundation

olavloite:
Adds the foundation for read/write transactions. The API that is added in this change is not (yet) public, as the eventual API for read/write transactions will include a transaction runner. That is, a method that allows the application developer to submit the transaction logic as a lambda that should be executed in the context of a read/write transaction, and that should automatically be retried by the client if Spanner aborts the transaction.

### DIFF
--- a/src/spanner/src/read_write_transaction.rs
+++ b/src/spanner/src/read_write_transaction.rs
@@ -125,8 +125,9 @@ impl ReadWriteTransaction {
             .ok_or_else(|| crate::error::internal_error("No stats returned"))?;
         match stats.row_count {
             Some(RowCount::RowCountExact(c)) => Ok(c),
-            Some(RowCount::RowCountLowerBound(c)) => Ok(c),
-            None => Ok(0),
+            _ => Err(crate::error::internal_error(
+                "ExecuteSql returned an invalid or missing row count type for a read/write transaction",
+            )),
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/spanner/src/read_write_transaction.rs
```diff
@@ -0,0 +1,440 @@
...
+            seqno: std::sync::atomic::AtomicI64::new(1),
+        })
+    }
+}
+
+/// A read-write transaction.
+#[derive(Debug)]
+pub(crate) struct ReadWriteTransaction {
+    pub(crate) context: ReadContext,
+    seqno: std::sync::atomic::AtomicI64,
```
coryan, line 89: Do you expect a single transaction to be used from multiple threads, concurrently?

If so, then this may be appropriate, but otherwise you should consider a simple `i64` and use `&mut self` for the functions that mutate its value.

src/spanner/src/read_write_transaction.rs
```diff
@@ -0,0 +1,440 @@
...
+            .spanner
+            .execute_sql(request, crate::RequestOptions::default())
+            .await?;
+
+        let stats = response
+            .stats
+            .ok_or_else(|| crate::error::internal_error("No stats returned"))?;
+        match stats.row_count {
+            Some(RowCount::RowCountExact(c)) => Ok(c),
+            Some(RowCount::RowCountLowerBound(c)) => Ok(c),
```
coryan: This is dropping some information.  Maybe consider something like:

https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.size_hint

src/spanner/src/read_write_transaction.rs
olavloite: Yes, we want to allow applications to use both read-only and read/write transactions concurrently. Specifically for DML statements, this is only 'half-supported' by Spanner at the moment, as DML statements must arrive in order on Spanner (although this could be relaxed in the future). Queries and reads are always allowed to be executed in parallel for both read-only and read/write transactions.

src/spanner/src/read_write_transaction.rs
olavloite: Ah, good point. Spanner should never return a lower-bound row count in a case like this. The result type should always be an RowCountExact. I've modified this block, so that it returns an internal error if the return type is anything other than RowCountExact.

PR #5042, opened 2026-03-17
fix(storage): BidiWriteObject PUT truncation on RAPID/zonal buckets + buffer optimizations

russfellows:
# Fix code to operate correctly with Rapid storage gRPC api's
### Summary

This PR fixes a critical data-correctness bug where objects uploaded via `BidiWriteObject` to GCS RAPID (zonal) buckets are **silently truncated** — e.g., a 16 MiB PUT results in objects of 2–14 MiB. It also includes buffer allocation optimizations that reduce per-PUT memory overhead by ~47%.

### Problem

Two independent issues combine to cause truncation on RAPID/zonal buckets:

1. **Finalize race condition**: The `finish_write=true` message can reach the server before all preceding HTTP/2 DATA frames have been delivered. The server finalizes the object with whatever bytes it has received, discarding the rest. This does *not* manifest on standard (multi-region/dual-region) buckets because their write path is more tolerant of ordering.

2. **Spec-ack trap**: On appendable writes, the server sends an initial `WriteObjectResponse` containing `Resource { size: 0 }` as a "specification acknowledgment." The reader task interprets this as the final `Resource`, signaling the producer to short-circuit — preventing the flush probe from ever observing a real `PersistedSize`.

### Fix

**File: `src/storage/src/storage/transport.rs`** (core fix)

- **Two-phase flush probe**: After sending all data chunks, send `flush=true, state_lookup=true, finish_write=false` and wait until `PersistedSize >= total_len` before sending `finish_write=true`. This guarantees all data is persisted before finalization.
- **Spec-ack skip**: The reader task now skips the initial `Resource { size: 0 }` on appendable writes (guarded by `is_appendable && !initial_resource_skipped && !seen_persisted_size`).
- **Keep-alive**: A background task sends periodic empty messages during the flush-wait phase to prevent server-side idle timeout.
- Communication between producer/reader via a `tokio::sync::watch` channel carrying `PersistedSize` updates.

### Buffer Optimizations

**File: `src/storage/src/storage/write_object.rs`**

- When the caller provides the entire payload as a single `Bytes` buffer (the common case), `send_grpc()` now uses it directly instead of allocating a `BytesMut` and copying. **Eliminates 16 MiB alloc + 16 MiB memcpy per PUT** for single-chunk payloads.

**File: `src/gax-internal/src/grpc.rs`**

- Added `SizedProstCodec<T, U>` — a thin `tonic::codec::Codec` wrapper that delegates to `ProstCodec::raw_encoder()` / `raw_decoder()` with a page-aligned initial buffer (2 MiB + 4 KiB). Applied to `bidi_stream_with_status()`. **Eliminates ~8 `BytesMut` reallocations** on the first encoded message (default is 8 KiB, which must grow to 2+ MiB).

Combined effect: per-PUT allocation drops from ~34 MiB to ~18 MiB, copies from ~32 MiB to ~16 MiB.

### Supporting Changes

**File: `src/gax-internal/src/gcs_constants.rs`** (new)

- Centralized GCS/gRPC tuning constants: `DEFAULT_GRPC_WRITE_CHUNK_SIZE` (2 MiB), `MAX_GRPC_WRITE_CHUNK_SIZE` (~3.94 MiB), server max message size, HTTP/2 initial window size, and environment variable overrides.

**File: `src/storage/src/constants.rs`** (new)

- Storage-layer constants: write chunk size, producer channel capacity.

**Files: `src/storage/src/storage/bidi/connector.rs`, `bidi/transport.rs`, `bidi/worker.rs`**

- HTTP/2 window size tuning and reconnect telemetry for RAPID bucket connections.

### Testing

Validated on a GCS RAPID (zonal) bucket (`US-CENTRAL1`) with 16 MiB object PUTs:
- **Before fix**: Objects consistently truncated (2–14 MiB)
- **After fix**: All 8/8 objects at exactly 16,777,216 bytes — PUT, stat, and GET all confirmed correct
- Buffer optimizations tested on the same cluster workload with no regressions

### Documentation

- `doc/RAPID_PUT_TRUNCATION_FIX_HISTORY.md` — detailed history of 6 fix attempts with protocol traces and failure analysis
- `doc/RAPID_BUFFER_ALLOCATION_ANALYSIS.md` — end-to-end buffer allocation chain analysis across 6 layers
- `doc/Changelog.md` — summary of all changes in this fork

---


coryan:
Is this targeting the right repository?

coryan:
@joshuatants FYI...

russfellows:
Yes, this is the correct repo.  Also, I just signed the CLA.

coryan:
Oh wow, thanks!

I have moved away from the development of GCS clients.  @joshuatants should review and ultimately approve this.  I will make a brief pass and get the CI builds started.

src/gax-internal/src/gcs_constants.rs
```diff
@@ -0,0 +1,82 @@
...
+/// configuration at or below 100 Gbps.
+///
+/// Set `S3DLIO_GRPC_INITIAL_WINDOW_MIB=0` to fall back to the protocol default.
+pub const DEFAULT_WINDOW_MIB: u64 = 128;
+
+/// Environment variable that overrides [`DEFAULT_WINDOW_MIB`] at runtime.
+///
+/// Value is interpreted as an integer number of MiB.  Set to `0` to use the
+/// HTTP/2 protocol default (65 KB).
+pub const ENV_GRPC_INITIAL_WINDOW_MIB: &str = "S3DLIO_GRPC_INITIAL_WINDOW_MIB";
```
coryan, line 29: In general we avoid environment variables for tunable parameters.  And in any case we would not name it `S3DLIO_GRPC_*`.  I think we should remove these and just rely on options provided in the request (or when the client is initialized).

src/gax-internal/src/gcs_constants.rs
```diff
@@ -0,0 +1,82 @@
...
+/// RESOURCE_EXHAUSTED: SERVER: Received message larger than max (N vs. 4194304)
+/// ```
+/// This limit applies to the **entire serialized protobuf message** — including
+/// field tags, varint-encoded lengths, `write_offset`, `ChecksummedData` wrapper,
+/// per-chunk CRC32C, and (on the first message) `WriteObjectSpec`.  Observed
+/// framing overhead is ~88–90 bytes for middle chunks.
+///
+/// **Do not use this as a data chunk size.**  Use [`MAX_GRPC_WRITE_CHUNK_SIZE`]
+/// for the data payload limit; it already accounts for overhead.
+pub const GCS_SERVER_MAX_MESSAGE_SIZE: usize = 4 * 1024 * 1024; // 4,194,304 bytes
```
coryan, line 46: This kind of GCS-specific constant needs to be in the GCS library.

src/gax-internal/src/gcs_constants.rs
```diff
@@ -0,0 +1,82 @@
...
+/// aligned; non-aligned values will be rounded down to the nearest 64 KiB
+/// boundary by the transport layer.
+pub const DEFAULT_GRPC_WRITE_CHUNK_SIZE: usize = 2 * 1024 * 1024; // 2,097,152 bytes (32 × 64 KiB)
+
+/// Environment variable that overrides [`DEFAULT_GRPC_WRITE_CHUNK_SIZE`] at runtime.
+///
+/// Value is interpreted as an integer number of bytes.  Values exceeding
+/// [`MAX_GRPC_WRITE_CHUNK_SIZE`] (4,128,768) are silently clamped to the maximum.
+/// Example: `export S3DLIO_GRPC_WRITE_CHUNK_SIZE=4128768` for 4 MiB - 64 KiB chunks.
+pub const ENV_GRPC_WRITE_CHUNK_SIZE: &str = "S3DLIO_GRPC_WRITE_CHUNK_SIZE";
```
coryan, line 82: As above, we would not use an environment variable for this, and would not name it `S3DLIO_*`...

src/storage/src/storage/bidi/connector.rs
```diff
@@ -92,17 +97,53 @@ where
         let client = self.client.clone();
         let options = self.options.clone();
         let spec = self.spec.clone();
-        let sleep = async |backoff| tokio::time::sleep(backoff).await;
+        let sleep = async |backoff| {
+            tracing::info!(
+                "gcs bidi read backoff: sleeping {:?} before next connect/reconnect attempt",
+                backoff
+            );
```
coryan, line 104: I am sure this is useful for troubleshooting, but please remove. Eventually the backoff policy will be made public and applications can (safely) write their own backoff policy with logging.

src/storage/src/storage/bidi/connector.rs
```diff
@@ -92,17 +97,53 @@ where
...
+            tokio::time::sleep(backoff).await;
+        };
         let default_timeout = self.options.bidi_attempt_timeout;
         // Move the copies and invoke the retry loop to call `connect_attempt()`.
         let inner = async move |d: Option<Duration>| {
             let attempt_timeout = std::cmp::min(default_timeout, d.unwrap_or(default_timeout));
+            tracing::info!(
+                "gcs bidi read connect attempt started (attempt-timeout={:?})",
+                attempt_timeout
+            );
```
coryan, line 114: Ditto, please remove the troubleshooting logs.

src/storage/src/storage/bidi/connector.rs
```diff
@@ -120,9 +161,20 @@ where
         let error = handle_redirect(self.spec.clone(), status);
         // This is a new attempt, increment the count.
         self.reconnect_attempts += 1;
+        tracing::info!(
+            "gcs bidi read reconnect requested (attempt={}, status={})",
+            self.reconnect_attempts,
+            error
+        );
```
coryan, line 168: Ditto.

src/storage/src/storage/bidi/connector.rs
```diff
@@ -182,7 +234,7 @@ where
                 s + &format!("&routing_token={token}")
             });
 
-        let (tx, rx) = tokio::sync::mpsc::channel::<BidiReadObjectRequest>(100);
+        let (tx, rx) = tokio::sync::mpsc::channel::<BidiReadObjectRequest>(BIDI_REQUEST_CHANNEL_CAPACITY);
```
coryan, line 237: Thanks. FWIW, the intent is to make this number a configuration option to `open_object()` though it is unclear if it has any measurable effect.

src/storage/src/storage/write_object.rs
```diff
@@ -993,6 +1019,55 @@ where
     /// println!("response details={response:?}");
     /// # Ok(()) }
     /// ```
+    /// Upload via gRPC BidiWriteObject, for RAPID/zonal buckets.
+    ///
+    /// Collects the full payload into a contiguous `Bytes` buffer, then calls
+    /// `write_object_grpc` which streams the data in 2 MiB chunks over
+    /// gRPC with per-chunk CRC32C and whole-object checksums.
```
coryan, line 1026: That seems like a bad idea uploads can be large (up to 5 TiB).  You probably want something more akin to `send_buffered()` where the data is sent and on an upload error the upload resumes from the last byte successfully uploaded.

src/storage/src/constants.rs
```diff
@@ -0,0 +1,42 @@
+// Copyright 2025 Google LLC
```
coryan, line 1: nit: 2026

src/storage/src/constants.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// Channel capacity for bidi request fan-out between workers and range readers.
+pub const BIDI_REQUEST_CHANNEL_CAPACITY: usize = 100;
+
+/// Batch size used by worker `recv_many()` when draining new range requests.
+pub const BIDI_REQUEST_RECV_MANY_BATCH_SIZE: usize = 16;
+
+/// Log interval (seconds) for "still waiting" progress messages during
+/// bidi connect/reconnect attempts.
+pub const BIDI_CONNECT_PROGRESS_LOG_INTERVAL_SECS: u64 = 5;
```
coryan, line 42: Since logging should not be the default, we should remove this constant too.

src/storage/src/constants.rs
```diff
@@ -0,0 +1,42 @@
...
+//! - [`BIDI_REQUEST_CHANNEL_CAPACITY`]
+//!   - `storage/bidi/connector.rs` internal request stream channel.
+//!   - `storage/bidi/transport.rs` worker/read-range channels.
+//! - [`BIDI_REQUEST_RECV_MANY_BATCH_SIZE`]
+//!   - `storage/bidi/worker.rs` batching for incoming read-range requests.
+//! - [`BIDI_CONNECT_PROGRESS_LOG_INTERVAL_SECS`]
+//!   - `storage/bidi/connector.rs` progress log cadence while connect attempts wait.
+
+/// Default per-attempt timeout (seconds) for bidi open/reconnect attempts.
+pub const DEFAULT_BIDI_ATTEMPT_TIMEOUT_SECS: u64 = 60;
```
coryan, line 32: Note that this number was found empirically, there is no reason to believe it is perfect.  The only thing we know is that anything over 300 seconds returns in `Code::Unknown` errors with a "408: Timeout" message.

src/storage/src/constants.rs
```diff
@@ -0,0 +1,42 @@
...
+
+/// Channel capacity for bidi request fan-out between workers and range readers.
+pub const BIDI_REQUEST_CHANNEL_CAPACITY: usize = 100;
+
+/// Batch size used by worker `recv_many()` when draining new range requests.
+pub const BIDI_REQUEST_RECV_MANY_BATCH_SIZE: usize = 16;
```
coryan, line 38: Likewise, these numbers are pure guesses, please add a comment so "future us" don't believe that they are somewhat sacred.

russfellows:
In general, I agree with all of your comments.  Should I make these changes, or is it better / easier if you do?  I was primarily providing this as a working example of how to get gRPC and Rapid storage working.  

One note:  I think the constants I am setting were hard coded originally.  I agree, hardcoded magic numbers in code is VERY bad.  So, I typically try to pull them all out and define things in a "constants.rs" file somewhere, which is what I have done.  That is better than being embedded in the code, but not perfect.  If you have a better way to implement, then please, be my guest.  

I would like to get some good set of code for Rapid storage integrated into your branch.  IT should make your future efforts go faster by starting with a working code base.  It can always be improved, but working is always a good start.

coryan:
> In general, I agree with all of your comments. Should I make these changes, or is it better / easier if you do? I was primarily providing this as a working example of how to get gRPC and Rapid storage working.

I will let @cjc25 and @joshuatants answer that question.

> I would like to get some good set of code for Rapid storage integrated into your branch. IT should make your future efforts go faster by starting with a working code base. It can always be improved, but working is always a good start.

I am afraid it is not that easy.  If we add any public APIs to the `main`  branch the APIs will go out with the next release. Once released, any changes are breaking to somebody somewhere, and will require bumping the major version. And bumping the major version delays adoption.

There are ways to avoid these problems: (a) work on a branch other than main, (b) protect all the public APIs with a `--cfg google_cloud_unstable_storage_bidi` flag (the name can be changed).  Again, I will let @joshuatants and @cjc25 decide how they want to handle that.

cjc25:
Thanks @russfellows! For now I think we will close this PR - @joshuatants will take on the write path afterwards. I agree with @coryan that we should not start out with this code in public APIs. For example finalization should be optional for appendable objects, and we want to make sure that the public API doesn't overly constrain us. We probably also want to use a name other than `send_grpc` to keep function names usage-descriptive rather than implementation-specific, although I am not dogmatic about that.

We should regardless pick up the `SizedProstCodec` extra-4K allocation and zero-copy improvements with gRPC implementation of `write_object`.

PR #5043, opened 2026-03-17
cleanup(pubsub): remove unused doc link

PhongChuong:


PhongChuong:
Forgot to remove this link in https://github.com/googleapis/google-cloud-rust/pull/5032

dbolduc:
I opened #5044 for the build failure. I think our default lease extension is too tight, so the server is expiring messages.

PR #5046, opened 2026-03-17
feat(spanner): helper for transaction retries

olavloite:
Adds a helper function for retrying aborted transactions.

### DIFF
--- a/src/spanner/src/partitioned_dml_transaction.rs
+++ b/src/spanner/src/partitioned_dml_transaction.rs
@@ -77,7 +77,7 @@ impl PartitionedDmlTransactionBuilder {
     ///
     /// The client will retry the entire transaction if it is aborted by Spanner.
     /// This policy can be used to customize whether a transaction should be retried
-    /// or not. The default is to retry indefinetly until the transaction succeeds.
+    /// or not. The default is to retry indefinitely until the transaction succeeds.
     pub fn with_retry_policy<P: TransactionRetryPolicy + 'static>(mut self, policy: P) -> Self {
         self.retry_policy = Box::new(policy);
         self


src/spanner/src/transaction_retry_policy.rs
```diff
@@ -0,0 +1,314 @@
...
+    Fut: std::future::Future<Output = crate::Result<T>>,
+{
+    let start_time = tokio::time::Instant::now();
+    let mut attempts: u32 = 0;
+
+    // This backoff is only used if Spanner does not return a retry delay.
+    let backoff = ExponentialBackoffBuilder::new()
+        .with_initial_delay(Duration::from_millis(10))
+        .with_maximum_delay(Duration::from_secs(1))
+        .with_scaling(1.3)
```
coryan, line 90: This is a common value in a lot of Google code, but I cannot find a source explaining why it is better than 2.0... I think it is cargo culted from one service to the next.

src/spanner/src/transaction_retry_policy.rs
```diff
@@ -0,0 +1,314 @@
...
+}
+
+/// Helper method to execute an asynchronous closure, retrying it if the
+/// transaction is aborted by the server.
+///
+/// This is used for operations like Partitioned DML transactions in Cloud Spanner, where
+/// the server may abort the transaction due to transient issues, indicating that the client
+/// should re-attempt the entire operation.
+#[allow(dead_code)]
+pub(crate) async fn retry_aborted<T, F, Fut>(
```
coryan, line 75: You could probably rewrite this using:

https://github.com/googleapis/google-cloud-rust/blob/62730bde8fb509dd966c520774a163f61778e435/src/gax/src/retry_loop_internal.rs#L51

🤷 

Just in passing, there are reasons why you may want to throttle retries:

https://github.com/googleapis/google-cloud-rust/blob/62730bde8fb509dd966c520774a163f61778e435/src/gax/src/retry_throttler.rs#L22-L42

olavloite:
> You are making the retry strategy very different for this service vs. all the other services in the Rust SDK.
> 
> I think it is valuable to reuse the same concepts across the SDK.

See also the comment here: https://github.com/googleapis/google-cloud-rust/pull/5047#discussion_r2952107999

This retry function is specifically for retrying __aborted transactions in Spanner__. That is different from retrying RPC invocations in a number of ways:
1. The entire transaction must be retried. This means that the entire application business logic of the transaction must be retried, which again could mean that other RPCs than the initial attempt will be retried.
2. There is no 'choice' as to when the transaction should be retried (e.g. 'also retry the transaction if `Unavailable` was returned by the Commit RPC' is not a valid choice). It must be retried (or abandoned) when Spanner returns `Aborted`. Period.
3. Retrying a transaction also requires the client library to include the ID of the previous transaction if it wants Spanner to bump the priority of the transaction that is being retried. (This is not included in this PR, it will be added in a separate PR that adds more logic for the 'multiplexed session read/write transaction protocol').
4. Spanner will in all well-behaved cases include a retry delay in the status that is returned when the transaction is aborted. The client-side backoff logic in this retry helper is therefore purely a fallback for unknown/ill-behaved cases, and should normally not be used.

coryan:
> > You are making the retry strategy very different for this service vs. all the other services in the Rust SDK.
> > I think it is valuable to reuse the same concepts across the SDK.
> 
> See also the comment here: [#5047 (comment)](https://github.com/googleapis/google-cloud-rust/pull/5047#discussion_r2952107999)
> 
> This retry function is specifically for retrying **aborted transactions in Spanner**. That is different from retrying RPC invocations in a number of ways:

I was not clear.  When I said "the same concepts" I was thinking about something like:

```rust
pub trait TransactionRetryPolicy {
   fn on_abort() -> bool; // I would probably return an enum
}

pub trait DefaultTransactionRetryPolicy;
impl TransactionRetryPolicy {
  fn on_abort() -> bool { ... }
}

pub trait TransactionRetryPolicyExt {
  fn with_attempt_count_limit(self, limit: u32) -> LimitedAttemptCountPolicy<Self> {
    LimitedAttemptCountPolicy::new(self, limit)
  }
  fn with_time_limit(self, maximum: Duration) -> LimitedTimePolicy<Self> {
    LimitedTimePolicy::new(self, maximum)
  }
}

impl TransactionRetryPolicyExt for T where T: TransactionRetryPolicy {}
```

Not the same types as retry policies, not even the same APIs, but at least feel more similar.

olavloite:
@coryan Thanks for the suggestion. I've refactored the implementation to use a policy instead of only a simple 'configuration struct'. Would you mind taking another look?

src/spanner/src/partitioned_dml_transaction.rs
```diff
@@ -247,8 +249,7 @@ mod tests {
         assert_eq!(res, 500);
     }
 
-    #[tokio::test]
-    #[ignore = "Transaction retries will be implemented in a subsequent PR"]
+    #[tokio::test(start_paused = true)]
```
coryan, line 252: Nice.

PR #5047, opened 2026-03-17
feat(spanner): partitioned DML transactions

olavloite:
Adds support for executing Partitioned DML transactions. Automatic retries of aborted errors will be added in a separate pull request.

### DIFF
--- a/src/spanner/src/database_client.rs
+++ b/src/spanner/src/database_client.rs
@@ -121,6 +121,14 @@ impl DatabaseClient {
     pub fn partitioned_dml_transaction(&self) -> PartitionedDmlTransactionBuilder {
         PartitionedDmlTransactionBuilder::new(self.clone())
     }
+
+    /// Returns a builder for a read-write transaction.
+    #[allow(dead_code)]
+    pub(crate) fn read_write_transaction(
+        &self,
+    ) -> crate::read_write_transaction::ReadWriteTransactionBuilder {
+        crate::read_write_transaction::ReadWriteTransactionBuilder::new(self.clone())
+    }
 }
 
 /// A builder for [DatabaseClient].
--- a/src/spanner/src/error.rs
+++ b/src/spanner/src/error.rs
@@ -34,3 +34,7 @@ impl SpannerInternalError {
         Self::UnexpectedData(message.into())
     }
 }
+
+pub(crate) fn internal_error(message: impl Into<String>) -> crate::Error {
+    crate::Error::deser(SpannerInternalError::new(message))
+}
--- a/src/spanner/src/lib.rs
+++ b/src/spanner/src/lib.rs
@@ -42,6 +42,7 @@ pub(crate) mod from_value;
 pub(crate) mod key;
 pub(crate) mod partitioned_dml_transaction;
 pub(crate) mod read_only_transaction;
+pub(crate) mod read_write_transaction;
 pub(crate) mod result_set;
 pub(crate) mod result_set_metadata;
 pub(crate) mod row;
--- a/src/spanner/src/read_only_transaction.rs
+++ b/src/spanner/src/read_only_transaction.rs
@@ -289,8 +289,8 @@ impl MultiUseReadOnlyTransaction {
 
 #[derive(Clone, Debug)]
 pub(crate) struct ReadContext {
-    client: DatabaseClient,
-    transaction_selector: crate::model::TransactionSelector,
+    pub(crate) client: DatabaseClient,
+    pub(crate) transaction_selector: crate::model::TransactionSelector,
 }
 
 impl ReadContext {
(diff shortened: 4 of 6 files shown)


src/spanner/src/transaction_retry_settings.rs
coryan, line 1: nit: the module name is unfortunate.  `transaction_retry_settings.rs` seems less likely to become a dumping ground.

src/spanner/src/transaction_retry_settings.rs
```diff
@@ -0,0 +1,51 @@
...
+use std::time::Duration;
+
+/// Configuration for automatically retrying a transaction when it is aborted.
+#[derive(Clone, Debug)]
+pub struct TransactionRetrySettings {
+    /// The maximum number of attempts to make. If 0, this field is ignored.
+    pub max_attempts: u32,
+    /// The total maximum time to spend retrying. If 0, this field is ignored.
+    pub total_timeout: Duration,
+}
```
coryan, line 24: I encourage you to think about "retry policies" instead of "settings". Before you know it, you want to log every retry attempt, or increment some global counter, or applications want to stop early if they get 3 "unavailable" errors, but want to continue even after 10 "aborted" errrors.

src/spanner/src/partitioned_dml_transaction.rs
```diff
@@ -0,0 +1,382 @@
...
+) -> crate::Result<i64> {
+    let mut lower_bound: Option<i64> = None;
+    while let Some(message_res) = stream.next_message().await {
+        let prs = message_res?;
+        if let Some(stats) = prs.stats {
+            if let Some(RowCountLowerBound(val)) = stats.row_count {
+                lower_bound = Some(val);
+            }
+        }
+    }
```
coryan: I think you can simplify this to:

```suggestion
    while let Some(prs) = stream.next_message().await.transpose()? {
        if let Some(RowCountLowerBound(val)) = prs.stats.and_then(|s| s.row_count) {
                lower_bound = Some(val);
        }
    }
```

src/spanner/src/partitioned_dml_transaction.rs
```diff
@@ -0,0 +1,382 @@
...
+    match lower_bound {
+        Some(lb) => Ok(lb),
+        None => Err(crate::Error::deser(
+            crate::error::SpannerInternalError::new(
+                "ExecuteStreamingSql completed successfully but no row_count_lower_bound was returned",
+            ),
+        )),
+    }
```
coryan: I probably missed a parenthesis:

```suggestion
    lower_bound.ok_or_else(|| crate::Error::deser(
            crate::error::SpannerInternalError::new(
                "ExecuteStreamingSql completed successfully but no row_count_lower_bound was returned",
            )))
```

src/spanner/src/transaction_retry_settings.rs
olavloite: Done

src/spanner/src/transaction_retry_settings.rs
olavloite: This is actually a very conscious choice. These settings are only intended for retrying __aborted transactions__. This part is indeed something that often causes some confusion and application developer errors when working with Spanner, as there are two different types of retries that need to be handled by either the client library and/or the application:
1. Retrying a transient error for a single RPC: The most common is that an RPC fails with an `Unavailable` error, and the RPC is retried N times. For this, it is natural that an application developer might want some more say in both when the RPC should be retried, and how that should happen, meaning that something like a `RetryPolicy` makes sense.
2. Retrying aborted transactions: This is a Spanner-specific thing and requires that __the entire transaction__ is retried. There is also no choice for the application developer when it comes to __what conditions should lead to a potential retry__. The only valid condition is when Spanner returns an `Aborted` error. Retrying a transaction because one of the RPCs returned `Unavailable` is for example not  a valid retry condition, at least not in the Spanner API.
3. There could be __application business logic__ that looks a bit like point 2 above, where an application says something like 'if one of the RPCs fails with `DeadlineExceeded`, then I want the application to retry the transaction'. But this type of retry would technically be different, because:
3.1. It would require the application to manually rollback the transaction. A `DeadlineExceeded` error does not invalidate the transaction on Spanner, and failing to manually rollback the transaction would keep it alive and have it hold on to its locks. An __aborted transaction__ is however no longer active on Spanner and does not need to be rolled back.
3.2. Spanner internally keeps track of how often a transaction has been aborted by Spanner, and when a transaction is retried, it will be assigned a higher priority by Spanner. This again will reduce the probability that the transaction will be aborted again by Spanner, which again means that the probability that the transaction retry will succeed is higher. Rolling back a transaction and 'retrying it manually' does not give the transaction higher priority.

All of this means that we deliberately limit the scope of these retry settings (e.g. we do not allow an application to say something like 'I also want you to retry `Unavailable` errors).

src/spanner/src/partitioned_dml_transaction.rs
olavloite: Done

src/spanner/src/partitioned_dml_transaction.rs
olavloite: Done

PR #5052, opened 2026-03-18
feat(spanner): add read operation support

olavloite:
Adds support for the key/value-based Read API in Spanner. Read operations are mapped to the StreamingRead RPC, which returns the same type of stream of PartialResultSet as the ExecuteStreamingSql RPC.


src/spanner/src/read_only_transaction.rs
```diff
@@ -142,6 +142,42 @@ impl SingleUseReadOnlyTransaction {
...
+    /// ```
+    /// # use google_cloud_spanner::client::{Spanner, Read, KeySet};
+    /// # use google_cloud_spanner::key;
+    /// # async fn run(spanner: Spanner) -> Result<(), google_cloud_spanner::Error> {
+    /// let db_client = spanner.database_client("projects/p/instances/i/databases/d").build().await?;
+    /// let transaction = db_client.single_use().build();
+    ///
+    /// // Read using the primary key
+    /// let read_by_pk = Read::new("Users", vec!["Id", "Name"]).with_keys(KeySet::all());
+    /// let mut result_set = transaction.read(read_by_pk).await?;
```
coryan: This feels awkward.  I think you should aim for something like:


```suggestion
    /// let mut result_set = transaction.read("Users", ["Id", "Name"]).with_keys(KeySet::all()).send().await?;
```

olavloite: Hmm.... I see what you mean there. My intention was to keep reads consistent with how queries are handled. I've therefore refactored this a bit to the following:

1. Renamed the method from `read` to `execute_read`
2. Renamed `struct Read` to `struct ReadRequest`. A `ReadRequest` is comparable to a `Statement` for queries.
3. Introduced builders for `ReadRequest`.

There are a couple of goals I wanted to achieve with this approach:
1. A `ReadRequest` should be independent of a transaction. Just like a `Statement`, it is something that an application should be able to construct outside the context of a transaction, and possibly also re-use across multiple transactions.
2. `ReadRequest` has a small quirk when it comes to how it is configured; You must always supply a table name, column names, and a `KeySet`. You can optionally also supply an index name. The `KeySet` is either used to filter based on the primary key of the table (if no index has been specified), or based on the secondary index (if specified). That is the reason for linking `KeySet` with `with_keys` and `with_index`, and requiring that only one of them is used.

Does this make sense to you?

src/spanner/src/read.rs
```diff
@@ -0,0 +1,211 @@
...
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_spanner::client::{ReadRequest, KeySet};
+    /// let request = ReadRequest::builder("Users", vec!["Id", "Name"]).with_keys(KeySet::all());
+    /// ```
+    ///
+    /// The `keys` parameter identifies the rows to be yielded by naming the primary keys
+    /// of the rows in the table. Rows are yielded in table primary key order.
+    pub fn with_keys(self, keys: impl Into<KeySet>) -> ConfiguredReadRequestBuilder {
```
coryan, line 56: Seems like you are rediscovering:

https://cliffle.com/blog/rust-typestate/

also in:

https://docs.rs/bon/latest/bon/

Thought I would probably not use bon for a single builder, and I would be wary of adding it as a public API.

PR #5056, opened 2026-03-18
cleanup(pubsub): fail sample on error

dbolduc:
Q: should we prefer `eprintln!` for errors?

### DIFF
--- a/src/pubsub/examples/src/publisher/publish_with_ordering_keys.rs
+++ b/src/pubsub/examples/src/publisher/publish_with_ordering_keys.rs
@@ -41,7 +41,7 @@ pub async fn sample(project: &str, topic_id: &str) -> anyhow::Result<()> {
     for publish_future in publish_futures {
         let message_id = publish_future
             .await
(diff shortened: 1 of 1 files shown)


src/pubsub/examples/src/publisher/publish_with_ordering_keys.rs
```diff
@@ -39,10 +39,10 @@ pub async fn sample(project: &str, topic_id: &str) -> anyhow::Result<()> {
     });
 
     for publish_future in publish_futures {
-        match publish_future.await {
-            Ok(message_id) => println!("published message with ID: {message_id}"),
-            Err(e) => println!("error publishing message: {e}"),
-        }
+        let message_id = publish_future
+            .await
+            .inspect_err(|e| println!("error publishing message: {e}"))?;
```
coryan:
```suggestion
            .inspect_err(|e| eprintln!("error publishing message: {e:?}"))?;
```

dbolduc: Fixed, thanks.

PR #5060, opened 2026-03-18
cleanup(auth): supress warning on unused idtoken specific code

alvarowolfx:


src/auth/src/credentials/user_account.rs
```diff
@@ -393,6 +393,7 @@ pub(crate) struct UserTokenProvider {
 }
 
 #[derive(PartialEq)]
+#[allow(dead_code)]
 enum UserTokenSource {
     IdToken,
```
coryan, line 398: Is this an alternative? Does it produce other problems?


```suggestion
enum UserTokenSource {
#[cfg(feature = "idtoken")]
IdToken,
```

alvarowolfx: it produces other problems, because some code branches match against the enum values. I tried to put it under the feature first. For now allowing the dead_code is easier.

PR #5068, opened 2026-03-19
feat(spanner): add read/write transaction runner

olavloite:
Adds a TransactionRunner for read/write transactions. This runner automatically retries the transaction if it is aborted by Spanner.

### DIFF
--- a/src/spanner/src/transaction_runner.rs
+++ b/src/spanner/src/transaction_runner.rs
@@ -30,11 +30,11 @@ use crate::transaction_retry_policy::{
 /// let db_client = client.database_client("projects/p/instances/i/databases/d").build().await?;
 /// let runner = db_client.read_write_transaction().build().await?;
 ///
-/// let result = runner.run(|transaction| Box::pin(async move {
+/// let result = runner.run(async |transaction| {
 ///     let statement = Statement::builder("UPDATE MyTable SET MyColumn = 'MyValue' WHERE Id = 1").build();
 ///     transaction.execute_update(statement).await?;
 ///     Ok(42)
-/// })).await?;
+/// }).await?;
 /// # Ok(())
 /// # }
 /// ```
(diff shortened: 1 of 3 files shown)


src/spanner/src/transaction_runner.rs
```diff
@@ -0,0 +1,492 @@
...
+    /// If the transaction is aborted by Spanner, the closure will be retried
+    /// automatically according to the configured `TransactionRetryPolicy`.
+    ///
+    /// The transaction is automatically committed if the closure returns `Ok`.
+    /// If the closure returns `Err`, the transaction will be rolled back and
+    /// the error will be propagated.
+    pub async fn run<T, F, Fut>(self, mut work: F) -> crate::Result<T>
+    where
+        F: FnMut(ReadWriteTransaction) -> Fut,
+        Fut: std::future::Future<Output = crate::Result<T>>,
```
coryan: Since we are using Rust >= 1.85 I think you can use:

https://doc.rust-lang.org/std/ops/trait.AsyncFnMut.html

src/spanner/src/transaction_runner.rs
```diff
@@ -0,0 +1,492 @@
...
+    {
+        let start_time = tokio::time::Instant::now();
+        let mut attempts: u32 = 0;
```
coryan, line 194: The robots says this code is not tested, are they confused?

olavloite: Yes, very much so. There are plenty of unit tests in this file that cover the various scenarios.

We don't have any end-to-end tests for this yet, though. The reason for that is that there is actually one part of the underlying protocol for read/write transactions that I have not yet implemented. Multiplexed sessions require transactions to keep track of a 'pre-commit token'. I'll do that in a later PR, and then also include end-to-end tests.

src/spanner/src/transaction_runner.rs
olavloite: Thanks, done!

Issue #4335, opened 2026-01-22
refactor(auth): use generated IAM client instead of handwritten interactions with the service


alvarowolfx:
Similar to #4334, we have multiple interactions with the IAM Service in the auth crate, but so far we have been handwriting those interactions with the `reqwest` create and missing nice things like built in retry and type safety. 

One issue is that all crates depend on auth, so there is a circular dependency problem here to be solved. @coryan gave some ideas to circumvent this that I don't recall anymore.

coryan:
Should we close this?  I think you said there was no ROI on this feature?

alvarowolfx:
> Should we close this? I think you said there was no ROI on this feature?

yeah, let's close it

PR #5069, opened 2026-03-19
impl(auth): introduce auth-internal crate to allow auth crate to use generated clients

alvarowolfx:
Towards https://github.com/googleapis/google-cloud-rust/issues/4335


alvarowolfx:
closing for now, it's branching to too many changes and starting to not worth the investment

PR #5070, opened 2026-03-19
feat(spanner): write-only transactions

olavloite:
Adds write-only transaction support for the Spanner client. Write-only transactions can be used for blind writes using mutations. These can be both with and without replay protection.


src/spanner/src/write_only_transaction.rs
```diff
@@ -0,0 +1,421 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::client::{DatabaseClient, Mutation};
+use crate::model::{
+    BeginTransactionRequest, CommitRequest, CommitResponse, RequestOptions, TransactionOptions,
+    transaction_options::ReadWrite,
+};
```
coryan, line 19: FWIW, the Google style guide suggests:


```suggestion
use crate::model::{
    BeginTransactionRequest, CommitRequest, CommitResponse, RequestOptions, TransactionOptions,
};
use crate::model::transaction_options::ReadWrite;
```

I have found that particular guideline useful, but not enough to enforce it automatically (yet).

PR #5071, opened 2026-03-19
feat(o11y):  extract and execute T4 span logs

### LINKED ISSUE #4946: Per-RPC Error Logging
Logs should be emitted for errors on HTTP or gRPC requests, attached to each T4 span.  In addition to common transport level attributes like response code and local error messages, these logs have the following attributes:

* `error.type` e.g. RATE_LIMIT_EXCEEDED
* `gcp.error.domain` e.g. `"pubsub.googleapis.com"`
* `gcp.error.metadata.<key> e.g. `gcp.error.metadata.zone: "us-east1-a"`

These values should be parsed from the ErrorInfo attached to the response as `error`, if any.

I suggest they are logged at "INFO" since we will be recording them on retryable requests.

haphungw:
Parse and map AIP 193 `ErrorInfo` telemetry payloads to retain `domain` and `metadata` components.
Create an internal helper `emit_error_log` to bind attributes into OpenTelemetry spans for T4 logging, integrated for both http and grpc tracing.

The next PR will add an integration test for T4 logging for http tracing.

Fixes #4946

### DIFF
--- a/src/gax-internal/Cargo.toml
+++ b/src/gax-internal/Cargo.toml
@@ -141,7 +141,7 @@ test-case.workspace               = true
 tokio                             = { workspace = true, features = ["test-util"] }
 tracing                           = { workspace = true, features = ["std"] }
 tracing-opentelemetry             = { workspace = true }
-tracing-subscriber                = { workspace = true, features = ["registry", "std", "json"] }
+tracing-subscriber                = { workspace = true, features = ["json", "registry", "std"] }
 # Local crates
 echo-server = { path = "echo-server" }
 grpc-server = { path = "grpc-server" }


src/gax-internal/src/observability/errors.rs
```diff
@@ -86,6 +99,65 @@ impl ErrorType {
...
+                domain, metadata, ..
+            }
+            | ErrorType::RpcError {
+                domain, metadata, ..
+            } => (domain.as_deref(), metadata.as_ref()),
+            _ => (None, None),
+        };
+
+        if let Some(domain) = domain {
+            span.record(GCP_ERRORS_DOMAIN, domain);
```
westarle: we shouldn't mutate the T4 span this way, we can remove this.

src/gax-internal/src/observability/errors.rs
```diff
@@ -86,6 +99,65 @@ impl ErrorType {
...
+
+            let ctx = span.context();
+            let otel_span = ctx.span();
+            for (k, v) in meta {
+                otel_span.set_attribute(opentelemetry::KeyValue::new(
+                    format!("{GCP_ERRORS_METADATA_PREFIX}{k}"),
+                    v.clone(),
+                ));
+            }
+        }
```
westarle: We should operate in the domain of tracing exclusively and not require OTel to pack this data.  I know you can't dynamically construct a event! in the way we'd hope.

If you have access to the raw JSON string for metadata from the ErrorInfo, you could push it into a field called `gcp.errors.metadata` (or `gcp.errors.metadata.json`).  Then downstream processors can unpack it if needed.

src/gax-internal/src/observability/errors.rs
```diff
@@ -86,6 +99,65 @@ impl ErrorType {
...
+
+        match http_status_code {
+            Some(http_code) => tracing::event!(
+                parent: span,
+                tracing::Level::INFO,
+                { ERROR_TYPE } = error_str,
+                { GCP_ERRORS_DOMAIN } = domain,
+                { RPC_RESPONSE_STATUS_CODE } = rpc_status_code,
+                { HTTP_RESPONSE_STATUS_CODE } = http_code,
+                "{err:?}"
```
westarle: this should be status message if available: https://github.com/haphungw/google-cloud-rust/blob/d04c1e8b92a4fb8af63df16cd258d9c3d87baf4a/src/gax/src/error/rpc.rs#L32-L35

westarle (changes requested):

src/gax-internal/src/observability/errors.rs
haphungw: Oh now I understand why you mentioned JSON yesterday.

src/gax-internal/src/observability/errors.rs
haphungw: done

src/gax-internal/src/observability/errors.rs
haphungw: done

src/gax-internal/src/observability/attributes.rs
```diff
@@ -85,10 +85,10 @@ pub mod keys {
     ///
     /// Example: pubsub.googleapis.com
     pub const GCP_ERRORS_DOMAIN: &str = "gcp.errors.domain";
-    /// The prefix for Google Cloud error metadata entries.
+    /// A key-value map representing the metadata embedded into the `ErrorInfo` payload.
     ///
-    /// Example: "gcp.errors.metadata.zone"
-    pub const GCP_ERRORS_METADATA_PREFIX: &str = "gcp.errors.metadata.";
+    /// It is flattened into a JSON string implicitly evaluating standard OTel semantics.
```
westarle: nit: add an example please

src/gax-internal/src/observability/errors.rs
```diff
@@ -86,6 +99,67 @@ impl ErrorType {
...
+                .map(|(k, v)| {
+                    format!(
+                        "\"{}\":\"{}\"",
+                        k.replace('"', "\\\""),
+                        v.replace('"', "\\\"")
+                    )
+                })
+                .collect();
+            format!("{{{}}}", fields.join(","))
+        });
```
westarle: Can we do something like:

```suggestion
let metadata_json = metadata.map(|m| {
    serde_json::to_string(&m).unwrap_or_else(|_| "{}".to_string())
});
```

This might be idiomatic (and faster.)

src/gax-internal/src/observability/errors.rs
```diff
@@ -86,6 +99,67 @@ impl ErrorType {
     }
 }
 
+pub(crate) fn emit_error_log(span: &tracing::Span, err: &Error) {
```
westarle, line 102: can you write a test?

haphungw: done

src/gax-internal/src/observability/attributes.rs
haphungw: done

src/gax-internal/src/observability/errors.rs
```diff
@@ -86,36 +99,88 @@ impl ErrorType {
...
+            None => tracing::event!(
+                parent: span,
+                tracing::Level::INFO,
+                { ERROR_TYPE } = error_str,
+                { GCP_ERRORS_DOMAIN } = domain,
+                { GCP_ERRORS_METADATA } = metadata_json,
+                { RPC_RESPONSE_STATUS_CODE } = rpc_status_code,
+                "{log_msg}"
+            ),
+        }
```
coryan: I recently learned you can simplify this:


```suggestion
    tracing::event!(
                parent: span,
                tracing::Level::INFO,
                { ERROR_TYPE } = error_str,
                { GCP_ERRORS_DOMAIN } = domain,
                { GCP_ERRORS_METADATA } = metadata_json,
                { RPC_RESPONSE_STATUS_CODE } = rpc_status_code,
                { HTTP_RESPONSE_STATUS_CODE } = http_code_status_code,
                "{log_msg}"
            );
```

Basically `Option<T>` implements `Value` if `T` does:

https://docs.rs/tracing/latest/tracing/field/trait.Value.html#impl-Value-for-Option%3CT%3E

And the implementation skips `None`.

haphungw: looks very neat. til.

src/gax-internal/src/observability/errors.rs
```diff
@@ -150,15 +204,110 @@ pub(crate) mod tests {
...
+                type Writer = Self;
+                fn make_writer(&'a self) -> Self::Writer {
+                    self.clone()
+                }
+            }
+
+            let output = Arc::new(Mutex::new(Vec::new()));
+            let subscriber = tracing_subscriber::fmt()
+                .with_writer(MockWriter(output.clone()))
+                .with_ansi(false)
```
westarle: Nit: you could use .json() and make more structured assertions about the contents.  I think this is OK for a future PR.

src/gax-internal/src/observability/errors.rs
```diff
@@ -86,6 +99,45 @@ impl ErrorType {
...
+
+        let error_str = error_type.as_str();
+        let log_msg = err
+            .status()
+            .map(|s| s.message.clone())
+            .unwrap_or_else(|| format!("{err:?}"));
+
+        tracing::event!(
+            parent: span,
+            tracing::Level::INFO,
```
westarle: I'm relatively confident this should be logged at DEBUG at this point.

src/gax-internal/src/observability/errors.rs
haphungw: I updated the assertions with json.

src/gax-internal/src/observability/errors.rs
haphungw: sounds good. i updated the logging level to DEBUG accordingly.

src/gax-internal/src/observability/errors.rs
```diff
@@ -150,15 +204,129 @@ pub(crate) mod tests {
...
+            );
+            assert_eq!(
+                fields["gcp.errors.domain"].as_str().expect("domain field"),
+                "test.googleapis.com"
+            );
+            assert_eq!(
+                fields["gcp.errors.metadata"]
+                    .as_str()
+                    .expect("metadata field"),
+                "{\"k1\":\"v1\",\"k2\":\"v2\"}"
```
westarle: I'm not sure if the (hash?)map -> json code is stable.  You might want to json decode this string and check each key to avoid flakes.

haphungw: Noted. I now parse the `metadata` string back to json before checking for the keys.

PR #5072, opened 2026-03-20
chore: update discovery and googleapis SHA circa 2026-03-20

JoeWang1127:


librarian.yaml
```diff
@@ -12,18 +12,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 language: rust
-version: v0.8.4-0.20260318163223-0ba84a3917a6
+version: v0.8.4-0.20260319233454-ec03ba51db7f
```
coryan: Please don't change the librarian version at the same time as the googleapis and/or discovery doc SHAs.  Revert this change and I can approve.

JoeWang1127: Do you want update the librarian version at a separate PR or don't want the update at all?

coryan: Separate PR.  Please always use a different PR for librarian changes from googleapis+discovery SHA changes. The rationale is:

* Reviewing PRs with only googleapis SHA and discovery SHA changes is easy, all the changes happen because some proto changed. And the PRs will ensure that nothing else is sneaking in. Basically a rubber stamp.
* Reviewing PRs with only librarian version changes is easy: if any code changes at all then we need to look at librarian's history and see if the changes are expected.
* Reviewering PRs with both is very hard.... I cannot tell if a change to a model file is an accidental change in librarian or an intentional change in a proto.  I need to look at each line, and even then it is very hard.


Could you change the playbooks to include this guideline and its rationale?

JoeWang1127: Updated the playbook to always creates separate PRs for librarian version updates and source updates.

PR #5073, opened 2026-03-20
docs(storage): add bucket encryption examples

joshuatants:


src/storage/examples/src/buckets/create_bucket_with_encryption_enforcement.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2025 Google LLC
```
coryan, line 1: nit: 2026

src/storage/examples/src/buckets/set_bucket_encryption_enforcement.rs
```diff
@@ -0,0 +1,57 @@
+// Copyright 2025 Google LLC
```
coryan, line 1: nit: 2026

joshuatants: Sorry about that, and thank you for fixing it for me:  I had enabled auto-merge.

PR #5075, opened 2026-03-20
test(storage-w1r3): add OpenTelemetry tracing

joshuatants:
Integrates OpenTelemetry tracing into the w1r3 storage benchmark using the `integration-tests-o11y` crate. When built with `--cfg google_cloud_unstable_tracing` and run with the `--project-id` flag, traces will be exported to Google Cloud Trace.

### DIFF
--- a/src/storage/benchmarks/w1r3/src/main.rs
+++ b/src/storage/benchmarks/w1r3/src/main.rs
@@ -39,6 +39,7 @@ use google_cloud_storage::retry_policy::RetryableErrors;
 use humantime::parse_duration;
 use instrumented_future::Instrumented;
 use instrumented_retry::DebugRetry;
+use opentelemetry_sdk::trace::{SdkTracerProvider, TraceError};
 use rand::{
     RngExt,
     distr::{Alphanumeric, Uniform},
@@ -48,10 +49,6 @@ use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{Duration, Instant};
 use tokio::sync::mpsc::Sender;
 use tracing::Instrument;
-
-#[cfg(google_cloud_unstable_tracing)]
-static TRACER_PROVIDER: std::sync::OnceLock<opentelemetry_sdk::trace::SdkTracerProvider> =
-    std::sync::OnceLock::new();
 
 const DEFAULT_TIMEOUT: Duration = Duration::from_secs(900);
 
@@ -64,11 +61,9 @@ async fn main() -> anyhow::Result<()> {
     if args.min_delete_batch > args.max_delete_batch {
         return Err(anyhow::Error::msg("invalid delete batch size range"));
     }
-    if args.reqwest_logs {
-        tracing_log::LogTracer::init()?;
-    }
+    tracing_log::LogTracer::init()?;
     let credentials = CredentialsBuilder::default().build()?;
-    enable_tracing(&args, &credentials).await?;
+    let tracer_provider = enable_tracing(&args, &credentials).await?;
     tracing::info!("{args:?}");
 
     let handle = tokio::runtime::Handle::current();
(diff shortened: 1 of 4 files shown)


src/storage/benchmarks/w1r3/src/main.rs
```diff
@@ -643,21 +663,48 @@ fn enable_tracing(args: &Args) {
     // formatter so that a delimiter is added between fields.
     .delimited("; ");
 
-    let subscriber = tracing_subscriber::fmt()
+    let level_filter = if args.reqwest_logs {
+        tracing_subscriber::filter::LevelFilter::TRACE
+    } else {
+        tracing_subscriber::filter::LevelFilter::INFO
+    };
```
coryan: If you are changing this code, maybe read about `EnvFilter`:

https://docs.rs/tracing-subscriber/latest/tracing_subscriber/filter/struct.EnvFilter.html

Now that I know how it works, I would use that instead.

src/storage/benchmarks/w1r3/src/main.rs
```diff
@@ -643,21 +663,48 @@ fn enable_tracing(args: &Args) {
...
+    let registry = tracing_subscriber::Registry::default().with(fmt_layer);
+
+    #[cfg(google_cloud_unstable_tracing)]
+    if let Some(project_id) = &args.project_id {
+        let tracer_provider =
+            integration_tests_o11y::otlp::trace::Builder::new(project_id, "storage-w1r3")
+                .with_credentials(_credentials.clone())
+                .build()
+                .await
+                .map_err(|e| anyhow::anyhow!("failed to create tracer provider: {e}"))?;
```
coryan: Consider:


```suggestion
                .await
                .inspect_err(|e| eprintln!("failed to create tracer provider: {e:?}"))?;
```

Which prints an error message but preserves the error types

src/storage/benchmarks/w1r3/src/main.rs
```diff
@@ -728,6 +775,14 @@ struct Args {
     /// Enable logs for the retry policies.
     #[arg(long)]
     debug_retry: bool,
+
+    /// The Google Cloud project ID.
```
coryan:
```suggestion
    /// A Google Cloud project ID used to send tracing data.
```

src/storage/benchmarks/w1r3/src/main.rs
```diff
@@ -47,6 +47,11 @@ use std::collections::BTreeMap;
 use std::sync::atomic::{AtomicU64, Ordering};
 use std::time::{Duration, Instant};
 use tokio::sync::mpsc::Sender;
+use tracing::Instrument;
+
+#[cfg(google_cloud_unstable_tracing)]
+static TRACER_PROVIDER: std::sync::OnceLock<opentelemetry_sdk::trace::SdkTracerProvider> =
+    std::sync::OnceLock::new();
```
coryan: I do not think you need this, you control the full program you can tell if this is initialized once.

coryan: After looking at the code again, yes, you should not be using a global variable.  Just return the provider (or an `Option<Provider>`) from `enable_tracing()`.

src/storage/benchmarks/w1r3/src/main.rs
joshuatants: Done

src/storage/benchmarks/w1r3/src/main.rs
joshuatants: Thanks for the tip, updated.

src/storage/benchmarks/w1r3/src/main.rs
joshuatants: Done.

src/storage/benchmarks/w1r3/src/main.rs
joshuatants: Awesome tip, thank you.

PR #5085, opened 2026-03-21
test(spanner): add integration test for write_at_least_once

olavloite:
Note: This pull request looks bigger than it 'actually' is, as it moves a lot of the integration test code from a single file into separate files.

Adds an integration test for write_at_least_once and refactors the integration test setup to follow the same pattern as the storage client. This cleans up the driver.rs file significantly, and makes it easier to add more integration tests in the future.

* The existing integration tests have all been moved to tests/spanner/src/query.rs
* The setup code for the emulator and test client have moved to tests/spanner/src/client.rs
* The new integration test for write_at_least_once has been added to tests/spanner/src/write.rs

This change also changes the `Delete` mutation to use the handwritten `KeySet` struct, and adds `is_null` and `try_is_null` to the `Row` struct to more easily check if a value is null.

### DIFF
--- a/src/spanner/src/row.rs
+++ b/src/spanner/src/row.rs
@@ -82,7 +82,7 @@ impl Row {
     /// # Example
     /// ```
     /// # use google_cloud_spanner::client::{Spanner, Statement};
-    /// # async fn test_doc() -> Result<(), Box<dyn std::error::Error>> {
+    /// # async fn test_doc() -> anyhow::Result<()> {
     /// let client = Spanner::builder().build().await?;
     /// let db_client = client.database_client("projects/p/instances/i/databases/d").build().await?;
     /// let transaction = db_client.single_use().build();
@@ -114,7 +114,7 @@ impl Row {
     /// # Example
     /// ```
     /// # use google_cloud_spanner::client::{Spanner, Statement};
-    /// # async fn test_doc() -> Result<(), Box<dyn std::error::Error>> {
+    /// # async fn test_doc() -> anyhow::Result<()> {
     /// let client = Spanner::builder().build().await?;
     /// let db_client = client.database_client("projects/p/instances/i/databases/d").build().await?;
     /// let transaction = db_client.single_use().build();
@@ -142,7 +142,7 @@ impl Row {
     /// # Example
     /// ```
     /// # use google_cloud_spanner::client::{Spanner, Statement};
-    /// # async fn test_doc() -> Result<(), Box<dyn std::error::Error>> {
+    /// # async fn test_doc() -> anyhow::Result<()> {
     /// let client = Spanner::builder().build().await?;
     /// let db_client = client.database_client("projects/p/instances/i/databases/d").build().await?;
     /// let transaction = db_client.single_use().build();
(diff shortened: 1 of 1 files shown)


src/spanner/src/row.rs
```diff
@@ -77,19 +77,79 @@ impl Row {
         &self.values
     }
 
+    /// Returns true if the value at the specified column name or index is null.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_spanner::client::{Spanner, Statement};
+    /// # async fn test_doc() -> Result<(), Box<dyn std::error::Error>> {
```
coryan: save yourself some typing:


```suggestion
    /// # async fn test_doc() -> anyhow::Result<()> {
```

`anyhow` is already a dev dependency and therefore accessible to doc tests.

src/spanner/src/row.rs
```diff
@@ -77,19 +77,79 @@ impl Row {
...
+    /// # Returns
+    ///
+    /// * `Ok(bool)` if the value is null or not.
+    /// * `Err(Error)` if the column name or index is invalid.
+    pub fn try_is_null<I: ColumnIndex>(&self, index: I) -> crate::Result<bool> {
+        let (_, value) = self.get_value(index)?;
+        Ok(value.kind() == crate::value::Kind::Null)
+    }
+
+    /// Returns true if the value at the specified column name or index is null, panicking on error.
```
coryan, line 112: FWIW, we typically implement just the non-panicking version. Applications can write:

```
  let y = foo.is_null(index).expect("the index exists");
  // or
  let x = foo.is_null(index).unwrap(); // YOLO
```

If they want the panic.

tests/spanner/Cargo.toml
```diff
@@ -25,17 +25,17 @@ license.workspace = true
...
-google-cloud-gax     = { workspace = true }
-google-cloud-lro     = { workspace = true }
-google-cloud-spanner = { workspace = true, features = [] }
-tokio.workspace      = true
-tracing.workspace    = true
-
-[dev-dependencies]
-reqwest    = { workspace = true, features = ["json"] }
-serde_json = { workspace = true }
+anyhow.workspace      = true
```
coryan, line 28: If you only intend to use `anyhow` in tests you should keep this in the `dev.dependencies` group.   That prevents accidental use outside the tests.

src/spanner/src/row.rs
olavloite: Having the 'plain' method panic and a separate `try_get` is an established practice for Rust database drivers. So I would prefer not to deviate from that. `is_null` (and `try_is_null`) are normally not implemented by other database drivers, but it is something that is heavily used by Spanner application developers in the other languages. And I do think that we should keep the semantics of these methods in Rust the same as the `get` / `try_get` methods.

src/spanner/src/row.rs
olavloite: Done

tests/spanner/Cargo.toml
olavloite: Note that this is the `Cargo.toml` file in `tests/spanner`, not in `src/spanner`.

My understanding is that now that a lot of the test logic has moved to `tests/spanner/src`, we need this in the `[dependencies]` section in order to be able to use it in the test logic, but that it will not leak into the actual library code, as it belongs to the `integration-tests-spanner`.

src/spanner/src/row.rs
coryan: Ack.

tests/spanner/Cargo.toml
coryan: Blegh... Somehow I convinced myself it was `src/spanner/Cargo.toml`.

PR #5103, opened 2026-03-23
test(spanner): provision the Spanner Emulator only once

### LINKED ISSUE #5096: Flake in spanner integration tests
The Rust CI failed on the `main` branch.

**Failure in Google Cloud Build:**
- **Check Suite:** https://api.github.com/repos/googleapis/google-cloud-rust/check-suites/61641661855
- **Commit:** 215edeb21c90f72e451afbe660bb0af986d950f0

olavloite:
The function for provisioning the Spanner Emulator with an instance and database should be executed only once.

The flake would happen if this part of the provisioning code would be executed during the `write_at_least_once` test:

https://github.com/olavloite/google-cloud-rust/blob/3d1daec70cb826dea45438e80f0dd336c28544f4/tests/spanner/src/client.rs#L144-149

```rust
    let write_tx = db_client.write_only_transaction().build();
    let mutation = Mutation::delete("AllTypes", KeySet::all());
    write_tx
        .write_at_least_once(vec![mutation])
        .await
        .expect("Failed to delete all data from AllTypes");
```

Fixes #5096

### DIFF
--- a/tests/spanner/Cargo.toml
+++ b/tests/spanner/Cargo.toml
@@ -34,7 +34,7 @@ google-cloud-spanner  = { workspace = true, features = [] }
 prost-types.workspace = true
 reqwest               = { workspace = true, features = ["json"] }
 serde_json            = { workspace = true }
-tokio.workspace       = true
+tokio                 = { workspace = true, features = ["sync"] }
 tracing.workspace     = true
 
 [lints]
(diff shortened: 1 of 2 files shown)


olavloite:
> I see, the process uses hard-coded ports. I see how this PR improves things but:
> 
> 1. It will fail if we have more than one build per VM / container
> 2. It will fail if I run two of these things on my laptop (unlikely)
> 3. It will fail if anything else is using those ports.
> 
> Is there any way to have the server use a random port?

I'll take a look at that. However, I think that if we want to support running two builds in parallel on the same VM/host, then we would also need to generate a random database name to use per VM to prevent those builds from interfering with each other.

PR #5108, opened 2026-03-23
ci(spanner): isolate integration tests using the Emulator

### LINKED ISSUE #5106: Spanner Emulator integration tests should use random port and random database
The integration tests for Spanner that use the Spanner Emulator should:
1. Start the Emulator on a random port to prevent port clashes.
2. Use a random database name for each test run, to prevent multiple parallel builds from interfering with each other. The use of a random database name can be overridden by setting an environment variable with a fixed database name. This can be used to debug integration test issues that need inspection of the database after the tests have run.

olavloite:
Isolates integration tests for Spanner that use the Spanner Emulator further by:
1. Not mapping the ports of the Emulator Docker container to localhost, but rather connecting directly to the dynamically assigned IP address of the container.
2. Generating a random database name for each test run. Tests can also use a fixed database name by setting an environment variable, for example for local testing where the user wants to manually inspect the database after a test run.

Fixes #5106

### DIFF
--- a/tests/spanner/src/client.rs
+++ b/tests/spanner/src/client.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use google_cloud_spanner::client::{KeySet, Mutation, Spanner};
+use google_cloud_test_utils::resource_names::LowercaseAlphanumeric;
 
 const PROJECT_ID: &str = "test-project";
 const INSTANCE_ID: &str = "test-instance";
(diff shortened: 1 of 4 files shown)


tests/spanner/src/client.rs
```diff
@@ -38,6 +37,17 @@ pub async fn wait_for_emulator(endpoint: &str) {
...
+
+async fn get_database_id() -> &'static str {
+    DATABASE_ID
+        .get_or_init(|| async {
+            std::env::var("SPANNER_EMULATOR_TEST_DB").unwrap_or_else(|_| {
+                format!("db-{}", &uuid::Uuid::new_v4().simple().to_string()[..20])
+            })
+        })
+        .await
+}
```
coryan, line 50: We have helpers to create unique ids which are (maybe) easier to use... Something like:

https://github.com/googleapis/google-cloud-rust/blob/56942569eee3be0c5854d9c233632b29029642ae/src/test-utils/src/resource_names.rs#L74-L78

You don't need to put the code in that file if you don't want to.

olavloite: Saves a dependency, so SGTM. Done.

PR #5111, opened 2026-03-23
feat(spanner): implement precommit token protocol for r/w transactions

olavloite:
Implements the pre-commit token protocol for read/write transactions.


src/spanner/src/precommit.rs
```diff
@@ -0,0 +1,158 @@
...
+    fn into_model(self) -> crate::model::MultiplexedSessionPrecommitToken {
+        crate::model::MultiplexedSessionPrecommitToken::new()
+            .set_precommit_token(self.precommit_token)
+            .set_seq_num(self.seq_num)
+    }
+}
+
+#[derive(Clone, Debug, Default)]
+pub(crate) enum PrecommitTokenTracker {
+    #[default]
```
dbolduc: It is idiomatic for `default()` to be the same as `new()`, but this is the same as `new_noop()`.

Can we get rid of the `impl Default` ?

src/spanner/src/result_set.rs
```diff
@@ -590,4 +597,38 @@ pub(crate) mod tests {
...
     }
+
+    #[tokio::test]
+    async fn test_result_set_precommit_token_tracked() {
+        let mut rs = run_mock_query(vec![PartialResultSet {
+            metadata: metadata(1),
+            values: vec![],
+            chunked_value: false,
+            resume_token: vec![],
+            stats: None,
```
dbolduc: nit: these are all default values. Consider omitting them, unless you want to draw attention to their value:

```rs
PartialResultSet {
  metadata: metadata(1),
  precommit_token: Some(...),
  last: true,
  ..Default::default(),
}
```

src/spanner/src/result_set.rs
```diff
@@ -590,4 +597,38 @@ pub(crate) mod tests {
...
+            last: true,
+            cache_update: None,
+        }])
+        .await;
+
+        // Force tracking mode since run_mock_query uses a ReadOnly transaction (NoOp).
+        rs.precommit_token_tracker = PrecommitTokenTracker::new();
+
+        // Read a row to trigger precommit token extraction
+        assert!(rs.next().await.is_none());
```
dbolduc: nit: more informative errors

```suggestion
        let row = rs.next().await;
        assert!(row.is_none(), "{row:?}");
```

src/spanner/src/read_write_transaction.rs
```diff
@@ -141,9 +149,11 @@ impl ReadWriteTransaction {
     /// Commits the transaction.
     pub(crate) async fn commit(self) -> crate::Result<wkt::Timestamp> {
         let transaction_id = self.transaction_id()?;
+        let precommit_token = self.context.precommit_token_tracker.get();
         let request = CommitRequest::default()
             .set_session(self.context.client.session.name.clone())
-            .set_transaction_id(transaction_id);
+            .set_transaction_id(transaction_id.clone())
```
dbolduc, line 155: Aside: we should change the signature of `transaction_id` to avoid converting between `bytes::Bytes` and `Vec<u8>` unnecessarily.

```diff
-    fn transaction_id(&self) -> crate::Result<Vec<u8>> {
+    fn transaction_id(&self) -> crate::Result<bytes::Bytes> {
         match &self.context.transaction_selector.selector {
-            Some(Selector::Id(id)) => Ok(id.to_vec()),
+            Some(Selector::Id(id)) => Ok(id.clone()),
             _ => Err(crate::error::internal_error("Transaction ID is missing")),
         }
     }
```

and note that `bytes::Bytes` is cheap to clone, and `Vec<u8>` is not.

src/spanner/src/precommit.rs
```diff
@@ -0,0 +1,158 @@
...
+        assert_eq!(retrieved.precommit_token, "token2");
+        assert_eq!(retrieved.seq_num, 2);
+    }
+
+    #[test]
+    fn test_tracker_default_derivation() {
+        // Enforce that default returns the NoOp variant properly
+        let tracker = PrecommitTokenTracker::default();
+        assert!(tracker.get().is_none());
+    }
```
dbolduc: ... `PrecommitTokenTracker::default()` is only used in a test. And I assume this is an internal-only type. So remove?

src/spanner/src/precommit.rs
```diff
@@ -0,0 +1,158 @@
...
+
+pub(crate) trait AsPrecommitToken {
+    fn into_model(self) -> crate::model::MultiplexedSessionPrecommitToken;
+}
+
+impl AsPrecommitToken for crate::model::MultiplexedSessionPrecommitToken {
+    fn into_model(self) -> crate::model::MultiplexedSessionPrecommitToken {
+        self
+    }
+}
```
dbolduc: I feel like this can only make the code more confusing. We always know whether an object is from our model, or a prost/protobuf message.

src/spanner/src/precommit.rs
```diff
@@ -0,0 +1,158 @@
...
+
+impl AsPrecommitToken for crate::google::spanner::v1::MultiplexedSessionPrecommitToken {
+    fn into_model(self) -> crate::model::MultiplexedSessionPrecommitToken {
+        crate::model::MultiplexedSessionPrecommitToken::new()
+            .set_precommit_token(self.precommit_token)
+            .set_seq_num(self.seq_num)
+    }
+}
```
dbolduc: There is a generated conversion for this:

https://github.com/olavloite/google-cloud-rust/blob/fc4b41a10522945f89b7d1d54a9a924313a6bea0/src/spanner/src/generated/convert/spanner/convert.rs#L2354-L2362

Should be able to say `pb_message.to_proto().expect("simple conversion never fails")` or something.

src/spanner/src/precommit.rs
```diff
@@ -0,0 +1,158 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::sync::{Arc, RwLock};
+
+pub(crate) trait AsPrecommitToken {
```
dbolduc: I would vote against this trait.

src/spanner/src/read_write_transaction.rs
olavloite: Done (the change is minimal, so I included it in this PR)

src/spanner/src/precommit.rs
olavloite: Done

src/spanner/src/result_set.rs

src/spanner/src/result_set.rs
olavloite: Done (here and elsewhere)

src/spanner/src/precommit.rs
olavloite: Removed

src/spanner/src/precommit.rs
olavloite: Removed (along with the entire trait)

src/spanner/src/precommit.rs
olavloite: yeah, that makes the conversion even simpler, and the trait fully superfluous.

src/spanner/src/precommit.rs
olavloite: Removed

src/spanner/src/precommit.rs
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    /// Creates a no-op tracker for read-only transactions.
+    pub(crate) fn new_noop() -> Self {
+        Self::NoOp
+    }
+
+    /// Updates the tracker with an optional precommit token from a response.
+    pub(crate) fn update(&self, token: Option<crate::model::MultiplexedSessionPrecommitToken>) {
+        let Some(token) = token else { return };
```
dbolduc: nit: `let token = token?;`

src/spanner/src/precommit.rs
```diff
@@ -0,0 +1,121 @@
...
+        let Self::Track(tracker) = self else { return };
+
+        let new_token = token;
```
dbolduc: nit: this is a noop. Can just use `token` below

src/spanner/src/precommit.rs
olavloite: That would not be compatible with the current return type of this function, which is `()` (and not `Option<()>`). But we can rewrite it to this:

```rust
        if let (Some(token), Self::Track(tracker)) = (token, self) {
            let mut guard = tracker.write().unwrap();
            if guard.as_ref().is_none_or(|c| c.seq_num < token.seq_num) {
                *guard = Some(token);
            }
        }
```

src/spanner/src/precommit.rs
olavloite: see above

src/spanner/src/precommit.rs
dbolduc: d'oh. You're right. Good call.

PR #5113, opened 2026-03-23
test(o11y): add integration test for t4 span debug log

haphungw:
Add integration testing to guarantee that detailed error logs correctly propagate telemetry data at the DEBUG (L4) tracing level.

Logs are picked up by `opentelemetry_appender_tracing`, separated from trace span.

### DIFF
--- a/tests/o11y/src/http_tracing.rs
+++ b/tests/o11y/src/http_tracing.rs
@@ -246,7 +246,6 @@ pub async fn to_otlp_debug_event() -> anyhow::Result<()> {
         log_event.trace_id, attempt_span.trace_id,
         "Log traceId correlation failed"
     );
-    // Span ID might be a deep nested span, but the traceId must correlate perfectly.
     assert_eq!(
         log_event.span_id, attempt_span.span_id,
         "Log spanId correlation failed"


tests/o11y/src/http_tracing.rs
```diff
@@ -132,6 +132,141 @@ pub async fn to_otlp() -> anyhow::Result<()> {
...
+        .expect("should have scope spans");
+    let spans = &scope_span.spans;
+
+    let attempt_span = spans
+        .iter()
+        .find(|s| s.kind == 3 /* CLIENT */)
+        .expect("Should have found a SPAN_KIND_CLIENT span");
+
+    let error_event = attempt_span
+        .events
```
westarle: we should try to pick up logs with https://docs.rs/opentelemetry-appender-tracing/latest/opentelemetry_appender_tracing/

Then inspect the logs sent to the OTLP endpoint to make sure the injected traceId and parent span are correct.

I'd prefer if logs are not also picked up as trace events, it might mean configuring a filter on the opentelemetry_tracing subscriber.  Not a big deal if you can't manage that as part of this PR.

tests/o11y/src/http_tracing.rs
```diff
@@ -132,6 +133,163 @@ pub async fn to_otlp() -> anyhow::Result<()> {
...
+        .find(|l| l.attributes.iter().any(|kv| kv.key == "error.type"))
+        .expect("Should have logged an ErrorInfo log natively mapped conditionally");
+
+    // Ensure the log was correctly correlated back to the original client request trace buffer (attempt span)
+    assert_eq!(
+        log_event.trace_id, attempt_span.trace_id,
+        "Log traceId correlation failed"
+    );
+    // Span ID might be a deep nested span, but the traceId must correlate perfectly!
+    // We omit asserting exact span_id equal to attempt_span because the macro could be inside an inner retry span.
```
westarle: the span of the log event should be the CLIENT span (these should be the lowest level of retry.)  what are you seeing?

haphungw: yes you're right, i missed that and the http attempt span was not entered at all. that is why the `span_id` of the outer `client_request` span was assigned to the log event instead of the `http_attempt` span.

haphungw: I've updated the code to use `span.in_scope(|| ...)` so that the L4 span is pushed back onto the context stack just long enough to emit the log. This guarantees that the error log is correctly correlated with the exact transport attempt that failed.

tests/o11y/src/http_tracing.rs
```diff
@@ -132,6 +133,166 @@ pub async fn to_otlp() -> anyhow::Result<()> {
...
+        .flat_map(|sl| sl.log_records)
+        .find(|l| l.attributes.iter().any(|kv| kv.key == "error.type"))
+        .expect("Should have logged an ErrorInfo log natively mapped conditionally");
+
+    // Ensure the log was correctly correlated back to the original client request trace buffer (attempt span)
+    assert_eq!(
+        log_event.trace_id, attempt_span.trace_id,
+        "Log traceId correlation failed"
+    );
+    // Span ID might be a deep nested span, but the traceId must correlate perfectly.
```
westarle: nit: I don't think this comment makes sense anymore.  Can this and trace go in the map below?

haphungw: thanks for catching this. i removed the outdated comment.

Regarding your other question: we can't put `trace_id` and `span_id` into the map because they aren't part of the attributes list. They are standalone fields separated from custom attributes in OTel, so we have to assert them separately.

PR #5116, opened 2026-03-23
feat(o11y): adapt l3 actionable error logs to WARN severity and OTel attributes

haphungw:
Align with cross language design for actionable errors.
Decrease `tracing::event!` from `ERROR` to `WARN` to prevent terminal SDK failures from triggering false-positive production alerts.
Add new `exception.type` and `exception.message` into the tracing block.

### DIFF
--- a/tests/o11y/src/e2e/signals.rs
+++ b/tests/o11y/src/e2e/signals.rs
@@ -205,6 +205,7 @@ fn check_logs(project_id: &str, buffer: Buffer, trace_id: TraceId) -> anyhow::Re
         "rpc.system.name": "http",
         "url.domain": "localhost:7469", // the showcase domain...
         "url.template": "", // TODO(#...)
+        "exception.type": "404",
     });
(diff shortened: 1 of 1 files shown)


tests/o11y/src/e2e/signals.rs
```diff
@@ -206,7 +206,25 @@ fn check_logs(project_id: &str, buffer: Buffer, trace_id: TraceId) -> anyhow::Re
...
-    assert_eq!(Some(&fields), want.as_object(), "{value:?}");
+    for (key, expected) in want.as_object().unwrap() {
+        assert_eq!(
+            fields.get(key),
+            Some(expected),
+            "mismatch for {key:?} in {value:?}"
+        );
+    }
+    assert_eq!(
+        fields.get("exception.type").unwrap().as_str().unwrap(),
```
westarle: nit: can you add this to `expected` above?

haphungw: thanks, i missed that. fixed.

PR #5120, opened 2026-03-24
feat(o11y): add actionable log suppression option

haphungw:
Set up the `disable_actionable_error_logging` to prevent actionable error logs from polluting the pipeline.

Depends on #5113
Must be merged after #5113 and keep the `log_layer` logic.

### DIFF
--- a/src/gax-internal/src/observability/client_signals.rs
+++ b/src/gax-internal/src/observability/client_signals.rs
@@ -22,6 +22,13 @@ pub use duration_metric::DurationMetric;
 pub use request_start::RequestStart;
 pub use with_client_signals::WithClientSignals;
 
+/// An extension to disable terminal actionable error logging.
+///
+/// If this extension is present in the `RequestOptions` supplied to a GAX call,
+/// the terminal application logs will be suppressed.
+#[derive(Clone, Copy, Debug)]
+pub struct SuppressActionableErrorLog;
+
 /// Creates a [Span] and [RequestStart] for a client request.
 ///
 /// # Parameters
--- a/src/gax-internal/src/observability/client_signals/with_client_signals.rs
+++ b/src/gax-internal/src/observability/client_signals/with_client_signals.rs
@@ -135,7 +135,7 @@ where
                 let err_msg = error.to_string();
 
                 // TODO(#4795) - use the correct name and target
-                if !this.start.info().disable_actionable_error_logging {
+                if !this.start.disable_actionable_error_logging() {
                     tracing::event!(
                         name: NAME,
                         target: TARGET,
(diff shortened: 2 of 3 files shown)


src/gax-internal/src/observability/client_signals/with_client_signals.rs
```diff
@@ -135,20 +135,22 @@ where
...
-                    { URL_DOMAIN } = this.start.info().default_host,
-                    { URL_TEMPLATE } = this.start.url_template(),
-                    { RPC_METHOD } = this.start.method(),
-                    { RPC_RESPONSE_STATUS_CODE } = rpc_status_code,
-                    { HTTP_RESPONSE_STATUS_CODE } = error.http_status_code(),
-                    { EXCEPTION_TYPE } = error_str,
-                    { EXCEPTION_MESSAGE } = err_msg,
-                    "{error:?}"
-                );
+                if !this.start.info().disable_actionable_error_logging {
```
coryan: This is fine to merge but be aware that I am thinking about some changes in these helpers.  Instead of a single `WithClientSignals` helper I think we should have a `WithLogging`, `WithMetric` and `WithClientSpan` helpers where they will be initialized as:

```
let (recorder, span) = client_request_span!(&INSTRUMENTATION_INFO);
let inner: Future<_> = client.get_foo();
let wrapped: WithSpan<WithMetric<WithLogging<_>>> =
    inner.with_logging().with_metric(self.metric.clone()).with_span(span);
let result = recorder.scope(wrapped).await; // Run the wrapped future in the context of `recorder`.
```

If that works, what we would do is skip the `WithLogging<>` step when disabled, instead of saving that state in `start()`.

src/gax-internal/src/options.rs
```diff
@@ -31,6 +31,8 @@ pub struct InstrumentationClientInfo {
     pub client_artifact: &'static str,
     /// The default hostname of the service.
     pub default_host: &'static str,
+    /// Whether to disable emission of actionable error logs.
+    pub disable_actionable_error_logging: bool,
```
coryan: So far this type is static and unchanged in each request.  In principle we could have a single `&'static` instance in all the code.  Now we need to change it in each request.

I am not sure it is a big deal.  It does create two ways to carry options, which seems like a bad thing.

src/gax/src/client_builder.rs
```diff
@@ -475,6 +477,19 @@ pub mod internal {
         builder.config.disable_follow_redirects = !v;
         builder
     }
+
+    /// Configure actionable error logging.
+    ///
+    /// By default, the client libraries emit actionable error logs for top-level
+    /// failures. Internal users can disable this behavior if they are wrapping
+    /// this client and want to emit their own actionable errors.
+    pub fn with_actionable_error_logging<F, Cr>(
```
coryan: If we start using this in gaxi or anywhere really, this becomes part of the public API of the library, and if we remove it, we have to take a breaking change.

It also seems strange, because the configuration of the client (which this struct is supposed to set) does not change.

And finally, the previous two functions here (`with_follow_redirects()` and `with_automatic_decompression()` ) are not used anywhere (and yet we cannot remove them).  I would rather not add a function that we may not use after all.

Can we guard with with the `#[cfg(google_cloud_unstable_tracing)]` thing **or** maybe use the request option extensions to pass through this configuration?

src/gax-internal/src/observability/client_signals/with_client_signals.rs
haphungw: Good sugggestion. I'm thinking to save the refactoring for another PR.

src/gax/src/client_builder.rs
haphungw: I've dropped the changes to `ClientBuilder` and now use request option extension.

src/gax-internal/src/observability/client_signals.rs
```diff
@@ -22,6 +22,13 @@ pub use duration_metric::DurationMetric;
 pub use request_start::RequestStart;
 pub use with_client_signals::WithClientSignals;
 
+/// An extension to disable terminal actionable error logging.
+///
+/// If this extension is present in the `RequestOptions` supplied to a GAX call,
+/// the terminal application logs will be suppressed.
+#[derive(Clone, Copy, Debug)]
+pub struct SuppressActionableErrorLog;
```
coryan, line 30: Note to self: this is safe to add because the whole module is protected by `#[cfg(google_cloud_unstable_tracing)]`

Issue #5024, opened 2026-03-16
Applications can signal and await subscriber shutdown


dbolduc:
Related to #4869, #4868 

Those are more concerned with the shutdown behavior. This issue is about how the application can signal and await a shutdown.

## Background

Messages can get "frozen" for 10 minutes if a program exits before the subscriber can flush pending nacks. And there may be unnecessary redeliveries if the subscriber does not flush pending acks. Letting an application await shutdown can avoid/mitigate these problems.

It is also somewhat difficult for an application to signal a shutdown. They have to `drop(stream)`, which means they need to own the stream. It is more convenient if applications can spawn a background task, and signal shutdown from it.

## Design

We should add a `ShutdownToken`, with an accessor for it on the `MessageStream`:

```rs
#[derive(Clone)]
pub struct ShutdownToken {
  inner: CancellationToken,
  fut: Shared<BoxFuture<'static, ()>>,
}

impl ShutdownToken {
  // I think it is convenient to shutdown and await.
  pub async fn shutdown(&self) {
    self.inner.cancel();
    self.fut.await
  }

  // We could also add this to just signal. But initially, I would not add this.
  pub fn signal_shutdown(&self) {
    self.inner.cancel();
  }
}

impl MessageStream {
  pub fn shutdown_token(&self) -> ShutdownToken {
    self.shutdown.clone()
  }
}
```

## Work

- [x] Hold lease loop as shared future
- [x] Split `MessageStream`
- [x] Add `CancellationToken` to `MessageStream`
- [x] Introduce `ShutdownToken`
- [x] Add `shutdown_token()` accessor
- [x] Support shutdown outside of the stream by holding weak senders for channels
- [x] Errors initiate a shutdown
- [x] Initiate a shutdown on `drop(stream)`
- [x] Make all this stuff public
- [x] Simplify any examples, integration tests.
- [x] Remove `MessageStream::close()`

dbolduc:
I think the current design I have in my head is missing something. It allows for applications to write code like this:

```rs
let mut stream = ...;
let shutdown_token = stream.shutdown_token();

while let Some(_) = stream.next().await {
  // use the message...
  if (done) {
    shutdown_token.cancel();
    break;
  }
}

// Gah! This never completes. Application needs to `drop(stream)` or call `stream.next().await` again.
shutdown_token.shutdown().await;
```

[...]

PR #5123, opened 2026-03-24
impl(pubsub): stream can be cancelled

dbolduc:
Part of the work for #5024 

Add a cancellation token to the `MessageStream`. If a shutdown is signaled, the next streamed item will be `None`.

A follow up PR will expose a new `ShutdownToken` type so the application can signal and await a shutdown.

### DIFF
--- a/src/pubsub/src/subscriber/message_stream.rs
+++ b/src/pubsub/src/subscriber/message_stream.rs
@@ -111,8 +111,9 @@ pub struct MessageStreamImpl {
 #[allow(clippy::large_enum_variant)]
 #[derive(Debug)]
 enum StreamState {
-    /// The stream failed with a permanent error. It should not be re-opened.
-    Failed,
+    /// The stream was cancelled or failed with a permanent error. It should not
+    /// be re-opened.
+    Closed,
     /// The stream is active.
     Active(Stream<Transport>),
 }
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -1452,4 +1469,84 @@ mod tests {
...
+        assert!(end.is_none(), "Shutdown should end the stream, got {end:?}");
+
+        Ok(())
+    }
+
+    #[tokio_test_no_panics(start_paused = true)]
+    async fn cancel_midstream() -> anyhow::Result<()> {
+        let (response_tx, response_rx) = channel(10);
+        let (ack_tx, mut ack_rx) = unbounded_channel();
+        let (nack_tx, mut nack_rx) = unbounded_channel();
```
dbolduc, line 1553: FYI for the reviewer: Now that we have the `tokio_unstable` build, we don't need these channels. We could set our assertions from within the mock, and the `tokio_unstable` build would catch a background panic.

I decided to keep using the channels so that a simple `cargo test` would still catch errors.

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -185,7 +191,18 @@ impl MessageStream {
...
-        self.inner.next().await
+        let next = tokio::select! {
+            biased;
+            _ = self.shutdown.cancelled() => None,
+            n = self.inner.next() => n,
+        };
+        if next.is_none() {
+            // Permanently close the stream, and drop messages that we haven't
+            // delivered to the application yet.
+            self.inner.stream = Some(StreamState::Failed);
```
suzmue: nit: Failed seems a little odd for the state.

suzmue (approved):

src/pubsub/src/subscriber/message_stream.rs
dbolduc: You're right. Renamed that state: `Closed`.

PR #5124, opened 2026-03-24
WIP - impl(gax-internal): recorder and logging

coryan:
Introduce `RequestRecorder` to capture the details of a request through multiple layers. Show how this is used in a logging wrapper for futures.

Part of the work for  #4772

### DIFF
--- a/src/gax-internal/src/observability.rs
+++ b/src/gax-internal/src/observability.rs
@@ -42,7 +42,9 @@ pub mod grpc_tracing;
 mod client_signals;
 
 #[cfg(google_cloud_unstable_tracing)]
-pub use client_signals::{ClientSignalsExt, DurationMetric, RequestRecorder, RequestStart};
+pub use client_signals::{
+    ClientSignalsExt, DurationMetric, RequestRecorder, RequestStart, WithLogging,
+};
 
 #[cfg(google_cloud_unstable_tracing)]
 mod client_tracing;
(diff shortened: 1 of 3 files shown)


src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -0,0 +1,337 @@
...
+    ///     let Err(e) = result else { return; };
+    ///     let Some(recorder) = RequestRecorder::current() else { return; };
+    ///     let snapshot = recorder.t3_snapshot();
+    ///     tracing::error!(
+    ///         { "url.domain" } = snapshot.default_host(),
+    ///         // use more things from snapshot here.
+    ///     );
+    /// }
+    /// ```
+    pub fn t3_snapshot(&self) -> T3Snapshot {
```
westarle: I think we need to ~bikeshed~ choose a name for these that outlives the current program:

Maybe one of:
* client -> transport
* request -> attempt

coryan: Yes, and I like ~the color of that bikeshed~ those names.

coryan: Yes, and I like ~the color of that bikeshed~ those names.

westarle: Oops just so it's clear there were two options there:

1. t3 = client & t4 = transport
2. t3 = request & t4 = attempt

My preference is #1.

PR #5125, opened 2026-03-24
impl(pubsub): add shutdown token

dbolduc:
Part of the work for #5024 

Add a type that can signal and await shutdown.

I think applications will typically signal and await together, but we separate the functions to allow for graceful shutdown without signaling shutdown (e.g. from an external error):

```rs
let shutdown_token = stream.shutdown_token();
while let Some((m, h)) = stream.next().await { ... }

// There must have been an external error. Wait for shutdown before returning.
shutdown_token.shutdown().await;
```

### DIFF
--- a/src/pubsub/src/subscriber/shutdown_token.rs
+++ b/src/pubsub/src/subscriber/shutdown_token.rs
@@ -1,0 +1,120 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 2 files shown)


src/pubsub/src/subscriber/shutdown_token.rs
```diff
@@ -0,0 +1,120 @@
...
+///
+///   // Signal a shutdown of the stream.
+///   token.cancel();
+///
+///   // Await a shutdown of the stream.
+///   token.shutdown().await;
+/// }
+/// ```
+#[derive(Clone, Debug)]
+pub struct ShutdownToken {
```
dbolduc, line 35: FYI: this will be a public type.

src/pubsub/src/subscriber/shutdown_token.rs
```diff
@@ -0,0 +1,120 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use futures::future::{BoxFuture, Shared};
+use tokio_util::sync::CancellationToken;
+
+/// A token to signal and await shutdown of a stream.
+///
+/// # Example
+/// ```no_rust
```
suzmue: Can you pair this with a TODO so we don't forget to make this sample runnable?

dbolduc:
@suzmue - I am merging this. Nothing is public yet.

PR #5126, opened 2026-03-24
test(spanner): integration tests for Read operations

olavloite:
Adds integration tests for Read operations.

### DIFF
--- a/tests/spanner/src/read.rs
+++ b/tests/spanner/src/read.rs
@@ -14,17 +14,15 @@
 
 use google_cloud_spanner::client::{DatabaseClient, KeyRange, KeySet, Mutation, ReadRequest};
 use google_cloud_spanner::key;
+use google_cloud_test_utils::resource_names::LowercaseAlphanumeric;
 
 pub async fn read_single_key(db_client: &DatabaseClient) -> anyhow::Result<()> {
-    let id1 = format!(
-        "read-single-{}",
-        google_cloud_test_utils::resource_names::LowercaseAlphanumeric.random_string(10)
-    );
+    let id1 = format!("read-single-{}", LowercaseAlphanumeric.random_string(10));
     let mutation = Mutation::new_insert_or_update_builder("AllTypes")
         .set("Id")
         .to(&id1)
         .set("ColString")
-        .to(&"single".to_string())
+        .to(&"single")
         .build();
     let write_tx = db_client.write_only_transaction().build();
     write_tx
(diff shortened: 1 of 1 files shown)


tests/spanner/src/read.rs
```diff
@@ -0,0 +1,298 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use google_cloud_spanner::client::{DatabaseClient, KeyRange, KeySet, Mutation, ReadRequest};
+use google_cloud_spanner::key;
+
+pub async fn read_single_key(db_client: &DatabaseClient) -> anyhow::Result<()> {
+    let id1 = format!(
+        "read-single-{}",
+        google_cloud_test_utils::resource_names::LowercaseAlphanumeric.random_string(10)
```
dbolduc: style nit: `use google_cloud_test_utils::resource_names::LowercaseAlphanumeric;` at the top might be easier on the eyes.

tests/spanner/src/read.rs
```diff
@@ -0,0 +1,298 @@
...
+    );
+    let mutation = Mutation::new_insert_or_update_builder("AllTypes")
+        .set("Id")
+        .to(&id1)
+        .set("ColString")
+        .to(&"single".to_string())
```
dbolduc: nit: I assume just `&"single"` works?

There is an `impl ToValue for &str`

dbolduc: here and below.

tests/spanner/src/read.rs
```diff
@@ -0,0 +1,298 @@
...
+        .build();
+    let mut result_set = db_client
+        .single_use()
+        .build()
+        .execute_read(read)
+        .await
+        .expect("Failed to execute read all keys");
+    let mut rows = Vec::new();
+    while let Some(row) = result_set
+        .next()
```
dbolduc, line 90: FYI, not for this PR: I think we will want a way to turn the `ResultSet` into a `futures::Stream`. It integrates with the Rust ecosystem well, and lets applications write slick code.

---

Background:

We have a few types that look like stream. i.e. they have an API like `async fn next(&mut self) -> Option<Result<T>>`. We offer a way to turn them into a `futures::Stream`, e.g. https://docs.rs/google-cloud-storage/latest/google_cloud_storage/read_object/struct.ReadObjectResponse.html#method.into_stream

---

Then if we have that, we can maybe write this expectation like the following:

```rs
use futures::TryStreamExt;
let rows = result_set.into_stream().try_filter(|row| {
  let id = row.get::<String, _>("Id");
  std::future::ready(id == id1 || id == id2)
})?;
assert_eq!(rows.len(), 2);
```

^ (untested)

tests/spanner/src/read.rs
```diff
@@ -0,0 +1,298 @@
...
+
+    let write_tx = db_client.write_only_transaction().build();
+    write_tx
+        .write_at_least_once(vec![mutation_1, mutation_2, mutation_3])
+        .await
+        .expect("Failed to write to AllTypes");
+
+    // Read range from id1 (inclusive) to id3 (exclusive) -> should return id1 and id2
+    let keyset = KeySet::builder()
+        .add_range(KeyRange::closed_open(key![id1.clone()], key![id3.clone()]))
```
dbolduc, line 141: comment: I think `key!` is our first public macro. I see that it is a convenient shorthand for key builders.

tests/spanner/src/read.rs
olavloite: Ack, I'll take a look if something like that is doable.

tests/spanner/src/read.rs
olavloite: Done

tests/spanner/src/read.rs
olavloite: Done (here and elsewhere)

Issue #5048, opened 2026-03-17
Improve subscriber lease extension logic


dbolduc:
The current implementation is to extend all leases every N seconds. It is an ok strategy. The customers aren't charged for these RPCs. It could be a lot better though, and has potential to clog the lease event loop before we have #3975.

We should:
- keep in the LeaseInfo the last time we attempted to extend a lease. (or `None`, for new messages).
- only extend leases for messages where 80% of the current lease has been used.

This looks like modifying the `Leases::retain` functions. And probably updating `MessageStream` / `LeaseState` tests that assume a certain behavior.

dbolduc:
I did the at-least-once half. The exactly-once half remains.

We could be more clever, and record when the last lease extension was successful, instead of just when it was attempted. But that probably won't make any difference, given that exactly-once lease extensions also have retries. I would probably copy the at-least-once logic.

Unassigning myself because I am not actively working on this.

PR #5127, opened 2026-03-24
perf(pubsub): concurrent lease extensions

dbolduc:
Part of the work for #3975 

Send lease extension RPCs concurrently. They no longer block the lease loop.

## Is it safe?

RPCs can start racing when we send things concurrently. e.g. a nack RPC can race with an extend RPC.

Note that this is fine. If the two RPCs race, the nack always wins (which is what we want when we send a nack).

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -237,6 +237,9 @@ where
             self.pending_extends
                 .spawn(async move { leaser.extend(ack_ids).await });
         }
+
+        // TODO(#5048) - we could process the results as a lease event.
+        while self.pending_extends.try_join_next().is_some() {}
     }
 
     /// Shutdown the leaser
@@ -735,6 +738,34 @@ pub(super) mod tests {
             state.process(ExactlyOnceNack(test_id(i)));
         }
         extend_and_await(&mut state).await;
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn pending_extends_size_management() {
+        let mut mock = MockLeaser::new();
+        mock.expect_extend()
+            .times(2)
+            .withf(|v| *v == vec![test_id(1)])
+            .returning(|_| ());
+
+        let mut state = LeaseState::new(Arc::new(mock), LeaseOptions::default());
+
+        state.add(test_id(1), at_least_once_info());
+        state.extend();
+        // Yield execution so the extend attempt can execute.
+        tokio::task::yield_now().await;
+
+        // TODO(#5048) - We currently clean up the completed pending extends in
+        // `LeaseState::extend()`. If we decide to clean up the pending extends
+        // elsewhere, this test will need an update.
+        state.extend();
+        let pending_extends = state.pending_extends.len();
+        assert!(
+            pending_extends < 2,
+            "The first lease extension attempt should have completed. We should not hold onto it."
+        );
+
+        let _ = state.pending_extends.join_all().await;
     }
 
     #[tokio::test]


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -215,16 +223,19 @@ where
...
     ///
     /// Drops messages whose lease deadline cannot be extended any further.
-    pub(super) async fn extend(&mut self) {
-        // TODO(#3975) - send RPCs concurrently.
+    pub(super) fn extend(&mut self) {
         let batches = self.leases.retain(self.max_lease);
         for ack_ids in batches {
-            self.leaser.extend(ack_ids).await;
+            let leaser = self.leaser.clone();
+            self.pending_extends
```
suzmue, line 230: This is going to grow indefinitely until the subscriber shuts down, can we clean these up (can be in a future PR)?

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -846,9 +846,12 @@ mod tests {
...
         mock.expect_modify_ack_deadline().returning(move |r| {
-            extend_tx
-                .send(r.into_inner())
-                .expect("sending on channel always succeeds");
+            let r = r.into_inner();
+            if r.ack_deadline_seconds != 0 {
+                extend_tx
+                    .send(r)
+                    .expect("sending on channel always succeeds");
+            }
```
suzmue, line 854: I don't see how the changes here should have impacted the results of this test. Was this just something that needed to be cleaned up?

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -115,6 +116,12 @@ where
     extend_interval: Interval,
     // How long messages can be kept under lease
     max_lease: Duration,
+
+    // In flight lease extension operations.
+    //
+    // These are held separate from pending acks/nacks because we do not need to
+    // await them on shutdown.
+    pending_extends: JoinSet<()>,
```
suzmue, line 124: nit: It's a little strange that we are holding on to these just for the sake of testing.

For extend_and_await we could just have extend return the JoinSet or something and just throw it away when we call it, but this might complicate our other tests because that rely on the extensions happening before shutdown too much.

suzmue (approved):

src/pubsub/src/subscriber/lease_state.rs
dbolduc: Thanks for catching! This was sloppy from me. Fixed in this PR.

src/pubsub/src/subscriber/message_stream.rs
dbolduc: Nacks and lease extensions are the same RPC (`ModifyAckDeadline`). Nacks have an `ack_deadline_seconds == 0`, and lease extensions have an `ack_deadline_seconds > 0`.

Previously the lease extensions would go through, then the nacks. Now they can race, so we need to sort them out.

src/pubsub/src/subscriber/lease_state.rs
dbolduc: > just for the sake of testing.

We might want to have these lease extension tasks return a result, and await the result as an event in the event loop. Then we can mark successful lease extensions as part of #5048.

I am going to keep it for this reason (which I did not think of until recently).

> just throw it away when we call it

@PhongChuong had a suggestion like this too. Where if we make a subsequent lease extension attempt, just throw away the old pending lease extension attempts. And we could still join the latest attempt for tests.

PR #5130, opened 2026-03-24
impl(gax-internal): introduce `RequestRecorder`

coryan:
A request recorder is a task local variable used to collect the observability information about a request. Using a task local variable allows us to pass *and return* information about the request through multiple layers without having to change all the APIs.

The name is supposed to evoke "flight recorder".

Part of the changes for #4772

### DIFF
--- a/src/gax-internal/src/observability/client_signals/recorder.rs
+++ b/src/gax-internal/src/observability/client_signals/recorder.rs
@@ -267,10 +267,25 @@ impl ClientSnapshot {
     }
 
     /// Returns the server address used in the last low-level request.
-    pub fn server_address(&self) -> Option<SocketAddr> {
+    ///
+    /// If no address is known, use the target address from `info.default_host`.
+    pub fn server_address(&self) -> String {
         self.transport_snapshot
             .as_ref()
             .and_then(|s| s.server_address)
+            .map(|a| a.ip().to_string())
+            .unwrap_or_else(|| self.info.default_host.to_string())
+    }
+
+    /// Returns the server port used in the last low-level request.
+    ///
+    /// If no port is known, use the port implied by `info.default_host`.
+    pub fn server_port(&self) -> u16 {
+        self.transport_snapshot
+            .as_ref()
+            .and_then(|s| s.server_address)
+            .map(|a| a.port())
+            .unwrap_or(443)
     }
 
     /// Returns the URL template (e.g. "/v1/storage/b/{bucket}") used in the last low-level request.
@@ -394,7 +409,9 @@ mod tests {
 
         assert_eq!(snap.attempt_count, 1, "{snap:?}");
         assert_eq!(snap.http_status_code(), Some(404), "{snap:?}");
-        assert_eq!(snap.server_address(), Some(server.addr()), "{snap:?}");
+        let addr = server.addr();
+        assert_eq!(snap.server_address(), addr.ip().to_string(), "{snap:?}");
+        assert_eq!(snap.server_port(), addr.port(), "{snap:?}");
     }
 
     #[tokio::test(start_paused = true)]
@@ -419,7 +436,12 @@ mod tests {
 
         assert_eq!(snap.attempt_count, 1, "{snap:?}");
         assert!(snap.http_status_code().is_none(), "{snap:?}");
-        assert!(snap.server_address().is_none(), "{snap:?}");
+        assert_eq!(
+            snap.server_address().as_str(),
+            TEST_INFO.default_host,
+            "{snap:?}"
+        );
+        assert_eq!(snap.server_port(), 443, "{snap:?}");
     }
 
     async fn simulate_storage_client_transport_layer(url: &str) -> Result<String, Error> {
@@ -466,6 +488,8 @@ mod tests {
         assert_eq!(snap.url(), Some(url.as_str()), "{snap:?}");
         assert_eq!(snap.attempt_count, 1, "{snap:?}");
         assert_eq!(snap.http_status_code(), Some(404), "{snap:?}");
-        assert_eq!(snap.server_address(), Some(server.addr()), "{snap:?}");
+        let addr = server.addr();
+        assert_eq!(snap.server_address(), addr.ip().to_string(), "{snap:?}");
+        assert_eq!(snap.server_port(), addr.port(), "{snap:?}");
     }
 }


src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -0,0 +1,345 @@
...
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub struct TransportSnapshot {
+    start: Instant,
+    server_address: Option<SocketAddr>,
+    rpc_system: Option<&'static str>,
+    rpc_method: Option<&'static str>,
```
haphungw: I think `rpc_method` should be moved up to `ClientSnapshot` since we can't access it at the transport level?

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -0,0 +1,345 @@
...
+        let options = RequestOptions::default().insert_extension(PathTemplate("/v7/{funny}"));
+        let client = reqwest::Client::new();
+        let request = client.get("http://127.0.0.1:1/v7/will-not-work").build()?;
+
+        recorder.on_http_request(&options, &request);
+        let snap = recorder.client_snapshot();
+        assert_eq!(snap.start, Instant::now(), "{snap:?}");
+        assert_eq!(snap.url_template(), Some("/v7/{funny}"), "{snap:?}");
+        assert_eq!(snap.rpc_system(), Some("http"), "{snap:?}");
+        assert!(snap.rpc_method().is_none(), "{snap:?}");
```
haphungw: why should `rpc_method()` return none? it also can't be populated in the response since `reqwest::Response` does not have this field.

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -0,0 +1,345 @@
...
+/// each client-level request (T3) and transport-level request (T4). To meet the requirements we
+/// need to capture information as the request makes progress. For example, the client request
+/// telemetry includes information about the last low-level request, such as the remote server IP
+/// address and port. It is difficult to carry this information through the different layers without
+/// breaking changes APIs.
+///
+/// This type solves that problem by setting a task-local (think "thread local" but for
+/// asynchronous tasks) variable valid for the full request. Each layer adds information to this
+/// variable. Once the telemetry layer is ready to emit a signal it consults the variable and uses
+/// the latest snapshot to populate the attributes of the signal.
```
westarle, line 43: maybe something like:

> the task_local is intended to be scoped to a single "client request", i.e. its lifetime matches the time between deciding to make a specific request (specific RPC method and endpoint) and the time a final success or failure result is available, including retried transport requests.

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -0,0 +1,345 @@
...
+#[non_exhaustive]
+pub struct ClientSnapshot {
+    pub start: Instant,
+    pub info: InstrumentationClientInfo,
+    pub attempt_count: u32,
+    pub transport_snapshot: Option<TransportSnapshot>,
+}
+
+impl ClientSnapshot {
+    fn new(info: InstrumentationClientInfo) -> Self {
```
westarle, line 242: I think we should capture everything from client_request_signals! here, including rpc method.

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -0,0 +1,345 @@
...
+        guard.attempt_count += 1;
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub struct ClientSnapshot {
+    pub start: Instant,
+    pub info: InstrumentationClientInfo,
+    pub attempt_count: u32,
```
westarle: if we don't pub these we have more flexibility in where we source them from fn's and I think a better opportunity to document them.

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -0,0 +1,345 @@
...
+            .as_ref()
+            .and_then(|s| s.url.as_deref())
+    }
+}
+
+#[derive(Clone, Debug, PartialEq)]
+#[non_exhaustive]
+pub struct TransportSnapshot {
+    start: Instant,
+    server_address: Option<SocketAddr>,
```
westarle, line 332: better to use fn's here too, I think.

We'd want "server_address" to resolve to the host in URL for e.g. dns errors that don't return HTTP status.

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -0,0 +1,345 @@
...
+    ///
+    /// In this context, responses that return an error status code are considered successful,
+    /// we just need them to capture their data for the spans and metrics.
+    #[cfg(feature = "_internal-http-client")]
+    pub fn on_http_response(&self, response: &reqwest::Response) {
+        let mut guard = self.inner.lock().expect("never poisoned");
+        guard.attempt_count += 1;
+        if let Some(s) = guard.transport_snapshot.as_mut() {
+            s.server_address = response.remote_addr();
+            s.http_status_code = Some(response.status().as_u16());
```
westarle, line 179: Will we eventually set error details here and in on_http_error?

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -0,0 +1,345 @@
...
+///     recorder.scope(transport_layer()).await
+/// }
+///
+/// fn info() -> InstrumentationClientInfo {
+/// # panic!();
+/// }
+/// async fn transport_layer() -> google_cloud_gax::Result<String> {
+/// # panic!("")
+/// }
+/// ```
```
westarle, line 75: the comment is already long, but it might still be helpful to describe the lifecycles.

> Client library implementations must call RequestRecorder::new and bring it in scope before making any requests.  Transports must make matched calls to on_x_request and exactly one of on_x_error or on_x_response.

> Instrumentation should read values directly after calling on_x_error or on_x_response to populate logs, traces and metrics.

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -0,0 +1,345 @@
...
+        guard.attempt_count += 1;
+        if let Some(s) = guard.transport_snapshot.as_mut() {
+            s.server_address = response.remote_addr();
+            s.http_status_code = Some(response.status().as_u16());
+        }
+    }
+
+    /// Call when it was not possible to send an HTTP request.
+    #[cfg(feature = "_internal-http-client")]
+    pub fn on_http_error(&self, _err: &Error) {
```
westarle, line 185: I'm wondering about the case where a transport impl calls on_http_error and then wants to update a 'T4' span, log and metric based on data in this recorder.  I think it is possible (and helpful since the recorder can have full info and clean API.)

Is it possible that a on_http_success can arrive before the transport can call client_snapshot()?

src/gax-internal/src/observability/client_signals/recorder.rs
coryan: Done.

If you mean the `gax-internal` clients, note that the if the transport is gRPC we certainly can:

https://github.com/googleapis/google-cloud-rust/blob/5c4749f7c859fbfc39e49fdf0c8e04f065fa2ce2/src/storage/src/generated/gapic/transport.rs#L71

For HTTP-based services we need to decide if this is injected by the transport or tracing layers.  Either one could do it, the generator knows what it is.

src/gax-internal/src/observability/client_signals/recorder.rs
coryan: I have changed all the tests to use the normal, expected flow for an HTTP-based request.  Hopefully it makes more sense now.

src/gax-internal/src/observability/client_signals/recorder.rs
coryan: Hmmm... I don't think we can:

- The URL template and resource name is not know until we resolved the binding in the `transport.rs` code.
- The RPC method may be `None` for the Storage library.

I have changed the tests to outline how the different layers will fill up the data, maybe that explains the intent?

src/gax-internal/src/observability/client_signals/recorder.rs
coryan: Done.

src/gax-internal/src/observability/client_signals/recorder.rs
coryan: Yes.

src/gax-internal/src/observability/client_signals/recorder.rs
coryan: > Is it possible that a on_http_success can arrive before the transport can call client_snapshot()?

I would expect that the things generating T4 spans, logs and metrics call `transport_snapshot()` (a thing we have not yet defined) after each transport-level request and *before* the next retry attempt.

Basically the gaxi abstractions will return a future decorated to generate those T4-level things before the future returns to the retry loop.

Does that make sense?

src/gax-internal/src/observability/client_signals/recorder.rs
coryan: Done, with some edits.

src/gax-internal/src/observability/client_signals/recorder.rs
coryan: That seems to duplicate the comments you suggested below, skipping for now.

src/gax-internal/src/observability/client_signals/recorder.rs
westarle: as long as its a retry loop, this should be fine.

I was trying to think about request hedging or something where multiple writers might interfere with each other.

src/gax-internal/src/observability/client_signals/recorder.rs
westarle: I think this one is an actual bug, we should record the logical server.address in requests that fail without an HTTP response without requiring the user to record/parse url.full or anything like that.

https://opentelemetry.io/docs/specs/semconv/registry/attributes/server/

coryan: Done.

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -267,10 +267,25 @@ impl ClientSnapshot {
...
     /// Returns the server address used in the last low-level request.
-    pub fn server_address(&self) -> Option<SocketAddr> {
+    ///
+    /// If no address is known, use the target address from `info.default_host`.
+    pub fn server_address(&self) -> String {
+        self.transport_snapshot
+            .as_ref()
+            .and_then(|s| s.server_address)
+            .map(|a| a.ip().to_string())
+            .unwrap_or_else(|| self.info.default_host.to_string())
```
westarle, line 277: This should parse `self.url` to use the actual endpoint we tried

i.e. if the customer set something bad in `with_endpoint` they should see that with connection timeouts etc.

lmk if you want me to file an issue.

coryan: Please do.  Keep in mind that `self.url` maybe `None` too, e.g., for gRPC.  So we need another level of fallback.

PR #5134, opened 2026-03-25
feat(storage): skip unnecessary checksum computation during read_object

vsharonlynn:
Continuation to issue #4285.

### DIFF
--- a/src/storage/src/storage/read_object/resumable.rs
+++ b/src/storage/src/storage/read_object/resumable.rs
@@ -39,13 +39,14 @@ impl ResumableResponse {
         let full = reader.request.read_offset == 0 && reader.request.read_limit == 0;
         let headers = response.headers();
         let response_checksums = checksums_from_response(full, response.status(), headers);
-
         let range = response_range(&response).map_err(Error::deser)?;
         let generation =
             parse_http_response::response_generation(&response).map_err(Error::deser)?;
 
         let highlights = parse_http_response::object_highlights(generation, headers)?;
 
+        // If the server response has an empty checksum (e.g. on range read),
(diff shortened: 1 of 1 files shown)


vsharonlynn:
Double-checking this PR.

vsharonlynn:
More details about this PR:

This change optimises for read ranges by short-circuiting the computation if the server returns empty checksums (regardless of request options). It solves "checksums should be disabled by default when downloading a range" in the [issue](https://github.com/googleapis/google-cloud-rust/issues/4285#issuecomment-4023080799).

src/storage/src/storage/read_object/resumable.rs
```diff
@@ -82,10 +94,12 @@ impl ResumableResponse {
...
             Ok(Some(chunk)) => {
-                self.reader
-                    .options
-                    .checksum
-                    .update(self.range.start, &chunk);
+                if self.should_track_checksums {
+                    self.reader
+                        .options
+                        .checksum
+                        .update(self.range.start, &chunk);
```
coryan: Isn't this calling:

https://github.com/googleapis/google-cloud-rust/blob/a929ca0ed9bb23d626ddebdb2bf110cd3a4cafa1/src/storage/src/storage/checksum/details.rs#L72-L79

Which already skips the computation if `crc32c` and `md5` are `None`?  Why do we need the `should_track_checksums` variable?

vsharonlynn: Yes, you're right that Checksum::update already skips computation when both fields are None.
I originally added the flag to avoid calling update on every chunk, but since the benefit is minimal, I've removed the flag to simplify the code.
Thank you.

PR #5135, opened 2026-03-25
test(spanner): integration tests for read/write transactions

olavloite:
Adds integration tests for read/write transactions. Verifies that the transaction runner automatically commits the transaction if the transaction is successful, and automatically rollbacks the transaction in case of an error.


dbolduc:
We added a new required build, so this probably needs a rebase + a `cargo fmt -p google-cloud-spanner`

PR #5139, opened 2026-03-25
feat(spanner): Batch DML

olavloite:
Adds support for Batch DML statements. Normally, a Batch DML statement returns an array of update counts containing the exact update count of each statement.

A Batch DML statement can be partly executed by Spanner and fail after N statements. In that case, the client returns a specific error that contains the update counts of the first N statements, and an error for the N+1th statement.

It is also possible that a Batch DML statement is partly executed, and that the read/write transaction is aborted by Spanner after N statements. This is propagated as a 'normal' error to trigger a transaction retry.

### DIFF
--- a/src/spanner/src/transaction_runner.rs
+++ b/src/spanner/src/transaction_runner.rs
@@ -501,43 +501,45 @@ mod tests {
 
         expect_begin_transaction(&mut mock, 2, vec![9, 9, 9]);
 
-        let mut attempt = 0;
+        let mut seq = mockall::Sequence::new();
         mock.expect_execute_batch_dml()
-            .times(2)
+            .once()
+            .in_sequence(&mut seq)
             .returning(move |_req| {
-                attempt += 1;
-                if attempt == 1 {
-                    // Return a successful response but with an embedded aborted status.
-                    let status = Status {
-                        code: Code::Aborted as i32,
-                        message: "transaction aborted".to_string(),
-                        ..Default::default()
-                    };
+                // Return a successful response but with an embedded aborted status.
+                let status = Status {
+                    code: Code::Aborted as i32,
+                    message: "transaction aborted".to_string(),
+                    ..Default::default()
+                };
 
-                    Ok(tonic::Response::new(v1::ExecuteBatchDmlResponse {
-                        result_sets: vec![v1::ResultSet {
-                            stats: Some(v1::ResultSetStats {
-                                row_count: Some(RowCount::RowCountExact(1)),
-                                ..Default::default()
-                            }),
+                Ok(tonic::Response::new(v1::ExecuteBatchDmlResponse {
+                    result_sets: vec![v1::ResultSet {
+                        stats: Some(v1::ResultSetStats {
+                            row_count: Some(RowCount::RowCountExact(1)),
                             ..Default::default()
-                        }],
-                        status: Some(status),
+                        }),
                         ..Default::default()
-                    }))
-                } else {
-                    // Return success after the retry.
-                    Ok(tonic::Response::new(v1::ExecuteBatchDmlResponse {
-                        result_sets: vec![v1::ResultSet {
-                            stats: Some(v1::ResultSetStats {
-                                row_count: Some(RowCount::RowCountExact(5)),
-                                ..Default::default()
-                            }),
+                    }],
+                    status: Some(status),
+                    ..Default::default()
+                }))
+            });
+        mock.expect_execute_batch_dml()
+            .once()
+            .in_sequence(&mut seq)
+            .returning(move |_req| {
+                // Return success after the retry.
+                Ok(tonic::Response::new(v1::ExecuteBatchDmlResponse {
+                    result_sets: vec![v1::ResultSet {
+                        stats: Some(v1::ResultSetStats {
+                            row_count: Some(RowCount::RowCountExact(5)),
                             ..Default::default()
-                        }],
+                        }),
                         ..Default::default()
-                    }))
-                }
+                    }],
+                    ..Default::default()
+                }))
             });
 
         mock.expect_commit()


src/spanner/src/batch_dml.rs
```diff
@@ -0,0 +1,289 @@
...
+            ..Default::default()
+        };
+        let rs = ResultSet {
+            stats: Some(stats),
+            ..Default::default()
+        };
+
+        // Note: crate::model::Status is the common type for status embedded in generated grpc responses.
+        let err_status = google_cloud_rpc::model::Status::default()
+            .set_code(3) // INVALID_ARGUMENT
```
dbolduc: nit:

```suggestion
            .set_code(Code::InvalidArgument as i32)
```

src/spanner/src/batch_dml.rs
```diff
@@ -0,0 +1,289 @@
...
+            status: Some(err_status),
+            ..Default::default()
+        };
+
+        let result = process_response(response);
+        let err = result.expect_err("should return error");
+        let batch_err = BatchUpdateError::extract(&err).expect("should extract BatchUpdateError");
+
+        assert_eq!(batch_err.update_counts, vec![3]);
+        assert_eq!(batch_err.status.status().expect("status").code, 3.into());
```
dbolduc: nit:
```suggestion
        assert_eq!(batch_err.status.status().expect("status").code, Code::InvalidArgument);
```

src/spanner/src/batch_dml.rs
```diff
@@ -0,0 +1,289 @@
...
+            row_count: Some(RowCount::RowCountExact(3)),
+            ..Default::default()
+        };
+        let rs = ResultSet {
+            stats: Some(stats),
+            ..Default::default()
+        };
+
+        let err_status = google_cloud_rpc::model::Status::default()
+            .set_code(google_cloud_gax::error::rpc::Code::Aborted as i32)
```
dbolduc: nit:
```suggestion
            .set_code(Code::Aborted as i32)
```

src/spanner/src/batch_dml.rs
```diff
@@ -0,0 +1,289 @@
...
+        let response = ExecuteBatchDmlResponse {
+            result_sets: vec![rs],
+            status: Some(err_status),
+            ..Default::default()
+        };
+
+        let result = process_response(response);
+        let err = result.expect_err("should return error");
+        let batch_err = BatchUpdateError::extract(&err);
+        assert!(batch_err.is_none());
```
dbolduc: nit: 
```suggestion
        assert!(batch_err.is_none(), "{batch_err:?}");
```

src/spanner/src/batch_dml.rs
```diff
@@ -0,0 +1,289 @@
...
+        assert_eq!(
+            err.status().expect("status").code,
+            google_cloud_gax::error::rpc::Code::Aborted
```
dbolduc: nit: 
```suggestion
            Code::Aborted
```

src/spanner/src/error.rs
```diff
@@ -38,3 +41,87 @@ impl SpannerInternalError {
 pub(crate) fn internal_error(message: impl Into<String>) -> crate::Error {
     crate::Error::deser(SpannerInternalError::new(message))
 }
+
+/// An error that occurs when a `execute_batch_update` partially succeeds.
+/// It contains the update counts for each statement evaluated prior to the failure,
```
dbolduc: nit: separate these with a newline. Our docs typically have a one-liner short description, then the rest of the details in their own paragraph(s)

dbolduc: also s/when a/when an/

src/spanner/src/error.rs
```diff
@@ -38,3 +41,87 @@ impl SpannerInternalError {
...
+///
+/// Statements are executed serially in the order provided in the batch.
+/// The `update_counts` correspond to the executed statements in the original
+/// request based on their relative order. The statement that failed is the
+/// one that follows directly after the last statement with an update count.
+/// Execution stops at the first failed statement, and the remaining statements
+/// are not executed.
+#[derive(thiserror::Error, Debug)]
+#[error("{status}")]
+pub struct BatchUpdateError {
```
dbolduc, line 59: Q: will this ever grow?

Adding fields is a breaking change, unless we tag the struct with `#[non_exhaustive]`. If you are unsure, the safe thing to do is add `#[non_exhaustive]`.

src/spanner/src/error.rs
```diff
@@ -38,3 +41,87 @@ impl SpannerInternalError {
...
+    /// The error that caused the batch to fail.
+    #[source]
+    pub status: crate::Error,
+}
+
+impl BatchUpdateError {
+    /// Extracts a `BatchUpdateError` from a `google_cloud_spanner::Error`, if present.
+    pub fn extract(err: &crate::Error) -> Option<&Self> {
+        err.source()
+            .and_then(|source| source.downcast_ref::<BatchUpdateError>())
```
dbolduc, line 71: comment: hmm this is assuming one level of nesting. Probably ok, just keep this in the back of your mind.

src/spanner/src/error.rs
```diff
@@ -38,3 +41,87 @@ impl SpannerInternalError {
...
+    /// The number of rows modified by each successful statement before the failure.
+    pub update_counts: Vec<i64>,
+    /// The error that caused the batch to fail.
+    #[source]
+    pub status: crate::Error,
+}
+
+impl BatchUpdateError {
+    /// Extracts a `BatchUpdateError` from a `google_cloud_spanner::Error`, if present.
+    pub fn extract(err: &crate::Error) -> Option<&Self> {
```
dbolduc, line 69: Is this worth a public fn? We have not added one in other libraries that have nested errors.

src/spanner/src/batch_dml.rs
```diff
@@ -0,0 +1,289 @@
...
+            .stats
+            .ok_or_else(|| internal_error("No stats returned for a successful statement"))
+            .and_then(|stats| match stats.row_count {
+                Some(RowCount::RowCountExact(c)) => Ok(c),
+                _ => Err(internal_error(
+                    "ExecuteBatchDml returned an invalid or missing row count type",
+                )),
+            });
+
+        update_counts.push(exact_count?);
```
dbolduc, line 83: comment (for my understanding): This discards partial results, but I assume the "no stats" and "missing row count" error cases are ones that never happen in practice unless things have gone horribly wrong. And we still need to write the code to handle them.

src/spanner/src/read_write_transaction.rs
```diff
@@ -122,27 +128,139 @@ impl ReadWriteTransaction {
...
+    ///             println!("The entire batch failed, or an internal error occurred: {}", e);
+    ///         }
+    ///     }
+    /// }
+    /// # Ok(())
+    /// # }).await?;
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn execute_batch_update(&self, batch: BatchDml) -> crate::Result<Vec<i64>> {
```
dbolduc, line 210: It's a bit weird to me that there is a `Vec<i64>` normally, and in the error case.


If `update_counts` should always be returned, did you consider something like:

```rs
pub struct BatchUpdateResponse {
  pub update_counts: Vec<i64>,
  pub result: crate::Result<()>,
}
```

It's a little clunky, but worth considering. Maybe you are afraid it does not force the application to look at the `result` field?

---

Q: So the application can learn something from the `update_counts` in the success case, and that is why we return it. Is that right?

If not, we could return a `std::result::Result<(), BatchUpdateError>`. Where the source of the `BatchUpdateError` might be a `crate::Error`, with an empty `update_counts` if there is some gRPC error.

src/spanner/src/read_write_transaction.rs
```diff
@@ -455,6 +574,126 @@ mod tests {
...
+                ..Default::default()
+            }))
+        });
+
+        let (db_client, _server) = setup_db_client(mock).await;
+
+        let tx = ReadWriteTransactionBuilder::new(db_client)
+            .begin_transaction()
+            .await
+            .expect("Failed to build transaction");
```
dbolduc: style nit: we prefer to use `?` and have our tests return `anyhow::Result<()>`

src/spanner/src/read_write_transaction.rs
```diff
@@ -455,6 +574,126 @@ mod tests {
...
+
+        let batch = BatchDml::builder()
+            .add_statement("UPDATE Users SET Name = 'Alice' WHERE Id = 1")
+            .add_statement("UPDATE Users SET Name = 'Bob' WHERE Id = 2");
+
+        let counts = tx.execute_batch_update(batch.build()).await.unwrap();
```
dbolduc: same.

src/spanner/src/transaction_runner.rs
```diff
@@ -488,4 +488,83 @@ mod tests {
...
+        let mut mock = create_session_mock();
+
+        expect_begin_transaction(&mut mock, 2, vec![9, 9, 9]);
+
+        let mut attempt = 0;
+        mock.expect_execute_batch_dml()
+            .times(2)
+            .returning(move |_req| {
+                attempt += 1;
+                if attempt == 1 {
```
dbolduc: please use `mockall::Sequence` with separate `mock.expect_foo()` calls.

Here is an example:

https://github.com/googleapis/google-cloud-rust/blob/eecdd62d51e69a3db9314e4c35a48148ec0de336/src/pubsub/src/subscriber/stream.rs#L315

src/spanner/src/read_write_transaction.rs
```diff
@@ -122,27 +128,139 @@ impl ReadWriteTransaction {
...
+    ///         if let Some(batch_error) = BatchUpdateError::extract(&e) {
+    ///             println!(
+    ///                 "Batch failed halfway. Successful update counts before failure: {:?}",
+    ///                 batch_error.update_counts
+    ///             );
+    ///         } else {
+    ///             println!("The entire batch failed, or an internal error occurred: {}", e);
+    ///         }
+    ///     }
+    /// }
```
dbolduc, line 204: I think this would look like:

```rs
let resp = transaction.execute_batch_update(batch).await;
match resp.result {
  Ok(()) => println!("All statements succeeded. Update counts: {:?}", resp.update_counts),
  Err(e) => println!("Batch failed. Successful update counts before failure: {:?}", resp.update_counts),
}
```

(The case where `update_counts.is_empty()` does not seem like a special case to me 🤷)

src/spanner/src/error.rs
olavloite: Done

src/spanner/src/error.rs
olavloite: It has been stable for the other clients since it was launched, but there is no 100% certainty, so added `#[non_exhaustive]`.

src/spanner/src/error.rs
olavloite: I think it is in this specific case. When using `execute_batch_update`, `BatchUpdateError` is one of the primary error that an application developer would want to use. I think that having a simple way of extracting that error, without needing to know exactly how it is nested, make this more application developer friendly.

src/spanner/src/error.rs
olavloite: Ack. This error is also only intended to be created by this client. We could enforce that even further by making the fields of the struct private and add getters, but I felt that the additional boilerplate was not really worth it in this case.

src/spanner/src/batch_dml.rs
olavloite: Correct. Spanner is according to the API contract required to only return result sets with a `RowCount::RowCountExact`. Anything else would be a violation of the contract (and in reality does not happen).

src/spanner/src/batch_dml.rs
olavloite: Done

src/spanner/src/batch_dml.rs
olavloite: Done

src/spanner/src/batch_dml.rs
olavloite: Done

src/spanner/src/batch_dml.rs
olavloite: Done

src/spanner/src/batch_dml.rs
olavloite: Done

src/spanner/src/read_write_transaction.rs
olavloite: > Q: So the application can learn something from the update_counts in the success case, and that is why we return it. Is that right?

The application can learn something from it in both the success and failure case:
1. Success: Verify that the number of rows affected is what you expect. This is often used by query generators and other frameworks (e.g. ORMs) to verify that the update they generated actually updated the number of rows that they expected.
2. Failure: In theory, you might want to verify the same as in the success case. But the primary use case is to understand which statement caused the failure.

The way that applications then can determine which statement failed is a bit clunky, but this is actually a relatively standard way of handling this type of errors across many programming languages. This is the standard in Java, .NET, and Python for all database drivers. And to a certain degree also Go, but as Go has multiple return arguments, it is there returned as both a result __and__ an error, instead of putting the update counts in the error.

> Maybe you are afraid it does not force the application to look at the result field?

Yes, this is the primary reason why I would prefer not to return a `BatchUpdateResponse` with a result or status field. I am absolutely sure that there will be applications that get that wrong and think that their batch succeeded when it didn't.

> If not, we could return a std::result::Result<(), BatchUpdateError>. Where the source of the BatchUpdateError might be a crate::Error, with an empty update_counts if there is some gRPC error.

I actually considered that. The primary reason I did not choose that was to keep the signature of this method as equal as possible to the other `execute_...` methods that we have (in this case it would be `std::result::Result<Vec<i64>, BatchUpdateError>`, as we definitely want to return the update counts in case of success).

One (tiny) additional drawback of that approach is that it is harder for an application to distinguish between a gRPC error and an actual `BatchUpdateError` that failed for the first statement. Although we could consider making the list of `update_counts` optional and return `None` in that case.

src/spanner/src/read_write_transaction.rs

src/spanner/src/read_write_transaction.rs

src/spanner/src/read_write_transaction.rs
dbolduc: > Yes, this is the primary reason why I would prefer not to return a BatchUpdateResponse with a result or status field. I am absolutely sure that there will be applications that get that wrong and think that their batch succeeded when it didn't.

Alright, makes sense to me. I suspected something like this.

src/spanner/src/transaction_runner.rs
olavloite: Done (missed this one yesterday)

PR #5140, opened 2026-03-25
perf(pubsub): concurrent acks/nacks

dbolduc:
Most of the work for #3975 

Send ack/nack RPCs in a background task. The keen eye will notice that the lease loop is no longer blocked on any `await` points! (Well, except for shutdown, where we do want to await these RPCs).

I will send a follow up PR to garbage collect the pending lease extensions in `next_event()` instead of `extend()`. It seems cleaner to me.

Then I will do some benchmarking.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -26,6 +26,7 @@ use tokio::sync::oneshot::Sender;
 // Use a `tokio::time::Instant` to facilitate time-based unit testing.
 use tokio::task::JoinSet;
 use tokio::time::{Duration, Instant, Interval, interval_at};
+use tokio_util::task::TaskTracker;
 
 // An ack ID is less than 200 bytes. The limit for a request is 512kB. It should
 // be safe to fit 2500 Ack IDs in a single RPC.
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -117,6 +117,9 @@ where
     // How long messages can be kept under lease
     max_lease: Duration,
 
+    // In flight acks and nacks.
```
PhongChuong, line 121: Consider using tokio_util [TaskTracker](https://docs.rs/tokio-util/latest/tokio_util/task/struct.TaskTracker.html#comparison-to-joinset)

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -165,6 +169,9 @@ where
             return LeaseEvent::Flush;
         }
 
+        // Clean up the state of pending operations.
+        while self.pending_acks_nacks.try_join_next().is_some() {}
```
PhongChuong: Consider using tokio_util [TaskTracker](https://docs.rs/tokio-util/latest/tokio_util/task/struct.TaskTracker.html#comparison-to-joinset) so this is no longer needed.

dbolduc: TIL, thanks.

Yeah, based on the documentation `TaskTracker` seems slightly more appropriate. I used it.

PR #5143, opened 2026-03-25
feat(o11y): align L3 client request logs with payload schema

haphungw:
Add fields to Client Request logs so that they comply with the standard OTel log payload schema.

Note: this establishes the correct schema mapping so impending refactors can inherit the formatting for Client logs.


dbolduc (approved):
LGTM, but note that I am not paying attention to the `o11y` project.

westarle (changes requested):
I think this is a positive change to the existing client instrumentation, but I think we should pause and help with #5148 and future PRs.

haphungw:
These attributes have been updated in the refactored implementation: https://github.com/googleapis/google-cloud-rust/pull/5172. Closing for redundancy.

PR #5147, opened 2026-03-25
feat(auth): add support for custom env, fs, and http implementations

codeshaunted:
This is a draft of a PR to add support for pluggable custom implementations of environment variable fetching, filesystem I/O and HTTP clients in `google-cloud-auth`.                                                                                                                       
                                                                                                                                                                                                                                                                                            
We are implementing a virtual machine with custom implementations of HTTP requests, environment variables, and filesystem I/O that will also need to be able to auth gcloud requests. Ideally we'd use this official crate for this, but to do so it needs modifications to support the custom I/O impls mentioned above.                                                                                                                                                                                                                                                           
                                                                                                                                                                                                                                                                                            
I'm mostly posting this PR for feedback to see if modifications like this have any chance of being merged into the official crate. The implementation here should not require a major version bump for the public stable API, but it goes deep under the hood to thread custom I/O implementations throughout the codebase.


coryan:
> This is a draft of a PR to add support for pluggable custom implementations of environment variable fetching, filesystem I/O and HTTP clients in `google-cloud-auth`.
> 
> We are implementing a virtual machine with custom implementations of HTTP requests, environment variables, and filesystem I/O that will also need to be able to auth gcloud requests. Ideally we'd use this official crate for this, but to do so it needs modifications to support the custom I/O impls mentioned above.

Thanks for the additional context, that helps.

I assume "we" is your company ? "BoundaryML"?

> I'm mostly posting this PR for feedback to see if modifications like this have any chance of being merged into the official crate.

I think there is a chance...  Can you give us some time to think through the maintenance effort for these new APIs? As you say, they are not breaking, but we also want to think about how they could change in the future and whether those future chance may make them breaking.

The PR is also larger than we would like. Don't do this until we get back to you, but we would probably want to break this down into smaller PRs.

> The implementation here should not require a major version bump for the public stable API, but it goes deep under the hood to thread custom I/O implementations throughout the codebase.

Ack.

codeshaunted:
> I assume "we" is your company ? "BoundaryML"?

Yes! We're working on a programming language that among other things will allow users to easily build strongly typed interfaces around LLM responses, and first-class support for GCP's Vertex is one of our goals.

> I think there is a chance... Can you give us some time to think through the maintenance effort for these new APIs? As you say, they are not breaking, but we also want to think about how they could change in the future and whether those future chance may make them breaking.
> 
> The PR is also larger than we would like. Don't do this until we get back to you, but we would probably want to break this down into smaller PRs.

All good. Appreciate you replying so quickly!

coryan:
> > I assume "we" is your company ? "BoundaryML"?
> 
> Yes! We're working on a programming language that among other things will allow users to easily build strongly typed interfaces around LLM responses, and first-class support for GCP's Vertex is one of our goals.
> 
> > I think there is a chance... Can you give us some time to think through the maintenance effort for these new APIs? As you say, they are not breaking, but we also want to think about how they could change in the future and whether those future chance may make them breaking.
> > The PR is also larger than we would like. Don't do this until we get back to you, but we would probably want to break this down into smaller PRs.
> 
> All good. Appreciate you replying so quickly!

Can you answer my questions in #5168 ?  I will probably have some follow ups.

Also, if I misrepresented the intent, please comment on that bug and set the record straight.

coryan:
Closing for now. When you have a new PR for #5168 please send it our way.

PR #5149, opened 2026-03-26
fix(pubsub): batch nacks on shutdown to respect service limits

### LINKED ISSUE #4847: Subscriber nacks on shutdown need to respect quota limits
https://github.com/googleapis/google-cloud-rust/blob/df77ebc841b24f23dc52cd747e07bc82e61975ec/src/pubsub/src/subscriber/lease_state.rs#L221

While `self.to_nack.len() <= ACK_IDS_PER_RPC`, that stops being true when we extend it by the messages currently under lease management.

We need to break this up into batches.

suzmue:
Ensure that when the Pub/Sub subscriber shuts down, it does not exceed the MAX_IDS_PER_RPC limit (2,500 IDs) when sending negative acknowledgments (nacks) for messages still under lease management. It also refactors the lease eviction logic to reduce code duplication.

Fixes #4847

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -34,8 +34,8 @@ use tokio_util::task::TaskTracker;
 // https://docs.cloud.google.com/pubsub/quotas
 const MAX_IDS_PER_RPC: usize = 2500;
 
-// Helper function to chunk ack ids to nack into chunks of MAX_IDS_PER_RPC.
-fn batch_drained(ack_ids: Vec<String>) -> Vec<Vec<String>> {
+// Helper function to chunk ack ids into chunks of MAX_IDS_PER_RPC.
+fn batch(ack_ids: Vec<String>) -> Vec<Vec<String>> {
     ack_ids
         .chunks(MAX_IDS_PER_RPC)
         .map(|c| c.to_vec())
--- a/src/pubsub/src/subscriber/lease_state/at_least_once.rs
+++ b/src/pubsub/src/subscriber/lease_state/at_least_once.rs
@@ -104,7 +104,7 @@ impl Leases {
     /// Called during shutdown, if configured to `NackImmediately`.
     pub fn evict_and_drain(mut self) -> (Vec<String>, Vec<Vec<String>>) {
         self.to_nack.extend(self.under_lease.into_keys());
-        (self.to_ack, super::batch_drained(self.to_nack))
+        (self.to_ack, super::batch(self.to_nack))
     }
 }
 
@@ -288,14 +288,11 @@ mod tests {
             leases
         );
 
-        let (mut to_ack, to_nack) = leases.evict_and_drain();
+        let (to_ack, to_nack) = leases.evict_and_drain();
+        assert_eq!(sorted(&to_ack), test_ids(0..10));
 
-        to_ack.sort();
-        assert_eq!(to_ack, test_ids(0..10));
-
-        let mut to_nack = NackBatches::flatten(to_nack).ack_ids;
-        to_nack.sort();
-        assert_eq!(to_nack, test_ids(10..30));
+        let to_nack = NackBatches::flatten(to_nack);
+        assert_eq!(sorted(&to_nack.ack_ids), test_ids(10..30));
     }
 
     #[test]
@@ -319,19 +316,15 @@ mod tests {
             leases
         );
 
-        let (mut to_ack, to_nack) = leases.evict_and_drain();
-
-        to_ack.sort();
-        assert_eq!(to_ack, test_ids(0..10));
+        let (to_ack, to_nack) = leases.evict_and_drain();
+        assert_eq!(sorted(&to_ack), test_ids(0..10));
 
         let to_nack = NackBatches::flatten(to_nack);
         assert_eq!(
             to_nack.counts,
             vec![MAX_IDS_PER_RPC, MAX_IDS_PER_RPC, MAX_IDS_PER_RPC - 10]
         );
-        let mut to_nack_ids = to_nack.ack_ids;
-        to_nack_ids.sort();
-        assert_eq!(to_nack_ids, test_ids(10..MAX_IDS_PER_RPC * 3));
+        assert_eq!(sorted(&to_nack.ack_ids), test_ids(10..MAX_IDS_PER_RPC * 3));
     }
 
     #[test]
--- a/src/pubsub/src/subscriber/lease_state/exactly_once.rs
+++ b/src/pubsub/src/subscriber/lease_state/exactly_once.rs
@@ -137,7 +137,7 @@ impl Leases {
                 .send(Err(AckError::Shutdown(NACK_SHUTDOWN_ERROR.into())));
             self.to_nack.push(ack_id);
         }
-        (self.to_ack, super::batch_drained(self.to_nack))
+        (self.to_ack, super::batch(self.to_nack))
     }
 }
 
@@ -165,7 +165,7 @@ impl PartialEq<Leases> for super::tests::TestLeases {
 
 #[cfg(test)]
 mod tests {
-    use super::super::tests::{NackBatches, TestLeases, test_id, test_ids};
+    use super::super::tests::{NackBatches, TestLeases, sorted, test_id, test_ids};
     use super::*;
     use std::collections::HashSet;
     use tokio::sync::oneshot::channel;
@@ -667,9 +667,8 @@ mod tests {
         let (to_ack, to_nack) = leases.evict_and_drain();
         assert!(to_ack.is_empty(), "{to_ack:?}");
 
-        let mut to_nack = NackBatches::flatten(to_nack).ack_ids;
-        to_nack.sort();
-        assert_eq!(to_nack, test_ids(1..4));
+        let to_nack = NackBatches::flatten(to_nack);
+        assert_eq!(sorted(&to_nack.ack_ids), test_ids(1..4));
 
         let err = result_rx1.await?.expect_err("error should be returned");
         assert!(matches!(err, AckError::Shutdown(_)), "{err:?}");
@@ -695,9 +694,7 @@ mod tests {
 
         let to_nack = NackBatches::flatten(to_nack);
         assert_eq!(to_nack.counts, vec![MAX_IDS_PER_RPC, 20]);
-        let mut to_nack_ids = to_nack.ack_ids;
-        to_nack_ids.sort();
-        assert_eq!(to_nack_ids, test_ids(0..MAX_IDS_PER_RPC + 20));
+        assert_eq!(sorted(&to_nack.ack_ids), test_ids(0..MAX_IDS_PER_RPC + 20));
 
         Ok(())
     }


src/pubsub/src/subscriber/lease_state/at_least_once.rs
```diff
@@ -288,15 +289,80 @@ mod tests {
...
                 under_lease: Vec::new(),
                 to_ack: test_ids(0..10),
                 to_nack: test_ids(10..30),
             },
+            Leases {
+                to_ack,
+                to_nack: to_nack.into_iter().flatten().collect(),
+                ..Default::default()
+            }
+        );
```
dbolduc: Consider:

```rs
assert_eq!(to_ack, test_ids(0..10));
assert_eq!(to_nack, vec![test_ids(10..30)]);
```

No need to put them back in a `Leases` object

src/pubsub/src/subscriber/lease_state/at_least_once.rs
```diff
@@ -288,15 +289,80 @@ mod tests {
...
+                under_lease: Vec::new(),
+                to_ack: test_ids(0..10),
+                to_nack: test_ids(10..MAX_IDS_PER_RPC * 3),
+            },
+            Leases {
+                to_ack: to_ack,
+                to_nack: to_nack.into_iter().flatten().collect(),
+                ..Default::default()
+            },
+        );
```
dbolduc: Your call on how to make the `to_nack` comparison most readable. Not sure this is any better, but could do something like:

```rs
struct NackBatches {
  counts: Vec<usize>,
  ack_ids: Vec<String>,
}

fn flatten(to_nacks: Vec<Vec<String>>) -> NackBatches {
  NackBatches {
    counts: to_nacks.map(|v| v.len()).collect(),
    ack_ids: sorted(to_nacks.into_iter().flatten().collect()),
  }
}
```

then:

```rs
let got = flatten(to_nack);
assert_eq!(got.counts, vec![MAX_IDS_PER_RPC, MAX_IDS_PER_RPC, MAX_IDS_PER_RPC - 10]);
assert_eq!(got.ack_ids, test_ids(10..MAX_IDS_PER_RPC * 3));
```

🤷

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -664,13 +665,94 @@ mod tests {
...
                 under_lease: Vec::new(),
                 to_ack: Vec::new(),
                 to_nack: test_ids(1..4),
             },
+            Leases {
+                to_ack,
+                to_nack: to_nack.into_iter().flatten().collect(),
+                ..Default::default()
+            }
+        );
```
dbolduc: same.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -664,13 +665,94 @@ mod tests {
...
+        let (result_tx, result_rx2) = channel();
+        leases.add(
+            test_id(2),
+            ExactlyOnceInfo {
+                receive_time: Instant::now(),
+                result_tx,
+                pending: false,
+            },
+        );
+
```
dbolduc: Checking the results on the channel is covered by the `evict()` test. No need to check it in the batching test.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -664,13 +665,94 @@ mod tests {
...
+        // Add MAX_IDS_PER_RPC + 10 messages under lease management. Nack some and evict.
+        for i in 3..MAX_IDS_PER_RPC + 20 {
```
dbolduc:
```suggestion
        for i in 0..MAX_IDS_PER_RPC + 20 {
```

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -664,13 +665,94 @@ mod tests {
...
+            leases.add(test_id(i), test_info());
+            if i % 2 == 0 {
+                leases.nack(test_id(i));
+            }
+        }
+        let (to_ack, to_nack) = leases.evict_and_drain();
+        assert_eq!(to_nack.len(), 2);
+        assert_eq!(to_nack[0].len(), MAX_IDS_PER_RPC as usize);
+        assert_eq!(to_nack[1].len(), 19);
```
dbolduc: ... then this is 20, not 19. Less mental math the reader needs to do.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
dbolduc: Oh, now I remember. The `impl PartialEq` for this thing sorts the lists for us. Yuck.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -34,6 +34,18 @@ use tokio_util::task::TaskTracker;
 // https://docs.cloud.google.com/pubsub/quotas
 const MAX_IDS_PER_RPC: usize = 2500;
 
+// Helper function to chunk nack ids into chunks of MAX_IDS_PER_RPC.
```
dbolduc: maybe "ack IDs to nack" instead of "nack ids" ?

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -127,17 +127,18 @@ impl Leases {
     }
 
     /// Nacks all messages under lease management that have not been acked by
-    /// the application.
+    /// the application and drains all messages from the lease state.
     ///
     /// Called during shutdown, if configured to `NackImmediately`.
-    pub fn evict(&mut self) {
+    pub fn evict_and_drain(&mut self) -> (Vec<String>, Vec<Vec<String>>) {
```
dbolduc: nit: I think we are done with the struct after this, so it should be :  `evict_and_drain(mut self)`

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -127,17 +127,18 @@ impl Leases {
...
         let under_lease = std::mem::take(&mut self.under_lease);
```
dbolduc: and no need to replace this. we can just move it out.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
```diff
@@ -99,12 +99,13 @@ impl Leases {
     }
 
     /// Nacks all messages under lease management that have not been acked by
-    /// the application.
+    /// the application and drains all messages from the lease state.
     ///
     /// Called during shutdown, if configured to `NackImmediately`.
-    pub fn evict(&mut self) {
+    pub fn evict_and_drain(&mut self) -> (Vec<String>, Vec<Vec<String>>) {
         let under_lease = std::mem::take(&mut self.under_lease);
```
dbolduc: same

src/pubsub/src/subscriber/lease_state/at_least_once.rs
```diff
@@ -335,34 +319,19 @@ mod tests {
...
-                to_nack: test_ids(10..MAX_IDS_PER_RPC * 3),
-            },
-            Leases {
-                to_ack: to_ack,
-                to_nack: to_nack.into_iter().flatten().collect(),
-                ..Default::default()
-            },
-        );
+        to_ack.sort();
+        assert_eq!(to_ack, test_ids(0..10));
```
dbolduc: nit: I think I have a shorthand to sort these containers.

```rs
        assert_eq!(sorted(&to_ack), test_ids(0..10));
```

src/pubsub/src/subscriber/lease_state/at_least_once.rs
```diff
@@ -102,10 +102,9 @@ impl Leases {
     /// the application and drains all messages from the lease state.
     ///
     /// Called during shutdown, if configured to `NackImmediately`.
-    pub fn evict_and_drain(&mut self) -> (Vec<String>, Vec<Vec<String>>) {
-        let under_lease = std::mem::take(&mut self.under_lease);
-        self.to_nack.extend(under_lease.into_keys());
-        super::batch_drained(self.drain())
+    pub fn evict_and_drain(mut self) -> (Vec<String>, Vec<Vec<String>>) {
+        self.to_nack.extend(self.under_lease.into_keys());
+        (self.to_ack, super::batch_drained(self.to_nack))
```
dbolduc: comment: I think it is cleaner now that the common fn only takes a `Vec<String>` instead of handling the tuple. 👍

src/pubsub/src/subscriber/lease_state/at_least_once.rs
dbolduc: here and elsewhere.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -34,16 +34,12 @@ use tokio_util::task::TaskTracker;
...
-    (
-        drained.0, // acks
-        drained
-            .1
-            .chunks(MAX_IDS_PER_RPC)
-            .map(|c| c.to_vec())
-            .collect(), // nacks
-    )
+// Helper function to chunk ack ids to nack into chunks of MAX_IDS_PER_RPC.
+fn batch_drained(ack_ids: Vec<String>) -> Vec<Vec<String>> {
```
dbolduc: naming nit: maybe just `batch` now?

dbolduc:
Thanks for the fix!

src/pubsub/src/subscriber/lease_state/at_least_once.rs
suzmue: Yeah this is why I used it. It is kind of weird now that the lease is consumed so I just updated the tests to sort.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
suzmue: done.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
suzmue: done.

src/pubsub/src/subscriber/lease_state.rs
suzmue: done.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
suzmue: done.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
suzmue: done.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
suzmue: done.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
suzmue: Yeah, it was a convenience to save typing when forwarding the result of drain, but since we can take things directly was able to simplify.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
suzmue: done. thanks!

src/pubsub/src/subscriber/lease_state.rs
suzmue: done

dbolduc (approved):

PR #5155, opened 2026-03-26
impl(gax-internal): fallback on URL for server address

### LINKED ISSUE #5144: ClientSnapshot::server_address should return actual endpoint attempted
Server address should be populated in connection issues like timeouts and DNS errors to help customers debug bad endpoints.  See the HTTP implementation:

https://github.com/googleapis/google-cloud-rust/blob/main/src/gax-internal/src/observability/http_tracing.rs#L85-L88

coryan points out that it might be None for gRPC, too.

> This [the implementation of `server_address`] should parse `self.url` to use the actual endpoint we tried
> 
> i.e. if the customer set something bad in `with_endpoint` they should see that with connection timeouts etc.

 _Originally posted by @westarle in [#5130](https://github.com/googleapis/google-cloud-rust/pull/5130/changes/6b3a3a9c0cf908438a44a86db6b6ba42adb1de90#r2990826346)_

coryan:
Fixes #5144

### DIFF
--- a/src/gax-internal/src/observability/client_signals/recorder.rs
+++ b/src/gax-internal/src/observability/client_signals/recorder.rs
@@ -419,6 +419,7 @@ mod tests {
     use httptest::responders::status_code;
     use httptest::{Expectation, Server};
     use pretty_assertions::assert_eq;
+    use std::net::{Ipv4Addr, SocketAddrV4};
 
     const TEST_METHOD_NAME: &str = "google.test.v1.Service/SomeMethod";
     const TEST_PATH_TEMPLATE: &str = "/v42/{parent}";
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -300,22 +303,40 @@ impl ClientSnapshot {
...
+                return host;
+            }
+        }
+        self.info.default_host.to_string()
     }
 
     /// Returns the server port used in the last low-level request.
     ///
     /// If no port is known, use the port implied by `info.default_host`.
     pub fn server_port(&self) -> u16 {
```
westarle, line 325: nit: a test could make sure this and `server_address` agree on the source of the data.

lgtm right now.

coryan: Done.

PR #5160, opened 2026-03-26
chore: major version bump for dataplex

JoeWang1127:


src/generated/cloud/dataplex/v1/Cargo.toml
```diff
@@ -16,7 +16,7 @@
 
 [package]
 name                   = "google-cloud-dataplex-v1"
-version                = "1.9.0"
+version                = "2.0.0"
```
coryan, line 19: I think you missed the `librarian generate google-cloud-dataplex-v1` step to update the README file too.

PR #5162, opened 2026-03-26
impl(gax-internal): implement `WithClientSpan` decorator

### LINKED ISSUE #5158: Implement `WithClientSpan`
This will be a decorator for `std::future::Future<Output = google_cloud_gax::Result<T>>`. It will report the duration metric when the future is ready.

Most of the information for the span attributes comes from the `RequestRecorder`. The error fields from the result.

The span is created in the `tracing.rs` file.

haphungw:
Introduce the `WithClientSpan` future decorator. This is part of the ongoing refactoring to break down the monolithic `WithClientSignals` into independent, composable decorators: `WithClientLogging`, `WithClientMetric`, and `WithClientSpan`.

Fixes #5158

### DIFF
--- a/src/gax-internal/src/observability/client_signals/with_client_span.rs
+++ b/src/gax-internal/src/observability/client_signals/with_client_span.rs
@@ -27,7 +27,6 @@ use crate::observability::errors::ErrorType;
 use google_cloud_gax::error::Error;
 use opentelemetry_semantic_conventions::attribute::{
     HTTP_RESPONSE_STATUS_CODE, RPC_METHOD, RPC_SYSTEM, SERVER_ADDRESS, SERVER_PORT, URL_FULL,
-    URL_PATH, URL_QUERY,
 };
 use pin_project::pin_project;
 use std::future::Future;
@@ -67,9 +66,6 @@ where
             return Poll::Ready(output);
         };
 
-        let url_path = snapshot.url().map(|u| u.path());
-        let url_query = snapshot.url().and_then(|u| u.query());
-
         match &output {
             Ok(_) => {
                 tracing::record_all!(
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/observability/client_signals/with_client_span.rs
```diff
@@ -0,0 +1,286 @@
...
+        let output = futures::ready!(this.inner.poll(cx));
+
+        let Some(snapshot) = RequestRecorder::current().map(|r| r.client_snapshot()) else {
+            return Poll::Ready(output);
+        };
+
+        let parsed_url = snapshot
+            .sanitized_url()
+            .and_then(|u| u.parse::<http::Uri>().ok());
+        let url_path = parsed_url.as_ref().map(|u| u.path());
```
coryan: We want these? TIL....  If that is the case we should consider capturing a `http::Url` in the recorder, seems silly to convert from URL to String back to Url.

haphungw: hmm I put them here since OTel define `url.path` and `url.query` as standard span attributes for http requests.
and done.

coryan: Thanks.  We could have done the URL changes in a follow up PR too.  In fact, I fear you may run into problems with the `_internal-http-client` feature disabled... Let's see what the build says.

westarle: url.query and url.path are required in SERVER spans and not listed for CLIENT spans.

I can't say exactly why that's the case, but I would remove them for now until they show up in the CLIENT span table.

https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-client-span
https://opentelemetry.io/docs/specs/semconv/http/http-spans/#http-server-span

haphungw: oh i see. thanks!

PR #5164, opened 2026-03-26
impl(gax-internal): introduce `WithClientMetric` decorator

### LINKED ISSUE #5157: Implement `WithClientMetric`
This will be a decorator for `std::future::Future<Output = google_cloud_gax::Result<T>>`. It will report the duration metric when the future is ready.

Most of the information for the metric attributes comes from the `RequestRecorder`. The error fields from the result.

coryan:
This decorator reports the client duration metric for the decorated future. It uses a `RequestRecorder` to obtain the request attributes, including its starting time.

Fixes #5157

### DIFF
--- a/src/gax-internal/src/observability/client_signals/recorder.rs
+++ b/src/gax-internal/src/observability/client_signals/recorder.rs
@@ -259,6 +259,10 @@ impl ClientSnapshot {
     }
 
     /// Returns the client request duration.
+    ///
+    /// This measures the time since the instance was created. Client libraries should initialize an
+    /// instance at the beginning of the request, before any RPCs or attempts to create or fetch
+    /// authentication tokens.
     pub fn client_duration(&self) -> Duration {
         self.start.elapsed()
     }


src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -257,6 +258,11 @@ impl ClientSnapshot {
         }
     }
 
+    /// Returns the client request duration.
```
westarle, line 261: nit
```suggestion
    /// Returns the client request duration (the duration since `new` was called.)
```

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -637,4 +643,13 @@ mod tests {
...
         Ok(())
     }
+
+    #[tokio::test(start_paused = true)]
+    async fn client_duration() {
+        const DURATION: Duration = Duration::from_millis(123456);
+        let recorder = RequestRecorder::new(TEST_INFO);
+        tokio::time::sleep(DURATION).await;
+        let snap = recorder.client_snapshot();
+        assert_eq!(snap.client_duration(), DURATION, "{snap:?}");
```
westarle, line 657: is the tokio timer accurate or do you need an epsilon here?

src/gax-internal/src/observability/client_signals/recorder.rs
coryan: Done, with some edits.

src/gax-internal/src/observability/client_signals/recorder.rs
coryan: tests with `start_paused = true` use a virtual clock, the time is accurate.

PR #5165, opened 2026-03-27
impl(spanner): internal retries for ResultSet

olavloite:
Adds support for transparently resuming the stream of PartialResultSets if it fails halfway with a transient error. Spanner emits resume tokens that can be used to resume the stream from a given position. This allows the ResultSet to continue to return rows, even if the underlying stream breaks and needs to be restarted.

This change focuses purely on the technical implementation of resuming a stream. Configuration of the RetryPolicy that should be used, will be added in a follow-up PR.

### DIFF
--- a/src/spanner/src/result_set.rs
+++ b/src/spanner/src/result_set.rs
@@ -346,7 +346,7 @@ impl ResultSet {
         Ok(())
     }
 
-    // TODO: Make the retry policy configurable.
+    // TODO(#5185): Make the retry policy configurable.
     fn should_retry(&self, e: &crate::Error) -> bool {
         if self.retry_count >= 10 {
             return false;


src/spanner/src/result_set.rs
```diff
@@ -111,87 +145,211 @@ impl ResultSet {
...
+                    .send()
+                    .await?;
+                self.stream = stream;
+            }
         }
+        Ok(())
+    }
 
-        None
+    // TODO: Make the retry policy configurable.
```
coryan: Please assign an issue to track it.  When you do design the retry keep this in mind:

- In the storage client we call things that "retry" an interrupted response a "resume policy".
- We also use policies for resuming interrupted streams, it would be nice if clients for different services were consistent.

olavloite: Done. Added #5185

PR #5173, opened 2026-03-27
chore(spanner): pass prev transaction ID to retry

olavloite:
When retrying an aborted transaction, the retry should include the previous transaction ID. This instructs Spanner to increase the lock priority of the retry, which increases the chances that it will succeed, as it gets priority in case of new lock conflicts during the retry. This ensures that transactions are able to make progress, also when they are repeatedly aborted by Spanner.

### DIFF
--- a/src/spanner/src/read_write_transaction.rs
+++ b/src/spanner/src/read_write_transaction.rs
@@ -36,7 +36,7 @@ use crate::statement::Statement;
 use std::sync::atomic::Ordering;
 
 /// A builder for [ReadWriteTransaction].
-#[derive(Clone)]
+#[derive(Clone, Debug)]
 pub(crate) struct ReadWriteTransactionBuilder {
     client: DatabaseClient,
     options: TransactionOptions,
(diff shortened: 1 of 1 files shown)


src/spanner/src/read_write_transaction.rs
```diff
@@ -36,6 +36,7 @@ use crate::statement::Statement;
 use std::sync::atomic::Ordering;
 
 /// A builder for [ReadWriteTransaction].
+#[derive(Clone)]
```
coryan: FWIW, you almost always want `#[derive(Debug)]` too, that is handy when running tests and (eventually) if you start adding debug logging to the client.

src/spanner/src/write_only_transaction.rs
```diff
@@ -154,19 +156,25 @@ impl WriteOnlyTransaction {
         let mutations_proto: Vec<_> = mutations.into_iter().map(|m| m.build_proto()).collect();
         let mutation_key = Mutation::select_mutation_key(&mutations_proto);
         let client = self.client;
+        let previous_transaction_id = Arc::new(Mutex::new(Bytes::new()));
```
coryan, line 159: Feels like we should be able to avoid the Arc and mutex... only one thread is using this thing at a time, the access is serialized by the async runtime...  But I cannot come up with a clever solution.

olavloite: Yeah, I was struggling with that one as well for a while, and I was also not able to come up with an elegant solution. All of them just ended up adding a lot of additional boiler plate code.

src/spanner/src/read_write_transaction.rs

PR #5175, opened 2026-03-27
impl(gax-internal): `client_request_signals` macro

coryan:
This macro will be used by generated code and hand-crafted clients to emit client-request signals.

Part of the work for #4772


src/gax-internal/src/observability/client_signals.rs
```diff
@@ -131,6 +131,24 @@ macro_rules! client_request_signals {
         }
         (start, span)
     }};
+    (metric: $metric:expr, info: $info:expr, method: $method:literal, $inner:expr) => {{
+        use ::tracing::instrument::Instrument;
+        let recorder = $crate::observability::RequestRecorder::new($info);
+        let span = $crate::client_request_span!(info: $info, method: $method);
+        // TODO(#5158) - add the span decorator.
```
westarle: done?

coryan: Good catch, fixed.

PR #5176, opened 2026-03-28
impl(gax-internal): use recorder in `ReqwestClient`

coryan:
Change the `ReqwestClient` to record interesting events in the `RequestRecorder`.

We need to change the generated code at the same time because the integration tests depend on this generated code and `gax-internal` working together.

Part of the work for #4772

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -27,7 +27,7 @@ pub mod reqwest;
 use crate::as_inner::as_inner;
 use crate::attempt_info::AttemptInfo;
 #[cfg(google_cloud_unstable_tracing)]
-use crate::observability::create_http_attempt_span;
+use crate::observability::{HttpResultExt, RequestRecorder, create_http_attempt_span};
 use google_cloud_auth::credentials::{
     Builder as CredentialsBuilder, CacheableResource, Credentials,
 };
@@ -261,22 +261,33 @@ impl ReqwestClient {
         options: RequestOptions,
         attempt_info: AttemptInfo,
     ) -> Result<reqwest::Response> {
-        use crate::observability::HttpResultExt;
         let span = create_http_attempt_span(
             &request,
             &options,
             self.instrumentation,
             attempt_info.attempt_count,
         );
-        let (method, url) = (request.method().clone(), request.url().clone());
-        self.execute_http_inner(request)
+        if let Some(recorder) = RequestRecorder::current() {
+            recorder.on_http_request(&request);
+        }
+        let result = self
+            .execute_http_inner(request)
             .instrument(span.clone())
-            .await
-            .record_http(&span, attempt_info.attempt_count, method, url)
+            .await;
+        if let Some(recorder) = RequestRecorder::current() {
+            match &result {
+                Ok(r) => recorder.on_http_response(r),
+                Err(e) => recorder.on_http_error(e),
+            }
+        }
+        result.record_http(&span)
     }
 
     #[cfg_attr(not(google_cloud_unstable_tracing), allow(unused_mut))]
     async fn execute_http_inner(&self, mut request: reqwest::Request) -> Result<reqwest::Response> {
+        // We want to send the tracing propagation headers even if tracing is disabled in the
+        // client. A global trace (say from the incoming HTTP request to Cloud Run) could be
+        // propagated.
         #[cfg(google_cloud_unstable_tracing)]
         crate::observability::propagation::inject_context(
             &tracing::Span::current(),
@@ -410,13 +421,14 @@ impl ReqwestClient {
         options: &RequestOptions,
         attempt_count: u32,
     ) -> Result<reqwest::Response> {
-        use crate::observability::HttpResultExt;
         let span = create_http_attempt_span(&request, options, self.instrumentation, attempt_count);
-        let (method, url) = (request.method().clone(), request.url().clone());
+        if let Some(recorder) = RequestRecorder::current() {
+            recorder.on_http_request(&request);
+        }
         self.request_attempt_inner(request)
             .instrument(span.clone())
             .await
-            .record_http(&span, attempt_count, method, url)
+            .record_http(&span)
     }
 
     #[cfg_attr(not(google_cloud_unstable_tracing), allow(unused_mut))]
@@ -424,12 +436,23 @@ impl ReqwestClient {
         &self,
         mut request: reqwest::Request,
     ) -> Result<reqwest::Response> {
+        // We want to send the tracing propagation headers even if tracing is disabled in the
+        // client. A global trace (say from the incoming HTTP request to Cloud Run) could be
+        // propagated.
         #[cfg(google_cloud_unstable_tracing)]
         crate::observability::propagation::inject_context(
             &tracing::Span::current(),
             request.headers_mut(),
         );
-        let response = self.inner.execute(request).await.map_err(map_send_error)?;
+        let result = self.inner.execute(request).await.map_err(map_send_error);
+        #[cfg(google_cloud_unstable_tracing)]
+        if let Some(recorder) = RequestRecorder::current() {
+            match &result {
+                Ok(r) => recorder.on_http_response(r),
+                Err(e) => recorder.on_http_error(e),
+            }
+        }
+        let response = result?;
         if !response.status().is_success() {
             return self::to_http_error(response).await;
         }
@@ -554,24 +577,14 @@ async fn to_http_response<O: serde::de::DeserializeOwned + Default>(
 #[cfg(test)]
 mod tests {
     use super::*;
-    #[cfg(google_cloud_unstable_tracing)]
-    use crate::client_request_span;
     use crate::options::ClientConfig;
     use crate::options::InstrumentationClientInfo;
     use google_cloud_auth::credentials::anonymous::Builder as Anonymous;
-    #[cfg(google_cloud_unstable_tracing)]
-    use google_cloud_test_utils::test_layer::TestLayer;
     use http::{HeaderMap, HeaderValue, Method};
-    #[cfg(google_cloud_unstable_tracing)]
-    use opentelemetry_semantic_conventions::trace as otel_trace;
     use test_case::test_case;
-    #[cfg(google_cloud_unstable_tracing)]
-    use tracing::Instrument;
-
-    type TestResult = std::result::Result<(), Box<dyn std::error::Error>>;
 
     #[tokio::test]
-    async fn client_http_error_bytes() -> TestResult {
+    async fn client_http_error_bytes() -> anyhow::Result<()> {
         let http_resp = http::Response::builder()
             .header("Content-Type", "application/json")
             .status(400)
@@ -593,7 +606,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn client_error_with_status() -> TestResult {
+    async fn client_error_with_status() -> anyhow::Result<()> {
         use google_cloud_gax::error::rpc::{Code, Status, StatusDetails::LocalizedMessage};
         let body = serde_json::json!({"error": {
             "code": 404,
@@ -634,7 +647,7 @@ mod tests {
     #[test_case(reqwest::StatusCode::OK, "{}"; "200 with empty object")]
     #[test_case(reqwest::StatusCode::NO_CONTENT, "{}"; "204 with empty object")]
     #[test_case(reqwest::StatusCode::NO_CONTENT, ""; "204 with empty content")]
-    async fn client_empty_content(code: reqwest::StatusCode, content: &str) -> TestResult {
+    async fn client_empty_content(code: reqwest::StatusCode, content: &str) -> anyhow::Result<()> {
         let response = resp_from_code_content(code, content)?;
         assert!(response.status().is_success());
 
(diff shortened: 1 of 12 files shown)


tests/o11y/src/e2e/signals.rs
```diff
@@ -198,31 +199,23 @@ fn check_logs(project_id: &str, buffer: Buffer, trace_id: TraceId) -> anyhow::Re
...
-        assert_eq!(
-            fields.get(key),
-            Some(expected),
-            "mismatch for {key:?} in {value:?}"
-        );
-    }
-
-    assert!(
-        fields
-            .get("exception.message")
```
coryan, line 220: Removing this field seems suspect but I cannot find where it was introduced before.

@westarle or @haphungw can you help me out?

tests/o11y/src/e2e/signals.rs
```diff
@@ -198,31 +199,23 @@ fn check_logs(project_id: &str, buffer: Buffer, trace_id: TraceId) -> anyhow::Re
...
+        "gcp.client.repo": "googleapis/google-cloud-rust",
+        "gcp.client.service": "showcase",
+        "error.type": "404",
+        "http.request.method": "POST",
+        "rpc.method": "google.showcase.v1beta1.Echo/Echo",
+        "rpc.service": "showcase",
         "rpc.system.name": "http",
         "url.domain": "localhost:7469", // the showcase domain...
-        "url.template": "", // TODO(#...)
-        "exception.type": "404",
```
coryan, line 208: Ditto.

src/gax-internal/src/http.rs
```diff
@@ -778,79 +791,9 @@ mod tests {
         Ok(())
     }
 
-    #[tokio::test]
-    #[cfg(google_cloud_unstable_tracing)]
-    async fn test_t3_span_enrichment() {
```
coryan, line 783: I removed this test because it is no longer relevant.  There is no "enrichment" of the T3 spans, the `WithClientTracing` decorator does the job based on the recorder data (for which I added tests).

src/gax-internal/src/http.rs
```diff
@@ -410,26 +421,38 @@ impl ReqwestClient {
...
             &tracing::Span::current(),
             request.headers_mut(),
         );
-        let response = self.inner.execute(request).await.map_err(map_send_error)?;
+        let result = self.inner.execute(request).await.map_err(map_send_error);
+        #[cfg(google_cloud_unstable_tracing)]
+        if let Some(recorder) = RequestRecorder::current() {
+            match &result {
+                Ok(r) => recorder.on_http_response(r),
+                Err(e) => recorder.on_http_error(e),
```
coryan, line 452: Note that only two places need to call the `recorder.on_http_error()`...  The higher level errors are handled in the `record_http` function.

src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -650,111 +587,52 @@ mod tests {
...
-            .unwrap_or_else(|| panic!("cannot find `test_span` span: {captured:?}"));
-        let attributes = &test.attributes;
-
-        // Should NOT be recorded because the span lacks the necessary marker.
-        assert!(!attributes.contains_key(otel_trace::HTTP_RESPONSE_STATUS_CODE));
-        assert!(!attributes.contains_key(otel_trace::HTTP_REQUEST_RESEND_COUNT));
-    }
-
-    #[tokio::test]
-    async fn test_record_intermediate_client_request_error() {
```
coryan, line 720: This test is no longer relevant, as long as the information is saved in the recorder we are fine.

src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -153,76 +160,6 @@ impl ResultExt for Result<reqwest::Response> {
     }
 }
 
-/// Records HTTP transport attributes on the current span.
-///
-/// This function is used to enrich the Client Request Span (T3) with attributes
-/// from the transport attempt that are only available before the response is consumed.
-pub fn record_intermediate_client_request(
```
coryan, line 160: This is no longer needed, the annotations are stored in the recorder and used in `WithClientSpan`

src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -153,76 +161,6 @@ impl ResultExt for Result<reqwest::Response> {
...
-        final_url
-            .host_str()
-            .map(|h| h.trim_start_matches('[').trim_end_matches(']'))
-            .unwrap_or(""),
-    );
-    span.record(
-        otel_trace::SERVER_PORT,
-        final_url
-            .port_or_known_default()
-            .map(|p| p as i64)
```
coryan, line 199: You may notice a lot of changes for ports from i64 to u64... We used to cast `u16` to `i64` here, now we provide the `u16` directly to the span which converts to `u64`...

westarle: `opentelemetry_tracing` preserves the type of `i64` data [[ref](https://docs.rs/tracing-opentelemetry/latest/src/tracing_opentelemetry/layer.rs.html#283-298)] but doesn't do so for `u64`.  I think it falls back to String [[tracing](https://docs.rs/tracing-core/0.1.36/src/tracing_core/field.rs.html#295-298)][[opentelemetry_tracing](https://docs.rs/tracing-opentelemetry/latest/src/tracing_opentelemetry/layer.rs.html#551-554)].  The OpenTelemetry standard doesn't have an unsigned type [[ref](https://opentelemetry.io/docs/specs/otel/common/#integers)] so they probably just wanted to avoid truncation however unlikely.

All that to say, it's OK to carry the data as any type we like, but the data applied to the span needs to be `i64` for all signed and unsigned integers; if we make the Recorder value `i64` it could be a good place to include the note above in a comment.

tests/o11y/src/e2e/signals.rs
westarle: better remove them, thanks!

src/gax-internal/src/observability/http_tracing.rs
coryan: Ack I will send a PR to fix that first.  Should we do the same with the `u16` for HTTP status code?

coryan: I think I can answer my own question if I read a few lines below: yes.

westarle: Yes, all the numbers need to be traced as int64:

- server.port
- http.response.status_code
- http.request.resend_count

Maybe removing or panicking https://github.com/googleapis/google-cloud-rust/blob/main/src/test-utils/src/test_layer.rs#L60 would help.

coryan: Fixed in https://github.com/googleapis/google-cloud-rust/pull/5180 and will rebase this PR.

src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -650,111 +588,52 @@ mod tests {
...
+
+        let snapshot = recorder.client_snapshot();
+        assert!(snapshot.http_resend_count().is_none(), "{snapshot:?}");
+        assert_eq!(snapshot.http_status_code(), Some(200), "{snapshot:?}");
+        assert_eq!(snapshot.server_address(), "example.com", "{snapshot:?}");
+        assert_eq!(snapshot.server_port(), 443, "{snapshot:?}");
     }
 
     #[tokio::test]
-    async fn test_record_intermediate_client_request_no_marker() {
```
coryan, line 688: This test is also no needed, we no longer depend on the marker to annotate or not annotate the T3 span.

src/gax-internal/src/observability/http_tracing.rs
```diff
@@ -650,111 +588,52 @@ mod tests {
...
-        let url = "https://example.com/test".parse::<reqwest::Url>().unwrap();
+    async fn test_record_intermediate_client_request_error() {
+        let recorder = RequestRecorder::new(TEST_INFO);
+        let request =
+            reqwest::Request::new(Method::GET, "https://example.com/test".parse().unwrap());
+        recorder.on_http_request(&request);
+        // Simulate a 404 error
         let response = http::Response::builder()
-            .status(StatusCode::OK)
+            .status(StatusCode::NOT_FOUND)
```
coryan, line 622: This looks odd, but it is because the test is completely different, they just shared enough code to confuse GitHub.

src/gax-internal/tests/http_observability.rs
```diff
@@ -106,6 +113,7 @@ mod tests {
             (GCP_CLIENT_REPO, "googleapis/google-cloud-rust".into()),
             (GCP_CLIENT_ARTIFACT, TEST_ARTIFACT.into()),
             (GCP_CLIENT_LANGUAGE, "rust".into()),
+            (GCP_RESOURCE_NAME, TEST_RESOURCE.into()),
```
coryan, line 116: New!

src/gax-internal/tests/http_observability.rs
```diff
@@ -355,37 +363,24 @@ mod tests {
...
-            { { otel_trace::HTTP_REQUEST_RESEND_COUNT } } = field::Empty,
-            "gax.client.span" = field::Empty,
-        );
-
-        // Execute the request within the T3 span
-        let _response: Result<Response<TestResponse>> = client
-            .execute(request, None::<NoBody>, options)
-            .instrument(t3_span.clone())
-            .await;
+        let result = generated_tracing_stub(&client).await;
```
coryan, line 372: Note that this test implementation is very different.  I introduced a few functions to "simulate" what the generator should do, using the same helpers the generator would use, and then tested the expected behavior. The prior implementation created spans manually.

src/gax-internal/tests/http_observability.rs
```diff
@@ -524,15 +639,65 @@ mod tests {
         assert_eq!(got, want);
     }
 
+    // Verify the path starting from `execute_http()` records the request and send errors.
```
coryan, line 642: We had no tests for the code paths used by storage, I thought it was time to add them.

tests/o11y/src/http_tracing.rs
```diff
@@ -438,9 +446,7 @@ pub async fn parse_error() -> anyhow::Result<()> {
             "gcp.client.artifact",
             "google-cloud-showcase-v1beta1".into(),
         ),
-        ("gcp.client.language", "rust".into()),
```
coryan, line 441: Note that these are gone... we rely on OpenTelemetry's resource detectors to include the language (they do).

coryan:
PTAL.  Don't be scared by the PR size, it is mostly generated code in showcase.

src/gax-internal/src/observability/http_tracing.rs
coryan: Done.

src/gax-internal/tests/http_observability.rs
```diff
@@ -546,19 +711,93 @@ mod tests {
...
+            recorder.on_client_request(
+                ClientRequestAttributes::default()
+                    .set_rpc_method(TEST_RPC_METHOD)
+                    .set_url_template(TEST_URL_TEMPLATE)
+                    .set_resource_name(TEST_RESOURCE.to_string()),
+            );
+        }
+        let builder = client.http_builder(reqwest::Method::GET, "/test");
+        let attempt_info = AttemptInfo::new(0);
+        builder.send(options, attempt_info).await
```
westarle, line 790: Should we convert response errors to gax errors here?  I think if you had a 404 it would return Ok and fail to trigger the error paths at the client level.

maybe a test like `execute_http_recorder_server_error` would (dis?)prove it.

src/gax-internal/src/http.rs
```diff
@@ -410,26 +421,38 @@ impl ReqwestClient {
...
         crate::observability::propagation::inject_context(
             &tracing::Span::current(),
             request.headers_mut(),
         );
-        let response = self.inner.execute(request).await.map_err(map_send_error)?;
+        let result = self.inner.execute(request).await.map_err(map_send_error);
+        #[cfg(google_cloud_unstable_tracing)]
+        if let Some(recorder) = RequestRecorder::current() {
+            match &result {
+                Ok(r) => recorder.on_http_response(r),
```
westarle, line 451: Nit: I get a mildly worried about matching the on_x_response to the on_x_requests when they are distributed.  I think our tests cover things practicality right now, but I never know what complexity we'll add in the future.

A couple ideas:

- in RequestRecorder have on_http_request return a guard that you must_use to call finish(result)
- Have ClientSnapshot do accounting and put a debug assert in drop

src/gax-internal/tests/http_observability.rs
coryan: > Should we convert response errors to gax errors here?

I do not think so.

> I think if you had a 404 it would return Ok and fail to trigger the error paths at the client level.

That is intentional.... This is simulating the `execute_http()` path, used by Storage. In that case we do not convert HTTP status errors to errors (308 is the most memorable case, but there are others).

> 
> maybe a test like `execute_http_recorder_server_error` would (dis?)prove it.

I think this test is checking how the `ReqwestClient` should work.  I think you caught a bug in my changes for storage (#5171).

src/gax-internal/src/http.rs
coryan: Seems like a good cleanup #5182

src/gax-internal/tests/http_observability.rs
coryan: > I think this test is checking how the `ReqwestClient` should work. I think you caught a bug in my changes for storage (#5171).

I was wrong. Capturing the errors happens in  `WithClient{Span, Metric, Logging}`.

PR #5178, opened 2026-03-28
chore: fix versions for recent libraries

coryan:
A few recent libraries had the wrong initial version and the wrong initial version bump. This bumps all those libraries to 1.0.0 where librarian can start maintaining the numbers correctly.

See https://github.com/googleapis/librarian/issues/4881 for details of why this happens and how we are going to prevent the problem in the future.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1307,7 +1307,7 @@ dependencies = [
 
 [[package]]
 name = "google-cloud-apiregistry-v1"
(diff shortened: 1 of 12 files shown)


coryan:
> just curious: is the bump all manual or did you update the yaml and call some command?

The latter. Update `librarian.yaml` and then call `librarian@${V} generate --all`.

PR #5180, opened 2026-03-28
impl(gax-internal): use `i64` for numeric o11y attributes

coryan:
Some of the o11y attributes are numbers. `opentelemetry_tracing` converts `u64` to strings [[ref]], and it is only too easy to automatically convert`u16` or `u32` to `u64`.

Fortunately all the numbers we want fit in a `i64` which are faithfully preserved by `opentelemetry_tracing`. This change fixes some places I was sloppy and used `u64` when I shouldn't.

More context in the original [this code review comment](https://github.com/googleapis/google-cloud-rust/pull/5176/changes#r3004829460).

[ref]:
https://docs.rs/tracing-opentelemetry/latest/src/tracing_opentelemetry/layer.rs.html#283-298

### DIFF
--- a/src/gax-internal/src/observability/client_signals/recorder.rs
+++ b/src/gax-internal/src/observability/client_signals/recorder.rs
@@ -335,7 +335,7 @@ impl ClientSnapshot {
     ///
     /// If no port is known, use the port implied by `info.default_host`.
     ///
-    /// Use with the "server.port" attribute.
+    /// Use with the "server.port" attribute after casting to `i64`.
     pub fn server_port(&self) -> u16 {
         if let Some(port) = self
             .transport_snapshot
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/observability/client_signals/with_client_span.rs
```diff
@@ -259,7 +259,7 @@ mod tests {
         assert_eq!(get_attr("error.type"), Some(&Value::String("404".into())));
         assert_eq!(
             get_attr("http.response.status_code"),
-            Some(&Value::String("404".into()))
+            Some(&Value::I64(404))
```
westarle, line 262: 🎉

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -332,6 +334,8 @@ impl ClientSnapshot {
     /// Returns the server port used in the last low-level request.
     ///
     /// If no port is known, use the port implied by `info.default_host`.
+    ///
+    /// Use with the "server.port" attribute.
```
westarle: nit
```suggestion
    /// Use with the "server.port" attribute after casting to i64.
```

coryan: Done, and in the other numeric cases too.

PR #5181, opened 2026-03-29
feat(spanner): batch read-only transactions

olavloite:
This is step 1 in adding support for batch read-only transactions to the Spanner client. Batch read-only transactions are a specialized sort of multi-use read-only transactions that allow you to partition queries or reads, and then execute each partition separately. The partitions can be distributed to multiple processes (including processes on a separate machine) and executed in parallel. This can reduce the time that is needed to for example bulk-extract a large amount of data.

This change adds support for creating a BatchReadOnlyTransaction and partitioning queries and read operations. Follow-up pull requests will add support for executing partitions and distributing partitions to other hosts.

### DIFF
--- a/src/spanner/src/batch_read_only_transaction.rs
+++ b/src/spanner/src/batch_read_only_transaction.rs
@@ -1,0 +1,439 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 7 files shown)


src/spanner/src/batch_read_only_transaction.rs
```diff
@@ -0,0 +1,439 @@
...
+                    session_name: self.inner.context.client.session.name.clone(),
+                    read_request: read.clone(),
+                },
+            })
+            .collect())
+    }
+}
+
+/// Defines the segments of data to be read in a partitioned read or query.
+/// These partitions can be serialized and processed across several
```
coryan, line 232: I know this is mostly a placeholder at the moment, but eventually applications will need to serialize and deserialize these (otherwise, how do they send them to another process).  When you do, consider using `serde` as the serialization library.

PR #5183, opened 2026-03-29
impl(storage): add spans to `ObjectDescriptor::read_range`

### LINKED ISSUE #3178: Tracing for `Storage` client

coryan:
An object descriptor may be used in multiple `read_range()` operations.
Application developers will be interested in the performance of these
calls.

Fixes #3178

### DIFF
--- a/tests/o11y/src/e2e/storage.rs
+++ b/tests/o11y/src/e2e/storage.rs
@@ -156,6 +156,15 @@ async fn storage_data_operations(bucket_name: &str) -> anyhow::Result<()> {
         tracing::info!("reading from open object");
         let mut response = descriptor.read_range(ReadRange::head(16)).await;
         while response.next().await.transpose()?.is_some() {}
+        {
+            let _test_span = tracing::info_span!("test-span").entered();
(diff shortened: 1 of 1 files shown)


coryan:
> do you have a link with the resulting spans? I'm curious about their relationship to CLIENT spans.

Look at the `read_range` spans in:

https://console.cloud.google.com/traces/explorer;traceId=adcf64362227220daae829ebdb6d93d4?project=coryan-test

They are unconnected to the client span... That might be a good improvement for the future: @joshuatants FYI

westarle:
> They are unconnected to the client span... That might be a good improvement for the future: @joshuatants FYI

Let's chat about it sometime, I think it is a good thing that they aren't hierarchically related for now.

PR #5193, opened 2026-03-30
chore(auth): bump sha2 to 0.11.0 and hmac to 0.13.0

PhongChuong:


src/auth/src/credentials/external_account_sources/aws_sourced.rs
```diff
@@ -21,7 +21,7 @@ use crate::{
 };
 use chrono::Utc;
 use google_cloud_gax::error::CredentialsError;
-use hmac::{Hmac, Mac};
+use hmac::{Hmac, KeyInit, Mac};
```
alvarowolfx, line 24: why is this `KeyInit` being added ? seems like a unused import, but sounds like it was required due to the upgrade.

PhongChuong: `KeyInit` is a trait which can only be used if it's in scope.

[Here's](https://github.com/RustCrypto/traits/blob/1c96506e27d9e6bb8531f7c19b06112597692d02/digest/src/buffer_macros/fixed.rs#L375) where it's needed if you're curious.

coryan: Looks like `KeyInit` is a new trait:

https://docs.rs/hmac/0.12.1/hmac/#traits
https://docs.rs/hmac/0.13.0/hmac/#traits

It needs to be in scope to be used, which probably required the `use` declaration.

PR #5194, opened 2026-03-30
feat(spanner): execute a batch read-only transaction partition

olavloite:
Adds a method for executing a Partition that has previously been returned by a partition_read or partition_query operation. A Partition can be executed using any DatabaseClient that is connected to the same database.

### DIFF
--- a/src/spanner/src/batch_read_only_transaction.rs
+++ b/src/spanner/src/batch_read_only_transaction.rs
@@ -241,9 +241,7 @@ impl Partition {
     /// Executes this partition and returns a [ResultSet] that
     /// contains the rows that belong to this partition.
     ///
-    /// # Example
-    ///
-    /// Executing a query partition:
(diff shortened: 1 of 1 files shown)


src/spanner/src/batch_read_only_transaction.rs
```diff
@@ -241,17 +241,149 @@ impl Partition {
     /// Executes this partition and returns a [ResultSet] that
     /// contains the rows that belong to this partition.
     ///
-    /// # Panics
+    /// # Example
+    ///
+    /// Executing a query partition:
```
coryan: Consider:


```suggestion
    /// # Example: executing a query partition
```

And a similar name before.  That would probably render as two different sections and somewhat easier to grok for customers.

olavloite: Thanks, done.

PR #5198, opened 2026-03-30
fix(o11y): align `server.address` with OTel specs to prevent high cardinality

### LINKED ISSUE #5195: `server.address` should identify the endpoint used
In our current implementation, we calculate `server.address` from the actual network peer.  This can lead to unexpected cardinality, which the standards seem to reserve for `network.peer.address`.

It should track the actual request endpoint configured (the host or authority in the endpoint).  This can be a DNS name or IP, whatever is specified when the request is made.

This should more closely match our customers' expectations based on the [OTel standards](https://opentelemetry.io/docs/specs/semconv/dotnet/dotnet-http-metrics/#metric-httpclientrequestduration) where they might not expect the cardinality on this attribute (and it is the only place to find this information in metrics, since they lack URL information.)

haphungw:
In the current implementation, `server.address` was being populated with the globally resolved socket IP (network peer address). This led to high metric cardinality and violated OpenTelemetry standards.

This PR decouples the configured URL endpoint from the resolved TCP address.

Fixes #5195


src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -646,18 +649,6 @@ mod tests {
...
-            let mut guard = recorder.inner.lock().expect("never poisoned");
-            let s = guard.transport_snapshot.as_mut().expect("already set");
-            s.server_address = Some(SocketAddr::V4(SocketAddrV4::new(
-                Ipv4Addr::new(127, 0, 0, 234),
-                234,
-            )));
-        }
-        let snap = recorder.client_snapshot();
-        assert_eq!(snap.server_address(), "127.0.0.234", "{snap:?}");
-        assert_eq!(snap.server_port(), 234, "{snap:?}");
```
coryan, line 659: We still want a version of this test for the new `network_peer_address_*` functions?

src/gax-internal/src/observability/client_signals/duration_metric.rs
```diff
@@ -119,6 +120,14 @@ impl DurationMetric {
...
             (SERVER_ADDRESS, Some(snapshot.server_address().into())),
             (SERVER_PORT, Some((snapshot.server_port() as i64).into())),
+            (
+                NETWORK_PEER_ADDRESS,
+                snapshot.network_peer_address().map(|v| v.into()),
+            ),
+            (
+                NETWORK_PEER_PORT,
+                snapshot.network_peer_port().map(|v| v.into()),
+            ),
```
coryan, line 129: I thought the point was to keep these out of the metrics because their cardinality was too high?

src/gax-internal/src/observability/client_signals/with_client_metric.rs
```diff
@@ -201,8 +201,13 @@ mod tests {
                 ("url.template", TEST_URL_TEMPLATE),
                 ("http.response.status_code", "404"),
                 ("error.type", "404"),
-                ("server.address", server.addr().ip().to_string().as_str()),
+                ("server.address", "127.0.0.1"),
                 ("server.port", format!("{}", server.addr().port()).as_str()),
+                ("network.peer.address", "127.0.0.1"),
+                (
+                    "network.peer.port",
+                    format!("{}", server.addr().port()).as_str(),
```
coryan: optional: does this work? is it any better?


```suggestion
                    server.addr().port().to_string().as_str(),
```

src/gax-internal/src/observability/client_signals/duration_metric.rs
haphungw: makes sense. I removed these from `duration_metrics` and now they only exist on the trace spans.

PR #5208, opened 2026-03-30
fix(o11y): remove redundant `rpc.service` attribute and ensure `rpc.method` is present

haphungw:
For https://github.com/googleapis/google-cloud-rust/issues/4924


src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -566,7 +564,6 @@ mod tests {
                 crate::observability::attributes::RPC_SYSTEM_GRPC.into(),
             ),
             (OTEL_KIND, "Client".into()),
-            (otel_trace::RPC_SERVICE, "google.pubsub.v1.Publisher".into()),
             (otel_trace::RPC_METHOD, "Publish".into()),
```
westarle: they reused the rpc.method field 🙄 so this should be "google.pubsub.v1.Publisher/Publish"

src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -516,7 +515,6 @@ mod tests {
...
```
westarle: see below

src/gax-internal/src/observability/grpc_tracing.rs
```diff
@@ -375,7 +375,7 @@ fn create_grpc_span(
     attempt_count: Option<i64>,
     resource_name: Option<&str>,
 ) -> tracing::Span {
-    let (rpc_service, rpc_method) =
+    let (_, rpc_method) =
```
westarle: I think the rpc.method should be uri.path() now, minus a / at the start.  See my test case below.

src/gax-internal/src/observability/grpc_tracing.rs
haphungw: done

src/gax-internal/src/observability/grpc_tracing.rs
haphungw: done

PR #5209, opened 2026-03-30
cleanup(gax-internal): macros use literal attribute names

coryan:
By using the literal attribute names we can reduce the number of constants exported by `google-cloud-gax-internal`. And therefore reduce the number of future breaking changes.

Part of the work for #4772

### DIFF
--- a/src/gax-internal/src/observability/client_signals.rs
+++ b/src/gax-internal/src/observability/client_signals.rs
@@ -84,11 +84,11 @@ macro_rules! client_request_signals {
         // We use string literals for all the field names because it narrows the public API for
         // `google-cloud-gax-internal`. The exception are these values, which we expect may change
         // from time to time.
-        use $crate::observability::internal::{GCP_CLIENT_REPO_GOOGLEAPIS, OTEL_KIND_INTERNAL, SCHEMA_URL_VALUE};
+        use $crate::observability::{GCP_CLIENT_REPO_GOOGLEAPIS, SCHEMA_URL_VALUE};
         tracing::info_span!(
             "client_request",
              "otel.name"             = concat!(env!("CARGO_CRATE_NAME"), "::", $method),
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/observability/client_signals.rs
```diff
@@ -109,6 +79,38 @@ macro_rules! client_request_signals {
             .instrument(span.clone());
         (span, pending)
     }};
+    (info: $info:expr, method: $method:literal) => {{
+        use ::tracing::field::Empty;
+        // We use string literals for all the field names because it narrows the public API for
+        // `google-cloud-gax-internal`. The exception are these values, which we expect may change
+        // from time to time.
+        use $crate::observability::internal::{GCP_CLIENT_REPO_GOOGLEAPIS, OTEL_KIND_INTERNAL, SCHEMA_URL_VALUE};
```
westarle: Nit: Why include "internal" this list?  I thought it's pretty baked into the standard.

https://opentelemetry.io/docs/concepts/signals/traces/#span-kind and into https://docs.rs/tracing-opentelemetry/latest/src/tracing_opentelemetry/layer.rs.html#174-183

coryan: It might be part of the standard.  I want to minimize the visibility of these symbols.  But now that you mention it, this is the google-cloud-gax-**internal**.  I will remove the extra module.

westarle: oops -- I meant I think we could just put "internal" instead of OTEL_KIND_INTERNAL.

westarle: maybe "Internal" for consistency

coryan: Got it, I implemented both your suggestion and the thing about the extra `internal` module.

PR #5210, opened 2026-03-31
impl(gax-internal): `server_address()` with IPv6

coryan:
The code was leaving the brackets for IPv6 `server_address()`, as they are extracted from the Uri. Fixing that just requires some trimming of the extra characters.

The tests did not pass because they hard-coded 127.0.0.1.

TIL: my laptop has an IPv6 enabled. Or maybe I knew and Today I Remembered.


haphungw:
looks great, thanks!

PR #5211, opened 2026-03-31
feat(spanner): request and transaction tags

olavloite:
Adds support for setting request and transaction tags.

### DIFF
--- a/src/spanner/src/batch_dml.rs
+++ b/src/spanner/src/batch_dml.rs
@@ -38,9 +38,24 @@ impl BatchDmlBuilder {
         self
     }
 
-    /// Specifies the request options for this batch calculation.
-    pub fn with_request_options(mut self, request_options: RequestOptions) -> Self {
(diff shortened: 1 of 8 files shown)


src/spanner/src/batch_dml.rs
```diff
@@ -38,9 +38,24 @@ impl BatchDmlBuilder {
         self
     }
 
-    /// Specifies the request options for this batch calculation.
-    pub fn with_request_options(mut self, request_options: RequestOptions) -> Self {
```
olavloite, line 42: This builder by accident exposed a method for setting `crate::model::RequestOptions` directly. We do not want that, as not all `RequestOptions` are relevant/valid in all cases. Instead, we want to gate that through specific setters, such as `with_transaction_tag`.

coryan: SGTM. We do something similar in storage where a single `struct` is used for all builders, but the builders only expose the methods that make sense...

PR #5214, opened 2026-03-31
impl(cloud-run-o11y): observability demo

coryan:


coryan:
Ping

haphungw:
Quick question about the audience for this demo. Is this for us to see how to wire up o11y, or are we planning this as a public sample for external users?

The implementation relies on `integration_tests_o11y` which is not accessible for external users, so technically they can't copy paste this demo code right?

coryan:
> Quick question about the audience for this demo. Is this for us to see how to wire up o11y, or are we planning this as a public sample for external users?

It is for us, and it show us how difficult / easy it is to wire up o11y. We can also use it to generate cool screenshots.

> The implementation relies on `integration_tests_o11y` which is not accessible for external users, so technically they can't copy paste this demo code right?

They can't... but they could cut&paste that code too. nudge nudge, wink wink. Though #5223 may be a better approach for that the things in `integration-tests-o11y` do.

PR #5227, opened 2026-04-01
fix(pubsub): avoid ack batch size limit errors

### LINKED ISSUE #5222: Reduce max ack IDs per batch
We batch up to 2500 ack IDs in an `AcknowledgeRequest` / `ModifyAckDeadlineRequest`. Apparently this is not conservative enough, as @suzmue had a request fail for being too big.

We should lower the value to 1000.

dbolduc:
Fixes #5222

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -32,10 +32,6 @@ use tokio_util::task::TaskTracker;
 // on the size of an ack ID. We can safely fit 1000 ack IDs into a request.
 //
 // https://docs.cloud.google.com/pubsub/quotas
-//
-// Note that other languages use this same value.
-//
-// https://github.com/googleapis/java-pubsub/blob/876aed861576c370fcb4dd2c0e9440646b5cdc04/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java#L88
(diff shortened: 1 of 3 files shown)


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -28,11 +28,15 @@ use tokio::task::JoinSet;
...
-// be safe to fit 2500 Ack IDs in a single RPC.
+// Request sizes are limited to 512kB. 500 bytes is a conservative upper limit
+// on the size of an ack ID. We can safely fit 1000 ack IDs into a request.
 //
 // https://docs.cloud.google.com/pubsub/quotas
-const MAX_IDS_PER_RPC: usize = 2500;
+//
+// Note that other languages use this same value.
+//
+// https://github.com/googleapis/java-pubsub/blob/876aed861576c370fcb4dd2c0e9440646b5cdc04/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StreamingSubscriberConnection.java#L88
```
PhongChuong: nit. I don't think this comment is needed.

PhongChuong (approved):

src/pubsub/src/subscriber/lease_state.rs
dbolduc: True. Removed.

PR #5237, opened 2026-04-01
feat(o11y): implement RequestRecorder hooks for gRPC execution

### LINKED ISSUE #5232: Missing `rpc.system.name` on gRPC client metrics and logs
`grpc::Client` does not explicity hook into `RequestRecorder`. While the http implementation uses `recorder.on_http_request(&request)` to map base OpenTelemetry attributes, the gRPC client does not have an equivalent call.

We need to implement `on_grpc_request` and `on_grpc_error` equivalents in `RequestRecorder`.

Identified in https://github.com/googleapis/google-cloud-rust/pull/5221.

haphungw:
Previously, the gRPC execution layer did not explicitly hook into the RequestRecorder to populate transport-level data before issuing a request. As a result, `rpc.system.name` and explicit URL boundaries were missing from gRPC duration metrics and logs.

This PR maps the gRPC execution paths directly into the telemetry bucket.

Fixes #5232

### DIFF
--- a/src/storage/src/storage/transport.rs
+++ b/src/storage/src/storage/transport.rs
@@ -410,7 +410,7 @@ mod tests {
         check_debug_log(&captured, "read_object");
 
         #[cfg(google_cloud_unstable_tracing)]
-        client_request_span(&captured, "read_object", "404");
+        client_request_span(&captured, "read_object", "404", "http");
 
         Ok(())
     }
@@ -497,7 +497,7 @@ mod tests {
         check_debug_log(&captured, "write_object_buffered");
 
         #[cfg(google_cloud_unstable_tracing)]
-        client_request_span(&captured, "write_object", "404");
+        client_request_span(&captured, "write_object", "404", "http");
 
         Ok(())
     }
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -390,7 +420,7 @@ impl ClientSnapshot {
     pub fn http_method(&self) -> Option<&str> {
         self.transport_snapshot
             .as_ref()
-            .and_then(|s| s.http_method.as_ref().map(|m| m.as_str()))
+            .and_then(|s| s.http_method.as_ref().map(|m: &http::Method| m.as_str()))
```
coryan, line 423: Is that needed? The compiler knows the type, and the intent is obvious-ish?

coryan (approved):
The changes LGTM, but seems like you need to rebase and resolve some conflicts.

src/gax-internal/src/observability/client_signals/recorder.rs
haphungw:
```
#[cfg(feature = "_internal-http-client")]
use crate::observability::attributes::RPC_SYSTEM_HTTP;
```
it is needed because i added this check.

Issue #3840, opened 2025-11-18
Add a subscriber throughput benchmark


dbolduc:
We should add a throughput benchmark for the subscriber client, similar to https://cloud.google.com/blog/products/data-analytics/testing-cloud-pubsub-clients-to-maximize-streaming-performance

The benchmark should pull messages from a subscription and ack them, across N tasks.

(At least) the following should be configurable:
- test iterations
- test duration
- num clients
- num tasks
- num worker threads
- outstanding messages / bytes
- ack/nack flush duration

The benchmark should report
- the ack count for each iteration
- some sort of error rate
- (stretch) deadline expirations / message redeliveries
- (stretch) stream restarts?

We need to run this thing for a while (>1 day) and report the results here.

The target throughput is around 600 MB/s on a 16 vCPU VM. I am not sure about the target error rate / redelivery rate.

---

I have a **rough** sketch of such a benchmark: https://github.com/dbolduc/google-cloud-rust/tree/1b01c325ca05f5ded4d60f0948d0cff63993ab02

dbolduc:
Here is an updated version of the sketch, with a proper client surface and somewhat-proper internals:

https://github.com/dbolduc/google-cloud-rust/commit/81271cce46e6c376f9552a706105adde8ec9d949

dbolduc:
@suzmue did this

PR #5238, opened 2026-04-01
test(pubsub): add throughput benchmark

suzmue:
Add a new throughput benchmark for measuring the performance of publish and subscribe operations. The benchmark is implemented as a single binary with subcommands for the publisher and subscriber, allowing for unified metric formatting and simplified execution.

A version of this was used to benchmark our libraries to validate acceptable performance.

For #3840

### DIFF
--- a/src/pubsub/benchmarks/throughput/README.md
+++ b/src/pubsub/benchmarks/throughput/README.md
@@ -12,37 +12,11 @@ rates in messages per second and megabytes per second.
 cargo run --release -p pubsub-throughput -- [COMMAND] [OPTIONS]
 ```
 
-### Commands
+To see the commands and options use:
 
-- `publisher`: Measure publishing throughput.
-- `subscriber`: Measure subscribing throughput.
-
-### Common Options (Global)
-
-- `--project`: The Google Cloud project ID.
-- `--report-interval`: The interval between progress reports (e.g., `5s`, `1m`;
-  default: `5s`).
-- `--duration`: The total duration to run the benchmark (e.g., `5m`, `1h`;
-  default: `5m`).
-- `--grpc-channels`: The number of gRPC channels to use (default: `1`).
-- `--max-outstanding-messages`: The maximum number of unacknowledged messages
-  held in memory (default: `100000`).
-
-### Publisher Options
-
-- `--topic-id`: The ID of the Pub/Sub topic to publish to (Required).
-- `--payload-size`: The size of each message payload in bytes (default: `1024`).
-- `--batch-size`: The maximum number of messages in a batch (default: `1000`).
-- `--batch-bytes`: The maximum size of a batch in bytes (default: `10485760`
-  which is 10 MB).
-- `--batch-delay`: The maximum time to wait before sending a batch (default:
-  `100ms`).
-
-### Subscriber Options
-
-- `--subscription-id`: The ID of the Pub/Sub subscription to receive from
-  (Required).
-- `--streams`: The number of subscriber streams to run (default: `1`).
+```bash
+cargo run -p pubsub-throughput -- --help
+```
 
 ## Output Format
 
--- a/src/pubsub/benchmarks/throughput/Cargo.toml
+++ b/src/pubsub/benchmarks/throughput/Cargo.toml
@@ -28,7 +28,7 @@ workspace = true
 [dependencies]
 anyhow.workspace    = true
 bytes.workspace     = true
-clap                = { workspace = true, features = ["derive", "env", "std"] }
+clap                = { workspace = true, features = ["derive", "env", "help", "std"] }
 google-cloud-pubsub = { workspace = true, features = ["default-rustls-provider"] }
 futures.workspace   = true
 humantime.workspace = true


suzmue:
I can split this into further if requested.

src/pubsub/benchmarks/throughput/src/args.rs
```diff
@@ -0,0 +1,86 @@
...
+pub enum Commands {
+    /// Run the publisher benchmark
+    Publisher(PublisherArgs),
+    /// Run the subscriber benchmark
+    Subscriber(SubscriberArgs),
+}
+
+#[derive(Args, Debug, Clone)]
+pub struct CommonArgs {
+    #[arg(long, default_value = "", env = "GOOGLE_CLOUD_PROJECT")]
```
dbolduc, line 35: TIL: `env`. That's cool.

src/pubsub/benchmarks/throughput/src/main.rs
```diff
@@ -0,0 +1,87 @@
...
+        }
+    }
+
+    Ok(())
+}
+
+/// Returns true if the benchmark has exceeded its maximum allowed runtime.
+pub(crate) fn done(maximum_runtime: Duration, start: Instant) -> bool {
+    let now = Instant::now();
+    now >= start + maximum_runtime
```
dbolduc: nit:

```suggestion
    start.elapsed() >= maximum_runtime
```

src/pubsub/benchmarks/throughput/src/publisher.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Thread-safe counters for tracking publisher performance metrics.
+#[derive(Default)]
+struct Stats {
+    /// Total messages sent to the client library.
+    send_count: AtomicI64,
+    /// Total bytes sent to the client library.
+    send_bytes: AtomicI64,
+    /// Total messages successfully acknowledged by the Pub/Sub service.
+    ack_count: AtomicI64,
+    /// Total bytes successfully acknowledged by the Pub/Sub service.
```
dbolduc: optional nit: what about `received_count`, `received_bytes` ?

But maybe I am just stuck in subscriber mode where `ack_count` means something else.

src/pubsub/benchmarks/throughput/src/publisher.rs
```diff
@@ -0,0 +1,149 @@
...
+/// Creates a Publisher client configured with the benchmark settings.
+async fn create_publisher(config: Arc<PublisherArgs>, topic_name: String) -> Publisher {
+    Publisher::builder(topic_name)
+        .set_byte_threshold(config.batch_bytes)
+        .set_message_count_threshold(config.batch_size)
+        .set_delay_threshold(config.batch_delay)
+        .with_grpc_subchannel_count(config.common.grpc_channels)
+        .build()
+        .await
+        .unwrap()
```
dbolduc: nit: prefer to use `anyhow::Result<>` + `?`

src/pubsub/benchmarks/throughput/src/publisher.rs
```diff
@@ -0,0 +1,149 @@
...
+        config.common.max_outstanding_messages,
+    ));
+    let stats = Arc::new(Stats::default());
+
+    // Start a background task to publish messages.
+    let publisher_stats = stats.clone();
+    let publisher_handle = tokio::task::spawn(async move {
+        loop {
+            // Respect the max_outstanding_messages limit.
+            let permit = semaphore.clone().acquire_owned().await.unwrap();
```
dbolduc: nit: same with `?`

src/pubsub/benchmarks/throughput/src/publisher.rs
dbolduc: You should feel 0 obligation to change this.

src/pubsub/benchmarks/throughput/src/subscriber.rs
```diff
@@ -0,0 +1,142 @@
...
+/// Orchestrates the subscriber benchmark by spawning multiple processing streams.
+async fn run_subscriber(
+    config: Arc<SubscriberArgs>,
+    subscriber: Subscriber,
+    subscription_name: &str,
+) {
+    let stats = Arc::new(Stats::default());
+    let mut tasks = Vec::new();
+    let max_outstanding_per_task = config.common.max_outstanding_messages / config.streams;
+    if max_outstanding_per_task > 0 {
```
dbolduc: maybe just `assert!(max_outstanding_per_task > 0);` ?

Or remove the line altogether, because `0` means "no limit"

src/pubsub/benchmarks/throughput/README.md
```diff
@@ -0,0 +1,83 @@
...
+  which is 10 MB).
+- `--batch-delay`: The maximum time to wait before sending a batch (default:
+  `100ms`).
+
+### Subscriber Options
+
+- `--subscription-id`: The ID of the Pub/Sub subscription to receive from
+  (Required).
+- `--streams`: The number of subscriber streams to run (default: `1`).
+
```
dbolduc: Q: can `clap` do this stuff? That seems preferred, because the markdown can stray from the code.

It seems like we can put comments on the fields, and then a `--help` prints them out (somehow)

https://docs.rs/clap/latest/clap/#example

Then this section is just like "To see the configuration options, run the benchmark with ..."

dbolduc: Non-blocking for this PR. Especially if the subcommand tree makes things difficult

src/pubsub/benchmarks/throughput/README.md
```diff
@@ -0,0 +1,83 @@
...
+- `iteration`: The current iteration number.
+- `count`: The number of messages processed in the operation.
+- `msgs/s`: The number of messages per second.
+- `bytes`: The total number of bytes processed.
+- `MB/s`: The throughput in megabytes per second.
+- `errors`: The number of errors encountered (if any).
+- `errors/s`: The number of errors per second (if any).
+
+## Examples
+
```
dbolduc, line 37: optional: could have `gcloud` commands to create/delete the topic/subscription

But I am also fine with not having this. It is easily figured out.

src/pubsub/benchmarks/throughput/src/subscriber.rs
```diff
@@ -0,0 +1,142 @@
...
+    let subscription_name = format!(
+        "projects/{}/subscriptions/{}",
+        config.common.project, config.subscription_id
+    );
+
+    let subscriber = Subscriber::builder()
+        .with_grpc_subchannel_count(config.common.grpc_channels)
+        .build()
+        .await
+        .unwrap();
```
dbolduc: nit: `?`

src/pubsub/benchmarks/throughput/src/main.rs
suzmue: Done.

src/pubsub/benchmarks/throughput/src/publisher.rs
suzmue: Done.

src/pubsub/benchmarks/throughput/src/publisher.rs
suzmue: Done. Actually I just moved this into run() to mathch subscriber.

src/pubsub/benchmarks/throughput/src/publisher.rs
suzmue: This is slightly harder since its in a spawned task and we would then need to extract the result which I dont think is worth it for now. So I am updating this to log and break instead.

src/pubsub/benchmarks/throughput/src/subscriber.rs
suzmue: Done.

src/pubsub/benchmarks/throughput/src/subscriber.rs
suzmue: Done.

src/pubsub/benchmarks/throughput/README.md
suzmue: Done.

src/pubsub/benchmarks/throughput/README.md
suzmue: I would think so. I'll maybe send a follow up because I don't know how off the top of my head.

suzmue: Oh its easy I just have to enable the right feature. Will add to this PR.

suzmue: Done :) Thanks for the great suggestion.

PR #5242, opened 2026-04-02
impl(auth): add universe_domain resolution helper

alvarowolfx:

### DIFF
--- a/src/auth/src/universe_domain.rs
+++ b/src/auth/src/universe_domain.rs
@@ -14,11 +14,14 @@
 
 use crate::constants::DEFAULT_UNIVERSE_DOMAIN;
 use crate::credentials::Credentials;
-use crate::errors::CredentialsError;
 
+/// Returns `true` if the given universe domain is the Default Google Universe (GDU).
+///
+/// This serves as a feature gate for capabilities that are only supported in the GDU
+/// (e.g., `googleapis.com`). For example, Regional Access Boundaries should be disabled,
+/// and User Account credentials should return an error when running outside the GDU.
 #[allow(dead_code)]
-pub(crate) fn is_default_universe_domain<S: Into<String>>(universe_domain: Option<S>) -> bool {
-    let universe_domain = universe_domain.map(|s| s.into());
+pub(crate) fn is_default_universe_domain(universe_domain: Option<String>) -> bool {
     match universe_domain {
         Some(ud) => ud == DEFAULT_UNIVERSE_DOMAIN,
         None => true,
@@ -26,12 +29,12 @@ pub(crate) fn is_default_universe_domain<S: Into<String>>(universe_domain: Optio
 }
 
 #[allow(dead_code)]
-pub(crate) async fn resolve(cred: &Credentials) -> Result<String, CredentialsError> {
+pub(crate) async fn resolve(cred: &Credentials) -> String {
     let cred_universe = cred.universe_domain().await;
-    Ok(cred_universe
+    cred_universe
         .as_deref()
         .unwrap_or(DEFAULT_UNIVERSE_DOMAIN)
-        .to_string())
+        .to_string()
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


src/auth/src/universe_domain.rs
```diff
@@ -0,0 +1,79 @@
...
+
+use crate::constants::DEFAULT_UNIVERSE_DOMAIN;
+use crate::credentials::Credentials;
+use crate::errors::CredentialsError;
+
+#[allow(dead_code)]
+pub(crate) fn is_default_universe_domain<S: Into<String>>(universe_domain: Option<S>) -> bool {
+    let universe_domain = universe_domain.map(|s| s.into());
+    match universe_domain {
+        Some(ud) => ud == DEFAULT_UNIVERSE_DOMAIN,
```
dbolduc, line 26: comment: I could imagine us wanting to distinguish between an application not setting a universe domain and explicitly asking for the GDU.

I'll wait to see how we use this function though.

src/auth/src/universe_domain.rs
```diff
@@ -0,0 +1,79 @@
...
+            async fn universe_domain(&self) -> Option<String>;
+        }
+    }
+
+    #[tokio::test]
+    async fn test_resolve_default() {
+        let mut mock = MockCredentials::new();
+        mock.expect_universe_domain().return_const(None);
+        let cred = Credentials::from(mock);
+        let result = resolve(&cred).await.unwrap();
```
dbolduc: nit: I think we more commonly have the test return a `anyhow::Result<()>` and use `?` instead of `unwrap()`

EDIT: nvm if we change the signature

src/auth/src/universe_domain.rs
```diff
@@ -0,0 +1,79 @@
...
+    match universe_domain {
+        Some(ud) => ud == DEFAULT_UNIVERSE_DOMAIN,
+        None => true,
+    }
+}
+
+#[allow(dead_code)]
+pub(crate) async fn resolve(cred: &Credentials) -> Result<String, CredentialsError> {
+    let cred_universe = cred.universe_domain().await;
+    Ok(cred_universe
```
dbolduc: This thing only ever returns `Ok(...)`. Why is it a `Result` ? Will it grow?

src/auth/src/universe_domain.rs
```diff
@@ -0,0 +1,79 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::constants::DEFAULT_UNIVERSE_DOMAIN;
+use crate::credentials::Credentials;
+use crate::errors::CredentialsError;
+
+#[allow(dead_code)]
+pub(crate) fn is_default_universe_domain<S: Into<String>>(universe_domain: Option<S>) -> bool {
+    let universe_domain = universe_domain.map(|s| s.into());
```
dbolduc: nit: if the input to this is coming from the builder, then `S` should already be a `String`.

If the templating is just to accept a `&str` in the unit tests, I don't think that's worth it.

src/auth/src/universe_domain.rs
alvarowolfx: This is mostly to serve as a gate to some features that should be enabled or not if is running outside GDU. Access Boundaries for example needs to be disable. User account should also throw an error. That's the main use case for this function. Will add a comment

src/auth/src/universe_domain.rs
alvarowolfx: moved to `String`, indeed we are only gonna use it from data coming from Builders.

src/auth/src/universe_domain.rs
alvarowolfx: oversight from my end, I forgot that `universe_domain` don't return an error.

src/auth/src/universe_domain.rs
dbolduc: Ah. Thanks!

PR #5243, opened 2026-04-02
impl(gax-internal): add universe_domain mod helper 

alvarowolfx:

### DIFF
--- a/src/gax-internal/src/universe_domain.rs
+++ b/src/gax-internal/src/universe_domain.rs
@@ -26,26 +26,23 @@ pub(crate) async fn resolve(
 ) -> Result<String, CredentialsError> {
     let env_universe = std::env::var(UNIVERSE_DOMAIN_VAR).ok();
     let cred_universe = cred.universe_domain().await;
-
-    let universe_domain = env_universe
+    let cred_universe = cred_universe.as_deref().unwrap_or(DEFAULT_UNIVERSE_DOMAIN);
+    let client_universe = env_universe
         .as_deref()
         .or(universe_domain_client_override)
-        .unwrap_or(DEFAULT_UNIVERSE_DOMAIN)
-        .to_string();
+        .unwrap_or(DEFAULT_UNIVERSE_DOMAIN);
 
-    let cred_universe = cred_universe.as_deref().unwrap_or(DEFAULT_UNIVERSE_DOMAIN);
-
-    if universe_domain != cred_universe {
+    if !cred_universe.eq(client_universe) {
         return Err(CredentialsError::from_msg(
             false,
             format!(
                 "The configured universe domain ({}) does not match the universe domain found in the credentials ({}). If you haven't configured the universe domain explicitly, `googleapis.com` is the default.",
-                universe_domain, cred_universe
+                client_universe, cred_universe
             ),
         ));
     }
 
-    Ok(universe_domain)
+    Ok(client_universe.to_string())
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/universe_domain.rs
```diff
@@ -0,0 +1,110 @@
...
+        let mut provider = MockCredentials::new();
+        let cred_domain = cred_domain.map(|s| s.to_string());
+        provider
+            .expect_universe_domain()
+            .returning(move || cred_domain.clone());
+        let cred = Credentials::from(provider);
+
+        let universe_domain = resolve(client_override, &cred).await;
+        let expected = expected.map(|s| s.to_string());
+        match (universe_domain, expected) {
```
dbolduc: It will be cleaner to split this into two tests: one with all the success cases, and one with all the error cases.

src/gax-internal/src/universe_domain.rs
```diff
@@ -0,0 +1,110 @@
...
+
+    let universe_domain = env_universe
+        .as_deref()
+        .or(universe_domain_client_override)
+        .unwrap_or(DEFAULT_UNIVERSE_DOMAIN)
+        .to_string();
+
+    let cred_universe = cred_universe.as_deref().unwrap_or(DEFAULT_UNIVERSE_DOMAIN);
+
+    if universe_domain != cred_universe {
```
dbolduc: Consider:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=455177e8ce756466b58187ed50f03d46

src/gax-internal/src/universe_domain.rs
```diff
@@ -0,0 +1,110 @@
...
+    let env_universe = std::env::var(UNIVERSE_DOMAIN_VAR).ok();
+    let cred_universe = cred.universe_domain().await;
+
+    let universe_domain = env_universe
+        .as_deref()
+        .or(universe_domain_client_override)
+        .unwrap_or(DEFAULT_UNIVERSE_DOMAIN)
+        .to_string();
+
+    let cred_universe = cred_universe.as_deref().unwrap_or(DEFAULT_UNIVERSE_DOMAIN);
```
dbolduc: So if my creds have no universe, and my client has a non-GDU universe, that is a fail. Should it be?

(There is not a test case for this)

src/gax-internal/src/universe_domain.rs
dbolduc: ^ Note that this succeeds if the creds have no universe and the client has a non-GDU universe.

src/gax-internal/src/universe_domain.rs
alvarowolfx: Yes, it should be a fail. No universe means GDU. If the client provides a non GDU, it's mismatch. I've added a test case.

src/gax-internal/src/universe_domain.rs
alvarowolfx: good approach. But we need to take into consideration that None means GDU, so it doesn't improve code readability that much

src/gax-internal/src/universe_domain.rs
alvarowolfx: good call, will split it up

alvarowolfx (commented):
I pushed a commit addressing review comments and also answered the question around the credential having no universe domain (which mean it's GDU)

src/gax-internal/src/universe_domain.rs
```diff
@@ -0,0 +1,125 @@
...
+        expected: &str,
+    ) -> TestResult {
+        let _env = match env_domain {
+            Some(domain) => ScopedEnv::set(UNIVERSE_DOMAIN_VAR, domain),
+            None => ScopedEnv::remove(UNIVERSE_DOMAIN_VAR),
+        };
+        let cred = mock_credentials(cred_domain);
+
+        let universe_domain = resolve(client_override, &cred).await?;
+        assert_eq!(universe_domain.as_str(), expected, "{universe_domain:?}");
```
dbolduc, line 98: nit:

```suggestion
        assert_eq!(universe_domain, expected);
```

src/gax-internal/src/universe_domain.rs
dbolduc: >  No universe means GDU

This implies to me that we messed up the `Credentials::universe_domain()` API. It should return a `String` instead of an `Option<String>`... oh well, too late to fix.

PR #5246, opened 2026-04-02
impl(gax): add with_universe_domain builder method

alvarowolfx:

### DIFF
--- a/src/gax/src/client_builder.rs
+++ b/src/gax/src/client_builder.rs
@@ -252,7 +252,7 @@ impl<F, Cr> ClientBuilder<F, Cr> {
 
     /// Configure the universe domain.
     ///
-    /// The universe domain is the default service domain for a given Cloud universe.
+    /// The universe domain is the default service domain for a given cloud universe.
     /// The default value is "googleapis.com".
     // TODO(#3646): Make this public and let example run when universe domain support is done.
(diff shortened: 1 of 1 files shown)


src/gax/src/client_builder.rs
```diff
@@ -427,6 +449,7 @@ pub mod internal {
         pub disable_follow_redirects: bool,
         pub grpc_subchannel_count: Option<usize>,
         pub grpc_request_buffer_capacity: Option<usize>,
+        pub universe_domain: Option<String>,
```
dbolduc: style nit: put this next to `endpoint` because they are conceptually similar.

src/gax/src/client_builder.rs
```diff
@@ -250,6 +250,17 @@ impl<F, Cr> ClientBuilder<F, Cr> {
         self
     }
 
+    /// Configure the universe domain.
+    ///
+    /// The universe domain is the default service domain for a given Cloud universe.
```
dbolduc: Q: s/Cloud/cloud/ ?

Aside: I looked and I still don't see any cloud.google.com docs to link to about what a universe domain is. That is sad.

alvarowolfx: yeah, it lacks public docs.

PR #5248, opened 2026-04-02
impl(auth): support universe domain for service account creds

alvarowolfx:
Towards #3646

### DIFF
--- a/src/auth/src/credentials/service_account.rs
+++ b/src/auth/src/credentials/service_account.rs
@@ -918,10 +918,7 @@ mod tests {
         let credentials = Builder::new(json_value.clone()).build()?;
 
         let universe_domain = credentials.universe_domain().await;
-        assert_eq!(
-            universe_domain.as_deref(),
-            Some("some-universe-domain.com")
-        );
+        assert_eq!(universe_domain.as_deref(), Some("some-universe-domain.com"));
 
         let credentials = Builder::new(json_value)
             .with_universe_domain("other-universe-domain.com")


src/auth/src/credentials/service_account.rs
```diff
@@ -265,6 +267,31 @@ impl Builder {
...
+    /// # Example
+    /// ```
+    /// # use google_cloud_auth::credentials::service_account::Builder;
+    /// # use serde_json::json;
+    /// # async fn sample() -> anyhow::Result<()> {
+    /// # let config = json!({
+    /// #     "type": "service_account",
+    /// #     "client_email": "foo@bar.com",
+    /// #     "private_key": "---BEGIN---"
+    /// # });
```
dbolduc: nit: this could just be `json!({})` if we are hiding it.

src/auth/src/credentials/service_account.rs
```diff
@@ -866,6 +905,55 @@ mod tests {
...
+            "client_email": "test-client-email",
+            "private_key_id": "test-private-key-id",
+            "private_key": private_key,
+            "project_id": "test-project-id",
+        });
+
+        let credentials = Builder::new(json_value).build()?;
+
+        let universe_domain = credentials.universe_domain().await;
+        assert_eq!(universe_domain, None, "{universe_domain:?}");
```
dbolduc: nit: no need for the extra printing when using `assert_eq!`. It only adds value for like `assert!(universe_domain.is_none(), "{universe_domain:?}");`

here and below.

src/auth/src/credentials/service_account.rs
```diff
@@ -265,6 +267,31 @@ impl Builder {
...
+    /// #     "type": "service_account",
+    /// #     "client_email": "foo@bar.com",
+    /// #     "private_key": "---BEGIN---"
+    /// # });
+    /// let credentials = Builder::new(config)
+    ///     .with_universe_domain("googleapis.com")
+    ///     .build()?;
+    /// # Ok(()) }
+    /// ```
+    pub fn with_universe_domain<S: Into<String>>(mut self, universe_domain: S) -> Self {
```
dbolduc: comment: hmm, if we can override things with a builder function, why can't we override things with the `GOOGLE_CLOUD_UNIVERSE_DOMAIN` env var?

w/e. Not my spec.

src/auth/src/credentials/service_account.rs
```diff
@@ -265,6 +267,31 @@ impl Builder {
         self
     }
 
+    /// Sets the universe domain for this credentials.
+    ///
+    /// The universe domain is the default service domain for a given cloud universe.
```
dbolduc: nit: I don't think "default" is correct.

I asked Gemini to rewrite this and it spat out:

```
/// Sets the Google Cloud universe domain for these credentials.
///
/// A "universe" is an isolated Google Cloud environment, such as the public 
/// cloud or a sovereign/air-gapped deployment (e.g., Google Distributed Cloud). 
/// The universe domain acts as the base URL for constructing API endpoints 
/// within that environment.
///
/// By default, this is set to `googleapis.com`. You should only override this 
/// if your application is operating within a custom Cloud universe and needs 
/// to direct authentication and service requests to a different base endpoint.
```

Consider taking any pieces of that that look good.

dbolduc: Also, #5259.

Maybe this should be the documentation for the `Credentials::universe_domain` accessor, and here we just say "override the value" without needing to explain what is a universe.

dbolduc:
Also, double check the title. The builder option is public, so this should be a `feat`. If we are missing pieces of the implementation and the title is inaccurate (1) sorry, (2) please fix it :)

PR #5251, opened 2026-04-02
impl(pubsub): subscriber shutdown

dbolduc:
Part of the work for #5024

Support a shutdown, signaled by the application (via a `ShutdownToken`, or by `drop`ping the stream), and after a permanent error.

None of this is public. The next PR will make it public and add the proper documentation for the `shutdown_token()` accessor.

I couldn't really break this down without regressing existing behavior. And in the middle of the release, that seemed like a bad idea.

### DIFF
--- a/src/pubsub/src/subscriber/message_stream.rs
+++ b/src/pubsub/src/subscriber/message_stream.rs
@@ -72,9 +72,10 @@ pub struct MessageStream {
     /// A token that can detect a shutdown from the application.
     shutdown: CancellationToken,
 
-    /// Shutdown the stream gracefully if the application drops this struct.
+    /// Signal a shutdown if the application drops this struct.
     ///
-    /// It is simpler to hold a `DropGuard` than `impl Drop for MessageStream`.
+    /// This field is intentionally unused; it exists solely to trigger a
+    /// shutdown signal via its `Drop` implementation.
     _shutdown_guard: DropGuard,
 }
 
@@ -257,6 +258,7 @@ impl MessageStream {
     pub fn shutdown_token(&self) -> ShutdownToken {
         ShutdownToken {
             inner: self.shutdown.clone(),
+            // This future is ready when the lease loop shutdown completes.
             fut: self.lease_loop.clone(),
         }
     }


src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -145,6 +155,22 @@ impl MessageStream {
...
         let lease_loop = handle.map(|_| ()).boxed().shared();
 
+        let weak_message_tx = message_tx.downgrade();
+        let weak_ack_tx = ack_tx.downgrade();
+
+        let shutdown = CancellationToken::new();
+        let shutdown_clone = shutdown.clone();
+        let _shutdown_guard = shutdown.clone().drop_guard();
+        tokio::spawn(async move {
+            // Hold the strong senders for the channels, dropping them when an
```
suzmue, line 166: nit: can you make clear that this is the only place that is allowed to hold the strong version? That way we don't come back in later and try to do a refactor to pass the Strong Sender somewhere else, without realizing there are consequences.

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -69,6 +71,11 @@ pub struct MessageStream {
 
     /// A token that can detect a shutdown from the application.
     shutdown: CancellationToken,
+
+    /// Shutdown the stream gracefully if the application drops this struct.
+    ///
+    /// It is simpler to hold a `DropGuard` than `impl Drop for MessageStream`.
```
suzmue: Consider alternate wording:

```suggestion
    /// This field is intentionally unused; it exists solely to trigger a 
    /// shutdown signal via its `Drop` implementation when `MessageStream` is dropped.
```

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -228,17 +253,24 @@ impl MessageStream {
         }))
     }
 
-    #[cfg(test)]
+    #[cfg(test)] // TODO(#5024) - document and make public
+    pub fn shutdown_token(&self) -> ShutdownToken {
+        ShutdownToken {
+            inner: self.shutdown.clone(),
+            fut: self.lease_loop.clone(),
```
suzmue, line 262: nit: Returning self.lease_loop was unexpected when I read this. Maybe a comment to explain that "The lease_loop exits once shutdown is complete" or "shutdown is considered complete when the lease loop has ended".

src/pubsub/src/subscriber/message_stream.rs
```diff
@@ -250,7 +282,7 @@ impl MessageStreamImpl {
         loop {
             // Serve a message if we have one ready.
             if let Some((m, hi)) = self.pool.pop_front() {
-                return Some(Ok((m, hi.into_handler(self.ack_tx.clone()))));
+                return Some(Ok((m, hi.into_handler(self.ack_tx.upgrade()?))));
```
suzmue, line 287: TIL: Upgrade returns Option, and this gets evaluated and None is returned in this case. Pretty cool.

src/pubsub/src/subscriber/message_stream.rs
dbolduc: Done.

src/pubsub/src/subscriber/message_stream.rs
dbolduc: > without realizing there are consequences.

The unit tests will definitely fail.

> this is the only place that is allowed to hold the strong version?

Well, this is the only place in the `MessageStream`. The `Handler`s all get a strong `ack_tx`. And the lease loop might be holding a strong `message_tx`, depending on the configured shutdown behavior. 🫨 

As I write this, I agree with your main point that this is not clear, unnecessarily complicated. So, #5255.

I am going to merge this PR first, and get the feature done, then spend an hour or two trying to build the background task into the `LeaseLoop`. And have the `LeaseLoop` outputs be weak senders + the cancellation token. Which seems like a clearer/cleaner approach.

src/pubsub/src/subscriber/message_stream.rs
dbolduc: The member variable says exactly this:

https://github.com/googleapis/google-cloud-rust/blob/431dc2a4de391016d0ef1246b1cac9466f3a1fe9/src/pubsub/src/subscriber/message_stream.rs#L67-L68

I guess I can duplicate the comment.

Aside: it was probably surprising because my names are bad. Maybe `shutdown` should be `cancel` and `lease_loop` should be `shutdown`. WDYT?

PR #5257, opened 2026-04-03
feat(pubsub): add APIs to shutdown subscriber

dbolduc:
Most of the work for #5024 (the rest are cleanups)

Provide a way for applications to both signal and await a shutdown of a subscriber message stream.

---

Docs look like:

Screenshot From 2026-04-03 12-52-40

Screenshot From 2026-04-03 12-54-36


dbolduc:
> nit: PR title

Fixed.

PR #5260, opened 2026-04-03
cleanup(pubsub): use shutdown APIs in examples

dbolduc:
Part of the work for #5024

### DIFF
--- a/src/pubsub/examples/src/subscriber/subscriber_stream.rs
+++ b/src/pubsub/examples/src/subscriber/subscriber_stream.rs
@@ -30,7 +30,7 @@ pub async fn sample(project_id: &str, subscription_id: &str) -> anyhow::Result<(
 
     println!("listening for messages...");
 
-    let stream = stream.into_stream();
+    let mut stream = stream.into_stream();
     while let Some((message, handler)) = stream.try_next().await? {
         println!("received message: {message:?}");
         handler.ack();


src/pubsub/examples/src/subscriber/quickstart_subscriber.rs
```diff
@@ -14,24 +14,27 @@
...
-    while let Ok(Some(item)) = tokio::time::timeout_at(deadline.into(), stream.next()).await {
-        let (message, handler) = item?;
+    while let Some((message, handler)) = stream.next().await.transpose()? {
         println!("received message: {message:?}");
         handler.ack();
     }
+    shutdown.await?;
 
     println!("done listening for messages");
     Ok(())
```
suzmue, line 40: No action required. Here is an alternative that doesn't have a 10 second delay if the loop were to receive an error early. It's probably a worse example.

```suggestion
    // Terminate the example after 10 seconds.
    let delay = tokio::time::sleep(Duration::from_secs(10));
    tokio::pin!(delay);
    
    println!("listening for messages...");
    let loop_result: Result<()> = loop {
        tokio::select! {
            _ = &mut delay => {
                break Ok(());
            }
            maybe_msg = stream.next() => {
                match maybe_msg {
                    Some(Ok((message, handler))) => {
                        println!("received message: {message:?}");
                        handler.ack();
                    }
                    Some(Err(e)) => break Err(e),
                    None => break Ok(()),
                }
            }
        }
    };

   stream.shutdown_token().shutdown().await;

   println!("done listening for messages");
   loop_result
```

dbolduc: > doesn't have a 10 second delay if the loop were to receive an error early.

We would exit because of the `?` and skip awaiting a shutdown on an early error.

(which is fine, because there is a 99.999% chance the error is from starting the stream and shutdown wouldn't help.)

PR #5289, opened 2026-04-04
feat(o11y): enable `resource_name_heuristic` for compute, sql, and storage

haphungw:
Enable `resource_name_heuristic` in `librarian.yaml` for the `compute`, `cloud sql`, and `storage` libraries.


librarian.yaml
```diff
@@ -249,10 +249,9 @@ libraries:
           match: |2
                 - `folders/<folder-number>`
                 - `organizations/<organization-number>`
-          replace: |+
```
dbolduc, line 252: intentional?

It seems worse than before

haphungw: thank you, fixed it!

PR #5292, opened 2026-04-05
feat(o11y): remove `google-cloud-unstable-tracing` guards and regenerate

haphungw:

PR #5293, opened 2026-04-05
fix(o11y): use correct Tokio `task_local` API in `RequestRecorder`

haphungw:
`RequestRecorder::current()` was using `RECORDER.try_get()`, which is not a valid method on `tokio::task::LocalKey`, causing a compilation error. 

Surfaced in #5292 when `google_cloud_unstable_tracing` is removed.

### DIFF
--- a/src/gax-internal/src/observability/client_signals/recorder.rs
+++ b/src/gax-internal/src/observability/client_signals/recorder.rs
@@ -120,7 +120,7 @@ impl RequestRecorder {
     /// }
     /// ```
     pub fn current() -> Option<Self> {
-        RECORDER.try_get().ok()
+        RECORDER.try_with(|r| r.clone()).ok()
     }
 
     /// Returns the data captured for the client layer.


src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -120,7 +120,7 @@ impl RequestRecorder {
     /// }
     /// ```
     pub fn current() -> Option<Self> {
-        RECORDER.try_get().ok()
+        RECORDER.try_with(|r| r.clone()).ok()
```
coryan, line 123: This is the same as the previous code, no?

https://docs.rs/tokio/1.48.0/src/tokio/task/task_local.rs.html#285-287

What is the motivation for the change?  I could be missing something but seems like it does not change anything?

coryan: > `RequestRecorder::current()` was using `RECORDER.try_get()`, which is not a valid method on `tokio::task::LocalKey`, causing a compilation error.

Oh, I see, this only appeared in tokio 1.48.0:

https://docs.rs/tokio/1.48.0/tokio/task/struct.LocalKey.html#method.try_get

And we are only requiring 1.47.0:

https://github.com/googleapis/google-cloud-rust/blob/10e89c00a9ef89c04e5809780083397dd6999a21/Cargo.toml#L397

I think the right fix is to bump the minimum Tokio version requirement. Unless we think there is a good reason to support 1.47 ?

haphungw: https://docs.rs/tokio/1.47.0/tokio/task/struct.LocalKey.html#method.try_with

Tokio's does not expose a `try_get()`, only a `try_with()`.
When we started removing the unstable tracing guards, with `try_get(),` we encountered this compilation err:
```
error[E0599]: no method named `try_get` found for struct `tokio::task::LocalKey<T>` in the current scope
```

haphungw: I just realized it's a version thing. `try_get()` is supported in 1.50.0. https://docs.rs/tokio/latest/tokio/task/struct.LocalKey.html#method.try_get

Do we want to update Tokio's version instead?

haphungw: > > `RequestRecorder::current()` was using `RECORDER.try_get()`, which is not a valid method on `tokio::task::LocalKey`, causing a compilation error.
> 
> Oh, I see, this only appeared in tokio 1.48.0:
> 
> https://docs.rs/tokio/1.48.0/tokio/task/struct.LocalKey.html#method.try_get
> 
> And we are only requiring 1.47.0:
> 
> https://github.com/googleapis/google-cloud-rust/blob/10e89c00a9ef89c04e5809780083397dd6999a21/Cargo.toml#L397
> 
> I think the right fix is to bump the minimum Tokio version requirement. Unless we think there is a good reason to support 1.47 ?

agreed, sounds good

haphungw: are we bumping to 1.48.0 or the latest?

Issue #5159, opened 2026-03-26
Consider `WithTransportLogging`, `WithTransportMetric`, `WithTransportSpan`


coryan:
These are analogs to #5157 and #5158 and #5148 but for the transport layer.

I am not sure they are better than the code we have today, they might be. Maybe prototype one and if it works, then open bugs for each one.

coryan:
We can do this after the milestone., they would just just change the implementation details of `gax-internal`.

PR #5297, opened 2026-04-06
impl(gax-internal): refactor transport-level spans, metrics, and logging

haphungw:
Implement transport-level o11y decorators for http requests: `WithTransportSpan`, `WithTransportMetric`, and `WithTransportLogging`. 

For #5159

### DIFF
--- a/src/gax-internal/src/observability/client_signals/transport_metric.rs
+++ b/src/gax-internal/src/observability/client_signals/transport_metric.rs
@@ -263,4 +263,76 @@ mod tests {
 
         Ok(())
     }
+
+    #[tokio::test(start_paused = true)]
+    async fn record_error() -> anyhow::Result<()> {
+        let exporter = InMemoryMetricExporter::default();
+        let provider = SdkMeterProvider::builder()
+            .with_reader(PeriodicReader::builder(exporter.clone()).build())
+            .build();
+        let metric =
+            TransportMetric::new_with_provider(Some(&TEST_INFO), Arc::new(provider.clone()));
+        let recorder = RequestRecorder::new(TEST_INFO);
+        recorder.on_client_request(
+            ClientRequestAttributes::default()
+                .set_url_template("https://example.com/v1/projects/{project}/topics/{topic}")
+                .set_rpc_method(TEST_METHOD),
+        );
+
+        let request = reqwest::Request::new(
+            reqwest::Method::GET,
+            "https://example.com/v1/projects/p/topics/t".parse()?,
+        );
+
+        recorder.on_http_request(&request);
+        tokio::time::sleep(DELAY).await;
+
+        let err =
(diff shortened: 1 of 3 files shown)


src/gax-internal/src/observability/client_signals/transport_metric.rs
```diff
@@ -0,0 +1,266 @@
...
+];
+
+const METRIC_NAME: &str = "gcp.client.attempt.duration";
+const METRIC_UNIT: &str = "s";
+
+/// Simplify the creation of transport attempt duration metrics.
+#[derive(Clone, Debug)]
+pub struct TransportMetric(Option<Histogram<f64>>);
+
+impl TransportMetric {
```
coryan, line 46: Does this differ by anything other than the name from `DurationMetric`?  There may an opportunity to refactor them (in a future PR), maybe with a trait that captures the name what method in the snapshot is used to get the duration?

src/gax-internal/src/observability/client_signals/recorder.rs
```diff
@@ -297,6 +306,11 @@ impl ClientSnapshot {
         self.start.elapsed()
     }
 
+    /// Returns the transport attempt duration.
+    pub fn transport_duration(&self) -> Option<std::time::Duration> {
```
coryan, line 310: Please add unit tests for these new methods in ClientSnapshot.

src/gax-internal/src/observability/client_signals/with_transport_metric.rs
```diff
@@ -0,0 +1,60 @@
...
+use std::pin::Pin;
+use std::task::{Context, Poll};
+
+/// Implements a decorator for [Future] adding duration metrics for transport attempts.
+#[must_use = "futures do nothing unless you `.await` or poll them"]
+#[pin_project]
+pub struct WithTransportMetric<F> {
+    #[pin]
+    inner: F,
+    metric: TransportMetric,
```
coryan, line 28: If `TransportMetric` becomes `DurationMetric<TransportLatency>` (or something like it) then this struct could be refactored to be a generic thing too.

src/gax-internal/src/observability/client_signals/transport_metric.rs
haphungw: yes absolutely. there should be a lot of common things to extract.

haphungw:
My only concern is that gRPC will be more complicated than a direct port. Since `tonic` relies on Tower, we'd have to rewrite this same logic to fit that paradigm. That means we'll end up maintaining two different implementation styles for transport observability across HTTP and gRPC.

I'll write up the gRPC version to see how it looks.

haphungw:
> Separately, the title still says "WIP"

Should I try to merge the whole thing or split it into smaller PRs?

PR #5299, opened 2026-04-06
test(o11y): add unit test for gRPC retry loop

### LINKED ISSUE #5281: F1.10: gRPC T4 retries

haphungw:
To run:
```bash
RUSTFLAGS="--cfg google_cloud_unstable_tracing" cargo test -p google-cloud-gax-internal --features _internal-grpc-client --lib observability::client_signals::tests::grpc_client_request_retry
```

Fixes #5281

### DIFF
--- a/src/gax-internal/src/observability/client_signals.rs
+++ b/src/gax-internal/src/observability/client_signals.rs
@@ -396,6 +396,9 @@ mod tests {
 
     #[cfg(feature = "_internal-grpc-client")]
     pub(crate) async fn recorded_request_grpc_stub_retry(url: &str) -> Result<String, Error> {
+        use google_cloud_gax::retry_policy::{AlwaysRetry, RetryPolicyExt};
+        use std::sync::Arc;
+
         let recorder = RequestRecorder::current().expect("current recorder should be available");
         recorder.on_client_request(
             ClientRequestAttributes::default()
@@ -406,8 +409,7 @@ mod tests {
 
         let mut config = crate::options::ClientConfig::default();
         config.tracing = true;
-        use google_cloud_gax::retry_policy::{AlwaysRetry, RetryPolicyExt};
-        config.retry_policy = Some(std::sync::Arc::new(AlwaysRetry.with_attempt_limit(3)));
+        config.retry_policy = Some(Arc::new(AlwaysRetry.with_attempt_limit(3)));
 
         config.cred = Some(Anonymous::new().build());
 
@@ -427,8 +429,6 @@ mod tests {
             message: "test message".into(),
             ..Default::default()
         };
-
-        tokio::time::sleep(TEST_REQUEST_DURATION).await;
 
         let response: Result<tonic::Response<EchoResponse>, google_cloud_gax::error::Error> =
             client
@@ -548,7 +548,7 @@ mod tests {
             })),
         ])
         .await?;
-        
+
         let signals = SignalProviders::new();
 
         let metric = DurationMetric::new_with_provider(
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/observability/client_signals.rs
coryan, line 202: Note how this test uses `start_paused = true` that does three things: 

1) Makes Tokio use a virtual clock, so `Instant::now()` only changes when you advance the time or sleep, 
2) It makes the sleeps take zero wall clock time because tokio simply advances the virtual clock, and
3) It makes the clock measurements match exactly the desired time.

src/gax-internal/src/observability/client_signals.rs
```diff
@@ -394,6 +394,57 @@ mod tests {
...
+                "Echo",
+            ));
+            e
+        };
+        let request = EchoRequest {
+            message: "test message".into(),
+            ..Default::default()
+        };
+
+        tokio::time::sleep(TEST_REQUEST_DURATION).await;
```
coryan: ... this test sleeps, but the wall clock duration is not predictable (but at least 750ms) because ...

src/gax-internal/src/observability/client_signals.rs
```diff
@@ -486,6 +537,50 @@ mod tests {
         Ok(())
     }
 
+    #[cfg(feature = "_internal-grpc-client")]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn grpc_client_request_retry() -> anyhow::Result<()> {
```
coryan, line 542: The test does not start with a paused clock.

src/gax-internal/src/observability/client_signals.rs
```diff
@@ -486,6 +537,50 @@ mod tests {
...
+
+        let spans = signals.trace_exporter.get_finished_spans()?;
+
+        // Assert that at least one span has resend_count = 1
+        let retry_span = spans.iter().find(|s| {
+            s.attributes
+                .iter()
+                .any(|kv| kv.key.as_str() == "gcp.grpc.resend_count" && kv.value.to_string() == "1")
+        });
+        assert!(retry_span.is_some(), "expected a span with resend_count=1");
```
coryan: It also looks like you don't use the sleep, so maybe we can just remove it?

coryan: If this fails it would be nice to have the actual spans printed out.

src/gax-internal/src/observability/client_signals.rs
```diff
@@ -394,6 +394,57 @@ mod tests {
...
+        recorder.on_client_request(
+            ClientRequestAttributes::default()
+                .set_rpc_method(TEST_METHOD)
+                .set_url_template(TEST_URL_TEMPLATE)
+                .set_resource_name("//test.googleapis.com/test-only".to_string()),
+        );
+
+        let mut config = crate::options::ClientConfig::default();
+        config.tracing = true;
+        use google_cloud_gax::retry_policy::{AlwaysRetry, RetryPolicyExt};
```
coryan: Please move this to the beginning of the file or the function.

src/gax-internal/src/observability/client_signals.rs
```diff
@@ -394,6 +394,57 @@ mod tests {
...
+        config.retry_policy = Some(std::sync::Arc::new(AlwaysRetry.with_attempt_limit(3)));
```
coryan: Consider  a use declaration for `std::sync::Arc`

coryan (changes requested):
The easiest fix is to remove the sleep, but if you will need it in a future PR or something then we need to start the test paused.

src/gax-internal/src/observability/client_signals.rs
haphungw: done

haphungw: done

src/gax-internal/src/observability/client_signals.rs
haphungw: done

src/gax-internal/src/observability/client_signals.rs
haphungw: done

src/gax-internal/src/observability/client_signals.rs
haphungw: i see, makes sense

coryan (approved):

PR #5300, opened 2026-04-06
impl(auth): add universe_domain call to MDS client

alvarowolfx:
Also moving MDS client to a builder like interface, to allow extension later (e.g. retries) later.

Towards #3646


suzmue (approved):
nit: I would recommend having the refactor be in a separate PR in the future.

PR #5301, opened 2026-04-06
impl(auth): access boundaries only supported on GDU

alvarowolfx:

### DIFF
--- a/src/auth/src/access_boundary.rs
+++ b/src/auth/src/access_boundary.rs
@@ -242,13 +242,13 @@ where
         universe_domain: Option<String>,
     ) -> Self {
         let credentials = Arc::new(credentials);
-        let provider = IAMAccessBoundaryProvider {
-            credentials: credentials.clone(),
-            url: access_boundary_url,
-        };
 
         // Only enable access boundary for default universe domain
         let access_boundary = if is_default_universe_domain(universe_domain) {
+            let provider = IAMAccessBoundaryProvider {
+                credentials: credentials.clone(),
+                url: access_boundary_url,
+            };
             Arc::new(AccessBoundary::new(provider))
         } else {
             Arc::new(AccessBoundary::new_no_op())
@@ -1295,7 +1295,7 @@ pub(crate) mod tests {
         let access_boundary = get_access_boundary_from_headers(cached_headers);
         assert!(
             access_boundary.is_none(),
-            "Expected no access boundary header for non-default universe"
+            "Expected no access boundary header for non-default universe: {access_boundary:?}"
         );
 
         Ok(())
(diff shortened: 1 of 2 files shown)


coryan:
Try to avoid the double negative. Maybe "impl(auth): only GDU needs access boundaries" or "impl(auth): only GDU supports access boundaries".

alvarowolfx:
> Try to avoid the double negative. Maybe "impl(auth): only GDU needs access boundaries" or "impl(auth): only GDU supports access boundaries".

good call, updated the title.

src/auth/src/access_boundary.rs
```diff
@@ -235,13 +236,23 @@ impl<T> CredentialsWithAccessBoundary<T>
...
         let credentials = Arc::new(credentials);
         let provider = IAMAccessBoundaryProvider {
             credentials: credentials.clone(),
             url: access_boundary_url,
         };
-        let access_boundary = Arc::new(AccessBoundary::new(provider));
+
+        // Only enable access boundary for default universe domain
+        let access_boundary = if is_default_universe_domain(universe_domain) {
+            Arc::new(AccessBoundary::new(provider))
```
dbolduc, line 252: nit: move `let provider = ...` into here?

src/auth/tests/credentials.rs
```diff
@@ -574,6 +574,13 @@ mod tests {
             .times(0..)
             .respond_with(status_code(200).body("some-client-email")),
         ); // access boundaries might have enough time to fetch email
+        server.expect(
+            Expectation::matching(request::path(
+                "/computeMetadata/v1/universe/universe-domain",
+            ))
+            .times(0..)
+            .respond_with(status_code(404)),
+        ); // access boundaries might have enough time to fetch universe domain to check if should be enabled
```
dbolduc: consider:

```suggestion
        ); // The access boundary background task might query the MDS for its universe domain.
```

src/auth/src/access_boundary.rs
```diff
@@ -1248,6 +1266,67 @@ pub(crate) mod tests {
...
+            credentials: Arc::new(mock),
+            mds_client,
+            iam_endpoint_override: None,
+            url: OnceLock::new(),
+        };
+
+        let val = provider.fetch_access_boundary().await?;
+        assert!(
+            val.is_none(),
+            "Expected None for non-default universe domain"
```
dbolduc: nit: The error message isn't telling us more than the assertion. Printing `"{val:?}"` would tell us more.

src/auth/src/access_boundary.rs
```diff
@@ -1248,6 +1266,67 @@ pub(crate) mod tests {
...
+        );
+
+        let cached_headers = creds.headers(Extensions::new()).await?;
+        let token = get_token_from_headers(cached_headers.clone());
+        assert!(token.is_some(), "{token:?}");
+
+        let access_boundary = get_access_boundary_from_headers(cached_headers);
+        assert!(
+            access_boundary.is_none(),
+            "Expected no access boundary header for non-default universe"
```
dbolduc: same

alvarowolfx:
> I don't know what an access boundary is, but the code looks reasonable

thanks for the review, I've just pushed a commits addressing your comments

PR #5302, opened 2026-04-06
impl(gax-internal): use universe_domain when creating gRPC/HTTP transport

alvarowolfx:


src/gax-internal/src/grpc.rs
```diff
@@ -104,12 +104,19 @@ impl Client {
         >,
     ) -> ClientBuilderResult<Self> {
         let credentials = Self::make_credentials(&config).await?;
+
+        let universe_domain =
+            crate::universe_domain::resolve(config.universe_domain.as_deref(), &credentials)
+                .await
+                .map_err(BuilderError::transport)?;
```
dbolduc: Why is this a transport error? The client does not even try to initialize the transport layer. It fails before then because of bad configuration.

Should we expand the `ClientBuilder::Error` to have a case for mismatched universe domains?

src/gax-internal/src/http.rs
```diff
@@ -595,9 +607,24 @@ mod tests {
...
+    use google_cloud_auth::errors::CredentialsError;
     use http::{HeaderMap, HeaderValue, Method};
     use test_case::test_case;
 
+    type AuthResult<T> = std::result::Result<T, CredentialsError>;
+    type TestResult = anyhow::Result<()>;
+
+    mockall::mock! {
+        #[derive(Debug)]
+        Credentials {}
```
dbolduc, line 605: Aside, non-blocking: We don't need to define this more than once per crate.

And we could even put it in `google-cloud-test-utils` to define it once for all tests

src/gax-internal/src/host.rs
```diff
@@ -33,15 +38,21 @@ pub(crate) fn header(endpoint: Option<&str>, default_endpoint: &str) -> Result<S
...
+    origin_and_header(endpoint, default_endpoint, universe_domain).map(|(origin, _)| origin)
 }
 
 fn origin_and_header(
     endpoint: Option<&str>,
     default_endpoint: &str,
+    universe_domain: &str,
 ) -> Result<(Uri, String), HostError> {
-    let default_origin = Uri::from_str(default_endpoint).map_err(HostError::Uri)?;
+    let default_endpoint = default_endpoint.replace(DEFAULT_UNIVERSE_DOMAIN, universe_domain);
```
dbolduc: nit: this isn't really the "default" after we swap in a custom universe domain.

src/gax-internal/src/host.rs
```diff
@@ -57,9 +68,11 @@ fn origin_and_header(
         .ok_or_else(|| HostError::MissingAuthority(endpoint.to_string()))?
         .host()
         .to_string();
+
+    let custom_suffix = format!(".{}", universe_domain);
     let (Some(prefix), Some(service)) = (
-        custom_host.strip_suffix(".googleapis.com"),
-        default_host.strip_suffix(".googleapis.com"),
+        custom_host.strip_suffix(&custom_suffix),
+        default_host.strip_suffix(&custom_suffix),
```
dbolduc: This seems wrong. The universe domain is not necessarily tied to the endpoint right?

Like for the `.with_endpoint("foo.googleapis.com").with_universe_domain("my-universe-domain.com")` example.

The host should be "foo.googleapis.com" but this would return "foo.my-universe-domain.com"

dbolduc (changes requested):

alvarowolfx:
Split gax changes into https://github.com/googleapis/google-cloud-rust/pull/5419

src/gax-internal/src/host.rs
```diff
@@ -58,6 +58,14 @@ fn origin_and_header(
         .ok_or_else(|| HostError::MissingAuthority(endpoint.to_string()))?
         .host()
         .to_string();
+
+    let is_default_universe_domain = default_host.ends_with(".googleapis.com");
+    if !is_default_universe_domain {
+        // For non GDU environments, endpoint takes priority if provided.
+        // And we don't treat regional/locational endpoints specially.
+        return Ok((custom_origin, custom_host));
```
dbolduc: is `private.my-custom-ud.com` a thing? Or `restricted.my-custom-ud.com`?

In which case we would want to use like `<service>.my-custom-ud.com` as the host, but this would use `private.my-custom-ud.com`.

Duckie found b/361611906 for me. Seems like yes?

dbolduc: Don't be afraid to scrap this whole `origin_and_header` function. I wrote it and it was not my finest work.

(My memory is that I was trying to make it as efficient as possible, but (1) that is not so important for client-construction vs. like per request, (2) it made the code inflexible.).

src/gax-internal/src/host.rs
```diff
@@ -59,16 +70,10 @@ fn origin_and_header(
...
-        // For non GDU environments, endpoint takes priority if provided.
-        // And we don't treat regional/locational endpoints specially.
-        return Ok((custom_origin, custom_host));
-    }
-
+    let universe_suffix = format!(".{universe_domain}");
     let (Some(prefix), Some(service)) = (
-        custom_host.strip_suffix(".googleapis.com"),
-        default_host.strip_suffix(".googleapis.com"),
+        custom_host.strip_suffix(&universe_suffix),
```
dbolduc, line 75: I think this fails to strip the suffix on the `.with_endpoint("foo.googleapis.com").with_universe_domain("my-custom.ud")` case and then it returns `"foo.custom.ud"`, when it should return `"foo.googleapis.com"`.

Assuming I know what the caller is supplying to this function.

alvarowolfx: yeah, that's the conflicting thing that I'm stuck on this feature. You commented that before on https://github.com/googleapis/google-cloud-rust/pull/5302#discussion_r3047163947.

I'm trying to make the code look good and pass the test cases for localhost and the scenario that you described.

```
#[test_case("localhost:5678", "test.googleapis.com"; "emulator")]
#[test_case("https://localhost:5678", "test.googleapis.com"; "emulator with scheme")]
#[test_case("https://test.googleapis.com", "test.googleapis.com"; "GDU endpoint with universe domain") ]
#[test_case("https://test.another-universe.com", "test.another-universe.com"; "endpoint priority") ]
fn header_universe_domain(input: &str, want: &str) -> anyhow::Result<()> {
    let got = header(
        Some(input),
        "https://test.googleapis.com",
        "my-custom-universe.com",
    )?;
    assert_eq!(got, want, "input={input:?}");
    Ok(())
}
```

dbolduc:
Take inspiration from this:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=e3525652e19f18e517c6f64547d6a18e

The idea is just to give up if we don't recognize the universe domain as the GDU or the supplied universe domain option.

It fails if you have credentials for universe A, but then you try to make a request to universe B where neither are the GDU. Is that even a use case? It seems to defeat the purpose.

src/gax-internal/src/host.rs
dbolduc: FWIW, I think these tests are wrong. Not sure why I wrote them.

https://github.com/googleapis/google-cloud-rust/blob/822e21599ea8ea85e0345c0452146f21b285f04b/src/gax-internal/src/host.rs#L161-L163

Gemini tells me we should use "localhost" as the host when talking to an emulator, not the default service endpoint.

alvarowolfx: yup, I'm also finding that some previous assertions for universe domains might be incorrect, like https://github.com/googleapis/google-cloud-rust/blob/ffc6f3ebc96c8380eccda3de79512aa012ef275b/src/gax-internal/src/host.rs#L159, it should prioritize the endpoint override, so host should be `test.my-universe-domain.com`

alvarowolfx:
Split host resolution header into https://github.com/googleapis/google-cloud-rust/pull/5464

src/gax-internal/src/grpc.rs
```diff
@@ -624,8 +632,61 @@ where
...
+    use test_case::test_case;
+
+    type TestResult = anyhow::Result<()>;
+
+    #[tokio::test]
+    #[test_case(None, DEFAULT_UNIVERSE_DOMAIN, "https://test.googleapis.com/"; "default GDU")]
+    #[test_case(None, "my-custom-universe.com", "https://test.my-custom-universe.com/"; "default custom universe domain")]
+    #[test_case(Some("https://test.googleapis.com/"), DEFAULT_UNIVERSE_DOMAIN, "https://test.googleapis.com/"; "GDU override")]
+    #[test_case(Some("https://another-custom-universe.com/"), "my-custom-universe.com", "https://another-custom-universe.com/"; "custom endpoint override with universe domain")]
+    #[test_case(Some("https://rep.test.googleapis.com/"), "my-custom-universe.com", "https://rep.test.googleapis.com/"; "regional endpoint with universe domain")]
```
dbolduc: nit: I think this is more like

```suggestion
    #[test_case(Some("https://test.us-central1.rep.googleapis.com/"), "my-custom-universe.com", "https://test.us-central1.rep.googleapis.com/"; "regional endpoint with universe domain")]
```

dbolduc (approved):

Issue #5252, opened 2026-04-02
Add confirmed_nack and nack to the Subscriber


suzmue:
The Pub/Sub team would like us to expose the ExactlyOnce::confirmed_nack feature.

We also may want to take this opportunity to add an explicit AtLeastOnce::nack for improved code readability and navigation.

PR #5303, opened 2026-04-06
feat(pubsub): add confirmed_nack to ExactlyOnce handler

suzmue:
This feature allows the user to get the result of the ModifyAckDeadline request for the message.

For #5252

### DIFF
--- a/src/pubsub/src/subscriber/handler.rs
+++ b/src/pubsub/src/subscriber/handler.rs
@@ -41,6 +41,7 @@
 //! ```
 
 use crate::error::AckError;
+use crate::subscriber::lease_state::NACK_SHUTDOWN_ERROR;
 use tokio::sync::mpsc::UnboundedSender;
 use tokio::sync::oneshot::Receiver;
 
@@ -380,9 +381,7 @@ impl ExactlyOnceImpl {
     pub async fn confirmed_nack(self) -> AckResult {
         self.ack_tx
             .send(Action::ExactlyOnceNack(self.ack_id))
-            .map_err(|_| {
-                AckError::Shutdown(crate::subscriber::lease_state::NACK_SHUTDOWN_ERROR.into())
-            })?;
+            .map_err(|_| AckError::Shutdown(NACK_SHUTDOWN_ERROR.into()))?;
         self.result_rx
             .await
             .map_err(|e| AckError::Shutdown(e.into()))?
@@ -606,7 +605,7 @@ mod tests {
             err.source()
                 .expect("shutdown errors have a source")
                 .to_string(),
-            crate::subscriber::lease_state::NACK_SHUTDOWN_ERROR
+            NACK_SHUTDOWN_ERROR
         );
 
         Ok(())


src/pubsub/src/subscriber/handler.rs
```diff
@@ -308,6 +308,27 @@ impl ExactlyOnce {
...
+    /// async fn on_message(m: Message, h: ExactlyOnce) {
+    ///   match h.confirmed_nack().await {
+    ///     Ok(()) => println!("Confirmed nack for message={m:?}. The message will be redelivered."),
+    ///     Err(e) => println!("Failed to confirm nack for message={m:?} with error={e:?}"),
+    ///   }
+    /// }
+    /// ```
+    ///
+    /// If the result is an `Ok`, the message is guaranteed to be immediately
+    /// considered for redelivery.
```
dbolduc: The reader might think an error means the message will not get redelivered.

I think you should add "If an error occurs, the message will still be redelivered, but it may be held for the remainder of its `max_lease_extension`."

(Of course it may take that long anyway bc the server could hold it... 😒)

src/pubsub/src/error.rs
```diff
@@ -47,7 +47,7 @@ pub enum PublishError {
     Shutdown,
 }
 
-/// Represents an error that can occur when acknowledging a message.
+/// Represents an error that can occur when acknowledging or negatively acknowledging a message.
```
dbolduc: optional: maybe just "acking or nacking" ?

also, nit: 80 character lines?

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -115,6 +115,10 @@ pub(super) struct ExactlyOnceInfo {
     // report if a lease has expired. We do not want to mask a success with
     // a `LeaseExpired` error.
     pending: bool,
```
dbolduc, line 119: optional: Maybe this should be `acking` now?

No need to change in this PR. That seems noisy

src/pubsub/src/error.rs
```diff
@@ -75,15 +75,14 @@ pub enum AckError {
...
     #[error(
-        "shutdown before attempting the acknowledgement. The message was not acknowledged, and will be redelivered."
+        "shutdown before attempting the operation. The message was not acknowledged, and will be redelivered."
     )]
     ShutdownBeforeAck,
 
     /// Error during shutdown.
     ///
-    /// The result of the acknowledgement is unknown. The service may or may not
-    /// redeliver the message.
```
dbolduc, line 85: Meh this change says less than before.

Can we say "If you attempted to ack the message, the service may or may not redeliver it."

src/pubsub/src/subscriber/handler.rs
```diff
@@ -308,6 +308,27 @@ impl ExactlyOnce {
...
+    pub async fn confirmed_nack(mut self) -> std::result::Result<(), AckError> {
```
dbolduc, line 330: unit tests, please

src/pubsub/src/subscriber/handler.rs
```diff
@@ -353,6 +374,15 @@ impl ExactlyOnceImpl {
             .await
             .map_err(|e| AckError::Shutdown(e.into()))?
     }
+
+    pub async fn confirmed_nack(self) -> AckResult {
+        self.ack_tx
+            .send(Action::ExactlyOnceNack(self.ack_id))
+            .map_err(|_| AckError::ShutdownBeforeAck)?;
```
dbolduc: Q: Why should we return `ShutdownBeforeAck` for a nack?

I think `Shutdown(NACK_SHUTDOWN_ERROR)` is more appropriate

src/pubsub/src/subscriber/handler.rs
```diff
@@ -353,6 +374,15 @@ impl ExactlyOnceImpl {
...
+        self.result_rx
+            .await
```
dbolduc, line 386: Also, this can have a `LeaseExpired` error... doesn't that mean the nack was already successful? 🤔 

I am fine with not changing anything. It does indicate that the application should double check its settings. (And it's not like a nack error is actionable, anyways)

src/pubsub/src/subscriber/handler.rs
suzmue: done.

src/pubsub/src/subscriber/handler.rs
suzmue: Yeah I think its a little weird. I was considering doing that as well, will leave out of this PR tho.

src/pubsub/src/subscriber/lease_state.rs
suzmue: ack. will do in follow up.

src/pubsub/src/error.rs
suzmue: done.

src/pubsub/src/subscriber/handler.rs
suzmue: Done.

src/pubsub/src/error.rs
suzmue: Done.

src/pubsub/src/subscriber/handler.rs
```diff
@@ -353,13 +376,26 @@ impl ExactlyOnceImpl {
             .await
             .map_err(|e| AckError::Shutdown(e.into()))?
     }
+
+    pub async fn confirmed_nack(self) -> AckResult {
+        self.ack_tx
+            .send(Action::ExactlyOnceNack(self.ack_id))
+            .map_err(|_| {
+                AckError::Shutdown(crate::subscriber::lease_state::NACK_SHUTDOWN_ERROR.into())
```
dbolduc: style nit: probably worthy of a `use` declaration for `NACK_SHUTDOWN_ERROR`

src/pubsub/src/subscriber/handler.rs
```diff
@@ -515,6 +592,26 @@ mod tests {
...
+        assert_eq!(ack_rx.try_recv(), Err(TryRecvError::Empty));
+        drop(ack_rx);
+
+        let err = h.confirmed_nack().await.expect_err("nack should fail");
+        assert!(matches!(err, AckError::Shutdown(_)), "{err:?}");
+        assert_eq!(
+            err.source()
+                .expect("shutdown errors have a source")
+                .to_string(),
+            crate::subscriber::lease_state::NACK_SHUTDOWN_ERROR
```
dbolduc: same nit

dbolduc (approved):

src/pubsub/src/subscriber/handler.rs
suzmue: Switched to Shutdown(NACK_SHUTDOWN_ERROR).  Why do we have ShutdownBeforeAck to begin with?

src/pubsub/src/subscriber/handler.rs
suzmue: done. Thanks!

src/pubsub/src/subscriber/handler.rs
dbolduc: Because for acknowledgements it provides more information. You **know** your ack did not go through, because it was never attempted, if you see `ShutdownBeforeAck`.

Whereas if you get a `Shutdown` error then maybe the ack went through, maybe it didn't.

PR #5307, opened 2026-04-07
feat(spanner): inline BeginTransaction with the first statement (work in progress, do not review)

olavloite:
Work in progress, do not review: Implements inline-begin-transaction for the Spanner client. This is an optimization that reduces the number of round-trips to Spanner by 1 for most transactions.


olavloite:
No longer relevant. This is done in separate, smaller PRs.

PR #5309, opened 2026-04-07
impl(auth): IAM signer should respect universe domain

alvarowolfx:

### DIFF
--- a/src/auth/src/signer/iam.rs
+++ b/src/auth/src/signer/iam.rs
@@ -209,6 +209,7 @@ mod tests {
     use httptest::responders::{json_encoded, status_code};
     use httptest::{Expectation, Server};
     use serde_json::json;
+    use test_case::test_case;
     use tokio::time::Duration;
 
     type TestResult = anyhow::Result<()>;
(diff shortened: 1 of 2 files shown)


src/auth/src/signer/iam.rs
```diff
@@ -336,6 +350,52 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn test_sign_blob_url_with_override() -> TestResult {
+        let mock = MockCredentials::new();
+        let creds = Credentials::from(mock);
+        let signer = IamSigner::new(
+            "test@example.com".to_string(),
+            creds,
+            Some("http://example.com".to_string()),
```
dbolduc, line 365: nit: can you add a test case where we have a custom endpoint AND the credentials have a custom universe domain?

I know the mock credential's universe domain will be ignored, so it will basically be the same test. Maybe parameterize this for `None`, `Some("custom-universe.com")` ?

alvarowolfx: done

PR #5310, opened 2026-04-07
impl(auth): add retries for universe_domain endpoint on MDS client

alvarowolfx:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4013,6 +4013,7 @@ dependencies = [
  "google-cloud-gax-internal",
  "google-cloud-iam-v1",
  "google-cloud-pubsub",
+ "google-cloud-rpc",
  "google-cloud-test-macros",
  "google-cloud-wkt",
  "http",
@@ -5077,7 +5078,7 @@ dependencies = [
 
 [[package]]
 name = "google-cloud-wkt"
-version = "1.2.1"
+version = "1.3.0"
 dependencies = [
  "anyhow",
  "base64",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -468,8 +468,8 @@ tokio-stream      = { default-features = false, version = "0.1.16" }
 google-cloud-auth        = { default-features = false, version = "1.9.0", path = "src/auth" }
 google-cloud-gax         = { default-features = false, version = "1.9.1", path = "src/gax" }
 gaxi                     = { default-features = false, version = "0.7.12", path = "src/gax-internal", package = "google-cloud-gax-internal" }
-wkt                      = { default-features = false, version = "1", path = "src/wkt", package = "google-cloud-wkt" }
-google-cloud-wkt         = { default-features = false, version = "1", path = "src/wkt", package = "google-cloud-wkt" }
+wkt                      = { default-features = false, version = "1.3.0", path = "src/wkt", package = "google-cloud-wkt" }
+google-cloud-wkt         = { default-features = false, version = "1.3.0", path = "src/wkt", package = "google-cloud-wkt" }
 google-cloud-api         = { default-features = false, version = "1.5.0", path = "src/generated/api/types" }
 google-cloud-iam-v1      = { default-features = false, version = "1.8.0", path = "src/generated/iam/v1" }
 google-cloud-location    = { default-features = false, version = "1.8.0", path = "src/generated/cloud/location" }
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 language: rust
-version: v0.10.2-0.20260413154732-a2a6f2ea1322
+version: v0.10.2-0.20260413201001-f1215bd28dbc
 repo: googleapis/google-cloud-rust
 sources:
   conformance:
@@ -36,6 +36,10 @@ release:
     - .repo-metadata.json
   tools:
     cargo:
+      - name: cargo-hack
+        version: 0.6.44
+      - name: cargo-minimal-versions
+        version: 0.1.37
       - name: cargo-semver-checks
         version: 0.46.0
       - name: cargo-workspaces
@@ -1545,7 +1549,7 @@ libraries:
           output: src/storage/src/control/generated
           specification_format: none
           api_path: google/storage/v2
-          template: prost
+          template: storage
   - name: google-cloud-storagebatchoperations-v1
     version: 1.9.0
     copyright_year: "2025"
@@ -1649,7 +1653,7 @@ libraries:
     version: 1.8.0
     copyright_year: "2025"
   - name: google-cloud-wkt
-    version: 1.2.1
+    version: 1.3.0
     copyright_year: "2025"
     output: src/wkt
     roots:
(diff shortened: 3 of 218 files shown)


src/auth/src/mds/client.rs
```diff
@@ -123,6 +134,55 @@ impl Client {
...
+        &self,
+        request: reqwest::RequestBuilder,
+        error_message: &'static str,
+        retry_config: RetryConfig,
+    ) -> crate::Result<reqwest::Response> {
+        let sleep = async |d| tokio::time::sleep(d).await;
+
+        if !retry_config.has_retry_config() {
+            return self.send(request, error_message).await;
+        }
```
dbolduc: nit: This doesn't seem like a special case. I think we should always use a retry loop, and the default settings are policies that don't retry. That is less complexity IMO.

src/auth/src/mds/client.rs
```diff
@@ -123,6 +134,55 @@ impl Client {
...
+                    let err_headers = response.headers().clone();
+                    let err_payload = response.bytes().await.map_err(|e| {
+                        google_cloud_gax::error::Error::transport(err_headers.clone(), e)
+                    })?;
+                    return Err(google_cloud_gax::error::Error::http(
+                        status.as_u16(),
+                        err_headers,
+                        err_payload,
+                    ));
+                }
```
dbolduc: why are we duping the `send(...)` code? can we reuse it?

src/auth/src/mds/client.rs
```diff
@@ -105,6 +115,7 @@ impl Client {
     pub(crate) fn universe_domain(&self) -> UniverseDomainRequest {
         UniverseDomainRequest {
             client: self.clone(),
+            retry_config: RetryConfig::default(),
```
dbolduc: what about creating the policies once, and storing them on the client?

alvarowolfx: we want to have different policies per call, because the MDS Client is reused on the MDS Credentials provider. For calls to fetch tokens we don't want retries, because the TokenCache already handles retries. We only want retries here for universe_domain calls

src/auth/src/mds/client.rs
alvarowolfx: `send(...)` maps errors to Credentials Errors and is simpler (we don't need to clone requests). On the retry loop we need to map to gax errors and clone requests. I tried to reuse both, but code was not looking nicer. We could make `send(...)` already return gax errors and map later to credentials errors, but we can loose some information as gax errors are more specific

src/auth/src/mds/client.rs
dbolduc: What I want is to allocate the policies once.

If we only make one attempt to get the universe_domain per credentials, then it's fine to do it on that call.

alvarowolfx:
Split `from_gax_error` method into PR #5350

src/auth/src/mds/client.rs
```diff
@@ -105,37 +113,98 @@ impl Client {
...
+        let error_message_str = error_message.to_string().clone();
+        retry_loop(
+            async move |_| {
+                let req = request
+                    .try_clone()
+                    .expect("client libraries only create builders where `try_clone()` succeeds");
+                let response = req
+                    .send()
+                    .await
+                    .map_err(google_cloud_gax::error::Error::io)?;
```
dbolduc: style nit: a `use google_cloud_gax::error::Error as GaxError` could save you some typing.

src/auth/src/mds/client.rs
```diff
@@ -105,37 +113,98 @@ impl Client {
...
+    backoff_policy: BackoffPolicyArg,
+    retry_throttler: SharedRetryThrottler,
+}
+
+impl Default for RetryConfig {
+    fn default() -> Self {
+        Self {
+            retry_policy: NeverRetry.into(),
+            backoff_policy: ExponentialBackoff::default().into(),
+            retry_throttler: Arc::new(Mutex::new(AdaptiveThrottler::default())),
```
dbolduc, line 185: nit: we are allocating these on every call to `send()`. We only need to do it once per MDS client.

src/auth/src/mds/client.rs
```diff
@@ -479,4 +581,76 @@ mod tests {
...
+            .times(2)
+            .respond_with(cycle(responses)),
+        );
+
+        let retry_policy = AlwaysRetry.with_attempt_limit(2);
+        let backoff_policy = ExponentialBackoffBuilder::new()
+            .with_initial_delay(Duration::from_millis(1))
+            .with_maximum_delay(Duration::from_millis(1))
+            .build()
+            .unwrap();
```
dbolduc: nit: prefer `?` over `unwrap()`.

src/auth/src/mds/client.rs
```diff
@@ -479,4 +581,76 @@ mod tests {
...
+
+        let domain = client
+            .universe_domain()
+            .with_retry_policy(retry_policy.into())
+            .with_backoff_policy(backoff_policy.into())
+            .send()
+            .await
+            .unwrap();
```
dbolduc: same

src/auth/src/mds/client.rs
```diff
@@ -479,4 +581,76 @@ mod tests {
...
+            Expectation::matching(all_of![
+                request::method("GET"),
+                request::path(MDS_UNIVERSE_DOMAIN_URI),
+            ])
+            .times(2)
+            .respond_with(cycle(responses)),
+        );
+
+        let retry_policy = AlwaysRetry.with_attempt_limit(2);
+        let backoff_policy = ExponentialBackoffBuilder::new()
```
dbolduc: I'd use a `MockBackoff` that returns `Duration::ZERO`. That way we know the backoff policy gets called.

src/auth/src/mds/client.rs
```diff
@@ -479,4 +581,76 @@ mod tests {
...
+
+        let err = client
+            .universe_domain()
+            .with_retry_policy(retry_policy.into())
+            .with_backoff_policy(backoff_policy.into())
+            .send()
+            .await
+            .unwrap_err();
+
+        assert!(err.to_string().contains("failed to fetch universe domain"));
```
dbolduc, line 641: Q: do we expect applications to also match on the contents of this string?

alvarowolfx: the `universe_domain` method in the `CredentialsProvider` trait never returns an error, it's an `Option<String>`. So no, customers are not gonna match this string.

src/auth/src/mds/client.rs
alvarowolfx: good call, I'm gonna reuse other existing mocks for backoff, retry and throttler from the auth package

src/auth/src/mds/client.rs
alvarowolfx: But we would need multiple MDS Clients on the MDS Credential, because some calls (access_token, id_token) should not have retries and `universe_domain` needs retries. That's why the retry is allocated per call.

dbolduc: Why can't you store the `backoff_policy` and `retry_throttler` on the client and clone that for each `send()`? You are currently making new ones for every `send()` call, right? Am I missing something?

src/auth/src/mds/client.rs
dbolduc: Oh, so the error is just for us and maybe future logs when we write them.

src/auth/src/mds/client.rs
alvarowolfx: I think I got what you're saying, we are cloning `Client` anyway, so would make sense for the retry to live in the Client and be cloned from it and self contained, since the `send()` methods belongs to `Client`

PR #5317, opened 2026-04-07
impl(auth): add universe_domain support for External Accounts

alvarowolfx:

### DIFF
--- a/src/auth/src/credentials/external_account.rs
+++ b/src/auth/src/credentials/external_account.rs
@@ -1605,13 +1605,13 @@ mod tests {
         });
 
         let creds = Builder::new(contents)
-            .with_universe_domain("my-overrided-universe.com")
+            .with_universe_domain("my-overridden-universe.com")
             .build()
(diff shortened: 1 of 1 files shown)


alvarowolfx:
> Why do I smell "there should be a decorator trying to burst out of this thing"?

@coryan you have a point and I considered that since we recently have done this for access boundaries. But each credential type has a different source for the universe domain, so I think still makes sense for the resolution to be part of the credential provider. And the `universe_domain` method is already part of the `CredentialsProvider` trait.

* Service Accounts and External Accounts: read from their key file (each has a different format) 
* Impersonated: Should lookup inside source credential input (via file or programmatic) if a universe_domain was provided.
* MDS: Fetch from MDS endpoint. 
* User accounts: not supported

For access boundaries it was a bit different, since the source for fetching boundaries was mostly the same (only MDS is a bit different). But will think a bit more. The API is still not being made public, and this change can be made internally, so we can still change things up.

PR #5318, opened 2026-04-07
chore: force `h2` version to avoid minimal-versions deadlock

haphungw:
Fix a `gcb-pr-minimal-versions` build failure in `google-cloud-gax-internal` tests caused by `h2` version `0.4.2`, which surfaces after `google_cloud_unstable_tracing` guards are removed in #5292.

The Problem
During minimal-versions checks, `h2` resolves to `0.4.2`, which somehow leads to deadlocks or missing tracing spans in our gRPC tests.

Previous Attempts
I tried switching tests to `multi_thread` (PR #5308), which masked the issue but was not a good solution.
I tried forcing the version in the root `[workspace.dependencies]`, but `Cargo` ignored it during the isolated package build for `gax-internal`.

The Fix
Add `h2 = "0.4.13"` as a direct dependency in `src/gax-internal/Cargo.toml`. `0.4.13` is the version resolved in our normal Cargo.lock.

### DIFF
--- a/src/gax-internal/Cargo.toml
+++ b/src/gax-internal/Cargo.toml
@@ -47,6 +47,7 @@ _internal-http-client = ["_internal-common", "dep:hyper", "dep:reqwest", "dep:se
 # features of reqwest that `google-cloud-auth` does.
 _internal-grpc-client = [
   "_internal-common",
+  "dep:h2",
   "dep:http-body",
   "dep:prost",
   "dep:prost-types",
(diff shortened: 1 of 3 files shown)


src/gax-internal/Cargo.toml
```diff
@@ -95,6 +95,7 @@ _default-rustls-provider = ["google-cloud-auth?/default-rustls-provider", "tonic
 _internal-grpc-server-streaming = []
 
 [dependencies]
+h2                                 = "0.4.13"
```
coryan: Even if only used in one package, we want this in the top-level Cargo.toml file.

haphungw: done. thank you 🙇‍♀️

PR #5321, opened 2026-04-08
perf(spanner): inline begin transaction error handling (step 2)

olavloite:
Adds error handling for inline-begin-transaction. If the first statement in a
transaction fails, and that statement included a BeginTransaction option, then
the transaction has not been started. In order to keep the semantics of the
transaction consistent for an 'outside observer', we need to do the following:
1. Catch the error that was thrown by the initial statement.
2. Start the transaction using an explicit BeginTransaction RPC.
3. Retry the initial statement, but now using the transaction ID from step 2.
4. Return the error or result for the retried initial statement.

The above makes sure that:
1. The transaction is actually started when the first statement is executed, also
   when the statement failed.
2. The statement becomes part of the transaction, and the result of the statement
   is consistent with the read-timestamp of the transaction.

The second part is important in order to comply with Spanner's strong consistency
guarantees; If for example a statement returns a 'Table not found' error, then
that error is only valid for the read timestamp that was used for executing the
statement. This is the reason that we retry the statement after the BeginTransaction
RPC to be able to return a result that is guaranteed to be consistent with any
other queries/reads that will be executed in the same transaction.

This is step 2 of the implementation of inline-begin-transaction for Spanner. A preliminary view of the full implementation after all steps can be seen in https://github.com/googleapis/google-cloud-rust/pull/5307


dbolduc:
@olavloite can you have someone from @googleapis/spanner-team review this PR? I want to test if y'all can approve and merge PRs that only affect spanner. (#5341)

dbolduc:
Nice, looks like the shared ownership works. :slightly_smiling_face:

PR #5322, opened 2026-04-08
test(spanner): add integration test for inline-begin error handling (step 3)

olavloite:
Adds an integration test for error handling for inline-begin-transaction. This test uses
a gRPC proxy to intercept calls from the client to Spanner to be able to deterministically
emulate specific concurrency issues. This test shows how a query that failed during the
first attempt, and thereby also failed to start the transaction, could succeed during a
retry after the transaction has been started with an explicit BeginTransaction RPC.

This is step 3 of the implementation of inline-begin-transaction for Spanner. A preliminary view of the full implementation after all steps can be seen in https://github.com/googleapis/google-cloud-rust/pull/5307

### DIFF
--- a/tests/spanner/src/query.rs
+++ b/tests/spanner/src/query.rs
@@ -14,7 +14,7 @@
 
 use crate::client::{get_database_id, get_emulator_host};
 use crate::test_proxy::{InterceptedSpanner, SpannerInterceptor};
-use google_cloud_spanner::client::{DatabaseClient, Kind, Spanner, Statement};
+use google_cloud_spanner::client::{DatabaseClient, Kind, QueryOptions, Spanner, Statement};
 use google_cloud_test_utils::resource_names::LowercaseAlphanumeric;
 use spanner_grpc_mock::google::spanner::v1 as spanner_v1;
 use spanner_grpc_mock::google::spanner::v1::spanner_client::SpannerClient;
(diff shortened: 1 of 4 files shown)


tests/spanner/src/query.rs
```diff
@@ -408,3 +419,143 @@ fn verify_row_2(row: &google_cloud_spanner::client::Row) {
...
+    let local_addr = listener.local_addr()?;
+    let proxy_address = format!("{}:{}", local_addr.ip(), local_addr.port());
+
+    let proxy = DelayedBeginProxy {
+        emulator_client: raw_client,
+        latch: Arc::clone(&latch),
+        begin_transaction_entered_latch: Arc::clone(&begin_transaction_entered_latch),
+    };
+
+    let _server_handle = tokio::spawn(async move {
```
sakthivelmanii: should we shutdown the server handle when test is completed?

src/spanner/grpc-mock/src/lib.rs
```diff
@@ -64,6 +64,7 @@ pub mod google {
         include!("generated/protos/google.rpc.rs");
     }
     pub mod spanner {
+        #[allow(rustdoc::broken_intra_doc_links, rustdoc::bare_urls)]
```
dbolduc: `grpc-mock` is fully generated since https://github.com/googleapis/google-cloud-rust/pull/5433

If we need these `#[allow]`s, let me know, and I'll update the template

tests/spanner/src/query.rs
olavloite: Yeah, good point. I added that in https://github.com/googleapis/google-cloud-rust/pull/5322/changes/1d21817e054dc334ac866b65bb00660484ac20c0

olavloite:
@dbolduc This PR seems to require a review from you or one of the other owners.

dbolduc:
> @dbolduc This PR seems to require a review from you or one of the other owners.

Yeah, only my team owns the `deny.toml`

PR #5324, opened 2026-04-08
fix(spanner): prevent deadlock if ResultSet#next() is never called (step 5)

olavloite:
If a query included a BeginTransaction option and the application never called ResultSet#next(),
then the transaction ID would never be returned. This would block any other query from using
the transaction. This change refactors ResultSet to use a background worker to read from the
stream. This prevents that a deadlock can happen if the application does not call ResultSet#next().
It also allows the application to call ResultSet#metadata() without first calling ResultSet#next().
Finally, it also allows the ResultSet to decode data from the server asynchronously while the
application processes rows that it has already read from the ResultSet.


dbolduc:
(not looking at the code, just the description)

> without first calling ResultSet#next().

Do you think the problem is: https://github.com/hyperium/tonic/issues/515 ? (That a streaming read over tonic never yields until the first message is available) (which is a bug in the implementation of gRPC if you ask me).

If so, I ran into this in Pub/Sub and decided to lazy-initialize my streaming read response. Do you prefer that to buffering reads in a background task?.

https://github.com/googleapis/google-cloud-rust/blob/b175db6fdb8c3de63d023c6a96d0f0899a25aaa3/src/pubsub/src/subscriber/message_stream.rs#L90-L97

In Pub/Sub it made more sense, because we do not necessarily receive a message until someone else publishes one. But in Spanner if you are running a query, you are expected to get results soonish.

olavloite:
@dbolduc 

> Do you think the problem is: https://github.com/hyperium/tonic/issues/515 ?

No, this is purely a limitation in the design for this Spanner library. I should maybe not have added a 'fix' prefix. It boils down to the following:
1. We are implementing 'inline-begin-transaction' in this library. This is an optimization that we have in all of our client libraries.
2. In the Rust library, we have also made one other design decision that **does** deviate from the client libraries in the other languages: In Rust, When you call `execute_query`, we initiate the query stream directly. In the other languages, we delay the actual stream creation until the first call to `ResultSet#next()`. The latter often causes confusion among customers who are just doing a quick benchmark of either Spanner or the library. Their benchmark only calls `execute_query`, which finishes in a few microseconds, and never calls `ResultSet#next()`. This design also causes any query errors to only show up at the first call to `ResultSet#next()`. So to prevent all of that in Rust, we have decided to initiate the stream directly when `execute_query` is called.
3. The above design decision in combination with inline-begin-transaction means that if an application never calls `ResultSet#next()`, the client library will also never get the transaction ID, as it is returned together with the first query result.
4. That again could cause a deadlock if other queries were also started for the transaction, as these would be waiting for the transaction ID.

olavloite:
Closing in favor of #5654

PR #5325, opened 2026-04-08
perf(spanner): inline-begin for read/write transactions (step 6)

olavloite:
Adds support for inline-begin for read/write transactions. This reduces the
number of round-trips to Spanner by one for read/write transactions.


olavloite:
Replaced by #5662

PR #5327, opened 2026-04-08
chore(spanner): change begin transaction option to enum (step 8)

olavloite:
Change the begin-transaction option from a boolean to an enum to allow more values in
the future.


src/spanner/src/read_only_transaction.rs
```diff
@@ -186,6 +192,17 @@ impl SingleUseReadOnlyTransaction {
...
+/// Options for how to start a transaction.
+#[derive(Clone, Copy, Debug, PartialEq, Eq, Default)]
+#[non_exhaustive]
+pub enum BeginTransactionOption {
+    /// The transaction will be started inlined with the first statement.
+    #[default]
+    InlineBegin,
+    /// The transaction will be started explicitly using a `BeginTransaction` RPC.
+    ExplicitBegin,
+}
```
suzmue, line 204: drive by nit: Consider the following.

1. Renaming `BeginTransactionOption` to `BeginTransactionMode` or just `BeginTransaction`. This avoids confusion with std::option::Option and is more consistent with other  configuration options we have in google-cloud-rust.
2. Drop the `Begin` suffix on the two variants, which are redundant with the enum name.

The new code using this would then look like 
```rust
with_begin_transaction_mode(BeginTransactionMode::Inline)
```

PR #5328, opened 2026-04-08
impl(pubsub): prepare retry loop for confirmed_ack in Leaser

PhongChuong:
Update confirmed_ack to prepare for implementing retry_loop. This PR does not change existing behavior as retry_loop is set to NeverRetry.

Towards #4804

### DIFF
--- a/src/pubsub/src/subscriber/leaser.rs
+++ b/src/pubsub/src/subscriber/leaser.rs
@@ -144,14 +144,13 @@ where
         let leaser = self.clone();
         let mut ack_ids = ack_ids;
 
-        // Returns `Result<()>`.
         let attempt = async move |_| {
             let ids = std::mem::take(&mut ack_ids);
             let ack_ids = leaser.confirmed_ack_attempt(ids).await;
             if ack_ids.is_empty() {
                 Ok(())
             } else {
-                // Return a synthetic error to force retry.
+                // Return a synthetic error to indicate that we should retry.
                 Err(crate::Error::timeout("retry me"))
             }
         };


src/pubsub/src/subscriber/leaser.rs
```diff
@@ -137,24 +138,60 @@ where
...
+            false,
+            Arc::new(Mutex::new(
+                google_cloud_gax::retry_throttler::AdaptiveThrottler::default(),
+            )),
+            Arc::new(google_cloud_gax::retry_policy::NeverRetry),
+            Arc::new(google_cloud_gax::exponential_backoff::ExponentialBackoff::default()),
+        )
+        .await;
+
+        let _ = self.confirmed_tx.send(
```
dbolduc: quick feedback: we want to send partial results from inside the `inner` function of the loop instead of waiting for it to finish.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -137,24 +138,60 @@ where
...
+        let sleep = async |d| tokio::time::sleep(d).await;
+        // TODO(#4804): update retry_loop configuration and move to retry_policy.
+        // Result is ignored as inner always return Ok.
+        let _ = retry_loop(
+            inner,
+            sleep,
+            false,
+            Arc::new(Mutex::new(
+                google_cloud_gax::retry_throttler::AdaptiveThrottler::default(),
+            )),
```
dbolduc: `options` has one of these. I think we could clone it and save some allocations on every request.

Aside: the stub has a throttler, it seems sad to have a second layer of throttling, but probably better than rewriting the retry loop logic.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -137,24 +138,60 @@ where
...
-                )
-            })
-            .collect();
-        let _ = self.confirmed_tx.send(confirmed_acks);
+        use std::time::Duration;
+
+        let subscription = self.subscription.clone();
+        let stub = self.inner.clone();
+        let options = self.options.clone();
+        let confirmed_acks = Arc::new(Mutex::new(HashMap::<String, AckResult>::new()));
```
dbolduc: (I haven't tried, but I think) the `Arc`, `Mutex` here should be avoidable because the retry loop executes sequentially?

I imagine we could have some:

```rs
        let leaser = self.clone();
        let mut ack_ids = ack_ids;
        let attempt = async move |_| {
            let ids = std::mem::take(&mut ack_ids);
            let ack_ids = leaser.confirmed_ack_one_attempt(ids).await;
            if ack_ids.is_empty() {
                Ok(())
            } else {
                // Return a synthetic error. Better than rewriting the retry loop. :shrug:
                Err(crate::Error::timeout("retry me"))
            }
        };
```

with

```rs
impl DefaultLeaser {
  fn confirmed_ack_one_attempt(&self, ack_ids: Vec<String>) -> Vec<String> {
    // basically `confirmed_ack()` now, but where it returns ack IDs we need to retry at the end.
  }
}
```

It's a little goofy to have a synthetic error like this. We are sort of hacking the retry loop. But if it works, seems preferable to introducing `Mutex`es

dbolduc: Ok, I didn't trust myself, so I coded it to make sure it works:

https://github.com/dbolduc/google-cloud-rust/commit/f37f0114d0fcfd0567b5158bb832a4e676059739

Note that it is nice to isolate the logic for a single attempt from the retry loop.

PhongChuong: Thanks for the code suggestion. Moving the attempt to a helper function is a lot cleaner. It didn't occur to me that we can do: `self.clone();` to pass into the closure.

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Updated to use the throttler in self.options. PTAL.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -137,7 +141,58 @@ where
     /// keep the retry logic in the leaser, while allowing for partial results
     /// to be reported before the entire operation completes.
     async fn confirmed_ack(&self, ack_ids: Vec<String>) {
-        // TODO(#4804): implement retries
+        let leaser = self.clone();
+        let mut ack_ids = ack_ids;
+
+        // Returns `Result<()>`.
```
dbolduc: I'd remove this. This was a mental note for me that I never cleaned up.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -137,7 +141,58 @@ where
...
+        };
+
+        let sleep = async |d| tokio::time::sleep(d).await;
+        let _ = retry_loop(
+            attempt,
+            sleep,
+            true,
+            retry_throttler(&self.options),
+            retry_policy(),
+            backoff_policy(),
```
dbolduc, line 165: For all of these, I think ideally we want the leaser to hold them, as a minor optimization so we don't have to allocate them for each request.

But I think you will change these in follow up PRs anyway (e.g. to use a `MockBackoff` policy in tests)... so it's fine for now.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -137,7 +141,58 @@ where
...
+
+        // Returns `Result<()>`.
+        let attempt = async move |_| {
+            let ids = std::mem::take(&mut ack_ids);
+            let ack_ids = leaser.confirmed_ack_attempt(ids).await;
+            if ack_ids.is_empty() {
+                Ok(())
+            } else {
+                // Return a synthetic error to force retry.
+                Err(crate::Error::timeout("retry me"))
```
dbolduc, line 154: Aside: `Error::timeout` was just the quickest thing I could think of.

We used to have an `Error::other(...)` but I think that got removed.

I am fine with this because no one consumes the error.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -137,7 +141,58 @@ where
...
+        let mut ack_ids = ack_ids;
+
+        // Returns `Result<()>`.
+        let attempt = async move |_| {
+            let ids = std::mem::take(&mut ack_ids);
+            let ack_ids = leaser.confirmed_ack_attempt(ids).await;
+            if ack_ids.is_empty() {
+                Ok(())
+            } else {
+                // Return a synthetic error to force retry.
```
dbolduc: nit: maybe "indicate we should try again" instead of "force retry" ?

Maybe. It depends on how we cut off the retry loop. (Like whether the retry policy is responsible, or if we handle it manually).

We will need to figure out how to satisfy the remaining ack IDs in the case of the retry loop being exhausted. It could be something dumb like passing in a `final_attempt: bool` to `confirmed_ack_one_attempt()`. (IDK, I haven't thought this far ahead)

dbolduc: Another option is to have some custom error like:

```rs
// This thing impls std::error::Error
struct UnresolvedAckIds {
  ack_ids: Vec<String>,
  final_status: Arc<crate::Error>,
}
```

and then the retry policy shoves that in a `crate::Error::io()` if it is exhausted. And outside of the retry loop, we extract the error, and satisfy the final ack_ids.

^ This seems very convoluted. Hopefully we can do better.

PhongChuong: I was thinking of having a shared ack_ids that is updated with each attempt. After the retry_loop, we can check to see if there are any ids remaining and act accordingly (most likely return some retry exhausted error)

dbolduc: > (most likely return some retry exhausted error)

Yeah. This typically contains the final transient error from the RPC as its `source()`.

The RPC error never leaves `confirmed_ack_one_attempt()`, so I feel like we will need to either bubble that up or cut the loop off in `confirmed_ack_one_attempt()`. I am not sure.

> shared ack_ids

I feel like we should be able to design around shared data for a sequential operation like this.

PR #5329, opened 2026-04-08
impl(auth): add universe_domain support for MDS

alvarowolfx:

### DIFF
--- a/src/auth/src/credentials/mds.rs
+++ b/src/auth/src/credentials/mds.rs
@@ -403,9 +403,21 @@ where
 
         // No overrides and no cache. Try to fetch from MDS.
         let response = self.mds_client.universe_domain().send().await;
-        let universe_domain = response.ok();
-        let _ = self.universe_domain.set(universe_domain.clone());
-        universe_domain
+        match response {
+            Ok(universe_domain) => {
+                let _ = self.universe_domain.set(Some(universe_domain.clone()));
+                Some(universe_domain)
+            }
+            Err(e) => {
+                if !e.is_transient() {
+                    // Only cache None if the error is permanent (e.g., 404 on GDU)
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/mds.rs
```diff
@@ -375,6 +392,21 @@ where
...
+            return Some(ud.clone());
+        }
+        if let Some(ud) = self.universe_domain.get() {
+            return ud.clone();
+        }
+
+        // No overrides and no cache. Try to fetch from MDS.
+        let response = self.mds_client.universe_domain().send().await;
+        let universe_domain = response.ok();
+        let _ = self.universe_domain.set(universe_domain.clone());
```
suzmue: If there is an error, the `None` will get cached indefinitely since we can't overwrite it due to the OnceLock. Even though the fetching code will be run again it won't be actually able to store future successful results.

Can we also add a test covering this case?

suzmue: Nevermind it seems like this is intentional, and the Some() will hold an option inside it.

A test covering this would be nice though anyway.

alvarowolfx: This is going to be improved with retries. It's a tricky think since in normal environments (GDU - googleapis.com), this call will fail, since the universe domain endpoint doesn't exist (404) on MDS. But we don't want to keep hitting it on a GDU, since in some parts it can make calls slower if is always trying to resolve the universe domain. With retries, I'm only gonna save on the OnceLock if the error is not retryable. I'll at least add a comment saying that

alvarowolfx: actually, I can already write that code, since the error being returned is the same and I already can check if is transient. Working on it

PR #5352, opened 2026-04-09
feat(storage): enable transport-level tracing for gRPC

haphungw:
Update `from_parts` to use `Client::new_with_instrumentation()`.

Before, the http client was instrumented but the gRPC client was not, even when tracing was enabled. This resulted in missing transport-level spans for gRPC operations.


coryan (approved):
Yikes! Thanks!

PR #5354, opened 2026-04-09
test(pubsub): add sample for publisher batch settings

suzmue:
Fixes #5348

### DIFF
--- a/src/pubsub/examples/src/publisher/publish_with_batch_settings.rs
+++ b/src/pubsub/examples/src/publisher/publish_with_batch_settings.rs
@@ -31,10 +31,9 @@ pub async fn sample(project_id: &str, topic_id: &str) -> anyhow::Result<()> {
         .build()
         .await?;
 
-    let mut futures = vec![];
-    for i in 0..10 {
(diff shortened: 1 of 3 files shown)


src/pubsub/examples/src/publisher/publish_with_batch_settings.rs
```diff
@@ -0,0 +1,48 @@
...
+    // - total size reaches 1 KiB, OR
+    // - 100 milliseconds have passed.
+    let publisher = Publisher::builder(topic_name)
+        .set_message_count_threshold(10)
+        .set_byte_threshold(1024) // 1 KiB
+        .set_delay_threshold(Duration::from_millis(100))
+        .build()
+        .await?;
+
+    let mut futures = vec![];
```
dbolduc: super nit: I more commonly avoid the macro with:

```suggestion
    let mut futures = Vec::new();
```

Or use `.map(...).collect()` to one line it.

dbolduc (approved):

src/pubsub/examples/src/publisher/publish_with_batch_settings.rs
suzmue: done.

PR #5357, opened 2026-04-10
impl(pubsub): handle ErrorInfo in a single confirmed_ack attempt

PhongChuong:
Process server response ErrorInfo in process_ack_attempt to potentially retry ack_ids with transient errors or return Permanent ack error.

Towards #4804

### DIFF
--- a/src/pubsub/src/error.rs
+++ b/src/pubsub/src/error.rs
@@ -84,20 +84,6 @@ pub enum AckError {
     /// the message, the service may or may not redeliver it.
     #[error("error during shutdown. The result of the operation is unknown. {0}")]
     Shutdown(#[source] Box<dyn std::error::Error + Send + Sync + 'static>),
-
-    /// Permanent failure from the Pub/Sub backend server.
-    ///
-    /// Do not retry.
-    #[error("permanent error from the Pub/Sub backend server: {0}. Do not retry.")]
-    Permanent(String),
-
-    /// Transient failure from the Pub/Sub backend server.
-    ///
-    /// The client has exhausted all retry attempts.
-    #[error(
-        "transient error from the Pub/Sub backend server: {0}. The client has exhausted all retry attempts."
-    )]
-    Transient(String),
 }
 
 #[cfg(test)]
--- a/src/pubsub/src/error.rs
+++ b/src/pubsub/src/error.rs
@@ -84,6 +84,20 @@ pub enum AckError {
     /// the message, the service may or may not redeliver it.
     #[error("error during shutdown. The result of the operation is unknown. {0}")]
     Shutdown(#[source] Box<dyn std::error::Error + Send + Sync + 'static>),
+
+    /// Permanent failure from the Pub/Sub backend server.
+    ///
+    /// Do not retry.
+    #[error("permanent error from the Pub/Sub backend server: {0}. Do not retry.")]
+    Permanent(String),
+
+    /// Transient failure from the Pub/Sub backend server.
+    ///
+    /// The client has exhausted all retry attempts.
+    #[error(
+        "transient error from the Pub/Sub backend server: {0}. The client has exhausted all retry attempts."
+    )]
+    Transient(String),
 }
 
 #[cfg(test)]
--- a/src/pubsub/src/subscriber/leaser.rs
+++ b/src/pubsub/src/subscriber/leaser.rs
@@ -18,12 +18,13 @@ use super::stub::Stub;
 use crate::RequestOptions;
 use crate::error::AckError;
 use crate::model::{AcknowledgeRequest, ModifyAckDeadlineRequest};
+use crate::{Response, Result};
 use google_cloud_gax::error::rpc::StatusDetails;
 use google_cloud_gax::exponential_backoff::ExponentialBackoff;
 use google_cloud_gax::retry_loop_internal::retry_loop;
 use google_cloud_gax::retry_policy::NeverRetry;
 use google_cloud_gax::retry_throttler::CircuitBreaker;
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::sync::{Arc, Mutex};
 use tokio::sync::mpsc::UnboundedSender;
 
@@ -154,7 +155,7 @@ where
                 .set_ack_ids(ids.clone());
             let response = leaser.inner.acknowledge(req, leaser.options.clone()).await;
 
-            let (to_confirm, remaining) = Self::process_ack_attempt(ids, response);
+            let (to_confirm, remaining) = process_ack_attempt(ids, response);
             let _ = leaser.confirmed_tx.send(to_confirm);
             if remaining.is_empty() {
                 Ok(())
@@ -225,89 +226,79 @@ fn retry_throttler(
     })
 }
 
-impl<T> DefaultLeaser<T>
-where
-    T: Stub + 'static,
-{
-    fn process_ack_attempt(
-        ack_ids: Vec<String>,
-        response: std::result::Result<crate::Response<()>, google_cloud_gax::error::Error>,
-    ) -> (HashMap<String, AckResult>, Vec<String>) {
-        let e = match response {
-            Ok(_) => {
-                let to_confirm = ack_ids.into_iter().map(|id| (id, Ok(()))).collect();
-                return (to_confirm, Vec::new());
-            }
-            Err(e) => e,
-        };
-
-        let (transient_failures, permanent_failures) = Self::extract_failures(&e);
-        let shared_err = Arc::new(e);
-
-        // If the response lacks specific per ack_id failure info, we treat the response as all sharing the same
-        // RPC error.
-        if transient_failures.is_empty() && permanent_failures.is_empty() {
-            let to_confirm = ack_ids
-                .into_iter()
-                .map(|id| {
-                    (
-                        id,
-                        Err(AckError::Rpc {
-                            source: shared_err.clone(),
-                        }),
-                    )
-                })
-                .collect();
+fn process_ack_attempt(
+    ack_ids: Vec<String>,
+    response: Result<Response<()>>,
+) -> (HashMap<String, AckResult>, Vec<String>) {
+    let e = match response {
+        Ok(_) => {
+            let to_confirm = ack_ids.into_iter().map(|id| (id, Ok(()))).collect();
             return (to_confirm, Vec::new());
         }
+        Err(e) => e,
+    };
 
-        // Otherwise, we extract specific failures:
-        // - ack_ids with transient failures are to be retried.
-        // - ack_ids with permanent failures are resolved with the RPC error.
-        // - Unlisted ack_ids are considered successfully acknowledged.
-        let mut transient = Vec::new();
-        let mut to_confirm = HashMap::new();
-        for id in ack_ids {
-            if transient_failures.contains(&id) {
-                transient.push(id);
-            } else if permanent_failures.contains(&id) {
-                to_confirm.insert(
+    let (transient_failures, permanent_failures) = extract_failures(&e);
+    let shared_err = Arc::new(e);
+
+    // If the response lacks specific per ack_id failure info, we treat the response as all sharing the same
+    // RPC error.
+    if transient_failures.is_empty() && permanent_failures.is_empty() {
+        let to_confirm = ack_ids
+            .into_iter()
+            .map(|id| {
+                (
                     id,
                     Err(AckError::Rpc {
                         source: shared_err.clone(),
                     }),
-                );
-            } else {
-                to_confirm.insert(id, Ok(()));
-            }
-        }
-
-        (to_confirm, transient)
+                )
+            })
+            .collect();
+        return (to_confirm, Vec::new());
     }
 
-    fn extract_failures(
-        e: &crate::Error,
-    ) -> (
-        std::collections::HashSet<String>,
-        std::collections::HashSet<String>,
-    ) {
-        let mut transient = std::collections::HashSet::new();
-        let mut permanent = std::collections::HashSet::new();
-        if let Some(status) = e.status() {
-            for detail in &status.details {
-                if let StatusDetails::ErrorInfo(info) = detail {
-                    for (k, v) in &info.metadata {
-                        if v.starts_with("TRANSIENT_FAILURE_") {
-                            transient.insert(k.clone());
-                        } else if v.starts_with("PERMANENT_FAILURE_") {
-                            permanent.insert(k.clone());
-                        }
+    // Otherwise, we extract specific failures:
+    // - ack_ids with transient failures are to be retried.
+    // - ack_ids with permanent failures are resolved with the RPC error.
+    // - Unlisted ack_ids are considered successfully acknowledged.
+    let mut transient = Vec::new();
+    let mut to_confirm = HashMap::new();
+    for id in ack_ids {
+        if transient_failures.contains(&id) {
+            transient.push(id);
+        } else if permanent_failures.contains(&id) {
+            to_confirm.insert(
+                id,
+                Err(AckError::Rpc {
+                    source: shared_err.clone(),
+                }),
+            );
+        } else {
+            to_confirm.insert(id, Ok(()));
+        }
+    }
+
+    (to_confirm, transient)
+}
+
+fn extract_failures(e: &crate::Error) -> (HashSet<String>, HashSet<String>) {
+    let mut transient = HashSet::new();
+    let mut permanent = HashSet::new();
+    if let Some(status) = e.status() {
+        for detail in &status.details {
+            if let StatusDetails::ErrorInfo(info) = detail {
+                for (k, v) in &info.metadata {
+                    if v.starts_with("TRANSIENT_FAILURE_") {
+                        transient.insert(k.clone());
+                    } else if v.starts_with("PERMANENT_FAILURE_") {
+                        permanent.insert(k.clone());
                     }
                 }
             }
         }
-        (transient, permanent)
     }
+    (transient, permanent)
 }
 
 #[cfg(test)]
@@ -318,7 +309,7 @@ pub(super) mod tests {
     use super::*;
     use crate::{Error, Response};
     use google_cloud_gax::error::rpc::{Code, Status};
-    use std::collections::HashSet;
+    use google_cloud_rpc::model::ErrorInfo;
     use tokio::sync::Mutex;
     use tokio::sync::mpsc::unbounded_channel;
 
@@ -389,6 +380,15 @@ pub(super) mod tests {
         }
     }
 
+    fn response_with_error_info(info: ErrorInfo) -> Result<Response<()>> {
+        Err(Error::service(
+            Status::default()
+                .set_code(Code::FailedPrecondition)
+                .set_message("fail")
+                .set_details(vec![StatusDetails::ErrorInfo(info)]),
+        ))
+    }
+
     #[test]
     fn clone() {
         let (confirmed_tx, _confirmed_rx) = unbounded_channel();
@@ -409,7 +409,7 @@ pub(super) mod tests {
 
     #[test]
     fn extract_failures() {
-        let info = google_cloud_rpc::model::ErrorInfo::new()
+        let info = ErrorInfo::new()
             .set_reason("reason")
             .set_domain("domain")
             .set_metadata([
@@ -420,12 +420,8 @@ pub(super) mod tests {
                 ("ack_5", "PERMANENT_FAILURE_OTHER"),
             ]);
 
-        let status = Status::default()
-            .set_code(Code::FailedPrecondition)
-            .set_message("fail")
-            .set_details(vec![StatusDetails::ErrorInfo(info)]);
-        let err = Error::service(status);
-        let (transient, permanent) = DefaultLeaser::<MockStub>::extract_failures(&err);
+        let err = response_with_error_info(info).unwrap_err();
+        let (transient, permanent) = super::extract_failures(&err);
 
         assert_eq!(
             transient,
@@ -435,6 +431,26 @@ pub(super) mod tests {
             permanent,
             HashSet::from(["ack_4".to_string(), "ack_5".to_string()])
         );
+    }
+
+    #[test]
+    fn extract_failures_multiple_error_info() {
+        let info1 =
+            ErrorInfo::new().set_metadata([("ack_1", "TRANSIENT_FAILURE_UNORDERED_ACK_ID")]);
+        let info2 = ErrorInfo::new().set_metadata([("ack_2", "PERMANENT_FAILURE_INVALID_ACK_ID")]);
+
+        let status = Status::default()
+            .set_code(Code::FailedPrecondition)
+            .set_message("fail")
+            .set_details(vec![
+                StatusDetails::ErrorInfo(info1),
+                StatusDetails::ErrorInfo(info2),
+            ]);
+        let err = Error::service(status);
+        let (transient, permanent) = super::extract_failures(&err);
+
+        assert_eq!(transient, HashSet::from(["ack_1".to_string()]));
+        assert_eq!(permanent, HashSet::from(["ack_2".to_string()]));
     }
 
     #[tokio::test]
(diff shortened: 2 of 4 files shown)


src/pubsub/src/error.rs
```diff
@@ -84,6 +84,20 @@ pub enum AckError {
...
+    #[error("permanent error from the Pub/Sub backend server: {0}. Do not retry.")]
+    Permanent(String),
+
+    /// Transient failure from the Pub/Sub backend server.
+    ///
+    /// The client has exhausted all retry attempts.
+    #[error(
+        "transient error from the Pub/Sub backend server: {0}. The client has exhausted all retry attempts."
+    )]
+    Transient(String),
```
PhongChuong: Transient is currently unused but I added it here as it is paired with Permanent error.
It will be used in the next PR to propagate transient error back to the application after the final retry_loop attempt.

dbolduc: We definitely should not distinguish between permanent and transient errors, because they are no different to the application. The application can't modify the retry policies.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -222,23 +226,58 @@ where
...
-        // TODO(#4804): process the results, and return ack IDs that fail with
-        // transient errors here.
-        Vec::new()
+        transient_failures.into_iter().collect()
+    }
+
+    fn extract_failures(
+        info: &google_cloud_rpc::model::ErrorInfo,
+        transient: &mut std::collections::HashSet<String>,
+        permanent: &mut std::collections::HashMap<String, String>,
```
dbolduc: prefer to return these as values instead of using mutable out-params

https://abseil.io/tips/176

src/pubsub/src/error.rs
dbolduc: I was planning on putting all errors from the server into the `Rpc` branch. Why do you prefer a new error branch? Is it because the error info is gross and long in debug strings?

My plan was to extend the `Rpc` branch contain a `details: Option<String>` which we fill in if the ack ID is in the error metadata. Other languages use an enum, but mehhh, I feel like it's not very helpful. https://pkg.go.dev/cloud.google.com/go/pubsub#AcknowledgeStatus.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -195,6 +197,8 @@ where
 }
 
 fn retry_policy() -> Arc<NeverRetry> {
+    // TODO(#4804): Update the retry_policy to retry for the following error
```
dbolduc, line 209: I like that you cut the PR off here.

FYI, we can probably reuse the `StreamRetryPolicy`, which maybe should retry on `DEADLINE_EXCEEDED`: https://github.com/googleapis/google-cloud-rust/blob/9e5a76024e8a9392b2bed33c58db14c3b81d97bd/src/pubsub/src/subscriber/retry_policy.rs#L51

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -222,23 +226,58 @@ where
...
         let shared_result = response.map(|_| ()).map_err(Arc::new);
         let confirmed_acks = ack_ids
             .into_iter()
+            .filter(|id| !transient_failures.contains(id))
             .map(|id| {
-                (
-                    id,
+                let result = if let Some(err_str) = permanent_failures.get(&id) {
+                    Err(AckError::Permanent(err_str.clone()))
+                } else {
```
dbolduc: I was thinking this would all be `AckError::Rpc`, but would have some optional field for the `err_str`.

dbolduc (changes requested):
I could _maybe_ be convinced that we should use a separate branch for errors returned via metadata vs. those returned as the gRPC status.

But I think we would call it like `Service` or `ServiceFailure` or something? (Not sure. I still lean towards using `Rpc`).

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Thanks.

src/pubsub/src/error.rs
PhongChuong: My concern is that ErrorInfo can potentially contains a lot of information including those of other ack_ids that are "transient". If these are retried and is confirmed, then it can be confusing to users who are not familiar with the server behavior with regards to ErrorInfo. This is made worst by the fact that we might want to return the transient error if we exhausted the retry attempts.

I agree with the assessment that Permanent/Transient are no different to the application.

AckError::Service seems reasonable.

PhongChuong: As discussed off thread, switched to use AckError::Rpc instead. We can extend this if needed.

PhongChuong:
@dbolduc , thanks for the review. I've updated to use AckError::Rpc instead as discussed off thread.

I also noticed that the behavior for responses with error info with ack_ids not in the error info was incorrect. The behavior should be to return Ok. I've updated the code and unit tests to reflect this. PTAL.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -222,23 +226,71 @@ where
             .set_subscription(self.subscription.clone())
             .set_ack_ids(ack_ids.clone());
         let response = self.inner.acknowledge(req, self.options.clone()).await;
+
+        let mut has_error_info = false;
```
dbolduc: I feel like the accounting could be simplified if we have a `HashMap<String, Result<>>` for the ack_ids.

We can initialize them all to Ok. If we get a permanent error, we update the result. And if we get a transient error we pluck it from the map.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=a5b3ee2db5b079d64e2ea69039d14547

I cut a lot of corners modeling our types though, so maybe I am cheating on avoiding that complexity. Still, consider taking ideas from that if they make sense.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -323,6 +375,46 @@ pub(super) mod tests {
...
+            ("ack_2".to_string(), "GIBBERISH_IGNORE".to_string()),
+            ("ack_3".to_string(), "TRANSIENT_FAILURE_OTHER".to_string()),
+            (
+                "ack_4".to_string(),
+                "PERMANENT_FAILURE_INVALID_ACK_ID".to_string(),
+            ),
+            ("ack_5".to_string(), "PERMANENT_FAILURE_OTHER".to_string()),
+        ]
+        .into_iter()
+        .collect();
```
dbolduc: we have builders that make this more friendly

```suggestion
        let info = google_cloud_rpc::model::ErrorInfo::default()
            .set_reason("reason")
            .set_domain("domain")
            .set_metadata([
                ("ack_1", "TRANSIENT_FAILURE_UNORDERED_ACK_ID"),
                ("ack_2", "GIBBERISH_IGNORE"),
                ("ack_3", "TRANSIENT_FAILURE_OTHER"),
                ("ack_4", "PERMANENT_FAILURE_INVALID_ACK_ID"),
                ("ack_5", "PERMANENT_FAILURE_OTHER"),
            ]);
```

src/pubsub/src/subscriber/leaser.rs
dbolduc: Maybe to make the interfaces composable we do like:

```rs
// Returns (confirmed, transients)
fn process(ack_ids: HashMap<String, AckResult>, response: &GaxResponse) -> (HashMap<String, AckResult>, HashMap<String, AckResult>) {
  // blah blah blah, extraction code.

  (confirmed, transients)
}

File: src/pubsub/src/subscriber/leaser.rs [hunk 7]
```diff
@@ -486,4 +578,195 @@ pub(super) mod tests {
...
+
+        let leaser = DefaultLeaser::new(
+            Arc::new(mock),
+            confirmed_tx,
+            "projects/my-project/subscriptions/my-subscription".to_string(),
+            10,
+            1_usize,
+        );
+        let ack_ids = leaser
+            .confirmed_ack_attempt(vec!["ack_1".to_string(), "ack_2".to_string()])
```
dbolduc: I suggest using `test_ids(1..3)`. That is consistent with the existing tests.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -486,4 +578,195 @@ pub(super) mod tests {
...
+            .await;
+
+        assert!(ack_ids.is_empty());
```
dbolduc:
```suggestion
        assert!(ack_ids.is_empty(), "{ack_ids:?}");
```

Or `assert_eq!(ack_ids, Vec::new());`

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -486,4 +578,195 @@ pub(super) mod tests {
...
+
+        let confirmed_acks = confirmed_rx.recv().await.expect("results were not sent");
+        assert_eq!(confirmed_acks.len(), 2);
+        assert!(confirmed_acks.get("ack_1").unwrap().is_ok());
+        assert!(confirmed_acks.get("ack_2").unwrap().is_ok());
```
dbolduc: optional: this is fine, but I prefer to match the whole container instead of individual elements.

We could compare `HashMap`s directly if we

```rs
#[cfg(test)]
impl PartialEq for AckResult { ... }
```

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -222,23 +226,71 @@ where
...
             .set_ack_ids(ack_ids.clone());
         let response = self.inner.acknowledge(req, self.options.clone()).await;
+
+        let mut has_error_info = false;
+        let mut transient_failures = std::collections::HashSet::new();
+        let mut permanent_failures = std::collections::HashMap::new();
+        if let Err(e) = &response {
+            if let Some(status) = e.status() {
+                for detail in &status.details {
+                    if let google_cloud_gax::error::rpc::StatusDetails::ErrorInfo(info) = detail {
```
dbolduc: nit: might be worth a `use google_cloud_gax::error::rpc::StatusDetails;`

dbolduc (commented):
I think this works, but can you see if there is a simpler way to write it?

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Updated to use the suggested method of updating when we see a permanent error vs plucking it from the map for transient.

I ended up not initializing the result to Ok as I didn't see much improvement in doing because I branched earlier based on the response.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -217,40 +229,96 @@ impl<T> DefaultLeaser<T>
...
 pub(super) mod tests {
-    use super::super::lease_state::tests::{sorted, test_ids};
+    use super::super::lease_state::tests::{sorted, test_id, test_ids};
     use super::super::retry_policy::tests::verify_policies;
     use super::super::stub::tests::MockStub;
     use super::*;
     use crate::{Error, Response};
     use google_cloud_gax::error::rpc::{Code, Status};
-    use std::sync::Arc;
+    use std::collections::HashSet;
```
dbolduc: nit: could use this in `super` to save some typing

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -217,40 +229,96 @@ impl<T> DefaultLeaser<T>
...
-            .set_ack_ids(ack_ids.clone());
-        let response = self.inner.acknowledge(req, self.options.clone()).await;
-        let shared_result = response.map(|_| ()).map_err(Arc::new);
-        let confirmed_acks = ack_ids
-            .into_iter()
-            .map(|id| {
-                (
+    fn process_ack_attempt(
+        ack_ids: Vec<String>,
+        response: std::result::Result<crate::Response<()>, google_cloud_gax::error::Error>,
```
dbolduc:
```suggestion
        response: Result<Response<()>>,
```

plus

```rs
use crate::{Result, Response};
```

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -323,6 +407,36 @@ pub(super) mod tests {
...
+                ("ack_2", "GIBBERISH_IGNORE"),
+                ("ack_3", "TRANSIENT_FAILURE_OTHER"),
+                ("ack_4", "PERMANENT_FAILURE_INVALID_ACK_ID"),
+                ("ack_5", "PERMANENT_FAILURE_OTHER"),
+            ]);
+
+        let status = Status::default()
+            .set_code(Code::FailedPrecondition)
+            .set_message("fail")
+            .set_details(vec![StatusDetails::ErrorInfo(info)]);
```
dbolduc: Q: do you want to test the case where there are multiple `ErrorInfo`s ?

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -144,14 +145,21 @@ where
...
-            let ids = std::mem::take(&mut ack_ids);
-            let ack_ids = leaser.confirmed_ack_attempt(ids).await;
-            if ack_ids.is_empty() {
+            let ids = std::mem::take(&mut remaining_ids);
+            let req = AcknowledgeRequest::new()
+                .set_subscription(leaser.subscription.clone())
+                .set_ack_ids(ids.clone());
+            let response = leaser.inner.acknowledge(req, leaser.options.clone()).await;
+
+            let (to_confirm, remaining) = Self::process_ack_attempt(ids, response);
```
dbolduc: FYI: this is consuming the error. If there are leftover transient errors it will take extra work to get them out.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -217,40 +229,96 @@ impl<T> DefaultLeaser<T>
...
+    ) -> (HashMap<String, AckResult>, Vec<String>) {
+        let e = match response {
+            Ok(_) => {
+                let to_confirm = ack_ids.into_iter().map(|id| (id, Ok(()))).collect();
+                return (to_confirm, Vec::new());
+            }
+            Err(e) => e,
+        };
+
+        let (transient_failures, permanent_failures) = Self::extract_failures(&e);
```
dbolduc: It seems sub-optimal to construct these sets...

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -217,40 +229,96 @@ impl<T> DefaultLeaser<T>
...
+        // Otherwise, we extract specific failures:
+        // - ack_ids with transient failures are to be retried.
+        // - ack_ids with permanent failures are resolved with the RPC error.
+        // - Unlisted ack_ids are considered successfully acknowledged.
+        let mut transient = Vec::new();
+        let mut to_confirm = HashMap::new();
+        for id in ack_ids {
+            if transient_failures.contains(&id) {
+                transient.push(id);
+            } else if permanent_failures.contains(&id) {
```
dbolduc: ... then loop over every ack ID in the original set and look the ack ID up in both sets.

I think an optimal algorithm would do one look up per entry in the metadata.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -217,40 +229,96 @@ impl<T> DefaultLeaser<T>
...
+            Err(e) => e,
+        };
+
+        let (transient_failures, permanent_failures) = Self::extract_failures(&e);
+        let shared_err = Arc::new(e);
+
+        // If the response lacks specific per ack_id failure info, we treat the response as all sharing the same
+        // RPC error.
+        if transient_failures.is_empty() && permanent_failures.is_empty() {
+            let to_confirm = ack_ids
```
dbolduc: If this is a transient error, e.g. `UNAVAILABLE`, we should be retrying. Do we have a plan for that?

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -305,6 +373,22 @@ pub(super) mod tests {
         }
     }
 
+    impl PartialEq for AckError {
+        fn eq(&self, other: &Self) -> bool {
+            match (self, other) {
+                (AckError::LeaseExpired, AckError::LeaseExpired) => true,
+                (AckError::ShutdownBeforeAck, AckError::ShutdownBeforeAck) => true,
+                (AckError::Rpc { source: s1 }, AckError::Rpc { source: s2 }) => {
+                    format!("{:?}", s1) == format!("{:?}", s2)
```
dbolduc, line 373: 👍

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -486,4 +600,102 @@ pub(super) mod tests {
...
+            .collect();
+        assert_eq!(confirmed_acks, expected);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn process_ack_attempt_permanent_failure() -> anyhow::Result<()> {
+        let info = google_cloud_rpc::model::ErrorInfo::new()
+            .set_metadata([(test_id(1), "PERMANENT_FAILURE_INVALID_ACK_ID".to_string())]);
```
dbolduc:
```suggestion
            .set_metadata([(test_id(1), "PERMANENT_FAILURE_INVALID_ACK_ID")]);
```

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -486,4 +600,102 @@ pub(super) mod tests {
...
+            })
+            .collect();
+        assert_eq!(confirmed_acks, expected);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn process_ack_attempt_permanent_failure() -> anyhow::Result<()> {
+        let info = google_cloud_rpc::model::ErrorInfo::new()
```
dbolduc: nit: If I ever catch myself writing `google_cloud_foo`, I think about whether I should `use` that type.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -486,4 +600,102 @@ pub(super) mod tests {
...
+            .set_metadata([(test_id(1), "PERMANENT_FAILURE_INVALID_ACK_ID".to_string())]);
+
+        let response = Err(Error::service(
+            Status::default()
+                .set_code(Code::FailedPrecondition)
+                .set_message("fail")
+                .set_details(vec![StatusDetails::ErrorInfo(info.clone())]),
+        ));
+        let (confirmed_acks, remaining) =
+            DefaultLeaser::<MockStub>::process_ack_attempt(test_ids(1..3), response);
```
dbolduc: nit: why define this on the leaser instead of as a standalone fn?

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -486,4 +600,102 @@ pub(super) mod tests {
...
+            .collect();
+        assert_eq!(confirmed_acks, expected);
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn process_ack_attempt_transient_failure() -> anyhow::Result<()> {
+        let info = google_cloud_rpc::model::ErrorInfo::new()
+            .set_metadata([(test_id(1), "TRANSIENT_FAILURE_OTHER".to_string())]);
```
dbolduc:
```suggestion
            .set_metadata([(test_id(1), "TRANSIENT_FAILURE_OTHER")]);
```

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -486,4 +600,102 @@ pub(super) mod tests {
...
+
+        assert!(remaining.is_empty(), "{remaining:?}");
+
+        let err = AckError::Rpc {
+            source: Arc::new(Error::service(
+                Status::default()
+                    .set_code(Code::FailedPrecondition)
+                    .set_message("fail")
+                    .set_details(vec![StatusDetails::ErrorInfo(info)]),
+            )),
```
dbolduc: optional: could factor out into `test_error() -> crate::Error { Error::service(...) }`

src/pubsub/src/subscriber/leaser.rs
dbolduc: Maybe. I could imagine we have `N` ack IDs and then all `N` show up as transient failures.

My suggestion would do `N` inserts then `N` removes on the `HashMap`. That seems dumb.

But I think it is more likely to have a request with ~100 ack IDs, and 2 of them show up as transient failures.

dbolduc: No action required. Just something to think about it.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -217,40 +229,96 @@ impl<T> DefaultLeaser<T>
...
+            }
+            Err(e) => e,
+        };
+
+        let (transient_failures, permanent_failures) = Self::extract_failures(&e);
+        let shared_err = Arc::new(e);
+
+        // If the response lacks specific per ack_id failure info, we treat the response as all sharing the same
+        // RPC error.
+        if transient_failures.is_empty() && permanent_failures.is_empty() {
```
dbolduc: comment (thinking out loud): This is probably what we want... but I am only like 99% sure.

It would be weird if the service attached an `ErrorInfo` with no metadata to indicate that all the acks succeeded. I think they would return `OK` instead.

Might be worth confirming with Pub/Sub folks, though.

dbolduc (commented):
The only blocker is handling transient failures for the no ErrorInfo case. If your plan was to add that in a follow up, cool, lmk.

We should also look out for how to return the transient error.

src/pubsub/src/subscriber/leaser.rs
PhongChuong: The plan is to set the inner (client) retry policy to retry on the errors listed on line 209 and for this outer retry_loop to use the same retry throttler, retry policy and backoff policy.
If the error one of the listed (including `UNAVAILABLE`), then it the client directly retries it without the confirmed_ack attempt closure being involved until it exhausted the retry attempts. After the final attempt, confirmed_ack attempt closure is called from the outer retry loop and we return the AckError::Rpc.

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Confirmed with the PubSub folks and Ok is returned if all ack succeeded.

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Thanks.
I haven't fully committed to a particular solution for the left over transient errors. It's likely that I will update the process_ack_attempt to take Arc<Error> instead the response.

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Ack.
I think if the ErrorInfo is only in one status.details then we can directly use the ErrorInfo map.

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Added the test case.

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Moved out of Leaser.

PhongChuong:
> The only blocker is handling transient failures for the no ErrorInfo case. If your plan was to add that in a follow up, cool, lmk.

I plan to add this as a follow up PR.

> 
> We should also look out for how to return the transient error.

@dbolduc , thanks for the review. Updated the PR with the suggestions.

src/pubsub/src/subscriber/leaser.rs
dbolduc: That is more complicated than having one retry policy. I don't think we should do that.

The loops have separate `RetryState`s. How are we going to enforce `.with_attempt_limit(3)` which is the spec for lease extensions?

In theory, we could receive:

- 1 `UNAVAILABLE`s without `ErrorInfo`
- then 1 `UNAVAILABLE` with `ErrorInfo`

x3 iterations in the outer loop for 6 total RPCs. (This isn't a realistic scenario, but it seems easy to avoid)

---

I think if we get a transient error here, we would return all the ack IDs as transients.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -323,6 +407,45 @@ pub(super) mod tests {
...
+            HashSet::from(["ack_1".to_string(), "ack_3".to_string()])
+        );
+        assert_eq!(
+            permanent,
+            HashSet::from(["ack_4".to_string(), "ack_5".to_string()])
+        );
+    }
+
+    #[test]
+    fn extract_failures_multiple_error_info() {
```
dbolduc, line 437: 👍

dbolduc (approved):

src/pubsub/src/subscriber/leaser.rs
dbolduc: Ack, I think we will have to return the `Result` with the list of transients somehow. Doesn't have to be in this PR.

PR #5360, opened 2026-04-10
chore: bump and regenerate libraries

noahdietz:
Created by running `librarian bump --all` and `librarian generate --all`.


coryan (approved):

src/storage/README.md
```diff
@@ -31,12 +31,12 @@ should not introduce breaking changes to the client libraries.
 
 ## More Information
 
-- Read the [crate's documentation](https://docs.rs/google-cloud-storage/latest)
+- Read the [crate's documentation](https://docs.rs/google-cloud-storage/1.11.0)
```
coryan, line 34: @dbolduc this you make this change happen (pure guess you might have, you seemed interested in pinning the READMEs to the right version)? It is a neat change if you did, but why don't we see the same change in pubsub ?

/CC: @suzmue

dbolduc: I did not author any code to do this. I only changed the README mustache templates for GAPICs to encode the version in links.

I am guessing this is some pattern matching in librarian and the pubsub README.md does not have this pattern, but storage README.md does.

PR #5361, opened 2026-04-10
test(pubsub): add sample for publisher retry settings

suzmue:
Fixes #5349

### DIFF
--- a/src/pubsub/examples/src/publisher/publish_with_retry_settings.rs
+++ b/src/pubsub/examples/src/publisher/publish_with_retry_settings.rs
@@ -1,0 +1,48 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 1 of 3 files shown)


src/pubsub/examples/src/publisher/publish_with_retry_settings.rs
```diff
@@ -0,0 +1,48 @@
...
+use google_cloud_pubsub::client::Publisher;
+use google_cloud_pubsub::model::Message;
+use std::time::Duration;
+
+pub async fn sample(project_id: &str, topic_id: &str) -> anyhow::Result<()> {
+    let topic_name = format!("projects/{project_id}/topics/{topic_id}");
+
+    // Configure custom retry settings.
+    // In this example, we retry with a time limit of 10 minutes.
+    let retry_policy = AlwaysRetry.with_time_limit(Duration::from_secs(600));
```
dbolduc, line 27: Why not use `RetryableErrors` instead of `AlwaysRetry`?

https://github.com/googleapis/google-cloud-rust/blob/a7dba239703118ca375dbb7777659d1e08adb3f6/src/pubsub/src/publisher/retry_policy.rs#L57

Oh, because it is not public. Should it be public? It seems useful. We don't want applications retrying permanent errors for 10 minutes

dbolduc (approved):

src/pubsub/examples/src/publisher/publish_with_retry_settings.rs
suzmue: Yeah maybe we should add RetryableErrors to a retry_policy module like in [storage](https://docs.rs/google-cloud-storage/latest/google_cloud_storage/retry_policy/index.html).

Filed #5366

PR #5368, opened 2026-04-11
impl(gax-internal): integrate `RequestRecorder` with gRPC transport

haphungw:
Add hook to record when a gRPC request starts and completes (successfully or with error).

This PR implements changes to Unary streaming only. Bidirectional streaming and Server streaming will be addressed in subsequent PRs.

### DIFF
--- a/src/gax-internal/src/observability/client_signals.rs
+++ b/src/gax-internal/src/observability/client_signals.rs
@@ -18,10 +18,10 @@ mod transport_metric;
 mod with_client_logging;
 mod with_client_metric;
 mod with_client_span;
-#[cfg(feature = "_internal-http-client")]
+#[cfg(any(feature = "_internal-http-client", feature = "_internal-grpc-client"))]
 mod with_transport_logging;
 mod with_transport_metric;
-#[cfg(feature = "_internal-http-client")]
+#[cfg(any(feature = "_internal-http-client", feature = "_internal-grpc-client"))]
 mod with_transport_span;
 
 pub use duration_metric::DurationMetric;
@@ -30,10 +30,10 @@ pub use transport_metric::TransportMetric;
 pub use with_client_logging::WithClientLogging;
 pub use with_client_metric::WithClientMetric;
 pub use with_client_span::WithClientSpan;
-#[cfg(feature = "_internal-http-client")]
+#[cfg(any(feature = "_internal-http-client", feature = "_internal-grpc-client"))]
 pub use with_transport_logging::WithTransportLogging;
 pub use with_transport_metric::WithTransportMetric;
-#[cfg(feature = "_internal-http-client")]
+#[cfg(any(feature = "_internal-http-client", feature = "_internal-grpc-client"))]
 pub use with_transport_span::WithTransportSpan;
 
 /// Creates a [Span] and decorated future for a client request.
@@ -80,6 +80,15 @@ macro_rules! client_request_signals {
         use ::tracing::instrument::Instrument;
         let span = $crate::client_request_signals!(info: $info, method: $method);
         let recorder = $crate::observability::RequestRecorder::new($info);
+        if let Some(current) = $crate::observability::RequestRecorder::current() {
+            let snap = current.client_snapshot();
+            if let Some(rn) = snap.resource_name() {
+                recorder.on_client_request(
+                    $crate::observability::ClientRequestAttributes::default()
+                        .set_resource_name(rn.to_string())
+                );
+            }
+        }
         let pending = recorder
             .scope($crate::observability::WithClientSpan::new(
                 span.clone(),
--- a/src/gax-internal/src/observability/client_signals.rs
+++ b/src/gax-internal/src/observability/client_signals.rs
@@ -18,10 +18,10 @@ mod transport_metric;
 mod with_client_logging;
 mod with_client_metric;
 mod with_client_span;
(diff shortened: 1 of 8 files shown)


src/gax-internal/src/observability/client_signals.rs
```diff
@@ -80,6 +80,15 @@ macro_rules! client_request_signals {
         use ::tracing::instrument::Instrument;
         let span = $crate::client_request_signals!(info: $info, method: $method);
         let recorder = $crate::observability::RequestRecorder::new($info);
+        if let Some(current) = $crate::observability::RequestRecorder::current() {
```
coryan: I don't follow... this is where we create a request recorder and then set the recoder for a given future...  Why would we notify any existing recorder (which is also unexpected) of anything?

tests/o11y/tests/storage.rs
```diff
@@ -16,6 +16,7 @@
 mod storage {
     use google_cloud_test_utils::errors::anydump;
 
+    #[ignore]
```
coryan, line 19: Why is this disabled?

coryan (commented):
I don't think I understand this change.

tests/o11y/tests/storage.rs
haphungw: This requires `BidiReadObject` to pass. I'm not sure about the current implementation, so I wanna get this through before adding more code.
The test will be enabled again in the next PR when we add bidi streaming support.

haphungw:
The main motivation behind this PR (and also why I am so rushed) is to fix a regression in our gRPC observability logic. The generated gRPC transport still expects to find `resource_name` in `req.extensions()` because we already removed the insertions when we started the `RequestRecorder` refactor, leaving gRPC spans without resource names. https://github.com/googleapis/librarian/blob/db6b993937641a143240625be669eb97f3b24270/internal/sidekick/rust/templates/grpc-client/transport.rs.mustache#L191

If we don't want to refactor immediately, an alternative is to put `extensions.insert(gaxi::observability::ResourceName::new(rn));`
which I think is not ideal given our current approach to use `RequestRecorder`.
https://github.com/googleapis/librarian/pull/5298/changes: 

---

I found all of these issues when I was verifying the storage integration tests: 
https://github.com/westarle/google-cloud-rust/blob/0609d255b33cbc42d664a3e625e542c448643764/tests/o11y/src/storage_grpc_tracing.rs#L273-L288

We have three problems to deal with right now:

- State propagation: gRPC transport currently cannot access `resource_name` because it was removed from request extensions.
- RPC Types: gRPC needs implementations for Unary, Bidirectional, and Server Streaming RPCs (unlike http which is purely request-response).
- Missing resource names: `storage` (and `compute`) have many methods without resource names. They were just not picked up by our current heuristic: https://github.com/googleapis/librarian/pull/5299

What I'm trying to do in this PR is to see if this is the right way to give gRPC access to `RequestRecorder` for Unary, and if this works I'll extend it to Bidi and Server Streaming.

src/gax-internal/src/observability/client_signals.rs
haphungw: This is true, this block of code makes no sense. This was a helpless attempt to fix the [bug](https://github.com/westarle/google-cloud-rust/blob/0609d255b33cbc42d664a3e625e542c448643764/tests/o11y/src/storage_grpc_tracing.rs#L288) in the storage integration test, but now as I double check this happened because we were not able to generate the resource name for many `storage` methods.

tests/o11y/tests/storage.rs
coryan: Ack, thanks. Consider: 

```rust
#[ignore = "TODO(#....) - restore when bidi streaming works again"]
```

src/gax-internal/tests/grpc_observability.rs
```diff
@@ -314,6 +348,7 @@ mod tests {
         Ok(())
     }
 
+    #[ignore]
```
coryan, line 351: Presumably these need a `TODO(#...)` also.

coryan:
I think (but I am really not sure) we have two choices:

1) For gRPC, we use the options extension, set the resource name in the generated code (as shown in googleapis/librarian#5298) and then populate the RequestRecorder from that extension, or
2) Change the generated code to store the resource name directly into the RequestRecorder.

I am guessing (2) is cleaner, though maybe more risky.  Are you trying to implement (1) or some alternative I did not consider?

A third (less desirable  option) is to release without this attribute feature. It only affects the Storage admin operations which are not used that often 🤷

coryan (approved):
I don't think this hurts anything, so approved. I am not sure it solves anything either, but I may be wrong.

haphungw:
Yes, I was trying to do #1. I don't really think we have any other choices.

Re #2 being cleaner but more risky, what are the risks we are facing that is not regenerating with librarian?

westarle:
I'm ok with skipping the resource id attribute for storage control until after freeze to avoid risk to the release and do things 'right'

haphungw:
Since we are deferring the resource ID for storage control until post-freeze, what's the general sentiment on option 1 (`extensions`) vs option 2 (`RequestRecorder`) for the long term?

I'm already trying to make #1 work for Bidi streaming in a separate branch, but if that somehow misses the mark I hope reverting is not too bad 🤦‍♂️

PR #5369, opened 2026-04-11
chore: update version of prost-build

dbolduc:
Update the version of `prost_build` used to generate our protos from `0.13` to `0.14`.

The changes correspond to: https://github.com/googleapis/librarian/pull/5310


tests/protojson-conformance/src/generated/protos/conformance.rs
```diff
@@ -111,7 +111,7 @@ impl ::prost::Name for ConformanceResponse {
         "type.googleapis.com/conformance.ConformanceResponse".into()
     }
 }
-#[derive(Clone, Copy, PartialEq, ::prost::Message)]
+#[derive(Clone, Copy, PartialEq, Eq, Hash, ::prost::Message)]
```
coryan, line 114: If there is a flag to turn off the `Eq` and `Hash` in the generated code we should use it. We don't need those traits, and certainly don't need the extra code bloat.

coryan (approved):

tests/protojson-conformance/src/generated/protos/conformance.rs
dbolduc: Yeah. I was too lazy to look earlier, but it seems like it is not possible at the moment to configure the derived traits.

https://github.com/tokio-rs/prost/issues/1368

PR #5384, opened 2026-04-13
ci: remove rustdocfx gcb

PhongChuong:


coryan (approved):
You can manually disable the trigger or manually run terraform to delete it.

PR #5386, opened 2026-04-13
impl(auth): add universe_domain support for Impersonated

alvarowolfx:


alvarowolfx:
Split access boundaries changes in https://github.com/googleapis/google-cloud-rust/pull/5405

src/auth/src/credentials/impersonated.rs
```diff
@@ -1754,6 +1784,95 @@ mod tests {
         );
     }
 
+    #[tokio::test]
+    #[parallel]
+    async fn test_with_universe_domain() -> TestResult {
```
suzmue: nit: I think I would prefer these tests if they were organized as table driven tests. For every configuration, calling `with_universe_domain` should override whatever was set by the source (and without it should return the value from the the source that you set). 

Having a single test that checks this for credentials of all these types *might* be easier to read and verify that we have tested all use cases.

src/auth/src/credentials/impersonated.rs
```diff
@@ -1754,6 +1784,95 @@ mod tests {
...
+            creds.universe_domain().await,
+            Some("another-universe.com".to_string())
+        );
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    #[parallel]
+    async fn test_with_universe_domain_mds_source() -> TestResult {
```
suzmue: supernit: test names don't need to start with test in rust, its redundant with the module name.

src/auth/src/credentials/impersonated.rs
```diff
@@ -319,6 +322,16 @@ impl Builder {
         self
     }
 
+    /// Sets the Google Cloud universe domain for these credentials.
+    ///
+    /// Any value provided here overrides a `universe_domain` value from the input service account JSON.      
+    // TODO(#3646): Make this public and let example run when universe domain support is done.
+    #[allow(dead_code)]
```
suzmue, line 329: alternative: #[cfg(test)] might also be appropriate

suzmue (approved):

src/auth/src/credentials/impersonated.rs
suzmue: Nice refactor of the tests! Thanks :)

PR #5403, opened 2026-04-14
cleanup(storage): update rust storage module template to storage

jameslynnwu:
Librarian, https://github.com/googleapis/librarian/pull/5368, will
be updated to use template from rust module to generate
the storage module, rather than a custom "rust_storage" language.

Will follow up to remove language from rust modules, since its implied
by being in rust module.


dbolduc (approved):
Oops

Issue #5396, opened 2026-04-14
Use generated `grpc-mock`


dbolduc:
A follow up to actually use #5333 

The following libraries have handwritten `grpc-mock` frameworks. They should use the generated one.

- [x] storage
- [x] pubsub
- [ ] spanner
  - This one will take some test refactors, sadly.
  - The tests will end up cleaner, though.

PR #5404, opened 2026-04-14
cleanup(storage): use generated grpc-mock

dbolduc:
Part of the work for #5396 

Use sidekick to generate the `grpc-mock` fake test server for Storage.

I think it is silly to generate the test code, so instead we move the existing handwritten unit tests for storage into a single `tests/grpc-mock`.


coryan (dismissed):

librarian.yaml
```diff
@@ -1737,6 +1737,17 @@ libraries:
     copyright_year: "2025"
     output: src/storage/grpc-mock
```
dbolduc, line 1737: I think librarian deletes this unless I make it a module under `src/storage`. Which is fine, I can do that.

@coryan - does that sound right?

coryan: I think that sounds right. Worth trying anyway.

dbolduc: Nvm, that doesn't even work. It might be the `grpc-mock/` fighting with the `grpc-mock/src/generated/protos`

dbolduc: New theory: I didn't update librarian to a version that has my template, so no code is generated. :monkey: 

https://screenshot.googleplex.com/43YnfeyUA3jEztJ

This is what I get for trying to author a PR from my personal windows laptop with 0 tooling.

dbolduc:
PTAL

coryan (approved):
LGTM. Now that we can generate `grpc-mock` we can test the generated code once, and YOLO the other libraries.

dbolduc:
> the generated code once, and YOLO the other libraries.

Storage has a read, a write, and a bidi. I am not scared.

Issue #5408, opened 2026-04-14
PubSub: Subscriber Leaser.confirmed_ack Arc Mutex usage


PhongChuong:
confirmed_ack uses arc mutex so that it can process the remaining_ids with last_error after the final attempt of retry_loop. Since retry_loop execute the the attempt closure sequentially, it may be possible to avoid using arc mutex. Possible solutions are:
- mutable capture
- clever use of retry_loop return value
- custom retry_loop
- refactor retry_loop to use generic error and embedding remaining_ids in the returned error

dbolduc:
The current solution is perfectly reasonable. Yes, we could optimize the happy path more, but the only way to recover the failures when the loop terminates without a lock and without rewriting the retry loop is to clone the remaining ids. I don't think that is worth it.

PR #5409, opened 2026-04-14
impl(pubsub): process confirm_ack remaining ids after final attempt

PhongChuong:
Note: the policy to enable retry will be added in a followup PR

### DIFF
--- a/src/pubsub/src/subscriber/leaser.rs
+++ b/src/pubsub/src/subscriber/leaser.rs
@@ -18,7 +18,6 @@ use super::stub::Stub;
 use crate::RequestOptions;
 use crate::error::AckError;
 use crate::model::{AcknowledgeRequest, ModifyAckDeadlineRequest};
-use crate::{Response, Result};
 use google_cloud_gax::error::rpc::StatusDetails;
 use google_cloud_gax::exponential_backoff::ExponentialBackoff;
 use google_cloud_gax::retry_loop_internal::retry_loop;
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/leaser.rs
```diff
@@ -226,23 +268,18 @@ fn retry_throttler(
     })
 }
 
-fn process_ack_attempt(
+fn process_ack_attempt_error(
```
suzmue: Can we add a doc comment please that says what the return values represent?

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -145,24 +144,50 @@ where
...
+                    Ok(_) => (ids.into_iter().map(|id| (id, Ok(()))).collect(), Vec::new()),
+                    Err(e) => {
+                        let shared_err = Arc::new(e);
+                        let (to_confirm, remaining) =
+                            process_ack_attempt_error(ids, shared_err.clone());
+
+                        if !remaining.is_empty() {
+                            let mut err_guard =
+                                last_error.lock().expect("mutex should not be poisoned");
+                            *err_guard = Some(shared_err);
```
suzmue:
```suggestion
                            *last_error.lock().expect("mutex should not be poisoned") = Some(shared_err);
```

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -176,8 +201,27 @@ where
...
+
+        let final_remaining =
+            std::mem::take(&mut *remaining_ids.lock().expect("mutex should not be poisoned"));
+        if !final_remaining.is_empty() {
+            let err =
+                std::mem::take(&mut *last_error.lock().expect("mutex should not be poisoned"));
+            if let Some(shared_err) = err {
+                let confirmed_acks = final_remaining
+                    .into_iter()
+                    .map(|id| {
```
dbolduc, line 213: optional: At a guess, passing around a `HashMap<String, AckResult>` should reduce the number of places we need to write this map logic.

I count 3 places where we do this. 1 would be nice, but idk how practical it is.

dbolduc: No action required, this might fall naturally out of #5408

dbolduc (approved):

PR #5411, opened 2026-04-15
feat(pubsub): expose RetryableErrors for customized retry configuration

### LINKED ISSUE #5366: Expose RetryableErrors used for pubsub publisher
We should audit which of the retry policies across the publisher and subscriber we might want to expose and determine naming appropriately.

Likely this should be exposed in a retry_policy module like we do in [storage](https://docs.rs/google-cloud-storage/latest/google_cloud_storage/retry_policy/index.html).

suzmue:
This change introduces a new public `retry_policy` module in the `google-cloud-pubsub` crate, exporting the `RetryableErrors` struct. Users can now explicitly reference and decorate this policy to customize retry limits and durations on Publisher instances.

Fixes #5366


src/pubsub/src/retry_policy.rs
```diff
@@ -0,0 +1,207 @@
...
+#[derive(Clone, Debug)]
+pub struct RetryableErrors;
+
+impl RetryPolicy for RetryableErrors {
+    fn on_error(&self, _state: &RetryState, error: Error) -> RetryResult {
+        if error.is_transient_and_before_rpc() {
+            return RetryResult::Continue(error);
+        }
+
+        if error.is_io() || error.is_timeout() {
```
dbolduc, line 54: (I know this code already existed, but)

Why would we retry a timeout error? That is at best weird and at worst a bug.

src/pubsub/src/retry_policy.rs
```diff
@@ -0,0 +1,207 @@
...
+        // - 5xx: Internal Server Error, Bad Gateway, etc.
+        if let Some(408 | 429 | 499 | 500..600) = error.http_status_code() {
+            return RetryResult::Continue(error);
+        }
+
+        if let Some(status) = error.status() {
+            use google_cloud_gax::error::rpc::Code;
+            return match status.code {
+                Code::Aborted
+                | Code::Cancelled
```
dbolduc: This is suspicious. We don't have request cancelling, so not sure when we would receive a `CANCELLED` error. The docs seem to indicate they only expect `CANCELLED` coming from the client. They also say this could lead to a double publish, so I'd remove it.

"The client can retry the operation, but the operation may have been executed on the previous call."

src/pubsub/src/retry_policy.rs
```diff
@@ -0,0 +1,207 @@
...
+                | Code::DeadlineExceeded
```
dbolduc, line 79: This is suspicious.

The docs say "Typically the client would retry the operation. Keep in mind that this may result in the operation being executed more than once on the server."

I think it is better to avoid a double publish than retry on this error.

src/pubsub/src/retry_policy.rs
```diff
@@ -0,0 +1,207 @@
...
+            // retryable.
+            return RetryResult::Continue(error);
+        }
+
+        // Catch raw HTTP errors that may not have been mapped to a gRPC status.
+        // - 408: Request Timeout
+        // - 429: Resource Exhausted
+        // - 499: Cancelled Request
+        // - 5xx: Internal Server Error, Bad Gateway, etc.
+        if let Some(408 | 429 | 499 | 500..600) = error.http_status_code() {
```
dbolduc: > 500..600 

Can we enumerate the errors we should retry on?

e.g. `501` is "not implemented", `505` is "HTTP version not supported". While Pub/Sub probably doesn't return them, they definitely shouldn't be retried.

suzmue:
@dbolduc Will go through each of your comments too but this is also based on the retry implementations for Go and Java: https://github.com/googleapis/google-cloud-go/blob/9d47d1ca53dc1a3183352f67ebeb174b1d194846/pubsub/v2/service.go#L51 and https://github.com/googleapis/java-pubsub/blob/880f28a11d7dc2864ae958e59940bb24bb6fa555/google-cloud-pubsub/src/main/java/com/google/cloud/pubsub/v1/StatusUtil.java#L48

src/pubsub/src/retry_policy.rs
dbolduc: Nevermind, I think you are right and I am wrong.

> Why would we retry a timeout error?

There is an attempt timeout and an operation timeout. If we hit an attempt timeout, we should keep retrying. If we hit the operation timeout, we should stop.

This code only applies to attempt timeouts.

Operation timeouts are set using `RetryPolicyExt::with_time_limit`. And that has its own logic to override a `RetryResult::Continue` with a `RetryResult::Exhausted`.

https://github.com/googleapis/google-cloud-rust/blob/52388cce02ce4f62a8320023096076bc35cb03e2/src/gax/src/retry_policy.rs#L466-L468

dbolduc (approved):
I agree that we want to expose this exact API.

src/pubsub/src/retry_policy.rs
dbolduc: Thinking out loud... Might be worth bringing up in our team meeting or the pubsub meeting. You probably understand this better than I do.

> double publish

The more I think about publish retries, the more confused I am. Typically, we do not retry `POST`s like the `Publish` RPC because they are not idempotent.

The application knows whether the message is idempotent, but we don't in general. e.g. if the message is some event like "your build passed", that could be published twice without side effects. But if the message is for some chat, you would not want it published twice.

My guess is that applications probably set a nonce or a request ID in the pubsub message attributes if they want retries + "exactly once" publish. And let their subscribers check for dups. (Aside: then why is exactly-once delivery a subscriber feature when applications need to do this anyway for "exactly-once" publish?? 🫨)

So I think what I convinced myself of is that retrying on `DEADLINE_EXCEEDED` is no different from retrying on `UNAVAILABLE`. You could get a double publish in either case.

---

I am not sure if there is any action we should take. We could add a blurb in the docs saying the publisher retries by default and that applications can use `NeverRetry` to turn off retries? Maybe here:

https://docs.rs/google-cloud-pubsub/0.33.2/google_cloud_pubsub/builder/publisher/struct.PublisherBuilder.html#method.with_retry_policy

src/pubsub/src/retry_policy.rs
suzmue: Yes I will do this.

src/pubsub/src/retry_policy.rs
suzmue: Good point. Let's drop this one. I noticed Go also dropped it from the Java policy.

I am also dropping 499 to align with this decision.

src/pubsub/src/retry_policy.rs
suzmue: Yeah I think eliminating double publish is not a goal we have. Good point might be worth pointing out.

dbolduc (approved):

PR #5412, opened 2026-04-15
chore(spanner): add multiplexed session maintainer

olavloite:
Adds a background maintainer for the multiplexed session that is used by a database client. A session is created when a database client is created. The session is used for all operations that are executed using that client, and the session is rotated after 7 days. All transactions get the session name that they will be using when the transaction is created. This means that it is guaranteed that a transaction will continue to use the same session for its entire lifetime, even if the client rotates its session while a transaction is being executed.


olavloite:
(The build error is unrelated to this change)

coryan:
> (The build error is unrelated to this change)

Sorry about that. Fixed in https://github.com/googleapis/google-cloud-rust/pull/5413

src/spanner/src/session_maintainer.rs
```diff
@@ -0,0 +1,578 @@
...
+    async fn maintenance_loop(
+        maintainer: Weak<ManagedSessionMaintainer>,
+        interval: Duration,
+        age: Duration,
+    ) {
+        loop {
+            sleep(interval).await;
+            if !Self::maintain(&maintainer, age).await {
+                break;
+            }
```
dbolduc: Consider flattening this:

```rs
sleep(interval).await; // Only if you need/want this first sleep

while let Some(m) = maintainer.upgrade() {
    if let Err(e) = m.check_and_replace_session(age).await {
        tracing::warn!(
            "Failed to check and replace session for {}: {}. Retrying in 1 hour.",
            m.database_name,
            e
        );
    }
    sleep(interval).await;
}
```

dbolduc: You would not do this if `maintain` is going to grow. In that case you would do more like:

```rs
while let Some(m) = maintainer.upgrade() {
    Self::maintain(m).await;
    sleep(interval).await;
}
```

 Or `m.maintain().await;`. Or something.

src/spanner/src/session_maintainer.rs
```diff
@@ -0,0 +1,578 @@
...
+    }
+
+    async fn create_session(
+        spanner: &Spanner,
+        database_name: &str,
+        database_role: &str,
+        options: &RequestOptions,
+    ) -> Result<Session> {
+        let request = CreateSessionRequest::new()
+            .set_database(database_name.to_string())
```
dbolduc: nit: no need for `to_string()`

src/spanner/src/session_maintainer.rs
```diff
@@ -0,0 +1,578 @@
...
+            .set_session(
+                Session::new()
+                    .set_multiplexed(true)
+                    .set_creator_role(database_role.to_string()),
```
dbolduc: same.

src/spanner/src/session_maintainer.rs
```diff
@@ -0,0 +1,578 @@
...
+        true
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use gaxi::grpc::tonic::Response;
+    use google_cloud_auth::credentials::anonymous::Builder as Anonymous;
+    use spanner_grpc_mock::google::spanner::v1::Session as GrpcSession;
```
dbolduc, line 169: comment: I like the name choice.

dbolduc (approved):
The uses of `Arc`, `WeakArc`, `RwLock` all look right to me.

I just sorta squinted at the tests. I did not look at them closely.

src/spanner/src/session_maintainer.rs
```diff
@@ -0,0 +1,578 @@
...
+
+use crate::client::Spanner;
+use crate::model::{CreateSessionRequest, Session};
+use crate::{RequestOptions, Result};
+use std::sync::{Arc, RwLock, Weak};
+use std::time::Duration;
+use tokio::time::{Instant, sleep};
+
+/// The interval at which the background task checks if the session needs replacement.
+pub(crate) const SESSION_MAINTENANCE_INTERVAL: Duration = Duration::from_secs(3600);
```
sakthivelmanii, line 23: In java client, we have 10 mins by default. but 1hr is more than enough.

src/spanner/src/session_maintainer.rs
```diff
@@ -0,0 +1,578 @@
...
+                    .set_multiplexed(true)
+                    .set_creator_role(database_role.to_string()),
+            );
+
+        spanner.create_session(request, options.clone()).await
+    }
+
+    async fn maintenance_loop(
+        maintainer: Weak<ManagedSessionMaintainer>,
+        interval: Duration,
```
sakthivelmanii, line 142: why do we have to pass it here if we are not providing options to customer to alter it?

src/spanner/src/session_maintainer.rs
```diff
@@ -0,0 +1,578 @@
...
+                session: Arc::new(session),
+                created_at: Instant::now(),
+            }),
+            database_name,
+            database_role,
+            options,
+        });
+
+        let weak_maintainer = Arc::downgrade(&maintainer);
+        tokio::spawn(async move {
```
sakthivelmanii, line 78: how are we shutting down the maintainer when client is closed?

dbolduc: The client drops its strong `Arc<Maintainer>`.  The next time the maintenance loop wakes up and tries to `WeakArc::upgrade()`, it returns `None`, because there are 0 strong references.

src/spanner/src/session_maintainer.rs
dbolduc: We don't have to, but if you test the interface over a variable value, you can confidently change the default value later without rewriting any tests.

sakthivelmanii: In java client, we provide an option to change the value through configuration. I would like to know if that's the intention here.

src/spanner/src/session_maintainer.rs
sakthivelmanii: got it.

src/spanner/src/session_maintainer.rs
olavloite: No, my intention was not to provide a configuration option for customers. I only added this for testability, but in the end, it turned out to be hard to test this loop without introducing actual delays. tokio allows you to control time in tests, but that does not work well with the mock Spanner server. That was also the reason that I moved `maintain` to a separate method to make that testable.

I'll keep the argument here for now in the hope that we can find a way to test it in the future.

src/spanner/src/session_maintainer.rs
olavloite: I don't expect `maintain` to grow, but I wanted it in a separate method to be able to call it directly from tests. So I applied the second suggestion.

src/spanner/src/session_maintainer.rs
olavloite: yeah, considering that we start rotating already after 7 days, I don't think we need to try that often.

src/spanner/src/session_maintainer.rs
olavloite: removed

src/spanner/src/session_maintainer.rs
olavloite: removed

PR #5418, opened 2026-04-15
docs(auth): Credentials methods

### LINKED ISSUE #5259: Credentials missing documentation
https://docs.rs/google-cloud-auth/1.8.0/google_cloud_auth/credentials/struct.Credentials.html#implementations

`headers()` and `universe_domain()` are not documented.

alvarowolfx:
Fixes #5259

### DIFF
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -133,14 +133,22 @@ impl Credentials {
     ///
     /// Different auth tokens are sent via different headers. The
     /// [Credentials] constructs the headers (and header values) that should be
-    /// sent with a request.
+    /// sent with a request. Headers are cached in a background task and refreshed
+    /// periodically as the tokens expire.
     ///
     /// # Parameters
     /// * `extensions` - An `http::Extensions` map that can be used to pass additional
-    ///   context to the credential provider. For caching purposes, this can include
-    ///   an [EntityTag] from a previously returned [`CacheableResource<HeaderMap>`].
-    ///   If a valid `EntityTag` is provided and the underlying authentication data
-    ///   has not changed, this method returns `Ok(CacheableResource::NotModified)`.
+    ///   context to the credential provider. If the caller does not need to compute derived values
+    ///   from the headers then do not provide an `EntityTag`. The credentials will either return
+    ///   `Err(...)` or `Ok(CacheableResource::New {})` in this case. Since the credentials
+    ///   already cache the headers, then it can use the results directly. Some applications need
+    ///   to compute values derived from the result, and want to avoid that computation if the
+    ///   headers have not changed. In that case, provide the `EntityTag` returned from a previous
+    ///   call. If the underlying authentication data has not changed, this method returns
+    ///   `Ok(CacheableResource::NotModified)` and you can use the same derived data. If the
+    ///   caller provides an `EntityTag` and the underlying authentication data has changed, this
+    ///   function returns `Ok(CacheableResource::New { ... })`. That result invalidates the
+    ///   tag, and provides new values for the headers.
     ///
     /// # Returns
     /// A `Result` containing:
--- a/src/auth/src/credentials.rs
+++ b/src/auth/src/credentials.rs
@@ -133,8 +133,9 @@ impl Credentials {
     ///
     /// Different auth tokens are sent via different headers. The
     /// [Credentials] constructs the headers (and header values) that should be
-    /// sent with a request. Headers are cached in a background task and refreshed
-    /// periodically as the tokens expire.
+    /// sent with a request. If the authentication provider requires it, headers
+    /// are cached, and a background task periodically refreshes any expired
+    /// tokens.
     ///
     /// # Parameters
     /// * `extensions` - An `http::Extensions` map that can be used to pass additional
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials.rs
```diff
@@ -129,10 +129,42 @@ where
 }
 
 impl Credentials {
+    /// Asynchronously constructs the auth headers.
+    ///
+    /// Different auth tokens are sent via different headers. The
+    /// [Credentials] constructs the headers (and header values) that should be
+    /// sent with a request.
+    ///
```
coryan, line 139: I think you should mention that all credential types cache the headers, so applications don't need to cache them too.

src/auth/src/credentials.rs
```diff
@@ -129,10 +129,42 @@ where
...
+    /// # Parameters
+    /// * `extensions` - An `http::Extensions` map that can be used to pass additional
+    ///   context to the credential provider. For caching purposes, this can include
+    ///   an [EntityTag] from a previously returned [`CacheableResource<HeaderMap>`].
+    ///   If a valid `EntityTag` is provided and the underlying authentication data
+    ///   has not changed, this method returns `Ok(CacheableResource::NotModified)`.
```
coryan: Maybe:


```suggestion
    ///   context to the credential provider.
    ///  If the caller does not need to compute derived values from the headers then do not provide an `EntityTag`. The credentials will either return `Err(...)` or `Ok(CacheableResource::New {})` in this case.
    /* note to us: we should offer a simplified API for this case */
    /// Since the credentials already cache the headers, then it can use the results directly.
    ///
    ///   Some applications need to compute values derived from the result, and want to avoid
    /// that computation if the headers have not changed. 
    /// In that case, provide the `EntityTag` returned from a previous call. If the underlying authentication data
    ///   has not changed, this method returns `Ok(CacheableResource::NotModified)` and you can 
    /// use the same derived data.
    /// If the caller provides an `EntityTag` and the underlying authentication data has changed, this function returns `Ok(CacheableResource::New { ... })`.
    /// That result invalidates the tag, and provides new values for the headers.
```

alvarowolfx: > note to us: we should offer a simplified API for this case

We provide the AccessTokenCredentials to get tokens directly without any CacheableResource.

alvarowolfx (commented):
Applied your changes on the `extensions` parameter and added comment on headers being cached

src/auth/src/credentials.rs
coryan: > > note to us: we should offer a simplified API for this case
> 
> We provide the AccessTokenCredentials to get tokens directly without any CacheableResource.

Yes, and maybe we can have a thing that is "credentials with full headers, but easy to use if I don't need to do my own caching".

src/auth/src/credentials.rs
```diff
@@ -129,10 +129,50 @@ where
 }
 
 impl Credentials {
+    /// Asynchronously constructs the auth headers.
+    ///
+    /// Different auth tokens are sent via different headers. The
+    /// [Credentials] constructs the headers (and header values) that should be
+    /// sent with a request. Headers are cached in a background task and refreshed
+    /// periodically as the tokens expire.
```
coryan:
```suggestion
    /// sent with a request. If needed, headers are cached, and a background task
    /// periodically refreshes any expired tokens.
```

I would have said "If the authentication workflow requires it" instead of "if nedded", but I am not sure we have a fixed term for the different authentication workflows / protocols / what-is-the-right-word?

src/auth/src/credentials.rs
```diff
@@ -129,10 +129,50 @@ where
...
+    ///
+    /// # Parameters
+    /// * `extensions` - An `http::Extensions` map that can be used to pass additional
```
coryan, line 141: I think we should merge this as-is, but also consider refactoring to a common doc in a future PR.

src/auth/src/credentials.rs
```diff
@@ -129,10 +129,50 @@ where
...
+    /// * `Err(CredentialsError)`: If an error occurs while trying to fetch or
+    ///   generating the headers.
     pub async fn headers(&self, extensions: Extensions) -> Result<CacheableResource<HeaderMap>> {
         self.inner.headers(extensions).await
     }
 
+    /// Retrieves the universe domain associated with the credentials, if any.
+    ///
+    /// A "universe" is an isolated Google Cloud environment, such as the public
+    /// cloud or a sovereign/air-gapped deployment (e.g., Google Distributed Cloud).
```
coryan: Is "Google Distributed Cloud" a thing that appears in public docs? If so, please link it.  If not, let's use the public / marketing-approved product names.

coryan (approved):
A few small changes for you to make.

src/auth/src/credentials.rs
alvarowolfx: I used GDC here (and we have a public page), but thinking more and searching our docs, is not the right product to mention here. I'll remove the `e.g Google Distributed Cloud` part.

src/auth/src/credentials.rs
alvarowolfx: I try to call "authentication/credentials provider" as much as possible, but to be honest, not sure if is the best definition.

alvarowolfx (commented):
Addressed some comments.

PR #5419, opened 2026-04-15
feat(gax): add universe domain mismatch error

alvarowolfx:
Towards #3646

### DIFF
--- a/src/gax/src/client_builder.rs
+++ b/src/gax/src/client_builder.rs
@@ -94,6 +94,11 @@ impl Error {
         matches!(&self.0, ErrorKind::Transport(_))
     }
 
+    /// If true, the client universe domain does not match the credentials.
+    pub fn is_universe_domain_mismatch(&self) -> bool {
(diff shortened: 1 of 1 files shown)


src/gax/src/client_builder.rs
```diff
@@ -113,6 +124,8 @@ enum ErrorKind {
     DefaultCredentials(#[source] BoxError),
     #[error("could not initialize transport client")]
     Transport(#[source] BoxError),
+    #[error("universe domain mismatch")]
```
dbolduc, line 127: comment: At first I thought this should say more, but I think the `CredentialError` covers that.

optional: Mention how to fix the problem? Like "universe domain mismatch between the client and its credentials. Use `with_universe_domain()` to change the client's universe domain."

dbolduc (approved):

src/gax/src/client_builder.rs
alvarowolfx: the underlying CredentialError contains more details on how to fix it.

PR #5422, opened 2026-04-15
fix(wkt): implement camelCase conversion for FieldMask JSON mapping

### LINKED ISSUE #5410: Serialization / Deserialization of `FieldMask` may be incorrect
The documentation says:

```
In JSON, the same mask is represented as below:

{
  mask: "user.displayName,photo"
}
```

https://protobuf.dev/reference/protobuf/google.protobuf/#field-mask

That is not what we implement in Rust:

https://github.com/googleapis/google-cloud-rust/blob/330dc2f04bb620070f5acf17acf9226e6efe79a9/src/wkt/src/field_mask.rs#L297

Our tests only use single-word fields:

https://github.com/googleapis/google-cloud-rust/blob/330dc2f04bb620070f5acf17acf9226e6efe79a9/tests/protobuf/src/lib.rs#L85

So it is possible we have this wrong and the integration tests pass more by accident than design.

haphungw:
Fix a discrepancy between our implementation of FieldMask serialization/deserialization and the official Protocol Buffers JSON mapping specification.

According to the spec, in JSON, a FieldMask is represented as a string listing the paths separated by a comma, and each path is converted to camelCase. Our previous implementation simply joined and split paths with commas without performing any case conversion.

Fixes #5410

### DIFF
--- a/src/wkt/src/field_mask.rs
+++ b/src/wkt/src/field_mask.rs
@@ -294,7 +294,13 @@ impl serde::ser::Serialize for FieldMask {
     where
         S: serde::ser::Serializer,
     {
-        serializer.serialize_str(&self.paths.join(","))
+        let paths = self
+            .paths
+            .iter()
+            .map(|p| to_camel_case(p))
+            .collect::<Vec<_>>()
+            .join(",");
+        serializer.serialize_str(&paths)
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/wkt/src/field_mask.rs
```diff
@@ -326,11 +332,40 @@ impl serde::de::Visitor<'_> for PathVisitor {
...
+    for c in camel.chars() {
+        if c.is_ascii_uppercase() {
+            snake.push('_');
+            snake.push(c.to_ascii_lowercase());
+        } else {
+            snake.push(c);
+        }
+    }
+    snake
+}
```
haphungw, line 367: I implemented our own case conversion functions to avoid adding external dependencies, e.g., `heck`, which was considered (it's also pre-1.0).

tests/protobuf/src/lib.rs
```diff
@@ -68,6 +68,20 @@ pub async fn run(builder: ClientBuilder) -> Result<()> {
...
+        .add_secret_version()
+        .set_parent(&get.name)
+        .set_payload(
+            SecretPayload::new()
+                .set_data(bytes::Bytes::from(data))
+                .set_data_crc32c(checksum as i64),
+        )
+        .send()
+        .await?;
+
```
dbolduc, line 78: Remove? This doesn't do anything, and breaks the comment from the `update_*` method.

coryan (approved):
Thank you! I do like the addition of an integration test.

src/wkt/src/field_mask.rs
dbolduc: FTR, we can take internal dependencies on crates that are < 1.0. What we want to avoid is exposing types from crates that are < 1.0 in our public API.

dbolduc (approved):
Remove the extra RPC from the integration test. Or convince me it is necessary, but idts.

src/wkt/src/field_mask.rs
haphungw: ack. `heck` would require some dot handling (split the path by dots, convert each part, join them back), so i think this should still work.

tests/protobuf/src/lib.rs
haphungw: I kept the call because it is needed for the test. 

`update_secret()` maps "test-alias" to version 1i64. Since this is a new secret, if we don't create one version here, version 1 won't exist and the service will return an error when we try to set the alias: https://github.com/googleapis/google-cloud-rust/pull/5422/changes/e41c9a877356d23f37022644875c6ac3f395f1b2

I moved the comment to address the flow break.

dbolduc: Got it. Thanks

dbolduc (approved):

PR #5429, opened 2026-04-16
feat(spanner): add query_options to Statement

olavloite:
Adds a with_query_options to set the QueryOptions for a Statement.

### DIFF
--- a/src/spanner/src/client.rs
+++ b/src/spanner/src/client.rs
@@ -20,6 +20,7 @@ pub use crate::database_client::DatabaseClient;
 pub use crate::error::SpannerInternalError;
 pub use crate::from_value::{ConvertError, FromValue};
 pub use crate::key::{Key, KeyRange, KeySet, KeySetBuilder};
+pub use crate::model::execute_sql_request::QueryOptions;
 pub use crate::model::transaction_options::IsolationLevel;
 pub use crate::model::transaction_options::read_write::ReadLockMode;
 pub use crate::mutation::{Mutation, ValueBinder, WriteBuilder};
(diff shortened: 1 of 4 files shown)


sakthivelmanii (approved):

src/spanner/src/client.rs
```diff
@@ -20,6 +20,7 @@ pub use crate::database_client::DatabaseClient;
 pub use crate::error::SpannerInternalError;
 pub use crate::from_value::{ConvertError, FromValue};
 pub use crate::key::{Key, KeyRange, KeySet, KeySetBuilder};
+pub use crate::model::execute_sql_request::QueryOptions;
 pub use crate::model::transaction_options::IsolationLevel;
 pub use crate::model::transaction_options::read_write::ReadLockMode;
```
dbolduc, line 25: Why are we re-exporting these in the `mod client`?

Please open an issue to go through all of the re-exports in this module. Only one of them is a client.

We tend to have:
- `mod client` - clients
- `mod builder` - builders for RPC requests, and the client builder.
- `mod model` - all of the generated message types
- `mod model_ext` - handwritten message types that are fairly basic
- `mod error` - any custom errors specific to this crate
- `mod foo` some complicated thing foo. Like having a `mod result_set` is probably justified

We don't have any hard and fast rules about when to `mod foo` vs. when to add `foo` to `mod model_ext`.

dbolduc (approved):

src/spanner/src/client.rs
dbolduc: > Please open an issue

Nvm, I clicked the button: https://github.com/googleapis/google-cloud-rust/issues/5430

PR #5431, opened 2026-04-16
feat(pubsub): retry confirmed_ack partial failures in Subscriber

PhongChuong:

### DIFF
--- a/src/pubsub/src/subscriber/leaser.rs
+++ b/src/pubsub/src/subscriber/leaser.rs
@@ -374,8 +374,6 @@ pub(super) mod tests {
     use super::*;
     use crate::{Error, Response, Result};
     use google_cloud_gax::error::rpc::{Code, Status};
-    use google_cloud_gax::retry_result::RetryResult;
-    use google_cloud_gax::retry_state::RetryState;
     use google_cloud_rpc::model::ErrorInfo;
     use test_case::test_case;
     use tokio::sync::Mutex;
@@ -596,22 +594,12 @@ pub(super) mod tests {
     async fn confirmed_ack_success() -> anyhow::Result<()> {
         let (confirmed_tx, mut confirmed_rx) = unbounded_channel();
         let mut mock = MockStub::new();
-        mock.expect_acknowledge().times(1).return_once(|r, o| {
+        mock.expect_acknowledge().times(1).return_once(|r, _| {
             assert_eq!(
                 r.subscription,
                 "projects/my-project/subscriptions/my-subscription"
             );
             assert_eq!(sorted(&r.ack_ids), test_ids(0..10));
-            let retry = o.retry_policy().clone().unwrap();
-            let mut state = RetryState::default();
-            state.attempt_count = 1;
-            assert!(
-                !matches!(
-                    retry.on_error(&state, crate::Error::service(Status::default())),
-                    RetryResult::Continue(_)
-                ),
-                "Expected NeverRetry to not continue"
-            );
             Ok(Response::from(()))
         });
 
@@ -644,22 +632,12 @@ pub(super) mod tests {
     async fn confirmed_ack_failure() -> anyhow::Result<()> {
         let (confirmed_tx, mut confirmed_rx) = unbounded_channel();
         let mut mock = MockStub::new();
-        mock.expect_acknowledge().times(1).return_once(|r, o| {
+        mock.expect_acknowledge().times(1).return_once(|r, _| {
             assert_eq!(
                 r.subscription,
                 "projects/my-project/subscriptions/my-subscription"
             );
             assert_eq!(sorted(&r.ack_ids), test_ids(0..10));
-            let retry = o.retry_policy().clone().unwrap();
-            let mut state = RetryState::default();
-            state.attempt_count = 1;
-            assert!(
-                !matches!(
-                    retry.on_error(&state, crate::Error::service(Status::default())),
-                    RetryResult::Continue(_)
-                ),
-                "Expected NeverRetry to not continue"
-            );
             Err(Error::service(
                 Status::default()
                     .set_code(Code::FailedPrecondition)
--- a/src/pubsub/src/subscriber/leaser.rs
+++ b/src/pubsub/src/subscriber/leaser.rs
@@ -223,8 +223,6 @@ where
         };
 
         let sleep = async |d| tokio::time::sleep(d).await;
-        // Note: the retry policy is not directly used as attempt explicitly decides
-        // the retry logic by returning a synthetic error.
         let _ = retry_loop(
             attempt,
             sleep,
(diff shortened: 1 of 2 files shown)


src/pubsub/src/subscriber/retry_policy.rs
PhongChuong, line 52: Should we update this to include `Code::DeadlineExceeded`?

PhongChuong: As discussed offline, we should.
I'll update this in a followup PR.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -148,10 +153,14 @@ where
         let remaining_ids = Arc::new(Mutex::new(ack_ids));
         let last_error = Arc::new(Mutex::new(None));
 
+        let mut never_retry_options = self.exactly_once_options.clone();
+        never_retry_options.set_retry_policy(NeverRetry);
```
dbolduc: We should use `self.options` here. It only retries on stream reconnects.

dbolduc: I might have called this thing `at_least_once_options()` which was short-sighted. Maybe it should be called like `attempt_options()` ?

src/pubsub/src/subscriber/retry_policy.rs
```diff
@@ -101,6 +102,24 @@ pub(super) fn at_least_once_options(grpc_subchannel_count: usize) -> RequestOpti
     o
 }
 
+/// The policies for lease management RPCs in exactly-once delivery.
+pub(super) fn exactly_once_options(grpc_subchannel_count: usize) -> RequestOptions {
+    let mut o = RequestOptions::default();
+    o.set_retry_policy(
+        OnlyTransportErrors
+            .with_attempt_limit(grpc_subchannel_count as u32 + 1)
```
dbolduc: I think the inner retry loop should handle this.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -688,14 +715,60 @@ pub(super) mod tests {
...
+    #[tokio::test]
+    async fn confirmed_ack_partial_transient_failure_retry_success() -> anyhow::Result<()> {
+        let (confirmed_tx, mut confirmed_rx) = unbounded_channel();
+        let mut mock = MockStub::new();
+
+        let info = ErrorInfo::new().set_metadata([(test_id(1), "TRANSIENT_FAILURE_OTHER")]);
+        let err = response_with_error_info(vec![info]).unwrap_err();
+
+        mock.expect_acknowledge()
+            .times(1)
```
dbolduc, line 780: can we add a `mockall::Sequence` ? I never trust the ordering without it.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -186,19 +198,21 @@ where
...
         let _ = retry_loop(
             attempt,
             sleep,
             true,
-            retry_throttler(&self.options),
-            retry_policy(),
-            backoff_policy(),
+            retry_throttler(&self.exactly_once_options),
+            retry_policy(&self.exactly_once_options),
+            backoff_policy(&self.exactly_once_options),
```
dbolduc: Consider writing the interface in terms of a generic backoff policy, to avoid randomness and/or real sleeps in the test. Something like this:

https://github.com/googleapis/google-cloud-rust/blob/5d354cbb64eb5ff44bcfd7b729994a389839fcf2/src/pubsub/src/subscriber/stream.rs#L66-L94

The tradeoff is that you don't verify the default backoff policy in tests. Which I think is fine.

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Done.

src/pubsub/src/subscriber/retry_policy.rs
PhongChuong: Updated. PTAL.

src/pubsub/src/subscriber/leaser.rs
PhongChuong: I changed it to `rpc_options`.

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Updated to inject BackoffPolicy. PTAL.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -186,19 +217,21 @@ where
...
                     *ids_guard = remaining;
                     // Return a synthetic error to indicate that we should retry.
-                    Err(crate::Error::timeout("retry me"))
+                    Err(crate::Error::transport(http::HeaderMap::new(), "retry me"))
                 }
             }
         };
 
         let sleep = async |d| tokio::time::sleep(d).await;
+        // Note: the retry policy is not directly used as attempt explicitly decides
```
dbolduc: The retry policy is used right? It just retries on transport errors, which we return.

Maybe just remove the comment?

src/pubsub/src/subscriber/retry_policy.rs
```diff
@@ -94,13 +95,27 @@ impl BackoffPolicy for NoBackoff {
 /// Note that an RPC attempt that fails because a channel is closed may be
 /// retried on another channel that is closed. That is why we retry up to N
 /// times where N is the number of channels in the client.
-pub(super) fn at_least_once_options(grpc_subchannel_count: usize) -> RequestOptions {
+pub(super) fn rpc_options(grpc_subchannel_count: usize) -> RequestOptions {
```
dbolduc, line 98: Thanks

src/pubsub/src/subscriber/retry_policy.rs
```diff
@@ -319,6 +340,107 @@ pub(super) mod tests {
...
+            "non-transport error should not be retried"
+        );
+
+        // Verify time limit
+        let r = retry.remaining_time(&RetryState::default()).unwrap();
+        assert!(
+            r > Duration::from_secs(599) && r <= Duration::from_secs(600),
+            "Expected remaining time ~600s, got {:?}",
+            r
+        );
```
dbolduc: This scares me. It just looks like a timing thing.

Maybe leave it out of this function and (optionally) verify the remaining time on the interface directly? Or like debug the type and verify the attempt limit as a string (also kinda yucky)

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -576,13 +624,12 @@ pub(super) mod tests {
...
         let (confirmed_tx, mut confirmed_rx) = unbounded_channel();
         let mut mock = MockStub::new();
-        mock.expect_acknowledge().times(1).return_once(|r, o| {
+        mock.expect_acknowledge().times(1).return_once(|r, _| {
             assert_eq!(
                 r.subscription,
                 "projects/my-project/subscriptions/my-subscription"
             );
             assert_eq!(sorted(&r.ack_ids), test_ids(0..10));
-            verify_policies(o, 16);
```
dbolduc, line 585: We can put this back right?

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -615,13 +662,12 @@ pub(super) mod tests {
...
```
dbolduc, line 624: Same

dbolduc (approved):

src/pubsub/src/subscriber/retry_policy.rs
PhongChuong: Good catch.
I've updated the test to use `#[tokio::test(start_paused = true)]` and tokio instant instead. Seems to work.

dbolduc: Ah perfect

PR #5440, opened 2026-04-16
docs: document safe local backend usage for terraform integration tests

haphungw:
During local testing, running Terraform without a local backend can inadvertently modify the shared project or disable services due to shared state confusion (which happened recently). This documentation provides a safe, isolated path for developers setting up their own test projects.

### DIFF
--- a/.gcb/builds/README.md
+++ b/.gcb/builds/README.md
@@ -1,65 +1,63 @@
 # Terraform for Integration Test Resources
 
-This directory contains Terraform configurations to set up the resources needed for integration tests (Firestore, KMS, Buckets, etc.).
+This directory contains Terraform configurations to set up the resources needed
+for integration tests (Firestore, KMS, Buckets, etc.).
 
 ## Safe Usage for Personal Test Projects
 
-By default, this configuration uses a remote GCS backend which tracks the state of the shared `rust-sdk-testing` project. **Do not run `terraform apply` directly with the remote backend if you are targeting a personal test project, as it may interfere with the shared state or disable services in the shared project.**
+By default, this configuration uses a remote GCS backend which tracks the state
+of the shared `rust-sdk-testing` project. **Do not run `terraform apply`
+directly with the remote backend if you are targeting a personal test project,
+as it may interfere with the shared state or disable services in the shared
+project.**
 
-To safely use these Terraform scripts to set up resources in your personal test project:
+To safely use these Terraform scripts to set up resources in your personal test
+project:
 
-1.  **Set your project variable**:
+1. **Set your project variable**:
 
-    ```bash
-    PROJECT=$(gcloud config get project)
-    ```
+   ```bash
+   PROJECT=$(gcloud config get project)
+   ```
 
-2.  **Override the backend to use local state**:
-    Create a `backend_override.tf` file. This tells Terraform to store the state locally on your machine instead of the shared GCS bucket.
+1. **Override the backend to use local state**: Create a `backend_override.tf`
+   file. This tells Terraform to store the state locally on your machine instead
+   of the shared GCS bucket.
 
-    ```bash
-    cat > backend_override.tf <<EOF
-    terraform {
-      backend "local" {
-        path = "${HOME}/${PROJECT}-rust-sdk.tfstate"
-      }
-    }
-    EOF
-    ```
+   ```bash
+   cat > backend_override.tf <<EOF
+   terraform {
+     backend "local" {
+       path = "${HOME}/${PROJECT}-rust-sdk.tfstate"
+     }
+   }
+   EOF
+   ```
 
-3.  **Initialize Terraform**:
+1. **Initialize Terraform**:
 
-    ```bash
-    terraform init -reconfigure
-    ```
+   ```bash
+   terraform init -reconfigure
+   ```
 
-4.  **Plan and Apply**:
+1. **Plan and Apply**:
 
-    ### Option A: For Local Testing (Recommended)
+   ### Option A: For Local Testing (Recommended)
 
-    If you only need the resources for running tests locally (and do not need to set up GCB triggers), target only the `services` and `resources` modules. This avoids the need to create CI-specific service accounts:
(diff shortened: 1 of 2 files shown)


doc/contributor/howto-guide-set-up-development-environment.md
```diff
@@ -253,6 +253,9 @@ gcloud storage service-agent --project=${GOOGLE_CLOUD_PROJECT} \
     --authorize-cmek=projects/${GOOGLE_CLOUD_PROJECT}/locations/us-central1/keyRings/us-central1/cryptoKeys/storage-examples
 ```
 
+> [!NOTE]
+> Alternatively, if you prefer to use Terraform to automate this setup in your own project, see the instructions in [.gcb/builds/README.md](../../.gcb/builds/README.md) for a safe way to do this without affecting the shared project.
+
```
dbolduc: Can we delete the old instructions? These instructions seem simpler to me.

.gcb/builds/README.md
```diff
@@ -0,0 +1,59 @@
...
+    ```bash
+    gcloud iam service-accounts create integration-test-runner \
+        --display-name "Integration Test Runner"
+    ```
+    
+    Then run the full apply:
+    ```bash
+    terraform plan -var="project=${PROJECT}" -out="/tmp/builds.plan"
+    terraform apply "/tmp/builds.plan"
+    ```
```
dbolduc: Let's leave this out. It's not really relevant to test projects.

.gcb/builds/README.md
```diff
@@ -0,0 +1,59 @@
+# Terraform for Integration Test Resources
+
+This directory contains Terraform configurations to set up the resources needed for integration tests (Firestore, KMS, Buckets, etc.).
```
dbolduc: nit: flow lines to 80 characters? I think mdformat will complain.

.gcb/builds/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+3.  **Initialize Terraform**:
+    ```bash
+    terraform init -reconfigure
+    ```
+
+4.  **Plan and Apply**:
+    
+    ### Option A: For Local Testing (Recommended)
+    If you only need the resources for running tests locally (and do not need to set up GCB triggers), target only the `services` and `resources` modules. This avoids the need to create CI-specific service accounts:
```
dbolduc: I think there are some extra `resources` we don't need:

Like the build cache:

https://github.com/haphungw/google-cloud-rust/blob/0e9d81ae2f12527c566129d5cea2df1855b464c4/.gcb/builds/resources/main.tf#L24

and the GCB worker pool:

https://github.com/haphungw/google-cloud-rust/blob/0e9d81ae2f12527c566129d5cea2df1855b464c4/.gcb/builds/resources/main.tf#L86

Ideally we would move anything related to GCB into its own module, so we can ignore it when creating resources in our test project.

.gcb/builds/README.md
```diff
@@ -0,0 +1,59 @@
...
+
+2.  **Override the backend to use local state**:
+    Create a `backend_override.tf` file. This tells Terraform to store the state locally on your machine instead of the shared GCS bucket.
+    ```bash
+    cat > backend_override.tf <<EOF
+    terraform {
+      backend "local" {
+        path = "${HOME}/${PROJECT}-rust-sdk.tfstate"
+      }
+    }
```
coryan: The "Right Thing[tm]" may be to have your own terraform directory (maybe hosted in your personal fork or an internal google repo) where this file is committed, and then reference the existing modules using a github link:

https://developer.hashicorp.com/terraform/language/block/module#github-repository

coryan: This is not a reason to block anything. Your change is fine (modulo formatting), better than what we have today.

doc/contributor/howto-guide-set-up-development-environment.md
haphungw: done

.gcb/builds/README.md
haphungw: done

.gcb/builds/README.md
haphungw: noted. I'll keep that approach in mind as a future improvement.

.gcb/builds/README.md
haphungw: Yes that's a better structure. I'll handle the cleanup in another. I'll try to be extremely careful with it 🤦‍♀️

dbolduc: heads up on two things:

1. These modules have dependencies on each other to make sure things get defined in order
2. IIRC moving something from one module to another will delete and recreate it. So be careful about when/if you decide to try it. (There may be a way to migrate, but idk)

coryan: >  (There may be a way to migrate, but idk)

https://developer.hashicorp.com/terraform/language/block/moved

dbolduc: Ah, thanks.

haphungw: ack

.gcb/builds/README.md
```diff
@@ -0,0 +1,63 @@
...
+This directory contains Terraform configurations to set up the resources needed
+for integration tests (Firestore, KMS, Buckets, etc.).
+
+## Safe Usage for Personal Test Projects
+
+By default, this configuration uses a remote GCS backend which tracks the state
+of the shared `rust-sdk-testing` project. **Do not run `terraform apply`
+directly with the remote backend if you are targeting a personal test project,
+as it may interfere with the shared state or disable services in the shared
+project.**
```
coryan, line 12: Feels like we should be able to prevent accidental deletion of / creation of things.

Maybe we can have some kind of sigil resource that must exist (and is created outside terraform) before anything runs?  Or maybe something with the `prevent_destroy` flag?

Or maybe we just don't have permissions to delete these things, and only a robot can.

coryan: Created https://github.com/googleapis/google-cloud-rust/issues/5462 to track that idea.

coryan (approved):

PR #5443, opened 2026-04-16
chore: update librarian version and use list syntax for include_list

sofisl:
For https://github.com/googleapis/librarian/issues/4298

See https://github.com/googleapis/librarian/pull/5419 as reference for this work

This is PR 2 in this series: https://github.com/googleapis/librarian/issues/4769#issuecomment-4117482367


dbolduc (approved):

coryan:
In case the CI failures are hard to grok or access: we run `librarian tidy` as part of the CI, and it is printing:

```

2026/04/16 23:04:11 librarian: yaml: unmarshal errors:
--
  | line 805: cannot unmarshal !!seq into string
  | line 1495: cannot unmarshal !!seq into string
  | line 1683: cannot unmarshal !!seq into string
  | line 1719: cannot unmarshal !!seq into string
  | line 1726: cannot unmarshal !!seq into string
  | line 1732: cannot unmarshal !!seq into string
  | line 1740: cannot unmarshal !!seq into string
  | exit status 1

```

librarian.yaml
```diff
@@ -801,7 +801,8 @@ libraries:
             - .google.firestore.v1.ExplainStats
           api_path: google/firestore/v1
           template: convert-prost
-        - include_list: latlng.proto
+        - include_list:
+            - latlng.proto
```
coryan, line 805:
```suggestion
        include_list:
          - latlng.proto
```

librarian.yaml
```diff
@@ -1490,7 +1491,9 @@ libraries:
           output: src/storage/src/generated/convert/storage
           api_path: google/storage/v2
           template: convert-prost
-        - include_list: date.proto,expr.proto
+        - include_list:
+            - date.proto
+            - expr.proto
```
coryan, line 1496:
```suggestion
        include_list:
          - date.proto
          - expr.proto
```

I can only tell because I made the same mistake for Swift.

librarian.yaml
```diff
@@ -1676,7 +1679,11 @@ libraries:
       - protobuf-src
     rust:
       modules:
-        - include_list: api.proto,source_context.proto,type.proto,descriptor.proto
+        - include_list:
```
coryan, line 1682: Ditto.

librarian.yaml
```diff
@@ -1708,25 +1715,29 @@ libraries:
     skip_release: true
     rust:
       modules:
-        - include_list: conformance.proto
+        - include_list:
```
coryan, line 1718: Here too

librarian.yaml
```diff
@@ -1708,25 +1715,29 @@ libraries:
...
-          include_list: conformance.proto
+          include_list:
+            - conformance.proto
           module_path: crate::generated::gapic::model
           output: tests/protojson-conformance/src/generated/gapic
           api_path: conformance
           template: grpc-client
-        - include_list: conformance.proto
+        - include_list:
+            - conformance.proto
```
coryan, line 1732:
```suggestion
          include_list:
            - conformance.proto
```

coryan (dismissed):

librarian.yaml
sofisl: thank you!

librarian.yaml
sofisl: So, it seems like librarian tidy wants `include_list` to be first, so it sets it as the beginning of the list :|.

sofisl:
@dbolduc @coryan it seems like the generated diff is failing because showcase was moved to `stable`:

```

diff --git a/src/generated/showcase/.repo-metadata.json b/src/generated/showcase/.repo-metadata.json
--
  | index a20e3a3..6164273 100644
  | --- a/src/generated/showcase/.repo-metadata.json
  | +++ b/src/generated/showcase/.repo-metadata.json
  | @@ -8,6 +8,6 @@
  | "library_type": "GAPIC_AUTO",
  | "name": "showcase",
  | "name_pretty": "Client Libraries Showcase",
  | -    "release_level": "stable",
  | +    "release_level": "preview",
  | "repo": "googleapis/google-cloud-rust"
  | }
  | \ No newline at end of file
  | diff --git a/src/generated/showcase/README.md b/src/generated/showcase/README.md
  | index 4955432..0292796 100644
  | --- a/src/generated/showcase/README.md
  | +++ b/src/generated/showcase/README.md
  | @@ -2,6 +2,14 @@
  |  
  | <!-- Code generated by sidekick. DO NOT EDIT. -->
  |  
  | +**FEEDBACK WANTED:** We believe the APIs in this crate are stable, and
  | +do not anticipate any breaking changes are needed. We are looking for
  | +feedback before labeling the APIs "1.0". Changes (even breaking changes)
  | +are still possible, but not expected.
  | +
  | +We also believe the implementation is ready for production, bugs are
  | +still possible, but not expected.
  | +
  | **WARNING:** some RPCs have no corresponding Rust function to call them.
  | Typically these are streaming RPCs. We expect adding these RPCs in a
  | way that does not break the existing APIs or changes their behavior in a
  | @@ -12,12 +20,6 @@ GitHub repository.
  | Showcase represents both a model API and an integration testing surface for
  | client library generator consumption.
  |  
  | -The client library types and functions are stable and not expected to change.
  | -Please note that Google Cloud services do change from time to time. The client
  | -libraries are designed to preserve backwards compatibility when the service
  | -changes in compatible ways. For example, adding RPCs, or fields to messages
  | -should not introduce breaking changes to the client libraries.
  | -
  | ## Quickstart
  |  
  | The main types to work with this crate are the clients:
  | diff --git a/src/generated/showcase/src/lib.rs b/src/generated/showcase/src/lib.rs
  | index 639a423..b4fea17 100644
  | --- a/src/generated/showcase/src/lib.rs
  | +++ b/src/generated/showcase/src/lib.rs
  | @@ -16,6 +16,14 @@
  |  
  | //! Google Cloud Client Libraries for Rust - Client Libraries Showcase API
  | //!
  | +//! **FEEDBACK WANTED:** We believe the APIs in this crate are stable, and
  | +//! do not anticipate any breaking changes are needed. We are looking for
  | +//! feedback before labeling the APIs "1.0". Changes (even breaking changes)
  | +//! are still possible, but not expected.
  | +//!
  | +//! We also believe the implementation is ready for production, bugs are
  | +//! still possible, but not expected.
  | +//!
  | //! **WARNING:** some RPCs have no corresponding Rust function to call them.
  | //! Typically these are streaming RPCs. We expect adding these RPCs in a
  | //! way that does not break the existing APIs or changes their behavior in a
  | @@ -26,12 +34,6 @@
  | //! This crate contains traits, types, and functions to interact with Client Libraries Showcase API
  | //! Most applications will use the structs defined in the [client] module.
  | //!
  | -//! The client library types and functions are stable and not expected to change.
  | -//! Please note that Google Cloud services do change from time to time. The client
  | -//! libraries are designed to preserve backwards compatibility when the service
  | -//! changes in compatible ways. For example, adding RPCs, or fields to messages
  | -//! should not introduce breaking changes to the client libraries.
  | -//!
  | //! # Available Clients
  | //!
  | //! * [Compliance](client/struct.Compliance.html)


<br class="Apple-interchange-newline">
```

This seems unrelated to my change but I can check to see if there were any expected diffs for Librarian. If that's the case, is the recommended approach to run regenerate? I'll also ask in the Librarian chat to cross-reference.

dbolduc:
> @dbolduc @coryan it seems like the generated diff is failing because showcase was moved to `stable`:

At a guess, it was https://github.com/googleapis/librarian/pull/5276

> This seems unrelated to my change but I can check to see if there were any expected diffs for Librarian. If that's the case, is the recommended approach to run regenerate? I'll also ask in the Librarian chat to cross-reference.

Yeah, we should regenerate. This is the safe way:

https://github.com/googleapis/google-cloud-rust/blob/main/doc/contributor/howto-guide-generated-code-maintenance.md#all-libraries

But if you want to YOLO showcase only, you can do:

```
V=$(sed -n 's/^version: *//p' librarian.yaml)
go run github.com/googleapis/librarian/cmd/librarian@${V} generate google-cloud-showcase-v1beta1
```

dbolduc:
If we wanted to be super proper, we would send one PR updating librarian to a commit that flips showcase from `stable` to `preview` and merge that, then send this PR.... but mehhh who has time for that.

dbolduc:
> but mehhh who has time for that.

Me, apparently. I just sent #5445. We should merge that then rebase this PR.

sofisl:
@dbolduc thank you so much for your help and all the PRs!! Appreciate it, especially since I'm ramping up on Librarian and Rust. :)

coryan (approved):

dbolduc:
> @dbolduc thank you so much for your help and all the PRs!! Appreciate it, especially since I'm ramping up on Librarian and Rust. :)

No problem, thanks for the update!

Issue #5430, opened 2026-04-16
Audit exports from the spanner client mod


dbolduc:
We should go through all of the re-exports in `google-cloud-spanner`'s `client` module. Only one of them is a client.

We tend to have:
- `mod client` - clients
- `mod builder` - builders for RPC requests, and the client builder.
- `mod model` - all of the generated message types
- `mod model_ext` - handwritten message types that are fairly basic
- `mod error` - any custom errors specific to this crate
- `mod foo` some complicated thing foo. Like having a `mod result_set` is probably justified

We don't have any hard and fast rules about when to `mod foo` vs. when to add `foo` to `mod model_ext`.

_Originally posted by @dbolduc in https://github.com/googleapis/google-cloud-rust/pull/5429#discussion_r3094628357_

PR #5446, opened 2026-04-17
feat(spanner): support directed reads

olavloite:
Add support for setting DirectedReadOptions for SQL statements and read operations.


src/spanner/src/client.rs
```diff
@@ -20,6 +20,8 @@ pub use crate::database_client::DatabaseClient;
 pub use crate::error::SpannerInternalError;
 pub use crate::from_value::{ConvertError, FromValue};
 pub use crate::key::{Key, KeyRange, KeySet, KeySetBuilder};
+pub use crate::model::DirectedReadOptions;
+pub use crate::model::directed_read_options;
```
olavloite, line 24: @dbolduc I would like your opinion on this: `DirectedReadOptions` contains multiple nested options. Should we expose the module like this? The advantage being that any potential new option that is added to the module in the future will automatically also be exposed? Or should we explicitly list the nested options?
We would want to expose all nested options as it is today, and I also do not foresee any options that we might add in the future that we would not want to expose here.

dbolduc: > to expose here

Here is not the right place.

We can expose them in a `pub mod crate::model` instead of from the `mod crate::client`.

---

Do you only want to expose a subset of the types in the generated `mod model`? You can pick and choose things to export in the `src/lib.rs`. `src/storage/src/lib.rs` might be inspiring.

Also, there are generator options to skip generating a message or RPC. I think there is also an option to generate a message as `pub(crate)` if it is only used internally. (We can also add new generator options if needed).

With that being said, there is not any maintenance burden associated with exposing a generated model type we don't use in our API. It is just ideal not to do that.

olavloite: I've exposed it now in a `pub mod crate::model`.

-----

> Do you only want to expose a subset of the types in the generated mod model?

Yes, I think so. But let me pick this up as part of #5430. I think that:
1. We only want to expose the types that are actually being used in the public API of the Spanner client.
2. In some cases, like for `directed_read_options`, we want to expose all nested types, including types that might be added in the future.

sakthivelmanii (approved):

src/spanner/src/client.rs
dbolduc: Makes sense, that's what I figured.

> We only want to expose the types that are actually being used in the public API of the Spanner client.

+1.

>  we want to expose all nested types, including types that might be added in the future.

I can help you with that when we are ready. The generator has a flag for `included_ids` that we could potentially use. I designed it to be intuitive. So you could just specify the service/method/message you want, and it generates all the necessary components for that scope.

For example, in Pub/Sub, we say "do the whole publisher service, but for the subscriber service, only do these 3 RPCs."

https://github.com/googleapis/google-cloud-rust/blob/782279d29bd2cce63ce630d7f2cdd439dc849157/librarian.yaml#L1206-L1210

(Details on the algorithm here: https://github.com/googleapis/librarian/blob/962f7b29d821817d9b7ef915c77551e22a49e800/internal/sidekick/api/dependencies.go#L21-L56)

dbolduc: > I've exposed it now in a pub mod crate::model.

:+1:

PR #5447, opened 2026-04-17
feat(spanner): add option for DataBoost

olavloite:
Adds an option to use DataBoost with partitioned queries and partitioned reads.


src/spanner/src/batch_read_only_transaction.rs
```diff
@@ -230,6 +230,42 @@ impl BatchReadOnlyTransaction {
...
+    /// # let partitions = transaction
+    /// #     .partition_query(
+    /// #         Statement::builder("SELECT * FROM Users").build(),
+    /// #         PartitionOptions::default(),
+    /// #     )
+    /// #     .await?;
+    /// // On a worker receiving a partition, execute it with Data Boost:
+    /// let options = PartitionExecuteOptions::default()
+    ///     .with_data_boost(true);
+    /// let mut result_set = partitions[0].execute(&db_client, options).await?;
```
dbolduc, line 259: We tend to use builders so applications don't have to supply default options everywhere.

Maybe like `p.execute(&db_client).set_data_boost(true).send().await`

Thoughts?

dbolduc (approved):

src/spanner/src/batch_read_only_transaction.rs
olavloite: Fair point. However, our other `execute(..)` methods all really execute the statement, and I would prefer this method to be consistent with that pattern. So I've changed this so that `Partition` has a `with_data_boost` method (and could also hold more options in the future if needed). The call then looks like this:

```rust
    let mut result_set = partitions[0].clone()
        .with_data_boost(true)
        .execute(&db_client)
        .await?;
```

sakthivelmanii (approved):

src/spanner/src/batch_read_only_transaction.rs
dbolduc: Makes sense, better than my suggestion

dbolduc (approved):

src/spanner/src/batch_read_only_transaction.rs
```diff
@@ -645,4 +673,71 @@ pub(crate) mod tests {
...
+            .set_session("projects/p/instances/i/databases/d/sessions/123")
+            .set_transaction(crate::model::TransactionSelector {
+                selector: Some(crate::model::transaction_selector::Selector::Id(
+                    b"tx_id_2".to_vec().into(),
+                )),
+                ..Default::default()
+            })
+            .set_table("Users")
+            .set_columns(vec!["Id".to_string(), "Name".to_string()])
+            .set_partition_token(b"partition_token_456".to_vec());
```
dbolduc: Simplify this.

Also if you find yourself typing `crate::model::Foo`, consider `use crate::model::Foo` for readability.

olavloite: Done

Issue #5450, opened 2026-04-17
Flaky test `pubsub::run_ordered_roundtrip` timeouts on `CreateSubscription`


haphungw:
[Error Log:](https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/6c26b2e2-4e02-4fb5-83c8-183bbf724e33?project=rust-sdk-testing&e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=allow_workbench_image_override)

```
2026-04-17T14:45:56.819678Z WARN ThreadId(07) experimental.client.request: Error { kind: Timeout, source: Some(reqwest::Error { kind: Request, url: "https://pubsub.googleapis.com/v1/projects/rust-sdk-testing/subscriptions/...", source: TimedOut }) } rpc.system.name="http" rpc.method="google.pubsub.v1.Subscriber/CreateSubscription" gcp.client.version="0.33.2" gcp.client.repo="googleapis/google-cloud-rust" gcp.client.artifact="google-cloud-pubsub" gcp.schema.url="https://opentelemetry.io/schemas/1.39.0" url.domain="pubsub" ... error.type="CLIENT_TIMEOUT"
```

dbolduc:
`pubsub::run_exactly_once_nack_then_confirmed_ack` also timed out creating a subscription:

https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/e4b90621-5654-4bc1-9d59-0b996117f6cb;step=2?e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=allow_workbench_image_override&project=rust-sdk-testing

```
full error: the request exceeded the request deadline error sending request for url (https://pubsub.googleapis.com/v1/projects/rust-sdk-testing/subscriptions/subscription-GA8dylGytVwfKJdsKxjQbd1idXBiH00Sta2DfZBhqbxb9cakmIe45WRBJV3pw9OFjqeD7FE06utVGQXIN0VU0nQpo5HYVpYHv01DgGixumZXitniACAwJxI6kPKbEGfEQUmGQ76rQ5volVKhIhSHISS4ATDgqfAb9LH9hknERiFABVrPxdRQtubv0PX8COhdBr8301sutI2QYWzuA2brSmH5kEzBzncYWTJcnG1rZd0UFmdUIW?%24alt=json%3Benum-encoding%3Dint)
Caused by:
[...]

PR #5451, opened 2026-04-17
ci(pubsub): disable flaky test

haphungw:
For https://github.com/googleapis/google-cloud-rust/issues/5450

### DIFF
--- a/tests/pubsub/tests/driver.rs
+++ b/tests/pubsub/tests/driver.rs
@@ -138,6 +138,7 @@ mod pubsub {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[ignore = "flaky test, see #5450"]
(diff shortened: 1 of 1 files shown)


tests/pubsub/tests/driver.rs
```diff
@@ -138,6 +138,7 @@ mod pubsub {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[ignore = "flaky test, see #5450"]
```
coryan: I think we should use "TODO(#5450)", maybe followed by " - enable once it is deflaked".  That way we can search for all the TODOs with the same regular expression.

coryan (approved):

tests/pubsub/tests/driver.rs
haphungw: done

Issue #4685, opened 2026-02-13
Write additional hand crafted samples for Pub/Sub


suzmue:
The critical samples are complete (#3017).

The remaining samples are listed here. If you are choosing to add one, please create a sub-issue from the task item and assign yourself.

### Publish

- [ ] `pubsub_publish`
- [ ] `pubsub_publish_avro_records`
- [ ] `pubsub_publish_proto_messages`
- [ ] `pubsub_publish_custom_attributes`
- [ ] `pubsub_publish_with_error_handler`
- [ ] `pubsub_publish_messages_error_handler`
- [ ] `pubsub_publish_otel_tracing`
- [ ] `pubsub_publisher_with_compression`
- [ ] `pubsub_publisher_flow_control` 
- [ ] `pubsub_publisher_concurrency_control`

### Subscribe

- [ ] `pubsub_subscriber_async_pull`
- [ ] `pubsub_subscriber_sync_pull`
- [ ] `pubsub_subscriber_async_pull_custom_attributes`
- [ ] `pubsub_subscribe_avro_records`
- [ ] `pubsub_subscribe_avro_records_with_revisions`
- [ ] `pubsub_subscribe_proto_messages`
- [ ] `pubsub_subscribe_otel_tracing`
- [ ] `pubsub_optimistic_subscribe`

### Admin (Topics, Subscriptions, & Schemas)

- [ ] `pubsub_update_topic_type`
- [ ] `pubsub_update_topic_schema`
- [ ] `pubsub_get_topic_policy`
- [ ] `pubsub_set_topic_policy`
- [ ] `pubsub_test_topic_permissions`
- [ ] `pubsub_detach_subscription`
- [ ] `pubsub_get_subscription_policy`
- [ ] `pubsub_set_subscription_policy`
- [ ] `pubsub_test_subscription_permissions`
- [ ] `pubsub_create_push_subscription`
- [ ] `pubsub_create_unwrapped_push_subscription`
- [ ] `pubsub_update_push_configuration`
- [ ] `pubsub_create_bigquery_subscription`
- [ ] `pubsub_create_cloud_storage_subscription`
- [ ] `pubsub_create_subscription_with_filter`
- [ ] `pubsub_dead_letter_create_subscription`
- [ ] `pubsub_dead_letter_update_subscription`
- [ ] `pubsub_dead_letter_remove`
- [ ] `pubsub_create_topic_with_schema`
- [ ] `pubsub_create_topic_with_schema_revisions`
[...]

PR #5454, opened 2026-04-17
cleanup(pubsub): run subscriber samples in parallel

dbolduc:
Our subscriber samples each run for 10s. We currently have 3, and I am going to add 3 more.

We should run them all in parallel to save time. At the moment this shaves ~20s off the build time.

Related to #4685

### DIFF
--- a/src/pubsub/examples/src/lib.rs
+++ b/src/pubsub/examples/src/lib.rs
@@ -45,24 +45,43 @@ pub async fn run_subscription_samples(
 ) -> anyhow::Result<()> {
     let client = SubscriptionAdmin::builder().build().await?;
     let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
-
     let topic_id = topic_name.split("/").last().unwrap();
+
+    // Our subscriber examples each take 10 seconds to complete, so we run them
+    // in parallel.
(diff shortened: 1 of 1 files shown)


src/pubsub/examples/src/lib.rs
```diff
@@ -45,24 +45,43 @@ pub async fn run_subscription_samples(
...
         .await?;
-    subscriber::exactly_once::sample(&project_id, &id).await?;
+    slow_tasks.spawn({
+        let (project_id, id) = (project_id.clone(), id.clone());
+        async move { subscriber::exactly_once::sample(&project_id, &id).await }
+    });
+
+    // Await the result of the slow subscriber examples.
+    while let Some(task) = slow_tasks.join_next().await {
+        task??;
```
dbolduc, line 79: It occurs to me that if any of these tasks fail, we will not know which one.

Maybe I should return like a `(name: &'static str, result: Result<()>)` pair, so I can at least print out which test it was.

Let me take a TODO here, because I have a few PRs stacked on top of this one: https://screenshot.googleplex.com/7MjqtogDDgq3kAN. I can address this at the end.

suzmue (approved):

PR #5460, opened 2026-04-17
refactor(pubsub): return weak senders from lease loop

### LINKED ISSUE #5255: Consider yielding weak senders + cancellation token from lease loop
> nit: can you make clear that this is the only place that is allowed to hold the strong version? That way we don't come back in later and try to do a refactor to pass the Strong Sender somewhere else, without realizing there are consequences. 

 _Originally posted by @suzmue in [#5251](https://github.com/googleapis/google-cloud-rust/pull/5251/changes#r3030789036)_

---

This code could be clearer. The reader has to keep too much in their head about who holds strong vs. weak senders for these channels, as the context for this is split across multiple structs.

It would be simpler if only the `MessageStream` only ever interacts with weak senders.

dbolduc:
`LeaseLoop` owns the shutdown behavior, instead of having the logic split between the `MessageStream` and the `LeaseLoop`.

Fixes #5255

### DIFF
--- a/src/pubsub/src/subscriber/lease_loop.rs
+++ b/src/pubsub/src/subscriber/lease_loop.rs
@@ -118,18 +118,6 @@ impl LeaseLoop {
             cancel,
         }
     }
-
-    #[cfg(test)]
-    #[track_caller]
-    fn strong_ack_tx(&self) -> UnboundedSender<Action> {
-        self.ack_tx.upgrade().expect("shutdown has not begun")
-    }
-
-    #[cfg(test)]
-    #[track_caller]
-    fn strong_message_tx(&self) -> UnboundedSender<NewMessage> {
-        self.message_tx.upgrade().expect("shutdown has not begun")
-    }
 }
 
 // Shuts down lease management.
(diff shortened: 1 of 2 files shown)


src/pubsub/src/subscriber/lease_loop.rs
```diff
@@ -91,10 +113,23 @@ impl LeaseLoop {
...
+    #[track_caller]
+    fn strong_ack_tx(&self) -> UnboundedSender<Action> {
+        self.ack_tx.upgrade().expect("shutdown has not begun")
+    }
+
+    #[cfg(test)]
+    #[track_caller]
+    fn strong_message_tx(&self) -> UnboundedSender<NewMessage> {
+        self.message_tx.upgrade().expect("shutdown has not begun")
+    }
```
suzmue: nit: We can add an impl LeaseLoop {} block in mod tests to define these helpers. Then we don't need to add these cfg(test) guarded blocks in the main code.

suzmue (approved):

suzmue:
> Fixes #5255
> 
> `LeaseLoop` owns the shutdown behavior, instead of having the logic split between the `MessageStream` and the `LeaseLoop`. Seems slightly cleaner.

nit: PR description cleanup, put fixes at end. seems slightly cleaner is also implied by the fact you are doing it. 

LeaseLoop owns the shutdown behavior, instead of having the logic split between the MessageStream and the LeaseLoop.

Fixes #5255

src/pubsub/src/subscriber/lease_loop.rs
coryan: Darn... that is brilliant. We should do it more...

dbolduc: Done. Cool idea.

dbolduc:
> PR description

Updated

> put fixes at end

I am an old dog, this may take me a while.

> seems slightly cleaner is also implied by the fact you are doing it.

:laughing: 

Removed. My intention was to communicate to the reviewer that I do not think this PR is the most consequential refactor of the 21st century.

PR #5464, opened 2026-04-18
impl(gax-internal): host resolution for custom universe domains

alvarowolfx:
Towards #3646


src/gax-internal/src/host.rs
```diff
@@ -34,15 +39,21 @@ pub(crate) fn header(endpoint: Option<&str>, default_endpoint: &str) -> Result<S
...
+    origin_and_header(endpoint, default_endpoint, universe_domain).map(|(origin, _)| origin)
 }
 
 fn origin_and_header(
     endpoint: Option<&str>,
     default_endpoint: &str,
+    universe_domain: &str,
 ) -> Result<(Uri, String), HostError> {
-    let default_origin = Uri::from_str(default_endpoint).map_err(HostError::Uri)?;
+    let service_endpoint = default_endpoint.replace(DEFAULT_UNIVERSE_DOMAIN, universe_domain);
```
dbolduc: we should determine the `service` before doing this substitution.

(Can/should we provide the `service` to this function instea of the `default_endpoint`? I don't remember what the callers look like)

src/gax-internal/src/host.rs
```diff
@@ -58,17 +69,30 @@ fn origin_and_header(
...
-    let (Some(prefix), Some(service)) = (
-        custom_host.strip_suffix(".googleapis.com"),
-        default_host.strip_suffix(".googleapis.com"),
-    ) else {
+
+    let default_suffix = format!(".{DEFAULT_UNIVERSE_DOMAIN}");
+    let universe_suffix = format!(".{universe_domain}");
+
+    let service = default_host
+        .strip_suffix(&universe_suffix)
```
dbolduc: then we don't need this line.

src/gax-internal/src/host.rs
```diff
@@ -58,17 +69,30 @@ fn origin_and_header(
         .ok_or_else(|| HostError::MissingAuthority(endpoint.to_string()))?
         .host()
         .to_string();
-    let (Some(prefix), Some(service)) = (
-        custom_host.strip_suffix(".googleapis.com"),
-        default_host.strip_suffix(".googleapis.com"),
-    ) else {
+
+    let default_suffix = format!(".{DEFAULT_UNIVERSE_DOMAIN}");
```
dbolduc: nit: this allocation is unnecessary. I'd rather spell out `".googleapis.com"` than reuse the const and allocate an extra string when this is called.

src/gax-internal/src/host.rs
```diff
@@ -58,17 +69,30 @@ fn origin_and_header(
...
+    let Some(prefix) = prefix else {
+        return Ok((custom_origin, custom_host));
+    };
     let parts: Vec<&str> = prefix.split(".").collect();
     match &parts[..] {
-        // This is a regional endpoint. It should be used as the host.
+        // This is a global or regional endpoint. It should be used as the host.
         // `{service}.{region}.rep.googleapis.com`
-        [s, _, "rep"] if *s == service => Ok((custom_origin, custom_host)),
+        [s] | [s, _, "rep"] if *s == service => Ok((custom_origin, custom_host)),
```
dbolduc: comment: hmm, so this is needed now for the `.with_endpoint("<service>.googleais.com").with_universe_domain("my-ud.com")` case.

Otherwise we would fallback to the default host for the custom universe.

src/gax-internal/src/host.rs
```diff
@@ -123,15 +149,16 @@ mod tests {
     #[test_case("http://restricted.googleapis.com", "test.googleapis.com"; "VPC-SC restricted")]
     #[test_case("http://test-my-private-ep.p.googleapis.com", "test.googleapis.com"; "PSC custom endpoint")]
     #[test_case("https://us-central1-test.googleapis.com", "us-central1-test.googleapis.com"; "locational endpoint")]
-    #[test_case("https://us-central1-wrong.gooogleapis.com", "test.googleapis.com"; "locational but with bad service")]
-    #[test_case("https://us-central1test.gooogleapis.com", "test.googleapis.com"; "maybe locational with missing dash")]
-    #[test_case("https://-test.gooogleapis.com", "test.googleapis.com"; "maybe locational with missing location")]
```
dbolduc, line 128: why remove these?

src/gax-internal/src/host.rs
dbolduc: This seems too complicated. I think it is a result of substituting the universe domain in the default endpoint. Can we stop doing that?

We will also mess up theoretical[^1] cases like `.with_endpoint("private.googleapis.com").with_universe_domain("my-ud.com")`. I think it should use `"<service>.googleapis.com"` as the host, not `"<service>.my-ud.com"`.

[^1]: While I don't think it is a real example, it seems more consistent to enforce this behavior.

src/gax-internal/src/host.rs
```diff
@@ -58,17 +69,30 @@ fn origin_and_header(
         .ok_or_else(|| HostError::MissingAuthority(endpoint.to_string()))?
         .host()
         .to_string();
-    let (Some(prefix), Some(service)) = (
-        custom_host.strip_suffix(".googleapis.com"),
-        default_host.strip_suffix(".googleapis.com"),
-    ) else {
+
+    let default_suffix = format!(".{DEFAULT_UNIVERSE_DOMAIN}");
+    let universe_suffix = format!(".{universe_domain}");
```
dbolduc: nit: we could avoid this allocation if we pass in the universe suffix. Your call on if that hurts readability too much.

dbolduc (changes requested):
Can you try skipping this substitution? I think it will simplify the logic.

https://github.com/alvarowolfx/google-cloud-rust/blob/9e13b1b2be5138d4fc52782c66009615fa09f8f0/src/gax-internal/src/host.rs#L55

src/gax-internal/src/host.rs
alvarowolfx: changing this function to received the `service` will span to changes in way too many places. The `default_endpoint` comes from the generated code and based on the service protos. A lot of tests also pass `localhost` or other variations of local servers, so we can't always derive a service from that, which is why we have this fallback https://github.com/googleapis/google-cloud-rust/pull/5464/changes#diff-1e0260d3179b4f69b243f2a46346bc7f6bc6425434fe481d0fd430f82c6353beR76

src/gax-internal/src/host.rs
alvarowolfx: going to add back. Those had `gooogleapis.com` as host, with 3 `o`, so was being treated as custom endpoint and not working. Will fix the typo and add back.

dbolduc: > Those had gooogleapis.com as host, with 3 o

LOL. Didn't even notice.

src/gax-internal/src/host.rs
```diff
@@ -34,35 +39,53 @@ pub(crate) fn header(endpoint: Option<&str>, default_endpoint: &str) -> Result<S
...
     default_endpoint: &str,
+    universe_domain: &str,
 ) -> Result<(Uri, String), HostError> {
     let default_origin = Uri::from_str(default_endpoint).map_err(HostError::Uri)?;
     let default_host = default_origin
         .authority()
         .expect("missing authority in default endpoint")
         .host()
         .to_string();
+    let scheme = default_origin.scheme().map(|s| s.as_str());
```
dbolduc, line 61: 👍

another variable i did not consider.

src/gax-internal/src/host.rs
```diff
@@ -34,35 +39,53 @@ pub(crate) fn header(endpoint: Option<&str>, default_endpoint: &str) -> Result<S
...
-    ) else {
-        return Ok((default_origin, default_host));
+
+    let universe_suffix = format!(".{universe_domain}");
+    let (prefix, universe_domain) = if let Some(prefix) = custom_host.strip_suffix(&universe_suffix) {
+        (prefix, universe_domain)
+    } else if let Some(prefix) = custom_host.strip_suffix(&default_suffix) {
+        (prefix, DEFAULT_UNIVERSE_DOMAIN)
+    } else {
+        return Ok((custom_origin, custom_host));
```
dbolduc, line 88: nit: Add a comment like "If we don't recognize the universe domain, use the custom endpoint."

src/gax-internal/src/host.rs
```diff
@@ -79,10 +102,26 @@ fn origin_and_header(
...
+    service: &str,
+    universe_domain: &str,
+) -> Result<(Uri, String), HostError> {
+    let host = format!("{service}.{universe_domain}");
+    let origin_str = if let Some(s) = scheme {
+        format!("{s}://{host}")
+    } else {
+        host.clone()
+    };
+    let origin = Uri::from_str(&origin_str).map_err(HostError::Uri)?;
```
dbolduc, line 121: nit: Cloning the host is unnecessary when we only need a reference to it.

Consider:

```suggestion
    let mut parts = Uri::Parts::default();
    parts.schema = scheme.map(|s| Uri::Schema::from_str(s).map_error(HostError::Uri)?);
    parts.authority = Some(Uri::Authority::from_str(&host).map_error(HostError::Uri)?);
    let origin = Uri::from_parts(parts);
```

src/gax-internal/src/host.rs
```diff
@@ -34,35 +39,53 @@ pub(crate) fn header(endpoint: Option<&str>, default_endpoint: &str) -> Result<S
...
     let default_host = default_origin
         .authority()
         .expect("missing authority in default endpoint")
         .host()
         .to_string();
+    let scheme = default_origin.scheme().map(|s| s.as_str());
 
-    let Some(endpoint) = endpoint else {
+    let default_suffix = format!(".{DEFAULT_UNIVERSE_DOMAIN}");
+    let service = default_host.strip_suffix(&default_suffix);
```
dbolduc, line 64: nit: I still vote for spelling out ".googleapis.com" as a `&'static str`.

src/gax-internal/src/host.rs
dbolduc: ^ I still think this too.

dbolduc (approved):
LGTM.

There are some micro-optimizations that no one would ever notice... But it's a good habit to incorporate them into your game. Across a code base, they can add up.

src/gax-internal/src/host.rs
dbolduc: meh, the `from_parts` idea doesn't work. Let me try again.

```rs
    let origin = if let Some(s) = scheme {
        let o = format!("{s}://{host}");
        Uri::from_str(&o)
    } else {
        Uri::from_str(&host)
    }.map_err(HostError::Uri)?;
```

This at least skips the clone of host for the missing schema case.

Issue #5459, opened 2026-04-17
Run missing_docs lint for handwritten libraries


suzmue:
https://doc.rust-lang.org/beta/nightly-rustc/rustc_lint/builtin/static.MISSING_DOCS.html can help us catch places where we have unintentionally omitted documentation.

Intentional omissions can still be allowed via `#[allow(missing_docs)]` annotations.

suzmue:
The two phases for making this happen are:

- Phase 1: Fix all warnings in the existing handwritten crates

Enable warn for each handwritten crate and fix any warnings. Also manually add allows for any contained generated crates.

- Phase 2: Turn on warnings by default for handwritten crates
 
We can do this by setting the workspace missing_docs lint to warn which new crates will inherit. But to do this we will need to update the generated libraries to NOT inherit this (likely by updating their generated cargo.toml to set allow).

suzmue:
At this point in time, we don't plan to turn on the lint for the whole workspace.

PR #5468, opened 2026-04-20
cleanup(pubsub): enable missing_docs warning and document Handler

suzmue:
For #5459

### DIFF
--- a/src/pubsub/src/subscriber/handler.rs
+++ b/src/pubsub/src/subscriber/handler.rs
@@ -127,11 +127,11 @@ pub enum Handler {
     ///
     /// ```
     /// # use google_cloud_pubsub::subscriber::handler::{Handler, ExactlyOnce};
-    /// # async fn on_message(h: Handler) {
+    /// # async fn on_message(h: Handler) -> Result<(), Box<dyn std::error::Error>> {
     /// if let Handler::ExactlyOnce(h) = h {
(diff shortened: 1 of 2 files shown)


coryan (approved):

src/pubsub/src/subscriber/handler.rs
```diff
@@ -108,7 +108,31 @@ pub(super) enum Action {
...
+    ///
+    /// The handler type is determined by the subscription configuration.
+    ///
+    /// ```
+    /// # use google_cloud_pubsub::subscriber::handler::{Handler, ExactlyOnce};
+    /// # async fn on_message(h: Handler) {
+    /// if let Handler::ExactlyOnce(h) = h {
+    ///     let _ = h.confirmed_ack().await;
+    /// }
+    /// # }
```
dbolduc: Seems odd to want a confirmed ack, but then discard the result.

```suggestion
    /// # async fn on_message(h: Handler) -> anyhow::Result<()> {
    /// if let Handler::ExactlyOnce(h) = h {
    ///     h.confirmed_ack().await?;
    /// }
    /// # Ok(()) }
```

^ don't assume this compiles.

dbolduc (approved):

src/pubsub/src/subscriber/handler.rs
suzmue: good point will fix.

PR #5480, opened 2026-04-20
cleanup(storage): enable missing_docs warning

suzmue:
For #5459

### DIFF
--- a/src/storage/src/model_ext.rs
+++ b/src/storage/src/model_ext.rs
@@ -304,9 +304,9 @@ pub(crate) enum RequestedRange {
 #[non_exhaustive]
 #[allow(dead_code)]
 pub struct WriteObjectRequest {
-    /// The specification for the write operation.
+    /// The object attributes and pre-conditions for the write operation.
     pub spec: crate::model::WriteObjectSpec,
-    /// The common request parameters.
+    /// Additional request parameters that are not part of the object attributes.
     pub params: Option<crate::model::CommonObjectRequestParams>,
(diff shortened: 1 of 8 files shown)


src/storage/src/model_ext.rs
```diff
@@ -304,7 +304,9 @@ pub(crate) enum RequestedRange {
 #[non_exhaustive]
 #[allow(dead_code)]
 pub struct WriteObjectRequest {
+    /// The specification for the write operation.
```
coryan, line 307: Maybe something like:


```suggestion
    /// The object attributes and pre-conditions for the write operation.
```

src/storage/src/model_ext.rs
```diff
@@ -304,7 +304,9 @@ pub(crate) enum RequestedRange {
...
     pub spec: crate::model::WriteObjectSpec,
+    /// The common request parameters.
```
coryan, line 309:
```suggestion
    /// Additional request parameters that are not part of the object attributes.
```

src/storage/src/storage.rs
```diff
@@ -19,6 +19,7 @@ pub(crate) mod common_options;
 pub(crate) mod open_object;
 pub(crate) mod perform_upload;
 pub(crate) mod read_object;
+/// Request options for the storage service.
```
coryan, line 22: In general we put these inside the module, can we do the same here?

coryan (approved):
A few small nits.

PR #5485, opened 2026-04-21
chore: update sources and regenerate

noahdietz:
Update googleapis and discovery API sources, regenerate clients.

[googleapis changes since last update](https://github.com/googleapis/googleapis/compare/f5cb7afc40b63d52f43bc306cb9b64a87b681aea..939ba3bf8408af83f0f73ae35c76c4b11a8c8c8d)
[discovery changes since last update](https://github.com/googleapis/discovery-artifact-manager/compare/a9217d2c6b2ce003fde50c0729c2bb8434434b5b..1d266e76e732e6835d6df5002dc790430567d0c2#diff-b71d5fba26f70ab075bc30ab7d71b0c36e6cfee4ef33ed4941ff6b1ed8999004) (just for compute)


Cargo.lock
```diff
@@ -2094,6 +2094,7 @@ dependencies = [
  "google-cloud-gax-internal",
  "google-cloud-longrunning",
  "google-cloud-lro",
+ "google-cloud-rpc",
```
noahdietz, line 2097: @coryan this looks....suspect?

noahdietz: I believe it is because `chronicle` add a direct dependency on `google.rpc.Status` in their API surface and that has shown up in the generated updates here.

noahdietz: That tracks based on the diff of their new proto and the BUILD.bazel file getting `google/rpc/status_proto` as a new dep

image

coryan: yes, `sidekick` detects many new dependencies automatically.  There is a corresponding change in `generated/cloud/chronicle/v1/Cargo.toml`...

src/generated/cloud/chronicle/v1/Cargo.toml
```diff
@@ -47,6 +47,7 @@ gaxi                               = { workspace = true, features = ["_internal-
 google-cloud-gax.workspace         = true
 google-cloud-longrunning.workspace = true
 google-cloud-lro.workspace         = true
+google-cloud-rpc.workspace         = true
```
coryan, line 50: @noahdietz see, this is where the generator caught the new dep.  The changes in `Cargo.lock` are just a side-effect.

coryan (commented):
I need to run.. sorry.

src/generated/cloud/chronicle/v1/Cargo.toml
noahdietz: Very good, thank you !

Cargo.lock
noahdietz: Great, all good then. TY!

noahdietz:
> I need to run.. sorry.

No problem

PhongChuong (approved):

Issue #3306, opened 2025-09-09
Support User Project option for StorageControl API


suzmue:
Attempting to run the samples for disabling and getting status on a requester pays enabled object (#3291) reveal that we should have a way to provide an option (like the C++ library does) to set the x-goog-user-project header for storage control requests.

Related to #3189 and they should likely use the same design.

dbolduc:
Should we add user project to `RequestOptions`, `RequestOptionsBuilder` ?

https://github.com/googleapis/google-cloud-rust/blob/e40933ae224e4f7db2c0aadd6bb1f70216958e18/src/gax/src/options.rs#L43

and handle it in the `gaxi` clients

https://github.com/googleapis/google-cloud-rust/blob/e40933ae224e4f7db2c0aadd6bb1f70216958e18/src/gax-internal/src/grpc.rs#L215

https://github.com/googleapis/google-cloud-rust/blob/e40933ae224e4f7db2c0aadd6bb1f70216958e18/src/gax-internal/src/http.rs#L96

suzmue:
> Should we add user project to RequestOptions, RequestOptionsBuilder ?

I agree that seems like the right place for it.

joshuatants:
Samples written in #5594

Issue #3189, opened 2025-08-29
Support User Project option for storage::ReadObject


suzmue:
Set the x-goog-user-project header. See https://cloud.google.com/storage/docs/json_api/v1/parameters#xgooguserproject.

There are other options we should consider in https://github.com/googleapis/google-cloud-cpp/blob/main/google/cloud/internal/rest_set_metadata.cc#L36

joshuatants:
Samples written in #5594

PR #5421, opened 2026-04-15
impl(gax): add source universe domains on mismatch err

alvarowolfx:
Towards #3646

PR #5490, opened 2026-04-22
feat(storage): add user_project option

joshuatants:
Adds a `with_user_project` method to `Storage` and `StorageControl` request builders to support Requester Pays buckets. When set, the internal gRPC and HTTP transports emit an `x-goog-user-project` header which takes precedence over any `quota_project_id` configured in the client credentials.

For #3189 and #3306.


joshuatants:
Hi @googleapis/cloud-sdk-rust-team in this PR I add functionality to `gax` to support the request level setting of the `UserProject`. This seems like it should require a bump in the minor number (hence the failing minimal-versions checks). Looking around it _seems_ that this is handled automatically by the tooling. I'm wondering what I should do here?

1. Split this PR into two: one for the `gax` change which goes in first, another for the `storage` change.
2. Keep the PR as in but add another `feat(storage)` line to cover storage.

Thank you for the advice.

dbolduc:
> I'm wondering what I should do here?

The tests are failing because there is new logic in `gax-internal` that is not present in the latest released `gax-internal`.

The tests should pass if you bump the `gax-internal` version for the crate:

https://github.com/googleapis/google-cloud-rust/blob/562640dd4d223edd11fd5839a7c9e0eedf146cdd/src/gax-internal/Cargo.toml#L17

... as well as the minimum version of `gax-internal` set in the workspace (which is being used by storage in the failing tests):

https://github.com/googleapis/google-cloud-rust/blob/562640dd4d223edd11fd5839a7c9e0eedf146cdd/Cargo.toml#L473

The change is purely additive, so make the version `0.7.13`.

(The same would have been true for adding a public API to `gax` but note that we already did one of these version bumps since the latest release in #5421)

src/gax/src/options.rs
```diff
@@ -262,6 +274,31 @@ pub mod internal {
     #[derive(Debug, Clone, Default, PartialEq)]
     pub struct ResourceName(pub String);
 
+    /// Per-request billing project. When present, `gax-internal`'s gRPC and
+    /// HTTP transports emit an `x-goog-user-project` header carrying this
+    /// value and drop any `x-goog-user-project` header the credentials
+    /// provider would have emitted from its configured `quota_project_id`,
+    /// so the wire carries exactly one `x-goog-user-project`.
+    #[derive(Debug, Clone, PartialEq)]
+    pub struct UserProject(http::HeaderValue);
```
dbolduc: I don't want to plumb this via the `extensions`. Can you just add a field + setters/getters similar to `user_agent`, `user_agent()`, `set_user_agent()`, `with_user_agent()` ?

src/gax/src/options.rs
```diff
@@ -262,6 +274,31 @@ pub mod internal {
...
+
+    impl UserProject {
+        /// Creates a new `UserProject` extension.
+        ///
+        /// # Panics
+        ///
+        /// Panics if the project ID contains non-visible ASCII characters.
```
dbolduc: This is valid, but not customary for us. I think we can avoid it if we treat it like a user agent. The error may be less helpful, but eh.

coryan:
> 
> The change is purely additive, so make the version `0.7.13`.
> 
> (The same would have been true for adding a public API to `gax` but note that we already did one of these version bumps since the latest release in #5421)

I would prefer if we split these changes to `gax` and `gax-internal` to separate PRs.  Purely because it will make it easier to grok the history of the changes, not because it would fix any builds or anything.

src/gax-internal/src/grpc.rs
```diff
@@ -516,17 +518,27 @@ impl Client {
...
             .headers(http::Extensions::new())
             .await
             .map_err(Error::authentication)?;
 
-        let CacheableResource::New { data, .. } = h else {
+        let CacheableResource::New { mut data, .. } = h else {
             unreachable!("headers are not cached");
         };
 
+        if let Some(up) = options.get_extension::<UserProject>() {
```
dbolduc: Meh, don't love that this is here. But it's fine.

If we keep this, can you add a comment like:

```
// The client-supplied user project must override the credential's user project. So we insert it after setting the credential's headers.
```

---

But I think I would prefer is to add this logic to `Self::make_headers()`

And then we just flip this to:

```rs
// Note that client headers override credential headers (e.g. for the `x-goog-user-project`).
data.extend(headers);
Ok(data)
```

src/gax-internal/src/http.rs
```diff
@@ -392,7 +395,12 @@ impl ReqwestClient {
 
         builder = match self.cred.headers(Extensions::new()).await {
             Err(e) => return Err(Error::authentication(e)),
-            Ok(CacheableResource::New { data, .. }) => builder.headers(data),
+            Ok(CacheableResource::New { mut data, .. }) => {
+                if let Some(up) = options.get_extension::<UserProject>() {
+                    data.insert(X_GOOG_USER_PROJECT, up.as_value().clone());
+                }
+                builder.headers(data)
```
dbolduc: The logic doesn't belong in the match statement. Add a block like the user agent block right after this.

dbolduc: Note that you need to use the `builder.headers()` method to `insert` the key and overwrite the previous entry. `builder.header()` is an `append` :shaking_face: 

---

optional: at this point, we might want to construct a `http::HeaderMap` and call `builder.headers()` once. 🤷

dbolduc (changes requested):
> Split this PR into two

This is typically a good idea. I would be grateful if you did this.

I did not look at the storage changes because I have enough feedback on the `gax` half.

src/gax-internal/src/http.rs
```diff
@@ -392,7 +395,12 @@ impl ReqwestClient {
...
+            }
             Ok(CacheableResource::NotModified) => unreachable!("headers are not cached"),
         };
         builder.build().map_err(map_send_error)
```
alvarowolfx, line 417: same as @dbolduc, I think we should add a method to the builder to allow overriding request level quota/user project. And maybe some similar code can be used in the gRPC part.

```rust
builder = quota_project(options)
            .into_iter()
            .fold(builder, |b, qp| b.header(X_GOOG_USER_PROJECT, qp));

fn quota_project(options: &crate::options::RequestOptions){
   options.get_extension::<UserProject>()
}

```

joshuatants:
Thank you all for the feedback. Will split the PR and address the comments. ETA  for the `gax` changes is early next week.

joshuatants:
@dbolduc @alvarowolfx I've split the PR into two and kept only the `gax` changes in this PR. 

`gcb-pr-semver-checks` is failing but I'm not sure how to fix it. By my understanding the default implementation the semver checker is complaining about is provided in `impl<T> RequestOptionsBuilder for T where T: internal::RequestBuilder`. Is that correct?

src/gax-internal/src/http.rs
joshuatants: Done

src/gax-internal/src/http.rs
joshuatants: Done

src/gax-internal/src/grpc.rs
joshuatants: Done

src/gax/src/options.rs
joshuatants: Done

src/gax/src/options.rs
joshuatants: Done

dbolduc:
@joshuatants  - Sorry, you're right. I offered a bad suggestion.

> `gcb-pr-semver-checks` is failing but I'm not sure how to fix it.

It is going to be a tough fix. Let me think about it for a bit and get back to you with a plan.

> By my understanding the default implementation the semver checker is complaining about is provided in `impl<T> RequestOptionsBuilder for T where T: internal::RequestBuilder`. Is that correct?

Yep. This was my mistake. At the time I thought I could make "breaking changes" such as adding a new trait function to public types marked as `internal`. Since then, we have learned that if a type is public, it is public. It doesn't matter if we hide a module from the documentation and tell external crates not to use it -- breaking the interface costs us a major version bump.

dbolduc:
> It is going to be a tough fix.

Actually, I found an easy fix, and wrote it up in go/cloud-rust:adding-request-options

Basically, I think we can provide a default implementation for any new `fns` we add to `RequestOptionsBuilder` and specialize them.

So we would say like:

```rs
pub trait RequestOptionsBuilder : internal::RequestBuilder {
    // <current fns omitted...>

    fn with_user_project<V: Into<String>>(self, _v: V) -> Self
    where Self: Sized { self }
}
```

instead of:

```rs
pub trait RequestOptionsBuilder : internal::RequestBuilder {
    // <current fns omitted...>

    fn with_user_project<V: Into<String>>(self, v: V) -> Self;
}
```

... but maybe hold off on making changes in this PR until that doc gets approved. My team often thinks of things that I missed.

joshuatants:
Thanks for the help, will wait for updates.

dbolduc:
> Thanks for the help, will wait for updates.

Doc approved. Proceed with the suggestion in: https://github.com/googleapis/google-cloud-rust/pull/5490#issuecomment-4315724228

joshuatants:
Updated. PTAL.

src/gax/src/options.rs
```diff
@@ -92,6 +93,22 @@ impl RequestOptions {
         &self.user_agent
     }
 
+    /// Sets the user project for the request.
+    ///
+    /// When present, `gax-internal`'s gRPC and HTTP transports emit an
+    /// `x-goog-user-project` header carrying this value and drop any
+    /// `x-goog-user-project` header the credentials provider would have
+    /// emitted from its configured `quota_project_id`, so the wire
+    /// carries exactly one `x-goog-user-project`.
```
dbolduc:
```suggestion
    /// Sets the [quota project] for the request.
    ///
    /// This adds the `x-goog-user-project` header to the request. Note that
    /// setting this option overrides a credentials' quota project.
    ///
    /// [quota project]: https://docs.cloud.google.com/docs/quotas/quota-project
```

I don't want to mention `gax-internal`, and I think the name for this thing is called a "quota project" even though the header is called "user project" 🫨

src/gax/src/options.rs
```diff
@@ -92,6 +93,22 @@ impl RequestOptions {
...
+    pub fn set_user_project<T: Into<String>>(&mut self, v: T) {
```
dbolduc: change `user_project` to `quota_project` here and elsewhere

src/gax/src/options.rs
```diff
@@ -189,6 +206,23 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
 
     /// Sets the polling backoff policy configuration.
     fn with_polling_backoff_policy<V: Into<PollingBackoffPolicyArg>>(self, v: V) -> Self;
+
+    // Methods with a default implementation.
+    // See https://github.com/googleapis/google-cloud-rust/pull/5490 for context.
```
dbolduc, line 210: Good thinking.

src/gax/src/options.rs
```diff
@@ -189,6 +206,23 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
...
+
+    /// Sets the user project for the request.
+    ///
+    /// When present, `gax-internal`'s gRPC and HTTP transports emit an
+    /// `x-goog-user-project` header carrying this value and drop any
+    /// `x-goog-user-project` header the credentials provider would have
+    /// emitted from its configured `quota_project_id`, so the wire
+    /// carries exactly one `x-goog-user-project`.
```
dbolduc: update these docs too.

dbolduc (changes requested):
Looks great! I just think we should call this thing a "quota_project", not a "user_project".

And I had a suggestion for the public docs.

src/gax/src/options.rs
joshuatants: 😂
Sure thing. Updated the docs.

src/gax/src/options.rs
joshuatants: Thank you

src/gax/src/options.rs
joshuatants: Done

src/gax/src/options.rs
joshuatants: Done

dbolduc (approved):

PR #5491, opened 2026-04-22
perf(spanner): add gRPC channel pooling

olavloite:
Applications that use Spanner for high throughput often need more than one gRPC channel in order to achieve the best possible performance. Having more than one channel both increases the number of concurrent streams that an application can have open at any time (the limit is 100 per gRPC channel), and distributes the load that is generated by the application across multiple Spanner frontends.

See also go/spanner-rust-channel-pool


src/spanner/src/client.rs
```diff
@@ -61,20 +62,19 @@ impl google_cloud_gax::client_builder::internal::ClientFactory for Factory {
     type Credentials = Credentials;
 
     async fn build(self, config: ClientConfig) -> crate::ClientBuilderResult<Self::Client> {
-        let transport =
-            crate::generated::gapic_dataplane::transport::Spanner::new(config.clone()).await?;
-        let grpc_client = transport.inner.clone();
+        let num_channels = std::env::var("SPANNER_NUM_CHANNELS")
```
olavloite, line 70: We will add a 'normal' configuration option for this as well in a follow-up PR (assuming that we go with this solution)

coryan: I am not sure if this helps, but consider:

https://github.com/googleapis/google-cloud-rust/blob/902a7cb0d77be8ebd1aac0bd049eae180fdcb051/src/gax/src/client_builder.rs#L509

olavloite: Discussed offline in [go/spanner-rust-channel-pool](http://goto.google.com/spanner-rust-channel-pool)

sakthivelmanii (approved):

PR #5492, opened 2026-04-22
feat(spanner): support QueryMode when executing statements

olavloite:
Spanner supports multiple QueryModes when executing a statement:
- PLAN: Only parse and plan the query. Parameter values may be omitted.
- NORMAL: Normal execution. Parameter values must be supplied.
- PROFILE: Execute and profile the query. Returns execution stats.


olavloite:
(The build error in gcb-pr-deny seems unrelated)

sakthivelmanii (approved):

PR #5498, opened 2026-04-22
impl(pubsub): use fewer exactly-once lease extensions

PhongChuong:
This PR extends the logic in https://github.com/googleapis/google-cloud-rust/pull/5241 to exactly-once leases. Specifically, we only extend exactly-once leases if they are within 5 seconds of expiring.

Towards #5048


dbolduc (approved):
Thanks!

PR #5499, opened 2026-04-22
chore(spanner): foundation for batch write transactions 

olavloite:
Adds the foundations for batch write transactions. The public API will be
added in a follow-up pull request.


olavloite:
Replaced by #5502

PR #5502, opened 2026-04-22
feat(spanner): add BatchWrite transaction

olavloite:
Adds support for BatchWrite transactions. BatchWrite can be used to write
large numbers of mutations to Spanner. A BatchWrite transaction is not
guaranteed to be atomic. However, each MutationGroup in a BatchWrite
transaction is guaranteed to be atomic.

This change adds a method that returns a stream of responses and delegates
all error handling to the application. A follow-up pull request will add a
managed execution method that handles errors automatically.

Depends on #5499


src/spanner/src/batch_write_transaction.rs
```diff
@@ -0,0 +1,207 @@
...
+    ///     }
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    ///
+    /// This method sends the mutation groups to Spanner and returns the responses as a stream.
+    /// Each response includes a status code that indicates whether the mutation groups that
+    /// it references were applied successfully.
+    /// The method does not handle any errors, including retryable errors like Aborted.
```
dbolduc, line 89: super nit: add a newline before this paragraph?

```suggestion
    ///
    /// The method does not handle any errors, including retryable errors like Aborted.
```

src/spanner/src/mutation.rs
```diff
@@ -13,9 +13,12 @@
 // limitations under the License.
 
 use crate::key::KeySet;
+use crate::model::batch_write_request::MutationGroup as ProtoMutationGroup;
```
dbolduc, line 16: why do we write our own type instead of exposing the generated type?

src/spanner/src/batch_write_transaction.rs
```diff
@@ -0,0 +1,207 @@
...
+pub struct BatchWriteResponseStream {
+    pub(crate) inner: BatchWriteStream,
+}
+
+impl BatchWriteResponseStream {
+    /// Fetches the next [BatchWriteResponse] from the stream.
+    ///
+    /// Returns `Some(Ok(BatchWriteResponse))` when a message is successfully received,
+    /// `None` when the stream concludes naturally, or `Some(Err(_))` on RPC errors.
+    pub async fn next_message(&mut self) -> Option<crate::Result<BatchWriteResponse>> {
```
dbolduc: naming nit: we tend to call this just `next()`

dbolduc: and add the `into_stream()` (eventually, not necessarily in this PR)

src/spanner/src/mutation.rs
```diff
@@ -266,6 +269,42 @@ impl ValueBinder {
     }
 }
 
+/// A group of mutations that are applied atomically in a [crate::BatchWriteTransaction].
+#[derive(Clone, Debug, PartialEq)]
+pub struct MutationGroup {
+    pub mutations: Vec<Mutation>,
+}
```
dbolduc, line 277: FYI: https://github.com/googleapis/google-cloud-rust/issues/5566

It is a breaking change to add any fields to this in the future. Are you sure it will not grow?

(if we are keeping the type)

dbolduc (commented):
Mostly curious about why we aren't using the type from the model.

src/spanner/src/mutation.rs
olavloite: Using the generated `MutationGroup` struct would also mean that customers would need to use the generated `Mutation` struct (or manually convert hand-written `Mutation` instances to the generated version before setting them on `MutationGroup`). And we certainly want to use a hand-written `Mutation` struct, as it allows us to encapsulate the way that Spanner expects the various data types to be encoded on the wire. A hand-written `Mutation` can be constructed like this:

```rust
let mutation = Mutation::new_insert_builder("Users")
    .set("UserId").to(&1)
    .set("UserName").to(&"Alice")
    .build();
```

Our `ToValue` trait then takes care or encoding these values the way that Spanner expects them.

src/spanner/src/batch_write_transaction.rs
olavloite: Done

src/spanner/src/batch_write_transaction.rs
olavloite: Renamed to `next()` and added `into_stream()`.

src/spanner/src/mutation.rs
olavloite: Added `#[non_exhaustive]`

src/spanner/src/mutation.rs
dbolduc: Ack

dbolduc (approved):

PR #5504, opened 2026-04-22
docs: Format resource names in RPC samples

amanda-tarafa:


amanda-tarafa:
Generated with https://github.com/googleapis/librarian/pull/5530 minus some `metadata.json` files that seem to have been generated by some pending changes in the generator.

coryan (commented):
I won't be able to review the code in sidekick, just a small observation on the output.

amanda-tarafa:
> I won't be able to review the code in sidekick, just a small observation on the output.

@coryan  Did you meant to leave a comment here about the output, or is this referring to the comments you left on the librarian PR (those are already addressed). I don't see any other comment here though. Thanks.

src/generated/api/apikeys/v2/src/client.rs
```diff
@@ -174,10 +175,10 @@ impl ApiKeys {
     /// use google_cloud_gax::paginator::ItemPaginator as _;
     /// use google_cloud_apikeys_v2::Result;
     /// async fn sample(
-    ///    client: &ApiKeys, parent: &str
+    ///    client: &ApiKeys, project_id: &str, location_id: &str
```
coryan, line 180: It would be nicer (I think) if the quickstart examples used the same format as these examples, that is, receive the project_id, location_id, etc. as arguments to a `sample()` function.

coryan:
> > I won't be able to review the code in sidekick, just a small observation on the output.
> 
> @coryan Did you meant to leave a comment here about the output, or is this referring to the comments you left on the librarian PR (those are already addressed). I don't see any other comment here though. Thanks.

Yes, but apparently I forgot how to use computers.  Fixed.

src/generated/api/apikeys/v2/src/client.rs
amanda-tarafa: See how this looks. A few notes:

- I had to show the sample method that we were hiding before but...
- I couldn't get the error part of the result to be any "prettier", I'm probably missing something.
- I can hide the sample method again for the quickstarts that would receive no parameters, but that adds complexity to the template, and will cover only cases of services that have no AIP Standard methods. And I'm planning on generating samples for custom methods next, so this seemed too much for too little.
- I left client initialization in the sample as opposed to receiving the client as a parameters like in the RPC ones because I think in the quickstart it makes sense to show how to initialize a client.
- I think we have to live with the trailing comma on the parameter list unless we want to make the annotation know about that last of the parameters. Or unless we want to receive the client as a parameter.

coryan: > See how this looks. A few notes:
> 
> * I had to show the sample method that we were hiding before but...

That sounds fine.

> * I couldn't get the error part of the result to be any "prettier", I'm probably missing something.

Cannot use `anyhow::Result<()>` ?

> * I can hide the sample method again for the quickstarts that would receive no parameters, but that adds complexity to the template, and will cover only cases of services that have no AIP Standard methods. And I'm planning on generating samples for custom methods next, so this seemed too much for too little.

Agree.

> * I left client initialization in the sample as opposed to receiving the client as a parameters like in the RPC ones because I think in the quickstart it makes sense to show how to initialize a client.

Agree.

> * I think we have to live with the trailing comma on the parameter list unless we want to make the annotation know about that last of the parameters. Or unless we want to receive the client as a parameter.

Trailing commas are fine, `rustfmt` adds them sometimes too, they should not bother anybody.

amanda-tarafa: > Cannot use anyhow::Result<()> ?

I tried this but I get "no external crate `anyhow`" that's on the `use anyhow;` What am I missing?

coryan: Ah, we are missing a stanza in the Cargo.toml file, something like:

```toml
[dev-dependencies]
anyhow.workspace = true
```

amanda-tarafa: Ah good, will add that as well.

src/pubsub/src/generated/gapic/client.rs
```diff
@@ -735,16 +741,18 @@ impl SubscriptionAdmin {
 /// ```
 /// # use google_cloud_pubsub::client::SchemaService;
 /// use google_cloud_gax::paginator::ItemPaginator as _;
-/// # async fn sample() -> Result<(), Box<dyn std::error::Error>> {
+/// async fn sample(
+///    parent: &str,
+/// ) -> anyhow::Result<()> {
```
coryan, line 746: For a future PR, I wonder why this parent (which I am sure is a project) is not captured.

coryan (approved):
For a future PR: the `PredictionService` in aiplatform is pretty important but also is super ad-hoc, so the examples are not as awesome as the other examples. Some `List*()` things are missing the parent format too.

src/pubsub/src/generated/gapic/client.rs
amanda-tarafa: Because it's a reference to a resource outside of this package, so we don't have the resource definition so we don't know the patterns. See [here](https://github.com/googleapis/googleapis/blob/05cca56fb2037c01576e2d2fdcf00e46eb86c777/google/pubsub/v1/schema.proto#L227).

Not sure if there's much we can do about it, except finding and loading this external resources. I think that's being done for mixins etc. already, but I'm not sure.

PR #5507, opened 2026-04-23
chore(spanner): mark unary RPCs by default idempotent

olavloite:
Mark all unary RPCs for Spanner as idempotent by default. An application can override this in the RequestOptions by setting an explicit value. If an application passes in a RequestOptions value without setting an explicit value for idempotency, then the client will also mark the RPC as idempotent.

Server-streaming RPCs are not marked as idempotent, because:
1. ExecuteStreamingSql and StreamingRead produce a ResultSet, which internally handles transient errors and retries of these. This internal handling is needed in order to supply the correct resume_token for the retry.
2. BatchWrite is not idempotent.

### DIFF
--- a/src/spanner/src/client.rs
+++ b/src/spanner/src/client.rs
@@ -159,7 +159,7 @@ impl Spanner {
         self.inner
             .create_session()
             .with_request(request)
-            .with_options(options)
+            .with_options(with_default_idempotency(options))
             .send()
             .await
     }
@@ -172,7 +172,7 @@ impl Spanner {
         self.inner
             .execute_sql()
             .with_request(request)
-            .with_options(options)
+            .with_options(with_default_idempotency(options))
             .send()
             .await
     }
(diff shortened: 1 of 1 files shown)


src/spanner/src/client.rs
```diff
@@ -824,6 +834,91 @@ mod tests {
...
     }
 
+    #[tokio::test]
+    async fn default_retry_respected() -> anyhow::Result<()> {
+        use crate::model::CreateSessionRequest;
+
+        // 1. Setup Mock Server
+        let mut mock = MockSpanner::new();
+        let mut seq = mockall::Sequence::new();
+        mock.expect_create_session()
```
dbolduc, line 763: comment: We are only testing default idempotency for `create_session` and assuming it is true for the other fns.

That is fine with me. Just calling it out.

dbolduc (approved):

src/spanner/src/client.rs
olavloite: Yeah, that is indeed a valid point, and also a bit in line with the other worry that we might forget this for new RPCs. But as all these RPC definitions use the exact same method body, we can move that to a macro. That way, it should not really be possible to forget this for any new RPC that might be added in the future. So I changed that in https://github.com/googleapis/google-cloud-rust/pull/5507/commits/43241ac315501592b360dcade2cda63770a552de.

PR #5511, opened 2026-04-23
impl(auth): Service account id token not supported on custom universes

alvarowolfx:
Towards #3646

### DIFF
--- a/src/auth/src/credentials/idtoken/service_account.rs
+++ b/src/auth/src/credentials/idtoken/service_account.rs
@@ -181,10 +181,10 @@ impl Builder {
             serde_json::from_value::<ServiceAccountKey>(self.service_account_key)
                 .map_err(BuilderError::parsing)?;
 
-        let universe_domain = service_account_key.universe_domain.clone();
+        let universe_domain = service_account_key.universe_domain.as_deref();
         if !crate::universe_domain::is_default_universe_domain(universe_domain) {
             return Err(BuilderError::not_supported(
-                "Service Account Credentials does not support getting an ID token in universes other than googleapis.com",
+                "Service Account Credentials do not support getting an ID token in universes other than googleapis.com",
             ));
         }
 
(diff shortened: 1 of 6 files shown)


src/auth/src/credentials/idtoken/service_account.rs
```diff
@@ -180,6 +180,14 @@ impl Builder {
         let service_account_key =
             serde_json::from_value::<ServiceAccountKey>(self.service_account_key)
                 .map_err(BuilderError::parsing)?;
+
+        let universe_domain = service_account_key.universe_domain.clone();
+        if !crate::universe_domain::is_default_universe_domain(universe_domain) {
```
coryan: Can we do this without cloning? Maybe a `.as_ref()?

src/auth/src/credentials/idtoken/service_account.rs
```diff
@@ -180,6 +180,14 @@ impl Builder {
...
+            return Err(BuilderError::not_supported(
+                "Service Account Credentials does not support getting an ID token in universes other than googleapis.com",
```
coryan:
```suggestion
                "Service Account Credentials do not support getting an ID token in universes other than googleapis.com",
```

But run this through a native speaking.  Maybe something like "Google Cloud only supports ID tokens and Service Account Credentials on the googleapis.com universe"....

Also, if I get this error message I may not know what a "universe" is...

coryan (changes requested):

src/auth/src/credentials/idtoken/service_account.rs
alvarowolfx: good call, I changed the `is_default_universe_domain` helper to accept an `&str`.

src/auth/src/credentials/idtoken/service_account.rs
alvarowolfx: I applied the verb correction. This exception message is the recommended one in the design docs. "Google Cloud only supports..." might be too broad, as the limitation is specific to this token flow, rather than all of Google Cloud.

About not knowing what an "universe" is: It is the official terminology for this feature (Google Default Universe vs. others), and it appears as `universe_domain` in the JSON key file. If a user is using a non-default universe, they should know what it is.

coryan (approved):

PR #5512, opened 2026-04-23
impl(auth): override impersonation url with universe domain on source credential flow

alvarowolfx:
Towards #3646


src/auth/src/credentials/idtoken/impersonated.rs
```diff
@@ -169,7 +169,7 @@ impl Builder {
             include_email: None,
             target_audience: target_audience.into(),
             service_account_impersonation_url: Some(format!(
-                "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/{}:generateIdToken",
+                "https://iamcredentials.{{universe_domain}}/v1/projects/-/serviceAccounts/{}:generateIdToken",
```
dbolduc: Inventing our own format string language looks scary to me.

Why not hold the `target_principal` as a field and build the string later? We are rebuilding the whole string anyway for each call to `token()`, so it's not like there are performance benefits to holding the full thing here.

src/auth/src/credentials/idtoken/impersonated.rs
dbolduc, line 161: comment: Ah, we cannot grab the source's universe domain here because this is a sync fn.

src/auth/src/credentials/idtoken/impersonated.rs
```diff
@@ -439,12 +439,26 @@ impl TokenProvider for ImpersonatedTokenProvider {
...
+            .contains("{universe_domain}")
+        {
+            let universe_domain = crate::universe_domain::resolve(&self.source_credentials).await;
+            replaced_url = self
+                .service_account_impersonation_url
+                .replace("{universe_domain}", &universe_domain);
+            replaced_url.as_str()
+        } else {
+            self.service_account_impersonation_url.as_str()
+        };
```
dbolduc: comment: Meh, I don't like rebuilding the URL on every call to token. The URL doesn't change right?

But meh, this thing takes a `&self`, so we cannot really mutate self (e.g. if we tried to cache the universe domain / URL).

So this is probably best. Can you add a comment saying why we are doing this approach?

- Fetching the universe domain is async, so we must do it in the calls to `token()`, not in the builder.
- Achieving inner-mutability with a lock is worse than just building the URL string for each request.

src/auth/src/credentials/idtoken/impersonated.rs
```diff
@@ -439,12 +439,26 @@ impl TokenProvider for ImpersonatedTokenProvider {
             }
         };
 
+        let replaced_url;
```
dbolduc: Very weird. Use `let replaced_url = ... ` later instead?

dbolduc: Well, this code is probably vanishing.

src/auth/src/credentials/idtoken/impersonated.rs
dbolduc: We can use some type like:

```rs
enum ImpersonationUrl {
  // Only the target principal is known. We need to build the URL using the source credentials' universe domain.
  TargetPrincipal(String),
  // We were given the exact URL to use.
  Exact(String),
}

impl ImpersonationUrl {
  // Builds the URL for requests.
  async fn url(&self, source_creds: &Credentials) -> String {
    match self {
      TargetPrincipal(ref tp) => todo!("build the url. omitted for brevity"),
      Exact(ref url) => url.clone(),
    }
  }
}
```

dbolduc (changes requested):
I would feel much safer with an enum representing state than a format string, if that is possible.

src/auth/src/credentials/idtoken/impersonated.rs
alvarowolfx: I've tested holding `target_principal`, but that would also required to add a set of extra functions/internal things to allow to override the URL for tests. But I liked the enum idea, let me try that here.

src/auth/src/credentials/idtoken/impersonated.rs
alvarowolfx: this was to avoid a `.clone` of `self.service_account_impersonation_url`

src/auth/src/credentials/idtoken/impersonated.rs
alvarowolfx: universe domain is cached on all providers, so is kind of safe and no costly to call from the source credentials. But will add more comments (and try the enum approach)

dbolduc: Right. I was more thinking about caching the full URL string

> I don't like rebuilding the URL on every call to token. 

Ignore me. It is probably the same performance to clone the full string vs. build it.

src/auth/src/credentials/idtoken/impersonated.rs
dbolduc: Oh, it is a lifetime thing. I see it now. Thanks.

I don't think my enum saved that clone. :(

alvarowolfx:
Split `ImpersonationUrl` thing into https://github.com/googleapis/google-cloud-rust/pull/5558

src/auth/src/credentials/impersonated.rs
```diff
@@ -188,31 +188,39 @@ impl ImpersonationUrl {
...
+        creds: &Credentials,
+        principal: &str,
+        method: &str,
+    ) -> String {
+        let universe_domain = crate::universe_domain::resolve(creds).await;
         let endpoint = match &self.endpoint {
-            Some(endpoint) => endpoint,
-            None => "https://iamcredentials.googleapis.com",
+            Some(endpoint) => endpoint.to_string(),
+            None => format!("https://iamcredentials.{}", universe_domain),
```
dbolduc, line 223: nit: eventually clippy will nag us on this:

```suggestion
            None => format!("https://iamcredentials.{universe_domain}"),
```

dbolduc (approved):
Thanks for breaking this up. Made it easy to review.

PR #5515, opened 2026-04-24
feat(spanner): support timeout, retry- and backoff policy per statement

olavloite:
Adds support for setting a timeout, retry policy, and backoff policy per statement and read. This allows an application to override the defaults for these options on a per statement-basis.

Note: This change adds the API for all types of statements. However, the retry and backoff policies are not respected for the streaming RPCs ExecuteStreamingSql and StreamingRead yet. This will be added in a follow-up pull request.

### DIFF
--- a/src/spanner/src/client.rs
+++ b/src/spanner/src/client.rs
@@ -253,14 +253,21 @@ mod tests {
     use gaxi::grpc::tonic::{Code as GrpcCode, Response, Status};
     use google_cloud_auth::credentials::anonymous::Builder as Anonymous;
     use google_cloud_gax::error::rpc::Code;
+    use google_cloud_test_macros::tokio_test_no_panics;
     use spanner_grpc_mock::google::rpc as mock_rpc;
     use spanner_grpc_mock::google::spanner::v1 as mock_v1;
     use spanner_grpc_mock::google::spanner::v1::Session;
     use spanner_grpc_mock::{MockSpanner, start};
     use static_assertions::{assert_impl_all, assert_not_impl_any};
-    use std::sync::Arc;
-    use std::sync::atomic::{AtomicUsize, Ordering};
     use std::time::Duration;
+
+    mockall::mock! {
+        #[derive(Debug)]
+        BackoffPolicy {}
+        impl google_cloud_gax::backoff_policy::BackoffPolicy for BackoffPolicy {
+            fn on_failure(&self, state: &google_cloud_gax::retry_state::RetryState) -> std::time::Duration;
+        }
+    }
 
     #[test]
     fn auto_traits() {
@@ -841,7 +848,7 @@ mod tests {
         Ok(())
     }
 
-    #[tokio::test]
+    #[tokio_test_no_panics]
     async fn timeout_respected() -> anyhow::Result<()> {
         use crate::batch_dml::BatchDml;
         use std::time::Duration;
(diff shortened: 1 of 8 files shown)


src/spanner/src/client.rs
```diff
@@ -838,6 +841,264 @@ mod tests {
...
+
+                Ok(())
+            })
+            .await?;
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn retry_policy_respected() -> anyhow::Result<()> {
```
dbolduc, line 997: We can mock the policies with `mockall`:

https://github.com/googleapis/google-cloud-rust/blob/ca4ea8a22835303c522ab3cf68142ece9dfb59fa/src/pubsub/src/subscriber/stream.rs#L167-L173

https://github.com/googleapis/google-cloud-rust/blob/ca4ea8a22835303c522ab3cf68142ece9dfb59fa/src/pubsub/src/subscriber/stream.rs#L364-L369

dbolduc: I don't like doing the same thing N different ways... but maybe I should stop caring in the age of the LLM.

dbolduc (approved):
suggested changes are optional.

src/spanner/src/client.rs
```diff
@@ -838,6 +841,264 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
```
dbolduc: Actually, I just remembered, this should use `#[tokio_test_no_panics]`

https://github.com/googleapis/google-cloud-rust/blob/ca4ea8a22835303c522ab3cf68142ece9dfb59fa/src/pubsub/src/subscriber/message_stream.rs#L531

https://github.com/googleapis/google-cloud-rust/blob/ca4ea8a22835303c522ab3cf68142ece9dfb59fa/src/pubsub/src/subscriber/message_stream.rs#L468

src/spanner/src/client.rs
```diff
@@ -838,6 +841,264 @@ mod tests {
...
+            }))
+        });
+
+        mock.expect_execute_streaming_sql().once().returning(|req| {
+            let metadata = req.metadata();
+            let timeout = metadata.get("grpc-timeout");
+            assert!(
+                timeout.is_some(),
+                "grpc-timeout header should be present for query"
+            );
```
dbolduc, line 879: Basically, if this assertion fails, the background task running the server panics, but that does not fail the test.

We have a special build that enables an unstable tokio feature to actually fail the test on a background panic.

Details in: https://github.com/googleapis/google-cloud-rust/issues/4733

dbolduc (commented):
Sorry, I missed that we should be enforcing stricter assertions on the fake server.

I imagine there are N such test cases we should update. So no reason to block this PR on it. I'll open an issue for you.

src/spanner/src/client.rs
olavloite: Thanks. This shows that 'never trust a test that has never failed' is really true... I've added `tokio_test_no_panics` to this test, and I'll go through the other tests in a separate pull request.

src/spanner/src/client.rs
olavloite: No, this is indeed better (not only for consistency, but also due to the fact that it gives more verification options). So I changed it to use `mockall`.

PR #5517, opened 2026-04-24
feat(spanner): support timeout, retry- and backoff policy for partitions

olavloite:
Adds support for setting an attempt timeout and a retry- and backoff policy for
a partition in a BatchReadOnlyTransaction. These settings are not serialized,
which means that any host that wants to apply any of these settings to its
execution of a partition must set these explicitly.

Depends on #5516


src/spanner/src/result_set.rs
```diff
@@ -936,6 +1016,99 @@ pub(crate) mod tests {
...
+    async fn test_result_set_custom_retry_policy() -> anyhow::Result<()> {
+        use gaxi::grpc::tonic::Response;
+        use gaxi::grpc::tonic::Status;
+
+        use spanner_grpc_mock::MockSpanner;
+        use spanner_grpc_mock::start;
+
+        use google_cloud_gax::retry_policy::Aip194Strict;
+        use google_cloud_gax::retry_policy::RetryPolicyExt;
+
```
dbolduc: nit: define these once for the whole `mod tests` ?

dbolduc (approved):

src/spanner/src/result_set.rs

PR #5519, opened 2026-04-24
feat(spanner): support read lock_hint and order_by

olavloite:
Adds two read options that were missed during the initial implementation.


src/spanner/src/read.rs
```diff
@@ -163,6 +170,57 @@ impl ConfiguredReadRequestBuilder {
...
+    /// ```
+    ///
+    /// By default, Spanner returns result rows in primary key order (or index key
+    /// order if reading via an index) except for `PartitionRead` requests.
+    ///
+    /// For applications that don't require rows to be returned in primary key
+    /// (`ORDER_BY_PRIMARY_KEY`) order, setting `ORDER_BY_NO_ORDER` option allows
+    /// Spanner to optimize row retrieval, resulting in lower latencies in certain
+    /// cases (for example, bulk point lookups).
+    pub fn with_order_by(mut self, order_by: OrderBy) -> Self {
```
dbolduc, line 198: super nit: I think we use `set_foo` for proto fields and `with_foo` for client configuration (like retries / endpoint / whatever)

This is an almost meaningless distinction, but it is what we have been doing 🤷

src/spanner/src/read.rs
```diff
@@ -163,6 +170,57 @@ impl ConfiguredReadRequestBuilder {
...
+    ///
+    /// Request exclusive locks judiciously because they block others from reading that
+    /// data for the entire transaction, rather than just when the writes are being performed.
+    pub fn with_lock_hint(mut self, lock_hint: LockHint) -> Self {
+        self.lock_hint = Some(lock_hint);
+        self
+    }
+
     /// Builds the configured `ReadRequest`.
     pub fn build(self) -> ReadRequest {
```
dbolduc, line 250: (I probably missed context, but) Can we use this generated builder?

https://github.com/olavloite/google-cloud-rust/blob/4a032d4e62db247a1edfb5994bde192482e6e411/src/spanner/src/generated/gapic_dataplane/builder.rs#L684-L880

Otherwise, we are asking for maintenance work.

dbolduc (approved):

src/spanner/src/read.rs
coryan: > super nit: I think we use `set_foo` for proto fields and `with_foo` for client configuration (like retries / endpoint / whatever)
> 
> This is an almost meaningless distinction, but it is what we have been doing 🤷

It is more intentional than you may think: separating the namespace avoids conflicts when the request protos get new fields.

dbolduc: Ah, smart.

src/spanner/src/read.rs
olavloite: If you mean as the builder that customer applications use: Then no. The generated `ReadRequest` builder contains a couple of fields that we do __not__ want an application to set directly, as they are managed by the client library. These are `session` and `transaction`.

We could (probably) use a generated `ReadRequestBuilder` as the backing data structure of our hand-written `ReadRequestBuilder`, instead of using separate fields for the fields that we want to support. I'm not sure it would simplify things much, though, especially as we use some other structs for `Read` that are also hand-written wrappers (e.g. `KeySet`).

src/spanner/src/read.rs
olavloite: > It is more intentional than you may think: separating the namespace avoids conflicts when the request protos get new fields.

But that would only create (hard) problems for the generated builders, right. This struct is hand-written, meaning that in case of name conflict, we can work around it.

I'm not necessarily against changing this to `set_order_by`, but I think that we then should do that throughout the entire client. E.g. this builder already has a method `with_limit` (which sets a proto field), and there are several other fields that already have `with_foo` setters.

olavloite: Created #5579 for this

PR #5521, opened 2026-04-24
feat(spanner): add options for RPC priority

olavloite:
Adds support for setting the RPC priority for SQL statements, reads, and the CommitRequest of a transaction.


src/spanner/src/read.rs
```diff
@@ -143,6 +144,24 @@ impl ConfiguredReadRequestBuilder {
...
+    /// ```
+    /// # use google_cloud_spanner::client::{ReadRequest, KeySet};
+    /// # use google_cloud_spanner::model::request_options::Priority;
+    /// let request = ReadRequest::builder("Users", vec!["Id"])
+    ///     .with_keys(KeySet::all())
+    ///     .with_priority(Priority::Low)
+    ///     .build();
+    /// ```
+    pub fn with_priority(mut self, priority: Priority) -> Self {
+        self.request_options
```
dbolduc, line 173: why is this field an `Option<>`? We could just initialize it to the default and set things directly.

src/spanner/src/read.rs
```diff
@@ -143,6 +144,24 @@ impl ConfiguredReadRequestBuilder {
...
+    /// # Example
+    /// ```
+    /// # use google_cloud_spanner::client::{ReadRequest, KeySet};
+    /// # use google_cloud_spanner::model::request_options::Priority;
+    /// let request = ReadRequest::builder("Users", vec!["Id"])
+    ///     .with_keys(KeySet::all())
+    ///     .with_priority(Priority::Low)
+    ///     .build();
+    /// ```
+    pub fn with_priority(mut self, priority: Priority) -> Self {
```
dbolduc, line 172: same nit from the other PR: `set_`, not `with_`.

And same question on if we can use a generated builder, to minimize maintenance.

dbolduc (approved):

src/spanner/src/read.rs
olavloite: We try to be pedantic and only include optional nested protos like this only if any of its fields have actually been set. And while one such optional field does not make much difference, it adds up in the end. `ExecuteSqlRequest` currently has 4 such optional fields that could be initialized with an empty default struct value.

src/spanner/src/read.rs
olavloite: Created #5579 for the question regarding `set_` vs `with_`.

Using a `ReadRequest` or its builder directly is not something that we want, due to a few fields that are fully managed by the client (`session` and `transaction`).

PR #5525, opened 2026-04-24
impl(pubsub): add update_last_extension fn to lease state

PhongChuong:
Add helper functions {update_last_extension, update_last_extension_eo} to LeaseState to update ack ids that were successfully extended. These functions are currently unused. The followup PR will plumb successful extends using LeaseEvent.

This PR also split pending_extends into pending_extends and eo_pending_extends to differentiate successful extends for at least once and exactly once leases.

Towards: https://github.com/googleapis/google-cloud-rust/issues/5048

### DIFF
--- a/src/pubsub/src/subscriber/lease_state/at_least_once.rs
+++ b/src/pubsub/src/subscriber/lease_state/at_least_once.rs
@@ -68,6 +68,16 @@ impl Leases {
         )
     }
 
+    /// Updates the `last_extension` timestamp for the given ack IDs with the
+    /// completion time of a successful extension RPC.
+    pub fn update_last_extension(&mut self, ack_ids: &[String], time: Instant) {
+        for id in ack_ids {
+            if let Some(info) = self.under_lease.get_mut(id) {
+                info.last_extension = Some(time);
+            }
+        }
+    }
+
     /// Returns batches of ack IDs to extend.
     ///
     /// Drops messages whose lease deadline cannot be extended any further.
@@ -100,6 +110,8 @@ impl Leases {
                     // Flush the batch when it is full.
                     batches.push(std::mem::take(&mut batch));
                 }
+                // TODO(#5048): Do not update last_extension here after update_last_extension fn
+                // is used to report successful extends.
                 info.last_extension = Some(now);
                 true
             }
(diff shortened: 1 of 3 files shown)


src/pubsub/src/subscriber/lease_state/at_least_once.rs
```diff
@@ -151,6 +163,59 @@ mod tests {
     // Cover the constant, converting it to an integer for convenience.
     const MAX_IDS_PER_RPC: i32 = super::MAX_IDS_PER_RPC as i32;
 
+    #[test]
```
dbolduc: super nit: why is this the first test in the fixture?

I'd probably put it after the `drain` test to match the order fns are defined on the struct.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
```diff
@@ -151,6 +163,59 @@ mod tests {
...
+    fn update_last_extension() {
+        let mut leases = Leases::default();
+        let now = Instant::now();
+
+        leases.add(test_id(1), AtLeastOnceInfo::new());
+        leases.add(test_id(2), AtLeastOnceInfo::new());
+
+        assert_eq!(
+            leases.under_lease.get(&test_id(1)).unwrap().last_extension,
```
dbolduc: nit: for readability, consider defining a helper in the test module:

```rs
impl Leases {
  fn last_extension(&self, id: &str) -> Option<Instant> {
    self.under_lease.get(id).expect("we only look up ack IDs under lease").last_extension
  }
}
```

And then we can say like:

```rs
        assert_eq!(leases.last_extension(&test_id(1)), None);
```

src/pubsub/src/subscriber/lease_state/at_least_once.rs
```diff
@@ -68,6 +68,16 @@ impl Leases {
         )
     }
 
+    /// Updates the `last_extension` timestamp for the given ack IDs with the
+    /// completion time of a successful extension RPC.
+    pub fn update_last_extension(&mut self, ack_ids: &[String], time: Instant) {
+        for id in ack_ids {
+            if let Some(info) = self.under_lease.get_mut(id) {
+                info.last_extension = Some(time);
```
dbolduc: optional nit: if we only ever pass `Instant::now()`, it might be simpler to omit the `time` parameter. Note that `Instant::now()` is already fake-able in unit tests.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
dbolduc: here and below, and in the `exactly_once::Leases` tests

dbolduc (approved):

src/pubsub/src/subscriber/lease_state/at_least_once.rs
PhongChuong: Moved.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
PhongChuong: Thanks. It's easier to read now.

src/pubsub/src/subscriber/lease_state/at_least_once.rs
PhongChuong: Removed the time parameter.

PR #5529, opened 2026-04-24
ci: move advisory check to github action

suzmue:
Advisories are usually caused by external every not changes in our code. Checking for these on pre-submits increases developer friction. This change moves that check to be a scheduled job that files an issue which maintainers can then address.


coryan (approved):
Keep in mind that GHA are not free if we go over some limits.  You need to monitor this and switch to GCB if it looks like we will get close to those limits.

PR #5530, opened 2026-04-24
impl(pubsub): update last extension for successful extend

### LINKED ISSUE #5048: Improve subscriber lease extension logic

PhongChuong:
Lease state propagate extension results back to LeaseLoop using by the LeaseEvent ExtendCompleted and ExtendCompletedEO. The result are used to update the last extension value in the lease state.

Fixes: https://github.com/googleapis/google-cloud-rust/issues/5048

### DIFF
--- a/src/pubsub/src/subscriber/lease_state.rs
+++ b/src/pubsub/src/subscriber/lease_state.rs
@@ -188,6 +188,10 @@ pub(super) enum LeaseEvent {
     Flush,
     /// Extend leases
     Extend,
+    /// Pending extensions completed
+    ExtendCompleted(Vec<String>),
+    /// Pending exactly-once extensions completed
+    ExtendCompletedEO(Vec<String>),
 }
 
 impl<L> LeaseState<L>
@@ -228,6 +232,22 @@ where
         tokio::select! {
             _ = self.flush_interval.tick() => LeaseEvent::Flush,
             _ = self.extend_interval.tick() => LeaseEvent::Extend,
+            res = self.pending_extends.join_next(), if !self.pending_extends.is_empty() => {
+                if let Some(Ok(ack_ids)) = res {
+                    LeaseEvent::ExtendCompleted(ack_ids)
+                } else {
+                    // swallow the JoinError.
+                    LeaseEvent::ExtendCompleted(Vec::new())
+                }
+            }
+            res = self.eo_pending_extends.join_next(), if !self.eo_pending_extends.is_empty() => {
+                if let Some(Ok(ack_ids)) = res {
+                    LeaseEvent::ExtendCompletedEO(ack_ids)
+                } else {
+                    // swallow the JoinError.
+                    LeaseEvent::ExtendCompletedEO(Vec::new())
+                }
+            }
         }
     }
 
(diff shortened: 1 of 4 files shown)


src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -228,6 +232,22 @@ where
         tokio::select! {
             _ = self.flush_interval.tick() => LeaseEvent::Flush,
             _ = self.extend_interval.tick() => LeaseEvent::Extend,
+            res = self.pending_extends.join_next(), if !self.pending_extends.is_empty() => {
+                if let Some(Ok(ack_ids)) = res {
+                    LeaseEvent::ExtendCompleted(ack_ids)
+                } else {
+                    // swallow the JoinError.
+                    LeaseEvent::ExtendCompleted(Vec::new())
+                }
```
dbolduc: Alternatively, put the `select!` in a `loop` and `continue` here.

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -305,6 +323,10 @@ where
     ///
     /// Drops messages whose lease deadline cannot be extended any further.
     pub(super) fn extend(&mut self) {
+        // Old pending extensions and their results are no longer needed.
+        self.pending_extends = JoinSet::new();
+        self.eo_pending_extends = JoinSet::new();
```
dbolduc: comment: Ah, this took me a second. If we didn't hear back from the last batch by this point, we are going to try to extend everything in the batch again now.

dbolduc: While it seems unlikely that an RPC gets lost for 3 seconds, I don't see the point of discarding the results.

WDYT about removing this?

src/pubsub/src/subscriber/lease_state.rs
```diff
@@ -488,6 +507,28 @@ pub(super) mod tests {
         assert_eq!(flattened.ack_ids, test_ids(2..3));
     }
 
+    #[tokio::test(start_paused = true)]
+    async fn extend_clears_pending() {
```
dbolduc: comment: I know this gives us coverage, but the behavior isn't mandatory.

dbolduc (approved):

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: Removed.

src/pubsub/src/subscriber/lease_state.rs
PhongChuong: Done.

PR #5536, opened 2026-04-24
cleanup: enable missing_docs by default

suzmue:
Alternatively, we document these things, but that can be done as follow up work. Cargo.toml does not allow inheriting some lints from the workspace and overriding others, so we have to put the annotations in the code.

For #5459


suzmue:
May come back to this but closing for now.

PR #5538, opened 2026-04-24
docs(pubsub): improve shutdown docs

suzmue:


src/pubsub/src/subscriber/shutdown_behavior.rs
```diff
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-/// The behavior on shutdown.
+/// The behavior on shutdown for a stream.
```
coryan, line 15: Still sounds a bit odd to me. What stream are we talking about?

coryan (approved):

src/pubsub/src/subscriber/shutdown_behavior.rs
dbolduc: A stream of messages

dbolduc (approved):

PR #5555, opened 2026-04-28
refactor(pubsub): make exactly once retry loop reusable

PhongChuong:
Move all of the exactly once retry logic such as process_attempt_error, extract_failures and the main retry loop in confirmed_ack into its own file to be reuse by confirmed_nack and exactly once extend.


PhongChuong:
This ended up larger than I expected. Going to break it down into smaller PRs.

PR #5558, opened 2026-04-28
refactor(auth): add ImpersonationUrl struct to handle logic to build url

alvarowolfx:


src/auth/src/credentials/impersonated.rs
```diff
@@ -160,6 +160,83 @@ pub struct Builder {
     is_access_boundary_enabled: bool,
 }
 
+#[derive(Debug, Clone)]
+pub(crate) struct ImpersonationUrl {
+    // optional endpoint override for testing
+    pub(crate) endpoint: Option<String>,
```
dbolduc, line 166: comment: ah, this feels sad, but I get why it's there.

There is a way to tell reqwest to [resolve](https://docs.rs/reqwest/latest/reqwest/struct.ClientBuilder.html#method.resolve) endpoints differently. As in like: `.resolve("iamcredentials.googleapis.com", server.endpoint())`

... but I have never tried it, and it is just another way to inject a custom endpoint.

If we had to do this sort of thing N times, maybe we would look into that. But for now, what you have is perfectly fine.

src/auth/src/credentials/impersonated.rs
```diff
@@ -2515,4 +2620,25 @@ mod tests {
 
         Ok(())
     }
+
+    #[test_case(None, "https://iamcredentials.googleapis.com", "https://iamcredentials.googleapis.com" ; "no override")]
+    #[test_case(Some("https://custom.endpoint".to_string()), "https://iamcredentials.googleapis.com", "https://custom.endpointgoogleapis.com" ; "with override")]
+    #[test_case(Some("https://custom.endpoint".to_string()), "https://iamcredentials.googleapis.com/v1/foo", "https://custom.endpointgoogleapis.com/v1/foo" ; "with override and path")]
```
dbolduc: these look funky to me.  `custom.endpointgoogleapis.com` ?

alvarowolfx: fixed, was a mistake from simplifying it from the version that handles `universe_domain` too. It should remove `iamcredentials.googleapis.com` on the `replace_url` func

src/auth/src/credentials/impersonated.rs
```diff
@@ -2515,4 +2611,14 @@ mod tests {
 
         Ok(())
     }
+
+    #[test_case(ImpersonationUrlKind::TargetPrincipal("test@example.com".to_string()), "test@example.com" ; "target principal")]
+    #[test_case(ImpersonationUrlKind::Exact("https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/test@example.com:generateAccessToken".to_string()), "test@example.com" ; "exact url")]
+    fn impersonation_url_client_email(kind: ImpersonationUrlKind, expected: &str) {
```
dbolduc, line 2617: Can we have simple unit tests like this for `access_token_url` and `id_token_url`?

dbolduc (approved):

src/auth/src/credentials/impersonated.rs
alvarowolfx: added. I was going to do it on the other PR. But we can already some here.

PR #5560, opened 2026-04-29
refactor(pubsub): make exactly once retry loop logic a function

PhongChuong:
This allow the retry loop to be reused for confirmed_nack and exactly once extend.


dbolduc (approved):
Nice, thanks for adding the test coverage.

PR #5563, opened 2026-04-29
feat(pubsub): retry confirmed_nack partial failures in Subscriber

PhongChuong:

### DIFF
--- a/src/pubsub/src/subscriber/leaser.rs
+++ b/src/pubsub/src/subscriber/leaser.rs
@@ -262,6 +262,14 @@ pub(super) mod tests {
         }
     }
 
+    mockall::mock! {
+        #[derive(Debug)]
+        pub BackoffPolicy {}
+        impl BackoffPolicy for BackoffPolicy {
+            fn on_failure(&self, state: &google_cloud_gax::retry_state::RetryState) -> std::time::Duration;
+        }
+    }
+
     #[async_trait::async_trait]
     impl Leaser for Arc<MockLeaser> {
         async fn ack(&self, ack_ids: Vec<String>) {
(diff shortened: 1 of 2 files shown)


src/pubsub/src/subscriber/leaser.rs
```diff
@@ -462,4 +463,46 @@ pub(super) mod tests {
         }
         Ok(())
     }
+
+    #[tokio::test]
+    async fn confirmed_nack() -> anyhow::Result<()> {
```
dbolduc, line 542: ah, this was missing. thanks for adding.

src/pubsub/src/subscriber/leaser.rs
```diff
@@ -462,4 +463,46 @@ pub(super) mod tests {
...
+        let mut mock = MockStub::new();
+        mock.expect_modify_ack_deadline()
+            .times(1)
```
dbolduc, line 545: Question, and the answer might be "nah": do we want to test that the retry loop is called? (here, and for confirmed_ack)

We wouldn't want to use the default policies or `verify_policies()`. So we'd probably have to make a `DefaultLeaser` then edit its options field or something.

dbolduc: I'm fine with not. Just pointing out that this test would pass without the code change to add the retry loop.

dbolduc (approved):

src/pubsub/src/subscriber/leaser.rs
PhongChuong: Added the tests.

PhongChuong:
@dbolduc , thanks for the review. I've also added a check in the retry loop logic. PTAL.

dbolduc (approved):

PR #5578, opened 2026-05-04
feat(storage): add quota_project option

joshuatants:
Adds a `with_quota_project` method to `Storage` and `StorageControl`
request builders to support Requester Pays buckets.

If set, this takes precedence over any `quota_project_id` configured in the client credentials.

For https://github.com/googleapis/google-cloud-rust/issues/3189 and https://github.com/googleapis/google-cloud-rust/issues/3306.

### DIFF
--- a/src/storage/src/builder_ext.rs
+++ b/src/storage/src/builder_ext.rs
@@ -15,6 +15,8 @@
 //! Extends [builder][crate::builder] with types that improve type safety and/or
 //! ergonomics.
 
+use google_cloud_gax::options::internal::{RequestOptionsExt, UserProject};
+
 /// An extension trait for `RewriteObject` to provide a convenient way
 /// to poll a rewrite operation until it is complete.
 #[async_trait::async_trait]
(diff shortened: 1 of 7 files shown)


src/storage/src/builder_ext.rs
```diff
@@ -71,13 +71,52 @@ impl RewriteObjectExt for crate::builder::storage_control::RewriteObject {
...
+/// [Requester Pays]: https://cloud.google.com/storage/docs/requester-pays
+/// [StorageControl]: crate::client::StorageControl
+pub trait QuotaProjectExt: Sized {
+    /// Sets the project that will be billed for this request.
+    fn with_quota_project(self, project: impl Into<String>) -> Self;
+}
+
+impl<T> QuotaProjectExt for T
+where
+    T: google_cloud_gax::options::internal::RequestBuilder,
```
dbolduc: Didn't you add this function to the request builder in the previous PR? Why do we need this trait?

joshuatants: `with_quota_project` was indeed added to `RequestOptionsBuilder` in `gax/src/options.rs`. However, I thought that in order to add this method to the auto-generated `StorageControl` builders, we need to do something more, hence I added this trait.

Is this the right way to add `with_quota_project` to the `StorageControl` builders?

dbolduc: You should remove the trait. If that doesn't work, we did something wrong

joshuatants: Done. Verified that it works correctly. Thank you for pointing this out.

dbolduc (approved):

Issue #5584, opened 2026-05-04
auth: Add support for Google Distributed Cloud service identity authentication


alvarowolfx:
This issue tracks the work required to implement Google Distributed Cloud (GDC) service identity authentication support in the `google-cloud-auth` crate. This will allow the SDK to function in private or disconnected cloud environments by supporting the `gdch_service_account` credential type, utilizing self-signed JWTs (ES256) exchanged via a Secure Token Service (STS) endpoint.

https://cloud.google.com/distributed-cloud

PR #5586, opened 2026-05-04
impl(auth): extend STS handler for gdch use case

alvarowolfx:
Towards #5584

### DIFF
--- a/src/auth/src/credentials/internal/sts_exchange.rs
+++ b/src/auth/src/credentials/internal/sts_exchange.rs
@@ -26,15 +26,26 @@ type Result<T> = std::result::Result<T, CredentialsError>;
 /// Reference: https://datatracker.ietf.org/doc/html/rfc8693
 #[derive(Debug, Default)]
 pub struct STSHandler {
-    use_json: bool,
+    body_encoding: BodyEncoding,
     ca_cert_path: Option<String>,
 }
 
-impl STSHandler {
-    /// Configures the handler to use JSON encoding instead of Form URL encoding.
+/// Defines the type of body serialization to use for requests to the STS endpoint.
+#[derive(Debug, Default)]
+pub(crate) enum BodyEncoding {
+    /// Default: Form URL encoding (application/x-www-form-urlencoded).
+    #[default]
+    UrlEncoded,
+    /// JSON encoding (application/json).
     #[allow(dead_code)]
-    pub(crate) fn with_json_body(mut self) -> Self {
-        self.use_json = true;
+    Json,
+}
+
+impl STSHandler {
+    /// Configures the handler to use the specified request type for token exchange.
+    #[allow(dead_code)]
+    pub(crate) fn with_body_encoding(mut self, body_encoding: BodyEncoding) -> Self {
+        self.body_encoding = body_encoding;
         self
     }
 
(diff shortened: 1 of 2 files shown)


src/auth/src/credentials/internal/sts_exchange.rs
```diff
@@ -96,6 +130,25 @@ impl STSHandler {
...
+    let cert_bytes = tokio::fs::read(&path).await.map_err(|e| {
+        CredentialsError::from_msg(
+            false,
+            format!("failed to read custom CA certificate from {}: {}", path, e),
+        )
+    })?;
+    let cert = reqwest::Certificate::from_pem(&cert_bytes).map_err(|e| {
+        CredentialsError::from_msg(
+            false,
+            format!("failed to parse custom CA certificate: {}", e),
```
coryan: nit:


```suggestion
            format!("failed to parse custom CA certificate: {e}"),
```

Note that this loses the `e` as a source and the chain of sources behind it. I thought there was a way to preserve that in `CredentialsError`?

src/auth/src/credentials/internal/sts_exchange.rs
```diff
@@ -96,6 +130,25 @@ impl STSHandler {
...
 }
 
+async fn add_root_cert(
+    builder: reqwest::ClientBuilder,
+    path: String,
+) -> Result<reqwest::ClientBuilder> {
+    let cert_bytes = tokio::fs::read(&path).await.map_err(|e| {
+        CredentialsError::from_msg(
+            false,
+            format!("failed to read custom CA certificate from {}: {}", path, e),
```
coryan: nit:

```suggestion
            format!("failed to read custom CA certificate from {path}: {e}"),
```

src/auth/src/credentials/internal/sts_exchange.rs
```diff
@@ -24,14 +24,35 @@ type Result<T> = std::result::Result<T, CredentialsError>;
 
 /// Handles OAuth2 Secure Token Service (STS) exchange.
 /// Reference: https://datatracker.ietf.org/doc/html/rfc8693
-pub struct STSHandler {}
+#[derive(Debug, Default)]
+pub struct STSHandler {
+    use_json: bool,
```
coryan: Feels like this should be an enum 🤷

src/auth/src/credentials/internal/sts_exchange.rs
```diff
@@ -60,25 +81,38 @@ impl STSHandler {
...
         let mut headers = headers.clone();
         client_auth.inject_auth(&mut headers)?;
 
-        let res = client
-            .post(url)
-            .form(&params)
-            .headers(headers)
+        let builder = client.post(url).headers(headers);
+        let builder = if self.use_json {
+            builder.json(&params)
```
coryan: Note to self:  I checked the `Cargo.toml` and both the `json` and `form` features are enabled.

coryan (approved):
A bunch of nits, please fix and merge.

src/auth/src/credentials/internal/sts_exchange.rs
alvarowolfx: I agree and I thought about it too, but I was too lazy. Added one now, just didn't like the naming.

src/auth/src/credentials/internal/sts_exchange.rs
alvarowolfx: good call. Changed to use CredentialsError::new, which we can keep the source error.

PR #5588, opened 2026-05-04
cleanup(pubsub): remove unnecessary dead_code annotation

suzmue:
eo_extend is now used so no longer needs to be decorated. This change also adds a comment to the batching_options to explain why it is unused (and checks the expectation) so it can be cleaned up if no longer necessary.


PhongChuong (approved):

PhongChuong:
Thanks for the catch.

PR #5589, opened 2026-05-05
docs(samples): replace ignore, no_run doctests with run_all_samples gating

### LINKED ISSUE #5500: Use feature gates instead of ignore tag for samples
Using the \`\`\`ignore tag for samples renders docs with a warning that "this code is not tested".

Instead, we can feature gate the code in the sample, e.g.:

```rs
/// ```
/// # #[cfg(feature = "run_all_samples")]
/// # async fn sample() -> anyhow::Result<()> { ... }
/// ```
```

And build/run all the samples with `cargo test --doc --features run_all_samples`

https://github.com/googleapis/google-cloud-rust/blob/562640dd4d223edd11fd5839a7c9e0eedf146cdd/.gcb/scripts/test-ignored-samples.sh#L32

This also would let us reclaim the \`\`\`ignore tag for samples that are Rust code, but may or may not compile. Currently we use \`\`\`no_rust for this case as a workaround. Note that this is not a big win, as no public docs are affected. It just seems more correct to me.

abhinavgautam01:
Fixes #5500.

## Summary

This PR migrates sample doctest handling from `ignore,no_run` to feature-gated samples.

### What changed

1. Replaced ` ```ignore,no_run ` Rust doc fences with feature-gated snippets:
  - `#[cfg(feature = "run_all_samples")]`
  - wrapped snippet bodies in hidden `# { ... # }` blocks

2. Added `run_all_samples = []` feature flags to affected crates so gated snippets compile consistently.

3. Updated post-merge CI wiring for samples:
  - renamed trigger key from `ignored-samples` to `run-all-samples`
  - replaced script `test-ignored-samples.sh` with `test-run-all-samples.sh`
  - switched doctest execution to:
    `cargo test --workspace --doc --all-features -- --show-output`
    (no `--ignored` path)

4. Updated contributor docs to make feature-gated samples the recommended approach and clarify that `ignore` is now for rare cases that must stay excluded from doc-test compilation.

## Why

Using `ignore` marks snippets as “not tested” and relies on a separate `--ignored` workflow.
Feature-gated samples keep snippets in normal doctest compilation while still allowing us to control when they are included.

## Validation

- Ran representative doctests with all features on:
 - `google-cloud-common`
 - `google-cloud-secretmanager-v1`
 - `google-cloud-storage`
 - `google-cloud-pubsub`
 - `google-cloud-firestore`

### DIFF
--- a/.gcb/builds/triggers/main.tf
+++ b/.gcb/builds/triggers/main.tf
@@ -162,11 +162,11 @@ locals {
       script = "workspace-minimal-versions"
     }
 
-    # Runs doctest for samples marked ignore.
+    # Runs doctest for feature-gated samples.
     # These are mostly generated samples which are many and take a long time to build.
-    ignored-samples = {
+    run-all-samples = {
       config = "complex.yaml"
-      script = "test-ignored-samples"
+      script = "test-run-all-samples"
       included_files = [
         "**/generated/**",
       ]
--- a/.gcb/scripts/test-run-all-samples.sh
+++ b/.gcb/scripts/test-run-all-samples.sh
@@ -26,10 +26,10 @@ rustup component add clippy
 cargo version
 rustup show active-toolchain -v
 
-echo "RUSTFLAGS in test-ignored-samples: ${RUSTFLAGS:-}"
-echo "RUSTDOCFLAGS in test-ignored-samples: ${RUSTDOCFLAGS:-}"
+echo "RUSTFLAGS in test-run-all-samples: ${RUSTFLAGS:-}"
+echo "RUSTDOCFLAGS in test-run-all-samples: ${RUSTDOCFLAGS:-}"
 
-cargo test --workspace --doc --all-features -- --ignored --show-output
+cargo test --workspace --doc --all-features -- --show-output
 
 echo "==== DONE ===="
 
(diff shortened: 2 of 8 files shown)


coryan:
Thanks for the PR.  As the name implies, all the files in the `src/generated/` directory are generated. The right fix is to change the generator and call it again.  Any manual changes such as the one in this PR will be lost the next time we generate the code, and therefore we cannot accept them.

I did not look at the other changes in any detail.

abhinavgautam01:
Thanks for the review @coryan! You're absolutely right—I made a critical mistake by manually editing files in `src/generated/`. Those files are auto-generated, so my changes would be lost the next time the generator runs.

I've now reverted all manual changes to `src/generated/` files. The PR now only includes:

* Hand-written code changes (`storage/open_object.rs`)
* Feature flag additions to hand-written crates (storage, pubsub, firestore, wkt)
* Documentation updates
* CI trigger and script updates

The correct approach is to modify the librarian generator templates to add the feature-gating, then re-run `librarian generate --all`. I'm currently working on this in parallel and will update this PR once the librarian changes are merged.

Please let me know if this approach looks correct or if you'd suggest a different direction.

dbolduc:
Thanks, but I think we should close this for the same reason as https://github.com/googleapis/librarian/pull/5819#issuecomment-4381294575. The juice isn't worth the squeeze yet.

We really appreciate the effort though.

abhinavgautam01:
Thanks for the clarification! That makes sense.
Appreciate you taking the time to review this...!

PR #5591, opened 2026-05-05
impl(auth): add GDCH credential and token provider

alvarowolfx:
Towards #5584

### DIFF
--- a/src/auth/src/credentials/gdch.rs
+++ b/src/auth/src/credentials/gdch.rs
@@ -14,14 +14,24 @@
 
 //! [Google Distributed Cloud] service identity authentication.
 
+use super::internal::sts_exchange::{ExchangeTokenRequest, STSHandler};
+use crate::constants::{GDCH_SERVICEACCOUNT_TOKEN_TYPE, TOKEN_EXCHANGE_TOKEN_TYPE};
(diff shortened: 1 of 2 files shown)


src/auth/src/credentials/gdch.rs
```diff
@@ -220,5 +288,119 @@ mod tests {
...
+            .ok_or_else(|| anyhow::anyhow!("missing auth header"))?;
+
+        assert_eq!(headers.len(), expected_len);
+        assert_eq!(
+            token_val,
+            http::HeaderValue::from_static("Bearer test-token")
+        );
+
+        if let Some(qp) = quota_project {
+            let quota_project_header = headers.get(crate::credentials::QUOTA_PROJECT_KEY).unwrap();
```
suzmue, line 362: nit: can we use ?  here?
```suggestion
            let quota_project_header = headers.get(crate::credentials::QUOTA_PROJECT_KEY)?;
```

suzmue (approved):

src/auth/src/credentials/gdch.rs
alvarowolfx: this is an Option, not a Result

PR #5593, opened 2026-05-05
impl(o11y): add feature flag and state for LRO tracing

### LINKED ISSUE #5623: Implement core state and options for LRO tracing
- Introduce `TracingDetails` and `PollerOptions`.
- Add `new_poller_with_options` taking `PollerOptions`.
- Use `#[cfg(google_cloud_unstable_tracing)]` for gating.

haphungw:
Define the `google-cloud-rust-unstable-tracing` flag to gate the unstable tracing implementation for complex flows (including LROs). Update the `PollerImpl` struct and `new_poller` function.

This change sets up the initialization logic required for implementing LRO spans.

Fixes https://github.com/googleapis/google-cloud-rust/issues/5623

### DIFF
--- a/src/lro/src/internal/aip151.rs
+++ b/src/lro/src/internal/aip151.rs
@@ -48,7 +48,60 @@ where
         + Send
         + 'static,
 {
-    PollerImpl::new(polling_error_policy, polling_backoff_policy, start, query)
+    new_poller_with_tracing(
+        polling_error_policy,
+        polling_backoff_policy,
+        start,
+        query,
+        "",
+        false,
+    )
+}
+
+pub fn new_poller_with_tracing<ResponseType, MetadataType, S, SF, Q, QF>(
+    polling_error_policy: Arc<dyn PollingErrorPolicy>,
+    polling_backoff_policy: Arc<dyn PollingBackoffPolicy>,
+    start: S,
+    query: Q,
+    method_name: &'static str,
+    enable_tracing: bool,
+) -> impl Poller<ResponseType, MetadataType>
+where
+    ResponseType: Message + serde::ser::Serialize + serde::de::DeserializeOwned + Send,
+    MetadataType: Message + serde::ser::Serialize + serde::de::DeserializeOwned + Send,
+    S: FnOnce() -> SF + Send + Sync,
+    SF: std::future::Future<Output = Result<Operation<ResponseType, MetadataType>>>
+        + Send
+        + 'static,
+    Q: Fn(String) -> QF + Send + Sync + Clone,
+    QF: std::future::Future<Output = Result<Operation<ResponseType, MetadataType>>>
+        + Send
+        + 'static,
+{
+    #[cfg(feature = "google-cloud-rust-unstable-tracing")]
+    let lro_span = if enable_tracing {
+        Some(tracing::info_span!(
+            "LRO Wait",
+            "gcp.rpc.method" = method_name,
+            "gcp.lro.operation_name" = tracing::field::Empty
+        ))
+    } else {
+        None
+    };
+
+    #[cfg(not(feature = "google-cloud-rust-unstable-tracing"))]
+    let _ = method_name;
+    #[cfg(not(feature = "google-cloud-rust-unstable-tracing"))]
+    let _ = enable_tracing;
+
+    PollerImpl::new(
+        polling_error_policy,
+        polling_backoff_policy,
+        start,
+        query,
+        #[cfg(feature = "google-cloud-rust-unstable-tracing")]
+        lro_span,
+    )
 }
 
 /// Creates a new `impl Poller<(), M>` from the closures created by the generator.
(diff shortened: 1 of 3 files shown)


src/lro/src/internal/aip151.rs
```diff
@@ -48,7 +50,30 @@ where
         + Send
         + 'static,
 {
-    PollerImpl::new(polling_error_policy, polling_backoff_policy, start, query)
+    #[cfg(feature = "google-cloud-rust-unstable-tracing")]
+    let lro_span = if enable_tracing {
+        Some(tracing::info_span!(
+            "LRO Wait",
+            method = method_name,
+            gcp.lro.operation_name = tracing::field::Empty
```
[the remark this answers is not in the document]
haphungw: fixed

src/lro/src/internal/aip151.rs
```diff
@@ -48,7 +48,60 @@ where
...
+        polling_error_policy,
+        polling_backoff_policy,
+        start,
+        query,
+        "",
+        false,
+    )
+}
+
+pub fn new_poller_with_tracing<ResponseType, MetadataType, S, SF, Q, QF>(
```
suzmue, line 65: Could you add a doc comment to this new public function please?

src/lro/src/internal/aip151.rs
suzmue, line 33: is the plan to deprecate this after switching usage over to new_poller_with_tracing?

src/lro/src/internal/aip151.rs
```diff
@@ -48,7 +48,60 @@ where
         + Send
         + 'static,
 {
-    PollerImpl::new(polling_error_policy, polling_backoff_policy, start, query)
+    new_poller_with_tracing(
+        polling_error_policy,
+        polling_backoff_policy,
+        start,
+        query,
+        "",
```
suzmue, line 56: It's kind of odd to have to do this (pass in this default value that goes unused), but since we are just using this in the generated code (and I assume those will alll have a method_name to pass?) then that seems all good. Otherwise would maybe suggest this as an option.

src/lro/src/internal/aip151.rs
```diff
@@ -48,7 +48,60 @@ where
...
+    QF: std::future::Future<Output = Result<Operation<ResponseType, MetadataType>>>
+        + Send
+        + 'static,
+{
+    #[cfg(feature = "google-cloud-rust-unstable-tracing")]
+    let lro_span = if enable_tracing {
+        Some(tracing::info_span!(
+            "LRO Wait",
+            "gcp.rpc.method" = method_name,
+            "gcp.lro.operation_name" = tracing::field::Empty
```
suzmue, line 90: Is this going to always be empty at initialization? Where will the operation id be coming from?

src/lro/src/internal/aip151.rs
```diff
@@ -48,7 +48,60 @@ where
...
+        ))
+    } else {
+        None
+    };
+
+    #[cfg(not(feature = "google-cloud-rust-unstable-tracing"))]
+    let _ = method_name;
+    #[cfg(not(feature = "google-cloud-rust-unstable-tracing"))]
+    let _ = enable_tracing;
```
suzmue: nit: Consider
```suggestion
    #[cfg(not(feature = "google-cloud-rust-unstable-tracing"))]
    {
      let _ = method_name;
      let _ = enable_tracing;
    }
```

src/lro/src/internal/aip151.rs
```diff
@@ -209,6 +262,9 @@ struct PollerImpl<S, Q> {
     query: Q,
     operation: Option<String>,
     state: PollingState,
+    #[cfg(feature = "google-cloud-rust-unstable-tracing")]
+    #[allow(dead_code)]
```
suzmue: consider:

```suggestion
    #[expect(dead_code)]
```

suzmue (approved):

src/lro/src/internal/aip151.rs
haphungw: done

src/lro/src/internal/aip151.rs
haphungw: done

src/lro/src/internal/aip151.rs
haphungw: Yes, I plan to deprecate `new_poller` after we update the sidekick templates.

src/lro/src/internal/aip151.rs
haphungw: I'm planning to clean this up afterwards.

src/lro/src/internal/aip151.rs
haphungw: It is empty because we don't have the Operation ID yet. We will record the actual operation name on the T2 parent span later inside `poll()` once the service returns the operation object.

src/lro/src/internal/aip151.rs
haphungw: done

suzmue (approved):

PR #5596, opened 2026-05-06
impl(auth): add GDCH credential builder

alvarowolfx:
Towards #5584

### DIFF
--- a/src/auth/src/credentials/gdch.rs
+++ b/src/auth/src/credentials/gdch.rs
@@ -26,6 +26,9 @@ use crate::token_cache::TokenCache;
 use crate::{Result, errors};
 use async_trait::async_trait;
 use base64::prelude::{BASE64_URL_SAFE_NO_PAD, Engine as _};
+use google_cloud_gax::backoff_policy::BackoffPolicyArg;
+use google_cloud_gax::retry_policy::RetryPolicyArg;
+use google_cloud_gax::retry_throttler::RetryThrottlerArg;
 use http::{Extensions, HeaderMap};
 use rustls::sign::Signer;
 use rustls_pki_types::PrivateKeyDer;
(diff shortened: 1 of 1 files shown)


alvarowolfx:
better docs and examples are going to be added in coming PRs, when making `mod gdch` and `gdch` feature public.

suzmue (commented):
My comments here since commenting is down:

Question: Why do we not need to do verification on the cred_type? We are always looking for a specific value there, so I was curious about why it isn't checked like the format_version.

The test that has no format version says "missing format_version and other required fields". The code path that is tested for missing the format_version appears to be a special case. If the other fields are required, a separate test to verify those fail would be nice.

The descriptive test names are great for the retry policy!

minor naming nit: the retry test names read like they’re asserting GDCH’s default retry semantics, but the tests are actually verifying that retry settings provided through the builder are applied to the token-fetch path. Maybe consider renaming to something more explicit, like gdch_applies_retry_settings_for_transient_failures

alvarowolfx:
thanks @suzmue 

> Why do we not need to do verification on the cred_type? We are always looking for a specific value there, so I was curious about why it isn't checked like the format_version

We don't check credential types in the credential itself, mainly because the happy path for customer is to use via ADC, and ADC routes to the right implementation via the credential type. And if they want to use an specific type, they can use it directly, in which the type itself is not that important, only the required fields. But GDCH is not being used on ADC flows, so maybe make sense to check the type. I added the check. 

> The test that has no format version says "missing format_version and other required fields". The code path that is tested for missing the format_version appears to be a special case. If the other fields are required, a separate test to verify those fail would be nice.

I split the test cases. Also at a given point we are just testing the behavior of `serde_json::from_value::<GdchServiceAccountKey>`, which fails if a non optional field is missing, so is hard to draw a line on how far should we go here. 

Also changes the name of the tests around retry policy.

src/auth/src/credentials/gdch.rs
```diff
@@ -191,6 +194,114 @@ struct GdchServiceAccountCredentials {
     quota_project_id: Option<String>,
 }
 
+/// A builder for [`GdchServiceAccountCredentials`].
+#[allow(dead_code)]
+pub(crate) struct Builder {
+    key: serde_json::Value,
```
coryan, line 200: Can we parse this into a more meaningful type during initialization?

coryan: Oh, I see, we validate when we call `build*()`... hmm, weird, maybe, fine if we do the same everywhere else.

coryan (approved):

src/auth/src/credentials/gdch.rs
alvarowolfx: yeah, I just followed the pattern from other builders, where the `Value` is validated on `build`

PR #5605, opened 2026-05-06
refactor(o11y): introduce `PollerOptions`  

haphungw:
Refactor LRO poller creation to use an options struct instead of positional arguments. This prepares for future extensions (like tracing) without breaking signatures. 

For https://github.com/googleapis/google-cloud-rust/issues/5623

### DIFF
--- a/src/lro/src/internal/aip151.rs
+++ b/src/lro/src/internal/aip151.rs
@@ -48,27 +48,33 @@ where
         + Send
         + 'static,
 {
-    new_poller_with_tracing(
+    new_poller_with_options(
         polling_error_policy,
         polling_backoff_policy,
         start,
         query,
-        "",
-        false,
+        PollerOptions::default(),
     )
 }
 
-/// Creates a new `impl Poller<R, M>` with tracing support enabled.
+/// Options for creating a new poller.
+#[derive(Default)]
+#[non_exhaustive]
+pub struct PollerOptions {
+    pub method_name: &'static str,
+    pub enable_tracing: bool,
+}
+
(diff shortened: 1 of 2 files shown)


dbolduc:
> introduce `PollerOptions` and deprecate `new_poller`

I suggest doing one thing at a time.

src/lro/src/internal/aip151.rs
```diff
@@ -48,27 +48,33 @@ where
...
     )
 }
 
-/// Creates a new `impl Poller<R, M>` with tracing support enabled.
+/// Options for creating a new poller.
+#[derive(Default)]
+#[non_exhaustive]
+pub struct PollerOptions {
+    pub method_name: &'static str,
+    pub enable_tracing: bool,
```
dbolduc: Alternative organization:

If the method name is only relevant for tracing, you could have like:

```rs
#[non_exhaustive]
pub struct TracingDetails {
  pub method_name: &'static str,
}

#[non_exhaustive]
pub struct PollerOptions {
  pub tracing: Option<TracingDetails>,
}
```

then `lro_span` is just a map of the `options.tracing` field. 🤷

dbolduc (approved):
This is better. Left an option for you to consider that I think makes the code simpler, but maybe it is overkill. WDYT?

Note that these are public APIs so we only have until the next release to lock in the interfaces.

src/lro/src/internal/aip151.rs
haphungw: This is definitely better for concern separation. Thanks!

PR #5612, opened 2026-05-07
fix(storage): rate limit bucket deletion in cleanup

### LINKED ISSUE #5219: ci: rate limit storage bucket deletion
Spawned from #5212:

When there are a large number of stale buckets, we can exceed the storage bucket creation/deletion API of `approximately one request every two seconds` [limit](https://docs.cloud.google.com/storage/quotas).

We can avoid this by serializing the bucket deletion in [cleanup_bucket](https://github.com/googleapis/google-cloud-rust/blob/2962a5c674a967448dcc4900f325b5ca50326a56/src/storage/examples/src/lib.rs#L905).

abhinavgautam01:
Fixes #5219
 
 ## Problem
 When there are many stale buckets in integration tests, the cleanup process 
 was deleting them in parallel and exceeding GCP's Storage API rate limit 
 (approximately one request every two seconds).
 
 ## Solution
 Serialize bucket deletion by removing parallel spawning (tokio::spawn and 
 join_all) and instead delete buckets sequentially with a 2-second delay 
 between each deletion to respect the API rate limit.
 
 ## Changes
 - Changed `cleanup_stale_buckets()` to delete buckets sequentially
 - Added 2-second delay between bucket deletions
 - Maintains proper error handling for each deletion attempt
 - Delay is not applied after the last bucket (no unnecessary wait)
 
 ## Testing
 This change should prevent rate limit errors during stale bucket cleanup 
 in integration tests.

### DIFF
--- a/src/storage/examples/src/lib.rs
+++ b/src/storage/examples/src/lib.rs
@@ -23,10 +23,8 @@ use google_cloud_gax::options::RequestOptionsBuilder;
 use google_cloud_gax::paginator::ItemPaginator as _;
 use google_cloud_gax::throttle_result::ThrottleResult;
 use google_cloud_gax::{
-    backoff_policy::BackoffPolicy,
-    error::rpc::Code,
-    exponential_backoff::ExponentialBackoffBuilder,
-    retry_policy::RetryPolicyExt,
+    backoff_policy::BackoffPolicy, error::rpc::Code,
+    exponential_backoff::ExponentialBackoffBuilder, retry_policy::RetryPolicyExt,
     retry_state::RetryState,
 };
 use google_cloud_storage::client::{Storage, StorageControl};
--- a/src/storage/examples/src/lib.rs
+++ b/src/storage/examples/src/lib.rs
@@ -23,7 +23,6 @@ use google_cloud_gax::options::RequestOptionsBuilder;
 use google_cloud_gax::paginator::ItemPaginator as _;
 use google_cloud_gax::throttle_result::ThrottleResult;
 use google_cloud_gax::{
-    backoff_policy::BackoffPolicy, error::rpc::Code,
     exponential_backoff::ExponentialBackoffBuilder, retry_policy::RetryPolicyExt,
     retry_state::RetryState,
 };
@@ -797,24 +796,13 @@ pub async fn cleanup_stale_buckets(
     Ok(())
 }
 
-fn delete_bucket_error_should_backoff(e: &google_cloud_gax::error::Error) -> bool {
-    if e.is_transient_and_before_rpc() {
-        return true;
-    }
-    matches!(
-        e.status().map(|s| s.code),
-        Some(
-            Code::ResourceExhausted
-                | Code::Unavailable
-                | Code::DeadlineExceeded
-                | Code::Aborted
-                | Code::Internal
-        )
-    )
+pub async fn cleanup_bucket(client: StorageControl, name: String) -> anyhow::Result<()> {
+    empty_bucket_contents(&client, &name).await?;
+    delete_bucket_with_error_backoff(&client, &name).await
 }
 
-/// Delete one bucket, sleeping with exponential backoff (initial delay ≥ 2s,
-/// capped) between attempts on retryable errors. See [`cleanup_stale_buckets`].
+/// Delete one bucket, using the client's retry loop with exponential backoff
+/// (initial delay >= 2s, capped). See [`cleanup_stale_buckets`].
 async fn delete_bucket_with_error_backoff(
     client: &StorageControl,
     name: &str,
@@ -823,24 +811,14 @@ async fn delete_bucket_with_error_backoff(
         .with_initial_delay(Duration::from_secs(2))
         .with_maximum_delay(Duration::from_secs(120))
         .build()?;
-    let mut state = RetryState::new(true);
-    let deadline = std::time::Instant::now() + Duration::from_secs(900);
-    loop {
-        match client.delete_bucket().set_name(name).send().await {
-            Ok(()) => return Ok(()),
-            Err(e) => {
-                if std::time::Instant::now() > deadline {
-                    return Err(e.into());
-                }
-                if !delete_bucket_error_should_backoff(&e) {
-                    return Err(e.into());
-                }
-                let wait = backoff.on_failure(&state);
-                tokio::time::sleep(wait).await;
-                state.attempt_count = state.attempt_count.saturating_add(1);
-            }
-        }
-    }
+    client
+        .delete_bucket()
+        .set_name(name)
+        .with_backoff_policy(backoff)
+        .with_idempotency(true)
+        .send()
+        .await?;
+    Ok(())
 }
 
 /// List and remove objects, managed folders, folders, and anywhere caches so
@@ -851,6 +829,9 @@ async fn empty_bucket_contents(client: &StorageControl, name: &str) -> anyhow::R
     use google_cloud_wkt::FieldMask;
 
     let current = client.get_bucket().set_name(name).send().await?;
+    // Configure the bucket to be garbage collected. Some buckets are created by
+    // sample code, which does not (and should not) include setting labels to
+    // automatically garbage collect the bucket.
     if current
         .labels
         .get("integration-test")
@@ -968,11 +949,6 @@ The default credentials (see below) are configured to use project {project:?}.
 {credentials:?}"#
     );
     Ok(true)
-}
-
-pub async fn cleanup_bucket(client: StorageControl, name: String) -> anyhow::Result<()> {
-    empty_bucket_contents(&client, &name).await?;
-    delete_bucket_with_error_backoff(&client, &name).await
 }
 
 fn enable_info_tracing() -> tracing::subscriber::DefaultGuard {
--- a/src/storage/examples/src/lib.rs
+++ b/src/storage/examples/src/lib.rs
@@ -749,8 +749,7 @@ pub async fn cleanup_stale_buckets(
         .list_buckets()
         .set_parent(format!("projects/{project_id}"))
         .by_item();
-    let mut pending = Vec::new();
-    let mut names = Vec::new();
+    let mut buckets_to_cleanup = Vec::new();
     while let Some(bucket) = buckets.next().await {
         let bucket = bucket?;
         if bucket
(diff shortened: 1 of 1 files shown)


src/storage/examples/src/lib.rs
```diff
@@ -759,20 +758,20 @@ pub async fn cleanup_stale_buckets(
...
-    println!("cleaning up {} buckets", pending.len());
-    let results = futures::future::join_all(pending).await;
-    let errors = results.into_iter().zip(names).filter(|(r, _)| r.is_err());
-    for (r, name) in errors {
-        println!("error deleting bucket {name}: {r:?}");
+    println!("cleaning up {} buckets", buckets_to_cleanup.len());
+    // Serialize bucket deletion to respect GCP rate limit (~1 deletion every 2 seconds)
+    for (idx, name) in buckets_to_cleanup.iter().enumerate() {
+        if let Err(e) = cleanup_bucket(client.clone(), name.clone()).await {
+            println!("error deleting bucket {name}: {e:?}");
```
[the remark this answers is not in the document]
abhinavgautam01: Thanks for the feedback! You're right—using structured logging is more consistent with the repository's standards. I've replaced the println! calls with tracing::info!() and tracing::error!() to align with the rest of the codebase. Updated and pushed.

src/storage/examples/src/lib.rs
```diff
@@ -759,20 +758,20 @@ pub async fn cleanup_stale_buckets(
...
+        }
+        // Add delay between deletions, but not after the last one
+        if idx < buckets_to_cleanup.len() - 1 {
+            tokio::time::sleep(tokio::time::Duration::from_secs(2)).await;
+        }
```
coryan: This is not an effective fix.  Multiple tests may (and will) be running at the same time, all of them will delete multiple clients at the same time, this sleep will do nothing to coordinate them.

A better (but not perfect) fix is to use a backoff policy that starts at (at least) 2 seconds. That way all clients slow down to at least 2 seconds when there is an error.

src/storage/examples/src/lib.rs
```diff
@@ -759,20 +758,20 @@ pub async fn cleanup_stale_buckets(
             .is_some_and(|v| v == "true")
             && bucket.create_time.is_some_and(|v| v < stale_deadline)
         {
-            let client = client.clone();
-            let name = bucket.name.clone();
-            pending.push(tokio::spawn(
-                async move { cleanup_bucket(client, name).await },
-            ));
-            names.push(bucket.name);
+            buckets_to_cleanup.push(bucket.name);
```
coryan, line 761: You may want to serialize the removal of the buckets, but the removal of the bucket contents should proceed in parallel.

coryan: BTW, the algorithm will look more complex. Add some comments explaining the motivation, because future-us will be confused when we see the complexity N years form now.

coryan (changes requested):

src/storage/examples/src/lib.rs
abhinavgautam01: Agreed — a fixed sleep only spaces out deletes in this process and doesn’t align with other CI workers. I removed that and retry DeleteBucket with ExponentialBackoffBuilder, with_initial_delay at least 2s, only on retryable / rate-limit-style errors (e.g. ResourceExhausted, Unavailable, plus pre-RPC transient), with a max total wait so we don’t spin forever.

src/storage/examples/src/lib.rs
abhinavgautam01: Done: bucket contents are still cleared with the same per-bucket join_all pattern; for stale cleanup we now run empty_bucket_contents for all buckets in parallel, then call DeleteBucket one bucket at a time with the backoff above. I split helpers (empty_bucket_contents / delete_bucket_with_error_backoff) and added a short comment block on why the flow looks asymmetric (many parallel test jobs, shared quotas, future readers).

abhinavgautam01:
ping @coryan

coryan:
Formatting is failing with:


```
diff --git a/src/storage/examples/src/lib.rs b/src/storage/examples/src/lib.rs
index 6c42795..22eedff 100644
--- a/src/storage/examples/src/lib.rs
+++ b/src/storage/examples/src/lib.rs
@@ -23,10 +23,8 @@ use google_cloud_gax::options::RequestOptionsBuilder;
 use google_cloud_gax::paginator::ItemPaginator as _;
 use google_cloud_gax::throttle_result::ThrottleResult;
 use google_cloud_gax::{
-    backoff_policy::BackoffPolicy,
-    error::rpc::Code,
-    exponential_backoff::ExponentialBackoffBuilder,
-    retry_policy::RetryPolicyExt,
+    backoff_policy::BackoffPolicy, error::rpc::Code,
+    exponential_backoff::ExponentialBackoffBuilder, retry_policy::RetryPolicyExt,
     retry_state::RetryState,
 };
 use google_cloud_storage::client::{Storage, StorageControl};
```

Remember to always use `cargo format` from the top-level directory so the configuration takes effect.  I think in this case  `cargo format -p storage-samples` would fix the problem.

abhinavgautam01:
> Formatting is failing with:
> 
> ```
> diff --git a/src/storage/examples/src/lib.rs b/src/storage/examples/src/lib.rs
> index 6c42795..22eedff 100644
> --- a/src/storage/examples/src/lib.rs
> +++ b/src/storage/examples/src/lib.rs
> @@ -23,10 +23,8 @@ use google_cloud_gax::options::RequestOptionsBuilder;
>  use google_cloud_gax::paginator::ItemPaginator as _;
>  use google_cloud_gax::throttle_result::ThrottleResult;
>  use google_cloud_gax::{
> -    backoff_policy::BackoffPolicy,
> -    error::rpc::Code,
> -    exponential_backoff::ExponentialBackoffBuilder,
> -    retry_policy::RetryPolicyExt,
> +    backoff_policy::BackoffPolicy, error::rpc::Code,
> +    exponential_backoff::ExponentialBackoffBuilder, retry_policy::RetryPolicyExt,
>      retry_state::RetryState,
>  };
>  use google_cloud_storage::client::{Storage, StorageControl};
> ```
> 
> Remember to always use `cargo format` from the top-level directory so the configuration takes effect. I think in this case `cargo format -p storage-samples` would fix the problem.

Good catch — ran the formatter from the repo root with cargo fmt -p storage-samples so the workspace rustfmt config applies; it collapsed the google_cloud_gax import list to match the diff you pasted.

src/storage/examples/src/lib.rs
```diff
@@ -759,50 +759,98 @@ pub async fn cleanup_stale_buckets(
...
+                }
+                if !delete_bucket_error_should_backoff(&e) {
+                    return Err(e.into());
+                }
+                let wait = backoff.on_failure(&state);
+                tokio::time::sleep(wait).await;
+                state.attempt_count = state.attempt_count.saturating_add(1);
+            }
+        }
+    }
```
coryan: The client library already has a retry loop, and it already includes the error types you want.  This can be replaced with:

```suggestion
    let backoff = ExponentialBackoffBuilder::new()
        .with_initial_delay(Duration::from_secs(2))
        .with_maximum_delay(Duration::from_secs(120))
        .build()?;
    client.delete_bucket().set_name(name).with_backoff_policy(backoff).with_idempotency(true).send()
```

And then you can remove the `delete_bucket_error_should_backoff()` function.

src/storage/examples/src/lib.rs
```diff
@@ -759,50 +759,98 @@ pub async fn cleanup_stale_buckets(
...
+/// List and remove objects, managed folders, folders, and anywhere caches so
+/// the bucket can be deleted. Per-resource deletes are still parallelized with
+/// `join_all` inside this bucket.
+async fn empty_bucket_contents(client: &StorageControl, name: &str) -> anyhow::Result<()> {
     use google_cloud_gax::{Result as GaxResult, paginator::ItemPaginator};
     use google_cloud_wkt::FieldMask;
 
-    // Configure the bucket to be garbage collected. Some buckets are created by
-    // sample code, which does not (and should not) include setting labels to
-    // automatically garbage collect the bucket.
```
coryan, line 804: Please restore these comments.

src/storage/examples/src/lib.rs
```diff
@@ -900,12 +948,33 @@ pub async fn cleanup_bucket(client: StorageControl, name: String) -> anyhow::Res
...
+The default credentials (see below) are configured to use project {project:?}.
+{credentials:?}"#
+    );
+    Ok(true)
+}
+
+pub async fn cleanup_bucket(client: StorageControl, name: String) -> anyhow::Result<()> {
+    empty_bucket_contents(&client, &name).await?;
+    delete_bucket_with_error_backoff(&client, &name).await
+}
```
coryan: group this with the other cleanup functions.

src/storage/examples/src/lib.rs
```diff
@@ -759,50 +759,98 @@ pub async fn cleanup_stale_buckets(
...
+                | Code::Unavailable
+                | Code::DeadlineExceeded
+                | Code::Aborted
+                | Code::Internal
+        )
+    )
 }
 
-pub async fn cleanup_bucket(client: StorageControl, name: String) -> anyhow::Result<()> {
+/// Delete one bucket, sleeping with exponential backoff (initial delay ≥ 2s,
```
coryan: We try to use ASCII in comments:

```suggestion
/// Delete one bucket, sleeping with exponential backoff (initial delay >= 2s,
```

coryan (changes requested):

src/storage/examples/src/lib.rs
abhinavgautam01: done... delete_bucket_with_error_backoff now builds the same exponential backoff and passes it through with_backoff_policy, sets with_idempotency(true) and send().await?. The custom delete_bucket_error_should_backoff helper is gone; the library’s retry policy handles the error classes we care about...

src/storage/examples/src/lib.rs
abhinavgautam01: Restored the three-line comment above the integration-test label update in empty_bucket_contents so the GC motivation next to that block is back.

src/storage/examples/src/lib.rs
abhinavgautam01: cleanup_bucket now sits right after cleanup_stale_buckets, with delete_bucket_with_error_backoff and empty_bucket_contents following as the private helpers used by both paths.

src/storage/examples/src/lib.rs
abhinavgautam01: Updated the doc comment to use >= 2s and adjusted the wording to describe backoff via the client retry path instead of “sleeping” in our own loop...

coryan (approved):

coryan:
This is looking good, I need a review from somebody in the @googleapis/gcs-team

dbolduc (approved):

PR #5614, opened 2026-05-07
feat(auth): make GDCH authentication pub

alvarowolfx:
Towards #5584


alvarowolfx:
cc @howardjohn this PR makes the gdch authentication available/public. It will be under the `gdch` feature.

src/auth/Cargo.toml
```diff
@@ -80,8 +80,8 @@ default = ["default-idtoken-backend", "default-rustls-provider"]
 # The `idtoken` feature enables support to create and validate OIDC ID Tokens.
 # See the create top-level documentation for more information.
 idtoken = ["dep:jsonwebtoken", "jsonwebtoken"]
-# The `__gdch` feature enables support for GDCH service account credentials.
-__gdch = ["dep:p256"]
+# The `gdch` feature enables support for GDCH service account credentials.
+gdch = ["dep:p256"]
```
howardjohn, line 84: We only use `der` from this. But we pull in p256 -> ecdsa -> der. might be worth pruning it down

howardjohn:
I tested this on my environment and it works as expected. thanks!

src/auth/Cargo.toml
alvarowolfx: I tried to use `der` directly, but I had to add some extra code that was not that easy to follow. I can try to use `ecdsa` directly maybe. Will check.

But that's part of the reason we gated this into a rust feature, so we don't take the penalty of the extra lib for the most common use cases of the auth library.

howardjohn: ah my bad, was reading it wrong.

You could use only der but would need a bit of logic like

```
   
fn ecdsa_der_to_jose(der: &[u8], field_len: usize) -> Result<Vec<u8>> {
   let mut reader = SliceReader::new(der).map_err(errors::non_retryable)?;
    let (r, s) = reader
        .sequence(|reader| {
            let r = reader.decode::<UintRef<'_>>()?;
            let s = reader.decode::<UintRef<'_>>()?;
            Ok((r, s))
        })
        .and_then(|signature| reader.finish(signature))
        .map_err(errors::non_retryable)?;

    let mut jose = Vec::with_capacity(field_len * 2);
    append_jose_integer(&mut jose, r.as_bytes(), field_len)?;
    append_jose_integer(&mut jose, s.as_bytes(), field_len)?;
    Ok(jose)
}

fn append_jose_integer(out: &mut Vec<u8>, value: &[u8], field_len: usize) -> Result<()> {
    if value.len() > field_len {
        return Err(errors::non_retryable_from_str(
            "invalid GDCH ECDSA signature integer length",
        ));
    }
    out.extend(std::iter::repeat_n(0, field_len - value.len()));
    out.extend_from_slice(value);
    Ok(())
}
```

`ecdsa` directly should be straightforward though

src/auth/src/credentials.rs
```diff
@@ -792,6 +792,9 @@ fn build_credentials(
                     universe_domain.clone(),
                     |b: external_account::Builder, s: Vec<String>| b.with_scopes(s)
                 ),
+                "gdch_service_account" => Err(BuilderError::not_supported(format!(
```
dbolduc, line 795: (I am probably just late to the party here)

why isn't GDCH supported in ADC? Is there some technical reason?

dbolduc (approved):

src/auth/src/credentials.rs
alvarowolfx: two factors:
* It requires a target `audience`, which we don't collect on the ADC flow and it's not on the GDCH service account file. Adding an `with_audience` on the ADC flow can cause some confusion, since just service accounts have a optional audience.
* We gated it with a `gdch` Rust feature.

dbolduc: Ack, thanks

src/auth/Cargo.toml
alvarowolfx: opened https://github.com/googleapis/google-cloud-rust/pull/5617, but there are other shenanigans on using `der` directly, like it being imported by other packages. Will work on that later.

PR #5617, opened 2026-05-07
refactor(auth): prune p256 dependency down to der crate

alvarowolfx:
Towards #5584


alvarowolfx:
close for now, we can optimize later.

PR #5630, opened 2026-05-08
feat(gax): add client-level attempt_timeout and use as transport default

### LINKED ISSUE #4466: Set per `attempt_timeout` on the client
We expose `.with_attempt_timeout` only on our request builders. They cannot be set on the client builder.

Note that similar properties like the retry policy can be set once per client, and overridden per request. That can simplify some application code that calls multiple RPCs with the same policies.

To support this, we need to:

- add a `fn with_attempt_timeout(...)` on the `gax::client_builder::ClientBuilder`.
- store the value in the `gax::internal::ClientConfig`
- plumb the value into our `gax-internal::http::ReqwestClient` and `gax-internal::grpc::GrpcClient`s.
- use the client default if one is not set per request in both the clients. (e.g. [this logic for HTTP retry policy](https://github.com/googleapis/google-cloud-rust/blob/52859ce38daf5fe721f6250fa872910a3c0689c8/src/gax-internal/src/http.rs#L209))
- and add tests, obviously.

abhinavgautam01:
Fixes #4466

Adds an optional client-level per-attempt timeout and applies it as a transport fallback when requests don't set their own per-attempt timeout.

What changed:

- Add attempt_timeout: Option<Duration> to internal ClientConfig and ClientBuilder::with_attempt_timeout(...)
- gax-internal/http & grpc: store client default and merge into RequestOptions when absent

Behavior:

- Per-request RequestOptions.attempt_timeout still takes precedence.

Tests:

- Ran google-cloud-gax, google-cloud-gax-internal and full workspace tests locally — all pass.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -470,7 +470,7 @@ tokio-stream      = { default-features = false, version = "0.1.16" }
 
 # Local packages used as dependencies.
 google-cloud-auth        = { default-features = false, version = "1.10.0", path = "src/auth" }
-google-cloud-gax         = { default-features = false, version = "1.10.1", path = "src/gax" }
+google-cloud-gax         = { default-features = false, version = "1.11.0", path = "src/gax" }
 gaxi                     = { default-features = false, version = "0.7.13", path = "src/gax-internal", package = "google-cloud-gax-internal" }
 wkt                      = { default-features = false, version = "1.4.0", path = "src/wkt", package = "google-cloud-wkt" }
 google-cloud-wkt         = { default-features = false, version = "1.4.0", path = "src/wkt", package = "google-cloud-wkt" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -470,7 +470,7 @@ tokio-stream      = { default-features = false, version = "0.1.16" }
 
 # Local packages used as dependencies.
 google-cloud-auth        = { default-features = false, version = "1.10.0", path = "src/auth" }
-google-cloud-gax         = { default-features = false, version = "1.10.0", path = "src/gax" }
+google-cloud-gax         = { default-features = false, version = "1.10.1", path = "src/gax" }
 gaxi                     = { default-features = false, version = "0.7.13", path = "src/gax-internal", package = "google-cloud-gax-internal" }
 wkt                      = { default-features = false, version = "1.4.0", path = "src/wkt", package = "google-cloud-wkt" }
 google-cloud-wkt         = { default-features = false, version = "1.4.0", path = "src/wkt", package = "google-cloud-wkt" }
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -77,6 +77,7 @@ pub struct Client {
     retry_throttler: SharedRetryThrottler,
     polling_error_policy: Arc<dyn PollingErrorPolicy>,
     polling_backoff_policy: Arc<dyn PollingBackoffPolicy>,
+    attempt_timeout: Option<Duration>,
 }
 
 impl Client {
@@ -140,6 +141,7 @@ impl Client {
             polling_backoff_policy: config
                 .polling_backoff_policy
                 .unwrap_or_else(|| Arc::new(ExponentialBackoff::default())),
+            attempt_timeout: config.attempt_timeout,
         })
     }
 
@@ -276,7 +278,13 @@ impl Client {
         let headers = self.add_auth_headers(headers).await?;
         let metadata = tonic::MetadataMap::from_headers(headers);
         let mut request = ::tonic::Request::from_parts(metadata, extensions, request);
-        if let Some(attempt_timeout) = options.attempt_timeout() {
+        let mut merged_options = options.clone();
+        if merged_options.attempt_timeout().is_none() {
+            if let Some(t) = self.attempt_timeout {
+                merged_options.set_attempt_timeout(t);
+            }
+        }
+        if let Some(attempt_timeout) = merged_options.attempt_timeout() {
             request.set_timeout(*attempt_timeout);
         }
         let codec = tonic_prost::ProstCodec::<Request, Response>::default();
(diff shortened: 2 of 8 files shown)


src/gax-internal/src/grpc.rs
```diff
@@ -276,7 +278,13 @@ impl Client {
...
-        if let Some(attempt_timeout) = options.attempt_timeout() {
+        let mut merged_options = options.clone();
+        if merged_options.attempt_timeout().is_none() {
+            if let Some(t) = self.attempt_timeout {
+                merged_options.set_attempt_timeout(t);
+            }
+        }
+        if let Some(attempt_timeout) = merged_options.attempt_timeout() {
             request.set_timeout(*attempt_timeout);
         }
```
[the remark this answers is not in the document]
abhinavgautam01: Thanks — great catch. Replaced the clone-based merge with a direct fallback that avoids cloning:

```rust id="8rjzhh"
if let Some(timeout) = crate::options::resolve_effective_timeout(&options, self.attempt_timeout, None) {
    request.set_timeout(timeout);
}
```

The helper lives in `src/gax-internal/src/options.rs`. This removes the extra clone and keeps semantics identical. Ran local tests (`google-cloud-gax` & `google-cloud-gax-internal`) — all passed.

src/gax-internal/src/grpc.rs
```diff
@@ -408,7 +416,13 @@ impl Client {
...
-        if let Some(timeout) = effective_timeout(options, remaining_time) {
+        let mut merged_options = options.clone();
+        if merged_options.attempt_timeout().is_none() {
+            if let Some(t) = self.attempt_timeout {
+                merged_options.set_attempt_timeout(t);
+            }
+        }
+        if let Some(timeout) = effective_timeout(&merged_options, remaining_time) {
             request.set_timeout(timeout);
         }
```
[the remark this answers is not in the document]
abhinavgautam01: Agreed about duplication and `Cow`. Extracted the merge logic into a single helper: `resolve_effective_timeout(...)`

(`src/gax-internal/src/options.rs`) and reused it from both `server_streaming_with_status` and `request_attempt`. This removes the duplicated clone/`Cow` logic and keeps the code concise and efficient. Local tests passed.

src/gax-internal/src/http.rs
```diff
@@ -388,7 +390,16 @@ impl ReqwestClient {
...
-        builder = effective_timeout(options, remaining_time)
+        let merged_options = if options.attempt_timeout().is_some() {
+            options.clone()
+        } else {
+            let mut o = options.clone();
+            if let Some(t) = self.attempt_timeout {
+                o.set_attempt_timeout(t);
+            }
+            o
+        };
```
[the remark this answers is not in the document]
abhinavgautam01: Thanks — applied the same approach in HTTP: removed the clone branch and use the helper to compute the timeout, then call:

```rust id="4kthxk"
builder = builder.timeout(timeout)
```

when present. Also removed the now-unused `effective_timeout` import. Changes are in `src/gax-internal/src/http.rs` and `src/gax-internal/src/options.rs`.

All local tests pass.

abhinavgautam01:
ping @dbolduc

src/gax/src/client_builder.rs
```diff
@@ -386,6 +386,16 @@ impl<F, Cr> ClientBuilder<F, Cr> {
         self
     }
 
+    /// Configure the per-attempt timeout used as the client default.
+    ///
+    /// When set, this timeout will be used for each attempt of a request unless
+    /// a per-request attempt timeout is provided via RequestOptions. Per-request
+    /// settings take precedence.
+    pub fn with_attempt_timeout<V: Into<std::time::Duration>>(mut self, v: V) -> Self {
```
dbolduc, line 394: Can you add a unit test for this function in this file?

dbolduc (commented):
Looks good, thanks! We just need to get past the CI.

I think we need a

```sh
cargo fmt -p google-cloud-gax
cargo fmt -p google-cloud-gax-internal
```

And then because we added a new API to `gax` which `gax-internal` depends on, we need to bump the version of `gax`.

That happens in:
- Cargo.toml
- src/gax/Cargo.toml

Then we need to update the generated `Cargo.lock` with like: `cargo check -p google-cloud-gax`

src/gax/src/client_builder.rs
abhinavgautam01: yep...on it..!

src/gax/Cargo.toml
```diff
@@ -18,7 +18,7 @@ name = "google-cloud-gax"
 #     version of all downstream dependencies. For details see:
 #         https://github.com/googleapis/google-cloud-rust/issues/3237
 #         https://github.com/googleapis/google-cloud-rust/issues/3265
-version     = "1.10.0"
+version     = "1.10.1"
```
coryan: This should be

```suggestion
version     = "1.11.0"
```

A good rule of thumb: if reverting the change is breaking, it must be larger than patch.

Cargo.toml
```diff
@@ -470,7 +470,7 @@ tokio-stream      = { default-features = false, version = "0.1.16" }
 
 # Local packages used as dependencies.
 google-cloud-auth        = { default-features = false, version = "1.10.0", path = "src/auth" }
-google-cloud-gax         = { default-features = false, version = "1.10.0", path = "src/gax" }
+google-cloud-gax         = { default-features = false, version = "1.10.1", path = "src/gax" }
```
coryan:
```suggestion
google-cloud-gax         = { default-features = false, version = "1.11.0", path = "src/gax" }
```

coryan (commented):
@dbolduc I think we also need to make the version change in `librarian.yaml` or `librarian bump` will be confused.

src/gax/Cargo.toml
abhinavgautam01: good stuff, i’ll definitely keep that in mind..!

abhinavgautam01:
ping @dbolduc

dbolduc:
> ping @dbolduc

@coryan is right. The version should be `1.11.0`, and I forgot we need to update it in `librarian.yaml` as well.

https://github.com/googleapis/google-cloud-rust/blob/04350be46d2307ad6333376c1e58d7f140547130/librarian.yaml#L829

Can you fix that?

Cargo.toml
abhinavgautam01: done...

abhinavgautam01:
> > ping @dbolduc
> 
> @coryan is right. The version should be `1.11.0`, and I forgot we need to update it in `librarian.yaml` as well.
> 
> https://github.com/googleapis/google-cloud-rust/blob/04350be46d2307ad6333376c1e58d7f140547130/librarian.yaml#L829
> 
> Can you fix that?

done..!

dbolduc (approved):
LGTM, thanks! Let's see what the builds say.

abhinavgautam01:
> LGTM, thanks! Let's see what the builds say.

hopefully the builds stay green...🤧

coryan (approved):

PR #5631, opened 2026-05-08
feat(storage): add feature flag and models for appendable upload

vsharonlynn:
- Add model extensions for appendable upload.
- Reuses `google_cloud_unstable_storage_bidi` feature flag to gate the feature.

### DIFF
--- a/.gcb/builds/triggers/main.tf
+++ b/.gcb/builds/triggers/main.tf
@@ -35,7 +35,6 @@ locals {
   gcb_secret_name = "projects/${var.project}/secrets/github-github-oauthtoken-319d75/versions/latest"
 
   unstable_flags = join(" ", [
-    "--cfg google_cloud_unstable_storage_appendable_upload",
     "--cfg google_cloud_unstable_trust_boundaries",
     "--cfg google_cloud_unstable_storage_bidi",
     "--cfg google_cloud_unstable_grpc_server_streaming"
--- a/.gcb/builds/triggers/main.tf
+++ b/.gcb/builds/triggers/main.tf
@@ -35,6 +35,7 @@ locals {
   gcb_secret_name = "projects/${var.project}/secrets/github-github-oauthtoken-319d75/versions/latest"
 
   unstable_flags = join(" ", [
+    "--cfg google_cloud_unstable_storage_appendable_upload",
     "--cfg google_cloud_unstable_trust_boundaries",
     "--cfg google_cloud_unstable_storage_bidi",
     "--cfg google_cloud_unstable_grpc_server_streaming"
--- a/src/storage/src/model_ext.rs
+++ b/src/storage/src/model_ext.rs
@@ -318,7 +318,7 @@ pub struct WriteObjectRequest {
 #[non_exhaustive]
 pub struct OpenAppendableObjectRequest {
     /// The object attributes and pre-conditions for the open operation.
-    pub spec: Box<crate::model::WriteObjectSpec>,
+    pub spec: crate::model::WriteObjectSpec,
     /// Additional request parameters.
     pub params: Option<crate::model::CommonObjectRequestParams>,
 }
@@ -489,7 +489,7 @@ pub(crate) mod tests {
     #[test]
     fn test_open_appendable_object_request() {
         let req = OpenAppendableObjectRequest {
-            spec: Box::new(crate::model::WriteObjectSpec::default()),
+            spec: crate::model::WriteObjectSpec::default(),
             params: None,
         };
         assert_eq!(req.spec.resource, None);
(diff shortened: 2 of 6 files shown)


.gcb/builds/triggers/main.tf
```diff
@@ -35,6 +35,7 @@ locals {
   gcb_secret_name = "projects/${var.project}/secrets/github-github-oauthtoken-319d75/versions/latest"
 
   unstable_flags = join(" ", [
+    "--cfg google_cloud_unstable_storage_appendable_upload",
```
coryan: You can use `unstable_storage_bidi`, it is not being used for anything else.

src/storage/src/model_ext.rs
```diff
@@ -310,6 +310,69 @@ pub struct WriteObjectRequest {
...
+#[cfg(google_cloud_unstable_storage_appendable_upload)]
+/// The first message sent in an appendable upload bidi stream.
+#[derive(Debug, PartialEq)]
+#[non_exhaustive]
+pub enum AppendObjectFirstMessage {
+    /// Create a new appendable object.
+    Create(Box<crate::model::WriteObjectSpec>),
+    /// Take over an existing appendable object.
+    Takeover(AppendObjectSpec),
+}
```
coryan: This seems like it should be two separate methods, not one method with an enum type.

coryan (commented):
This points to a very low-level abstraction for appendable objects... you are defining public APIs that are direct mappings of the proto messages.

Ultimately, this is your call, but I think you want to raise the level of abstraction.  Maybe think about what logical operations the application would want to do (I think "open an object for exclusive appends" and "append to an open object", and maybe "reopen an object for appends").  Implement those abstractions and hide the details from your customers.

FYI: @joshuatants

vsharonlynn:
> This points to a very low-level abstraction for appendable objects... you are defining public APIs that are direct mappings of the proto messages.
> 
> Ultimately, this is your call, but I think you want to raise the level of abstraction. Maybe think about what logical operations the application would want to do (I think "open an object for exclusive appends" and "append to an open object", and maybe "reopen an object for appends"). Implement those abstractions and hide the details from your customers.
> 
> FYI: @joshuatants

Thanks for your input Carlos! Sounds like a good idea. We're actively looking into this. I'm currently drafting the APIs to see how it looks. I'll get back soon.

src/storage/src/model_ext.rs
```diff
@@ -310,6 +310,61 @@ pub struct WriteObjectRequest {
...
 
+#[cfg(google_cloud_unstable_storage_bidi)]
+/// Represents the parameters of a request to open a new object for exclusive appends.
+///
+/// This type might be used in mocks of the `Storage` client.
+#[derive(Debug, PartialEq)]
+#[non_exhaustive]
+pub struct OpenAppendableObjectRequest {
+    /// The object attributes and pre-conditions for the open operation.
+    pub spec: Box<crate::model::WriteObjectSpec>,
```
coryan: Unclear why you are using `Box<T>` here, but that is not enough reason to block the change.

src/storage/src/model_ext.rs
```diff
@@ -310,6 +310,61 @@ pub struct WriteObjectRequest {
     pub params: Option<crate::model::CommonObjectRequestParams>,
 }
 
+#[cfg(google_cloud_unstable_storage_bidi)]
+/// Represents the parameters of a request to open a new object for exclusive appends.
+///
+/// This type might be used in mocks of the `Storage` client.
```
coryan: nit:

```suggestion
/// Consumers of the `google-cloud-storage` crate rarely have a need to use this type directly, the most common exception is when mocking of the `Storage` client.
```

src/storage/src/model_ext.rs
```diff
@@ -310,6 +310,61 @@ pub struct WriteObjectRequest {
...
+    /// The object attributes and pre-conditions for the open operation.
+    pub spec: Box<crate::model::WriteObjectSpec>,
+    /// Additional request parameters.
+    pub params: Option<crate::model::CommonObjectRequestParams>,
+}
+
+#[cfg(google_cloud_unstable_storage_bidi)]
+/// Represents the parameters of a request to reopen an existing object for appends.
+///
+/// This type might be used in mocks of the `Storage` client.
```
coryan: same nit:

```suggestion
/// Consumers of the `google-cloud-storage` crate rarely have a need to use this type directly, the most common exception is when mocking of the `Storage` client.
```

coryan (approved):
Consider the nits and then merge.

joshuatants:
Looks good, thank you @vsharonlynn . Have a minor nit but we can discuss that offline.

joshuatants (approved):

src/storage/src/model_ext.rs
vsharonlynn: Cleaned it up. Thank you! Honestly, the only reason is my current Rust illiteracy 😅 Still getting the hang of the language and completely missed that during modifications. Thanks for the grace.

PR #5655, opened 2026-05-12
feat(storage): send CRC32C checksum on final chunk of buffered resumable uploads

vsharonlynn:
Compute and send the CRC32C checksum on the final chunk of JSON resumable uploads with unknown size to prevent finalization on mismatch. Also includes a minor optimization to avoid unnecessary clones in checksum computation.


joshuatants:
Looking good. Are there tests?

vsharonlynn:
@joshuatants 

> Looking good. Are there tests?

Added.

PTAL. Thanks!

joshuatants (approved):

PR #5660, opened 2026-05-13
chore(deps): consolidate otel updates

dbolduc:
The real version of #5657

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -26,3 +26,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      opentelemetry-crates:
+        patterns:
+          - "*opentelemetry*"


.github/dependabot.yml
```diff
@@ -26,3 +26,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+    groups:
+      opentelemetry-crates:
+        patterns:
+          - "*opentelemetry*"
```
PhongChuong, line 32:
```suggestion
          - "opentelemetry*"
```
Nit.

dbolduc: I want to include `tracing-opentelemetry`

coryan, line 32: Would this work? Seems more controlled and still gives you want you want:

```suggestion
          - "opentelemetry*"
          - "tracing-opentelemetry*"
```

coryan (approved):

.github/dependabot.yml
dbolduc: > Would this work?

Yes, but I don't prefer it.

I think it's more likely that if we depend on some new `foo-opentelemetry` crate we want it to be updated in step with the others rather than on its own.

And if that's wrong, it's super easy to fix.

PR #5672, opened 2026-05-14
fix(storage): provide user project for requester pays samples

### LINKED ISSUE #3291: Tests run for storage samples for requester pays enabled require a user project and none is provided
See error from integration tests run:
```
Error: the service reports an error with code INVALID_ARGUMENT described as: Bucket is a requester pays bucket but no user project provided.
```

This happens when trying to get_requester_pays_status::sample is run and when disable_requester_pays_statys::sample is run.

These are being added to CI for #3288 and are being skipped.

Code location: https://github.com/googleapis/google-cloud-rust/blob/main/src/storage/examples/src/lib.rs

joshuatants:
Fix the  `get_requester_pays_status` and `disable_requester_pays` samples by adding a user project identifier. This allows the samples to successfully execute against a requester-pays bucket.

Fixes #3291.

### DIFF
--- a/src/storage/examples/src/lib.rs
+++ b/src/storage/examples/src/lib.rs
@@ -814,25 +814,6 @@ async fn delete_bucket_with_error_backoff(
     Ok(())
 }
 
-async fn disable_requester_pays_for_cleanup(client: &StorageControl, name: &str, project_id: &str) {
-    use google_cloud_storage::model::bucket::Billing;
-
-    if let Err(e) = client
-        .update_bucket()
-        .set_bucket(
-            Bucket::default()
-                .set_name(name)
-                .set_billing(Billing::default().set_requester_pays(false)),
-        )
-        .set_update_mask(FieldMask::default().set_paths(["billing.requester_pays"]))
-        .with_quota_project(project_id)
-        .send()
-        .await
-    {
-        tracing::error!("error disabling requester_pays for bucket {name}: {e:?}");
-    }
-}
-
 pub async fn cleanup_bucket(
     client: StorageControl,
     name: String,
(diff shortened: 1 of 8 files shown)


src/storage/examples/src/lib.rs
```diff
@@ -177,17 +177,34 @@ pub async fn run_bucket_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
...
+    // By capturing the result each call, we ensure that a best-effort
+    // `disable_requester_pays` is always executed before returning any errors,
+    // resetting the bucket so the cleanup routines can delete it.
+    // If the whole test process dies/panics, it will still leak, but this
+    // handles standard runtime or assertion errors much more gracefully.
+    tracing::info!("running enable_requester_pays example");
+    let enable_res = buckets::enable_requester_pays::sample(&client, &id).await;
+    if let Err(e) = &enable_res {
+        tracing::error!("Failed to enable requester pays: {:?}", e);
+    }
```
coryan: Here and below, consider:


```suggestion
    let enable_res = buckets::enable_requester_pays::sample(&client, &id).await
        .inspect_err(|e| tracing::error!("Failed to enable requester pays: {e:?}");
```

src/storage/examples/src/lib.rs
```diff
@@ -177,17 +177,34 @@ pub async fn run_bucket_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
...
-        tracing::info!("running get_requester_pays_status example");
-        buckets::get_requester_pays_status::sample(&client, &id).await?;
-        tracing::info!("running disable_requester_pays example");
-        buckets::disable_requester_pays::sample(&client, &id).await?;
+
+    // By capturing the result each call, we ensure that a best-effort
+    // `disable_requester_pays` is always executed before returning any errors,
+    // resetting the bucket so the cleanup routines can delete it.
+    // If the whole test process dies/panics, it will still leak, but this
+    // handles standard runtime or assertion errors much more gracefully.
```
coryan: Consider moving the call to disable_requester_pays to the caller of this function, then the code here can remain relatively simple.

src/storage/examples/src/lib.rs
joshuatants: No longer needed with new cleanup logic (see other comment).

src/storage/examples/src/lib.rs
joshuatants: I updated the general `cleanup_bucket` function so that if a bucket has `requester pays` enabled, it's disabled before the bucket is removed. The upside here is that we never have to think about this again, though the downside is that I had to update the call to `cleanup_bucket` in several places to pass in the `project_id`. WDYT?

src/storage/examples/src/lib.rs
```diff
@@ -821,17 +814,57 @@ async fn delete_bucket_with_error_backoff(
...
+    let current = client
+        .get_bucket()
+        .set_name(name)
+        .with_quota_project(project_id)
+        .send()
+        .await?;
+
+    if current.billing.as_ref().is_some_and(|b| b.requester_pays) {
+        disable_requester_pays_for_cleanup(client, name, project_id).await;
+    }
```
coryan: You could update requester_pays and the label in a single RPC, that uses less quota.

coryan (approved):

src/storage/examples/src/lib.rs
joshuatants: Done, thank you.

PR #5675, opened 2026-05-14
cleanup: remove unused dependencies

alvarowolfx:
Used `cargo machete` tool to detect unused deps and manually resolved some false positives.

Command use:  `cargo machete --with-metadata --fix .`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4210,6 +4210,7 @@ dependencies = [
  "google-cloud-gax",
  "google-cloud-gax-internal",
  "google-cloud-iam-v1",
+ "google-cloud-pubsub",
  "google-cloud-rpc",
  "google-cloud-test-macros",
  "google-cloud-wkt",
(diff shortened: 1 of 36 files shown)


haphungw (approved):

PhongChuong (approved):
Thanks for this.
I was wondering about using this vs `unused_crate_dependencies` but it seems like they both have false positives and `cargo machete` seems to be better as we can define `package.metadata.cargo-machete`.

You probably already know but we should add this check to the ci.

alvarowolfx:
> Thanks for this. I was wondering about using this vs `unused_crate_dependencies` but it seems like they both have false positives and `cargo machete` seems to be better as we can define `package.metadata.cargo-machete`.
> 
> You probably already know but we should add this check to the ci.

@PhongChuong  I was wondering about adding as a CI check, but we will have to think how to exclude generated libraries (cargo machete don't have many options for filtering). Maybe we need a `package.metadata.cargo-machete` section on generated libs.

PhongChuong:
@alvarowolfx , ci sounds feasible with some work. Lets merge and discuss off thread.

PR #5676, opened 2026-05-14
feat(spanner): leader-aware routing

olavloite:
Adds support for leader-aware routing to the Spanner client. This requires a custom header to be included in the requests that should be routed to the leader.

This change also includes a small change to gaxi to support setting custom headers.

### DIFF
--- a/src/gax-internal/src/grpc.rs
+++ b/src/gax-internal/src/grpc.rs
@@ -557,6 +557,8 @@ impl Client {
         request_params: &str,
         options: &RequestOptions,
     ) -> Result<http::header::HeaderMap> {
+        use google_cloud_gax::options::internal::RequestOptionsExt as _;
+
         let mut headers = HeaderMap::new();
         if let Some(user_agent) = options.user_agent() {
             headers.insert(
@@ -585,6 +587,9 @@ impl Client {
                 http::header::HeaderName::from_static("x-goog-request-params"),
                 http::header::HeaderValue::from_str(request_params).map_err(Error::ser)?,
             );
+        }
+        if let Some(custom_headers) = options.get_extension::<HeaderMap>() {
+            headers.extend(custom_headers.clone());
         }
         Ok(headers)
     }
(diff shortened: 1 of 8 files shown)


src/spanner/src/database_client.rs
```diff
@@ -50,6 +50,7 @@ use std::sync::Arc;
 pub struct DatabaseClient {
     pub(crate) spanner: Spanner,
     pub(crate) session_maintainer: Arc<ManagedSessionMaintainer>,
+    pub(crate) leader_aware_routing_enabled: bool,
```
sakthivelmanii, line 53: Is there any usecase where customer needs to disable LAR?

olavloite: Yes, there are use-cases where disabling it can give you a small performance benefit: https://docs.cloud.google.com/spanner/docs/leader-aware-routing#limitations

src/spanner/src/read_write_transaction.rs
```diff
@@ -551,6 +549,9 @@ impl ReadWriteTransaction {
             };
             options.set_retry_policy(bounded_policy);
         }
+        if self.context.client.leader_aware_routing_enabled {
+            *options = options.clone().insert_extension((*LAR_HEADER_MAP).clone());
+        }
```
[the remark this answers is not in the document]

src/gax-internal/src/grpc.rs
```diff
@@ -717,4 +722,24 @@ mod tests {
         // We can't easily assert the internal state without exposing more internals,
         // but this verifies the method exists and runs.
     }
+
+    #[tokio::test]
+    async fn test_make_headers_with_custom_headers() {
```
dbolduc, line 727: tl;dr: Testing the feature here works for me. No action required.

---

Here is more background if you are curious...

We have been putting tests for observable behavior (e.g. adding headers) of the gRPC / HTTP clients here: https://github.com/olavloite/google-cloud-rust/tree/3d482798776bb62bb111646ed61508e93e5cffa3/src/gax-internal/tests

But I think it is getting unruly with one binary for each header key, with repetitive code in each.

I thing the existing tests + clippy's dead code detection gives us enough confidence that `make_headers()` is called. So I am satisfied with the coverage.

src/gax-internal/src/grpc.rs
```diff
@@ -586,6 +588,9 @@ impl Client {
                 http::header::HeaderValue::from_str(request_params).map_err(Error::ser)?,
             );
         }
+        if let Some(custom_headers) = options.get_extension::<HeaderMap>() {
```
dbolduc, line 591: comment: seems like the custom header for LAR is known at client creation time. So we could have made it a `ClientConfig` option.

I prefer your approach of per-request custom headers though. It is more general and we will need it in the future, if we ever write a Bigtable client.

dbolduc (approved):
Approved the `gax` changes. I did not look at the changes under `src/spanner`

Note to self: I should add custom header support to our HTTP client so it maintains feature parity with the gRPC client.

src/gax-internal/src/grpc.rs
olavloite: We might also need it in the future for Spanner, as it could be that we need to support enabling/disabling this feature on a per-transaction basis.

src/spanner/src/database_client.rs
olavloite: @sakthivelmanii Would you mind reviewing the Spanner-specific part of this PR (see also https://github.com/googleapis/google-cloud-rust/pull/5676#pullrequestreview-4299634793)

sakthivelmanii (approved):

Issue #5592, opened 2026-05-05
Implement bigquery-v2 crate


dbolduc:
We need to offer a client surface for the `google/cloud/bigquery/v2` APIs. We want it to include fully generated admin clients (generated from the proto spec), where the Job client has an extension to treat creating / awaiting jobs as an LRO.

This warrants a design doc, as there are several approaches we should consider (listed in order of preference):
- Leveraging common paths, such as the `discovery` poller.
- Manually defining extra modules and extending the `InsertJobBuilder`
- Teaching the generator how to recognize Jobs in bigquery as LROs.

When this is done, we should have at least one integration test creating a job.

Note that we currently generate this crate. We can flip the `publish` bit to `true` for this crate without having extended the GAPIC. We were conservatively making sure we wanted to expose this generated surface vs. cover it in a veneer.

Googlers can see: go/cloud-rust:bigquery-high-level-design

---

cc: @haphungw

abhinavgautam01:
Hi, I’d like to contribute to this issue...

I reviewed the repo structure and the current BigQuery code and I see that `google-cloud-bigquery-v2` already exists in `src/generated/cloud/bigquery/v2`, is currently not released, and there are already integration tests around `JobService::insert_job()`. From the issue description, it looks like the main work is deciding how the BigQuery Jobs flow should be exposed, especially treating create/await as an LRO.

Before I start, could you help clarify the preferred direction for an external contribution?

1. Should this be implemented entirely in `google-cloud-rust`, for example by adding handwritten extensions around the generated `bigquery-v2` client?
2. Or is the preferred solution to change the generator in `googleapis/librarian` and then regenerate code here?
3. Is there any public design guidance you can share for the approach to use, since the linked design doc appears to be internal?

Thanks.

dbolduc:
> clarify the preferred direction for an external contribution?

My preferred direction is to have my colleague @haphungw take on this work instead of an external contributor. It requires too much specific knowledge, and the solution is not obvious (at least to me).

If you were looking for an open issue, maybe try #4466? I think I can assign it to you if you comment on it.

antriksh-9:
@dbolduc 
Hi, I would like to know if there is any open issue to work upon ?

dbolduc:
> Hi, I would like to know if there is any open issue to work upon ?

[...]

PR #5681, opened 2026-05-15
impl(bigquery): add query metadata enums

alvarowolfx:
Add Query metadata on query creation and query polling. Add a test method that uses AST to check if the implementation is missing fields from the raw responses.

Towards #5592

### DIFF
--- a/src/bigquery/src/query/metadata.rs
+++ b/src/bigquery/src/query/metadata.rs
@@ -22,6 +22,7 @@ use google_cloud_bigquery_v2::model::{
 /// Depending on how the query was initiated, this metadata originates from either a `jobs.query`
 /// response or a `jobs.insert` job resource.
 #[derive(Debug, Clone)]
+#[non_exhaustive]
 #[allow(clippy::large_enum_variant)]
 pub enum QueryCreationMetadata {
     /// Metadata originating from a `jobs.query` API call.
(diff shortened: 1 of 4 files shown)


src/bigquery/src/query/metadata.rs
```diff
@@ -0,0 +1,714 @@
...
+        }
+    }
+
+    /// Returns the auto-generated ID for the query.
+    ///
+    /// This field only exists in the `JobsQuery` branch.
+    pub fn query_id(&self) -> Option<&str> {
+        match self {
+            Self::JobsQuery(res) => Some(&res.query_id),
+            Self::JobsInsert(_) => None,
```
dbolduc, line 75: Why do we want to reproduce all possible fields? Isn't it enough to give the application an enum where they can match on the branches? (if we even want an enum)

alvarowolfx: I was leaning on discoverability of the fields. To your point, those accessors for attributes that are unique to each branch might not be that useful, since users that want a more specific field might know that they only exist on a given branch. But for fields that exists on both branches (let's take total_rows as an example), I feel like it would lead to same boilerplate code every time that users want to read that. Like having to repeat this code a lot:

```rust
let total_rows = match self {
    QueryMetadata::GetQueryResultsResponse(res) => res.total_rows,
    QueryMetadata::JobsQuery(res) => res.total_rows,
}
```

Issue #5644, opened 2026-05-11
feat(storage): include custom object metadata in ObjectHighlights


metebalci:
**Is your feature request related to a problem? Please describe.**

[`ObjectHighlights`](https://docs.rs/google-cloud-storage/1.12.0/google_cloud_storage/model_ext/struct.ObjectHighlights.html) returned by `Storage::read_object(...).send().await?.object()` omits the `metadata: HashMap<String, String>` field of [`Object`](https://docs.rs/google-cloud-storage/1.12.0/google_cloud_storage/model/struct.Object.html).
Reading per-object custom metadata requires a follow-up `StorageControl::get_object` RPC — doubling the GET count on workloads that key on per-object metadata, even though the metadata is already on the wire
alongside the body:

- **gRPC**: [`google.storage.v2.ReadObjectResponse`](https://github.com/googleapis/googleapis/blob/master/google/storage/v2/storage.proto) field 4 is `metadata: Object`, sent on the first response message of the stream.
- **REST**: [`GET /storage/v1/b/{bucket}/o/{object}?alt=media`](https://cloud.google.com/storage/docs/json_api/v1/objects/get) surfaces custom metadata as `x-goog-meta-*` HTTP response headers.

**Describe the solution you'd like**

Add `metadata: HashMap<String, String>` to `ObjectHighlights`. Additive on a `#[non_exhaustive]` struct, no API break. Optionally also surface `event_based_hold`, `temporary_hold`, `retention` so the response can drive retention/hold gating without a follow-up call.

**Describe alternatives you've considered**

Issuing the follow-up `StorageControl::get_object` per download — the current workaround the doc comment recommends. Works, but doubles the request count.

**Additional context**

`ObjectHighlights` was introduced in [#2626](https://github.com/googleapis/google-cloud-rust/issues/2626) / [#2685](https://github.com/googleapis/google-cloud-rust/issues/2685) as a deliberately curated subset, so this is asking to expand the subset, not redesign the API. The two other major-cloud Rust SDKs already expose object metadata on their download responses:

- AWS — [`GetObjectOutput::metadata`](https://docs.rs/aws-sdk-s3/latest/aws_sdk_s3/operation/get_object/struct.GetObjectOutput.html#structfield.metadata): `Option<HashMap<String, String>>`
- Azure — [`BlobClient::download` response properties](https://docs.rs/azure_storage_blob/latest/azure_storage_blob/) surface `x-ms-meta-*`

[...]

PR #5687, opened 2026-05-18
feat(storage): include custom object metadata in ObjectHighlights

metebalci:
Adds `metadata: HashMap<String, String>` to `ObjectHighlights`,
populated from the `x-goog-meta-*` HTTP response headers on
`read_object()`. The metadata is already on the wire alongside the
object body, so no additional `StorageControl::get_object` RPC is
needed.

`ObjectHighlights` is `#[non_exhaustive]`, so this is additive — no
public API break.

## Changes

- `model_ext.rs`: new `metadata` field on `ObjectHighlights`.
- `parse_http_response.rs`: extract `x-goog-meta-*` headers into the
  field, via a small private helper.
- `bidi/range_reader.rs`: propagate `metadata` from the source
  `Object` (the bidi path already has it in memory; no extra cost).
- Test fixtures in `non_resumable.rs` and `bidi/range_reader.rs`
  updated for the new field.

## Out of scope

The issue mentions optionally also surfacing `event_based_hold`,
`temporary_hold`, and `retention`. The two hold flags are not exposed
as HTTP headers on the download response — surfacing them from
`read_object()` would require an additional metadata GET, defeating
the purpose. Per-object retention may map to
`x-goog-object-lock-mode` / `x-goog-object-lock-retain-until-date`
headers, but this needs empirical confirmation against a real GCS
endpoint with a retention-locked object. Both can be follow-up PRs if
the maintainers want them.

## Test plan

- `cargo test -p google-cloud-storage` — 215 passed, 0 failed.
- `cargo clippy --all-features --no-deps -p google-cloud-storage -- -D missing_docs -D clippy::exhaustive_enums` (the `clippy-strict` content) — clean.
- `cargo clippy --profile=test -p google-cloud-storage --all-targets -- -D warnings` — clean.
- Extended the existing `read_object_metadata` integration-style test to round-trip two `x-goog-meta-*` headers.
- Added two focused unit tests for the new `headers_to_metadata` helper covering case-insensitive header names, non-meta header filtering, and the empty-headers case.

For #5644

### DIFF
--- a/src/storage/src/storage/read_object/parse_http_response.rs
+++ b/src/storage/src/storage/read_object/parse_http_response.rs
@@ -51,7 +51,19 @@ pub fn object_highlights(generation: i64, headers: &HeaderMap) -> Result<ObjectH
                 .set_or_clear_crc32c(headers_to_crc32c(headers))
                 .set_md5_hash(headers_to_md5_hash(headers))
         }),
+        metadata: headers_to_metadata(headers),
     })
+}
+
+fn headers_to_metadata(headers: &HeaderMap) -> std::collections::HashMap<String, String> {
+    headers
+        .iter()
+        .filter_map(|(name, value)| {
+            let key = name.as_str().strip_prefix("x-goog-meta-")?;
+            let value = value.to_str().ok()?;
+            Some((key.to_string(), value.to_string()))
+        })
+        .collect()
 }
 
 pub(crate) fn headers_to_crc32c(headers: &HeaderMap) -> Option<u32> {
@@ -133,7 +145,9 @@ mod tests {
                     .append_header("content-language", "en")
                     .append_header("content-type", "text/plain")
                     .append_header("content-disposition", "inline")
-                    .append_header("etag", "etagval"),
+                    .append_header("etag", "etagval")
+                    .append_header("x-goog-meta-foo", "bar")
+                    .append_header("x-goog-meta-key2", "value2"),
             ),
         );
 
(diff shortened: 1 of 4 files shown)


joshuatants (approved):

src/storage/src/storage/read_object/parse_http_response.rs
```diff
@@ -51,9 +51,24 @@ pub fn object_highlights(generation: i64, headers: &HeaderMap) -> Result<ObjectH
...
+    let mut metadata = std::collections::HashMap::new();
+    for (name, value) in headers.iter() {
+        if let Some(key) = name.as_str().strip_prefix("x-goog-meta-") {
+            if !metadata.contains_key(key) {
+                if let Ok(val_str) = value.to_str() {
+                    metadata.insert(key.to_string(), val_str.to_string());
+                }
+            }
+        }
+    }
```
coryan, line 66: Consider:

```suggestion
    for (n, v) in headers.iter() {
        match (n.as_str().strip_prefix("x-goog-meta-"), v.to_str()) {
          (Some(key), Ok(val)) => { let _ = metadata.try_insert(key.to_string(), val.to_string()); },
          _ => {},
    }
```

coryan (approved):
A small nit for @joshuatants to consider.

src/storage/src/storage/read_object/parse_http_response.rs
joshuatants: Nice, though I think `try_insert` is ATM still a nightly-only experimental API. Perhaps 

```
metadata
.entry(key.to_string())
.or_insert_with(|| val_str.to_string());
```

would be better?

joshuatants: @metebalci Minor nit: why not, as coryan@ suggested, combine the two 'if let' statements into a single match?

` match (n.as_str().strip_prefix("x-goog-meta-"), v.to_str()) `

metebalci: @joshuatants clippy would complain:

```rust
error: you seem to be trying to use `match` for destructuring a single pattern.
  Consider using `if let`
    --> src/storage/src/storage/read_object/parse_http_response.rs:61:9
     |
  61 | /         match (name.as_str().strip_prefix("x-goog-meta-"), value.to_str()) {
  62 | |             (Some(key), Ok(val)) => {
  63 | |                 metadata
  64 | |                     .entry(key.to_string())
  ...  |
  67 | |             _ => {}
  68 | |         }
     | |_________^
     |
     = help: for further information visit
  https://rust-lang.github.io/rust-clippy/rust-1.95.0/index.html#single_match
     = note: `-D clippy::single-match` implied by `-D warnings`
     = help: to override `-D warnings` add `#[allow(clippy::single_match)]`
  help: try
     |
  61 ~         if let (Some(key), Ok(val)) =
  (name.as_str().strip_prefix("x-goog-meta-"), value.to_str()) {
  62 +             metadata
  63 +                 .entry(key.to_string())
  64 +                 .or_insert_with(|| val.to_string());
  65 +         }
     |
```

but I can collapse nested if:

```rust
  for (name, value) in headers.iter() {
      if let (Some(key), Ok(val)) =
          (name.as_str().strip_prefix("x-goog-meta-"), value.to_str())
      {
          metadata.entry(key.to_string()).or_insert_with(|| val.to_string());
      }
  }
```

coryan: > but I can collapse nested if:

Please do.

coryan (approved):

PR #5694, opened 2026-05-19
impl(o11y): add operation name support to `Poller` interface

haphungw:
Allow tracing decorators to dynamically query LRO operation IDs from underlying poller for telemetry recording.

Because `PollerImpl` is encapsulated as a generic inside the telemetry decorator, we can't do direct field access (like `self.inner.operation`) on the generic type. Declaring this standard method on the `Poller` trait supports LRO tracing attribute propagation without contaminating public structs.

### DIFF
--- a/src/lro/src/lib.rs
+++ b/src/lro/src/lib.rs
@@ -149,6 +149,9 @@ pub trait Poller<ResponseType, MetadataType>: Send + sealed::Poller {
         &mut self,
     ) -> impl Future<Output = Option<PollingResult<ResponseType, MetadataType>>> + Send;
 
+    /// Returns the name of the operation, if known.
+    fn operation_name(&self) -> Option<&str>;
+
     /// Sleep until the backoff time has elapsed.
     fn backoff(&mut self, state: &PollingState) -> impl Future<Output = ()> + Send;
 
--- a/src/lro/src/internal/aip151.rs
+++ b/src/lro/src/internal/aip151.rs
@@ -132,6 +132,9 @@ where
     async fn poll(&mut self) -> Option<PollingResult<(), M>> {
         self.poller.poll().await.map(self::map_polling_result)
     }
+    fn operation_name(&self) -> Option<&str> {
+        self.poller.operation_name()
+    }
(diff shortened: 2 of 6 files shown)


src/lro/src/lib.rs
```diff
@@ -149,6 +149,9 @@ pub trait Poller<ResponseType, MetadataType>: Send + sealed::Poller {
         &mut self,
     ) -> impl Future<Output = Option<PollingResult<ResponseType, MetadataType>>> + Send;
 
+    /// Returns the name of the operation, if known.
+    fn operation_name(&self) -> Option<String>;
```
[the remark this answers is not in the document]
haphungw: ack, done

src/lro/src/lib.rs
```diff
@@ -149,6 +149,9 @@ pub trait Poller<ResponseType, MetadataType>: Send + sealed::Poller {
         &mut self,
     ) -> impl Future<Output = Option<PollingResult<ResponseType, MetadataType>>> + Send;
 
+    /// Returns the name of the operation, if known.
+    fn operation_name(&self) -> Option<&str>;
```
coryan, line 153: This is fine. It is a new public API, but not a big deal.  But I just realized that `backoff()` is also public, and I wonder if we should have not made it so...

coryan (approved):

src/lro/src/lib.rs
haphungw: I found an alternative to avoid touching the public API: https://github.com/googleapis/google-cloud-rust/pull/5695. 

Instead of adding `operation_name()` to query the inner poller, `Tracing<P>::poll` now enters the T2 span scope before delegating the poll call. The T2 span is now active in the thread context, and the pollers can record the operation name directly onto it using [tracing::Span::current().record()](https://docs.rs/tracing/latest/tracing/struct.Span.html#method.current).

If you think this works, I can go ahead and close this PR.

haphungw:
replaced by https://github.com/googleapis/google-cloud-rust/pull/5695.

PR #5695, opened 2026-05-19
impl(o11y): record telemetry attributes on LRO span

haphungw:
In our previous iteration (https://github.com/googleapis/google-cloud-rust/pull/5694), we introduced an `operation_name(&self)` query method to the public `Poller` trait.

To avoid polluting the public API, we utilize thread-local span propagation: because the LRO span is active in the thread context, the internal pollers (`PollerImpl` and `DiscoveryPoller`) record the LRO operation name and destination resource ID directly onto the current active span using [tracing::Span::current().record()](https://docs.rs/tracing/latest/tracing/struct.Span.html#method.current).


src/lro/src/internal/tracing.rs
```diff
@@ -53,7 +53,6 @@ where
     fn into_stream(
         self,
     ) -> impl futures::Stream<Item = PollingResult<ResponseType, MetadataType>> + Unpin {
-        let span = self.span.clone();
-        crate::into_stream(self).instrument(span)
+        crate::into_stream(self)
```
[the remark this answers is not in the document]
haphungw: irrelevant now

src/lro/src/internal/aip151.rs
```diff
@@ -347,7 +389,7 @@ mod tests {
             start,
             query,
         );
-        let p0 = poller.poll().await;
+        let p0 = poller.poll().instrument(test_span()).await;
```
dbolduc, line 407: Shouldn't we verify the span contents?

Aside: consider writing new tests just for tracing. Often it is quickest to use existing tests to test a new feature... but when you do that enough times the tests get complicated and lose their focus.

haphungw: ack. I added unit tests for both aip151 and discovery to verify the 2 specific attributes we populate for LROs.

src/lro/src/internal/aip151.rs
```diff
@@ -268,6 +268,12 @@ where
         if let Some(start) = self.start.take() {
             let result = start().await;
             let (op, poll) = crate::details::handle_start(result);
+            #[cfg(google_cloud_unstable_tracing)]
+            if let Some(ref name) = op {
+                let span = tracing::Span::current();
+                span.record("gcp.longrunning.operation_name", name);
+                span.record("gcp.resource.destination.id", name);
```
westarle: are these ever different?

src/lro/src/internal/aip151.rs
```diff
@@ -268,6 +268,12 @@ where
         if let Some(start) = self.start.take() {
             let result = start().await;
             let (op, poll) = crate::details::handle_start(result);
```
westarle, line 281: Is it possible that the operation is marked done inside the start()?  If so we might want to extract the name from the result above handle_start?

src/lro/src/internal/aip151.rs
```diff
@@ -268,6 +268,12 @@ where
...
+            #[cfg(google_cloud_unstable_tracing)]
+            if let Some(ref name) = op {
+                let span = tracing::Span::current();
```
westarle: if tracing is disabled, this might interfere with the customers' spans.  Can we attach the span directly to the PollerImpl?

src/lro/src/internal/aip151.rs
haphungw: in the final design we decided that they're always the same. we expected that downstream tools (e.g., AppHub) will use `Operation ID` as the default.

src/lro/src/internal/aip151.rs
haphungw: That's a good point. I fixed the loops to extract the name directly from `result` (if `Ok`) before calling `handle_start`. Now we can make sure we don't miss these attributes.

I added a test for this case.

src/lro/src/internal/aip151.rs
haphungw: Yeah you're right. I considered a few approaches for context propagation: 
- adding `operation_name()` as a trait method (https://github.com/googleapis/google-cloud-rust/pull/5694): this pollutes our public API and intermediate decorators.
- direct span mutation (what you were looking at): this might interfere with customers' spans.
- `tokio::task_local!` (the newer commits).

We also used task-local back then for `gaxi` `RequestRecorder`, so I guess it's okay to adopt this.

haphungw: Regarding attaching the span directly to `PollerImpl`: I wanted to avoid adding a tracing field to `PollerImpl` because that forces our polling struct to actively manage and carry o11y variables even when tracing is disabled.

src/lro/src/internal/aip151.rs
westarle: thanks!

src/gax-internal/src/observability/client_signals.rs
```diff
@@ -117,6 +117,10 @@ macro_rules! client_request_signals {
             "http.request.resend_count"   = Empty,
             "http.response.status_code"   = Empty,
             "gcp.resource.destination.id" = Empty,
+            // Pre-declare LRO metrics fields to prevent tracing from silently ignoring them during runtime record()
```
westarle: I don't think this comment is necessary (it's the case for the rows above too)

src/lro/src/internal/discovery.rs
```diff
@@ -524,4 +569,131 @@ mod tests {
             self.name.as_ref()
         }
     }
+
+    #[cfg(google_cloud_unstable_tracing)]
+    struct TestLayer {
```
westarle: is it possible to use TestLayer in test-utils?

https://github.com/googleapis/google-cloud-rust/blob/7ca22629fe1919437f5326e11267d61e8718e55b/src/test-utils/src/test_layer.rs#L219-L260

src/lro/src/internal/tracing.rs
```diff
@@ -45,19 +75,171 @@ where
...
+        {
+            let attempt = if self.started {
+                self.poll_attempt_count += 1;
+                self.poll_attempt_count
+            } else {
+                self.started = true;
+                0 // Initial triggers record nothing
+            };
+
+            LRO_SPAN
```
westarle: would it make sense to combine into a single "LRO Recorder" that captures the LRO-level span and poll-attempt count?  You might be able to make it a struct so that it encapsulates the use of task-locals.

Only consider if it simplifies the code.

src/lro/src/internal/discovery.rs
haphungw: ooh, thanks

src/gax-internal/src/observability/client_signals.rs
haphungw: done

src/lro/src/internal/tracing.rs
haphungw: I think it does make the code looks more neat.

src/lro/src/internal/tracing.rs
```diff
@@ -12,20 +12,93 @@
...
 use google_cloud_gax::polling_state::PollingState;
 use tracing::{Instrument, Span, info_span};
 
+#[cfg(google_cloud_unstable_tracing)]
+use crate::POLL_ATTEMPT_COUNT;
+
+#[cfg(google_cloud_unstable_tracing)]
+tokio::task_local! {
+    pub(crate) static LRO_SPAN: Span;
+}
```
westarle, line 36: can these be unexported and use the LroRecorder API instead of LRO_SPAN directly?

haphungw: you're right, on it right now

src/gax-internal/src/observability/client_signals.rs
```diff
@@ -104,7 +104,6 @@ macro_rules! client_request_signals {
              "gcp.client.version"    = $info.client_version,
              "gcp.schema.url"        = SCHEMA_URL_VALUE,
              "otel.status_code"      = "UNSET",
-            // Fields to be recorded later
```
westarle, line 107: keep this comment

src/gax-internal/src/observability/client_signals.rs
```diff
@@ -117,6 +116,9 @@ macro_rules! client_request_signals {
             "http.request.resend_count"   = Empty,
             "http.response.status_code"   = Empty,
             "gcp.resource.destination.id" = Empty,
+            "gcp.longrunning.poll_attempt_count" = Empty,
+            "gcp.longrunning.done"               = Empty,
+            "gcp.longrunning.status_code"        = Empty,
```
westarle: Nit: Align these with the rows above.

src/lro/src/lib.rs
```diff
@@ -132,6 +132,11 @@ pub mod internal;
 #[cfg(google_cloud_unstable_tracing)]
 pub use internal::{PollerOptions, TracingDetails};
 
+#[cfg(google_cloud_unstable_tracing)]
+tokio::task_local! {
+    pub static POLL_ATTEMPT_COUNT: u32;
```
westarle, line 138: can this be pub(crate)?

src/lro/src/internal/tracing.rs
```diff
@@ -12,20 +12,118 @@
...
+#[cfg(google_cloud_unstable_tracing)]
+tokio::task_local! {
+    static LRO_RECORDER: LroRecorder;
+}
+
+#[cfg(google_cloud_unstable_tracing)]
+#[derive(Clone, Debug)]
+pub(crate) struct LroRecorder {
+    span: Span,
+    poll_attempt_count: u32,
```
westarle: I'm a little confused about how this and POLL_ATTEMPT_COUNT work together.  It might need a few comments, or maybe there's a way to simplify and consolidate to POLL_ATTEMPT_COUNT if there's a good way to use that instead of this.

A worry I have is if LroRecorder is cloned (in scope()) and then poll_attempt_count is mutated on the clone...

src/lro/src/internal/tracing.rs
```diff
@@ -12,20 +12,118 @@
...
+use crate::{Poller, PollingResult, Result, sealed};
 use google_cloud_gax::polling_state::PollingState;
 use tracing::{Instrument, Span, info_span};
 
+#[cfg(google_cloud_unstable_tracing)]
+use crate::POLL_ATTEMPT_COUNT;
+
+#[cfg(google_cloud_unstable_tracing)]
+tokio::task_local! {
+    static LRO_SPAN: Span;
```
westarle: is this used?

src/lro/src/internal/tracing.rs
haphungw: Yeah you're right, I did not consider the risk of cloning mutable struct. I was thinking that the mutation only occurred inside `record_poll`.

My solution is to make `LroRecorder` stateless by leaving it to wrap the readonly `Span` and no mutable fields, so no divergent state.

We move the stateful `poll_attempt_count` and `started` boolean to the `Tracing` decorator. The decorator's `poll` method takes `&mut self` so we are sure that `poll_attempt_count` can only be accessed by one flow. Then we copy `poll_attempt_count` into the `POLL_ATTEMPT_COUNT` task-local key only for the lifetime of the active poll future.

src/lro/src/internal/tracing.rs
haphungw: Yes. `LRO_SPAN` is to support both `aip151.rs` and `discovery.rs`.

Operation name is only discovered at runtime after a poll starts, so we need a way for pollers in both `aip151.rs` and `discovery.rs` to record it.

To avoid adding tracing parameters to the `Poller` trait signature, the `Tracing` decorator wraps the poll future inside `LRO_SPAN` scope, so both pollers can retrieve the active span across module boundaries and record operation name.

src/lro/src/lib.rs
haphungw: done

src/gax-internal/src/observability/client_signals.rs
haphungw: done

src/gax-internal/src/observability/client_signals.rs
haphungw: done

src/lro/src/lib.rs
haphungw: update:
Previously the `google_cloud_unstable_tracing` checks were missing in CI, so the build failure was missed.

Because the generated crates reference `POLL_ATTEMPT_COUNT`, it must be `pub` to compile. I've reverted it but added `#[doc(hidden)]` so it stays out of public API docs.

coryan: You could move it to `internal`, though that requires changing the generated code too...

Also, I find it odd that we need two `task_local` variables, can we combine them?

coryan (approved):
Feel free to merge so we can fix the CI and fix any issues in follow-up PRs.

haphungw:
I'm merging this PR to unblock CI, and address some issues in follow-ups:

1. Fix the unstable tracing compile error in `compute` with the `DiscoveryOperation` trait helper for Discovery-based LROs.
2. Refactor task-local by combining `POLL_ATTEMPT_COUNT` and `LRO_RECORDER` into a single unified task-local context inside `LroRecorder`. Attempt count should also be private, and we can expose `google_cloud_lro::poll_attempt_count()` to keep task-locals hidden from generated client APIs.

src/lro/src/lib.rs
haphungw: I fixed it here @ https://github.com/googleapis/google-cloud-rust/pull/5803

PR #5698, opened 2026-05-20
chore(spanner): directly execute queries and wait for first result

### LINKED ISSUE #5673: spanner: dropping a ResultSet for the first query in a transaction can lead to a deadlock
If the ResultSet that is returned by the first query in a transaction is dropped by the application before reading any data, it can lead to a deadlock, as the ResultSet will never process the transaction ID that is returned as part of the metadata of the query.

### LINKED ISSUE #5685: spanner: re-evaluate the design choice for execute_query not fetching the first PartialResultSet
The `execute_query` method in the Spanner client invokes the `ExecuteStreamingSql` RPC in Spanner. This initiates the gRPC stream that will receive the query results from Spanner. However, Spanner does not evaluate the query at all before setting up this stream, meaning that even the simplest syntax error will be delayed until the first time the application calls `ResultSet#next()`.

Possible alternatives:
- Keep it as it is now.
- Delay the initialization of the stream until the first time `ResultSet#next()` is called.
- Eagerly fetch the first `PartialResultSet` when `execute_streaming` is called to ensure that `execute_query` fully validates and executes the query.

The Spanner emulator __does__ evaluate the query when `ExecuteStreamingSql` is invoked. This currently causes a behavioral difference when using the Rust client with the Emulator vs with real Spanner.

olavloite:
This changes the way that queries are executed on Spanner from a model that uses a background task to read from the query stream, to a model that directly calls ExecuteStreamingSql when execute_query is called, and then waits for the first PartialResultSet or error to be received. This ensures that execute_query really executes the query and returns any errors that happens during query execution.

Closes #5685
Fixes #5673


tests/spanner/src/client.rs
```diff
@@ -23,6 +23,7 @@ const PROJECT_ID: &str = "test-project";
 const INSTANCE_ID: &str = "test-instance";
 
 pub fn get_emulator_host() -> Option<String> {
+    let _ = rustls::crypto::ring::default_provider().install_default();
```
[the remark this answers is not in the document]
olavloite: Done

src/spanner/src/result_set.rs
```diff
@@ -208,38 +161,84 @@ impl ResultSet {
...
+                if !is_started {
+                    return Err(internal_error(
+                        "Spanner failed to return a transaction ID for a query that included a BeginTransaction option",
+                    ));
+                }
+            }
+        }
+        self.local_metadata = Some(meta);
+        Ok(())
+    }
```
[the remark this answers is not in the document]

tests/spanner/Cargo.toml
```diff
@@ -38,6 +38,7 @@ prost                   = { workspace = true }
 prost-types.workspace   = true
 rand                    = { workspace = true }
 reqwest                 = { workspace = true, features = ["json"] }
+rustls                  = { workspace = true, features = ["ring"] }
```
olavloite: This change is not directly related to this change, but a change that is needed due to the dependency setups overall for the project.

coryan: Maybe look at [go/cloud-rust:prune-crypto-algos](https://goto.google.com/cloud-rust:prune-crypto-algos) and consider a similar solution to what we do in storage or pubsub:

https://github.com/googleapis/google-cloud-rust/blob/eeb17aa189f6357cd5c03ff6d1d80ad3f409cf84/tests/pubsub/Cargo.toml#L34

and

https://github.com/googleapis/google-cloud-rust/blob/eeb17aa189f6357cd5c03ff6d1d80ad3f409cf84/src/pubsub/Cargo.toml#L53-L59

olavloite: Thanks, done.

sakthivelmanii (approved):

PR #5701, opened 2026-05-20
chore(spanner): use Into<BatchDml> as input argument

olavloite:
Use Into<BatchDml> as the input argument for execute_batch_update. This allows an application to pass in a collection of statements, instead of having to manually construct a BatchDml struct.

### DIFF
--- a/src/spanner/src/read_write_transaction.rs
+++ b/src/spanner/src/read_write_transaction.rs
@@ -436,8 +436,11 @@ impl ReadWriteTransaction {
     /// # Ok(())
     /// # }
     /// ```
-    pub async fn execute_batch_update(&self, batch: BatchDml) -> crate::Result<Vec<i64>> {
-        let mut batch = batch;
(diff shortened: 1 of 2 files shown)


src/spanner/src/read_write_transaction.rs
```diff
@@ -1127,6 +1130,86 @@ mod tests {
         run_read_write_transaction_execute_batch_update(BeginTransactionOption::InlineBegin).await
     }
 
+    #[tokio_test_no_panics]
+    async fn read_write_transaction_execute_batch_update_vec() -> anyhow::Result<()> {
```
PhongChuong, line 1147: nit. Is it possible to reuse `run_read_write_transaction_execute_batch_update` in this test?

PhongChuong (approved):

src/spanner/src/read_write_transaction.rs
olavloite: Done

PR #5720, opened 2026-05-21
test(spanner): add panic handler to mock Spanner server

### LINKED ISSUE #5523: Stricter mock server assertions in spanner tests
Spanner tests using the `MockSpanner` server should use the `#[tokio_test_no_panics]` attribute macro to make sure there are no panics in the background task running the server.

https://github.com/googleapis/google-cloud-rust/blob/ca4ea8a22835303c522ab3cf68142ece9dfb59fa/src/pubsub/src/subscriber/message_stream.rs#L531

More details in: https://github.com/googleapis/google-cloud-rust/issues/4733

olavloite:
Adds a panic handler to the Spanner mock server that is used in tests. This removes the requirement to add the `tokio_test_no_panics` attribute to tests that use this server.

This is an alternative for #5718

Fixes #5523

### DIFF
--- a/.clippy.toml
+++ b/.clippy.toml
@@ -14,3 +14,7 @@
 
 # We need to support 1.87. Stop clippy from issuing newer warnings.
 msrv = "1.87.0"
+
+disallowed-methods = [
+    "spanner_grpc_mock::start"
+]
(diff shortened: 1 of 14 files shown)


olavloite:
Closing in favor of #5718, as this alternative would also require code to be added for each test, which defies the entire purpose of exploring this alternative.

PR #5726, opened 2026-05-23
chore(spanner): add default TLS provider

olavloite:
Adds a default TLS provider to Spanner in the same way as the other clients.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4691,6 +4691,7 @@ dependencies = [
  "google-cloud-gax",
  "google-cloud-gax-internal",
  "google-cloud-rpc",
+ "google-cloud-spanner",
  "google-cloud-test-macros",
  "google-cloud-wkt",
  "http",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4691,7 +4691,6 @@ dependencies = [
  "google-cloud-gax",
  "google-cloud-gax-internal",
  "google-cloud-rpc",
- "google-cloud-spanner",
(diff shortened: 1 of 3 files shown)


Cargo.lock
```diff
@@ -4691,6 +4691,7 @@ dependencies = [
  "google-cloud-gax",
  "google-cloud-gax-internal",
  "google-cloud-rpc",
+ "google-cloud-spanner",
```
sakthivelmanii: why are we adding `google-cloud-spanner` as dependency for package `google-cloud-spanner`

src/spanner/Cargo.toml
```diff
@@ -54,6 +60,7 @@ wkt                   = { workspace = true, features = ["time"] }
 
 [dev-dependencies]
 anyhow.workspace                   = true
+google-cloud-spanner               = { path = ".", features = ["default-rustls-provider"] }
```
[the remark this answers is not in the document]
olavloite: Removed

Cargo.lock
olavloite: This is (was) a side-effect of this: https://github.com/googleapis/google-cloud-rust/pull/5726#discussion_r3296341932

And that again was an earlier attempt at adding a default TLS provider, and that was not fully removed after implementing it in the same way as for the other clients in this project. It has now been removed.

sakthivelmanii (approved):

alvarowolfx (approved):

PR #5727, opened 2026-05-24
test(spanner): add more integration tests for read/write transactions

olavloite:
Adds more integration tests for read/write transactions, and fixes a bug where some commit options were not applied to the retry request if the original commit attempt returned a CommitResponse with a new precommit_token.

### DIFF
--- a/src/spanner/src/write_only_transaction.rs
+++ b/src/spanner/src/write_only_transaction.rs
@@ -16,7 +16,8 @@ use crate::client::{DatabaseClient, Mutation, amend_request_options_for_lar};
 use crate::model::request_options::Priority;
 use crate::model::transaction_options::ReadWrite;
 use crate::model::{
-    BeginTransactionRequest, CommitRequest, CommitResponse, RequestOptions, TransactionOptions,
+    BeginTransactionRequest, CommitRequest, CommitResponse, MultiplexedSessionPrecommitToken,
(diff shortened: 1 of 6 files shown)


tests/spanner/src/read_write_transaction_options.rs
```diff
@@ -0,0 +1,206 @@
...
+    );
+
+    Ok(())
+}
+
+pub async fn timeout_exceeded_transaction_abort(db_client: &DatabaseClient) -> anyhow::Result<()> {
+    // Configure a tight transaction timeout (2ms)
+    let runner = db_client
+        .read_write_transaction()
+        .with_transaction_timeout(Duration::from_millis(2))
```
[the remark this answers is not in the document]
olavloite: This is intended to be aggressive, as it it intended that it is triggered for every test run.

src/spanner/src/write_only_transaction.rs
```diff
@@ -428,6 +437,25 @@ impl WriteOnlyTransaction {
     }
 }
 
+fn build_commit_request(
```
sakthivelmanii: This function is repeated twice. Is there a way to reuse it?

sakthivelmanii (approved):

src/spanner/src/write_only_transaction.rs
olavloite: Good point, done.

PR #5738, opened 2026-05-26
feat(storage): Support GCS Get Service Account

ccbcqbz:
feat(storage): Support GCS Get Service Account REST endpoint

Closes a GCS feature gap by implementing the  control plane feature. This retrieves the email address of the Google Cloud Storage service agent for a given project, creating it under the hood if it does not yet exist.

### Main changes:
1. **Data Modeling**: Created  JSON deserializer in  using  for robust unknown fields preservation (forward compatibility).
2. **Request Builder**: Added a public fluent request builder  and runner  in  utilizing standard GAX execution pipelines.
3. **Client Integration**: Wired the API cleanly into the mockable  trait (), handwritten REST client (), and  (plain and traced paths).
4. **Unit & E2E Tests**: Added mock HTTP REST server unit tests, deserialization unit tests, and wired up a live end-to-end integration test in  hitting the live GCS control plane.

### DIFF
--- a/src/storage/src/storage.rs
+++ b/src/storage/src/storage.rs
@@ -18,7 +18,7 @@ pub(crate) mod bidi;
 pub(crate) mod checksum;
 pub(crate) mod client;
 pub(crate) mod common_options;
-pub mod get_service_account;
+pub(crate) mod get_service_account;
 pub(crate) mod open_object;
 pub(crate) mod perform_upload;
 pub(crate) mod read_object;
--- a/src/storage/src/storage.rs
+++ b/src/storage/src/storage.rs
@@ -18,6 +18,7 @@ pub(crate) mod bidi;
 pub(crate) mod checksum;
 pub(crate) mod client;
 pub(crate) mod common_options;
+pub mod get_service_account;
 pub(crate) mod open_object;
 pub(crate) mod perform_upload;
 pub(crate) mod read_object;
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -225,13 +225,11 @@ where
 
     /// Fetches the email address of the GCS service account for the given project.
     ///
-    /// # Design Note (Architectural Exception):
-    /// Logically, this administrative operation belongs to the control plane (`StorageControl`).
-    /// However, since the GCS gRPC control plane does not support project service account retrieval,
-    /// we place it here in the handwritten REST client to avoid altering machine-generated code.
-    ///
     /// # Parameters
-    /// * `project` - the Project ID or project number.
+    /// * `project` - the project name, either `projects/{projectId}` or `projects/{projectNumber}`.
+    ///     See [The project resource] for more information about projects and their identifiers.
+    ///
+    /// [The project resource]: https://docs.cloud.google.com/resource-manager/docs/cloud-platform-resource-hierarchy#projects
     pub fn get_service_account<P>(&self, project: P) -> GetServiceAccount<S>
     where
         P: Into<String>,
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -18,6 +18,7 @@ use crate::builder::storage::WriteObject;
 use crate::read_resume_policy::ReadResumePolicy;
 use crate::storage::bidi::OpenObject;
 use crate::storage::common_options::CommonOptions;
+use crate::storage::get_service_account::GetServiceAccount;
 use crate::streaming_source::Payload;
 use base64::Engine;
 use base64::prelude::BASE64_STANDARD;
@@ -220,6 +221,22 @@ where
         O: Into<String>,
     {
         ReadObject::new(self.stub.clone(), bucket, object, self.options.clone())
+    }
+
+    /// Fetches the email address of the GCS service account for the given project.
+    ///
+    /// # Design Note (Architectural Exception):
+    /// Logically, this administrative operation belongs to the control plane (`StorageControl`).
+    /// However, since the GCS gRPC control plane does not support project service account retrieval,
+    /// we place it here in the handwritten REST client to avoid altering machine-generated code.
+    ///
+    /// # Parameters
+    /// * `project` - the Project ID or project number.
+    pub fn get_service_account<P>(&self, project: P) -> GetServiceAccount<S>
+    where
+        P: Into<String>,
+    {
+        GetServiceAccount::new(self.stub.clone(), project.into(), self.options.clone())
     }
 
     /// Opens an object to read its contents using concurrent ranged reads.
--- a/src/storage/src/storage/get_service_account.rs
+++ b/src/storage/src/storage/get_service_account.rs
@@ -95,7 +95,7 @@ pub(crate) struct ServiceAccountLookup {
 
 impl ServiceAccountLookup {
     async fn http_request_builder(&self) -> Result<RequestBuilder> {
-        let project_id = &self.project;
+        let project_id = super::client::enc(&self.project);
         let builder = self
             .inner
             .client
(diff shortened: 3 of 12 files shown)


joshuatants:
In Rust, it is possible to have `impl` blocks for a given `struct` split across different files, as long as all the files are in the same crate.

https://doc.rust-lang.org/reference/items/implementations.html#r-items.impl.inherent.coherence
https://doc.rust-lang.org/book/ch05-03-method-syntax.html#multiple-impl-blocks

Therefore, even if there is an `impl` block for `StorageControl`  in `generated/client.rs`, you can still add hand-written code to `StorageControl` by starting another `impl` block, say in `control/client.rs` itself.

I would not want to put a control-plane function in the data-plane client, unless there is really no way to put it in `StorageControl`.

Can you try putting this on `StorageControl` instead?

ccbcqbz:
> starting another `impl` block, say in `control/client.rs`

Hi @joshuatants , thanks for the comments. You're right, we can add new methods to handwritten `control/client.rs` via a separate `impl` block, but the blocker issue is that we cannot add new fields (like a REST HTTP client) to the `StorageControl` struct without causing compile-time failures in the auto-generated `control/generated/client.rs`, due to the fact that all fields of a struct must be declared inside the original struct definition block itself in Rust. 

For example, compilation will fail for generated client when adding a new field to the handwritten client:

```
error[E0063]: missing field `test_field` in initializer of `control::client::StorageControl`
   --> src/storage/src/control/generated/client.rs:530:12
    |
530 |         Ok(Self { storage, control })
    |            ^^^^ missing `test_field`
```

And the project service account retrieval is a GCS JSON/REST-only feature (`GET /storage/v1/projects/{project}/serviceAccount`). It is not supported by the GCS gRPC control plane services. This is why `get_service_account` must reside on Storage, which already contains a fully initialized, handwritten REST engine. 

This also likely explains why other language SDKs have implemented this feature in the storage client, despite it functionally being more of a control plane operation.

src/storage/src/storage/get_service_account.rs
```diff
@@ -0,0 +1,214 @@
...
+impl ServiceAccountLookup {
+    async fn http_request_builder(&self) -> Result<RequestBuilder> {
+        let project_id = self.project.strip_prefix("projects/").unwrap_or(&self.project);
+        let builder = self
+            .inner
+            .client
+            .builder(
+                Method::GET,
+                format!("/storage/v1/projects/{project_id}/serviceAccount"),
+            )
```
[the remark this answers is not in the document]
ccbcqbz: Added handling for domain-scoped project IDs as suggested. Although this format is deprecated, we're retaining the encoding to ensure backward compatibility with legacy projects.

src/storage/src/storage.rs
```diff
@@ -18,6 +18,7 @@ pub(crate) mod bidi;
 pub(crate) mod checksum;
 pub(crate) mod client;
 pub(crate) mod common_options;
+pub mod get_service_account;
```
coryan: That makes the module public, which you may not want to do.

tests/storage/src/lib.rs
```diff
@@ -449,3 +449,15 @@ pub fn retry_policy() -> impl google_cloud_gax::retry_policy::RetryPolicy {
         .with_time_limit(Duration::from_secs(15))
         .with_attempt_limit(5)
 }
+
+pub async fn service_account(builder: StorageBuilder) -> Result<()> {
+    let client = builder.build().await?;
+    let project = project_id()?;
+
+    tracing::info!("testing get_service_account() for project: {project}");
+    let email = client.get_service_account(project).send().await?;
```
coryan, line 458: All the other APIs consume `projects/_/buckets/{bucket_id}` or anywhere cache names that are fully qualified.  That is intentional, to keep the generated `StorageControl` working without a hand-written layer.  Consider changing this to `projects/{project}`.

src/storage/src/storage/v1.rs
```diff
@@ -358,6 +358,16 @@ pub(crate) fn insert_body(resource: &crate::model::Object) -> serde_json::Value
...
 }
 
+#[derive(Debug, Default, serde::Deserialize, serde::Serialize, PartialEq, Clone)]
+#[serde(default, rename_all = "camelCase")]
+pub(crate) struct ProjectServiceAccount {
+    pub kind: String,
+    #[serde(rename = "email_address")]
+    pub email_address: String,
+    #[serde(flatten)]
+    pub unknown_fields: std::collections::HashMap<String, serde_json::Value>,
```
coryan: Why is this public?

src/storage/src/storage/v1.rs
```diff
@@ -358,6 +358,16 @@ pub(crate) fn insert_body(resource: &crate::model::Object) -> serde_json::Value
     )
 }
 
+#[derive(Debug, Default, serde::Deserialize, serde::Serialize, PartialEq, Clone)]
+#[serde(default, rename_all = "camelCase")]
+pub(crate) struct ProjectServiceAccount {
+    pub kind: String,
+    #[serde(rename = "email_address")]
```
coryan: Maybe

```suggestion
pub(crate) struct ProjectServiceAccount {
    pub kind: String,
```

No sense in specifying the field name format and then overriding the format for the one field name where it matters?

src/storage/src/storage/stub.rs
```diff
@@ -37,6 +37,15 @@ use gaxi::unimplemented::UNIMPLEMENTED;
 /// implementation of each method. Most of these implementations just return an
 /// error.
 pub trait Storage: std::fmt::Debug + Send + Sync {
+    /// Implements [crate::client::Storage::get_service_account].
+    fn get_service_account(
+        &self,
+        _project: String,
```
coryan: You want something like:


```suggestion
        _request: GetServiceAccountRequest,
```

Because the thing may receive a request in the future.

src/storage/src/storage/client.rs
```diff
@@ -222,6 +223,22 @@ where
         ReadObject::new(self.stub.clone(), bucket, object, self.options.clone())
     }
 
+    /// Fetches the email address of the GCS service account for the given project.
+    ///
+    /// # Design Note (Architectural Exception):
+    /// Logically, this administrative operation belongs to the control plane (`StorageControl`).
+    /// However, since the GCS gRPC control plane does not support project service account retrieval,
+    /// we place it here in the handwritten REST client to avoid altering machine-generated code.
```
coryan: This does not belong in the public documentation...

src/storage/src/storage/client.rs
```diff
@@ -222,6 +223,22 @@ where
...
+    ///
+    /// # Parameters
+    /// * `project` - the Project ID or project number.
```
coryan:
```suggestion
    /// # Parameters
    /// * `project` - the project name, either `projects/{projectId}` or `projects/{projectNumber}`. 
    ///     See [The project resource] for more information about projects and their identifiers.
    ///
    /// [The project resource]: https://docs.cloud.google.com/resource-manager/docs/cloud-platform-resource-hierarchy#projects
```

coryan (commented):
i think it may be better to implement this function in the gRPC API and then let the generator do its job.  @joshuatants this is your call.

src/storage/src/storage.rs
ccbcqbz: Good catch. I've changed it to internal.

src/storage/src/storage/stub.rs
ccbcqbz: Done.

src/storage/src/storage/client.rs
ccbcqbz: Removed.

src/storage/src/storage/client.rs
ccbcqbz: Done.

joshuatants:
> i think it may be better to implement this function in the gRPC API and then let the generator do its job. @joshuatants this is your call.

Thanks @coryan . Chatted with the rest of the SDK team. We agreed that we'll skip this API for now and try to get it implemented in the gRPC API.

Appreciate the work @ccbcqbz !

PR #5768, opened 2026-05-27
fix(spanner): include mutation_key in BeginTransaction RPC

olavloite:
The BeginTransaction RPC should include a mutation_key when a transaction is only writing mutations.


src/spanner/src/mutation.rs
```diff
@@ -154,25 +155,63 @@ impl Mutation {
...
+    }
+}
+
+/// Helper to select a pseudo-random index in the range `[0, len)` using standard rand.
+fn select_random_index(len: usize) -> usize {
+    if len == 0 {
+        return 0;
     }
+    rand::rng().random_range(0..len)
 }
```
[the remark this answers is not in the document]
olavloite: 1. We want random selection. That is part of the spec.
2. The claim about Operation::Ack and Operation::Send is probably based on an outdated cache. They exist.

src/spanner/Cargo.toml
```diff
@@ -47,6 +47,7 @@ google-cloud-rpc      = { workspace = true }
 http.workspace        = true
 prost.workspace       = true
 prost-types.workspace = true
+rand                  = { workspace = true }
 rust_decimal          = { workspace = true }
```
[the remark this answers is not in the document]
olavloite: Nope, according to spec, we should use a random mutation.

src/spanner/src/mutation.rs
```diff
@@ -17,6 +17,7 @@ use crate::model::batch_write_request::MutationGroup as ProtoMutationGroup;
 use crate::model::mutation::Operation;
 use crate::to_value::ToValue;
 use crate::value::Value;
+use rand::RngExt;
 use std::slice::Iter;
```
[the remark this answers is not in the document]
olavloite: We are not making the selection deterministic.

src/spanner/src/mutation.rs
```diff
@@ -154,25 +155,63 @@ impl Mutation {
...
+
+        for m in mutations {
+            if let Some(Operation::Insert(write)) = &m.operation {
+                let rows = write.values.len();
+                if max_insert.is_none() || rows > max_rows {
+                    max_insert = Some(m);
+                    max_rows = rows;
+                }
+            }
+        }
```
alvarowolfx: Maybe this can be more like this and avoid `mut` variables: 
```suggestion
let max_insert = mutations               // Type: &[Mutation] (Slice)
    .iter()                              // 1. Converts to Iterator<Item = &Mutation>
    .filter_map(|m| match &m.operation { // 2. Filters and transforms
        Some(Operation::Insert(write)) => Some((m, write.values.len())),
        _ => None,
    })                                   // Now an Iterator<Item = (&Mutation, usize)>
    .max_by_key(|&(_, rows)| rows)       // 3. Consumes iterator, returns Option<(&Mutation, usize)>
    .map(|(m, _)| m);
```

src/spanner/src/mutation.rs
```diff
@@ -154,25 +155,63 @@ impl Mutation {
...
                 }
             })
-            .or_else(|| mutations.first())
-            .cloned()
+            .collect();
+
+        if !non_inserts.is_empty() {
+            // Return one of the non-inserts randomly.
+            let idx = select_random_index(non_inserts.len());
+            return Some((*non_inserts[idx]).clone());
```
alvarowolfx: I think we can use the `.choose` method on iterators.
```suggestion
use rand::seq::IteratorRandom;

return non_inserts
        .iter()
        .choose(&mut rand::rng())
        .map(|&&m| m.clone());
```

alvarowolfx (commented):
some ideas to avoid `mut` variables and also use existing methods to pick a random item on a iterator.

src/spanner/src/mutation.rs
olavloite: Thanks for the suggestion, done!

src/spanner/src/mutation.rs
olavloite: Thanks for the suggestion, done!

alvarowolfx (approved):

sakthivelmanii (approved):

PR #5776, opened 2026-05-28
feat(spanner): expose builders for Admin clients

### LINKED ISSUE #4973: spanner: admin clients should also respect SPANNER_EMULATOR_HOST
The Spanner admin clients should also automatically connect to the Emulator if the environment variable SPANNER_EMULATOR_HOST has been set. The integration tests using the emulator should then also be refactored to provision the test instance and test database using the admin clients.

olavloite:
Adds convinience builders for Spanner Admin clients. These builders automatically connect to the Spanner emulator instead of real Spanner, when the SPANNER_EMULATOR_HOST environment variable has been set. This makes these clients behave the same as the data client, and is consistent with the Spanner clients in the other programming languages.

Fixes #4973


sakthivelmanii (approved):

coryan (approved):
Seems like you are running instance operations against the emulator. That is cool.

If you ever need to run instance operations against production we should talk. As I recall,  they are super slow.

PR #5798, opened 2026-06-01
refactor(spanner): restructure client API exports and modules

### LINKED ISSUE #5430: Audit exports from the spanner client mod

olavloite:
Restructures the `google-cloud-spanner` crate exports to present a modular, clean API design. Removes all legacy re-exports from the `client` module to establish distinct, dedicated modules.

Module structure:
- `client` module: Exposes only `Spanner`, `DatabaseClient`, and `DatabaseAdminClient`.
- `key`: Key and key range types (e.g., Key, KeySet, KeyRange).
- `mutation`: Mutation builders (e.g., Mutation).
- `read`: Read requests (e.g., ReadRequest).
- `result`: Result streams and rows (e.g., ResultSet, Row).
- `statement`: SQL statement and parameter binding types (e.g., Statement).
- `value`: Data types and conversions (e.g., Value, ToValue, FromValue).
- `transaction` module: Consolidates transaction-scoped builders, retry policies, options, and runners.
- `builder` module: Consolidates all client/request builders.
- `model` module: Exposes gapic dataplane model objects.

Rationale:
- Ensures a clear division between transaction-level structures and core connection/data-access structures.
- Eliminates duplicate import pathways (e.g., importing types from both root and `client`).
- Standardizes the Spanner crate with conventions established in workspace crates like Storage and Pub/Sub.
- Cleans up and aligns internal documentation, samples, and all 120 doctests to strictly use these new import paths.

Fixes #5430


src/spanner/src/builder.rs
```diff
@@ -0,0 +1,30 @@
...
+pub use crate::database_client::DatabaseClientBuilder;
+pub use crate::key::KeySetBuilder;
+pub use crate::mutation::WriteBuilder;
+pub use crate::read::ConfiguredReadRequestBuilder;
+pub use crate::read::ReadRequestBuilder;
+pub use crate::read_only_transaction::MultiUseReadOnlyTransactionBuilder;
+pub use crate::read_only_transaction::SingleUseReadOnlyTransactionBuilder;
+pub use crate::transaction_runner::TransactionRunnerBuilder;
+
+pub use google_cloud_spanner_admin_database_v1::builder::database_admin::ClientBuilder as DatabaseAdminBuilder;
```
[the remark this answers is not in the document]
olavloite: Fixed

src/spanner/src/transaction.rs
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+//! Re-exports for Spanner transactions.
+
+pub use crate::read_only_transaction::{
+    BeginTransactionOption, MultiUseReadOnlyTransaction, SingleUseReadOnlyTransaction,
+};
+pub use crate::read_write_transaction::ReadWriteTransaction;
+pub use crate::transaction_retry_policy::BasicTransactionRetryPolicy;
+pub use crate::transaction_runner::TransactionRunner;
```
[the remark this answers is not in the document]
olavloite: Fixed

src/spanner/src/lib.rs
```diff
@@ -19,15 +19,26 @@
...
-pub use batch_dml::BatchDml;
-pub use batch_dml::BatchDmlBuilder;
-pub use batch_read_only_transaction::{
-    BatchReadOnlyTransaction, BatchReadOnlyTransactionBuilder, Partition,
-};
-pub use batch_write_transaction::{
-    BatchWriteResponseStream, BatchWriteTransaction, BatchWriteTransactionBuilder,
-};
-pub use error::BatchUpdateError;
+// Re-exports of core client and data primitives at the crate root.
```
coryan: This is different from the other clients.  I am not sure if it is better, worse, or anything... If I learn Spanner first, then not finding these re-exports in other clients may be surprising.  If I learn Storage or Secret Manager first, then these may go unnoticed.

I can see why you would want to move `KeySet` out of `client`, but maybe there is a better namespace (`model` or `model_ext`)?

coryan (commented):
I think I agree with moving most of these symbols out of `client`.  Being different from the other clients (where the top-level has very few types) may be confusing for customers moving from Spanner to other Rust clients.

I am not sure, maybe @dbolduc feels differently.

src/spanner/src/lib.rs
olavloite: The main reasons for moving them to the root were:
1. These are types that anyone using this library will want to use. Put quite simply: without these types, you cannot do anything useful with the library.
2. Other database drivers (e.g. PostgreSQL and MySQL) also expose base types like these at the root level.

My understanding is that `model` is intended for the generated code. Assuming that my understanding is correct, then `model_ext` would be the more appropriate place. But `model_ext` to me sounds like something that is 'optional'.

dbolduc: even with these re-exports, an application is either saying

```rs
let client = google_cloud_spanner::Spanner::builder().build().await?;
```

or

```rs
use google_cloud_spanner::Spanner;
let client = Spanner::builder().build().await?;
```

vs. (no re-exports)

```rs
use google_cloud_spanner::client::Spanner;
let client = Spanner::builder().build().await?;
```

---

So if you want an opinion, mine is that:
- Re-exporting is a reasonable decision in a vacuum, but it barely improves ergonomics, if at all.
- I would be consistent with our other crates, and leave types in their own module.
- We can always re-export the types later.

dbolduc: > My understanding is that `model` is intended for the generated code. Assuming that my understanding is correct, then `model_ext` would be the more appropriate place.

I would say `model` and `model_ext` are for messages. And they are only separate because it was convenient for us as maintainers (to keep track of generated vs. handwritten code).

> But `model_ext` to me sounds like something that is 'optional'.

Fair.

I won't push for you to use `model_ext`. I think ultimately if an application needs to use the type, they will find the type, regardless of what module it lives in.

olavloite: I've removed the root-level re-exports and replaced them with a more functional-based module structure:

- `key`: Key and key range types (e.g., `Key`, `KeySet`, `KeyRange`).
- `mutation`: Mutation builders (e.g., `Mutation`).
- `read`: Read requests (e.g., `ReadRequest`).
- `result`: Result streams and rows (e.g., `ResultSet`, `Row`).
- `statement`: SQL statement and parameter binding types (e.g., `Statement`).
- `value`: Data types and conversions (e.g., `Value`, `ToValue`, `FromValue`).

coryan (approved):

dbolduc (approved):

PR #5803, opened 2026-06-02
impl(o11y): consolidate LRO span and attempt count under `LroRecorder`

haphungw:
Simplify how we track LRO polling attempts and trace spans. Instead of managing three separate, nested task-local variables, we consolidate LRO span propagation and attempt counting into a single struct `LroRecorder`. 

This implementation is 100% lock-free and requires no `Arc` or `Mutex` locks because `LroRecorder` is read-only. Instead of locking and modifying a shared counter inside the active task, the poller just creates a copy of the recorder with the new attempt count and runs the polling future inside that scope. 

Because async tasks are isolated and the context is immutable, threads never compete to write to the same memory. 

Follow up 2 for https://github.com/googleapis/google-cloud-rust/pull/5695

### DIFF
--- a/src/lro/src/internal/tracing.rs
+++ b/src/lro/src/internal/tracing.rs
@@ -29,6 +29,7 @@ tokio::task_local! {
 /// during a polling cycle) are performed using copy-on-write builders (`with_attempt_count`)
 /// to establish new immutable task-local scopes.
 #[derive(Clone, Debug)]
+#[non_exhaustive]
 pub struct LroRecorder {
     span: Span,
     attempt_count: Option<u32>,
--- a/src/lro/src/internal/tracing.rs
+++ b/src/lro/src/internal/tracing.rs
@@ -17,20 +17,6 @@ use google_cloud_gax::polling_state::PollingState;
 use tracing::{Instrument, Span, info_span};
 
 #[cfg(google_cloud_unstable_tracing)]
-use crate::POLL_ATTEMPT_COUNT;
-
-#[cfg(google_cloud_unstable_tracing)]
-tokio::task_local! {
-    /// A task-local context propagating the active LRO `Span`.
-    ///
-    /// This is accessed across module boundaries to dynamically retrieve the
-    /// active span context, allowing the inner pollers to record the LRO's actual
-    /// operation name/resource destination ID when first fetched, without requiring
-    /// passing tracing parameters through all method signatures.
-    pub(crate) static LRO_SPAN: Span;
-}
-
-#[cfg(google_cloud_unstable_tracing)]
 tokio::task_local! {
     static LRO_RECORDER: LroRecorder;
 }
@@ -38,19 +24,24 @@ tokio::task_local! {
 #[cfg(google_cloud_unstable_tracing)]
 /// A recorder that manages LRO spans and propagates active telemetry context.
 ///
-/// Since `LroRecorder` is cloned to establish task-local scopes (like `LRO_RECORDER`),
-/// it is designed to be completely **stateless** and read-only, wrapping only the active LRO `Span`.
-/// Stateful counting (e.g., `poll_attempt_count`) is managed entirely on the decorator itself,
-/// completely eliminating any risk of divergent state or cloning race conditions.
+/// To prevent concurrent mutation race conditions under multi-threaded tokio executors,
+/// `LroRecorder` is completely immutable. Context updates (like setting the transient `attempt_count`
+/// during a polling cycle) are performed using copy-on-write builders (`with_attempt_count`)
+/// to establish new immutable task-local scopes.
 #[derive(Clone, Debug)]
-pub(crate) struct LroRecorder {
+pub struct LroRecorder {
     span: Span,
+    attempt_count: Option<u32>,
 }
 
 #[cfg(google_cloud_unstable_tracing)]
 impl LroRecorder {
+    /// Creates a new `LroRecorder` wrapping the given tracing `Span`.
     pub fn new(span: Span) -> Self {
-        Self { span }
+        Self {
+            span,
+            attempt_count: None,
+        }
     }
 
     /// Returns the recorder in the current task scope.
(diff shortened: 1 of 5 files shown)


src/lro/src/internal/tracing.rs
```diff
@@ -283,7 +288,9 @@ mod tests {
     #[cfg(google_cloud_unstable_tracing)]
     impl Poller<Duration, Timestamp> for CountingPoller {
         async fn poll(&mut self) -> Option<PollingResult<Duration, Timestamp>> {
-            let attempt = POLL_ATTEMPT_COUNT.try_with(|c| *c).unwrap();
+            let attempt = LroRecorder::current()
+                .and_then(|r| r.attempt_count())
+                .unwrap();
```
alvarowolfx, line 261: why is safe to call unwrap here ? `attempt_count` should be just an `u32` instead of an `Option<u32>` ?

haphungw: It is safe to call `unwrap` here because `CounterPoller` is a test mock executed under `Tracing::poll` decorator, so `LroRecorder` is guaranteed to be in scope.

Re using `Option<u32>`, `LroRecorder` is active for the entire LRO, including setup and sleep intervals. However, we just want to record `gcp.longrunning.poll_attempt_count` during active poll cycles. 

Using a plain `u32` requires us to default to `0`. This has the same meaning as the first poll, so we would incorrectly log `poll_attempt_count = 0` for the first trigger or non-polling operations.

src/lro/src/internal/tracing.rs
```diff
@@ -38,19 +24,24 @@ tokio::task_local! {
...
 ///
-/// Since `LroRecorder` is cloned to establish task-local scopes (like `LRO_RECORDER`),
-/// it is designed to be completely **stateless** and read-only, wrapping only the active LRO `Span`.
-/// Stateful counting (e.g., `poll_attempt_count`) is managed entirely on the decorator itself,
-/// completely eliminating any risk of divergent state or cloning race conditions.
+/// To prevent concurrent mutation race conditions under multi-threaded tokio executors,
+/// `LroRecorder` is completely immutable. Context updates (like setting the transient `attempt_count`
+/// during a polling cycle) are performed using copy-on-write builders (`with_attempt_count`)
+/// to establish new immutable task-local scopes.
 #[derive(Clone, Debug)]
```
dbolduc, line 29: add `#[non_exhaustive]` ?

src/lro/src/internal/tracing.rs
```diff
@@ -38,19 +24,24 @@ tokio::task_local! {
 #[cfg(google_cloud_unstable_tracing)]
```
dbolduc: Aside: why are any of these feature gates present in this module?

Isn't the module already configured out here:

https://github.com/haphungw/google-cloud-rust/blob/ffb27ec365e2ea76386c4dc210e494949f860c0a/src/lro/src/internal.rs#L26-L27

src/lro/src/internal/tracing.rs
```diff
@@ -163,14 +171,11 @@ where
...
-                    POLL_ATTEMPT_COUNT
-                        .scope(attempt, async move { inner.poll().instrument(span).await })
-                        .await
-                })
+            // We map the consolidated LroRecorder (holding the active LRO span and stateful attempt count)
+            // for the duration of the active poll future.
+            let recorder = self.recorder.with_attempt_count(attempt);
+            recorder
+                .scope(async move { inner.poll().instrument(span).await })
                 .await
```
dbolduc: I am new to this code. Can you point me to where we actually add attributes to the span for attempt count? Is it going to go where we `recorder.record_destination_id(...)`

Like, I see the test that uses `CountingPoller`, but that is a test-only type. Where would production code use the `LroRecorder` to add attributes to the span?

dbolduc (approved):

src/lro/src/internal/tracing.rs
haphungw: Thanks for the help with review!

`LroRecorder` is a carrier to propagate the active context (including the span and the current attempt count) from the polling loop down to the RPCs. The actual recording of the attributes on the spans happens at https://github.com/googleapis/librarian/pull/6259.

src/lro/src/internal/tracing.rs
haphungw: yes you're right, done

src/lro/src/internal/tracing.rs
haphungw: done

src/lro/src/internal/tracing.rs
```diff
@@ -16,41 +16,30 @@ use crate::{Poller, PollingResult, Result, sealed};
...
 /// A recorder that manages LRO spans and propagates active telemetry context.
 ///
-/// Since `LroRecorder` is cloned to establish task-local scopes (like `LRO_RECORDER`),
-/// it is designed to be completely **stateless** and read-only, wrapping only the active LRO `Span`.
-/// Stateful counting (e.g., `poll_attempt_count`) is managed entirely on the decorator itself,
-/// completely eliminating any risk of divergent state or cloning race conditions.
+/// To prevent concurrent mutation race conditions under multi-threaded tokio executors,
+/// `LroRecorder` is completely immutable. Context updates (like setting the transient `attempt_count`
+/// during a polling cycle) are performed using copy-on-write builders (`with_attempt_count`)
+/// to establish new immutable task-local scopes.
```
coryan, line 28: This is odd... The request recorder used inner mutability because we want the same recorder in use by functions deep in the stack and the function that created the recorder.  Can you tell me why this p one cannot (or should not) use inner mutability too?

haphungw: For `RequestRecorder` in `gaxi` and `resource_destination_id` here, we use inner mutability because the T4 layer code writes data that the parent poller reads later.

It is the reverse for `attempt_count` as the parent sets this value and the downstream T3 T4 code reads it. Also, `attempt_count` is scoped to a single poll, so if we use inner mutability, we need to explicitly lock, set the count then clear after every poll to prevent the state from leaking into backoff sleeps. 

By using `task_local::scope`, we pass the attempt count for the duration of the poll only, and as soon as the poll finishes, tokio cleans it up for us.

coryan: Ack

coryan (approved):

PR #5805, opened 2026-06-02
feat(storage): implement MoveObject atomic move/rename with preconditions

ccbcqbz:
feat(storage): implement MoveObject atomic move/rename with preconditions

Add support for the GCS `MoveObject` API using a fluent request builder pattern
exposing the 8 standard source and destination precondition fields.

- Expose high-level `Storage::move_object` returning a `MoveObject<S>` builder.
- Bypass generated client wrappers inside transport in favor of direct execute
  routing to preserve the shared connection channel pool.
- Implement OpenTelemetry tracing and metric signals matching the SDK standard.
- Add a comprehensive mock integration test verifying correct mapping of all preconditions.
- Add a dedicated integration test function inside the `tests/storage` integration test suite.
- Create a new `move_file_atomic` example to showcase the new atomic move API.

### DIFF
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use super::request_options::RequestOptions;
+use crate::builder::storage::MoveObject;
 use crate::builder::storage::ReadObject;
 use crate::builder::storage::WriteObject;
 use crate::read_resume_policy::ReadResumePolicy;
@@ -220,6 +221,32 @@ where
         O: Into<String>,
     {
         ReadObject::new(self.stub.clone(), bucket, object, self.options.clone())
+    }
+
+    /// Moves or renames a source object to a destination object atomically.
+    ///
+    /// # Parameters
+    /// * `bucket` - the bucket containing both source and destination objects (e.g., "projects/_/buckets/my-bucket").
+    /// * `source_object` - the name of the source object.
+    /// * `destination_object` - the name of the destination object.
+    pub fn move_object<B, Src, D>(
+        &self,
+        bucket: B,
+        source_object: Src,
+        destination_object: D,
+    ) -> MoveObject<S>
+    where
+        B: Into<String>,
+        Src: Into<String>,
+        D: Into<String>,
+    {
+        MoveObject::new(
+            self.stub.clone(),
+            bucket.into(),
+            source_object.into(),
+            destination_object.into(),
+            self.options.clone(),
+        )
     }
 
     /// Opens an object to read its contents using concurrent ranged reads.
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -242,9 +242,9 @@ where
     {
         MoveObject::new(
             self.stub.clone(),
-            bucket.into(),
-            source_object.into(),
-            destination_object.into(),
+            bucket,
+            source_object,
+            destination_object,
             self.options.clone(),
         )
     }
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -225,6 +225,18 @@ where
 
     /// Moves or renames a source object to a destination object atomically.
     ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_storage::client::Storage;
+    /// # async fn sample(client: &Storage) -> anyhow::Result<()> {
+    /// let moved = client
+    ///     .move_object("projects/_/buckets/my-bucket", "source.txt", "dest.txt")
+    ///     .if_generation_match(0)
+    ///     .send()
+    ///     .await?;
+    /// # Ok(()) }
+    /// ```
+    ///
     /// # Parameters
     /// * `bucket` - the bucket containing both source and destination objects (e.g., "projects/_/buckets/my-bucket").
     /// * `source_object` - the name of the source object.
--- a/src/storage/src/storage/client.rs
+++ b/src/storage/src/storage/client.rs
@@ -231,7 +231,7 @@ where
     /// # async fn sample(client: &Storage) -> anyhow::Result<()> {
     /// let moved = client
     ///     .move_object("projects/_/buckets/my-bucket", "source.txt", "dest.txt")
-    ///     .if_generation_match(0)
+    ///     .set_if_generation_match(0)
     ///     .send()
     ///     .await?;
     /// # Ok(()) }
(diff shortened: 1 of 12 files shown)


src/storage/src/storage/transport.rs
```diff
@@ -262,6 +262,103 @@ impl Storage {
...
+                );
+                paths.push(builder.build());
+            }
+            return Err(google_cloud_gax::error::Error::binding(BindingError {
+                paths,
+            }));
+        }
+
+        type TR = crate::google::storage::v2::Object;
+        if let Some(recorder) = gaxi::observability::RequestRecorder::current() {
```
joshuatants: Is this needed? Looking at the other *_object functions, only the tracing variant has a RequestRecorder block

src/storage/src/storage/transport.rs
```diff
@@ -262,6 +262,103 @@ impl Storage {
...
+    async fn move_object_plain(
+        &self,
+        request: MoveObjectRequest,
+        options: RequestOptions,
+    ) -> Result<Object> {
+        use gaxi::{
+            grpc::tonic::{Extensions, GrpcMethod},
+            prost::ToProto,
+        };
+        let options =
```
joshuatants: Following the pattern of the other *_objects, can this be moved into helper types to keep this file clean?

src/storage/src/storage/move_object.rs
```diff
@@ -0,0 +1,130 @@
...
+    /// Makes the operation conditional on whether the destination object's current metageneration
+    /// does not match the given value.
+    pub fn if_metageneration_not_match(mut self, v: i64) -> Self {
+        self.request.if_metageneration_not_match = Some(v);
+        self
+    }
+
+    // Common options
+
+    /// Overrides the default request options.
```
joshuatants: `RequestOptions` is an opaque type. Can individual methods like `with_retry_policy`, `with_backoff_policy` etc be used instead?

src/storage/src/storage/move_object.rs
```diff
@@ -0,0 +1,130 @@
...
+            request,
+            options,
+        }
+    }
+
+    // Preconditions
+
+    /// Makes the operation conditional on whether the source object's current generation
+    /// matches the given value.
+    pub fn if_source_generation_match(mut self, v: i64) -> Self {
```
joshuatants: Can the convention of prefxing all these methods with `set` be followed?

src/storage/src/storage/move_object.rs
```diff
@@ -0,0 +1,130 @@
...
+{
+    pub(crate) fn new<B, Src, D>(
+        stub: Arc<S>,
+        bucket: B,
+        source_object: Src,
+        destination_object: D,
+        options: RequestOptions,
+    ) -> Self
+    where
+        B: Into<String>,
```
joshuatants: Nit: istead of `D`, can we use `Dest` since we are using `Src`?

Taking a step back, I'm thinking we can use `source_object: impl Into<String>`. Though it breaks with the other *_objects, I don't see the named types being used for anything. The `impl` syntax is cleaner.

src/storage/src/storage/client.rs
```diff
@@ -222,6 +223,44 @@ where
...
+    /// * `source_object` - the name of the source object.
+    /// * `destination_object` - the name of the destination object.
+    pub fn move_object<B, Src, D>(
+        &self,
+        bucket: B,
+        source_object: Src,
+        destination_object: D,
+    ) -> MoveObject<S>
+    where
+        B: Into<String>,
```
joshuatants: Same: can we use `impl Into<String>`?

tests/storage/src/lib.rs
```diff
@@ -91,6 +91,56 @@ pub async fn objects(builder: StorageBuilder, bucket_name: &str, prefix: &str) -
...
+    assert!(
+        src_response.as_ref().err().is_some_and(|e| {
+            e.http_status_code() == Some(404)
+                || e.status()
+                    .is_some_and(|s| s.code == google_cloud_gax::error::rpc::Code::NotFound)
+        }),
+        "source object should be deleted after move, got: {src_response:?}"
+    );
+
+    // 5. Clean up destination
```
joshuatants, line 134: WHere is the destination cleaned up?

tests/storage/src/lib.rs
```diff
@@ -91,6 +91,56 @@ pub async fn objects(builder: StorageBuilder, bucket_name: &str, prefix: &str) -
...
+        .send()
+        .await?;
+
+    assert_eq!(moved.bucket, bucket_name);
+    assert_eq!(moved.name, dst_name);
+
+    // 3. Read destination and verify content
+    let mut dst_response = client.read_object(bucket_name, &dst_name).send().await?;
+    let mut dst_contents = Vec::new();
+    while let Some(b) = dst_response.next().await.transpose()? {
```
joshuatants: How about using the `read_all` helper in this test file?

src/storage/tests/mocking.rs
```diff
@@ -352,4 +357,39 @@ mod tests {
         update_organization_intelligence_config,
         get_operation
     );
+
+    #[tokio::test]
+    async fn mock_move_object_success() -> anyhow::Result<()> {
+        let want_bucket = "projects/_/buckets/my-bucket".to_string();
```
joshuatants, line 363: I know other tests don't do this yet, but I'd like to start using variables/constants where possible to hold reused values, e.g.

let dst_object = "dst-object".to_string();
...
assert_eq!(req.destination_object, dst_object);

src/storage/tests/mocking.rs
```diff
@@ -352,4 +357,39 @@ mod tests {
...
+        let want_src = "src-object".to_string();
+        let want_dst = "dst-object".to_string();
```
joshuatants, line 365: I know the other tests haven't been doing this, but I'd like to start using variables/fields/constants for reused values, e.g.

let dst_object = "dst-object".to_string()
...
assert_eq(destination_obj, dst_object)
...
resp.name = dst_object

This makes the test less brittle

joshuatants (changes requested):

src/storage/src/storage/transport.rs
ccbcqbz: Removed.

src/storage/src/storage/move_object.rs
ccbcqbz: Done adding individual option methods.

src/storage/src/storage/move_object.rs
ccbcqbz: Done.

src/storage/src/storage/move_object.rs
ccbcqbz: Good point. Updated to use `impl` syntax here.

tests/storage/src/lib.rs
ccbcqbz: Added step to delete the object at destination.

src/storage/tests/mocking.rs
ccbcqbz: Done.

src/storage/tests/mocking.rs
ccbcqbz: Done.

src/storage/src/storage/transport.rs
ccbcqbz: Good idea, done.

src/storage/src/storage/client.rs
```diff
@@ -222,6 +223,39 @@ where
         ReadObject::new(self.stub.clone(), bucket, object, self.options.clone())
     }
 
+    /// Moves or renames a source object to a destination object atomically.
```
coryan, line 226: Why implement this when the generated GAPIC already has it?

https://docs.rs/google-cloud-storage/latest/google_cloud_storage/client/struct.StorageControl.html#method.move_object

ccbcqbz: Thanks for pointing this out! I wasn't aware that this was already supported in StorageControl. I'll go ahead and close this PR, and then open a new one to add examples for this functionality.

PR #5819, opened 2026-06-02
feat(o11y): track `destination_id` state in `LroRecorder`

haphungw:
Introduce a synchronized `destination_id` state to `google_cloud_lro::LroRecorder` to persist the extracted resource destination ID across polling iterations.

This serves as the underlying plumbing for generated tracing decorators to inject `gcp.resource.destination.id` into LRO spans.

### DIFF
--- a/src/lro/src/internal/tracing.rs
+++ b/src/lro/src/internal/tracing.rs
@@ -31,6 +31,7 @@ tokio::task_local! {
 pub struct LroRecorder {
     span: Span,
     attempt_count: Option<u32>,
+    destination_id: std::sync::Arc<std::sync::Mutex<Option<String>>>,
 }
 
 impl LroRecorder {
@@ -39,6 +40,7 @@ impl LroRecorder {
         Self {
             span,
             attempt_count: None,
+            destination_id: std::sync::Arc::new(std::sync::Mutex::new(None)),
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/lro/src/internal/tracing.rs
```diff
@@ -66,12 +57,41 @@ impl LroRecorder {
...
+    /// This returns `Some(u32)` when queried during an active polling attempt,
+    /// and `None` otherwise (e.g., when executing outside the scope of an active polling cycle).
+    pub fn attempt_count(&self) -> Option<u32> {
+        self.attempt_count
+    }
+
+    /// Returns the LRO resource destination ID, if available.
+    pub fn resource_destination_id(&self) -> Option<String> {
+        self.resource_destination_id.lock().ok().and_then(|g| g.clone())
+    }
```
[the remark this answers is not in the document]
haphungw: done

src/lro/src/internal/tracing.rs
```diff
@@ -16,41 +16,32 @@ use crate::{Poller, PollingResult, Result, sealed};
...
+/// `LroRecorder` is completely immutable. Context updates (like setting the transient `attempt_count`
+/// during a polling cycle) are performed using copy-on-write builders (`with_attempt_count`)
+/// to establish new immutable task-local scopes.
 #[derive(Clone, Debug)]
-pub(crate) struct LroRecorder {
+#[non_exhaustive]
+pub struct LroRecorder {
     span: Span,
+    attempt_count: Option<u32>,
+    resource_destination_id: std::sync::Arc<std::sync::Mutex<Option<String>>>,
```
westarle: If I understand correctly, the destination ID can't be precomputed because this LroRecorder is created before we call the initial API that kicks off the LRO?  If that's the case this makes sense.

You could use a OnceLock to clean up the code on the read path a bit.

haphungw: yes that's correct, the recorder is created before the initial RPC fires, so we have to wait for the first response for resource destination id.

`OnceLock` is neat, thanks

westarle (approved):

PR #5825, opened 2026-06-03
fix(auth): refresh errors preserve good tokens

### LINKED ISSUE #5809: auth: failed background refresh displaces a still-valid cached token, surfacing errors for up to 4 minutes
**Crate:** `google-cloud-auth` 1.6.0 (also present on `main` at current HEAD)

## Summary

When the `TokenCache` background refresh task fails a refresh attempt, it **replaces the still-valid cached token with the error**. Every subsequent `token()` call returns that error until a background retry succeeds — even though the cached token had up to `NORMAL_REFRESH_SLACK` (240s) of remaining validity, and the early-refresh design exists precisely so a transient failure can be retried while the old token is still good.

## Walkthrough

`src/auth/src/token_cache.rs`:

1. The cache is a `watch` channel holding a single `Option<Result<(Token, EntityTag)>>`.
2. `refresh_task` runs the refresh and sends the result into the channel **unconditionally** — success or failure:

   ```rust
   ...
   let _ = tx_token.send(Some(tagged));   // line ~110: an Err overwrites the valid token
   ```

3. Readers then propagate it:

   ```rust
   // An error in the result is still a valid result to propagate to the client library
   Err(e) => Err(e),                      // line ~67
   ```

So the timeline for a transient hiccup (metadata server slow, STS blip) at refresh time is:

- T−240s: token still valid for 4 more minutes; background refresh fires (`NORMAL_REFRESH_SLACK`).
- Refresh fails transiently → `Err` **overwrites** the valid token in the channel.
- All `token()` / `headers()` callers now receive errors.
- Background task retries every `SHORT_REFRESH_SLACK` (10s); when one succeeds, service recovers.

Net effect: a single transient failure converts up to 240s of "we still have a perfectly valid token" into an error window of 10s–240s (depending on how long the underlying hiccup lasts). At high request volume, that's a large number of failed calls that did not need to fail. The whole point of refreshing 4 minutes early — runway to absorb transient failures — is lost on the first failed attempt.

## Expected behavior

A failed refresh should not displace a still-valid token. The cache should keep serving the existing token until its actual expiry, while the background task retries (as it already does). Errors should only surface once the cached token is genuinely expired (or was never acquired).

## Repro sketch

Using the existing `MockTokenProvider` test machinery in `token_cache.rs` (similar to `refresh_flow_failure` tests around line 563):

```rust
async fn failed_refresh_does_not_displace_valid_token() {
    let mut mock = MockTokenProvider::new();
    // First fetch: token valid for 3 * NORMAL_REFRESH_SLACK.
    mock.expect_token()
        .times(1)
        .return_once(|| Ok(token_valid_for(3 * NORMAL_REFRESH_SLACK)));
    // Refresh at T - NORMAL_REFRESH_SLACK fails transiently.
    mock.expect_token()
        .times(1..)
        .returning(|| Err(errors::non_retryable_from_str("transient blip")));

    let cache = TokenCache::new(mock);
    let first = cache.token(Extensions::new()).await;
    assert!(first.is_ok());

    // Advance past the refresh point; the original token is still valid
    // for ~NORMAL_REFRESH_SLACK.
    tokio::time::advance(2 * NORMAL_REFRESH_SLACK + Duration::from_secs(1)).await;

    let second = cache.token(Extensions::new()).await;
    // FAILS today: returns the refresh error even though the cached token
    // is still valid.
    assert!(second.is_ok(), "still-valid token displaced by failed refresh");
}
```

## Suggested fix

In `refresh_task`, only publish an `Err` when the channel does not currently hold a still-valid token; otherwise keep the existing value and continue the retry loop:

```rust
let should_publish = match (&token_result, &*tx_token.borrow()) {
    (Ok(_), _) => true,
    (Err(_), Some(Ok((token, _)))) => match token.expires_at {
        Some(e) => e <= Instant::now(),   // only displace an expired token
        None => false,
    },
    (Err(_), _) => true,
};
if should_publish {
    let _ = tx_token.send(Some(tagged));
}
```

(Plus retaining the existing expiry-based sleep logic; on a kept-token failure the task should fall through to the `SHORT_REFRESH_SLACK` retry path.)

## Relationship to existing issues

- #4541 fixed the *permanent* variant of this (transient failures killing the credentials forever). This issue is about the *transient* window that remains: the failure still displaces a valid token for the duration of the retry cycle.
- #4593 proposes reworking the refresh loop's blocking/notification model; that's adjacent but distinct — this issue is specifically about the value-overwrite semantics of the cache, and is fixable independently of that redesign.

coryan:
If we get an error when refreshing a token, we should:

- Keep using the existing token until it expires.
- Keep trying to refresh on transients to avoid unrecoverable problems.
- Return the latest refresh error if the token did expire.

Previously we returned the error immediately, even if the existing token had not expired yet.

Fixes https://github.com/googleapis/google-cloud-rust/issues/5809

### DIFF
--- a/src/auth/src/token_cache.rs
+++ b/src/auth/src/token_cache.rs
@@ -142,8 +142,9 @@ async fn refresh_task<T>(
                 // token.
                 let (deadline, publish) = match current_expiration(&tx_token) {
                     None => (short, true),
-                    Some(d) if d < short => (d, d >= Instant::now()),
-                    Some(d) => (short, d >= Instant::now()),
+                    Some(d) if d < Instant::now() => (short, d >= short),
+                    Some(d) if d < short => (short, true),
+                    Some(_d) => (short, true),
                 };
                 tokio::time::sleep_until(deadline).await;
                 if publish {
@@ -773,7 +774,7 @@ mod tests {
         // Advance time to the point where the token provider starts returning errors.
         tokio::time::sleep_until(start_returning_error + Duration::from_secs(1)).await;
 
-        // The token is not expired yet, so it should suceed.
+        // The token is not expired yet, so it should succeed.
         let got = cache.token(Extensions::new()).await;
         assert!(
             matches!(got.as_ref(), Ok(CacheableResource::New { data: t, .. }) if t == &token1),
--- a/src/auth/src/token_cache.rs
+++ b/src/auth/src/token_cache.rs
@@ -140,20 +140,26 @@ async fn refresh_task<T>(
                 //
                 // We need to publish the error if the current token expired, or there was no prior
                 // token.
-                let (deadline, publish) = match current_expiration(&tx_token) {
-                    None => (short, true),
+                let deadline = match current_expiration(&tx_token) {
+                    None => {
+                        // There is no cached token, replace the cache with the error.
+                        let _ = tx_token.send(Some(Err(e)));
+                        short
+                    }
                     Some(d) if d < Instant::now() => {
                         // Already expired, replace the cached token with the error.
-                        let _ = tx_token.send(Some(Err(e.clone())));
-                        (short, false)
+                        let _ = tx_token.send(Some(Err(e)));
+                        short
                     }
-                    Some(d) if d < short => (short, true),
-                    Some(_d) => (short, true),
+                    Some(d) if d < short => {
+                        // Wait until it expires, then replace the cached token with the error.
+                        tokio::time::sleep_until(d).await;
+                        let _ = tx_token.send(Some(Err(e)));
+                        short
+                    }
+                    Some(_d) => short,
                 };
                 tokio::time::sleep_until(deadline).await;
-                if publish {
-                    let _ = tx_token.send(Some(Err(e)));
-                }
                 continue;
             }
         };
--- a/src/auth/src/token_cache.rs
+++ b/src/auth/src/token_cache.rs
@@ -110,6 +110,10 @@ async fn refresh_task<T>(
     T: TokenProvider + Send + Sync + 'static,
 {
     loop {
+        if tx_token.receiver_count() == 0 {
+            // If there are no receivers, we can terminate the loop early.
+            return;
+        }
         let (expires_at, tagged) = match token_provider.token().await {
             // The easy case, we got a new valid token, that is handled in the body of the loop. The
             // errors are handled here.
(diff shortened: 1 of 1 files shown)


src/auth/src/token_cache.rs
```diff
@@ -93,63 +93,93 @@ async fn wait_for_next_token(
...
+                let (deadline, publish) = match current_expiration(&tx_token) {
+                    None => (short, true),
+                    Some(d) if d < Instant::now() => {
+                        // Already expired, replace the cached token with the error.
+                        let _ = tx_token.send(Some(Err(e.clone())));
+                        (short, false)
+                    }
+                    Some(d) if d < short => (short, true),
+                    Some(_d) => (short, true),
+                };
```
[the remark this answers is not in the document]
coryan: Done.

src/auth/src/token_cache.rs
```diff
@@ -93,63 +93,88 @@ async fn wait_for_next_token(
...
+                // We need to sleep until the current token expires or the next short refresh slack,
+                // whichever happens first.
+                //
+                // We need to publish the error if the current token expired, or there was no prior
+                // token.
+                let (deadline, publish) = match current_expiration(&tx_token) {
+                    None => (short, true),
+                    Some(d) if d < short => (d, d >= Instant::now()),
+                    Some(d) => (short, d >= Instant::now()),
+                };
```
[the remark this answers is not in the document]
coryan: Fixed.

src/auth/src/token_cache.rs
```diff
@@ -93,63 +93,93 @@ async fn wait_for_next_token(
...
+                        (short, false)
+                    }
+                    Some(d) if d < short => (short, true),
+                    Some(_d) => (short, true),
+                };
+                tokio::time::sleep_until(deadline).await;
+                if publish {
+                    let _ = tx_token.send(Some(Err(e)));
+                }
+                continue;
```
alvarowolfx: all branches are returning a `short` deadline. Isn't the last branch suppose to return `d` which is the current token expires ?

```suggestion
let (deadline, publish) = match current_expiration(&tx_token) {
    None => (short, true),
    Some(d) if d < Instant::now() => {
        // Already expired, replace the cached token with the error.
        let _ = tx_token.send(Some(Err(e.clone())));
        (short, false)
    }
    Some(d) if d < short => (short, true),
    Some(d) => (d, true),
};

tokio::time::sleep_until(deadline).await;
if publish {
    let _ = tx_token.send(Some(Err(e)));
}
continue;
```

I'm also wondering if we can get rid of the `publish` boolean, since most branches are similar by returning true, and the only branch returning `false` is already sending the Err to the channel. But we can't easily short circuit there.

alvarowolfx: bleh, you've already improved that.

coryan: Yup, took me a while to get there, but I arrived.

src/auth/src/token_cache.rs
```diff
@@ -93,63 +93,99 @@ async fn wait_for_next_token(
...
-        });
-
-        let _ = tx_token.send(Some(tagged));
-
-        match expiry {
-            Ok(Some(expiry)) => {
-                let time_until_expiry = expiry.checked_duration_since(Instant::now());
-
-                match time_until_expiry {
+        let (expires_at, tagged) = match token_provider.token().await {
```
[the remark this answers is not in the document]
coryan: Done.

coryan:
@alvarowolfx PTAL, make sure the loop makes sense to you, I made multiple passes on it and may have missed something.

src/auth/src/token_cache.rs
```diff
@@ -93,63 +93,104 @@ async fn wait_for_next_token(
...
-
-        match expiry {
-            Ok(Some(expiry)) => {
-                let time_until_expiry = expiry.checked_duration_since(Instant::now());
-
-                match time_until_expiry {
+        if tx_token.receiver_count() == 0 {
+            // If there are no receivers, we can terminate the loop early.
+            return;
+        }
```
alvarowolfx, line 116: interesting. Wonder if we might need to look into more places where a similar pattern exists. Access Boundaries is one that I remember now that also has a background task

src/auth/src/token_cache.rs
```diff
@@ -93,63 +93,104 @@ async fn wait_for_next_token(
...
-                        }
+                    Some(d) if d < Instant::now() => {
+                        // Already expired, replace the cached token with the error.
+                        let _ = tx_token.send(Some(Err(e)));
+                    }
+                    Some(d) if d < short => {
+                        // First, wait until it expires, then replace the cached token with the
+                        // error, and then continue sleeping.
+                        tokio::time::sleep_until(d).await;
+                        let _ = tx_token.send(Some(Err(e)));
```
alvarowolfx, line 160: I was confuse on why would surface the error here, but I think that's correct. It's a transient error but the Retry policy was already exhausted here.

alvarowolfx (approved):
LGTM

src/auth/src/token_cache.rs
coryan: And because returning the latest refresh error is (I think) more informative than returning an expired token.

src/auth/src/token_cache.rs
coryan: Ack.

Issue #5826, opened 2026-06-03
Flake in spanner integration test


coryan:
```
Error: the service reports an error with code INTERNAL described as: Schema generation 0 was not registered with the Action Manager
Caused by:
    code: 'Internal error', message: "Schema generation 0 was not registered with the Action Manager"
failures:
    spanner::run_write_tests
test result: FAILED. 13 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.60s
error: test failed, to rerun pass `-p integration-tests-spanner --test driver`
```
https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/6d725db9-866d-4f0d-bba4-bcf775082f4b?project=rust-sdk-testing&e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=allow_workbench_image_override

PR #5827, opened 2026-06-03
test(spanner): disable flaky test

coryan:
See #5826 for the log lines.

### DIFF
--- a/tests/spanner/tests/driver.rs
+++ b/tests/spanner/tests/driver.rs
@@ -78,11 +78,14 @@ mod spanner {
         }
 
         async fn run_write_tests(db_client: &DatabaseClient) -> anyhow::Result<()> {
-            integration_tests_spanner::write::write_only_transaction(db_client).await?;
-            integration_tests_spanner::write::write(db_client).await?;
-            integration_tests_spanner::write::all_data_types_roundtrip(db_client).await?;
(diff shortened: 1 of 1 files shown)


tests/spanner/tests/driver.rs
```diff
@@ -78,11 +78,15 @@ mod spanner {
         }
 
         async fn run_write_tests(db_client: &DatabaseClient) -> anyhow::Result<()> {
-            integration_tests_spanner::write::write_only_transaction(db_client).await?;
-            integration_tests_spanner::write::write(db_client).await?;
-            integration_tests_spanner::write::all_data_types_roundtrip(db_client).await?;
-            integration_tests_spanner::write::all_data_types_parameter_binding(db_client).await?;
-            integration_tests_spanner::write::interval_parameter_binding(db_client).await?;
+            let _ = db_client;
+            #[cfg(false)] // TODO(#5826) - disable because it flakes
```
alvarowolfx, line 82: nit: the downside of using `#[cfg(false)]` is that is not gonna appears on test logs that it's disabled/ignored like we have with `#[ignore]`

coryan: Yeah, but these tests are defined inside macros. I have no idea how to use `#[ignore]` for just one of them.

haphungw (approved):

PR #5835, opened 2026-06-03
feat(bigquery-v2): add lro polling support for bq jobs - WIP - prototype

### LINKED ISSUE #5751: Prototype BQ Jobs LRO

haphungw:
Introduce Long-Running Operation (LRO) polling support for BigQuery Jobs. 

Unlike standard Google Cloud APIs, bq v2 jobs don't use the standard `google.longrunning.Operation` proto, so they cannot be automatically wrapped by the generator.

To fix this, we implement `DiscoveryOperation`, which lets the framework check the `DONE` state and pull the `job_id` straight out of `Job` and `GetQueryResultsResponse` objects. 

For the actual polling, we disable standard RPC retries with `NeverRetry` policy to avoid nested retry loops by letting the LRO's own retry policy handle transient network issues. 

`librarian.yaml` is updated to have `google-cloud-lro` as an dependency and leave the custom `operation.rs` alone when regenerating code.

Note on error handling: bq jobs don't throw standard RPC errors when they fail. Instead, they put the error details inside `JobStatus.error_result` field. Therefore, the LRO will tell us the poll completed successfully `(Ok(Job))` even when the job failed. We'll need to manually check `error_result` field to know if the job actually succeeded.

Fixes https://github.com/googleapis/google-cloud-rust/issues/5751

### DIFF
--- a/src/generated/cloud/bigquery/v2/src/lib.rs
+++ b/src/generated/cloud/bigquery/v2/src/lib.rs
@@ -110,3 +110,6 @@ pub(crate) use google_cloud_gax::client_builder::internal::new_builder as new_cl
 pub(crate) use google_cloud_gax::options::RequestOptions;
 pub(crate) use google_cloud_gax::options::internal::RequestBuilder;
 pub(crate) use google_cloud_gax::response::Response;
+
+#[allow(missing_docs)]
+pub mod operation;
--- a/src/generated/cloud/bigquery/v2/src/operation.rs
+++ b/src/generated/cloud/bigquery/v2/src/operation.rs
@@ -25,3 +25,62 @@ impl DiscoveryOperation for Job {
             })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::model::{ErrorProto, JobReference, JobStatus};
+
+    #[test]
+    fn test_done() {
+        let mut job = Job::default();
+        assert!(!job.done(), "missing status should not be done");
+
+        job.status = Some(JobStatus {
+            state: "RUNNING".to_string(),
+            ..Default::default()
+        });
+        assert!(!job.done(), "RUNNING should not be done");
+
+        job.status = Some(JobStatus {
+            state: "DONE".to_string(),
+            ..Default::default()
+        });
+        assert!(job.done(), "DONE should be done");
+    }
+
+    #[test]
+    fn test_name() {
+        let mut job = Job::default();
+        assert_eq!(job.name(), None, "missing job_reference should yield None");
+
+        job.job_reference = Some(JobReference {
+            job_id: "my_job".to_string(),
+            ..Default::default()
+        });
+        assert_eq!(job.name(), Some(&"my_job".to_string()), "should return job_id");
+    }
+
+    #[test]
+    fn test_error() {
+        let mut job = Job::default();
+        assert!(job.error().is_none(), "missing status should yield no error");
+
+        job.status = Some(JobStatus::default());
+        assert!(job.error().is_none(), "missing error_result should yield no error");
+
+        job.status = Some(JobStatus {
+            error_result: Some(ErrorProto {
+                message: "some error".to_string(),
+                ..Default::default()
+            }),
+            ..Default::default()
+        });
+
+        let err = job.error().expect("should have error");
+        assert_eq!(err.message, "some error");
+        // Unknown code is 2.
+        assert_eq!(err.code, (google_cloud_gax::error::rpc::Code::Unknown as i32).into());
+    }
+}
+
--- a/src/generated/cloud/bigquery/v2/src/operation.rs
+++ b/src/generated/cloud/bigquery/v2/src/operation.rs
@@ -1,0 +1,27 @@
+use crate::model::Job;
+use google_cloud_lro::internal::DiscoveryOperation;
+
+impl DiscoveryOperation for Job {
+    fn done(&self) -> bool {
+        self.status
+            .as_ref()
+            .map(|s| s.state == "DONE")
+            .unwrap_or(false)
+    }
+
+    fn name(&self) -> Option<&String> {
+        self.job_reference.as_ref().map(|r| &r.job_id)
+    }
+
+    fn error(&self) -> Option<google_cloud_gax::error::rpc::Status> {
+        self.status
+            .as_ref()
+            .and_then(|s| s.error_result.as_ref())
+            .map(|err| {
+                let mut status = google_cloud_gax::error::rpc::Status::default();
+                status.code = (google_cloud_gax::error::rpc::Code::Unknown as i32).into();
+                status.message = err.message.clone();
+                status
+            })
+    }
+}
--- a/tests/bigquery/src/lib.rs
+++ b/tests/bigquery/src/lib.rs
@@ -21,6 +21,7 @@ use google_cloud_bigquery_v2::model::{
     Dataset, DatasetReference, Job, JobConfiguration, JobConfigurationQuery, JobReference,
 };
 use google_cloud_gax::{error::rpc::Code, paginator::ItemPaginator};
+use google_cloud_lro::Poller;
 use google_cloud_test_utils::runtime_config::project_id;
 use rand::{RngExt, distr::Alphanumeric};
 
@@ -198,7 +199,8 @@ pub async fn job_service() -> Result<()> {
                         .set_query(JobConfigurationQuery::new().set_query(query)),
                 ),
         )
-        .send()
+        .poller()
+        .until_done()
         .await?;
     println!("CREATE JOB = {job:?}");
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2118,6 +2118,7 @@ dependencies = [
  "bytes",
  "google-cloud-gax",
  "google-cloud-gax-internal",
+ "google-cloud-lro",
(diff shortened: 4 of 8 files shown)


src/generated/cloud/bigquery/v2/src/lib.rs
```diff
@@ -79,6 +79,9 @@ pub mod client;
 /// Request builders.
 pub mod builder;
 
+#[allow(missing_docs)]
+pub mod operation;
```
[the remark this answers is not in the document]
haphungw: ignoring.
`#[allow(missing_docs)]` is hardcoded into librarian global template for `keep` modules. can't remove it from the template.

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -0,0 +1,154 @@
...
+}
+
+pub trait InsertJobBuilderExt {
+    fn poller(
+        self,
+        client: &JobService,
+        project_id: impl Into<String>,
+        location: Option<String>,
+    ) -> impl Poller<Job, Job>;
+}
```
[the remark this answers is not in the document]

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -0,0 +1,154 @@
...
+
+pub trait GetQueryResultsBuilderExt {
+    fn poller(
+        self,
+        client: &JobService,
+        project_id: impl Into<String>,
+        location: Option<String>,
+    ) -> impl Poller<GetQueryResultsResponse, GetQueryResultsResponse>;
+}
```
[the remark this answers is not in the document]

tests/bigquery/src/lib.rs
```diff
@@ -196,12 +198,24 @@ pub async fn job_service() -> Result<()> {
...
+    // Also test polling for query results
+    let results_poller = client
+        .get_query_results()
+        .set_project_id(&project_id)
+        .set_job_id(&job_id)
+        .poller(&client, &project_id, None);
+
+    let results = results_poller.until_done().await?;
+    println!("QUERY RESULTS (POLLED) = {results:?}");
+    assert_eq!(results.job_complete, Some(true));
```
alvarowolfx: this looks super interesting. Make life much easier to wait for a query to be executed, even with GAPICs. 

One thing that's looks odd is that `.poller` right now is accepting the same parameters used right before it, like `client` and `project_id`. Our other `poller` don't have that and are just a method without any parameters. 

https://googleapis.github.io/google-cloud-rust/working_with_long_running_operations.html#automatically-polling-a-long-running-operation

I wonder if we can access the client from inside GetQueryResult/InsertJob (https://github.com/googleapis/google-cloud-rust/blob/1af353501839a688d13f75a70639c1f28ebe81c3/src/generated/cloud/bigquery/v2/src/builder.rs#L1335) and don't need that.

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -0,0 +1,213 @@
...
+///     Ok(())
+/// }
+/// ```
+pub trait GetQueryResultsBuilderExt {
+    /// Returns a poller to monitor the status of the query results.
+    fn poller(
+        self,
+        client: &JobService,
+        project_id: impl Into<String>,
+        location: Option<String>,
```
alvarowolfx: as mentioned on another comments, I think accepting those parameters here again sounds like duplication, since the get_query_results() call accepts a `set_project_id` and `set_location`

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -0,0 +1,213 @@
...
+    ) -> impl Poller<Job, Job>;
+}
+
+impl InsertJobBuilderExt for InsertJob {
+    fn poller(
+        self,
+        client: &JobService,
+        project_id: impl Into<String>,
+        location: Option<String>,
```
alvarowolfx: as mentioned on another comments, I think accepting those parameters here again sounds like duplication, since the get_job() call accepts a `set_project_id` and `set_location`, and maybe we can grab the JobService from inside InsertJob https://github.com/googleapis/google-cloud-rust/blob/main/src/generated/cloud/bigquery/v2/src/builder.rs#L937

alvarowolfx (commented):
just some thoughts by reading it real quick. I think looks promising, but I wonder if we can make it even easier by not requiring extra parameters on the `.poller` method.

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -0,0 +1,213 @@
...
+            async move {
+                let mut b = client
+                    .get_query_results()
+                    .set_project_id(project_id)
+                    .set_job_id(name);
+                if let Some(loc) = location {
+                    b = b.set_location(loc);
+                }
+                let mut options = google_cloud_gax::options::RequestOptions::default();
+                options.set_retry_policy(google_cloud_gax::retry_policy::NeverRetry);
```
alvarowolfx: we need to have a way to pass a retry policy, because we can actually need to retry in some scenarios (see https://docs.cloud.google.com/bigquery/docs/error-messages) and even crazier, in some cases we only detect that the job has to be retried by pooling it. In some cases we need to re send the query (call jobs.query or jobs.insert again) with a different ID (see jobBackendError at https://docs.cloud.google.com/bigquery/docs/error-messages). We don't need to solve this right away, just pointing out that we need to take into account that we do need to customize retry policies here.

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -0,0 +1,213 @@
...
+
+        let query = move |name: String| {
+            let client = client_clone.clone();
+            let project_id = project_id.clone();
+            let location = location.clone();
+            async move {
+                let mut b = client
+                    .get_query_results()
+                    .set_project_id(project_id)
+                    .set_job_id(name);
```
alvarowolfx: We should pass maxResults=0 when using getQueryResults to wait for a query to finish. For large rows, it can add many seconds to the request time, even when no rows are actually returned.

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -0,0 +1,213 @@
...
+            let req = self;
+            async move { req.send().await }
+        };
+
+        let query = move |name: String| {
+            let client = client_clone.clone();
+            let project_id = project_id.clone();
+            let location = location.clone();
+            async move {
+                let mut b = client.get_job().set_project_id(project_id).set_job_id(name);
```
alvarowolfx: this is using `jobs.get` to poll for all job types, but `jobs.getQueryResults` is much better for query jobs. So maybe we need to check if this is a query job and use a different method (or call the jobs.getQueryResults poller)

tests/bigquery/src/lib.rs
alvarowolfx: looking more closely, this is polling the job two times, one using `jobs.get` via `let job = poller.until_done().await?;` and another one via `jobs.getQueryResults` via `let results = results_poller.until_done().await?;`. This is not very useful, since after is done by checking with `jobs.get`, it is already going to be done, so the other poller is not required. 

We only need one poller for a job, but different job types, might use different polling mechanisms. 
* Query jobs - jobs.getQueryResults
* Copy/Load/Extract - jobs.get

so we only need this for all jobs:

Query job:
```
let poller = client
    .insert_job()
    .set_project_id(&project_id)
    .set_job(
        Job::new()
            .set_job_reference(JobReference::new().set_job_id(&job_id))
            .set_configuration(
                JobConfiguration::new()
                    .set_labels([(INSTANCE_LABEL, "true")])
                    .set_query(JobConfigurationQuery::new().set_query(query)),
            ),
    )
    .poller();

let job = poller.until_done().await?;
println!("CREATE QUERY JOB (POLLED) = {job:?}");
```

Copy job:
```
let src_table_id = "some_table";
let dest_table_id = "dest_table";
let poller = client
    .insert_job()
    .set_project_id(&project_id)
    .set_job(
        Job::new()
            .set_job_reference(JobReference::new().set_job_id(&job_id))
            .set_configuration(
                JobConfiguration::new()
                    .set_labels([(INSTANCE_LABEL, "true")])
                    .set_copy(JobConfigurationCopy::new(). set_source_table(src_table_id). set_destination_table(dest_table_id)),
            ),
    )
    .poller();

let job = poller.until_done().await?;
println!("CREATE COPY JOB (POLLED) = {job:?}");
```

alvarowolfx (commented):
more thoughts

tests/bigquery/src/lib.rs
alvarowolfx: as we discussed internally, even better for queries, they can just insert it and poll for getQueryResults:

```
let job = client
    .insert_job()
    .set_project_id(&project_id)
    .set_job(
        Job::new()
            .set_job_reference(JobReference::new().set_job_id(&job_id))
            .set_configuration(
                JobConfiguration::new()
                    .set_labels([(INSTANCE_LABEL, "true")])
                    .set_query(JobConfigurationQuery::new().set_query(query)),
            ),
    )
    .send()
    .await?;

println!("CREATE QUERY JOB = {job:?}");

let results_poller = client
    .get_query_results()
    .set_project_id(&project_id)
    .set_job_id(&job_id)
    .poller(&client, &project_id, None);

let results = results_poller.until_done().await?;
println!("QUERY RESULTS (POLLED) = {results:?}");
```

PR #5840, opened 2026-06-04
test(spanner): retry internal errors from the Emulator

### LINKED ISSUE #5826: Flake in spanner integration test

olavloite:
The latest version of the Spanner Emulator can return an INTERNAL error if DDL statements and read/write transactions are executed concurrently. This change adds a specific error to the set of errors that are retried by the Spanner client when it is running agains the Emulator. This additional check is only done if the client is connected to the Emulator, and not if the client is connected to real Spanner.

Note: The test that failed and triggered the creation of this ticket, is not the actual culprit, and a similar failure could in theory happen for any of the other tests for the Spanner client that execute a read/write transaction.

Fixes #5826

### DIFF
--- a/src/spanner/src/client.rs
+++ b/src/spanner/src/client.rs
@@ -20,7 +20,9 @@ use crate::model::{
 };
 use crate::server_streaming::builder;
 use gaxi::options::{ClientConfig, Credentials};
+use google_cloud_auth::credentials::anonymous;
 use google_cloud_gax::client_builder::ClientBuilder as GaxClientBuilder;
+use google_cloud_gax::client_builder::internal::new_builder;
 use google_cloud_gax::options::{
     RequestOptions as GaxRequestOptions, internal::RequestOptionsExt as _,
 };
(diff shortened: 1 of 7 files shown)


coryan (approved):
I know this is hand-wavy, but I wonder if you could have implemented this with a retry policy decorator, and maybe have set that policy on the tests using the emulator or just automatically when the emulator is enabled.

sakthivelmanii (approved):

olavloite:
> I know this is hand-wavy, but I wonder if you could have implemented this with a retry policy decorator, and maybe have set that policy on the tests using the emulator or just automatically when the emulator is enabled.

I considered (and tried) that. But: We do not want to expose a public option for setting the retry policy for when a __transaction__ should be retried. And in the end, as this is for the integration tests, any such `RetryPolicy` option would need to be public.

PR #5843, opened 2026-06-04
feat(o11y): add error extraction to `DiscoveryOperation` and fix macro compatibility

haphungw:
Introduce telemetry injections for Discovery LROs.

This ensures Discovery LROs can record rich error details upon completion, matching the behavior of standard LROs, and provides a cleaner interface for code generation.

Used at https://github.com/googleapis/librarian/pull/6304.

### DIFF
--- a/src/lro/src/internal/tracing.rs
+++ b/src/lro/src/internal/tracing.rs
@@ -67,7 +67,31 @@ impl LroRecorder {
     pub fn attempt_count(&self) -> Option<u32> {
         self.attempt_count
     }
+}
 
+/// Helper macro to record telemetry for Discovery LROs.
+#[macro_export]
+#[doc(hidden)]
+macro_rules! record_discovery_polling_result {
+    ($span:expr, $op:expr) => {
+        let span = &$span;
+        let op = &$op;
+        let done = $crate::internal::DiscoveryOperation::done(op);
+        span.record("gcp.longrunning.done", done);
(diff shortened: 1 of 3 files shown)


westarle (changes requested):
Could you please add tests for these?

src/lro/src/internal/tracing.rs
```diff
@@ -98,6 +123,28 @@ impl LroRecorder {
...
+        {
+            if let Some(recorder) = $crate::LroRecorder::current() {
+                if let Some(attempt) = recorder.attempt_count() {
+                    let span = &$span;
+                    span.record("gcp.longrunning.poll_attempt_count", attempt);
+                    span.record("gcp.longrunning.done", false);
+                }
+            }
+        }
+        #[cfg(not(google_cloud_unstable_tracing))]
```
westarle: is this codepath reachable in this tracing module?

westarle (commented):
I'm not seeing the tests yet, maybe missing a commit?

src/lro/src/internal/tracing.rs
haphungw: you're right, these gates are dead code. i removed them.

westarle (commented):
this LGTM but I would like another Rust team member to review adding a dep from lro to gaxi before merging.

westarle (approved):

Issue #4677, opened 2026-02-13
Unexpected EOF / TLS close_notify error when calling Vertex AI generate content API


nvnmandadhi:
**Description**

Intermittent (or persistent) connection failure when attempting to send a generateContent request to the Vertex AI endpoint. The client library fails during the TLS handshake/session phase, reporting an unexpected EOF.

**Error sending request for url**

```
(https://us-central1-aiplatform.googleapis.com/v1/projects/<PROJECT_ID>/locations/us-central1/publishers/google/models/gemini-2.5-flash-lite:generateContent?%24alt=json%3Benum-encoding%3Dint)
client error (SendRequest)
connection error
peer closed connection without sending TLS close_notify: https://docs.rs/rustls/latest/rustls/manual/_03_howto/index.html#unexpected-eof
```

coryan:
What method?

coryan:
Are you printing the debug output for the error? (that is, using `{:?}`) or the display output (that is, using `{}`)? I think our errors are much better with `{:?}` and may have more details to troubleshoot things.

coryan:
This is a `PredictionService::generate_content()` call. The client is configured with `AlwaysRetry.with_attempt_limit(3)`, and the request included `.with_attempt_timeout(Duration::from_secs(300))` ...

nvnmandadhi:
> This is a `PredictionService::generate_content()` call. The client is configured with `AlwaysRetry.with_attempt_limit(3)`, and the request included `.with_attempt_timeout(Duration::from_secs(300))` ...

Updated the description including client details.

coryan:
I left this running to see if it will repro:

https://github.com/googleapis/google-cloud-rust/compare/main...coryan:google-cloud-rust:repro-bug-4677

If you think we should change the repro in any way, please let us know.

coryan:
No repro in 1,000 attempts.  We need your help to change the repro: I suspect we are not using large-enough inputs or outputs.

dbolduc:
I said I have seen this error before. I should mention it was from a bidirectional gRPC stream. (of course `reqwest` and `tonic` are both built on top of `hyper` / `h2`)

```
[...]

PR #5866, opened 2026-06-10
fix(auth): retry more transient errors

### LINKED ISSUE #5864: auth: one misclassified transport error permanently disables token refresh for the process lifetime
`google-cloud-auth` (observed on 1.6.0 in production; both code paths verified unchanged at 1.12.0). Related: #5809 — the fix for that issue preserves the previous good token across a failed refresh, but does not address what's below; it only delays the onset until the preserved token expires.

This is two composable bugs; either alone is survivable, together they take a healthy process permanently offline after a single torn network read.

### Bug 1 — `is_retryable` classifies status-less transport errors as permanent (`src/auth/src/errors.rs:128`)

```rust
fn is_retryable(err: &reqwest::Error) -> bool {
    if err.is_connect() {
        return true;
    }
    match err.status() {
        Some(code) => is_retryable_code(code),
        None => false,   // <-- the bug
    }
}
```

A connection that establishes and then dies mid-response — `hyper::Error(IncompleteMessage)`, connection reset by peer, an idle keep-alive race — has no HTTP status and is not a connect error, so it falls through to `None => false` and becomes `CredentialsError { is_transient: false }`.

This contradicts the crate's own `is_transient` documentation ("Timeouts and network problems are good candidates for `is_transient() == true`"). A torn read is *more* transient than connection-refused, which is already retried via the `is_connect()` arm.

### Bug 2 — the token cache makes any "permanent" error unrecoverable for the process lifetime (`src/auth/src/token_cache.rs:121`)

The background refresh task is the only code that ever fetches tokens; callers read a watch channel it publishes into. On an error classified non-transient it waits out the current token's validity, publishes the error, and **returns** — the task exits permanently:

```rust
Err(e) if !e.is_transient() => {
    // ... wait until the current token expires ...
    let _ = tx_token.send(Some(Err(e)));
    return;
}
```

Every subsequent `headers()`/`token()` call returns the latched error instantly with zero network I/O ("Subsequent calls with this credential will also fail"). There is no recovery path: no API to reset the cache, no periodic re-probe — nothing short of constructing a new `Credentials`, which most applications only do at startup. The code comment acknowledges the risk ("If the error was misclassified as permanent, that is a bug in the retry policy and better fixed there") — but the blast radius of one misclassification is a permanently dead credential, which is disproportionate.

### Production impact

Two incidents in one week, on different services, from ordinary network weather:

1. A long-lived cache server had a single token-exchange request to a local (in-pod, healthy) STS endpoint die mid-response during a transient network event. Bug 1 classified it permanent; Bug 2 latched it. The process could not fetch a GCS token for **22 hours** — the token endpoint was healthy the entire time and was *never contacted again* (its logs show zero requests from the affected client post-latch) — until the pod was manually restarted.
2. Two API-server processes hit the same latch via `Connection reset by peer (os error 104)` from the GCE metadata server during node-level network degradation. Other (non-Rust) Google auth clients on the same hosts retried through the same weather and recovered; the Rust SDK latched. Impact was masked only because routine worker recycling happens to replace the affected processes.

The #5809 fix would have changed incident 1's onset by ~50 minutes (until the preserved token expired); the dead refresh task is the unrecoverable part.

### Suggested fixes

1. In `is_retryable`, classify status-less, non-connect `reqwest` errors as **transient** — they are network-layer failures squarely within the `is_transient` doc's own definition.
2. Independently: don't `return` from the refresh loop on permanent errors. Keep retrying at a long backoff (capped, e.g. minutes), or expose a reset/re-probe hook on `Credentials`. A process-lifetime latch is never the right outcome for a credential that worked five minutes earlier — and as the in-code comment notes, the cost of a retry-policy misclassification should be bounded, not terminal.

dbolduc:
Classify more auth errors as transient, as a single permanent error can poison the token cache.

Note that this PR only affects auth. We should do a similar thing in our `gaxi::http::ReqwestClient`.

### Implementation nits:

I duplicated the `as_inner` function from `gaxi` twice in this PR. I liked the idea of an e2e test outside of the crate. And I wanted to verify the errors produced by the bad servers, not just that we classify them as transient.

Fixes #5864, related to #4677

### DIFF
--- a/src/auth/src/errors.rs
+++ b/src/auth/src/errors.rs
@@ -145,10 +145,10 @@ fn is_retryable(err: &reqwest::Error) -> bool {
         }
         if let Some(io_err) = as_inner::<std::io::Error, _>(hyper_err) {
             use std::io::ErrorKind;
-            match io_err.kind() {
-                ErrorKind::ConnectionReset | ErrorKind::ConnectionAborted => return true,
-                _ => return false,
-            }
+            return matches!(
+                io_err.kind(),
+                ErrorKind::ConnectionReset | ErrorKind::ConnectionAborted
+            );
         }
         return false;
     }
--- a/src/auth/src/errors.rs
+++ b/src/auth/src/errors.rs
@@ -133,10 +133,29 @@ fn is_retryable(err: &reqwest::Error) -> bool {
     if err.is_connect() {
         return true;
     }
-    match err.status() {
-        Some(code) => is_retryable_code(code),
-        None => false,
+    if err.is_request() {
+        // This implementation is inspired by the ideas from:
+        // http://github.com/TrueLayer/reqwest-middleware/blob/main/reqwest-retry/src/retryable_strategy.rs#L138-L184
+
+        let Some(hyper_err) = as_inner::<hyper::Error, _>(err) else {
+            return false;
+        };
+        if hyper_err.is_incomplete_message() || hyper_err.is_canceled() {
+            return true;
+        }
+        if let Some(io_err) = as_inner::<std::io::Error, _>(hyper_err) {
+            use std::io::ErrorKind;
+            match io_err.kind() {
+                ErrorKind::ConnectionReset | ErrorKind::ConnectionAborted => return true,
+                _ => return false,
+            }
+        }
+        return false;
     }
+    if let Some(code) = err.status() {
+        return is_retryable_code(code);
+    }
+    false
 }
 
 fn is_retryable_code(code: StatusCode) -> bool {
@@ -149,6 +168,25 @@ fn is_retryable_code(code: StatusCode) -> bool {
         | StatusCode::TOO_MANY_REQUESTS => true,
         _ => false,
     }
+}
+
+/// Extract the first source error of type `T`.
+fn as_inner<T, E>(error: &E) -> Option<&T>
+where
+    T: std::error::Error + 'static,
+    E: std::error::Error,
+{
+    let mut e = error.source()?;
+    // Prevent infinite loops due to cycles in the `source()` errors. This seems
+    // unlikely, and it would require effort to create, but it is easy to
+    // prevent.
+    for _ in 0..32 {
+        if let Some(value) = e.downcast_ref::<T>() {
+            return Some(value);
+        }
+        e = e.source()?;
+    }
+    None
 }
 
 #[cfg(test)]
--- a/src/auth/src/errors.rs
+++ b/src/auth/src/errors.rs
@@ -126,36 +126,17 @@ pub(crate) fn is_gax_error_retryable(err: &GaxError) -> bool {
 }
 
 fn is_retryable(err: &reqwest::Error) -> bool {
-    // Connection errors are transient more often than not. A bad configuration
-    // can point to a non-existing service, and that will never recover.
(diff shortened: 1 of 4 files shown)


src/auth/tests/transient.rs
```diff
@@ -0,0 +1,141 @@
...
+        // unlikely, and it would require effort to create, but it is easy to
+        // prevent.
+        for _ in 0..32 {
+            if let Some(value) = e.downcast_ref::<T>() {
+                return Some(value);
+            }
+            e = e.source()?;
+        }
+        None
+    }
```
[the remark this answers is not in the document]
dbolduc: > consider making it public

That's exactly what I want to avoid.

src/auth/src/errors.rs
```diff
@@ -133,10 +133,26 @@ fn is_retryable(err: &reqwest::Error) -> bool {
...
+        };
+        if hyper_err.is_incomplete_message() || hyper_err.is_canceled() {
+            return true;
+        }
+        if let Some(io_err) = as_inner::<std::io::Error, _>(hyper_err) {
+            use std::io::ErrorKind;
+            match io_err.kind() {
+                ErrorKind::ConnectionReset | ErrorKind::ConnectionAborted => return true,
+                _ => return false,
+            }
```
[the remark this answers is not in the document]
dbolduc: Good call, done.

src/auth/src/errors.rs
```diff
@@ -133,10 +133,29 @@ fn is_retryable(err: &reqwest::Error) -> bool {
...
+            return true;
+        }
+        if let Some(io_err) = as_inner::<std::io::Error, _>(hyper_err) {
+            use std::io::ErrorKind;
+            return matches!(
+                io_err.kind(),
+                ErrorKind::ConnectionReset | ErrorKind::ConnectionAborted
+            );
+        }
+        return false;
```
coryan: Why not 

```suggestion
    if err.is_request() {
        return true;
```

All HTTP errors (without a status code) seem retryable to me... None of the credential requests have side-effects, they are all idempotent by definition...

coryan (changes requested):

src/auth/src/errors.rs
dbolduc: I am happy to go with the simpler logic. We would definitely rather make the mistake of classifying a permanent error as transient than the other way around.

Just browsing the `hyper::Error` API, this one seems like it can only ever fail, although we probably would never hit it in practice: https://docs.rs/hyper/latest/hyper/struct.Error.html#method.is_parse_too_large

Does that change your mind at all? If not, I will simplify.

coryan: Agree, `is_parse_too_large` sounds like a thing you cannot recover from, it also seems to be a server-side error (meaning hyper servers create this, not hyper clients). Two alternatives:

1) Return retryable for everything except `is_parse_too_large()` (and maybe grow the list of things that are definitely permanent errors over time)
2) Ignore the problem and let the retry loop and token expiration take care of the problem.

I think (2) is safer for now...

dbolduc: Ack, and I agree. Done, thanks.

dbolduc: I left in the tests for the two specific types of request errors.

coryan (approved):

Issue #5844, opened 2026-06-05
Implement BigQuery Query Client


alvarowolfx:
Internal go/cloud-rust:bigquery-query-design

PR #5869, opened 2026-06-10
impl(bigquery): add client builder

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery/src/client.rs
+++ b/src/bigquery/src/client.rs
@@ -57,19 +57,3 @@ impl BigQuery {
         Ok(BigQuery { job_service })
     }
 }
-
-#[cfg(test)]
-mod tests {
-    use super::BigQuery;
-    use google_cloud_auth::credentials::anonymous::Builder as Anonymous;
-
-    #[tokio::test]
-    async fn test_bigquery_builder() -> anyhow::Result<()> {
-        let client = BigQuery::builder()
-            .with_credentials(Anonymous::new().build())
-            .build()
-            .await?;
-        assert!(std::sync::Arc::strong_count(&client.job_service) >= 1);
-        Ok(())
-    }
-}
--- a/src/bigquery/src/client.rs
+++ b/src/bigquery/src/client.rs
@@ -57,3 +57,18 @@ impl BigQuery {
         Ok(BigQuery { job_service })
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::BigQuery;
+    use google_cloud_auth::credentials::anonymous::Builder as Anonymous;
+
+    #[tokio::test]
+    async fn test_bigquery_builder() -> anyhow::Result<()> {
+        let _client = BigQuery::builder()
+            .with_credentials(Anonymous::new().build())
+            .build()
+            .await?;
+        Ok(())
+    }
+}
--- a/src/bigquery/src/client.rs
+++ b/src/bigquery/src/client.rs
@@ -1,0 +1,75 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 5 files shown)


src/bigquery/src/client.rs
```diff
@@ -0,0 +1,75 @@
...
+
+use crate::ClientBuilderResult as BuilderResult;
+use crate::client_builder::ClientBuilder;
+use google_cloud_bigquery_v2::client::JobService;
+use std::sync::Arc;
+
+/// A high-level BigQuery client for executing queries and managing jobs.
+#[derive(Clone, Debug)]
+pub struct BigQuery {
+    #[allow(dead_code)]
```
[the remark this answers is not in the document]
alvarowolfx: gonna be used right in the next PR.

src/bigquery/src/client.rs
```diff
@@ -0,0 +1,75 @@
...
+    use super::BigQuery;
+    use google_cloud_auth::credentials::anonymous::Builder as Anonymous;
+
+    #[tokio::test]
+    async fn test_bigquery_builder() -> anyhow::Result<()> {
+        let client = BigQuery::builder()
+            .with_credentials(Anonymous::new().build())
+            .build()
+            .await?;
+        assert!(std::sync::Arc::strong_count(&client.job_service) >= 1);
```
dbolduc: This is a meaningless assertion? It can never be 0.

Remove?

src/bigquery/src/client_builder.rs
```diff
@@ -0,0 +1,154 @@
...
+impl ClientBuilder {
+    /// Creates a new default `ClientBuilder`.
+    pub fn new() -> Self {
+        Self {
+            config: ClientConfig::default(),
+        }
+    }
+
+    /// Sets the BigQuery v2 endpoint.
+    pub fn with_endpoint<V: Into<String>>(mut self, v: V) -> Self {
```
dbolduc: Q: what is the plan for the bigquerystorage endpoint?

Also, a link would probably be helpful to explain what is "BigQuery v2". I dont expect customers to know off the top of their heads

src/bigquery/src/client_builder.rs
```diff
@@ -0,0 +1,154 @@
...
+    pub fn with_backoff_policy<V: Into<BackoffPolicyArg>>(mut self, v: V) -> Self {
+        self.config.backoff_policy = Some(v.into().into());
+        self
+    }
+
+    /// Configure the retry throttler.
+    pub fn with_retry_throttler<V: Into<RetryThrottlerArg>>(mut self, v: V) -> Self {
+        self.config.retry_throttler = v.into().into();
+        self
+    }
```
dbolduc: My default would be to start with a smaller API surface and add these in when we know they are needed.

If we want any sort of custom policies, we might regret exposing them.

dbolduc (approved):

src/bigquery/src/client_builder.rs
alvarowolfx: For BigQuery Storage we are going to add a `with_storage_endpoint`. I removed it for now because only makes sense to add it once we start supporting Storage acceleration. Having it here might mislead customer to think it is already supported.

also, in regards to docs, I was gonna leave to have better docs later when we can add examples and things like that. But will improve it a bit by adding a link

src/bigquery/src/client_builder.rs
alvarowolfx: ack, I added the retry stuff as a last minute thing on this PR, will remove and add later.

src/bigquery/src/client.rs
alvarowolfx: yeah, was hard to find any meaningful test for this module haha will remove this as well

src/bigquery/src/client.rs
```diff
@@ -57,19 +57,3 @@ impl BigQuery {
...
-    use super::BigQuery;
-    use google_cloud_auth::credentials::anonymous::Builder as Anonymous;
-
-    #[tokio::test]
-    async fn test_bigquery_builder() -> anyhow::Result<()> {
-        let client = BigQuery::builder()
-            .with_credentials(Anonymous::new().build())
-            .build()
-            .await?;
-        assert!(std::sync::Arc::strong_count(&client.job_service) >= 1);
```
dbolduc: I think the test is good to have, it is just the `assert!()` line that was kind of dead code.

dbolduc (approved):

PR #5870, opened 2026-06-10
impl(bigquery): add query runner builder

alvarowolfx:
Towards #5844


src/bigquery/src/error.rs
```diff
@@ -0,0 +1,90 @@
...
+    },
+
+    /// The underlying RPC failed.
+    #[non_exhaustive]
+    #[error("the operation failed. RPC error: {source}")]
+    Rpc {
+        /// The error returned by the service for the request.
+        #[source]
+        source: Arc<Error>,
+    },
```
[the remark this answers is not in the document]
alvarowolfx: I've copied this pattern from pubsub `Rpc` error https://github.com/googleapis/google-cloud-rust/blob/817555e2fa73af75fb0b07d210bb13d5e6d25c9d/src/pubsub/src/error.rs#L64

@dbolduc I don't recall exactly why we decided to make it an Arc here.

src/bigquery/src/generated/run_query_request.rs
alvarowolfx, line 1: Clippy is not happy with these generated files. Will add those allow tags on sidekick.

edit: https://github.com/googleapis/librarian/pull/6373

src/bigquery/src/error.rs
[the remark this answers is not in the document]
dbolduc: In Pub/Sub, N messages are batched into one RPC, and the RPC gets 1 error. We wanted to share that error across N messages.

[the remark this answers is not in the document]
alvarowolfx: gotcha. In Bigquery case here it's a 1:1 mapping (one Query operation -> one RPC), so I think can drop then Arc.

alvarowolfx:
issues on this PR will be solved when https://github.com/googleapis/google-cloud-rust/pull/5874 is merged and also I split errors to be merged on a separated PR: https://github.com/googleapis/google-cloud-rust/pull/5873

src/bigquery/src/query/run_query.rs
```diff
@@ -0,0 +1,177 @@
...
+}
+
+impl RunQuery {
+    /// Creates a new `RunQuery` builder for the given SQL query.
+    pub fn new(job_service: Arc<JobService>, sql: String) -> Self {
+        Self {
+            job_service,
+            request: RunQueryRequest::default()
+                .set_query(sql)
+                .set_use_legacy_sql(wkt::BoolValue::from(false)),
```
dbolduc: non-blocking Q: default to job creation mode optional?

src/bigquery/src/query/run_query.rs
```diff
@@ -0,0 +1,177 @@
...
+#[derive(Clone)]
+pub struct RunQuery {
+    pub(crate) job_service: Arc<JobService>,
+    pub(crate) request: RunQueryRequest,
+    pub(crate) project_id: Option<String>,
+}
+
+impl RunQuery {
+    /// Creates a new `RunQuery` builder for the given SQL query.
+    pub fn new(job_service: Arc<JobService>, sql: String) -> Self {
```
dbolduc: nit: `pub(crate)` ?

src/bigquery/src/query/run_query.rs
```diff
@@ -0,0 +1,177 @@
...
+                .set_query_request(query_request);
+
+            // TODO(#5844): implement jobs.query query execution
+            unimplemented!("jobs.query query execution not yet implemented");
+        }
+    }
+}
+
+include!("../generated/run_query_builder.rs");
+include!("../generated/run_query_request.rs");
```
dbolduc, line 91: style comment: we should prefer to have a `mod generated` over `include!()`, e.g.: https://github.com/googleapis/google-cloud-rust/blob/main/src/storage/src/generated.rs

The compiler can do smarter things if we build the module tree vs. pasting code in. (I think, I could understand the why better).

I think you would have to mess around with the modules of the generated types though, so feel free to take a TODO and not block this PR.

src/bigquery/src/query/run_query.rs
```diff
@@ -0,0 +1,177 @@
...
+        let _ = run_query.run().await;
+    }
+
+    #[test]
+    fn test_force_job_path() {
+        let job_service = create_job_service();
+        let mut run_query = RunQuery::new(job_service, "SELECT 1".to_string());
+        assert!(!run_query.request.force_job_path());
+
+        // Set allow_large_results
```
dbolduc: remove? comment is not adding to the code.

src/bigquery/src/query/run_query.rs
```diff
@@ -0,0 +1,177 @@
...
+    }
+
+    /// Sets the project ID to override the default client project ID.
+    pub fn with_project_id<S: Into<String>>(mut self, project_id: S) -> Self {
+        self.project_id = Some(project_id.into());
+        self
+    }
+
+    /// Executes the SQL query, routing internally to `jobs.query` (fast path)
+    /// or `jobs.insert` (job path) depending on configured fields.
```
dbolduc: nit:
```suggestion
    /// Executes the SQL query
    ///
    /// The implementation routes internally to `jobs.query` (fast path)
    /// or `jobs.insert` (job path) depending on configured fields. If the fast path is available, the client library takes it. If not, it falls back to creating a job, which is typically slower.
```

^ probably need to reformat. And links are helpful.

dbolduc (approved):
I think the `new()` should be `pub(crate)`, but nothing else is blocking.

src/bigquery/src/query/run_query.rs
alvarowolfx: ack and added a TODO

src/bigquery/src/query/run_query.rs
alvarowolfx: good catch, my internal draft had it as pub(crate), not sure when I moved to `pub` 🤦

src/bigquery/src/query/run_query.rs
```diff
@@ -0,0 +1,191 @@
...
+                google_cloud_bigquery_v2::model::DataFormatOptions::new()
+                    .set_use_int64_timestamp(true),
+            );
+            let _req = PostQueryRequest::new()
+                .set_project_id(project_id)
+                .set_query_request(query_request);
+
+            // TODO(#5844): implement jobs.query query execution
+            unimplemented!("jobs.query query execution not yet implemented");
+        }
```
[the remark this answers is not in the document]
alvarowolfx: will improve that later when actually doing query execution

PR #5871, opened 2026-06-10
feat(o11y): propagate LRO destination ID to child spans

haphungw:
Introduce stateful propagation of the LRO destination ID (resource ID/operation name) across LRO polling attempts inside `google-cloud-lro`.

### DIFF
--- a/src/lro/src/internal/tracing.rs
+++ b/src/lro/src/internal/tracing.rs
@@ -31,6 +31,7 @@ tokio::task_local! {
 pub struct LroRecorder {
     span: Span,
     attempt_count: Option<u32>,
+    destination_id: std::sync::Arc<std::sync::Mutex<Option<String>>>,
 }
 
 impl LroRecorder {
@@ -39,6 +40,7 @@ impl LroRecorder {
         Self {
             span,
             attempt_count: None,
+            destination_id: std::sync::Arc::new(std::sync::Mutex::new(None)),
         }
     }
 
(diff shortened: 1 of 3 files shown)


src/lro/src/internal/tracing.rs
westarle: I'm not sure its as immutable anymore as it mutates the destination ID.

src/lro/src/internal/tracing.rs
```diff
@@ -31,6 +31,7 @@ tokio::task_local! {
 pub struct LroRecorder {
     span: Span,
     attempt_count: Option<u32>,
+    destination_id: std::sync::Arc<std::sync::Mutex<Option<String>>>,
```
westarle: can this be a oncelock to simplify reads (I feel like we did this somewhere before?)

haphungw: you're right, fixed!

src/lro/src/internal/tracing.rs
haphungw: fixed the comment.

As for the immutability, when I was planning this in the first place, I forgot that the destination ID is only available after the first response comes back. Only then can we propagate it to the clones of the recorder. So it can't be strictly immutable anymore, but the ID is only set once.

haphungw:
Closing this PR in favor of #5819. I have force-pushed the latest updates (including the rename to `destination_id` and the Discovery LRO macro fixes) directly to the branch on #5819 to preserve the existing approvals and review history. Apologies for the confusion and duplication."

PR #5873, opened 2026-06-11
impl(bigquery): add query errors

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery/src/error.rs
+++ b/src/bigquery/src/error.rs
@@ -1,0 +1,88 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Custom errors for the Cloud BigQuery query client.
+
+use google_cloud_bigquery_v2::model::ErrorProto;
+use google_cloud_gax::error::Error;
+
+/// Errors that can occur during query configuration, execution, or polling.
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+pub enum QueryError {
+    /// The project ID was not provided or could not be determined.
+    #[error("no project ID was provided")]
+    MissingProjectId,
+
+    /// Only query jobs are supported by this client.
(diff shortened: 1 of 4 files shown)


alvarowolfx:
Split from https://github.com/googleapis/google-cloud-rust/pull/5870

src/bigquery/src/error.rs
```diff
@@ -0,0 +1,88 @@
...
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+pub enum QueryError {
+    /// The project ID was not provided or could not be determined.
+    #[error("no project ID was provided")]
+    MissingProjectId,
+
+    /// Only query jobs are supported by this client.
+    #[error("only query jobs are supported")]
+    UnsupportedJobType,
```
dbolduc, line 30: Q: is this for the `attach_job` path?

The `query` path can only be a query job, right?

src/bigquery/src/error.rs
```diff
@@ -0,0 +1,88 @@
...
+    #[test]
+    fn test_job_failed_display() {
+        let err = QueryError::JobFailed {
+            reason: "invalidQuery".to_string(),
+            message: "Syntax error: Unexpected end of input".to_string(),
+            errors: vec![
+                ErrorProto::new()
+                    .set_reason("invalidQuery")
+                    .set_message("Syntax error: Unexpected end of input"),
+            ],
```
dbolduc: Q: are we extracting one of the reason / messages from the `ErrorProto`s?

Just curious why this stuff is duplicated

src/bigquery/src/error.rs
```diff
@@ -0,0 +1,88 @@
...
+        assert_eq!(
+            err.to_string(),
+            "query job failed: invalidQuery - Syntax error: Unexpected end of input"
+        );
+    }
+
+    #[test]
+    fn test_rpc_display() {
+        let status = google_cloud_gax::error::rpc::Status::default()
+            .set_code(google_cloud_gax::error::rpc::Code::InvalidArgument)
```
dbolduc: unimportant style nit: I like to `use google_cloud_gax::error::rpc::{Code, Status};` even though it is gonna add a line of code.

dbolduc (approved):

src/bigquery/src/error.rs
alvarowolfx: Yes, `attach_job` is only path now that would trigger that. But previous I was using InsertJobRequest and customers could pass a non-query job.

src/bigquery/src/error.rs
alvarowolfx: good point, I was making it easier to just see the primary reason on the JobFailed debug msg (`#[error("query job failed: {reason} - {message}")]`. Not sure what is the best way to debug a vec in the macro.

But customers can just do `.first()` themselves, so might be better to just have the `errors` vec.

PR #5885, opened 2026-06-15
test(o11y): add GAPIC LRO tracing success integration test

haphungw:
Initialize LRO tracing integration test suite and add the first test case verifying standard GAPIC LRO success paths.

### DIFF
--- a/tests/o11y/src/lro_tracing.rs
+++ b/tests/o11y/src/lro_tracing.rs
@@ -50,51 +50,26 @@ pub async fn lro_tracing_testlayer() -> anyhow::Result<()> {
     Ok(())
 }
 
-/// Helper to assert attributes on the T2 "LRO Wait" span.
-fn assert_lro_wait_span(
-    span: &CapturedSpan,
-    expected_method: &str,
-    expected_destination_id: Option<&str>,
-    expected_status: Option<&str>,
-    expected_description_contains: Option<&str>,
-) {
-    assert_eq!(
-        span.attributes.get("otel.name").and_then(|v| v.as_string()),
-        Some(expected_method.to_string())
-    );
-    assert_eq!(
-        span.attributes
-            .get("gcp.rpc.method")
-            .and_then(|v| v.as_string()),
-        Some(expected_method.to_string())
-    );
-    if let Some(dest_id) = expected_destination_id {
-        assert_eq!(
-            span.attributes
-                .get("gcp.resource.destination.id")
-                .and_then(|v| v.as_string()),
-            Some(dest_id.to_string())
-        );
-    }
-    assert_eq!(
-        span.attributes
-            .get("otel.status_code")
-            .and_then(|v| v.as_string()),
-        expected_status.map(|s| s.to_string())
-    );
-    if let Some(desc) = expected_description_contains {
-        let actual_desc = span
-            .attributes
-            .get("otel.status_description")
-            .and_then(|v| v.as_string())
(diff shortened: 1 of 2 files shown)


tests/o11y/src/lro_tracing.rs
```diff
@@ -0,0 +1,265 @@
...
+
+    // Run the successful LRO polling path test.
+    test_lro_success(&guard, &server, client.clone()).await?;
+    server.verify_and_clear();
+
+    Ok(())
+}
+
+/// Helper to assert attributes on the T2 "LRO Wait" span.
+fn assert_lro_wait_span(
```
dbolduc: nit: using a helper function to assert will hide the line numbers of failures.

You could add `#[track_caller]`, to know which line number called into this. But then you probably want to add a description to each of the assertions so you know which one failed.

https://rustc-dev-guide.rust-lang.org/backend/implicit-caller-location.html

tests/o11y/src/lro_tracing.rs
```diff
@@ -0,0 +1,265 @@
...
+        Some(&AttributeValue::UInt64(expected_poll_attempt))
+    );
+    if let Some(dest_id) = expected_destination_id {
+        assert_eq!(
+            span.attributes
+                .get("gcp.resource.destination.id")
+                .and_then(|v| v.as_string()),
+            Some(dest_id.to_string())
+        );
+    }
```
dbolduc: why not like:

```rs
        assert_eq!(
            span.attributes
                .get("gcp.resource.destination.id")
                .and_then(|v| v.as_string()).as_deref(),
            expected_destination_id);
```

tests/o11y/src/lro_tracing.rs
dbolduc: I don't think this thing is very helpful. The optional parameters are a code smell that we are trying too hard to unify things.

You could have like a:

```rs
fn get_string(span: &CapturedSpan, attribute: &'static str) -> Option<String> {
  span.attributes.get(attribute).and_then(|v| v.as_string())
}
```

and then just write these in line:

```rs
 assert_eq!(
        get_string(span, "otel.name").as_deref(),
        Some("google_cloud_showcase_v1beta1::client::Echo::wait::until_done")
    );
```

tests/o11y/src/lro_tracing.rs
```diff
@@ -0,0 +1,265 @@
...
+    // GetOperation attempt 1
+    assert_get_operation_span(
+        get_op_spans[0],
+        1,
+        Some("operations/wait-1"),
+        Some(false),
+        None,
+        "UNSET",
+        None,
+    );
```
dbolduc, line 173: it's hard for someone looking at the code to know what any of these parameters mean.

tests/o11y/src/lro_tracing.rs
haphungw: thanks for the suggestion, i did not consider this. changed the assertions to in-line.

tests/o11y/src/lro_tracing.rs
haphungw: done. i added an `attribute_str` helper to make it cleaner.

tests/o11y/src/lro_tracing.rs
haphungw: fixed as now the assertions are written in-line.

dbolduc (approved):

PR #5887, opened 2026-06-15
impl(bigquery): add jobs.insert query execution path

alvarowolfx:
Towards #5844

PR #5888, opened 2026-06-15
impl(bigquery): add jobs.query query execution path

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery/src/query/execution.rs
+++ b/src/bigquery/src/query/execution.rs
@@ -1,0 +1,145 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::error::QueryError;
+use crate::query::{Query, Result};
+use google_cloud_bigquery_v2::client::JobService;
+use google_cloud_bigquery_v2::model::{InsertJobRequest, Job, PostQueryRequest};
+use std::sync::Arc;
+
+pub(crate) struct PostQueryExecutor {
+    pub(crate) job_service: Arc<JobService>,
+    pub(crate) request: PostQueryRequest,
+}
+
+impl PostQueryExecutor {
+    pub(crate) fn new(job_service: Arc<JobService>, request: PostQueryRequest) -> Self {
+        Self {
+            job_service,
+            request,
+        }
+    }
+
+    pub(crate) async fn execute(self) -> Result<Query> {
+        let res = self
+            .job_service
+            .query()
+            .with_request(self.request)
+            .send()
+            .await
+            .map_err(|e| QueryError::Rpc { source: e })?;
+
+        if !res.errors.is_empty() {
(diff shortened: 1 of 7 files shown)


src/bigquery/src/query/run_query.rs
```diff
@@ -76,12 +77,13 @@ impl RunQuery {
...
+            let req = PostQueryRequest::new()
                 .set_project_id(project_id)
                 .set_query_request(query_request);
 
-            // TODO(#5844): implement jobs.query query execution
-            unimplemented!("jobs.query query execution not yet implemented");
+            PostQueryExecutor::new(self.job_service.clone(), req)
+                .execute()
+                .await
         }
```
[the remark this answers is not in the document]
alvarowolfx: will improve once we have both paths.

src/bigquery/src/query/run_query.rs
```diff
@@ -93,21 +95,16 @@ include!("../generated/run_query_request.rs");
...
-    #[derive(Debug)]
-    struct MockJobService;
-    impl google_cloud_bigquery_v2::stub::JobService for MockJobService {}
-
-    fn create_job_service() -> Arc<JobService> {
-        Arc::new(JobService::from_stub::<MockJobService>(Arc::new(
-            MockJobService,
-        )))
-    }
+    type TestResult = anyhow::Result<()>;
```
dbolduc, line 103: optional super nit: I just `use anyhow::Result;` or write out `anyhow::Result<()>` these days.

src/bigquery/src/query/execution.rs
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    pub(crate) async fn execute(self) -> Result<Query> {
+        let res = self
+            .job_service
+            .query()
+            .with_request(self.request)
+            .send()
+            .await
+            .map_err(|e| QueryError::Rpc { source: e })?;
```
dbolduc: nit: https://github.com/googleapis/google-cloud-rust/pull/5873#discussion_r3397457562

I think this is what the bot was talking about. It could just be `.await?`

src/bigquery/src/query/execution.rs
```diff
@@ -0,0 +1,145 @@
...
+
+        if !res.errors.is_empty() {
+            return Err(QueryError::JobFailed { errors: res.errors });
+        }
+
+        let completed = res.job_complete.unwrap_or(false);
+        let job_ref = res.job_reference.clone();
```
dbolduc: Q: why are we holding this separately when we could look up `self.initial_response.as_ref().and_then(|r| r.job_reference.as_ref())` ?

Are we going to move out `self.initial_response` at some point?

dbolduc: Or it is that other variations will not have an `initial_response`, and that is why the field is optional.

If so, this all makes sense. Just let me know.

dbolduc (approved):

src/bigquery/src/query/execution.rs
alvarowolfx: the other execution branch will have a `initial_job` `Job` instead of a `QueryResponse`, but both can generate an optional JobReference, that's why I extract it here.

See the other branch here: https://github.com/googleapis/google-cloud-rust/pull/5887/changes#diff-e1e7ca001a0917feb3cc17c75c06543d330a5c8b9dfa99dd89bc309225bcaeb0R24

src/bigquery/src/query/execution.rs
alvarowolfx: will try adding the `#[from]` to see if it magically converts. I think I will still need `.map_err(QueryError::from)`

alvarowolfx: TIL: the `?` also calls the `From` method, so yeah, you're correct, just `.await?` did the trick

PR #5886, opened 2026-06-15
test(o11y): add GAPIC LRO tracing logical error integration test

haphungw:
Stacked on top of https://github.com/googleapis/google-cloud-rust/pull/5885

Adds an integration test validating the tracing behavior of GAPIC LROs that complete with a logical error (i.e. status code 3 / INVALID_ARGUMENT inside the operation body).

PR #5895, opened 2026-06-16
impl(o11y): record `error.type` attribute on LRO span

haphungw:
Add `error.type` attribute to the LRO (T2) span and populate it from the status code if the long-running operation fails with a status code. This allows observability tools to track error rates by error type for long-running operations.

This lack of T2 error type recording is surfaced in the `error.type` assertions requested for LRO tracing integration tests (https://github.com/googleapis/google-cloud-rust/pull/5886).

### DIFF
--- a/src/lro/Cargo.toml
+++ b/src/lro/Cargo.toml
@@ -52,6 +52,7 @@ google-cloud-wkt         = { workspace = true }
 
 [target.'cfg(google_cloud_unstable_tracing)'.dependencies]
 tracing = { workspace = true }
+gaxi    = { workspace = true, features = ["_internal-common"] }
 
 [target.'cfg(google_cloud_unstable_tracing)'.dev-dependencies]
 tracing-subscriber      = { workspace = true, features = ["registry"] }
--- a/src/lro/src/internal/tracing.rs
+++ b/src/lro/src/internal/tracing.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use crate::{Poller, PollingResult, Result, sealed};
+use gaxi::observability::errors::error_type;
 use google_cloud_gax::polling_state::PollingState;
 use tracing::{Instrument, Span, info_span};
 
@@ -122,6 +123,7 @@ impl LroRecorder {
     pub fn record_error(&self, err: &crate::Error) {
         self.span.record("otel.status_code", "ERROR");
         self.span.record("otel.status_description", err.to_string());
+        self.span.record("error.type", error_type(err));
     }
 
     pub async fn record_action<F, Fut, T>(&self, f: F) -> T
--- a/src/lro/src/internal/ext.rs
+++ b/src/lro/src/internal/ext.rs
@@ -53,7 +53,8 @@ where
                 "gcp.rpc.method" = method_name,
                 "gcp.resource.destination.id" = tracing::field::Empty,
                 "otel.status_code" = tracing::field::Empty,
-                "otel.status_description" = tracing::field::Empty
+                "otel.status_description" = tracing::field::Empty,
+                "error.type" = tracing::field::Empty
             );
             let traced = Tracing::new(self, span);
             return Either::Right(traced);


src/lro/src/internal/tracing.rs
```diff
@@ -122,6 +122,11 @@ impl LroRecorder {
     pub fn record_error(&self, err: &crate::Error) {
         self.span.record("otel.status_code", "ERROR");
         self.span.record("otel.status_description", err.to_string());
+        if let Some(status) = err.status() {
```
westarle: Can we use https://github.com/googleapis/google-cloud-rust/blob/d059878adc986eb7e41c8e4e87e7319d722a3315/src/gax-internal/src/observability/errors.rs#L22 without too much modification?

src/lro/src/internal/tracing.rs
```diff
@@ -426,9 +431,8 @@ mod tests {
             .unwrap();
 
         assert!(
-            got.attributes
-                .get("gcp.longrunning.poll_attempt_count")
-                .is_none()
+            !got.attributes
```
westarle: please print the key (or attributes) if you find a key.

haphungw: since the clippy cleanups are out of scope for this PR (and will be addressed when we remove the unstable tracing gating flag altogether), I reverted those changes and we are back to using the original get().is_none() assertions for now.

src/lro/src/internal/tracing.rs
haphungw: i'd have to modify `gaxi` to expose `ErrorType`

src/lro/src/internal/tracing.rs
westarle: I don't see them reverted here...

src/lro/src/internal/tracing.rs
```diff
@@ -122,6 +122,10 @@ impl LroRecorder {
     pub fn record_error(&self, err: &crate::Error) {
         self.span.record("otel.status_code", "ERROR");
         self.span.record("otel.status_description", err.to_string());
+        self.span.record(
+            "error.type",
+            gaxi::observability::errors::ErrorType::from_gax_error(err).as_str(),
```
westarle: use gaxi::observability::errors::ErrorType above.

westarle (approved):
I think it looks good, but I'd like @dbolduc or someone familiar with Rust to comment on any downside to exposing the errors module pub in gaxi and adding the dependency here.

src/lro/src/internal/tracing.rs
haphungw: ack

src/lro/Cargo.toml
```diff
@@ -52,6 +52,7 @@ google-cloud-wkt         = { workspace = true }
 
 [target.'cfg(google_cloud_unstable_tracing)'.dependencies]
 tracing = { workspace = true }
+gaxi    = { workspace = true, features = ["_internal-grpc-client"] }
```
dbolduc: This is going to include gRPC components in the build for all of our GAPICs that have LROs.

You want: `"_internal-common"`.

src/gax-internal/src/observability.rs
```diff
@@ -22,7 +22,7 @@ pub(crate) mod propagation;
 pub(crate) mod attributes;
 
 #[cfg(any(feature = "_internal-http-client", feature = "_internal-grpc-client"))]
```
dbolduc: Same here. Use the `_internal-common` feature.

src/gax-internal/src/observability/errors.rs
```diff
@@ -42,7 +42,7 @@ pub enum ErrorType {
 }
 
 impl ErrorType {
```
dbolduc, line 45: before we forget, you should make this type `#[non_exhaustive]`

src/lro/src/internal/tracing.rs
```diff
@@ -122,6 +123,8 @@ impl LroRecorder {
     pub fn record_error(&self, err: &crate::Error) {
         self.span.record("otel.status_code", "ERROR");
         self.span.record("otel.status_description", err.to_string());
+        self.span
+            .record("error.type", ErrorType::from_gax_error(err).as_str());
```
dbolduc: exposing `ErrorType` seems unnecessary, at least at the moment. How do you expect this to grow/change over time?

If it isn't going to grow, then we can reduce the API surface, by just offering a function e.g.:

```rs
mod observability {
  pub fn to_string(e: &gax::Error) -> String {
    ErrorType::from_gax_error(err).as_str()
  }
}
```

src/gax-internal/src/observability/errors.rs
```diff
@@ -78,7 +78,7 @@ impl ErrorType {
         }
     }
 
-    pub(crate) fn as_str(&self) -> String {
+    pub fn as_str(&self) -> String {
```
dbolduc: This method name is wrong. It should be called `to_string()`.

"as" means you are yielding a cheap reference. "to" means you are making a new thing, leaving the original thing unchanged. https://rust-lang.github.io/api-guidelines/naming.html#c-conv

And `str` is a very different type from `String`.

src/gax-internal/src/observability/errors.rs
dbolduc: Or not expose it at all.

dbolduc (changes requested):
I am fine with exposing something from `gaxi`, but in a vacuum, I would minimize the surface.

Will any other layer need to know what is a strongly-typed `ErrorType`? Or do they just need to know how to encode it as an attribute?

haphungw:
@dbolduc I added this refactoring PR https://github.com/googleapis/google-cloud-rust/pull/5908 to fix those problems in `gaxi`.

dbolduc (approved):

PR #5896, opened 2026-06-16
impl(bigquery): add query polling and complete query

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery/src/query/query_handle.rs
+++ b/src/bigquery/src/query/query_handle.rs
@@ -18,6 +18,7 @@ use google_cloud_bigquery_v2::client::JobService;
 use google_cloud_bigquery_v2::model::{
     GetQueryResultsRequest, GetQueryResultsResponse, Job, JobReference, QueryResponse,
 };
+use google_cloud_gax::backoff_policy::BackoffPolicy;
 use google_cloud_gax::exponential_backoff::ExponentialBackoffBuilder;
 use google_cloud_gax::polling_backoff_policy::PollingBackoffPolicy;
 use google_cloud_gax::polling_state::PollingState;
@@ -42,24 +43,18 @@ impl Query {
             return Ok(CompleteQuery::from_query_response(self, initial_response));
         }
 
-        let res = poll_query_results(&self.job_service, self.job_ref.as_ref()).await?;
+        let job_ref = self
+            .job_ref
+            .as_ref()
+            .expect("query job should have job reference at this point");
+        let backoff_policy = Arc::new(
+            ExponentialBackoffBuilder::default()
+                .with_initial_delay(std::time::Duration::from_secs(10))
+                .build()
+                .expect("valid backoff configuration"),
+        );
+        let res = poll_query_results(&self.job_service, job_ref, backoff_policy).await?;
         Ok(CompleteQuery::from_get_query_results_response(self, res))
-    }
-
-    /// Returns the stateless query ID of the query, if available.
-    pub fn query_id(&self) -> Option<&str> {
-        self.initial_response.as_ref().and_then(|res| {
-            if res.query_id.is_empty() {
-                None
-            } else {
-                Some(res.query_id.as_str())
-            }
-        })
-    }
-
-    /// Returns the underlying job reference for this query.
-    pub fn job_reference(&self) -> Option<&JobReference> {
-        self.job_ref.as_ref()
     }
 }
 
@@ -94,18 +89,12 @@ impl CompleteQuery {
 /// Helper function to poll getQueryResults until a job finishes.
 pub(crate) async fn poll_query_results(
     job_service: &JobService,
-    job_ref: Option<&JobReference>,
+    job_ref: &JobReference,
+    backoff_policy: Arc<dyn PollingBackoffPolicy>,
 ) -> Result<GetQueryResultsResponse> {
-    let backoff_policy = ExponentialBackoffBuilder::default()
-        .with_maximum_delay(std::time::Duration::from_secs(10))
-        .build()
-        .unwrap();
-
     let mut state = PollingState::default();
 
     loop {
-        let job_ref = job_ref.expect("query job should have job reference at this point");
-
         let mut req = GetQueryResultsRequest::new()
             .set_max_results(0u32)
             .set_project_id(job_ref.project_id.clone())
@@ -118,10 +107,10 @@ pub(crate) async fn poll_query_results(
             .get_query_results()
             .with_request(req)
             .send()
-            .await
-            .map_err(|e| QueryError::Rpc { source: e })?;
+            .await?;
 
         if !res.errors.is_empty() {
+            // TODO(#5592): handle jobBackendError and other transient/retryable errors.
             return Err(QueryError::JobFailed { errors: res.errors });
         }
 
@@ -132,21 +121,25 @@ pub(crate) async fn poll_query_results(
 
         let delay = backoff_policy.wait_period(&state);
         tokio::time::sleep(delay).await;
+        // TODO(#5592): limit retry attempts or add cancellation mechanism
         state.attempt_count += 1;
     }
 }
 
 #[cfg(test)]
 mod tests {
+    use std::time::Duration;
+
     use super::*;
-    use crate::query::tests::{MockJobService, create_job_service};
+    use crate::query::tests::{
+        MockBackoffPolicy, MockJobService, create_job_service, create_test_backoff_policy,
+    };
     use google_cloud_bigquery_v2::model::{
         ErrorProto, GetQueryResultsResponse, JobReference, QueryResponse,
     };
     use google_cloud_gax::error::Error as GaxError;
     use google_cloud_gax::error::rpc::{Code, Status};
     use google_cloud_gax::response::Response;
-    use std::sync::atomic::{AtomicU32, Ordering};
 
     use test_case::test_case;
 
@@ -179,16 +172,18 @@ mod tests {
     #[tokio::test]
     async fn test_query_until_done_polls_success() -> TestResult {
         let mut mock = MockJobService::new();
-        mock.expect_get_query_results().returning(|req, _| {
-            assert_eq!(req.project_id, "some_project");
-            assert_eq!(req.job_id, "some_job_id");
-            assert_eq!(req.max_results, Some(0));
-            assert_eq!(req.location, "us-central1");
-            let res = GetQueryResultsResponse::new()
-                .set_job_complete(true)
-                .set_job_reference(JobReference::new().set_job_id(req.job_id));
-            Ok(Response::from(res))
-        });
+        mock.expect_get_query_results()
+            .returning(|req, _| {
+                assert_eq!(req.project_id, "some_project");
+                assert_eq!(req.job_id, "some_job_id");
+                assert_eq!(req.max_results, Some(0));
+                assert_eq!(req.location, "us-central1");
+                let res = GetQueryResultsResponse::new()
+                    .set_job_complete(true)
+                    .set_job_reference(JobReference::new().set_job_id(req.job_id));
+                Ok(Response::from(res))
+            })
+            .times(1);
         let job_service = create_job_service(mock);
         let job_ref = JobReference::new()
             .set_project_id("some_project")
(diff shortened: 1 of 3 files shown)


src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+            } else {
+                Some(res.query_id.as_str())
+            }
+        })
+    }
+
+    /// Returns the underlying job reference for this query.
+    pub fn job_reference(&self) -> Option<&JobReference> {
+        self.job_ref.as_ref()
+    }
```
dbolduc: (non-blocking, just remind me), Q: why do we prefer mutually exclusive optional accessors over an enum:

```rs
enum QueryReference {
  QueryId(String),
  JobReference(JobReference),
}

pub fn query_reference(&self) -> QueryReference; // Or maybe `Option<QueryReference>` if `None` is still possible.
```

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+
+/// Helper function to poll getQueryResults until a job finishes.
+pub(crate) async fn poll_query_results(
+    job_service: &JobService,
+    job_ref: Option<&JobReference>,
+) -> Result<GetQueryResultsResponse> {
+    let backoff_policy = ExponentialBackoffBuilder::default()
+        .with_maximum_delay(std::time::Duration::from_secs(10))
+        .build()
+        .unwrap();
```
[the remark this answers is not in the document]
dbolduc: We need to tell the bot that `expect` is allowed for things that cannot fail.

https://github.com/googleapis/google-cloud-rust/pull/5905

Still, I would prefer `expect("this is a valid configuration")` over `unwrap()`. Or:

```rs
// This is a valid configuration
.unwrap()
```

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+
+    let mut state = PollingState::default();
+
+    loop {
+        let job_ref = job_ref.expect("query job should have job reference at this point");
```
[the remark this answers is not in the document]
dbolduc: > query job should have job reference at this point

Then consider passing in a `&JobReference`. Ack that the `expect` has to happen at some point. I would probably do it immediately upon receiving the message from the server.

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+        if let Some(location) = job_ref.location.clone() {
+            req = req.set_location(location);
+        }
+
+        let res = job_service
+            .get_query_results()
+            .with_request(req)
+            .send()
+            .await
+            .map_err(|e| QueryError::Rpc { source: e })?;
```
dbolduc:
```suggestion
            .await?
```

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+        }
+
+        let completed = res.job_complete.unwrap_or(false);
+        if completed {
+            return Ok(res);
+        }
+
+        let delay = backoff_policy.wait_period(&state);
+        tokio::time::sleep(delay).await;
+        state.attempt_count += 1;
```
dbolduc, line 125: There should be a way to cut off this loop. Fine to take a TODO on it.

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+        }
+    }
+}
+
+/// Helper function to poll getQueryResults until a job finishes.
+pub(crate) async fn poll_query_results(
+    job_service: &JobService,
+    job_ref: Option<&JobReference>,
+) -> Result<GetQueryResultsResponse> {
+    let backoff_policy = ExponentialBackoffBuilder::default()
```
dbolduc: comment: It is fine to hardcode policies for now. Of note, there is no PollingPolicy, and this is not using our polling loop logic.

I think not using our LRO polling loop logic is probably the right call.

For the policy business, we will need to talk with @haphungw and come up with a plan. It would be nice if a query polling policy is familiar to a job polling policy, or an LRO polling policy.

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+        let res = job_service
+            .get_query_results()
+            .with_request(req)
+            .send()
+            .await
+            .map_err(|e| QueryError::Rpc { source: e })?;
+
+        if !res.errors.is_empty() {
+            return Err(QueryError::JobFailed { errors: res.errors });
+        }
```
dbolduc, line 115: do we need a TODO on classifying some of these errors as transient, and continuing / restarting the loop?

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+        let completed = query.until_done().await?;
+        assert_eq!(completed.job_ref.unwrap().job_id, "some_job_id");
+
+        Ok(())
+    }
+
+    #[test_case(Some("query_123"), Some("query_123"); "with query id")]
+    #[test_case(Some(""), None; "empty query id")]
+    #[test_case(None, None; "no query response")]
+
```
dbolduc: nit: remove newline

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+            completed: false,
+            initial_job: None,
+            initial_response,
+        };
+        assert_eq!(query.query_id(), expected_id);
+    }
+
+    #[test_case(true; "with job reference")]
+    #[test_case(false; "without job reference")]
+
```
dbolduc: same

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+    use super::*;
+    use crate::query::tests::{MockJobService, create_job_service};
+    use google_cloud_bigquery_v2::model::{
+        ErrorProto, GetQueryResultsResponse, JobReference, QueryResponse,
+    };
+    use google_cloud_gax::error::Error as GaxError;
+    use google_cloud_gax::error::rpc::{Code, Status};
+    use google_cloud_gax::response::Response;
+    use std::sync::atomic::{AtomicU32, Ordering};
+
```
dbolduc, line 143: same random newline

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+        let completed = query.until_done().await?;
+        assert_eq!(completed.job_ref.unwrap().job_id, "some_job_id");
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn test_query_until_done_polls_success() -> TestResult {
+        let mut mock = MockJobService::new();
+        mock.expect_get_query_results().returning(|req, _| {
```
dbolduc: nit: add `.times(1)` to make sure we do not keep polling after a success.

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+            initial_response: None,
+        };
+        assert_eq!(query.job_reference(), job_ref.as_ref());
+    }
+
+    #[tokio::test(start_paused = true)]
+    async fn test_poll_query_results_loops_until_complete() -> TestResult {
+        let mut mock = MockJobService::new();
+        let call_count = Arc::new(AtomicU32::new(0));
+        let call_count_clone = call_count.clone();
```
dbolduc: Use `mockall::Sequence`, e.g.:

https://github.com/googleapis/google-cloud-rust/blob/d9b1879e0fc3a06681e7ae2eb28fc7c44bf2f092/src/auth/src/retry.rs#L269-L287

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+            .set_project_id("some_project")
+            .set_job_id("some_job_id");
+
+        let start = tokio::time::Instant::now();
+        let res = poll_query_results(&job_service, Some(&job_ref)).await?;
+        let elapsed = start.elapsed();
+
+        assert!(res.job_complete.unwrap(), "{res:?}");
+        assert_eq!(call_count.load(Ordering::SeqCst), 3);
+        assert!(elapsed >= std::time::Duration::from_secs(1), "{elapsed:?}");
```
dbolduc: I am not looking closely at the policies, but are we sure this is always true? I thought our backoff policies were random.

dbolduc: What you really want is to pass in a backoff policy to the function signature. Then in the test you use a mock backoff policy and confirm it is called.

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+        };
+
+        let err = query.until_done().await.unwrap_err();
+        let errors = match err {
+            QueryError::JobFailed { errors } => errors,
+            _ => panic!("expected QueryError::JobFailed, got {err:?}"),
+        };
+        assert_eq!(errors.len(), 1);
+        assert_eq!(errors[0].reason, "invalidQuery");
+        assert_eq!(errors[0].message, "Syntax error");
```
dbolduc: style nit: I always prefer to match the container.

```rs
assert_eq!(errors, [ErrorProto::new()
                .set_reason("invalidQuery")
                .set_message("Syntax error")]);
```

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +33,322 @@ pub struct Query {
...
+    }
+
+    #[tokio::test]
+    async fn test_query_until_done_polls_success() -> TestResult {
+        let mut mock = MockJobService::new();
+        mock.expect_get_query_results().returning(|req, _| {
+            assert_eq!(req.project_id, "some_project");
+            assert_eq!(req.job_id, "some_job_id");
+            assert_eq!(req.max_results, Some(0));
+            assert_eq!(req.location, "us-central1");
```
dbolduc: opinion: I might have one test that verifies the fields in the initial request (including the optional location) and then never verify them again in other tests. IMO it can be kind of bloat-y to verify the same thing again and again and detracts from the focus of the tests. I don't think Carlos agrees with my opinion, so take it or leave it.

dbolduc (changes requested):
The only blocker is not using `mockall::Sequence` in the test.

src/bigquery/src/query/query_handle.rs
alvarowolfx: that's a good point, I thought about doing that by introducing this enum https://github.com/googleapis/google-cloud-rust/blob/main/src/bigquery/src/query/job_reference.rs previously, but I was almost giving up using it and creating an concept on top of the service. 

Maybe I use that and rename it to QueryReference

src/bigquery/src/query/query_handle.rs
alvarowolfx: good call. Yeah, depending on the error we need to reissue the job like in the `jobBackendError` 

https://docs.cloud.google.com/bigquery/docs/error-messages

src/bigquery/src/query/query_handle.rs
alvarowolfx: ack

src/bigquery/src/query/query_handle.rs
alvarowolfx: We've talked about this in the design doc around having a cancellation mechanism. Do you think we should have such mechanism or a hard timeout case here after trying enough ? I'm afraid of throwing a timeout when we know some queries can take actually take a long time to run.

src/bigquery/src/query/query_handle.rs
alvarowolfx: will do this in a separated PR. Will remove those methods for now

src/bigquery/src/query/query_handle.rs
dbolduc: To be clear, I am not necessarily talking about a cancellation mechanism. The application should have some way to say "give up after 10 minutes", or "give up after 10 polls" or something. (But that does not mean we need to go and send a `CancelJob` RPC. It might mean we exit the loop with the Job in an unknown state).

This all depends on the policies we expose. I am just saying that there is a "polling exhausted" branch of logic that is missing from the loop. It doesn't have to be added now, but eventually we will want it.

alvarowolfx: > "give up after 10 minutes", or "give up after 10 polls" 

what I'm calling cancellation mechanism is what you described here. I'm not talking about job cancellation.  Is to provide a way for customer to cancel the `until_done` call.

alvarowolfx (commented):
@dbolduc addressed your review comments and removed the JobReference/QueryId stuff to send. and discuss in another PR

src/bigquery/src/query/query_handle.rs
dbolduc: Ack

src/bigquery/src/query/query_handle.rs
```diff
@@ -25,3 +34,290 @@ pub struct Query {
...
+    /// Returns an error if a remote service or connection failure happens during polling.
+    pub async fn until_done(&self) -> Result<CompleteQuery> {
+        if let (true, Some(initial_response)) = (self.completed, &self.initial_response) {
+            return Ok(CompleteQuery::from_query_response(self, initial_response));
+        }
+
+        let job_ref = self
+            .job_ref
+            .as_ref()
+            .expect("query job should have job reference at this point");
```
dbolduc, line 49: If it is always there, why is the field an `Option<T>` ?

dbolduc (approved):

src/bigquery/src/query/query_handle.rs
alvarowolfx: stateless queries are not polled and don't have a job reference

PR #5905, opened 2026-06-17
ci(gemini): hone review bot

dbolduc:
(1) The bot is flagging too many valid `expect()`s because our language is too strict. Sometimes `expect()` is correct and preferred. So refine that standard.

(2) Tell the bot to look for unnecessary data copies, unnecessary locks. I swear this is like 50% of the "value" I add in a review.

### DIFF
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -50,7 +50,6 @@ When reviewing or generating code, apply rigorous scrutiny:
   preferable to returning an error that can't arise. In tests, prefer `?`. If
   that is not possible use `.expect()`. Only use `.unwrap()` as a last resort.
 
-
 ## Async & Concurrency
 
 - **Send & Sync:** New public `async fn` and Future types MUST be `Send` and
--- a/.gemini/styleguide.md
+++ b/.gemini/styleguide.md
@@ -45,11 +45,10 @@ When reviewing or generating code, apply rigorous scrutiny:
   meaningful distinctions for the user. Error variants should be actionable.
 - **Panics:** `unwrap()` and `expect()` should typically be avoided in
   production code and examples (use `?` or handle errors). If an invariant is
-  known, (e.g. a builer always returns a valid configuration, or the service
-  always returns a message with presence), it is fine to `expect()`, and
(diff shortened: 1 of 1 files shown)


.gemini/styleguide.md
```diff
@@ -26,16 +26,28 @@ When reviewing or generating code, apply rigorous scrutiny:
...
-  `?` operator for propagation. Custom error types should imply meaningful
-  distinctions for the user.
+  `?` operator for propagation. The library should almost always prefer to
+  return a `Result<>` over an internal `panic!`. Custom error types should imply
+  meaningful distinctions for the user. Error variants should be actionable.
+- **Panics:** `unwrap()` and `expect()` should typically be avoided in
+  production code and examples (use `?` or handle errors). If an invariant is
+  known, (e.g. a builer always returns a valid configuration, or the service
+  always returns a message with presence), it is fine to `expect()`, and
+  often preferable to returning an error that can't arise.
```
coryan: Do you want to say something like "In tests, prefer `?` if that is not possible using `.expect()` and only use `.unwrap()` as a last resort."

dbolduc: Yeah, good advice. Done.

alvarowolfx (approved):

.gemini/styleguide.md
```diff
@@ -26,16 +26,31 @@ When reviewing or generating code, apply rigorous scrutiny:
...
-  acceptable in tests.
 - **Error Handling:** Public functions should return `Result<T, Error>`. Use the
-  `?` operator for propagation. Custom error types should imply meaningful
-  distinctions for the user.
+  `?` operator for propagation. The library should almost always prefer to
+  return a `Result<>` over an internal `panic!`. Custom error types should imply
+  meaningful distinctions for the user. Error variants should be actionable.
+- **Panics:** `unwrap()` and `expect()` should typically be avoided in
+  production code and examples (use `?` or handle errors). If an invariant is
+  known, (e.g. a builer always returns a valid configuration, or the service
```
westarle:
```suggestion
  known, (e.g. a builder always returns a valid configuration, or the service
```

dbolduc: Fixed, thanks.

PR #5908, opened 2026-06-17
refactor(o11y): minimize internal API surface and gate errors under common feature

haphungw:
Refactor `gaxi` errors tracing observability module to minimize the public API surface and optimize feature gating for downstream usage.

This is a prerequisite for adding `error.type` span attributes recording to the LRO span (#5895), ensuring `google-cloud-lro` can fetch error tracing categories without bloating its dependency footprint.

### DIFF
--- a/src/gax-internal/src/observability.rs
+++ b/src/gax-internal/src/observability.rs
@@ -39,11 +39,8 @@ mod client_signals;
 #[cfg(any(feature = "_internal-http-client", feature = "_internal-grpc-client"))]
 pub use client_signals::{
     ClientRequestAttributes, DurationMetric, RequestRecorder, TransportMetric, WithClientLogging,
-    WithClientMetric, WithClientSpan, WithTransportMetric,
+    WithClientMetric, WithClientSpan, WithTransportLogging, WithTransportMetric, WithTransportSpan,
 };
-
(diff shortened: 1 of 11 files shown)


src/gax-internal/src/observability.rs
```diff
@@ -33,13 +33,16 @@ pub(crate) use http_tracing::{ResultExt as HttpResultExt, create_http_attempt_sp
...
 mod client_signals;
 
+#[cfg(any(feature = "_internal-http-client", feature = "_internal-grpc-client"))]
 pub use client_signals::{
     ClientRequestAttributes, DurationMetric, RequestRecorder, TransportMetric, WithClientLogging,
     WithClientMetric, WithClientSpan, WithTransportMetric,
 };
 
+#[cfg(any(feature = "_internal-http-client", feature = "_internal-grpc-client"))]
 pub use client_signals::{WithTransportLogging, WithTransportSpan};
```
dbolduc: why are these separate?

dbolduc (approved):

src/gax-internal/src/observability.rs
haphungw: done.
These were separate `pub use` blocks from the original T3/T4 telemetry implementation that haven't been cleaned up

PR #5912, opened 2026-06-17
fix(spanner): make types module public for typed param construction

### LINKED ISSUE #5911: [Spanner] `types` module is `pub(crate)` — blocks external use of `add_typed_param` with primitive types
## Summary

The `google_cloud_spanner::types` module is declared `pub(crate)` at `src/spanner/src/lib.rs:92`, which makes its contents — including the documented public-looking constructors `int64()`, `string()`, `timestamp()`, `bool()`, `bytes()`, `date()`, `float32()`, `float64()`, `json()`, `numeric()`, `uuid()`, `interval()`, and the PostgreSQL variants `pg_numeric()`, `pg_jsonb()`, `pg_oid()` — unreachable from external crates.

This blocks the documented `StatementBuilder::add_typed_param` API from being used with primitive types in downstream code. Currently the only public path requires going through `google_cloud_spanner::model::Type` (a `#[non_exhaustive]` generated protobuf struct) and field mutation, which is fragile and unergonomic.

## Environment

- Crate: `google-cloud-spanner`
- Pin: `rev d0d2f50f099248d4795e4f4efa32fbac67e28024` (or any rev that has the same `pub(crate)` declaration in `src/spanner/src/lib.rs:92`)
- Build target: any external Rust crate that depends on `google-cloud-spanner`

## What you observed

External code attempting to bind a typed parameter to a Spanner query — e.g. a change-stream TVF like:

```sql
SELECT ChangeRecord FROM READ_my_stream(
  start_timestamp => @start_timestamp,
  end_timestamp => @end_timestamp,
  partition_token => @partition_token,
  heartbeat_milliseconds => @heartbeat_milliseconds
)
```

cannot use the documented public-looking path:

```rust
use google_cloud_spanner::types;
// ...
.add_typed_param("start_timestamp", &start_ts, types::timestamp())
```

The Rust compiler rejects this with:

```
error[E0603]: module `types` is private
 --> src/connector/src/source/spanner_cdc/source/reader.rs:665:5
  |
6 | use google_cloud_spanner::types;
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^ private module
  |
  = note: the module `types` is defined here:
          .../google-cloud-spanner/src/lib.rs:92
  = help: the module is `pub(crate)`, so it can only be used within the same crate
```

The same error fires for `types::string()`, `types::int64()`, and the other primitive constructors.

## Root cause

`src/spanner/src/lib.rs:92`:

```rust
pub(crate) mod types;
```

The `pub(crate)` visibility makes the entire module invisible to anything outside the `google-cloud-spanner` crate, regardless of the `pub fn` markers on the individual items inside. This is a classic Rust visibility gotcha: the items are *declared* public, but the *module containing them* is crate-private.

The constructors themselves are correctly marked public and well-tested inside the crate. For example, `src/spanner/src/types.rs:207, 212, 242`:

```rust
pub fn int64() -> Type { ... }
pub fn string() -> Type { ... }
pub fn timestamp() -> Type { ... }
```

And `src/spanner/src/statement.rs:390-394` (the SDK's own tests) use them:

```rust
crate::types::string()
```

The visibility is contradictory: the SDK uses these constructors publicly internally and hides them externally. This looks like an oversight rather than a deliberate design choice.

## Reproduction

Minimal external-crate reproduction (any binding code is sufficient):

```rust
// External crate, e.g. a CDC source connector
use google_cloud_spanner::types;  // compile error: module `types` is private

fn bind_typed_param(stmt: google_cloud_spanner::statement::StatementBuilder) -> _ {
    stmt.add_typed_param("ts", &time::OffsetDateTime::now_utc(), types::timestamp())
}
```

The `types` import is the failing line. Removing it and using the workaround below compiles.

## Current workaround (what downstream projects are forced to do)

```rust
use google_cloud_spanner::model::{Type as ModelType, TypeCode};
use google_cloud_spanner::value::Type as ValueType;

let spanner_ts: ValueType = {
    let mut t = ModelType::new();           // Default::default()
    t.code = TypeCode::Timestamp;           // mutate the one field
    t.into()                                // From<model::Type> for value::Type
};
let spanner_str: ValueType = {
    let mut t = ModelType::new();
    t.code = TypeCode::String;
    t.into()
};
let spanner_i64: ValueType = {
    let mut t = ModelType::new();
    t.code = TypeCode::Int64;
    t.into()
};
// ... 12 lines for what should be 3
```

This workaround has been discovered independently by at least two downstream projects (one internal to RisingWave's Spanner CDC connector, and an earlier internal migration at commit `aa494b3cab` against a different SDK rev that hit the same issue). The pattern is identical because the constraint is identical.

### Why the workaround is fragile

`google_cloud_spanner::model::Type` is `#[non_exhaustive]`. We can't use struct-literal syntax from outside the crate, so we mutate fields after `new()`. If a future SDK rev adds a required field to the protobuf `Type` message, our code keeps compiling (because we never wrote a struct literal) but might silently produce a value Spanner rejects. Mitigation: the SDK's own internal code at `types.rs:288-297` uses the same `Default::default()` + field-mutation pattern, so upstream tests would break first.

## Suggested fix

Change `pub(crate) mod types;` to `pub mod types;` in `src/spanner/src/lib.rs:92`.

```diff
- pub(crate) mod types;
+ pub mod types;
```

That's it. The constructors are already `pub fn` and documented. After the change, downstream code can use the clean public path:

```rust
use google_cloud_spanner::types;
// ...
.add_typed_param("start_timestamp", &start_ts, types::timestamp())
.add_typed_param("partition_token", &token, types::string())
.add_typed_param("heartbeat_milliseconds", &n, types::int64())
```

The `model::Type` workaround becomes unnecessary, the `use google_cloud_spanner::model::...` imports drop, the helper `let` bindings drop, and the `#[non_exhaustive]` fragility goes away for the TVF binding use case.

## Risk assessment of the fix

- **API surface impact**: zero. The constructors are already public inside the crate; making the module public just makes them visible outside.
- **Behavioral change**: none. The functions return the same `Type` values either way.
- **Documentation**: the constructors are already documented at their declaration sites (`src/spanner/src/types.rs:200-280`). No doc updates needed.
- **Tests**: the SDK's own tests at `statement.rs:390-394` use the constructors. If the visibility change broke anything, those tests would fail.
- **Cargo features**: no feature flag change needed. The module was always compiled; only its visibility changes.

## Alternative fixes (if `pub mod types` is undesirable for some reason)

1. Re-export the constructors from `lib.rs` directly:
   ```rust
   pub use types::{int64, string, timestamp, /* ... */};
   ```
   This achieves the same effect without exposing the entire `types` module.
2. Add a builder method to `Type` itself: `Type::timestamp()`, `Type::string()`, etc. as associated functions. Same outcome, different location.
3. Re-export the constructors under a `pub mod` namespace, e.g. `pub mod typed { pub use crate::types::*; }`. Opt-in visibility.

The author's recommendation is option 1 (one-line re-export) because it preserves module encapsulation while exposing the public API.

## Why this matters

The `add_typed_param` API on `StatementBuilder` is part of the documented public surface. Its third argument is `param_type: Type`. External users have a right to expect they can construct that `Type` value without going through `model::Type` (a generated protobuf struct) and field mutation. The current forced workaround:

- Is verbose (12 lines for what should be 3).
- Couples external code to a generated protobuf type that the SDK explicitly marks as `#[non_exhaustive]`.
- Has been independently discovered and re-implemented by at least two downstream projects — a strong signal that the API surface is incomplete.

Filing this as a "missing public API" rather than a "bug" because no existing functionality is broken; only a public API is missing.

## Related observations

- The same SDK rev has `pub mod value;` and `pub mod model;` declared `pub` (not `pub(crate)`), so `value::Type` and `model::Type` are reachable. Only `types` is hidden.
- The `From<model::Type> for value::Type` impl that the workaround relies on is `pub`, so the conversion is intended to be usable from external code. Combined with the hidden `types` constructors, the design intent is unclear — the public paths are inconsistent.

## Test cases for the fix

After changing `pub(crate)` → `pub`, the following external-crate test should compile and pass:

```rust
#[test]
fn external_uses_typed_param_constructors() {
    use google_cloud_spanner::statement::Statement;
    use google_cloud_spanner::types;

    let stmt = Statement::builder("SELECT @ts, @s, @n")
        .add_typed_param("ts", &time::OffsetDateTime::now_utc(), types::timestamp())
        .add_typed_param("s", &"hello".to_string(), types::string())
        .add_typed_param("n", &42i64, types::int64())
        .build();

    assert!(stmt.sql().contains("SELECT @ts, @s, @n"));
}
```

Currently this fails to compile with "module `types` is private". After the fix, it should compile and pass.

buu-nguyen:
The `google_cloud_spanner::types` module was declared `pub(crate)`, so the
documented primitive constructors (`int64()`, `string()`, `timestamp()`, `bool()`, `bytes()`, `date()`, `float32()`, `float64()`, `json()`, `numeric()`, `uuid()`, `interval()`, plus the PostgreSQL variants `pg_numeric()`, `pg_jsonb()`, `pg_oid()`) were unreachable from external crates. Downstream code calling `StatementBuilder::add_typed_param` was forced to construct values via the generated
`google_cloud_spanner::model::Type`
protobuf struct and field mutation, coupling external code to a `#[non_exhaustive]` generated type.

The module visibility was the only thing wrong: every constructor is already
`pub fn` inside the module, already documented at its declaration site in
`src/spanner/src/types.rs`, and already used internally by the SDK's own tests at `src/spanner/src/statement.rs:390-394` via `crate::types::string()`.
This was a classic Rust visibility gotcha where items declared `pub` were
hidden by a `pub(crate)` module.

Changed `pub(crate) mod types` to `pub mod types` at `src/spanner/src/lib.rs:92`.
Added an integration test (`tests/types_module_visibility.rs`) that exercises
the constructors from an external-crate perspective, so any future regression
that re-hides the module fails compilation with the same `E0603` error downstream users reported.

Fixes #5911

### DIFF
--- a/src/spanner/src/lib.rs
+++ b/src/spanner/src/lib.rs
@@ -32,6 +32,8 @@ pub mod read;
 pub mod result;
 /// SQL statement builders and parameter bindings.
 pub mod statement;
+/// Spanner primitive type constructors for typed parameter binding.
+pub mod types;
 /// Type and value representations and conversion traits.
 pub mod value;
 
@@ -89,7 +91,6 @@ pub(crate) mod timestamp_bound;
 pub(crate) mod to_value;
 pub(crate) mod transaction_retry_policy;
 pub(crate) mod transaction_runner;
-pub(crate) mod types;
 pub(crate) mod write_only_transaction;
 
 mod status;
--- a/src/spanner/tests/types_module_visibility.rs
+++ b/src/spanner/tests/types_module_visibility.rs
@@ -1,0 +1,59 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Integration tests that exercise the public `google_cloud_spanner::types`
+//! module from an external-crate perspective.
+//!
+//! If the `types` module is ever declared `pub(crate)` again, these tests
+//! fail to compile with `error[E0603]: module 'types' is private`. That
+//! compile-time failure is the regression guard: it locks the public
+//! constructors (`types::timestamp()`, `types::string()`, `types::int64()`,
+//! etc.) reachable from external crates that depend on
+//! `google-cloud-spanner`.
+
+use google_cloud_spanner::statement::Statement;
+use google_cloud_spanner::types;
+
+#[test]
+fn external_uses_typed_param_constructors() {
+    let stmt = Statement::builder("SELECT @ts, @s, @n")
+        .add_typed_param("ts", &time::OffsetDateTime::now_utc(), types::timestamp())
+        .add_typed_param("s", &"hello".to_string(), types::string())
+        .add_typed_param("n", &42i64, types::int64())
+        .build();
+
+    assert!(stmt.sql().contains("SELECT @ts, @s, @n"));
+}
+
+#[test]
+fn external_reaches_all_primitive_constructors() {
+    // Touch each documented primitive constructor so any future visibility
+    // regression that hides even one of them fails compilation here.
+    let _ = types::int64();
+    let _ = types::string();
+    let _ = types::timestamp();
+    let _ = types::bool();
+    let _ = types::bytes();
+    let _ = types::date();
+    let _ = types::float32();
+    let _ = types::float64();
+    let _ = types::json();
+    let _ = types::numeric();
+    let _ = types::uuid();
+    let _ = types::interval();
+    let _ = types::pg_numeric();
+    let _ = types::pg_jsonb();
+    let _ = types::pg_oid();
+    let _ = types::array(types::int64());
+}


buu-nguyen:
> /gcbrun

@olavloite can you re-trigger this? Thanks.
I forgot to run `cargo fmt`, mb. Just did it and committed.

olavloite (approved):

sakthivelmanii (approved):

PR #5914, opened 2026-06-18
feat(storage): implement V4 Signed Policy Documents

### LINKED ISSUE #3079: storage_generate_signed_post_policy_v4

### LINKED ISSUE #4746: Feature parity for generateSignedPostPolicyV4
Bring the Rust SDK up to feature parity with the other SDKs for this feature.

Public docs are at https://docs.cloud.google.com/storage/docs/xml-api/post-object-forms (includes samples for other languages)

xlai20:
## Description
This PR introduces support for generating **V4 Signed Policy Documents** (`PostPolicyV4Builder`), bringing the Rust SDK into feature parity with other official GCS SDKs for POST form object uploads.

**Design Doc:** [go/rust-sdk-feature-v4-signed-policy-document-implementation_plan](http://go/rust-sdk-feature-v4-signed-policy-document-implementation_plan)

## Key Additions
* **`PostPolicyV4Builder` API:** A fluent builder API to configure URL styles, exact-match fields, `starts-with` prefix conditions, and `content-length-range` limits.
* **Strict JSON Serialization:** Implemented exact-byte character escaping using UTF-16 surrogate pairs (`\uXXXX`) to guarantee flawless Base64 JSON encoding, satisfying GCS's stringent signature validation requirements.

## Testing & Validation
* **Manual Validation:** I have manually run the added `generate_signed_post_policy_v4` sample code against a real Cloud Storage bucket and verified the upload worked correctly.
* **Conformance Tests Passed:** Integrated with the official `v4_signatures.json` test harness, successfully passing **100% (11/11)** of the `postPolicyV4Tests` (covering Path Style, Virtual Hosted Style, Bucket Bound Hostnames, and Unicode conditions).
* **Cross-Language Parity Check:** Verified that `$key` condition formulation and the maximum 7-day expiration bounds behave identically to the official Go and C++ SDKs.
* **Integration Tests:** Added `signed_post_policies_v4` in `tests/storage/src/lib.rs` which dynamically constructs a policy and performs a live upload using a `reqwest` multipart form. It ran successfully against a real Cloud bucket.


src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,502 @@
...
+}
+
+/// The result of signing a V4 POST Policy Document.
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct PostPolicyV4Result {
+    /// The destination URL for the POST request.
+    pub url: String,
+    /// The form fields (hidden inputs) that must be included in the multipart POST request.
+    pub fields: HashMap<String, String>,
+}
```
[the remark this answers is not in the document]
xlai20: Done.

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,502 @@
...
+
+    /// Adds a starts-with condition constraint (e.g. "$key", "uploads/").
+    pub fn with_starts_with<F: Into<String>, P: Into<String>>(
+        mut self,
+        field: F,
+        prefix: P,
+    ) -> Self {
+        self.starts_with_conditions.push((field.into(), prefix.into()));
+        self
+    }
```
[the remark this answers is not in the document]
xlai20: Done.

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,502 @@
...
+        let bucket_name = self.bucket_name();
+
+        let resolved_host = match self.endpoint.as_ref() {
+            Some(e) => {
+                let url = url::Url::parse(e)
+                    .map_err(|err| SigningError::invalid_parameter("endpoint", err))?;
+                url.host_str()
+                    .ok_or_else(|| SigningError::invalid_parameter("endpoint", "Missing host"))?
+                    .to_string()
+            }
```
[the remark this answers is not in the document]
xlai20: Done

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,502 @@
...
+                format!("{}://{}/", self.scheme, clean_hostname)
+            }
+        };
+
+        Ok(url)
+    }
+
+    /// Sign the policy document.
+    pub async fn sign_with(self, signer: &Signer) -> Result<PostPolicyV4Result, SigningError> {
+        let now = self.timestamp.unwrap_or_else(Utc::now);
```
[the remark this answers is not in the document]
xlai20: Done.

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,502 @@
...
+        let request_timestamp = now.format("%Y%m%dT%H%M%SZ").to_string();
+        let datestamp = now.format("%Y%m%d");
+        let credential_scope = format!("{datestamp}/auto/storage/goog4_request");
+
+        let client_email = if let Some(email) = self.client_email.clone() {
+            email
+        } else {
+            signer.client_email().await.map_err(SigningError::signing)?
+        };
```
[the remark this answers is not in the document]
xlai20: Done.

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,502 @@
...
+        fields.insert("x-goog-credential".to_string(), credential);
+        fields.insert("x-goog-date".to_string(), request_timestamp);
+        fields.insert("x-goog-signature".to_string(), signature_hex);
+        fields.insert("policy".to_string(), encoded_policy);
+
+        // Add user-supplied fields (including custom metadata or x-ignore- fields)
+        for (key, value) in &self.fields {
+            fields.insert(key.clone(), value.clone());
+        }
+
```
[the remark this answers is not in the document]
xlai20: Done.

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,582 @@
...
+    }
+
+    /// Sets the URL formatting style.
+    pub fn with_url_style(mut self, url_style: UrlStyle) -> Self {
+        self.url_style = url_style;
+        self
+    }
+
+    /// Sets the URL protocol scheme (e.g. "http" or "https").
+    pub fn with_scheme<S: Into<String>>(mut self, scheme: S) -> Self {
```
joshuatants: `SignedUrlBuilder` only has `with_endpoint` and parses the scheme from there: https://github.com/googleapis/google-cloud-rust/blob/03f5ffa4c12bd1a255a5e1a5ed34571525abc950/src/storage/src/storage/signed_url.rs#L486 . Is there a reason not to follow that pattern?

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,582 @@
...
+                }
+                host
+            }
+            None => {
+                let universe_domain = self.universe_domain.as_deref().unwrap_or("googleapis.com");
+                format!("storage.{universe_domain}")
+            }
+        };
+
+        let url = match self.url_style {
```
joshuatants, line 329: what happens if set_scheme is not called by the user, and instead eth user calls `set_endpoint(https://test.com)`?

As pointed out in an earliec comment this deviates from the behavioru in SignedUrlBuilder

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,582 @@
...
+        );
+        fields.insert("x-goog-credential".to_string(), credential);
+        fields.insert("x-goog-date".to_string(), request_timestamp);
+        fields.insert("x-goog-signature".to_string(), signature_hex);
+        fields.insert("policy".to_string(), encoded_policy);
+
+        // Add user-supplied fields (including custom metadata or x-ignore- fields)
+        for (key, value) in &self.fields {
+            fields.insert(key.clone(), value.clone());
+        }
```
joshuatants: What happens if we have 

`PostPolicyV4Builder::for_object(bucket, "object")
  .with_field("key", "another_object");
`

Looking at the code right now, I think that what would happen is that the signing happens with the original "object" (since user supplied fields with reserved keywords are ignored when building `conditions`), but lines 310-312 end up mapping `key` to `another_object`, meaning that the signature no longer matches the contents of the form. GCS will probably reject this?

Compare with `SignedUrlBuilder` (https://github.com/googleapis/google-cloud-rust/blob/03f5ffa4c12bd1a255a5e1a5ed34571525abc950/src/storage/src/storage/signed_url.rs#L534): there, the user supplied fields are added first, then the reserved parameters. This has the effect that the user cannot override the reserved fields, even if they try to do so.

Perhaps the safest way is to, in both `SignedUrlBuilder` and here, maintain a list of reserved keywords and reject with an error if the user tries to use one. We can leave that decision till later since we'll have to update `SignedUrlBuilder`.

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,582 @@
...
+    pub fn with_content_length_range(mut self, min: u64, max: u64) -> Self {
+        self.content_length_range = Some((min, max));
+        self
+    }
+
+    fn bucket_name(&self) -> &str {
+        self.bucket
+            .strip_prefix("projects/_/buckets/")
+            .unwrap_or(&self.bucket)
+    }
```
joshuatants, line 279: `SignedUrlBuilder` expects the prefix (https://github.com/googleapis/google-cloud-rust/blob/03f5ffa4c12bd1a255a5e1a5ed34571525abc950/src/storage/src/storage/signed_url.rs#L147), while here we accept both formats. Let's require the prefix for API consistency.

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,582 @@
...
+        scheme: String,
+        bucket: String,
+        object: String,
+        expiration: u64,
+        timestamp: String,
+        url_style: Option<String>,
+        bucket_bound_hostname: Option<String>,
+        fields: Option<HashMap<String, String>>,
+        conditions: Option<PostPolicyV4TestConditions>,
+    }
```
joshuatants, line 539: THe builder has `with_endpoint` and `with_universe_domain`, so they should be added here and tested as well.

joshuatants (changes requested):

src/storage/src/storage/post_policy.rs
xlai20: Done. 

I have thoroughly check all the public functions in post_policy.rs against those in signed_url.rs for API consistency within Rust sdk, and added the same bucket prefix check here.

src/storage/src/storage/post_policy.rs
xlai20: Done.

I removed set_scheme() because this function is conflicting with the set_endpoint().

I had also thoroughly checked all other pub functions to make sure each function exists for a meaningful purpose; previously the AI is generating some functions to fit the tests, which had been removed in latest commit now.

src/storage/src/storage/post_policy.rs
xlai20: Done. 

Actually `with_endpoint` and `with_universe_domain` had been tested in the local edge_case tests already. They were not added here because the v4_conformance.json imported from GCS does not have a test that cover these two; but nevertheless, I add these into the conformance tests just in case in the future the conformance tests evolve to include more edge cases.

src/storage/src/storage/post_policy.rs
xlai20: Done.

I had made some changes to this `sign_with` method to address your concerns as follows:
1. In the output fields, I change the order in which reserved fields and user supplied fields are being added. Hence, now, the post_policy `sign_with` method is handling the output fields like what signed_url is handling the query paramters, that user cannot override the reserved fields. This also ensures that the signature will match the contents of the output form.
2. Yes, I am maintaining a list of reserved keywords and does not add to the input JSON document at all (there is a strict order requirement in the JSON input document before signing, so not adding them in the beginning is the best solution). Based on my research with other SDK implementation and a cross-check with the Rust SDK community, a silent ignoring here is good enough; we don't need to explicitly reject user's entire builder.

src/storage/src/storage/post_policy.rs
xlai20: Ditto my other comment.

`with_scheme` is removed. I have followed the way `signed-Url` is resolving end point in the latest commit.

src/storage/src/storage/post_policy.rs
joshuatants: LGTM. Let's also look at refactoring similar logic across all the SignedUrlBuilder functions. Can be done as a follow-up.

xlai20:
@joshuatants I've made 9 more commits since your review, to address your concerns and also further improved the code. These are the changes after your review:
*   **API Simplifications**:
    *   Removed `with_scheme()` and `with_bucket_bound_hostname()` methods from `PostPolicyV4Builder`.
    *   Unified endpoint processing logic in a new `resolve_endpoint()` helper that standardizes the `http/https` prefix before resolving the host.
*   **Bucket Name Validation**: Added `check_bucket_name()` to strictly enforce that the provided bucket name must begin with the `"projects/_/buckets/"` prefix.
*   **System Keys Protection**: Extended the protected `system_keys` check to include `"x-goog-signature"` and `"policy"`, guaranteeing that user-supplied malicious or conflicting fields are ignored in favor of the system-generated defaults.
*   **Documentation & Doc Tests**: Added comprehensive Rustdoc examples (`/// # Example`) and doc tests covering how to use `PostPolicyV4Builder` and each of its builder configuration methods.
*   **Test Updates**:
    *   Updated existing tests to accommodate the `"projects/_/buckets/"` prefix requirement and the unified endpoint behavior.
    *   Added a new `post_policy_v4_custom_fields` edge-case test to verify that custom system fields correctly drop/override conflicting user-supplied configurations while preserving standard custom fields (e.g., `x-goog-meta-*`).
*   **Example Script Updates**: Modified the example `generate_signed_post_policy_v4.rs` to format the bucket string properly with `"projects/_/buckets/"` and included an extra `"x-goog-meta-test"` metadata field demo.

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,854 @@
...
+use base64::prelude::BASE64_STANDARD;
+use chrono::{DateTime, Utc};
+use google_cloud_auth::signer::Signer;
+use std::collections::BTreeMap;
+use std::time::Duration;
+
+/// Creates [V4 Signed Policy Documents] (POST Object Forms).
+///
+/// This builder allows you to generate signed V4 POST policy documents for Google Cloud Storage.
+/// A [Signed Policy Document] enables unauthenticated users to upload files to GCS using an HTML form
```
joshuatants, line 26: I don't see a definition for `Signed Policy Document`. Does this link work?

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,854 @@
...
+
+        let scheme = url.scheme();
+        let mut host = url
+            .host_str()
+            .ok_or_else(|| SigningError::invalid_parameter("endpoint", "Missing host"))?
+            .to_string();
+
+        if let Some(port) = url.port() {
+            host.push_str(&format!(":{port}"));
+        }
```
joshuatants: Perhaps confusingly, if an explicitly specified port is the default port for a given scheme,`url::port` will not return it (https://docs.rs/url/2.5.8/url/struct.Url.html#method.port), i.e.

`https://host:443 -> host`
`http://host:80   -> host`

Based on https://github.com/googleapis/google-cloud-rust/blob/bbfc16322d84d7ede77b9ae227bd4806d0634185/src/storage/src/storage/signed_url.rs#L472-L477 , we need to keep the default port if it's specified. Let's reuse the logic from SignedUrlBuilder, perhaps refactoring if it makes sense.

src/storage/examples/src/objects/generate_signed_post_policy_v4.rs
```diff
@@ -0,0 +1,52 @@
...
+
+pub async fn sample(bucket_name: &str, object_name: &str) -> anyhow::Result<()> {
+    let signer = google_cloud_auth::credentials::Builder::default().build_signer()?;
+
+    let policy =
+        PostPolicyV4Builder::for_object(format!("projects/_/buckets/{bucket_name}"), object_name)
+            .with_expiration(Duration::from_secs(30 * 60)) // 30 minutes
+            .with_field("Content-Type", "text/plain")
+            .with_field("x-goog-meta-test", "data")
+            .with_starts_with("$key", "")
```
joshuatants, line 27: "" is the empty string, so`with_starts_with("$key", "")` is trivially true.

joshuatants (approved):

src/storage/examples/src/objects/generate_signed_post_policy_v4.rs
xlai20: I purposely wrote this. This is an example taken out from Google Cloud documentation (https://docs.cloud.google.com/storage/docs/authentication/signatures#policy-document-example), it says this means the file name can start with any valid character.

src/storage/src/storage/post_policy.rs
xlai20: Edited the link to "Signed Policy Document" name. Verified using rust_analyzer that this doc comment will work properly with valid links.

src/storage/src/storage/post_policy.rs
xlai20: Done. Copied this processing logic from SignedUrlBuilder into post_policy.rs.

I tried to do a refactoring so that signed_url.rs and post_policy.rs can have shared function; however, this end up to be quite a lot of file changes and also changes to the code in signed_url.rs. I don't want this PR to affect the existing function in signed_url.rs, so let's do the refactoring in a different PR.

joshuatants: SGTM

joshuatants (approved):

src/storage/src/storage/post_policy.rs
```diff
@@ -0,0 +1,858 @@
...
+    content_length_range: Option<(u64, u64)>,
+    fields: BTreeMap<String, String>,
+    client_email: Option<String>,
+    universe_domain: Option<String>,
+    endpoint: Option<String>,
+}
+
+/// The result of signing a V4 POST Policy Document.
+#[derive(Debug, Clone, serde::Serialize)]
+pub struct PostPolicyV4Result {
```
dbolduc, line 101: can you imagine this thing growing?

If so, tag the struct as `#[non_exhaustive]`

xlai20: Done. Tagged the struct as `#[non_exhaustive]`. Thanks for pointing this out.

We can't predict if there'll be product evolution in the future; for backward compatibility, tagging this public data structure to prevent user doing exhaustive pattern matching is a safer way.

dbolduc (approved):

PR #5915, opened 2026-06-18
impl(bigquery): generate unified query metadata

alvarowolfx:
Towards #5844


src/bigquery/src/generated/query_metadata.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// Output only. The number of rows affected by a DML statement. Present only
+    /// for DML statements INSERT, UPDATE or DELETE.
+    pub num_dml_affected_rows: std::option::Option<wkt::Int64Value>,
+
+    /// A token used for paging results.  When this token is non-empty, it
+    /// indicates additional results are available.
+    pub page_token: std::string::String,
+
+    /// Auto-generated ID for the query.
+    pub query_id: std::string::String,
```
dbolduc, line 89: Q: How do you feel about this and other fields having the default value when we have a `GetQueryResultsResponse`?

Seems like most fields are `Option<>`s, so `None` is an accurate representation. But this and `location` are `Strings`. Is it misleading?

src/bigquery/src/generated/query_metadata.rs
```diff
@@ -0,0 +1,453 @@
...
+    /// <https://cloud.google.com/bigquery/docs/locations>
+    pub location: std::string::String,
+
+    /// Output only. The number of rows affected by a DML statement. Present only
+    /// for DML statements INSERT, UPDATE or DELETE.
+    pub num_dml_affected_rows: std::option::Option<wkt::Int64Value>,
+
+    /// A token used for paging results.  When this token is non-empty, it
+    /// indicates additional results are available.
+    pub page_token: std::string::String,
```
dbolduc, line 86: should we expose `page_token`? We could make it `pub(crate)` and maybe spare ourselves from customers asking us what to do with their page tokens.

alvarowolfx:
https://github.com/googleapis/librarian/pull/6498 fixes the docs issue

src/bigquery/src/generated/query_metadata.rs
alvarowolfx: the sad thing is that on the service side they are non optional Strings. I had some code in the early prototype to make `query_id` and `location` optional, but then we have the problem that we add an extra thing for us to keep an eye and always maintain fields that should have this behavior. I think we should just keep it the way the service defined (for better or for worse)

src/bigquery/src/generated/query_metadata.rs
alvarowolfx: we would need some special code on Sidekick to make `page_token` `pub(crate)`, specially as now we just use the common templates. As of right now is easier to just filter it out. 

Maybe I need to check some TODOs and make those skipped fields via librarian.yaml

dbolduc: Yeah, we could do something like this, but for fields:

https://github.com/googleapis/librarian/blob/main/internal/sidekick/rust/codec.go#L332-L333

Not important though, just a thought.

dbolduc (approved):

PR #5916, opened 2026-06-18
impl(bigquery): basic row parsing

alvarowolfx:
Towards #5844


alvarowolfx:
Splitting errors into https://github.com/googleapis/google-cloud-rust/pull/5917. Later will split FromSql initial impl.

alvarowolfx:
Split FromSql trait and some basic conversion into PR https://github.com/googleapis/google-cloud-rust/pull/5924

src/bigquery/src/query/row.rs
```diff
@@ -12,25 +12,258 @@
...
+        let mut values = ListValue::new();
+        for (i, cell) in field_list.iter().enumerate() {
+            let value = get_field_value(cell)?;
+            match schema.get_field_by_index(i) {
+                Some(f) => {
+                    let field_type = f.r#type.clone();
+                    let schema = Arc::new(Schema::new_from_field(f.clone()));
+                    let value = convert_value(value, field_type, &schema)?;
+                    values.push(value);
+                }
```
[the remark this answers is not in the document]
alvarowolfx: we have to clone later for nested records

src/bigquery/src/query/row.rs
```diff
@@ -12,25 +12,298 @@
...
 #[derive(Clone, Debug)]
 pub struct Row {
-    pub(crate) values: wkt::Value,
+    pub(crate) values: Value,
     pub(crate) schema: Arc<Schema>,
 }
 
+pub(crate) mod private {
+    /// A sealed trait to prevent external implementation of `ColumnIndex`.
+    pub trait Sealed {}
```
dbolduc: I think we more commonly say:

```rs
mod sealed {
  pub trait ColumnIndex {}
}
```

src/bigquery/src/query/row.rs
```diff
@@ -12,25 +12,298 @@
...
+        "TIMESTAMP" => Ok(Value::String(value)),
+        "DATE" => Ok(Value::String(value)),
+        "TIME" => Ok(Value::String(value)),
+        "DATETIME" => Ok(Value::String(value)),
+        "NUMERIC" | "BIGNUMERIC" => Ok(Value::String(value)),
+        "BIGINT" => Ok(Value::String(value)),
+        "GEOGRAPHY" => Ok(Value::String(value)),
+        "JSON" => Ok(Value::String(value)),
+        "INTERVAL" => Ok(Value::String(value)),
+        "RANGE" => Ok(Value::String(value)),
```
dbolduc: optional style nit: combine all of these? `"STRING" | "BYTES" | ... | "RANGE" => Ok(Value::String(value))`

src/bigquery/src/query/row.rs
```diff
@@ -12,25 +12,298 @@
...
+                { "v": "James" },
+                { "v": "272793" },
+                { "v": "TRUE" },
+                { "v": null },
+                { "v": "64.0" },
+            ]
+        })
+        .as_object()
+        .expect("row is an object")
+        .clone();
```
dbolduc: optionally, you could make the json Map directly and avoid `expect()` and the `clone()`:

```rs
use serde_json::{json, Map};

let raw_row = Map::from_iter([(
    "f".to_string(),
    json!([
        { "v": "James" },
        { "v": "272793" },
        { "v": "TRUE" },
        { "v": null },
        { "v": "64.0" },
    ]),
)]);
```

src/bigquery/src/query/row.rs
```diff
@@ -12,25 +12,298 @@
...
+    #[test_case("FLOAT64", "-inf", Some(Value::String("-inf".to_string())); "float negative infinity")]
+    #[test_case("FLOAT", "abc", None; "float invalid")]
+    #[test_case("BOOLEAN", "true", Some(Value::Bool(true)); "boolean true lowercase")]
+    #[test_case("BOOLEAN", "TRUE", Some(Value::Bool(true)); "boolean true uppercase")]
+    #[test_case("BOOL", "false", Some(Value::Bool(false)); "bool false")]
+    #[test_case("BOOL", "invalid", None; "bool invalid")]
+    #[test_case("UNKNOWN", "value", None; "unknown type")]
+    fn convert_basic_type_cases(field_type: &str, value: &str, expected: Option<Value>) {
+        let res = convert_basic_type(value.to_string(), "test_col", field_type);
+        match expected {
```
dbolduc: Any `if`s in a test are a hint that we should separate the tests. We probably want 3 tests:

- `convert_basic_type_cases_success`
- `convert_basic_type_cases_conversion_fail`
- `convert_basic_type_cases_invalid_row_format`

src/bigquery/src/query/row.rs
```diff
@@ -12,25 +12,298 @@
...
+
+        let val = self
+            .values
+            .get(idx)
+            .ok_or_else(|| RowError::IndexOutOfRange {
+                index: idx,
+                len: self.schema.len(),
+            })?;
+
+        T::from_sql(val.clone()).map_err(|e| {
```
dbolduc, line 110: Aside: we could probably have an API like a:

```rs
pub fn take<T: FromSql, I: ColumnIndex>(&mut self, index: I) -> Result<T>
```

that can extract the value without needing to clone.

It has downsides of replacing with a default value though.

---

I think it is not necessary if we have a zero-clone way to convert the whole row into an application struct. Just a thought.

dbolduc (approved):

PR #5917, opened 2026-06-18
impl(bigquery): add row parsing errors

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery/src/error.rs
+++ b/src/bigquery/src/error.rs
@@ -48,7 +48,7 @@ pub enum QueryError {
     },
 }
 
-/// Errors that can occur when retrieving value cells from a [`Row`](crate::query::Row) or iterating over query results.
+/// Errors that can occur when retrieving value cells from a Row or iterating over query results.
 #[derive(thiserror::Error, Debug)]
 #[non_exhaustive]
 pub enum RowError {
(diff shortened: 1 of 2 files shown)


src/bigquery/src/error.rs
```diff
@@ -48,6 +48,74 @@ pub enum QueryError {
...
+    },
+}
+
+/// Represents failures when converting a raw BigQuery cell value (`wkt::Value`) to a Rust type.
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+pub enum ConvertError {
+    /// The value type did not match the expected type.
+    #[error("type mismatch, expected {expected}, got {got:?}")]
+    TypeMismatch {
```
dbolduc, line 99: Q: what is the difference between `TypeMismatch` and `Convert` ?

dbolduc (approved):

src/bigquery/src/error.rs
alvarowolfx: `TypeMismatch` is an error for when the value coming under the hood is not what we expect (like we expect a `TIMESTAMP` column to be a int64 in String format, but if we get another format, we don't even try to parse and raise `TypeMismatch`.

`Convert` is when the types are correct, but when actually trying to parse, it fails. Maybe we are parsing a `INTEGER`, which is expected to come as `String`, it comes like that, but inside the string there is a non integer value. 

So `TypeMismatch` doesn't even try to parse as the types are incorrect and `Convert` is that the types match, but parsing failed.

PR #5918, opened 2026-06-18
impl(o11y): box futures in `DiscoveryPoller` to prevent stack overflow

haphungw:
When tracing is enabled, `DiscoveryPoller` is wrapped inside nesting decorators. Because it uses generic type parameters `S` (start future) and `Q` (query future), the compiler generates extremely deeply nested type definitions for every generic type combination.

At runtime, executing these deeply nested future types requires a lot of stack memory, resulting in a stack overflow in LRO-heavy tests (`compute::run_compute_lro_errors`).

Therefore, instead of keeping the futures as generic type parameters, we store them as boxed futures on the heap (`Pin<Box<dyn Future + Send>>`). This erases their concrete types and flattens their size on the stack.

### DIFF
--- a/src/lro/src/internal/discovery.rs
+++ b/src/lro/src/internal/discovery.rs
@@ -77,9 +77,9 @@ where
     Q: FnMut(String) -> QF + Send + Sync + Clone + 'static,
     QF: std::future::Future<Output = Result<O>> + Send + 'static,
 {
-    let mut query_clone = query;
+    let mut query = query;
     let start_boxed = Box::pin(start());
-    let query_boxed = Box::new(move |name| Box::pin(query_clone(name)) as BoxedFuture<Result<O>>);
+    let query_boxed = Box::new(move |name| Box::pin(query(name)) as BoxedFuture<Result<O>>);
     DiscoveryPoller::new(
         polling_error_policy,
         polling_backoff_policy,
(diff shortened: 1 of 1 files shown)


src/lro/src/internal/discovery.rs
```diff
@@ -62,37 +62,47 @@ pub trait DiscoveryOperation {
     }
 }
 
+type BoxedFuture<T> = std::pin::Pin<Box<dyn std::future::Future<Output = T> + Send + 'static>>;
```
westarle, line 69: This should have a comment saying something about the tradeoffs we are making:

* Instrumented LRO futures can get very large and cause stack overflows, so we need to box them.
* The impact of moving these to the heap should be OK since polling is bound to network requests.

I'd also suggest having a Rust expert give you feedback as well, in case there are gotchas or alternatives we aren't considering.

src/lro/src/internal/discovery.rs
```diff
@@ -62,37 +62,51 @@ pub trait DiscoveryOperation {
     }
 }
 
+// Instrumented LRO futures can get very large and cause stack overflows
+// (especially when nested decorators are used for tracing), so we box them.
+// The performance impact of moving these to the heap is negligible because
+// polling is bound to network requests.
```
westarle:
```suggestion
// polling performance is bound to network requests.
```

westarle (approved):

PR #5924, opened 2026-06-22
impl(bigquery): add FromSql trait to read/convert columns from a Row

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery/src/query/from_sql.rs
+++ b/src/bigquery/src/query/from_sql.rs
@@ -1,0 +1,179 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::error::ConvertError;
+
+/// Converts BigQuery internal [wkt::Value] to Rust types.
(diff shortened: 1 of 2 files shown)


src/bigquery/src/query/from_sql.rs
```diff
@@ -0,0 +1,179 @@
...
+
+    impl From<ConvertError> for TestConvertError {
+        fn from(err: ConvertError) -> Self {
+            match err {
+                ConvertError::NotNull => Self::NotNull,
+                ConvertError::TypeMismatch { expected, .. } => Self::TypeMismatch(expected),
+                ConvertError::Convert(e) => Self::Convert(e.to_string()),
+            }
+        }
+    }
```
dbolduc, line 130: Can you try something like:

```rs
#[derive(...)]
#[cfg_attr(test, derive(PartialEq))]
enum ConvertError { ... }
```

src/bigquery/src/query/from_sql.rs
```diff
@@ -0,0 +1,179 @@
...
+
+impl FromSql for i64 {
+    fn from_sql(value: wkt::Value) -> Result<Self, ConvertError> {
+        match value {
+            wkt::Value::Number(n) => n
+                .as_i64()
+                .ok_or_else(|| ConvertError::Convert("number is not a valid i64".into())),
+            wkt::Value::String(s) => s
+                .parse::<i64>()
+                .map_err(|e| ConvertError::Convert(Box::new(e))),
```
dbolduc, line 50: comment: Ah, we try to accept both. That's good.

dbolduc (approved):

src/bigquery/src/query/from_sql.rs
alvarowolfx: PartialEq doens't work for `Box<dyn std::error::Error + Send + Sync + 'static>` in the `ConvertError::Convert` branch. That's why the TestConvertError does this: `ConvertError::Convert(e) => Self::Convert(e.to_string()),`

PR #5927, opened 2026-06-23
feat(storage): implement appendable upload traits and wrappers

vsharonlynn:

PR #5932, opened 2026-06-24
feat(storage): add error handling, interceptors, utilities

vsharonlynn:
This PR comes after PR #5927.


joshuatants:
I see a bunch of `#[allow(dead_code)]`, which I assume are there because the code will be used in follow-up PRs, is that right?

vsharonlynn:
> I see a bunch of `#[allow(dead_code)]`, which I assume are there because the code will be used in follow-up PRs, is that right?

Yes. These will be used in the next PR.

Without `#[allow(dead_code)]`, this command will fail:
 RUSTFLAGS="--cfg google_cloud_unstable_storage_bidi" cargo clippy --package google-cloud-storage --profile test --features unstable-stream -- -D warnings

src/storage/src/storage/bidi_write/redirect.rs
```diff
@@ -0,0 +1,182 @@
...
+                guard.write_handle = redirect.write_handle;
+                if let Some(generation) = redirect.generation {
+                    guard.generation = generation;
+                }
+                break;
+            }
+        }
+    }
+    to_gax_error(status)
+}
```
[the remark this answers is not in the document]
vsharonlynn: I'm directly copying https://github.com/googleapis/google-cloud-rust/blob/main/src/storage/src/storage/bidi/redirect.rs#L25-L37.

src/storage/src/storage/bidi_write/redirect.rs
```diff
@@ -0,0 +1,182 @@
...
+            generation: 1,
+            routing_token: Some("initial-token".into()),
+            write_handle: Some(BidiWriteHandle {
+                handle: bytes::Bytes::from_static(b"initial-handle"),
+            }),
+            ..Default::default()
+        };
+        let spec = Arc::new(Mutex::new(spec));
+
+        let got = handle_redirect(spec.clone(), status);
```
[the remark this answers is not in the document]
vsharonlynn: Ditto.

src/storage/src/storage/bidi_write/redirect.rs
```diff
@@ -0,0 +1,182 @@
...
+            bucket: "test-bucket".into(),
+            object: "test-object".into(),
+            generation: 1,
+            routing_token: Some("initial-token".into()),
+            write_handle: Some(initial_handle.clone()),
+            ..Default::default()
+        };
+        let spec = Arc::new(Mutex::new(spec));
+
+        let got = handle_redirect(spec.clone(), status);
```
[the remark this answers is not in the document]
vsharonlynn: Ditto.

src/storage/src/storage/bidi_write/retry_redirect.rs
```diff
@@ -0,0 +1,101 @@
...
+            RetryResult::Permanent(e) if is_redirect(&e) => RetryResult::Continue(e),
+            // Exhausted(), Continue() and other permanent errors pass thru.
+            result => result,
+        }
+    }
+
+    fn on_throttle(&self, state: &RetryState, error: Error) -> ThrottleResult {
+        self.inner.on_throttle(state, error)
+    }
+}
```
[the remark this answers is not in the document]
vsharonlynn: I'm closely following https://github.com/googleapis/google-cloud-rust/blob/main/src/storage/src/storage/bidi/retry_redirect.rs#L44

src/storage/src/storage/bidi_write/mocks.rs
```diff
@@ -12,8 +12,18 @@
...
+
+#[allow(dead_code)]
+pub type MockStream = tokio::sync::mpsc::Receiver<TonicResult<BidiWriteObjectResponse>>;
+#[allow(dead_code)]
+pub type MockStreamSender = Sender<TonicResult<BidiWriteObjectResponse>>;
+
+#[allow(dead_code)]
+pub fn mock_stream() -> (MockStreamSender, MockStream) {
+    tokio::sync::mpsc::channel(10)
+}
```
coryan: That seems like a lot of ceremony to say `mock_stream()` instead of `tokio::sync::mpsc::channel(10)`

src/storage/src/storage/bidi_write/redirect.rs
```diff
@@ -0,0 +1,182 @@
...
+                guard.routing_token = redirect.routing_token;
+                guard.write_handle = redirect.write_handle;
+                if let Some(generation) = redirect.generation {
+                    guard.generation = generation;
+                }
+                break;
+            }
+        }
+    }
+    to_gax_error(status)
```
coryan:
```suggestion
    let Ok(status) = RpcStatus::decode(status.details()) else {
        return to_gax_error(status);
    }
    for d in status.details {
        let Ok(redirect) = d.to_msg::<BidiWriteObjectRedirectedError>() else {
          continue;
        }
        let mut guard = spec.lock().expect("never poisoned");
        guard.routing_token = redirect.routing_token;
        guard.write_handle = redirect.write_handle;
        if let Some(generation) = redirect.generation {
          guard.generation = generation;
        }
        break;
    }
```

Though you could probably do something like:

```
if let Some(redirect) = status.details.find(|d| d.to_msg::<...>().ok()) {
  let mut guard = ...
}
```

src/storage/src/storage/bidi_write/redirect.rs
```diff
@@ -0,0 +1,182 @@
...
+}
+
+/// Determine if an error is a redirect error.
+#[allow(dead_code)]
+pub fn is_redirect(error: &Error) -> bool {
+    if error.status().is_none_or(|s| s.code != Code::Aborted) {
```
coryan, line 56: Maybe document why the `Code::Aborted` is necessary? Can the service return a redirect on a different status code? Why not?

src/storage/src/appendable_object_writer.rs
```diff
@@ -0,0 +1,252 @@
+// Copyright 2025 Google LLC
```
coryan: For any new files: 

```suggestion
// Copyright 2026 Google LLC
```

src/storage/src/storage/bidi_write/retry_redirect.rs
```diff
@@ -0,0 +1,101 @@
+// Copyright 2025 Google LLC
```
coryan:
```suggestion
// Copyright 2026 Google LLC
```

src/storage/src/storage/bidi_write/stub.rs
```diff
@@ -0,0 +1,55 @@
+// Copyright 2025 Google LLC
```
coryan:
```suggestion
// Copyright 2026 Google LLC
```

src/storage/src/storage/bidi_write.rs
```diff
@@ -0,0 +1,157 @@
+// Copyright 2025 Google LLC
```
coryan:
```suggestion
// Copyright 2026 Google LLC
```

coryan (changes requested):
Please make sure these changes are not exposed in any public API.  There are cargo tools to print the public API if you need them.

src/storage/src/storage/bidi_write/mocks.rs
vsharonlynn: Removed. Thanks!
Keeping the file empty, because this file will be used in the PR right after this one. Hope that's ok.

src/storage/src/storage/bidi_write/redirect.rs
vsharonlynn: Updated. Looks better. Thank you!

src/storage/src/storage/bidi_write/redirect.rs
vsharonlynn: Added documentation... PTAL. Thanks.

Is my understanding correct:
BidiWriteObjectRedirectError will always come from a response with an ABORTED error. The Cloud public doc on [BidiWriteObjectRedirectError](https://docs.cloud.google.com/storage/docs/reference/rpc/google.storage.v2#bidiwriteobjectredirectederror
) says “Error proto containing details for a redirected write. This error might be attached as details for an ABORTED response to BidiWriteObject.”

I also checked [Python](https://github.com/googleapis/google-cloud-python/blob/main/packages/google-cloud-storage/google/cloud/storage/asyncio/retry/_helpers.py#L100) and [Go](https://github.com/googleapis/google-cloud-go/blob/9968adfbea5b6bba137101ac446132fac8616734/storage/grpc_writer.go#L1536) explicitly check for ABORTED before looking for the BidiWriteObjectRedirectedError.

src/storage/src/storage/bidi_write/mocks.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright 2026 Google LLC
```
coryan: Seems like you could remove this file until the right PR, but feel free to leave it here if it helps you.

coryan (approved):

vsharonlynn:
> Please make sure these changes are not exposed in any public API. There are cargo tools to print the public API if you need them.

I ran:
```
$ cargo +nightly public-api -p google-cloud-storage diff origin/main..HEAD
```

The output is:
```
Removed items from the public API
=================================
(none)

Changed items in the public API
===============================
(none)

Added items to the public API
=============================
(none)
```

Thanks!

src/storage/src/storage/bidi_write/mocks.rs
vsharonlynn: Removed. The PR is cleaner when removed. Thank you.

coryan (approved):

PR #5952, opened 2026-06-26
fix(o11y): fix `record_discovery_polling_result` macro compilation failure

haphungw:
Update `record_discovery_polling_result!` macro to extract status code, message, and `error.type` using only public types, resolving compilation failures in generated crates.

### DIFF
--- a/src/lro/src/internal/tracing.rs
+++ b/src/lro/src/internal/tracing.rs
@@ -92,11 +92,8 @@ macro_rules! record_discovery_polling_result {
             if let Some(status) = error {
                 span.record("otel.status_code", "ERROR");
                 span.record("otel.status_description", &status.message);
-                let err = $crate::Error::service(status);
-                span.record(
-                    "error.type",
-                    ::gaxi::observability::errors::error_type(&err),
-                );
+                span.record("rpc.response.status_code", status.code as i32);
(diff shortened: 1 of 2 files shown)


src/lro/src/internal/tracing.rs
```diff
@@ -92,11 +92,8 @@ macro_rules! record_discovery_polling_result {
             if let Some(status) = error {
                 span.record("otel.status_code", "ERROR");
                 span.record("otel.status_description", &status.message);
-                let err = $crate::Error::service(status);
-                span.record(
-                    "error.type",
-                    ::gaxi::observability::errors::error_type(&err),
-                );
+                span.record("rpc.response.status_code", status.code as i32);
+                span.record("error.type", status.code.to_string());
```
westarle, line 96: Was this change intentional?

haphungw: Yes, it is. Since this macro is used in discovery-based clients (in this case, `compute`), calling `error_type` would force us to expose `gaxi` as a public dependency. 
Replacing it with `status.code.to_string()` avoids that leak while recording the exact same string (e.g., `"NOT_FOUND"`).

westarle (approved):

PR #5955, opened 2026-06-26
feat(o11y): extract error details and type for Discovery LROs

haphungw:
Now that the generator template and the LRO macro are fixed, we implement the actual telemetry error mapping for Compute Engine LROs and adds integration test assertions.

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 language: rust
-version: v0.22.0
+version: bb509a740d375c1ca1709e624e404b69924fdfa7
 repo: googleapis/google-cloud-rust
 sources:
   conformance:
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -165,7 +165,6 @@ libraries:
   - name: google-cloud-agentregistry-v1
     version: 1.0.0
     copyright_year: "2026"
-    rust: {}
   - name: google-cloud-aiplatform-v1
     version: 1.13.0
     copyright_year: "2025"
@@ -548,7 +547,6 @@ libraries:
   - name: google-cloud-ces-v1
     version: 1.3.0
     copyright_year: "2026"
-    rust: {}
   - name: google-cloud-chronicle-v1
     version: 1.12.0
     copyright_year: "2025"
@@ -866,7 +864,6 @@ libraries:
   - name: google-cloud-geminidataanalytics-v1
     version: 1.1.0
     copyright_year: "2026"
-    rust: {}
   - name: google-cloud-gkebackup-v1
     version: 1.12.0
     copyright_year: "2025"
@@ -945,7 +942,6 @@ libraries:
   - name: google-cloud-hypercomputecluster-v1
     version: 1.3.0
     copyright_year: "2026"
-    rust: {}
   - name: google-cloud-iam-admin-v1
     version: 1.10.0
     apis:
@@ -1347,7 +1343,6 @@ libraries:
   - name: google-cloud-securitycentermanagement-v1
     version: 1.2.0
     copyright_year: "2026"
-    rust: {}
   - name: google-cloud-securityposture-v1
     version: 1.12.0
     copyright_year: "2025"
@@ -1638,7 +1633,6 @@ libraries:
   - name: google-cloud-vectorsearch-v1
     version: 1.3.0
     copyright_year: "2026"
-    rust: {}
   - name: google-cloud-video-livestream-v1
     version: 1.12.0
     copyright_year: "2025"
@@ -1657,7 +1651,6 @@ libraries:
   - name: google-cloud-visionai-v1
     version: 1.3.0
     copyright_year: "2026"
-    rust: {}
   - name: google-cloud-vmmigration-v1
     version: 1.12.0
     copyright_year: "2025"
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 language: rust
-version: bb509a740d375c1ca1709e624e404b69924fdfa7
+version: v0.24.1-0.20260706153224-e26270201c2b
 repo: googleapis/google-cloud-rust
 sources:
   conformance:
--- a/src/generated/cloud/compute/v1/src/operation.rs
+++ b/src/generated/cloud/compute/v1/src/operation.rs
@@ -13,6 +13,7 @@
 // limitations under the License.
 
 use crate::model::Operation;
+use google_cloud_gax::error::rpc::{Code, Status};
 
 impl google_cloud_lro::internal::DiscoveryOperation for Operation {
     fn name(&self) -> Option<&String> {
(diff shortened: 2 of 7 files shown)


src/generated/cloud/compute/v1/src/operation.rs
```diff
@@ -21,4 +21,60 @@ impl google_cloud_lro::internal::DiscoveryOperation for Operation {
...
+                };
+            }
+            if let Some(err_msg) = &first_err.message {
+                message = err_msg.clone();
+            }
+        }
+
+        if code == google_cloud_gax::error::rpc::Code::Ok && http_status == 200 {
+            return None;
+        }
```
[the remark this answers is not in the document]
haphungw: Although `operation.rs` lives inside `src/generated`, it is a hand-written veneer file. Therefore, direct modification is correct and intended.

src/generated/cloud/compute/v1/src/operation.rs
```diff
@@ -21,4 +21,64 @@ impl google_cloud_lro::internal::DiscoveryOperation for Operation {
...
+            404 => google_cloud_gax::error::rpc::Code::NotFound,
+            409 => google_cloud_gax::error::rpc::Code::AlreadyExists,
+            429 => google_cloud_gax::error::rpc::Code::ResourceExhausted,
+            499 => google_cloud_gax::error::rpc::Code::Cancelled,
+            500 => google_cloud_gax::error::rpc::Code::Internal,
+            501 => google_cloud_gax::error::rpc::Code::Unimplemented,
+            503 => google_cloud_gax::error::rpc::Code::Unavailable,
+            504 => google_cloud_gax::error::rpc::Code::DeadlineExceeded,
+            _ => google_cloud_gax::error::rpc::Code::Unknown,
+        };
```
westarle: I'm a bit surprised to see this in generated code -- is it possible to re-use some existing mapping or move this to gaxi?  Same for string mapping below.

coryan: `operation.rs` is not generated, we have a small number of hand-crafted files in this crate:

https://github.com/googleapis/google-cloud-rust/blob/3872d2885c6da3a9463e85b50bf1fe8e9ddc1fa1/librarian.yaml#L571-L573

westarle: Thanks for the explanation!

src/generated/cloud/compute/v1/src/operation.rs
```diff
@@ -21,4 +21,64 @@ impl google_cloud_lro::internal::DiscoveryOperation for Operation {
...
+
+        let mut message = self.http_error_message.clone().unwrap_or_default();
+
+        if let Some(first_err) = self.error.as_ref().and_then(|err| err.errors.first()) {
+            if code == google_cloud_gax::error::rpc::Code::Ok {
+                code = google_cloud_gax::error::rpc::Code::Unknown;
+            }
+            if let Some(err_code) = &first_err.code {
```
westarle, line 58: If these are mappable to gRPC errors, can we use the standard conversion in gax (handling `QUOTA_EXCEEDED` and any other special values specially)?  This would make it more clear what is special about these errors.

https://github.com/googleapis/google-cloud-rust/blob/main/src/gax/src/error/rpc.rs#L295-L319

src/generated/cloud/dns/v1/src/tracing.rs
```diff
@@ -366,9 +366,7 @@ where
                 match &result {
                     Ok(response) => {
                         let op = response.body();
-                        // TODO(https://github.com/googleapis/librarian/issues/6286): Track recording error info for Discovery LROs
-                        let done = google_cloud_lro::internal::DiscoveryOperation::done(op);
-                        _span.record("gcp.longrunning.done", done);
+                        google_cloud_lro::record_discovery_polling_result!(&_span, op);
```
westarle, line 369: will you add an implementation of `error()` for DNS in the future?

librarian.yaml
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 language: rust
-version: v0.22.0
+version: bb509a740d375c1ca1709e624e404b69924fdfa7
```
noahdietz: This should be a pseudo-verison, not a commit

```
V=$(GOPROXY=direct go list -m -f '{{.Version}}' github.com/googleapis/librarian@main)
go run github.com/googleapis/librarian/cmd/librarian@${V} config set version $V
```

haphungw: thanks for the catch

src/generated/cloud/dns/v1/src/tracing.rs
haphungw: probably not. 

The DNS Operation resource doesn't give us error fields to map from (it just has a pending/done status). Since there's no error info to extract from it, the default trait implementation returning `None` already does what we need.

src/generated/cloud/compute/v1/src/operation.rs
haphungw: done. we now also handle `QUOTE_EXCEEDED` as a fallback.

src/generated/cloud/compute/v1/src/operation.rs
```diff
@@ -21,4 +21,56 @@ impl google_cloud_lro::internal::DiscoveryOperation for Operation {
...
+                if let Ok(c) = google_cloud_gax::error::rpc::Code::try_from(err_code.as_str()) {
+                    code = c;
+                } else if err_code == "QUOTA_EXCEEDED" {
+                    code = google_cloud_gax::error::rpc::Code::ResourceExhausted;
+                }
+            }
```
westarle, line 63: This might read better as:
```suggestion
    if let Some(err_code) = &first_err.code {
        code = match err_code.as_str() {
            "QUOTA_EXCEEDED" => Code::ResourceExhausted,
            other => Code::try_from(other).unwrap_or(code),
        };
    }
```

src/generated/cloud/compute/v1/src/operation.rs
```diff
@@ -21,4 +21,56 @@ impl google_cloud_lro::internal::DiscoveryOperation for Operation {
     fn done(&self) -> bool {
         self.status == Some(crate::model::operation::Status::Done)
     }
+    fn error(&self) -> Option<google_cloud_gax::error::rpc::Status> {
```
westarle: can we `use google_cloud_gax::error::rpc`?

src/generated/cloud/compute/v1/src/operation.rs
```diff
@@ -21,4 +21,56 @@ impl google_cloud_lro::internal::DiscoveryOperation for Operation {
...
+        if self.error.is_none()
+            && self.http_error_status_code.is_none()
+            && self.http_error_message.is_none()
+        {
+            return None;
+        }
+
+        let mut status = google_cloud_gax::error::rpc::Status::default();
+
+        let http_status = self.http_error_status_code.unwrap_or(200);
```
westarle, line 35: I think this should probably live in gax::error::rpc::Code but I don't think a new public API is warranted for this one use case.

Next time we need it, we should consider introducing it onto a shared type.

westarle (approved):

src/generated/cloud/compute/v1/src/operation.rs
haphungw: done

src/generated/cloud/compute/v1/src/operation.rs
haphungw: done

PR #5956, opened 2026-06-28
feat(storage): implement appendable upload connector

vsharonlynn:
This PR comes after PR #5932.

### DIFF
--- a/src/storage/src/storage/bidi_write.rs
+++ b/src/storage/src/storage/bidi_write.rs
@@ -14,9 +14,14 @@
 
 //! Internal traits and types for Appendable Object Write (Bidi Write).
 
+#[allow(dead_code)]
 pub(crate) mod connector;
+#[allow(dead_code)]
 mod redirect;
+#[allow(dead_code)]
 mod retry_redirect;
+#[allow(dead_code)]
+pub(crate) mod state;
 pub(crate) mod stub;
 
 use crate::google::storage::v2::{BidiWriteObjectRequest, BidiWriteObjectResponse};
--- a/src/storage/src/storage/bidi_write.rs
+++ b/src/storage/src/storage/bidi_write.rs
@@ -14,8 +14,6 @@
 
 //! Internal traits and types for Appendable Object Write (Bidi Write).
 
-#![allow(dead_code)]
-
 pub(crate) mod connector;
 mod redirect;
 mod retry_redirect;
@@ -30,6 +28,7 @@ use tokio::sync::mpsc::Receiver;
 /// A trait to mock `Streaming<T>` in the unit tests.
 ///
 /// This is not a public trait, we only need this for our own testing.
+#[allow(dead_code)]
 pub(crate) trait TonicStreaming: std::fmt::Debug + Send + 'static {
     fn next_message(
         &mut self,
@@ -46,6 +45,7 @@ impl TonicStreaming for Streaming<BidiWriteObjectResponse> {
 /// A trait to mock `gaxi::grpc::Client` in the unit tests.
 ///
 /// This is not a public trait, we only need this for our own testing.
+#[allow(dead_code)]
 pub(crate) trait Client: std::fmt::Debug + Send + 'static {
     type Stream: Sized;
     fn start(
@@ -139,12 +139,14 @@ pub(crate) mod tests {
         )
     }
 
+    #[allow(dead_code)]
     pub(crate) fn test_options() -> RequestOptions {
         let mut options = RequestOptions::new();
         options.backoff_policy = Arc::new(test_backoff());
         options
     }
 
+    #[allow(dead_code)]
     fn test_backoff() -> impl google_cloud_gax::backoff_policy::BackoffPolicy {
         use std::time::Duration;
         google_cloud_gax::exponential_backoff::ExponentialBackoffBuilder::new()
--- a/src/storage/src/storage/bidi_write.rs
+++ b/src/storage/src/storage/bidi_write.rs
@@ -14,6 +14,9 @@
 
 //! Internal traits and types for Appendable Object Write (Bidi Write).
 
+#![allow(dead_code)]
+
+pub(crate) mod connector;
 mod redirect;
 mod retry_redirect;
 pub(crate) mod stub;
@@ -27,7 +30,6 @@ use tokio::sync::mpsc::Receiver;
 /// A trait to mock `Streaming<T>` in the unit tests.
 ///
 /// This is not a public trait, we only need this for our own testing.
-#[allow(dead_code)]
 pub(crate) trait TonicStreaming: std::fmt::Debug + Send + 'static {
     fn next_message(
         &mut self,
@@ -44,7 +46,6 @@ impl TonicStreaming for Streaming<BidiWriteObjectResponse> {
 /// A trait to mock `gaxi::grpc::Client` in the unit tests.
 ///
 /// This is not a public trait, we only need this for our own testing.
-#[allow(dead_code)]
 pub(crate) trait Client: std::fmt::Debug + Send + 'static {
     type Stream: Sized;
     fn start(
@@ -81,6 +82,9 @@ impl Client for gaxi::grpc::Client {
         .await
     }
 }
+
+#[cfg(test)]
+mod mocks;
 
 #[cfg(test)]
 pub(crate) mod tests {
@@ -135,14 +139,12 @@ pub(crate) mod tests {
         )
     }
 
-    #[allow(dead_code)]
     pub(crate) fn test_options() -> RequestOptions {
         let mut options = RequestOptions::new();
         options.backoff_policy = Arc::new(test_backoff());
         options
     }
 
-    #[allow(dead_code)]
     fn test_backoff() -> impl google_cloud_gax::backoff_policy::BackoffPolicy {
         use std::time::Duration;
         google_cloud_gax::exponential_backoff::ExponentialBackoffBuilder::new()
--- a/src/storage/src/storage/bidi_write/connector.rs
+++ b/src/storage/src/storage/bidi_write/connector.rs
@@ -306,6 +306,14 @@ mod tests {
     use std::error::Error as _;
     use std::sync::Arc;
 
+    use super::super::mocks::{MockTestClient, SharedMockClient};
+    use super::super::tests::{permanent_error, redirect_status};
+    use gaxi::grpc::tonic::GrpcMethod;
+    use gaxi::grpc::tonic::Response as TonicResponse;
+    use gaxi::grpc::tonic::Result as TonicResult;
+    use google_cloud_gax::error::binding::{BindingError, SubstitutionFail};
+    use std::sync::Mutex;
+
     fn test_credentials() -> Credentials {
         Anonymous::new().build()
     }
@@ -359,7 +367,6 @@ mod tests {
     #[test_case::test_case("")]
     #[test_case::test_case("my-bucket")]
     async fn binding_error(bucket_name: &str) -> Result<()> {
-        use super::super::mocks::{MockTestClient, SharedMockClient};
         let mut mock = MockTestClient::new();
         // Binding errors are detected before a request is sent.
         mock.expect_start().never();
@@ -382,7 +389,6 @@ mod tests {
 
         let err = connector.connect_open(req).await.unwrap_err();
         assert!(err.is_binding(), "{err:?}");
-        use google_cloud_gax::error::binding::{BindingError, SubstitutionFail};
         let source = err.source().and_then(|e| e.downcast_ref::<BindingError>());
         assert!(matches!(source, Some(BindingError { .. })), "{err:?}");
         // Extract all the field names that did not match, and expect a single name:
@@ -413,11 +419,6 @@ mod tests {
 
     #[tokio::test]
     async fn start_error() -> Result<()> {
-        use super::super::mocks::{MockTestClient, SharedMockClient};
-        use super::super::tests::permanent_error;
-        use gaxi::grpc::tonic::GrpcMethod;
-        use std::sync::Mutex;
-
         let receivers = Arc::new(Mutex::new(Vec::new()));
         let save = receivers.clone();
         let mut mock = MockTestClient::new();
@@ -480,11 +481,6 @@ mod tests {
 
     #[tokio::test]
     async fn start_error_with_routing() -> Result<()> {
-        use super::super::mocks::{MockTestClient, SharedMockClient};
-        use super::super::tests::permanent_error;
-        use gaxi::grpc::tonic::GrpcMethod;
-        use std::sync::Mutex;
-
         let receivers = Arc::new(Mutex::new(Vec::new()));
         let save = receivers.clone();
         let mut mock = MockTestClient::new();
@@ -547,10 +543,6 @@ mod tests {
 
     #[tokio::test]
     async fn start_redirect_then_error() -> Result<()> {
-        use super::super::mocks::{MockTestClient, SharedMockClient};
-        use super::super::tests::{permanent_error, redirect_status};
-        use std::sync::Mutex;
-
         let mut seq = mockall::Sequence::new();
         let mut mock = MockTestClient::new();
         let receivers = Arc::new(Mutex::new(Vec::new()));
@@ -571,7 +563,6 @@ mod tests {
                 Err(permanent_error())
             });
         let client = SharedMockClient::new(mock);
-
         let mut connector = Connector::new(test_options(), client);
 
         let mut req = OpenAppendableObjectRequest {
@@ -652,10 +643,6 @@ mod tests {
 
     #[tokio::test]
     async fn start_immediately_closed() -> Result<()> {
-        use super::super::mocks::{MockTestClient, SharedMockClient};
-        use gaxi::grpc::tonic::Response as TonicResponse;
-        use gaxi::grpc::tonic::Result as TonicResult;
-
         let (tx1, rx1) = tokio::sync::mpsc::channel::<TonicResult<BidiWriteObjectResponse>>(5);
         let stream1 = TonicResponse::from(rx1);
         drop(tx1);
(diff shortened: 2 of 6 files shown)


src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,735 @@
...
+    ) -> Result<(BidiWriteObjectResponse, Connection<T::Stream>)> {
+        let resource = match req.spec.resource {
+            Some(r) => {
+                let object: crate::google::storage::v2::Object = gaxi::prost::ToProto::to_proto(r)
+                    .map_err(|e: gaxi::prost::ConvertError| Error::io(e.to_string()))?;
+                Some(object)
+            }
+            None => None,
+        };
+        self.params = req.params.map(|p| p.to_proto().expect("valid params"));
```
joshuatants: nit: in `connect_reopen` setting the params comes after the spec is created. let's follow the same order for both.

src/storage/src/storage/bidi_write/redirect.rs
```diff
@@ -0,0 +1,264 @@
...
+use crate::google::storage::v2::{AppendObjectSpec, BidiWriteObjectRedirectedError};
+use gaxi::as_inner::as_inner;
+use gaxi::grpc::from_status::to_gax_error;
+use gaxi::grpc::tonic::Status;
+use google_cloud_gax::error::rpc::Code;
+use prost::Message;
+use std::sync::{Arc, Mutex};
+
+pub fn handle_redirect(state: Arc<Mutex<AppendObjectSpecState>>, status: Status) -> crate::Error {
+    if let Ok(status) = RpcStatus::decode(status.details()) {
```
joshuatants: let's use coryan@'s proposed restructuring (the one in the previous commit)

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -0,0 +1,438 @@
...
+            ..BidiWriteObjectRequest::default()
+        };
+
+        self.tx
+            .send(UploadIntent::Append(request))
+            .await
+            .map_err(|e| crate::Error::io(e.to_string()))?;
+
+        self.write_offset += length;
+        self.crc32c_persisted = new_crc32c;
```
joshuatants: As discussed offline, `send` here only blocks until the UploadIntent is successfully sent. We ought to wait for the server to actually ack that it has persisted the bytes before updating these. This is probably best done via shared state (likely `Arc<Mutex<...>>`) between the transport and the worker.

Perhaps add a TODO here?

joshuatants (commented):
We're gone through this extensively offline.

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Done. Thanks.

src/storage/src/storage/bidi_write/redirect.rs
vsharonlynn: Done. Thanks.

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: After our meeting this morning:
This is for append, and for appends, the server does not necessarily return any response, so we can't wait for server ack before updating these values.
These values are for internal tracking for the data that has been queued / sent to the server, not the persisted values. Persisted size is in the variable persisted_size.
I renamed crc32c_persisted to running_crc32c.

joshuatants: Thanks for clarifying.

joshuatants (approved):

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,745 @@
...
+use gaxi::grpc::Client as GrpcClient;
+use gaxi::grpc::tonic::{Extensions, GrpcMethod, Streaming};
+use gaxi::prost::ToProto;
+use google_cloud_gax::error::binding::{
+    BindingError, PathMismatch, SubstitutionFail, SubstitutionMismatch,
+};
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+use tokio::sync::mpsc::Sender;
+/// Represents a bidirectional streaming connection.
```
coryan, line 40: nit: blank line after the `use` declarations.

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,745 @@
...
+    }
+
+    pub async fn connect_open(
+        &mut self,
+        req: crate::model_ext::OpenAppendableObjectRequest,
+    ) -> Result<(BidiWriteObjectResponse, Connection<T::Stream>)> {
+        let resource = match req.spec.resource {
+            Some(r) => {
+                let object: crate::google::storage::v2::Object = gaxi::prost::ToProto::to_proto(r)
+                    .map_err(|e: gaxi::prost::ConvertError| Error::io(e.to_string()))?;
```
coryan: This is not an I/O error, this is a deserialization error, so you want:

```suggestion
                    .map_err(Error::deser)?;
```

Please don't use `e.to_string()` when mapping errors, we want to preserve the error types, not just the error messages.

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,745 @@
...
+            routing_token: req.routing_token,
+            if_metageneration_match: req.if_metageneration_match,
+            if_metageneration_not_match: req.if_metageneration_not_match,
+            write_handle: req
+                .write_handle
+                .map(|handle| crate::google::storage::v2::BidiWriteHandle { handle }),
+        };
+        self.params = req
+            .params
+            .map(|p| p.to_proto().map_err(|e| Error::io(e.to_string())))
```
coryan: Here too:


```suggestion
            .map(|p| p.to_proto().map_err(|Error::ser)
```

src/storage/src/storage/bidi_write.rs
```diff
@@ -14,9 +14,14 @@
 
 //! Internal traits and types for Appendable Object Write (Bidi Write).
 
+#![allow(dead_code)]
```
coryan: Is there any way to have a smaller scope for this?

src/storage/src/storage/bidi_write/worker.rs
```diff
@@ -0,0 +1,252 @@
...
+    ) -> LoopResult<()> {
+        let (mut rx, mut tx) = (connection.rx, connection.tx);
+
+        let error = loop {
+            tokio::select! {
+                m = rx.next_message() => {
+                    match self.handle_response(m) {
+                        // Successful end of stream, return without error.
+                        None => break None,
+                        // An unrecoverable in the stream or its data, return
```
coryan:
```suggestion
                        // An unrecoverable error in the stream or its data, return
```

src/storage/src/storage/bidi_write/worker.rs
```diff
@@ -0,0 +1,252 @@
...
+use super::{Client, TonicStreaming};
+use crate::error::WriteError;
+use crate::google::storage::v2::{BidiWriteObjectRequest, BidiWriteObjectResponse};
+use gaxi::grpc::tonic::Result as TonicResult;
+use std::sync::Arc;
+use tokio::sync::mpsc::Receiver;
+use tokio::sync::oneshot;
+
+type WriteResult<T> = std::result::Result<T, WriteError>;
+type LoopResult<T> = std::result::Result<T, Arc<crate::Error>>;
```
coryan: Why is this an `Arc<crate::Error>` ?  That is only needed if more than one caller needs to receive the same error, which happens in bidi streaming reads, but it is not clear it could happen on bidi writes.  I would expect there is only one writer at a time?

coryan (changes requested):
This is still too large.  Aim for at **most** 200 lines of new code, which probably means 400 to 600 lines per PR.  1600 lines is simply too large.

vsharonlynn:
> This is still too large. Aim for at **most** 200 lines of new code, which probably means 400 to 600 lines per PR. 1600 lines is simply too large.

Ok. I am going to split this PR into smaller PRs.  I'll update once done. Thank you and sorry about this!

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Done. Thanks!

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Done. Thanks.

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Done. Thanks.

src/storage/src/storage/bidi_write.rs
vsharonlynn: Removed this one. Thanks and sorry about that.

vsharonlynn:
> > This is still too large. Aim for at **most** 200 lines of new code, which probably means 400 to 600 lines per PR. 1600 lines is simply too large.
> 
> Ok. I am going to split this PR into smaller PRs. I'll update once done. Thank you and sorry about this!

I split this PR into three PRs:
* This current PR is now for connector-related changes only, which includes some changes in redirect and mocks for testing.
* #6001 for background worker. Still in draft, will wait for this PR to be merged before rebasing and sending it out.
* #6002 for transport. Still in draft, will wait for 6001 to be merged before rebasing and sending it out.

Thank you.

src/storage/src/storage/bidi_write/worker.rs
vsharonlynn: Going to change in the new draft PR #6001 split from this PR.

src/storage/src/storage/bidi_write/worker.rs
vsharonlynn: Going to change in the new draft PR #6001 split from this PR.

src/storage/src/storage/bidi_write/worker.rs
vsharonlynn: Done in #6001. Thanks!

src/storage/src/storage/bidi_write/worker.rs
vsharonlynn: You are correct. I removed the `Arc` wrapper and now using just the `crate::Error` in #6001. Thank you.

src/storage/src/storage/bidi_write/redirect.rs
```diff
@@ -25,7 +26,7 @@ use prost::Message;
 use std::sync::{Arc, Mutex};
 
 #[allow(dead_code)]
-pub fn handle_redirect(spec: Arc<Mutex<AppendObjectSpec>>, status: Status) -> crate::Error {
+pub fn handle_redirect(state: Arc<Mutex<AppendObjectSpecState>>, status: Status) -> crate::Error {
```
coryan: Consider:


```suggestion
pub fn handle_redirect(state: &mut AppendObjectSpecState, status: Status) -> crate::Error {
```

or make `handle_redirect()` a member function of `AppendObjectSpecState`:

```rust
impl AppendObjectSpecState {
  pub fn handle_redirect(&mut self, status: Status) -> crate::Error {
  }
```

Then you can use this (in tests) without an `Arc<Mutex<>>` and still use it in the production code as you want.

src/storage/src/storage/bidi_write/redirect.rs
```diff
@@ -34,12 +35,47 @@ pub fn handle_redirect(spec: Arc<Mutex<AppendObjectSpec>>, status: Status) -> cr
...
+            }
+            AppendObjectSpecState::Append(spec) => {
+                let mut new_spec = spec.clone();
+                new_spec.routing_token = redirect.routing_token;
+                new_spec.write_handle = redirect.write_handle;
+                if let Some(g) = redirect.generation {
+                    new_spec.generation = g;
+                }
+                AppendObjectSpecState::Append(new_spec)
+            }
```
coryan: I suspect if you receive a `&mut self` then this can be implemented using simpler assignments.

src/storage/src/storage/bidi_write/redirect.rs
```diff
@@ -34,12 +35,47 @@ pub fn handle_redirect(spec: Arc<Mutex<AppendObjectSpec>>, status: Status) -> cr
...
-        guard.routing_token = redirect.routing_token;
-        guard.write_handle = redirect.write_handle;
-        if let Some(generation) = redirect.generation {
-            guard.generation = generation;
-        }
+        let mut guard = state.lock().expect("never poisoned");
+
+        let new_state = match &*guard {
+            AppendObjectSpecState::Write(spec, _) => {
+                if let Some(generation) = redirect.generation {
```
coryan: The amount of copying and cloning here makes me sad... you should be able to do something like:

```
let new_spec = spec.resource.map(|r| /* move the data out of |r| */);
```

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,750 @@
...
+    pub fn new(tx: Sender<BidiWriteObjectRequest>, rx: S) -> Self {
+        Self { tx, rx }
+    }
+}
+
+/// Represents the state of the initial request in the stream.
+#[derive(Clone, Debug)]
+#[allow(dead_code)]
+pub enum AppendObjectSpecState {
+    Write(Box<WriteObjectSpec>, Option<String>),
```
coryan: Why a `Box<WriteObjectSpec>` ?

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,750 @@
...
+    BindingError, PathMismatch, SubstitutionFail, SubstitutionMismatch,
+};
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+use tokio::sync::mpsc::Sender;
+
+/// Represents a bidirectional streaming connection.
+/// Contains the transmission channel for requests and the receiving stream for responses.
+#[derive(Debug)]
+#[allow(dead_code)]
```
coryan: Instead of repeating this line in each element, put it at the place the module is imported.

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,750 @@
...
+                paths: vec![mismatch],
+            };
+
+            return Err(crate::Error::binding(mismatch));
+        }
+
+        let mut x_goog_request_params = format!("bucket={bucket_name}");
+        if let Some(token) = routing_token {
+            x_goog_request_params.push_str("&routing_token=");
+            x_goog_request_params.push_str(&token);
```
coryan: This needs to be percent-encoded, unless you happen to know that it never needs the percent encoding, in which can a comment justifying that would be useful.

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,750 @@
...
+        }
+
+        let (tx, rx) = tokio::sync::mpsc::channel::<BidiWriteObjectRequest>(100);
```
coryan: The magical 100 should be a configuration parameter or a constant.  I apologize because I probably left a number of these in the code for `open_object()`.

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,750 @@
...
+                    }
+                    AppendObjectSpecState::Append(s) => {
+                        AppendObjectSpecState::Append(AppendObjectSpec {
+                            generation: new_generation.unwrap_or(s.generation),
+                            write_handle: m.write_handle.clone().or_else(|| s.write_handle.clone()),
+                            ..s.clone()
+                        })
+                    }
+                };
+                *guard = new_state;
```
coryan: You should refactor this to a member function like I outline for `handle_redirect()`

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,750 @@
...
+                problem,
+            };
+            let mismatch = PathMismatch {
+                subs: vec![mismatch],
+            };
+            let mismatch = BindingError {
+                paths: vec![mismatch],
+            };
+
+            return Err(crate::Error::binding(mismatch));
```
coryan: The `connect_attempt()` function is a bit long, consider refactoring things like this to helper functions.

coryan (changes requested):

src/storage/src/storage/bidi_write/redirect.rs
vsharonlynn: Thanks for the input. I changed it to `&mut` and made it a member function of `AppendObjectSpecState`.

Since `AppendObjectSpecState` struct is in `connector.rs` and the `handle_redirect` is here, I created a new file state.rs to put everything in the same file. This includes another comment related to `AppendObjectSpecState` [below](https://github.com/googleapis/google-cloud-rust/pull/5956#discussion_r3552143180).

Thank you!

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Yes. This is extracted into a member function of `AppendObjectSpecState` in state.rs (new file). Thanks!

src/storage/src/storage/bidi_write/redirect.rs
vsharonlynn: Yes. Done. Thank you.

src/storage/src/storage/bidi_write/redirect.rs
vsharonlynn: This looks way better. Thanks.

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: I added percent encoding. Thank you.

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Good idea. I moved this out.

And also moved the request preparation because this chunk is long, so moving it out to make this function simpler and easier to follow.

Thanks.

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Added a constant at the top of file.

Maybe after I can send a quick PR to make the 100 in `open_object` value a constant too :)

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Sure, done.
I was trying to make the scope of the `#[allow(dead_code)]` smaller/more specific but it became cluttered.
Thanks!

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Without `Box` it fails the `clippy::large_enum_variant` check.

`AppendObjectSpecState` can either be `Append` or `Write`.
`Write` contains `WriteObjectSpec`. `WriteObjectSpec` is large, contains `Option<Object>`. And `Object` is a large struct containing many fields.
In Rust, an enum takes a memory of the largest variant, so if `WriteObjectSpec` is not boxed, `AppendObjectSpecState` with `Append` variant will take up memory that fits `WriteObjectSpec` which is unnecessarily large.

Should I do this way with `Box`; or without `Box` but will need `#[allow(clippy::large_enum_variant)]` macro?

Thanks!

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,728 @@
...
+            Ok(Some(m)) => {
+                let mut guard = spec.lock().expect("never poisoned");
+                guard.handle_response(&m);
+
+                Ok((m, Connection::new(tx, stream)))
+            }
+            Ok(None) => Err(Error::io("bidi_write_object stream closed before start")),
+            Err(status) => {
+                let mut guard = spec.lock().expect("never poisoned");
+                Err(guard.handle_redirect(status))
```
coryan, line 224: optional: the code coverage says this case is not tested. Seems like an important behavior to test

coryan (approved):
LGTM.

For future reference, avoid PRs with 1,000 lines of code (such as this one). It makes it hard to review them thoroughly.  This may contradict what other reviewers want (a full picture of all the changes). Consider a draft PR with the full thing and smaller PRs that are part of the changes.

vsharonlynn:
> LGTM.
> 
> For future reference, avoid PRs with 1,000 lines of code (such as this one). It makes it hard to review them thoroughly. This may contradict what other reviewers want (a full picture of all the changes). Consider a draft PR with the full thing and smaller PRs that are part of the changes.

Noted. I'll do better in splitting the large PR into smaller PRs for the next one. Agree that it's difficult when it's this large of a PR. Thanks Carlos!

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Done. Thanks!

PR #5961, opened 2026-06-30
chore(actions): address zizmor findings

g-husam:
This PR is an auto-generated attempt to address zizmor findings. It may not catch everything, and should be reviewed by repository owners.

These changes were generated by running `zizmor --fix=all --gh-token=$(gh auth token) ./.github/workflows`, and then applying some fixes for any remaining issues reported by zizmor. See go/github-zizmor-help for instructions to install and run.

If this PR is unhelpful, feel free to close the PR and address separately. If it is helpful, feel free to approve and merge, or edit/modify as needed to get it to the right state. Repository owners must ultimately ensure compliance by 2026-07-13. The purpose of this PR is to provide some assistance with achieving that as a first pass. This will become a blocking check for new changes to github workflows on 2026-07-13 within the `googleapis` org.

There may be some ignored findings (with the comment `# zizmor: ignore[...]`), which you may fix if feasible.

### DIFF
--- a/.github/workflows/sdk.yaml
+++ b/.github/workflows/sdk.yaml
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 name: Rust SDK
-permissions: read-all
+permissions: read-all # zizmor: ignore[excessive-permissions]
 on:
   push:
     branches: [main, 'release-*']


.github/workflows/sdk.yaml
```diff
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 name: Rust SDK
-permissions: read-all
+permissions: read-all # zizmor: ignore[excessive-permissions]
```
dbolduc, line 16: Should we change the permissions instead?

g-husam: if possible to restrict, yeah that would be great! feel free to directly edit this PR. Do you know if permissions can be restricted?

dbolduc (approved):

.github/workflows/sdk.yaml
dbolduc: I am sure they can. I thought the tool might suggest a fix. If not, feel free to merge

g-husam: it doesn't seem to automate the fix, tho the docs suggest to almost always restrict the top level workflow permissions, and add the minimal necessary permissions to each job. 

https://docs.zizmor.sh/audits/#excessive-permissions

will merge for now

PR #5967, opened 2026-06-30
chore(spanner): add initial setup for built-in metrics

olavloite:
Sets up the initial boiler-plate code for adding built-in metrics for Spanner.

The feature is hidden behind a feature flag, meaning that it is currently invisible to customers.


src/spanner/src/observability/metrics.rs
```diff
@@ -0,0 +1,338 @@
...
+        };
+
+        let exporter = GcpMonitoringExporter::new(monitoring_client, project_id);
+
+        // Set up PeriodicReader
+        let reader = PeriodicReader::builder(exporter).build();
+
+        let meter_provider = SdkMeterProvider::builder().with_reader(reader).build();
+
+        let meter = meter_provider.meter("cloud.google.com/rust");
```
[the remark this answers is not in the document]
olavloite: We do not want to change this. We should follow the same naming scheme as the other Spanner clients.

src/spanner/src/observability/metrics.rs
```diff
@@ -0,0 +1,339 @@
...
+#[cfg(google_cloud_unstable_tracing)]
+fn result_to_status_str<T>(result: &crate::Result<T>) -> &'static str {
+    match result {
+        Ok(_) => "OK",
+        Err(e) => {
+            if let Some(status) = e.status() {
+                status.code.name()
+            } else {
+                "UNKNOWN"
+            }
```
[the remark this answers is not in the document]
olavloite: Nonsense

src/spanner/src/observability/metrics.rs
```diff
@@ -0,0 +1,339 @@
...
+
+        // Create the Google Cloud Monitoring client using the same config
+        let mut builder = MetricService::builder();
+
+        if let Some(ref cred) = config.cred {
+            builder = builder.with_credentials(cred.clone());
+        }
+        if let Some(ref ud) = config.universe_domain {
+            builder = builder.with_universe_domain(ud.clone());
+        }
```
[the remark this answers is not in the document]
olavloite: No, that would not be a good idea. It would mean that we would be trying to export metrics to Spanner or to the Spanner Emulator, both of which are not really useful for that....

src/spanner/src/observability/metrics.rs
```diff
@@ -0,0 +1,335 @@
...
+                    e
+                );
+                return Self::disabled();
+            }
+        };
+
+        let exporter = GcpMonitoringExporter::new(monitoring_client, project_id);
+
+        // Set up PeriodicReader
+        let reader = PeriodicReader::builder(exporter).build();
```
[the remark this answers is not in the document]
olavloite: Nonsense. If this was true, then the CI build would have failed.

src/spanner/Cargo.toml
```diff
@@ -31,8 +31,9 @@ default = ["default-rustls-provider"]
 # TLS and authentication. Applications with specific requirements for
 # cryptography (such as exclusively using the [ring] crate) should disable this
 # default and call `rustls::CryptoProvider::install_default()`.
-default-rustls-provider = ["gaxi/_default-rustls-provider"]
-unstable-stream         = ["dep:futures"]
+default-rustls-provider      = ["gaxi/_default-rustls-provider"]
+unstable-stream              = ["dep:futures"]
+experimental-builtin-metrics = ["dep:google-cloud-monitoring-v3", "dep:opentelemetry", "dep:opentelemetry_sdk"]
```
coryan: A thing we learnt the hard way: removing a public feature is a breaking change.   You may want to name this `_experimental-builtin-metrics` because that makes the feature private (by convention) and then removing it is fine.

src/spanner/Cargo.toml
```diff
@@ -60,6 +61,11 @@ tracing.workspace                      = true
 url.workspace                          = true
 wkt                                    = { workspace = true, features = ["time"] }
 
+# Optional dependencies for built-in metrics
+google-cloud-monitoring-v3 = { workspace = true, optional = true }
```
coryan, line 65: The new shinny is to use the OpenTelemetry Lightweight Protocol (OTLP). Google Cloud supports the protocol. There are Rust exporters for OTLP, they do lack a connector to inject the auth headers.

coryan (changes requested):

sakthivelmanii (approved):

src/spanner/Cargo.toml
olavloite: Changed the name to `_experimental-builtin-metrics` to make it private.

src/spanner/Cargo.toml
olavloite: We would prefer not to use the new shiny library here, as this is (partly) for internal metrics for Spanner, and we need some additional control over how the time series are being created. For that, we need the specific Google Cloud Monitoring client.

coryan: Ack. Your call. I hear that other Spanner teams are trying to use OTLP, but I don't have a dog in this fight.

src/spanner/Cargo.toml
coryan: Thanks.

coryan (approved):

PR #5992, opened 2026-07-03
chore: update Rust to 1.96.1

coryan:
This version includes fixes to retry HTTP requests, probably the root cause of the numerous flakes during the CI downloads.


coryan:
We may not need this, "1.96" means "latest version in the 1.96 series" for Docker images, which is what we use to pull the Rust toolchain.

Issue #5991, opened 2026-07-03
FR: Support the use of gRPC-Rust as a transport layer for Storage


joshuatants:
**Is your feature request related to a problem? Please describe.**

The official Google-maintained Rust implementation of gRPC (`gRPC-Rust`) has been released: https://crates.io/crates/grpc . We'd like to see what happens if Storage uses it.

**Describe the solution you'd like**
Storage control and data plane operations have an option to use `gRPC-Rust` instead of `tonic` as the transport.

**Additional context**

* b/526402680

PR #5996, opened 2026-07-07
refactor(storage): define the GrpcClient and GrpcStream aliases

joshuatants:
For #5991

### DIFF
--- a/src/storage/src/storage/bidi/connector.rs
+++ b/src/storage/src/storage/bidi/connector.rs
@@ -14,7 +14,7 @@
 
 use super::redirect::handle_redirect;
 use super::retry_redirect::RetryRedirect;
-use super::{Client, TonicStreaming};
+use super::{Client, GrpcClient, GrpcStream, TonicStreaming};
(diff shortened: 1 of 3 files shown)


src/storage/src/storage/bidi/connector.rs
```diff
@@ -23,15 +23,15 @@ use crate::request_options::RequestOptions;
 use crate::storage::bidi::resume_redirect::ResumeRedirect;
 use crate::storage::info::X_GOOG_API_CLIENT_HEADER;
 use crate::{Error, Result};
-use gaxi::grpc::Client as GrpcClient;
-use gaxi::grpc::tonic::{Extensions, GrpcMethod, Status, Streaming};
+
```
dbolduc: nit: remove the newline

dbolduc (approved):

src/storage/src/storage/bidi/connector.rs
joshuatants: Done

PR #5999, opened 2026-07-08
feat(gax): introduce GrpcRustClient skeleton

joshuatants:
For #5991.

Introducing a skeleton unblocks working on the changes needed for Storage for both the hand-written (https://github.com/googleapis/google-cloud-rust/pull/6000) and GAPIC (changes to be made to the relevant templates in `librarian`) clients.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -550,6 +550,7 @@ storage-grpc-mock         = { path = "src/storage/grpc-mock" }
 
 [workspace.lints.rust]
 unexpected_cfgs = { level = "deny", check-cfg = [
+  'cfg(google_cloud_unstable_grpc_rust)',
   'cfg(google_cloud_unstable_grpc_server_streaming)',
   'cfg(google_cloud_unstable_storage_bidi)',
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -549,6 +549,7 @@ storage-grpc-mock         = { path = "src/storage/grpc-mock" }
 
 [workspace.lints.rust]
 unexpected_cfgs = { level = "deny", check-cfg = [
+  'cfg(google_cloud_unstable_grpc_rust)',
```
coryan, line 553: I think you want to add this to:

https://github.com/googleapis/google-cloud-rust/blob/3774eba33c9c6cf017b3939eac18a41c3d7329bc/.gcb/builds/triggers/main.tf#L37-L42

So the new flag is tested.

coryan (changes requested):

Cargo.toml
joshuatants: Done.

coryan (approved):

PR #6001, opened 2026-07-08
feat(storage): implement appendable upload background worker

vsharonlynn:
This PR comes after #5956.

### DIFF
--- a/src/storage/src/storage/bidi_write/worker.rs
+++ b/src/storage/src/storage/bidi_write/worker.rs
@@ -19,23 +19,20 @@ use super::{Client, TonicStreaming};
 use crate::error::WriteError;
 use crate::google::storage::v2::{BidiWriteObjectRequest, BidiWriteObjectResponse};
 use gaxi::grpc::tonic::Result as TonicResult;
-use std::sync::Arc;
+
 use tokio::sync::mpsc::Receiver;
 use tokio::sync::oneshot;
 
 type WriteResult<T> = std::result::Result<T, WriteError>;
-type LoopResult<T> = std::result::Result<T, Arc<crate::Error>>;
+type LoopResult<T> = std::result::Result<T, crate::Error>;
 
-/// The intent sent from the user to the background worker.
+/// The intent sent from the foreground task to the background worker.
 pub enum UploadIntent {
-    /// Append a chunk of data.
     Append(BidiWriteObjectRequest),
-    /// Flush the data to the server.
     Flush(
         BidiWriteObjectRequest,
         oneshot::Sender<crate::Result<BidiWriteObjectResponse>>,
     ),
-    /// Finalize the upload.
     Finalize(
         BidiWriteObjectRequest,
         oneshot::Sender<crate::Result<BidiWriteObjectResponse>>,
@@ -47,6 +44,8 @@ pub struct Worker<C> {
     _connector: super::connector::Connector<C>,
     pending_flushes:
         std::collections::VecDeque<oneshot::Sender<crate::Result<BidiWriteObjectResponse>>>,
+    /// Tracks if the client intends to complete the upload, by sending a Finalize intent.
+    finalized: bool,
 }
 
 impl<C> Worker<C> {
@@ -54,6 +53,7 @@ impl<C> Worker<C> {
         Self {
             _connector: connector,
             pending_flushes: std::collections::VecDeque::new(),
+            finalized: false,
         }
     }
 }
@@ -76,12 +76,13 @@ where
                     match self.handle_response(m) {
                         // Successful end of stream, return without error.
                         None => break None,
-                        // An unrecoverable in the stream or its data, return
+                        // An unrecoverable error in the stream or its data, return
                         // the error.
                         Some(Err(e)) => break Some(e),
                         // New message on the stream handled successfully,
                         // continue.
                         Some(Ok(None)) => {},
+                        // TODO(#5716): Update when implementing reconnect logic.
                         // The stream reconnected successfully, update the local
                         // variables and continue.
                         Some(Ok(Some(connection))) => {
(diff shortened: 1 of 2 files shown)


src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,745 @@
...
+                            object: s
+                                .resource
+                                .as_ref()
+                                .map(|r| r.name.clone())
+                                .unwrap_or_default(),
+                            generation: new_generation.unwrap_or(0),
+                            write_handle: m.write_handle.clone(),
+                            ..Default::default()
+                        })
+                    }
```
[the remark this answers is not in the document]
vsharonlynn: Fixed in #5956.

src/storage/src/storage/bidi_write/worker.rs
```diff
@@ -0,0 +1,252 @@
...
+        let response = match message {
+            Ok(Some(msg)) => msg,
+            Ok(None) => {
+                if !self.pending_flushes.is_empty() {
+                    return Some(Err(Arc::new(crate::Error::io(
+                        "stream closed unexpectedly",
+                    ))));
+                }
+                return None;
+            }
```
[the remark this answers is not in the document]
vsharonlynn: Fixed.

src/storage/src/storage/bidi_write/worker.rs
```diff
@@ -0,0 +1,333 @@
...
+                                    self.pending_flushes.push_back(sender);
+                                    self.finalized = true;
+                                    req
+                                }
+                            };
+                            if let Err(e) = tx.send(request).await {
+                                break Some(crate::Error::io(e));
+                            }
+                        }
+                        None => {
```
joshuatants, line 103: IIUC this path is taken when the foreground writer drop its side of the channel? Reading https://github.com/googleapis/googleapis/blob/300fb478d0334baeefd134eba6ea30277477509b/google/storage/v2/storage.proto#L528 , it looks like the server will send a terminal `BidiWriteObjectResponse` afterward, and there is also the final status of the gRPC stream. However, looking at the code here we don't check for that.

src/storage/src/storage/bidi_write/worker.rs
```diff
@@ -0,0 +1,333 @@
...
+                            (rx, tx) = (connection.rx, connection.tx);
+                        }
+                    }
+                },
+                intent = requests.recv() => {
+                    match intent {
+                        Some(intent) => {
+                            let request = match intent {
+                                UploadIntent::Append(req) => req,
+                                UploadIntent::Flush(req, sender) => {
```
joshuatants: From https://github.com/googleapis/googleapis/blob/300fb478d0334baeefd134eba6ea30277477509b/google/storage/v2/storage.proto#L1554-L1561 , IIUC the server will only ack a flush if state_lookup is true. If false, it'll never ack it. 

If it's false, then should the pending_flushes still be updated?

src/storage/src/storage/bidi_write/worker.rs
```diff
@@ -0,0 +1,333 @@
...
+                        Some(Ok(None)) => {},
+                        // TODO(#5716): Update when implementing reconnect logic.
+                        // The stream reconnected successfully, update the local
+                        // variables and continue.
+                        Some(Ok(Some(connection))) => {
+                            (rx, tx) = (connection.rx, connection.tx);
+                        }
+                    }
+                },
+                intent = requests.recv() => {
```
joshuatants, line 95: On error, does this need to be drained similary to how pending_flushes is drain in L121/

joshuatants (changes requested):

src/storage/src/storage/bidi_write/worker.rs
vsharonlynn: Yes, thanks for catching that. I've updated the None arm to `drop(tx)` and call `wait_for_server_completion` (new function).

src/storage/src/storage/bidi_write/worker.rs
vsharonlynn: Good point.
Right now I force `state_lookup = true` because the SDK's flush API should return a persisted_size.

Another option is to return error when it's false, and this make sure transport (future PR) calls with state_lookup = true.
Wdyt?

Thanks.

src/storage/src/storage/bidi_write/worker.rs
vsharonlynn: Yes, thanks for catching that, too. I've created a function `drain_intents_on_error` to drain both `pending_flushes` for in-flight requrests, and drain requests queued in channel.

src/storage/src/storage/bidi_write/worker.rs
joshuatants: > Right now I force state_lookup = true because the SDK's flush API should return a persisted_size.

If the API retuns a persisted_size then yes we should do  state_lookup = true and report that.

> Another option is to return error when it's false, and this make sure transport (future PR) calls with state_lookup = true.

I don't understand what you mean.

joshuatants: Also, I realised that both the Flush and Finalize branches here don't set flush or finish_write (https://github.com/googleapis/googleapis/blob/69d78666e1c3b7c3b4f9d229550754c148638f98/google/storage/v2/storage.proto#L1564-L1578 ). I think they should be set? If the intent is to have the foreground writer set these, then the state_lookup = true should also be moved there.

joshuatants (changes requested):

src/storage/src/storage/bidi_write/worker.rs
vsharonlynn: The BidiWriteObjectRequest will be constructed in transport.rs ([next PR](https://github.com/googleapis/google-cloud-rust/pull/6002/changes#diff-65ffebcec58e4f7067de9a62568d6aa38fe10e766fd2adbee69cbd22ae607c0bR138)).

My dilemma is whether worker.rs here should override `state_lookup`, or let transport.rs be the sole constructor for the values in the request.

The other option is worker.rs here will check if `state_lookup == false`:
a. throw an error instead of overriding the `state_lookup` value with true
b. not enqueue to `pending_flushes`

vsharonlynn: > Also, I realised that both the Flush and Finalize branches here don't set flush or finish_write (https://github.com/googleapis/googleapis/blob/69d78666e1c3b7c3b4f9d229550754c148638f98/google/storage/v2/storage.proto#L1564-L1578 ). I think they should be set? If the intent is to have the foreground writer set these, then the state_lookup = true should also be moved there.

Similarly, this will be done in transport.rs in the [next PR](https://github.com/googleapis/google-cloud-rust/pull/6002/changes#diff-65ffebcec58e4f7067de9a62568d6aa38fe10e766fd2adbee69cbd22ae607c0bR163).

joshuatants: Thanks for clarifying.

> The other option is worker.rs here will check if state_lookup == false:
a. throw an error instead of overriding the state_lookup value with true
b. not enqueue to pending_flushes

If flush() expects to return the persisted size, I think a. is the better path, though we should assert()! and not throw an error because not setting state_lookup is our (SDK developer's) fault. If you throw an error here, what can the user do?

joshuatants: > > Also, I realised that both the Flush and Finalize branches here don't set flush or finish_write (https://github.com/googleapis/googleapis/blob/69d78666e1c3b7c3b4f9d229550754c148638f98/google/storage/v2/storage.proto#L1564-L1578 ). I think they should be set? If the intent is to have the foreground writer set these, then the state_lookup = true should also be moved there.
> 
> Similarly, this will be done in transport.rs in the [next PR](https://github.com/googleapis/google-cloud-rust/pull/6002/changes#diff-65ffebcec58e4f7067de9a62568d6aa38fe10e766fd2adbee69cbd22ae607c0bR163).

Ack. Let's set everything in transport.rs then, though we can assert()s here to make sure we don't pass around invalid state.

joshuatants (approved):

src/storage/src/storage/bidi_write/worker.rs
vsharonlynn: Agree. Added asserts. Thanks again!

vsharonlynn:
I'll merge this first to meet deadlines. Please feel free to leave feedback. I will incorporate them in a separate PR. Thank you.

PR #6002, opened 2026-07-08
feat(storage): implement appendable upload transport

vsharonlynn:
This PR comes after #6001.

### DIFF
--- a/src/storage/src/storage/bidi_write/transport.rs
+++ b/src/storage/src/storage/bidi_write/transport.rs
@@ -114,6 +114,7 @@ impl AppendableObjectWriterTransport {
         }
         // If persisted_size > 0 but the server didn't provide a checksum,
         // we can't reliably continue a running checksum, so it remains `None`.
+        // TODO(#5716): Check whether this is a valid case.
 
         let (tx, rx) = tokio::sync::mpsc::channel(100);
         let worker = Worker::new(connector);
--- a/src/storage/src/storage/bidi_write/transport.rs
+++ b/src/storage/src/storage/bidi_write/transport.rs
@@ -42,19 +42,6 @@ pub struct AppendableObjectWriterTransport {
 }
 
 impl AppendableObjectWriterTransport {
-    async fn extract_worker_error(&mut self, default_err_message: &str) -> Error {
-        if let Some(handle) = self.worker_handle.take() {
-            match handle.await {
-                Ok(Err(worker_err)) => return worker_err,
-                Ok(Ok(())) => {
-                    return Error::io("worker terminated successfully but channel was closed");
-                }
-                Err(join_err) => return Error::io(format!("worker task error: {join_err}")),
-            }
-        }
-        Error::io(default_err_message)
-    }
-
     pub async fn new_open<T>(
         mut connector: Connector<T>,
         req: OpenAppendableObjectRequest,
@@ -84,30 +71,49 @@ impl AppendableObjectWriterTransport {
         connector: Connector<T>,
         initial: BidiWriteObjectResponse,
         connection: Connection<<T as Client>::Stream>,
-        mut generation: i64,
+        generation: i64,
     ) -> Result<Self>
     where
         T: Client + Clone + Sync + Send + 'static,
         <T as Client>::Stream: TonicStreaming + Send + Sync,
     {
         let mut persisted_size = 0;
+        let mut generation = generation;
+
+        // The GCS backend returns `WriteStatus::Resource` in two scenarios:
+        // 1. Immediately upon creating a new appendable stream, where `finalize_time` is absent and the new `generation` is returned.
+        // 2. When the stream is fully finalized, where `finalize_time` is present.
+        // Otherwise, such as on stream reopens, the backend returns `WriteStatus::PersistedSize`.
         if let Some(WriteStatus::Resource(r)) = initial.write_status.as_ref() {
-            generation = r.generation;
+            if r.finalize_time.is_some() {
+                return Err(Error::io("object is already finalized"));
+            }
             persisted_size = r.size;
+            generation = r.generation;
         } else if let Some(WriteStatus::PersistedSize(s)) = initial.write_status.as_ref() {
             persisted_size = *s;
         }
 
+        // Determine whether we should do a full-object CRC32C checksum
+        // calculation. Default is `None`. We will then see if we can establish
+        // a valid CRC32C baseline.
         let mut running_crc32c = None;
         if persisted_size == 0 {
+            // A brand new object or takeover an existing object with 0 bytes written,
+            // so we start with a checksum of 0.
             running_crc32c = Some(0);
         } else if let Some(crc) = initial
             .persisted_data_checksums
             .as_ref()
             .and_then(|c| c.crc32c)
         {
+            // Takeover an existing object where the server returns the checksum
+            // of the persisted data. The running CRC32C checksum will start with
+            // this value.
             running_crc32c = Some(crc);
         }
+        // If persisted_size > 0 but the server didn't provide a checksum,
+        // we can't reliably continue a running checksum, so it remains `None`.
 
         let (tx, rx) = tokio::sync::mpsc::channel(100);
         let worker = Worker::new(connector);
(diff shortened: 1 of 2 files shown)


src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -0,0 +1,745 @@
...
+                                .unwrap_or_default(),
+                            object: s
+                                .resource
+                                .as_ref()
+                                .map(|r| r.name.clone())
+                                .unwrap_or_default(),
+                            generation: new_generation.unwrap_or(0),
+                            write_handle: m.write_handle.clone(),
+                            ..Default::default()
+                        })
```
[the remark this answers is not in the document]
vsharonlynn: Resolved in the previous PR #5956 .

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -0,0 +1,796 @@
...
+        let mut running_crc32c = None;
+        if persisted_size == 0 {
+            running_crc32c = Some(0);
+        } else if let Some(crc) = initial
+            .persisted_data_checksums
+            .as_ref()
+            .and_then(|c| c.crc32c)
+        {
+            running_crc32c = Some(crc);
+        }
```
joshuatants, line 114: can we add commenst here to explain the logic? makes it more readable

this also applies throughout the code

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -0,0 +1,796 @@
...
+            generation = r.generation;
+            persisted_size = r.size;
+        } else if let Some(WriteStatus::PersistedSize(s)) = initial.write_status.as_ref() {
+            persisted_size = *s;
+        }
+
+        let mut running_crc32c = None;
+        if persisted_size == 0 {
+            running_crc32c = Some(0);
+        } else if let Some(crc) = initial
```
joshuatants, line 105: just to make sure: if the server does not return a persisted_data_checksums, will checksumming be turned off entirely? if so, is that what we want (i think we'd still want to calculate chunk-level checksums)?

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -0,0 +1,796 @@
...
+        };
+        self.persisted_size = size;
+        Ok(size)
+    }
+
+    async fn finalize(mut self) -> Result<crate::model::Object> {
+        let (sender, receiver) = oneshot::channel();
+        let object_checksums = self.running_crc32c.map(|crc| ObjectChecksums {
+            crc32c: Some(crc),
+            md5_hash: vec![].into(),
```
joshuatants: can ` bytes::Bytes::new()` be used instead?

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -0,0 +1,796 @@
...
+        let size = match response.write_status {
+            Some(WriteStatus::PersistedSize(s)) => s,
+            Some(WriteStatus::Resource(r)) => r.size,
+            None => return Err(Error::io("flush response missing write_status")),
+        };
+        self.persisted_size = size;
+        Ok(size)
+    }
+
+    async fn finalize(mut self) -> Result<crate::model::Object> {
```
joshuatants, line 225: does finalize() also imply close()? 

if so, then similar logic to close() should be followed, i.e. dropping the transport and joining on the worker

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -0,0 +1,796 @@
...
+        let (initial, connection) = connector.connect_reopen(req).await?;
+        Self::start_worker(connector, initial, connection, generation)
+    }
+
+    fn start_worker<T>(
+        connector: Connector<T>,
+        initial: BidiWriteObjectResponse,
+        connection: Connection<<T as Client>::Stream>,
+        mut generation: i64,
+    ) -> Result<Self>
```
joshuatants, line 75: Why Result<Self>? Is it possible for the logic in this function to actually error?

joshuatants (changes requested):

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Yes, that's better. Thank you.

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Done. Thank you.

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Yes, done. Thanks.

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Only the running checksum will be `None`.

Append will still send chunk checksum.
```
async fn append(&mut self, chunk: Bytes) -> Result<()> {
        let length = chunk.len() as i64;
        let crc32c = crc32c::crc32c(&chunk);

        let new_running_crc32c = self
            .running_crc32c
            .map(|running| crc32c::crc32c_combine(running, crc32c, chunk.len()));

        let request = BidiWriteObjectRequest {
            write_offset: self.write_offset,
            data: Some(Data::ChecksummedData(ChecksummedData {
                content: chunk,
                crc32c: Some(crc32c),
            })),
            ..BidiWriteObjectRequest::default()
        };
        ...
}
```

I've added some comments. PTAL and see whether this makes sense. Thank you.

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Thanks for pointing this out. I made a mistake in the `start_worker` logic. When `Resource` is not empty, the object is already finalized and reopening should return an error.

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -0,0 +1,921 @@
...
+            .persisted_data_checksums
+            .as_ref()
+            .and_then(|c| c.crc32c)
+        {
+            // Takeover an existing object where the server returns the checksum
+            // of the persisted data. The running CRC32C checksum will start with
+            // this value.
+            running_crc32c = Some(crc);
+        }
+        // If persisted_size > 0 but the server didn't provide a checksum,
```
joshuatants, line 115: For my learning, is this a valid state? Can the server not provide a checksum?

src/storage/src/storage/bidi_write/transport.rs
joshuatants: Makes sense, but I thought running checksum is for object-level checksumming. What happens to the chunk-level checksum?

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -0,0 +1,921 @@
...
+        }
+
+        self.write_offset += length;
+        self.running_crc32c = new_running_crc32c;
+
+        Ok(())
+    }
+
+    async fn flush(&mut self) -> Result<i64> {
+        let (sender, receiver) = oneshot::channel();
```
joshuatants, line 199: Doesn't have to be addressed now: is it feasible to create a permenant tx/rx pair for responses from the worker? In both flush and close we create a oneshot channel every time, wondering if there might be some performance improvement if use a reusable channel.

joshuatants (approved):

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Good question... Some points that might benefit from creating `oneshot::channel` every time:
* cancellation safety. Say the user drops a flush().await future, a reusable channel would still receive obsolete background response.
* cleanup. Dropping the future drops the receiver, too, so stale stream response is cleanly discarded.

I'll list this in a list of possible performance improvements.
Thanks.

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: As discussed offline, adding TODO here to unblock the other PRs first. Thank you.

PR #6015, opened 2026-07-09
impl(bigquery): add row iterator

alvarowolfx:
Towards https://github.com/googleapis/google-cloud-rust/issues/5844

### DIFF
--- a/src/bigquery/src/query/iterator.rs
+++ b/src/bigquery/src/query/iterator.rs
@@ -59,7 +59,7 @@ impl RowIterator {
             return None;
         }
 
-        if let Some(token) = self.page_token.take() {
+        while let Some(token) = self.page_token.take() {
             match self.fetch_page(&token).await {
                 Ok((fetched_rows, next_token)) => {
                     self.page_token = next_token;
--- a/src/bigquery/src/query/iterator.rs
+++ b/src/bigquery/src/query/iterator.rs
@@ -28,21 +28,16 @@ pub struct RowIterator {
     schema: Arc<Schema>,
     page_token: Option<String>,
     rows: VecDeque<wkt::Struct>,
-    is_done: bool,
 }
 
 impl RowIterator {
     pub(crate) fn new(q: CompleteQuery) -> Self {
-        let rows = q.cached_rows;
-        let is_done = rows.is_empty() && q.page_token.is_none();
-
         Self {
             job_service: q.job_service,
             job_ref: q.job_ref,
             schema: q.schema,
             page_token: q.page_token,
-            rows,
-            is_done,
+            rows: q.cached_rows,
         }
     }
 
@@ -50,33 +45,30 @@ impl RowIterator {
     ///
     /// Returns `None` when all rows have been retrieved.
     pub async fn next(&mut self) -> Option<Result<Row>> {
-        if let Some(raw_row) = self.rows.pop_front() {
-            let row = Row::try_new(raw_row, &self.schema);
-            return Some(row);
-        }
+        loop {
+            if let Some(raw_row) = self.rows.pop_front() {
+                return Some(Row::try_new(raw_row, &self.schema));
+            }
 
-        if self.is_done {
-            return None;
-        }
+            if let Err(e) = self.try_fetch_page().await {
+                return Some(Err(e));
+            }
 
-        while let Some(token) = self.page_token.take() {
-            match self.fetch_page(&token).await {
-                Ok((fetched_rows, next_token)) => {
-                    self.page_token = next_token;
-                    self.rows.extend(fetched_rows);
-                    if let Some(raw_row) = self.rows.pop_front() {
-                        return Some(Row::try_new(raw_row, &self.schema));
-                    }
-                }
-                Err(e) => {
-                    self.page_token = Some(token);
-                    return Some(Err(e));
-                }
+            if self.rows.is_empty() && self.page_token.is_none() {
+                return None;
             }
         }
+    }
 
-        self.is_done = true;
-        None
+    async fn try_fetch_page(&mut self) -> Result<()> {
+        let Some(token) = &self.page_token.as_deref() else {
+            return Ok(());
+        };
+
+        let (fetched_rows, next_token) = self.fetch_page(token).await?;
+        self.page_token = next_token;
+        self.rows.extend(fetched_rows);
+        Ok(())
     }
 
     // Fetches the next page of results and the next page token.
@@ -140,7 +132,6 @@ mod tests {
         let q = create_test_complete_query(job_service, Some(create_test_job_ref()), vec![], None);
         let mut iter = q.read();
         assert!(iter.next().await.is_none(), "{iter:?}");
-        assert!(iter.is_done, "{iter:?}");
         Ok(())
     }
 
@@ -158,7 +149,6 @@ mod tests {
         assert_eq!(row2.get::<String, _>("col"), "second");
 
         assert!(iter.next().await.is_none(), "{iter:?}");
-        assert!(iter.is_done, "{iter:?}");
         Ok(())
     }
 
--- a/src/bigquery/src/query/iterator.rs
+++ b/src/bigquery/src/query/iterator.rs
@@ -1,0 +1,182 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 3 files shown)


src/bigquery/src/query/iterator.rs
```diff
@@ -0,0 +1,182 @@
...
+use crate::query::{CompleteQuery, Row, Schema};
+use google_cloud_bigquery_v2::client::JobService;
+use google_cloud_bigquery_v2::model::JobReference;
+use std::collections::VecDeque;
+use std::sync::Arc;
+
+pub type Result<T> = std::result::Result<T, RowError>;
+/// An iterator over rows returned by a query.
+#[derive(Debug)]
+pub struct RowIterator {
```
dbolduc, line 25: nit: this thing streams rows, so call it: `RowStream` ?

src/bigquery/src/query/iterator.rs
```diff
@@ -0,0 +1,182 @@
...
+    pub async fn next(&mut self) -> Option<Result<Row>> {
+        if let Some(raw_row) = self.rows.pop_front() {
+            let row = Row::try_new(raw_row, &self.schema);
+            return Some(row);
+        }
+
+        if self.is_done {
+            return None;
+        }
+
```
dbolduc: remove `self.is_done`. It is equivalent to `self.page_token.is_empty()` which is already checked in the while condition.

src/bigquery/src/query/iterator.rs
```diff
@@ -0,0 +1,182 @@
...
+            match self.fetch_page(&token).await {
+                Ok((fetched_rows, next_token)) => {
+                    self.page_token = next_token;
+                    self.rows.extend(fetched_rows);
+                    if let Some(raw_row) = self.rows.pop_front() {
+                        return Some(Row::try_new(raw_row, &self.schema));
+                    }
+                }
+                Err(e) => {
+                    self.page_token = Some(token);
```
dbolduc: If there is an error, why are we setting the page token?

Typically if we yield an error, it means the next item on the stream should be `None`.

At what level do you want to retry transient errors? I assumed that would happen inside `fetch_page(...)`

src/bigquery/src/query/iterator.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+
+        while let Some(token) = self.page_token.take() {
+            match self.fetch_page(&token).await {
+                Ok((fetched_rows, next_token)) => {
+                    self.page_token = next_token;
+                    self.rows.extend(fetched_rows);
+                    if let Some(raw_row) = self.rows.pop_front() {
+                        return Some(Row::try_new(raw_row, &self.schema));
+                    }
```
dbolduc: comment: There is a way to not repeat ourselves. If we set the code up like:

```rs
pub async fn next(&mut self) -> Option<Result<Row>> {
  loop {
    if let Some(r) = self.rows.pop_front() {
      return Some(Row::try_new(r, &self.schema));
    }
    if let Err(e) = self.try_fetch_rows().await {
      return Some(e);
    }
  }
}

async fn try_fetch_rows(&mut self) -> Result<()> {
  let token = self.page_token.take();
  let (fetched_rows, next_token) = self.fetch_page(token).await?;
  self.page_token = next_token;
  self.rows.extend(fetched_rows);
  Ok(())
}
```

Your call on what is more readable. [I used this pattern in pubsub](
https://github.com/googleapis/google-cloud-rust/blob/34f24e196bb0370d3204ab044233e7a0207d09a3/src/pubsub/src/subscriber/message_stream.rs#L276-L303).

src/bigquery/src/query/iterator.rs
```diff
@@ -0,0 +1,182 @@
...
+        }
+    }
+
+    /// Fetches the next row from the result set.
+    ///
+    /// Returns `None` when all rows have been retrieved.
+    pub async fn next(&mut self) -> Option<Result<Row>> {
+        if let Some(raw_row) = self.rows.pop_front() {
+            let row = Row::try_new(raw_row, &self.schema);
+            return Some(row);
```
dbolduc: super nit: we `return Some(Row::try_new(raw_row, &self.schema));` below but here we write the same thing in two lines.

src/bigquery/src/query/iterator.rs
```diff
@@ -0,0 +1,182 @@
...
+
+    #[tokio::test]
+    async fn test_row_iterator_row_conversion_error() -> TestResult {
+        let job_service = create_job_service(MockJobService::new());
+        let invalid_row = Map::from_iter([("f".to_string(), json!([]))]);
+        let q = create_test_complete_query(
+            job_service,
+            Some(create_test_job_ref()),
+            vec![invalid_row],
+            None,
```
dbolduc, line 163: Can we make this a `Some(_)`?

Then the expectation that we yield `None` after an error would fail.

src/bigquery/src/query/query_handle.rs
```diff
@@ -456,4 +461,37 @@ mod tests {
...
     }
+
+    #[tokio::test]
+    async fn test_complete_query_read() -> TestResult {
+        let job_service = create_job_service(MockJobService::new());
+        let job_ref = JobReference::new()
+            .set_project_id("some_project")
+            .set_job_id("some_job_id");
+        let schema = TableSchema::new().set_fields([
+            google_cloud_bigquery_v2::model::TableFieldSchema::new()
```
dbolduc: nit: `use google_cloud_bigquery_v2::model::TableFieldSchema;`

dbolduc (commented):
Consider calling it a `RowStream`.

And I had questions about where transient errors are retried. I think that will impact the shape of the code.

src/bigquery/src/query/iterator.rs
alvarowolfx: page token is being set back again because `self.page_token.take()` was used, which sets `self.page_token` to None. 

But yeah, the plan was to retry on `fetch_page`

src/bigquery/src/query/iterator.rs
alvarowolfx: I'm not super opinionated on this, but I prefer RowIterator. And for consistency, Spanner also calls it RowIterator

src/bigquery/src/query/iterator.rs
dbolduc: Why should we continue fetching rows after yielding an error to the application?

alvarowolfx: we can surface the error that retries were exhausted (maybe by transient errors) to fetch a given row/page, but the RowIterator can still be used and users can call `next` again to retry, so it needs to be on a good state.

dbolduc: > RowIterator can still be used and users can call `next` again to retry

I hear you that this is possible, I just don't think we should allow for it.

I think we should encourage them to set a better retry policy where exhausted actually means exhausted.

src/bigquery/src/query/iterator.rs
alvarowolfx: adding a `Some(_)` will make it try to fetch a new page, which is going to panic for now.

src/bigquery/src/query/iterator.rs
```diff
@@ -28,55 +28,47 @@ pub struct RowIterator {
...
-                        return Some(Row::try_new(raw_row, &self.schema));
-                    }
-                }
-                Err(e) => {
-                    self.page_token = Some(token);
-                    return Some(Err(e));
-                }
+            if self.rows.is_empty() && self.page_token.is_none() {
+                return None;
             }
```
dbolduc, line 59: nit: if you move this above `try_fetch_page()` it only needs to be `if self.page_token.is_none()`

---

Alternatively, if `try_fetch_page(...)` returns `Option<Result<()>>`, this if statement goes away... but then we can't use `?` with `self.fetch_page(token).await?`, so meh, not sure it helps.

src/bigquery/src/query/iterator.rs
dbolduc: Right. This is the same point about not continuing after errors where we just disagree on what the behavior should be.

src/bigquery/src/query/iterator.rs
dbolduc: @coryan - do you want to break the tie here?

dbolduc (approved):

src/bigquery/src/query/iterator.rs
alvarowolfx: merging this and will discuss more on next PR where the actual logic to fetch pages and retry will be added

coryan: > @coryan - do you want to break the tie here?

> we can surface the error that retries were exhausted (maybe by transient errors) to fetch a given row/page, but the RowIterator can still be used and users can call next again to retry, so it needs to be on a good state.

Uh? If you did not like the outcome of the retry policy, then adjust the policy... don't build a second retry loop around it.

PR #6028, opened 2026-07-10
feat(oracledatabase)!: bump major version

codyoss:
There was recently a change in the protos for this library which remove an RPC and messages:
https://github.com/googleapis/googleapis/commit/6b0172376aa81a3c584d5fca90dbfce3aec84d16

### DIFF
--- a/src/generated/cloud/oracledatabase/v1/src/stub/dynamic.rs
+++ b/src/generated/cloud/oracledatabase/v1/src/stub/dynamic.rs
@@ -531,7 +531,6 @@ pub trait OracleDatabase: std::fmt::Debug + Send + Sync {
         options: &crate::RequestOptions,
     ) -> std::sync::Arc<dyn google_cloud_gax::polling_backoff_policy::PollingBackoffPolicy>;
 
-    #[cfg(google_cloud_unstable_tracing)]
     fn get_poller_options(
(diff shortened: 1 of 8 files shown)


src/generated/cloud/oracledatabase/v1/src/stub/dynamic.rs
```diff
@@ -531,6 +531,7 @@ pub trait OracleDatabase: std::fmt::Debug + Send + Sync {
         options: &crate::RequestOptions,
     ) -> std::sync::Arc<dyn google_cloud_gax::polling_backoff_policy::PollingBackoffPolicy>;
 
+    #[cfg(google_cloud_unstable_tracing)]
```
coryan: This is unexpected, probably a librarian version mismatch somewhere.

coryan (changes requested):
We need to figure out the tracing changes.  I expect CI will fail until we do.

src/generated/cloud/oracledatabase/v1/src/stub/dynamic.rs
codyoss: The version was updated overnight. I pushed a new commit to generate with the updated version.

coryan (approved):

PR #6032, opened 2026-07-10
impl(bigquery): fetch more pages and limit buffered results

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery/src/query/iterator.rs
+++ b/src/bigquery/src/query/iterator.rs
@@ -15,7 +15,7 @@
 use crate::error::RowError;
 use crate::query::{CompleteQuery, Row, Schema};
 use google_cloud_bigquery_v2::client::JobService;
-use google_cloud_bigquery_v2::model::JobReference;
+use google_cloud_bigquery_v2::model::{GetQueryResultsRequest, JobReference};
 use std::collections::VecDeque;
 use std::sync::Arc;
 
@@ -28,6 +28,7 @@ pub struct RowIterator {
     schema: Arc<Schema>,
     page_token: Option<String>,
     rows: VecDeque<wkt::Struct>,
+    max_rows_buffered: Option<u32>,
 }
 
 impl RowIterator {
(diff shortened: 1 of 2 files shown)


src/bigquery/src/query/iterator.rs
```diff
@@ -72,9 +80,38 @@ impl RowIterator {
     }
 
     // Fetches the next page of results and the next page token.
-    async fn fetch_page(&self, _token: &str) -> Result<(Vec<wkt::Struct>, Option<String>)> {
-        // TODO(#5592): implement page fetching with jobs.getQueryResults
-        unimplemented!("pagination support not yet implemented")
+    async fn fetch_page(&self, token: &str) -> Result<(Vec<wkt::Struct>, Option<String>)> {
+        let job_ref = self.job_ref.as_ref().expect(
+            "only queries with a job reference should have page tokens and can fetch more pages",
+        );
```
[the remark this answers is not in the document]
alvarowolfx: this is an unreachable state. Only happens if the backend runs a query as a stateless job (no job reference) and provide a page token (which is not possible)

tests/bigquery/src/lib.rs
```diff
@@ -266,7 +269,13 @@ pub async fn query_client_multi_page() -> Result<()> {
 
     assert_eq!(complete_query.metadata().total_rows, Some(10000));
 
-    // TODO(#5592): iterate on rows with max_results = 1000
+    let mut iter = complete_query.read().set_max_rows_buffered(1000);
+    let mut count = 0;
+    while let Some(row) = iter.next().await {
+        let _row = row?;
```
dbolduc: nit: (unless you plan to use `row`)

```suggestion
    while let Some(row) = iter.next().await.transpose()? {
```

dbolduc (approved):

tests/bigquery/src/lib.rs
alvarowolfx: changed. I always forget this `.transpose` thing

PR #6037, opened 2026-07-12
perf(spanner): store buffered_values in a VecDeque

fornwall:
`process_partial_result_set()` drained values off the front of a `Vec` once per complete row, causing a performance issue (being an **order of magnitude slower** than e.g. the [python spanner client](https://docs.cloud.google.com/python/docs/reference/spanner/latest)) to due excessive `memmove` on larger result sets.

Fix this by changing `buffered_values` to a `VecDeque`.

See https://github.com/fornwall/spanner-client-perf-issue for a benchmark that contains before and after flamegraphs and concrete [numbers](https://github.com/fornwall/spanner-client-perf-issue#side-by-side).

Additionally, as a related minor enhancement, introduce a capacity reservation on `ready_rows` to pre-allocate memory based on the buffered values and column count.


olavloite (approved):

olavloite:
@fornwall Thanks for this fix. This does indeed make a significant difference.

> being an order of magnitude slower than e.g. the [python spanner client](https://docs.cloud.google.com/python/docs/reference/spanner/latest)

I was not able to reproduce this, though. At least not with real Spanner. It could be that the Emulator uses a different chunking strategy that real Spanner, making the comparison different. What I saw using real Spanner was this:

| Benchmark | Python | Rust Before | Rust After |
| --- | --- | --- | --- |
| 200,000 rows - 2 cols | 5.75s | 2.26s | 0.11s |
| 100,000 rows - 12 cols | 2.62s | 1.07s | 0.35s |

coryan (approved):

PR #6044, opened 2026-07-13
impl(bigquery): parse numeric types as decimal

alvarowolfx:
Towards #5844


src/bigquery/src/query/from_sql.rs
```diff
@@ -51,6 +51,25 @@ impl FromSql for String {
     }
 }
 
+impl FromSql for i32 {
+    fn from_sql(value: wkt::Value) -> Result<Self, ConvertError> {
+        match value {
+            wkt::Value::Number(n) => n
+                .as_i64()
+                .and_then(|v| i32::try_from(v).ok())
+                .ok_or_else(|| ConvertError::Convert("number is not a valid i32".into())),
```
dbolduc, line 60: nit: should `ConvertError::Convert` wrap the `i32::try_from()` error?

src/bigquery/src/query/from_sql.rs
```diff
@@ -69,6 +88,25 @@ impl FromSql for i64 {
     }
 }
 
+impl FromSql for f32 {
+    fn from_sql(value: wkt::Value) -> Result<Self, ConvertError> {
+        match value {
+            wkt::Value::Number(n) => n
+                .as_f64()
+                .map(|v| v as f32)
+                .ok_or_else(|| ConvertError::Convert("invalid f32 number".into())),
```
dbolduc: super nit: to match the other branch: "number is not a valid f32"

dbolduc (approved):

src/bigquery/src/query/from_sql.rs
alvarowolfx: `i32::try_from` error produces "out of range integral type conversion". It does not mention `i32`, not sure if is much better.

src/bigquery/src/query/from_sql.rs
alvarowolfx: ack

src/bigquery/src/query/from_sql.rs
dbolduc: Fair

PR #6046, opened 2026-07-14
feat(gax): implement grpc-rust client constructors

joshuatants:
for #5991

### DIFF
--- a/src/gax-internal/src/grpc/grpc_rust.rs
+++ b/src/gax-internal/src/grpc/grpc_rust.rs
@@ -12,31 +12,59 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+use super::grpc_helpers;
 use crate::grpc::tonic::{Extensions, Response as TonicResponse, Result as TonicResult};
+use crate::options::{ClientConfig, InstrumentationClientInfo};
+use crate::universe_domain::DEFAULT_UNIVERSE_DOMAIN;
+use google_cloud_auth::credentials::Credentials;
 use google_cloud_gax::Result as GaxResult;
-use google_cloud_gax::client_builder::Result as ClientBuilderResult;
+use google_cloud_gax::client_builder::{Error as BuilderError, Result as ClientBuilderResult};
 use google_cloud_gax::options::RequestOptions;
 use google_cloud_gax::polling_backoff_policy::PollingBackoffPolicy;
 use google_cloud_gax::polling_error_policy::PollingErrorPolicy;
+use grpc::client::{Channel, ChannelOptions};
+use grpc::credentials::LocalChannelCredentials;
+use grpc::credentials::rustls::client::{ClientTlsConfig, RustlsChannelCredendials};
+use http::Uri;
 use std::sync::Arc;
 
-#[derive(Clone, Debug)]
-pub struct GrpcRustClient {}
+/// A gRPC client backed by the [grpc-rust][grpc] crate.
+#[derive(Clone)]
+pub struct GrpcRustClient {
+    inner: Arc<GrpcRustClientInner>,
+}
+
+// TODO(#5991): Will be used by `bidi_stream_with_status` in an upcoming commit.
+#[allow(dead_code)]
+struct GrpcRustClientInner {
+    credentials: Credentials,
+    transport_metric: crate::observability::TransportMetric,
+    tracing_attributes: Option<super::TracingAttributes>,
+    endpoint: ResolvedGrpcEndpoint,
+    invoker: Channel,
+}
+
+/// A gRPC endpoint resolved from the client configuration.
+#[derive(Clone, Debug, PartialEq, Eq)]
+struct ResolvedGrpcEndpoint {
+    target: String,
+    authority: String,
+    tls: bool,
+    server_address: String,
+    server_port: Option<u16>,
+}
 
 impl GrpcRustClient {
-    pub async fn new(
-        _config: crate::options::ClientConfig,
-        _default_endpoint: &str,
-    ) -> ClientBuilderResult<Self> {
-        unimplemented!("not implemented yet")
+    pub async fn new(config: ClientConfig, default_endpoint: &str) -> ClientBuilderResult<Self> {
+        Self::build(config, default_endpoint, None).await
     }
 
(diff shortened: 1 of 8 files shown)


src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -100,6 +128,101 @@ impl GrpcRustClient {
...
+        let endpoint = resolve_endpoint(&config, default_endpoint, &universe_domain)?;
+        let tracing_attributes = tracing_enabled
+            .then(|| make_tracing_attributes(default_endpoint, &endpoint, instrumentation))
+            .transpose()?;
+
+        // TODO(#5991) [Parity gap]: `Client` applies a per-connection
+        // concurrency limit of 100 in src/gax-internal/src/grpc.rs:493.
+        // We can't do this here: while grpc-rust has an internal transport
+        // option for this, it's not exposed in ChannelOptions.
+        // Does it make sense for grpc-rust to support this?
```
dfawley: I don't think so.  We will honor the server's [HTTP/2 setting](https://httpwg.org/specs/rfc7540.html#SETTINGS_MAX_CONCURRENT_STREAMS) for this per connection (which I'd be amazed if Tonic doesn't, too).  I'm not sure what else this setting is doing for you, besides preventing you from dispatching more than 100 calls at a time no matter what the server uses.  You should be able to apply limits like that externally if you think you need to, potentially via an interceptor.

src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -100,6 +128,101 @@ impl GrpcRustClient {
...
+
+        // TODO(#5991) [Parity gap]: The existing Tonic-based `Client` uses
+        // `Channel::balance_channel()` and inserts
+        // max(1, grpc_subchannel_count) copies of the endpoint. This creates
+        // independently connected subchannels behind Tonic's load balancer.
+        // While grpc-rust internally supports a `round_robin` policy, it is not
+        // exposed for use.
```
dfawley, line 154: For other languages, a client-libraries maintained channel pool is generally what is done. I think we would recommend the same here for the time being.

src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -100,6 +128,101 @@ impl GrpcRustClient {
...
+        // TODO(#5991) [Parity gap]: grpc-rust only uses a single channel
+        // authority for both the HTTP/2 `:authority` and TLS SNI.
+        // (In Tonic these can be different values.). Separate values
+        // are needed for global and VPC-SC private/restricted VIPs and
+        // Private Service Connect. See `crate::host::origin` and its tests for
+        // examples.
+        //
+        // Until grpc-rust supports separate TLS server-name and
+        // request-authority settings, use the request origin as the shared
+        // authority.
```
dfawley: AIUI this is possible in some, but not all, other grpc languages. However, this is not really a supported feature. We'd need to talk more about this to make sure it's supported in all languages if it's a requirement for you.

src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -116,3 +239,333 @@ where
...
+    })
+}
+
+// TODO(#5991): [Parity gap]: Tonic first uses an installed process-default
+// CryptoProvider; otherwise, it selects the provider enabled through its Cargo
+// features. In contrast, grpc-rust's RustlsChannelCredendials::new requires a
+// process-wide rustls CryptoProvider to be installed.
+//
+// Can grpc-rust either mimic Tonic's behaviour or provide a way to inject a
+// CryptoProvider?
```
dfawley: This is going to take some more thought from our side.

We can't trivially accept a `rustls::ClientConfig` or a `rustls::CryptoProvider`.  First, they are unstable (0.x).  But even if they were stable, they are foreign APIs and so we would not want to depend directly on them.

We may be able to redesign things a little so we have a `grpc-rustls` crate that provides that capability.

And we may also want to provide some kind of feature flag(s) to enable + configure rustls.

src/gax-internal/src/grpc/grpc_rust.rs
joshuatants: Upon further investigation it looks like it was added to [avoid stalling in Pub/Sub](https://github.com/googleapis/google-cloud-rust/commit/9af52624321dd3d675a3fc47c3f4f06c6722368a). It may not be needed for Storage.

src/gax-internal/src/grpc/grpc_rust.rs
joshuatants: Dug a bit deeper: as it turns out the ability to configure TLS SNI independently from the HTTP/2 :authority is not needed for Storage. For Cloud Storage custom domains routed through [private/restricted VIPs](https://cloud.google.com/vpc/docs/configure-private-google-access#config-domain), Google requires using [storage.googleapis.com](http://storage.googleapis.com/) for both HTTP/2 :authority and TLS SNI. For [Private Service Connect](https://cloud.google.com/vpc/docs/configure-private-service-connect-apis#use-endpoint), Google permits the same configuration. Therefore, independently configurable values are not required for these cases.

Will raise again if this turns out to be incorrect.

src/gax-internal/src/grpc/grpc_rust.rs
joshuatants: Ack, as we discussed this morning we'll go with pick_first for and take measurements to decide next steps. And as you pointed out if we implement a connection pool we'll have to figure out to disable it for DirectPath/xDS.

src/gax-internal/src/grpc/grpc_rust.rs
joshuatants: As discussed, will wait for grpc-rust to think through this.

joshuatants:
@dfawley I cleaned up the comments after our discussion this morning. Could you take a look at the rest of the code? Thank you!

coryan (approved):

joshuatants:
> @dfawley I cleaned up the comments after our discussion this morning. Could you take a look at the rest of the code? Thank you!

@dfawley merging to `grpc_rust` first so that I can continue working on the bidi implementation. Happy to address any post-merge comments.

PR #6050, opened 2026-07-14
perf(pubsub): move batch flush timer from Dispatcher into batch actors

panimidi:
perf(pubsub): move batch flush timer from Dispatcher into batch actors

The Dispatcher runs a single batch-flush timer that is armed unconditionally
and reset on every fire, regardless of whether any batch actor has buffered
messages. Because each Publisher owns its own Dispatcher, an application with
many publishers performs continuous timer work even when completely idle. In
`Dispatcher::run` the timer is created once, and on every `delay_threshold`
interval it flushes all batch actors and then re-arms itself unconditionally,
so an idle publisher keeps scheduling and servicing a timer with nothing to
send. At high publisher cardinality this becomes a constant source of
time-wheel work; we observed it dominating idle CPU in a production service
that fans a single logical publisher out across several thousand topics.

For reference, the Java client's `Publisher.setupAlarm` only schedules the
alarm while a batch is non-empty and cancels it once the batch drains, so it
holds no timer at rest. This change brings the Rust client in line with that
behavior.

This change removes the shared timer from `Dispatcher::run`, leaving the
Dispatcher as a pure router of Publish, Flush, and ResumePublish commands, and
adds a lazy flush timer to both `ConcurrentBatchActor` and
`SequentialBatchActor`. Each actor arms its timer when the first message enters
an otherwise-empty batch, re-arms it after a timer flush that leaves work
pending, and disarms it once a flush drains the batch. An idle actor then
blocks in `recv` with no scheduled timer and therefore no time-wheel entry.

Batching semantics are unchanged: a batch is still flushed after at most
`delay_threshold`. There is no public API change and no change to item
visibility. Existing tests cover the behavior, including
`batch_sends_on_delay_threshold`, which disables the count and byte thresholds
and asserts that batches flush at exactly `delay_threshold` across repeated arm
and disarm cycles, for both keyed and unkeyed messages. The package test suite
passes.


panimidi:
For full disclosure: the branch, [`perf/pubsub-lazy-timers-full`](https://github.com/panimidi/google-cloud-rust/tree/perf/pubsub-lazy-timers-full), contains this timer change **plus** an additional optimization that isn't part of this PR: a fast path where `Publisher::publish` sends messages without an ordering key directly to a pre-created default batch actor, bypassing the `Dispatcher` entirely.

This is the exact code we ran our tests against, leaving it here for reference but I believe the "fast-path" adds a lot of noise like visibility changes and wasn't the main driver of performance improvement anyway.

suzmue:
Thank you for the PR! In addition to signing the CLA, would you please file an issue corresponding to this PR? (See [Contributing](https://github.com/googleapis/google-cloud-rust/blob/main/CONTRIBUTING.md#before-contributing-code)).

Edit: Filed #6052, no need to file an issue.

src/pubsub/src/publisher/actor.rs
```diff
@@ -276,23 +258,35 @@ impl ConcurrentBatchActor {
...
+                // Flush on timer — only active when messages are pending.
+                _ = async { timer.as_mut().unwrap().await }, if timer.is_some() => {
+                    self.flush(&mut inflight, &mut batch);
+                    if batch.is_empty() && inflight.is_empty() {
+                        timer = None;
+                    } else {
+                        timer.as_mut().unwrap().as_mut()
+                            .reset(tokio::time::Instant::now() + delay);
+                    }
+                }
```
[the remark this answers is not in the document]
suzmue: This is correct, the concurrent batch actor flushes all messages. Please fix.

src/pubsub/src/publisher/actor.rs
```diff
@@ -276,23 +258,35 @@ impl ConcurrentBatchActor {
...
                 msg = self.context.rx.recv() => {
                     match msg {
                         Some(ToBatchActor::Publish(msg)) => {
                             self.add_msg_and_flush(&mut inflight, &mut batch, msg);
+                            if timer.is_none() {
+                                timer = Some(Box::pin(tokio::time::sleep(delay)));
+                            }
                         },
```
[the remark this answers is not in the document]
suzmue: This is correct. The batch may be empty here. Either checking the batch to see if it is empty or modifying add_msg_and_flush to return whether it left messages in the batch would be reasonable options.

src/pubsub/src/publisher/actor.rs
```diff
@@ -265,6 +242,11 @@ impl ConcurrentBatchActor {
     /// Dispatcher drops the Sender.
     async fn run(mut self) {
         // We have multiple inflight batches concurrently.
+        let delay = self.context.batching_options.delay_threshold;
+        // Lazy timer: armed on the first message, disarmed after a flush that
+        // leaves the batch empty. An idle actor blocks in recv() with no time-
+        // wheel entry — zero scheduler overhead at rest.
+        let mut timer: Option<std::pin::Pin<Box<tokio::time::Sleep>>> = None;
         let mut inflight = JoinSet::new();
```
suzmue: The comment got split from the line of code it was referencing.


```suggestion
        let delay = self.context.batching_options.delay_threshold;
        // Lazy timer: armed on the first message, disarmed after a flush that
        // leaves the batch empty. An idle actor blocks in recv() with no time-
        // wheel entry — zero scheduler overhead at rest.
        let mut timer: Option<std::pin::Pin<Box<tokio::time::Sleep>>> = None;
        // We have multiple inflight batches concurrently.
        let mut inflight = JoinSet::new();
```

src/pubsub/src/publisher/actor.rs
```diff
@@ -413,25 +412,38 @@ impl SequentialBatchActor {
...
+                _ = async { timer.as_mut().unwrap().await }, if timer.is_some() => {
+                    self.flush(&mut inflight, &mut batch).await;
+                    inflight = JoinSet::new();
+                    if self.pending_msgs.is_empty() && batch.is_empty() {
+                        timer = None;
+                    } else {
+                        timer.as_mut().unwrap().as_mut()
+                            .reset(tokio::time::Instant::now() + delay);
+                    }
+                }
```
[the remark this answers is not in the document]
suzmue: This is also correct. We block until all messages are sent, so it will be empty at the time we reach this line.

src/pubsub/src/publisher/actor.rs
```diff
@@ -413,25 +412,38 @@ impl SequentialBatchActor {
...
                     match msg {
                         Some(ToBatchActor::Publish(msg)) => {
                             self.pending_msgs.push_back(msg);
                             if inflight.is_empty() {
                                 self.move_to_batch_and_flush(&mut inflight, &mut batch);
                             }
+                            if timer.is_none() {
+                                timer = Some(Box::pin(tokio::time::sleep(delay)));
+                            }
                         },
```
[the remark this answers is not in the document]
suzmue: This is correct as well. A flush may have just occurred so there may be no messages.

src/pubsub/src/publisher/actor.rs
```diff
@@ -413,25 +412,38 @@ impl SequentialBatchActor {
                     self.handle_inflight_join(join);
                     self.move_to_batch_and_flush(&mut inflight, &mut batch);
                 }
+                // Flush on timer — only active when messages are pending.
+                _ = async { timer.as_mut().unwrap().await }, if timer.is_some() => {
```
suzmue, line 438: I don't think it is necessary to wrap the timer.as_mut().unwrap().await in an async block.

If there is a reason that we do need to wrap it, please explain it in a comment.

src/pubsub/src/publisher/actor.rs
```diff
@@ -413,25 +412,38 @@ impl SequentialBatchActor {
...
+                    self.flush(&mut inflight, &mut batch).await;
+                    inflight = JoinSet::new();
+                    if self.pending_msgs.is_empty() && batch.is_empty() {
+                        timer = None;
+                    } else {
+                        timer.as_mut().unwrap().as_mut()
+                            .reset(tokio::time::Instant::now() + delay);
+                    }
```
suzmue: The flush code should look identical to the `Some(ToBatchActor::Flush(tx))` branch for both the ConcurrentBatchActor and the SequentialBatchActor.

src/pubsub/src/publisher/actor.rs
```diff
@@ -413,25 +412,38 @@ impl SequentialBatchActor {
...
                             self.flush(&mut inflight, &mut batch).await;
                             inflight = JoinSet::new();
+                            timer = None;
                             let _ = tx.send(());
                         },
                         Some(ToBatchActor::ResumePublish()) => {
                             // Nothing to resume as we are not paused.
                         },
                         None => {
-                            // This isn't guaranteed to execute if a user does not .await on the
```
suzmue, line 433: Extraneous diff, please restore comment.

src/pubsub/src/publisher/actor.rs
```diff
@@ -276,23 +258,35 @@ impl ConcurrentBatchActor {
...
                             inflight.join_all().await;
                             inflight = JoinSet::new();
+                            timer = None;
                             let _ = tx.send(());
                         },
                         Some(ToBatchActor::ResumePublish()) => {
                             // Nothing to resume as we do not pause without ordering key.
                         }
                         None => {
-                            // This isn't guaranteed to execute if a user does not .await on the
```
suzmue, line 294: Extraneous diff, please restore comment.

src/pubsub/src/publisher/actor.rs
```diff
@@ -167,35 +154,25 @@ impl Dispatcher {
...
-                                // Send down the same tx for the BatchActors to directly signal completion
-                                // instead of spawning a new task.
                                 if batch_actor.sender.send(ToBatchActor::ResumePublish()).is_err() {
                                     return; // Stop the dispatcher if a batch actor is dropped.
                                 }
                             }
                         }
                         None => {
                             // Gracefully shutdown since the Publisher has dropped the Sender.
-                            // By dropping the BatchActorHandles, they will individually handle the
```
suzmue, line 186: Extraneous diff, please restore comment.

src/pubsub/src/publisher/actor.rs
```diff
@@ -167,35 +154,25 @@ impl Dispatcher {
...
-                            // shutdown procedures.
                             drop(batch_actors);
-                            // When we drop actor_tasks, some batch actors may not have started execution
```
suzmue, line 189: Extraneous diff, please restore comment.

src/pubsub/src/publisher/actor.rs
```diff
@@ -167,35 +154,25 @@ impl Dispatcher {
...
                             }
                             tokio::spawn(async move {
-                                // Wait on all the flush operations.
                                 flush_set.join_all().await;
                                 let _ = tx.send(());
                             });
                         },
                         Some(ToDispatcher::ResumePublish(ordering_key)) => {
                             if let Some(batch_actor) = batch_actors.get_mut(&ordering_key) {
-                                // Send down the same tx for the BatchActors to directly signal completion
```
suzmue, line 177: Extraneous diff, please restore comment.

src/pubsub/src/publisher/actor.rs
```diff
@@ -167,35 +154,25 @@ impl Dispatcher {
                                 flush_set.spawn(rx);
                             }
                             tokio::spawn(async move {
-                                // Wait on all the flush operations.
```
suzmue, line 170: Extraneous diff, please restore comment.

src/pubsub/src/publisher/actor.rs
```diff
@@ -112,7 +112,10 @@ impl Dispatcher {
     ///    all BatchActors and awaits its completion.
     /// 3. A ResumePublish command from the `Publisher` is dispatched to the BatchActor
     ///    for its ordering key.
-    /// 4. A timer fire causes the Dispatcher to flush all BatchActors.
+    ///
+    /// Each batch actor owns its own lazy flush timer (see `ConcurrentBatchActor`
+    /// and `SequentialBatchActor`), so the Dispatcher no longer runs a timer of
+    /// its own.
```
suzmue: nit: this will live in our code forever and "no longer" isn't necessary to document the state.

```suggestion
    /// Each batch actor owns its own lazy flush timer (see `ConcurrentBatchActor`
    /// and `SequentialBatchActor`).
```

src/pubsub/src/publisher/actor.rs
```diff
@@ -265,6 +242,11 @@ impl ConcurrentBatchActor {
     /// Dispatcher drops the Sender.
     async fn run(mut self) {
         // We have multiple inflight batches concurrently.
+        let delay = self.context.batching_options.delay_threshold;
+        // Lazy timer: armed on the first message, disarmed after a flush that
+        // leaves the batch empty. An idle actor blocks in recv() with no time-
+        // wheel entry — zero scheduler overhead at rest.
```
suzmue: nit: When reviewing it was initially unclear to me why the tradeoff of using a heap allocation on (potentially) every batch was worth it compared to adding a boolean to guard the tokio::pin!. I believe from your comment it is because we want to fully get rid of the timer to get rid of the entries in the time wheel. I have included a suggestion for updating the comment that makes it a bit more explicit.

I am curious, did you also test this with a single timer that gets reset, as opposed to one that gets dropped? Was there a performance difference?

```suggestion
// Lazy timer: Armed when the first message enters a batch and dropped (`None`)
// when a flush drains the batch. Dropping the timer unregisters it from Tokio's 
// timer driver, ensuring idle actors hold zero time-wheel entries and incur 
// zero CPU wakeups at rest.
```

panimidi:
Thank you for the comments, I'll be reviewing them and sending fixes later.

I am also looking into the CLA, my company (MercadoLibre) doesn't currently have a CCLA signed with Google so it may take a little while.

src/pubsub/src/publisher/actor.rs
panimidi: I did not, I wanted to mirror the Java client library's behavior (canceling the scheduled task -> dropping it here), but it's a good point. Will test and report back later.

panimidi: Measured this with a small demo project (https://github.com/panimidi/pubsub-demo) for quick repro — 2000 publishers on one topic, delay_threshold = 1ms, against a local emulator — with a 10s idle phase and a 20s load phase at ~500 msg/s.

 Results reading the client process's own CPU via getrusage:

 | client                      | idle %CPU | load %CPU |
 |-----------------------------|-----------|-----------|
 | unmodified (before this PR) | 154.6%    | 228.6%    |
 | this PR — drop timer        | 0.0%      | 12.7%     |
 | alternative — reset timer   | 0.0%      | 12.6%     |

 (The 1ms delay and 2k publishers is a stress case to show the issue clearly)

 The reset variant shown on the table is on branch [perf/pubsub-lazy-timers-reset](https://github.com/panimidi/google-cloud-rust/tree/perf/pubsub-lazy-timers-reset) if you'd like to review, but I'd lean on keeping the drop version.

panimidi:
Force-pushed to re-author the commits for the corporate CLA. No content changes.

src/pubsub/src/publisher/actor.rs
suzmue: Ack. Thank you!

src/pubsub/src/publisher/actor.rs
suzmue: Thanks for checking!

I additionally ran some throughput benchmarks across different scenarios (unordered keys, many ordered keys (all active), many ordered keys (majority idle)) and saw improvements across all of them. Thank you for making this improvement!

I'll also see if we can incorporate something like your demo project into our repo so that we can make sure we don't have CPU usage regressions, thanks for linking to that.

suzmue (approved):

suzmue:
Thank you for this @panimidi! I went ahead and rebased onto main so we can get this merged.

src/pubsub/src/publisher/actor.rs
```diff
@@ -373,6 +384,11 @@ impl SequentialBatchActor {
     /// The loop terminates when the `rx` channel is closed, which happens when the
     /// Dispatcher drops the Sender.
     async fn run(mut self) {
+        let delay = self.context.batching_options.delay_threshold;
+        // Lazy timer: same pattern as ConcurrentBatchActor — armed when the
+        // first message arrives in a new batch cycle, disarmed after a flush
+        // that drains all pending messages.
+        let mut timer: Option<std::pin::Pin<Box<tokio::time::Sleep>>> = None;
```
coryan, line 391: Should we name this type? It is quite the mouthful and appears a couple of times in the file.

We have also been moving towards use declarations over fully qualified name, so this would be `Option<Pin<Box<Sleep>>>` with some suitable `use ` things at the top.

Feel free to make this changes in a followup PR.

coryan (approved):
A small nit to feel useful.

src/pubsub/src/publisher/actor.rs
suzmue: Let's do this in a follow up!

PR #6070, opened 2026-07-16
perf(spanner): add ToValue for &[u8]

fornwall:
`ToValue` was implemented for `Vec<u8>` but not for byte slices, so a caller holding a `&[u8]` had to allocate an owned `Vec<u8>` purely to encode it, adding a copy per binary cell on potentially hot binding paths.

Add `impl ToValue for &[u8]` holding the BYTES/base64 encoding, and have the `Vec<u8>` impl delegate to it via `self.as_slice()`.


src/spanner/src/to_value.rs
```diff
@@ -170,6 +170,12 @@ impl ToValue for f32 {
 }
 
 impl ToValue for Vec<u8> {
+    fn to_value(&self) -> Value {
+        self.as_slice().to_value()
+    }
+}
+
+impl ToValue for &[u8] {
```
olavloite: Can we change this to target `[u8]` instead?


```suggestion
impl ToValue for [u8] {
```

And then add this test:

```rust
    #[test]
    fn test_to_value_slice_trait_bound() {
        fn bind_param<T: ToValue + ?Sized>(val: &T) -> Value {
            val.to_value()
        }

        let slice: &[u8] = &[1, 2, 3];
        let v = bind_param(slice);
        assert_eq!(v.kind(), Kind::String);
        assert_eq!(v.as_string(), "AQID");
    }
```

fornwall: Nice! Applied your suggestion in b287bd4a8e601fc01b62557086e081cdc332a9d3 and added the test in afddb84777687ebe6ef94aef5aca66bb1e3d2aaf.

olavloite (approved):

sakthivelmanii (approved):

fornwall:
Replaced by https://github.com/googleapis/google-cloud-rust/pull/6184

PR #6075, opened 2026-07-16
impl(bigquery): add take method to Row

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery/src/query/row.rs
+++ b/src/bigquery/src/query/row.rs
@@ -128,9 +128,6 @@ impl Row {
 
     /// Takes ownership of a value from the row by column name or zero-based index.
     /// The value in the row is replaced with `Value::Null` in-place to avoid cloning.
-    ///
-    /// If type conversion fails, the original cell value is not restored into the row,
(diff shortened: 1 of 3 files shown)


src/bigquery/src/query/row.rs
```diff
@@ -298,6 +322,21 @@ mod tests {
         assert_eq!(row.get::<f64, _>(4), 64.0);
         assert_eq!(row.get::<f64, _>("some_float"), 64.0);
 
+        assert_eq!(row.take::<String, _>(0)?, "James");
+        assert_eq!(row.try_get::<Option<String>, _>(0)?, None);
```
dbolduc, line 326: comment: Alternatively, we could have `try_get::<String>` and verified the null value error, but this works too.

dbolduc (approved):

src/bigquery/src/query/row.rs
alvarowolfx: yeah, I was bouncing back and forth between the two options of checks

PR #6087, opened 2026-07-17
feat(spanner): implement ToValue for str

### LINKED ISSUE #6084: spanner: impl ToValue for str
The Spanner client should (also) implement `ToValue` for `str`:

```rust
impl ToValue for str {
    // ...
}
```

The client only implements the trait for `&str`, which forces users to write code like this:

```rust
    .add_param("name", &"Garcia")
```

fornwall:
`ToValue` was implemented for `&str` but not for the unsized `str`, meaning that callers had to pass an extra reference (`&"hello"`).

Fixes #6084.

### DIFF
--- a/src/spanner/src/to_value.rs
+++ b/src/spanner/src/to_value.rs
@@ -65,17 +65,21 @@ impl ToValue for ProtoValue {
 
 impl ToValue for String {
     fn to_value(&self) -> Value {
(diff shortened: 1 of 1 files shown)


fornwall:
Replaced by https://github.com/googleapis/google-cloud-rust/pull/6184.

Issue #4459, opened 2026-01-30
Flake in `auth` integration tests


coryan:
First time I see this:

```
failures:
  
---- driver::run_id_token_adc_impersonated stdout ----
 
thread 'driver::run_id_token_adc_impersonated' (7342) panicked at tests/integration-auth/src/lib.rs:601:10:
failed to get id token: CredentialsError { is_transient: true, message: Some("failed to fetch token, body=<{\n  \"error\": {\n    \"code\": 500,\n    \"message\": \"Unable to extract the resource from the request.\",\n    \"status\": \"INTERNAL\"\n  }\n}\n>"), source: Some(reqwest::Error { kind: Status(500, None), url: "https://iamcredentials.googleapis.com/v1/projects/-/serviceAccounts/impersonation-target@rust-auth-testing.iam.gserviceaccount.com:generateIdToken" }) }
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
 
```
https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/09072aa8-ba52-4d5b-9522-92c92cc0f047?project=rust-auth-testing&e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled&mods=allow_workbench_image_override

It may be a legitimate transient error that needs some kind of retry loop.

alvarowolfx:
that's gonna be improved by https://github.com/googleapis/google-cloud-rust/pull/4260, https://github.com/googleapis/google-cloud-rust/pull/4257 and https://github.com/googleapis/google-cloud-rust/pull/4256, which adds retry to idtoken.

dbolduc:
I think this is a similar enough error, but for a different ID token integration test.

```
---- driver::run_workload_identity_provider_programmatic_sourced stdout ----
thread 'driver::run_workload_identity_provider_programmatic_sourced' (743) panicked at tests/integration-auth/src/lib.rs:492:10:
[...]

PR #6090, opened 2026-07-17
feat(auth): add retry loop for generate_id_token

alvarowolfx:
Towards #4459

### DIFF
--- a/src/auth/src/credentials/idtoken/impersonated.rs
+++ b/src/auth/src/credentials/idtoken/impersonated.rs
@@ -427,11 +427,7 @@ async fn generate_id_token_call(
             .bytes()
             .await
             .map_err(|e| GaxError::transport(err_headers.clone(), e))?;
-        return Err(GaxError::http(
-            status.as_u16(),
-            err_headers,
-            format!("{MSG}: {err_payload:?}").into(),
-        ));
+        return Err(GaxError::http(status.as_u16(), err_headers, err_payload));
     }
 
     Ok(response)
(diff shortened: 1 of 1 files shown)


src/auth/src/credentials/idtoken/impersonated.rs
```diff
@@ -384,41 +390,130 @@ async fn generate_id_token(
...
+    GenerateIdTokenClient::new(
+        source_headers,
         delegates,
+        audience,
         include_email,
-    };
+        service_account_impersonation_url.to_string(),
+    )
+    .fetch()
+    .await
```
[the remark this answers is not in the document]
alvarowolfx: we can pass retry_policy and backoff_policy on later PRs

src/auth/src/credentials/idtoken/impersonated.rs
```diff
@@ -384,41 +390,130 @@ async fn generate_id_token(
...
+
+impl GenerateIdTokenClient {
+    fn new(
+        source_headers: HeaderMap,
+        delegates: Option<Vec<String>>,
+        audience: String,
+        include_email: Option<bool>,
+        url: String,
+    ) -> Self {
+        let retry_policy = Aip194Strict.with_time_limit(Duration::from_secs(60));
```
coryan: Consider:

```suggestion
        let retry_policy = Aip194Strict
            .continue_on_too_many_request()
            .continue_on_client_timeout()
            .with_time_limit(Duration::from_secs(60));
```

That will handle common errors that are retrying for these operations.

Also consider a attempt timeout of 10s or something similar. Otherwise this can stall for 60s on the first attempt.

coryan (approved):
All the suggestions are optional, this PR improves things, we can improve them further in a future PR

src/auth/src/credentials/idtoken/impersonated.rs
alvarowolfx: changed on this PR to call the `continue_on_*` methods and also added a per call limit of 10s. Will do the same in other places that have a similar pattern in the auth crate

PR #6091, opened 2026-07-17
test(auth): set idempotency for generate_id_token with IAM client

alvarowolfx:
Towards https://github.com/googleapis/google-cloud-rust/issues/4459 https://github.com/googleapis/google-cloud-rust/issues/6081

### DIFF
--- a/tests/integration-auth/src/lib.rs
+++ b/tests/integration-auth/src/lib.rs
@@ -32,6 +32,7 @@ use google_cloud_auth::credentials::{
 use google_cloud_auth::errors::SubjectTokenProviderError;
 use google_cloud_bigquery_v2::client::DatasetService;
 use google_cloud_gax::error::rpc::Code;
+use google_cloud_gax::options::RequestOptionsBuilder;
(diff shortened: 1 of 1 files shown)


tests/integration-auth/src/lib.rs
```diff
@@ -532,6 +533,7 @@ async fn generate_id_token(
 
     let res = client
         .generate_id_token()
+        .with_idempotency(true)
```
coryan, line 536: non-blocking: this should be true by default... even if the operation is a POST: creating the token has no side-effects, can we make that change (maybe it requires some changes in the generator and tuning).

coryan (approved):

tests/integration-auth/src/lib.rs
alvarowolfx: agree with you, but do we have hints in the service protos to derive that ?

coryan: No, we would need to add an annotation in  `librarian.yaml`.

PR #6093, opened 2026-07-17
fix(auth): add all 5xx status codes as retryable error

alvarowolfx:
Towards https://github.com/googleapis/google-cloud-rust/issues/6080

### DIFF
--- a/src/auth/src/errors.rs
+++ b/src/auth/src/errors.rs
@@ -140,17 +140,12 @@ fn is_retryable(err: &reqwest::Error) -> bool {
 }
 
 fn is_retryable_code(code: StatusCode) -> bool {
-    match code {
-        // Internal server errors do not indicate that there is anything wrong
-        // with our request, so we retry them.
-        StatusCode::INTERNAL_SERVER_ERROR
-        | StatusCode::BAD_GATEWAY
-        | StatusCode::SERVICE_UNAVAILABLE
-        | StatusCode::GATEWAY_TIMEOUT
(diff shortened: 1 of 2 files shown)


alvarowolfx:
I'm not sure if we should retry on 502/504. AIP 195  states that we shouldn't. Thoughts @coryan @dbolduc ?

src/auth/src/errors.rs
```diff
@@ -144,7 +144,9 @@ fn is_retryable_code(code: StatusCode) -> bool {
         // Internal server errors do not indicate that there is anything wrong
         // with our request, so we retry them.
         StatusCode::INTERNAL_SERVER_ERROR
+        | StatusCode::BAD_GATEWAY
```
coryan: Actionable: please just retry all 500s...  Worse things that happens is we keep running the background refresh loop, no? None of the operations here have side-effects that we should worry about.

optional: maybe we should revert this: list the error codes are *not* retryable (things like "403 - PERMISSION DENIED")?

coryan (changes requested):

coryan:
> I'm not sure if we should retry on 502/504. AIP 195 states that we shouldn't.

I think you mean 194... as I said in the other comment: AFAICT, these operations are safe to retry, they never have side-effects.

src/auth/src/errors.rs
alvarowolfx: added the other 500s codes by using `code.is_server_error()`. I prefer that than reverting it, I think is more explicit

coryan (approved):

Issue #2318, opened 2025-05-29
Design a solution for streaming RPCs in generated clients


coryan:
We may launch (some) clients before we have a full implementation of streaming RPCs, but we should have a solution designed before doing so.

The problems to solve include:

- What Rust surface do we expose for server-side, and bi-di streaming RPCs?
- Do we support client-side streaming RPCs? They are rare enough we may not need to (e.g. if only veneers have them).
- How do we automatically add a gRPC client to an existing REST client. We need this for bi-di streaming RPCs.
- Do we need to lazy initialize the inner gRPC client?
- Do we need to support progress timeouts?

coryan:
I think we can post-pone the detailed design.  I prototyped enough to convince myself that none of these changes will break existing clients.

Roughly the outline would be:

- If we have streaming RPCs the generator will inject `_internal-grpc-client` as a feature to the `gaxi` dependency.
   - It seems easier to hardcode this, but now that the version numbers are not part of the generated `Cargo.toml` file, hardcoding a few deps or their features is less brittle than before.
- If we have streaming RPCs `transport.rs.mustache` file add a `grpc::Client` to the transport. The initialization does not change.
- The trickiest part is generating the Prost types.
   - We should generate a `build.rs` file in $TMPDIR and have the generator call this code on the fly (see #2662).
   - We should add attributes to only enable the messages we need, as we do for split clients.
   - Generating the `include!()` things requires a new data structure to navigate the namespaces

dbolduc:
> Generating the `include!()` things requires a new data structure to navigate the namespaces

FYI: We can avoid this with:

https://github.com/googleapis/librarian/blob/aad765cf0b4be34999e8dd78ad045b1f5b597001/internal/sidekick/rust_prost/templates/tonic/build.rs.mustache#L40

https://github.com/googleapis/librarian/blob/aad765cf0b4be34999e8dd78ad045b1f5b597001/internal/sidekick/rust/templates/grpc-mock/src/lib.rs.mustache#L128

suzmue:
This work is completed for our generated clients and will be included in the next release for bidi and server streaming. There are no client side streaming methods in production code that we generate, so this work has been deferred. Please open a new issue if you have any issues or feature requests related to streaming.

PR #6101, opened 2026-07-20
impl(gax): add RequestSender and ResponseReceiver streaming types

suzmue:
Implement `RequestSender<Req>` and `ResponseReceiver<Resp>` wrapper types in `google-cloud-gax` for gRPC streaming handle management. For now, these are very simple wrappers around Tokio `mpsc` channel types, so that we can start generating code that uses them.

For #2318

### DIFF
--- a/src/gax/src/streaming.rs
+++ b/src/gax/src/streaming.rs
@@ -29,11 +29,14 @@ impl<Req> RequestSender<Req> {
     }
 
     /// Sends a request item over the stream.
-    pub async fn send(&self, item: Req) -> Result<(), crate::error::Error> {
+    pub async fn send(&self, item: Req) -> Result<(), crate::error::Error>
+    where
+        Req: Send + Sync + 'static,
+    {
         self.req_tx
             .send(item)
             .await
-            .map_err(|_| crate::error::Error::io("stream closed"))
+            .map_err(crate::error::Error::io)
     }
 }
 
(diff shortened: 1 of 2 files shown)


src/gax/src/streaming.rs
```diff
@@ -12,4 +12,68 @@
...
+    pub fn new(req_tx: mpsc::Sender<Req>) -> Self {
+        Self { req_tx }
+    }
+
+    /// Sends a request item over the stream.
+    pub async fn send(&self, item: Req) -> Result<(), crate::error::Error> {
+        self.req_tx
+            .send(item)
+            .await
+            .map_err(|_| crate::error::Error::io("stream closed"))
```
coryan: Shouldn't this be:


```suggestion
            .map_err(|e| crate::error::Error::io(e))
```

If you think the underlying error is an implementation detail, consider a comment for future-us.

coryan (approved):

src/gax/src/streaming.rs
```diff
@@ -12,4 +12,68 @@
...
+    /// Creates a new [`ResponseReceiver`].
+    pub fn new(rx: mpsc::Receiver<crate::Result<Resp>>) -> Self {
+        Self { rx }
+    }
+
+    /// Receives the next response item from the stream.
+    pub async fn recv(&mut self) -> Option<crate::Result<Resp>> {
+        self.rx.recv().await
+    }
+}
```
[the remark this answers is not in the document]
suzmue: Agreed, but lets do this in a future PR, and probably will be into_stream like the rest of our types.

src/gax/src/streaming.rs
suzmue: You are correct. Updated.

Issue #5997, opened 2026-07-07
FR: Support Custom Headers (HTTP & gRPC) in GCS Client


xlai20:
Add support for injecting custom HTTP headers into Google Cloud Storage (GCS) SDK requests. This is primarily used for advanced debugging, distributed tracing (e.g., `X-Client-Tracking-Id`), and custom middlebox routing.

**Requirements:**
 - **Global Client Configuration**: Allow developers to configure custom headers when instantiating the storage client (e.g., adding `with_custom_headers` to `ClientBuilder` in `google_cloud_storage` and updating `ClientConfig` in `google_cloud_gax`).
 - **Method/Request-Level Configuration**: Allow custom headers to be specified on specific API calls to supplement or override the client-level headers.
 - **Transport Support**: Changes must apply to both HTTP (`reqwest`) and gRPC (`tonic`) transports within `google_cloud_gax_internal`.
 
For more details on the use cases for this feature, please refer to the Google Cloud documentation: https://docs.cloud.google.com/storage/docs/troubleshooting#custom-headers.

PR #6103, opened 2026-07-20
fix(gax): enforce system-wins precedence for grpc headers

xlai20:
## Overview
This PR ensures that core system headers (like `User-Agent` and `x-goog-user-project`) always strictly override any custom headers provided by the user, preventing accidental duplicate keys.

Previously, custom headers were appended at the end of `make_headers`. If a user passed a custom header that collided with a system header, both would be sent over the wire. This could unintentionally pollute SDK telemetry or disrupt backend routing.

go/rust-sdk-custom-header

## Changes
* We now merge custom headers *before* applying system headers in `make_headers`.
* Swapped `.append()` to `.insert()` for system headers so they cleanly overwrite any colliding custom keys.
* Added a unit test to verify this "system-wins" precedence logic.


src/gax-internal/src/grpc/grpc_helpers.rs
```diff
@@ -305,4 +308,89 @@ mod tests {
...
+    #[test]
+    fn make_headers_enforces_system_wins_precedence() -> TestResult {
+        // Arrange
+        const USER_AGENT: &str = "system-user-agent/v1.2.3";
+        const QUOTA_PROJECT: &str = "system-quota-project";
+        const REQUEST_PARAMS: &str = "resource=projects%2Ftest";
+
+        let mut options = RequestOptions::default();
+        options.set_user_agent(USER_AGENT);
+        options.set_quota_project(QUOTA_PROJECT);
```
ccbcqbz: If the system doesn't set these, a user-provided custom header for the same key will still survive. Is it intended? 

Similarly, x-goog-request-params only overrides when !request_params.is_empty()

joshuatants: I think this can be fixed by first clearing these reserved headers after retrieving the custom_headers before proceeding.

joshuatants (changes requested):

src/gax-internal/src/grpc/grpc_helpers.rs
xlai20: I've added a block of code to unconditionally sanitize the custom headers so that all keys that conflict with system headers will be removed. (Note, it is outside the if block of custom headers retrieval, because in my next PR when I implement global-level custom headers, I need this sanitize block of code again)

And also adjust the test cases to cover this.

src/gax-internal/src/grpc/grpc_helpers.rs
```diff
@@ -305,4 +317,133 @@ mod tests {
...
+        );
+        assert_eq!(api_clients[0], API_CLIENT_HEADER);
+
+        // Assert the legitimate header is still present
+        assert_eq!(
+            headers
+                .get("x-legitimate-header")
+                .expect("x-legitimate-header should be present"),
+            "legitimate-value"
+        );
```
joshuatants: Is it possible to make this more concise? For example, assert!(headers.len() == 1) instead of checking them one by one?

joshuatants (approved):

src/gax-internal/src/grpc/grpc_helpers.rs
xlai20: Updated the test to make it more concise by directly asserting map contents.

xlai20:
This is also part of https://github.com/googleapis/google-cloud-rust/issues/5997

PR #6105, opened 2026-07-20
impl(bigquery): add FromRow derive macro

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery-derive/src/lib.rs
+++ b/src/bigquery-derive/src/lib.rs
@@ -20,7 +20,7 @@ use proc_macro::TokenStream;
 use quote::quote;
 use syn::{Data, DeriveInput, Fields, parse_macro_input};
 
-/// Derives standard library [`TryFrom`](std::convert::TryFrom) for converting a BigQuery `Row` into a struct.
+/// Derives standard library [TryFrom] for converting a BigQuery `Row` into a struct.
 ///
 /// Supports renaming attributes via `#[bigquery(rename = "new_name")]`.
 #[proc_macro_derive(FromRow, attributes(bigquery))]
--- a/src/bigquery-derive/src/lib.rs
+++ b/src/bigquery-derive/src/lib.rs
@@ -20,7 +20,7 @@ use proc_macro::TokenStream;
 use quote::quote;
 use syn::{Data, DeriveInput, Fields, parse_macro_input};
 
-/// Derives the `FromRow` trait for a struct.
+/// Derives standard library [`TryFrom`](std::convert::TryFrom) for converting a BigQuery `Row` into a struct.
 ///
 /// Supports renaming attributes via `#[bigquery(rename = "new_name")]`.
 #[proc_macro_derive(FromRow, attributes(bigquery))]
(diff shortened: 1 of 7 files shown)


src/bigquery-derive/src/lib.rs
```diff
@@ -15,3 +15,104 @@
...
+            }
+        }
+
+        impl<'a> std::convert::TryFrom<&'a google_cloud_bigquery::Row> for #name {
+            type Error = google_cloud_bigquery::RowError;
+
+            fn try_from(row: &'a google_cloud_bigquery::Row) -> std::result::Result<Self, Self::Error> {
+                google_cloud_bigquery::FromRow::from_row(row.clone())
+            }
+        }
```
dbolduc: Can we remove this? I don't think we need to offer this variant if the application can `row.clone().try_into()` themselves.

src/bigquery/src/query/row.rs
```diff
@@ -592,4 +600,109 @@ mod tests {
...
+            TableFieldSchema::new()
+                .set_name("some_null")
+                .set_type("INTEGER")
+                .set_mode("NULLABLE"),
+        ]);
+        let schema = Arc::new(Schema::new(schema));
+        let row = Row::try_new(raw_row, &schema)?;
+
+        let err = TestRow::try_from(row).unwrap_err();
+        assert!(matches!(err, RowError::ColumnNotFound(col) if col == "\"custom_int\""));
```
dbolduc: why is the column name `"foo"` instead of `foo` ?

src/bigquery-derive/src/lib.rs
```diff
@@ -15,3 +15,104 @@
...
+    });
+
+    let field_idents = fields
+        .iter()
+        .map(|f| f.ident.as_ref().expect("named field must have identifier"));
+
+    // TODO(#5592): check that the schema and this struct have same columns/attributes count.
+
+    let expanded = quote! {
+        impl google_cloud_bigquery::FromRow for #name {
```
dbolduc: Q: Why introduce a trait which is just a shorthand for `TryFrom<Row>` ?

I don't think we should have both, but I could be missing something.

dbolduc (commented):
LGTM, just have questions on why we need three trait impls instead of one. It seems like only `TryFrom<Row>` is necessary, and the others do not significantly help ergonomics.

src/bigquery-derive/src/lib.rs
alvarowolfx: that's a good point, I ended up introducing that because sqlx and tokio-postgres have equivalent traits. But in the end is just a shortcut to TryFrom<Row>. I'm checking if removing it causes any headache.

src/bigquery-derive/src/lib.rs
alvarowolfx: will remove that.

src/bigquery/src/query/row.rs
alvarowolfx: that's because for building a RowError::ColumnNotFound, `ColumnIndex` can be a number or a string, and  I ended up doing like this:

```
fn resolve_index<I: ColumnIndex>(&self, col: &I) -> Result<usize> {
    col.index(self)
        .ok_or_else(|| RowError::ColumnNotFound(format!("{:?}", col)))
}
```

probably I can use the non debug version to get rid of the quotes.

dbolduc: Ah.

dbolduc (approved):

PR #6120, opened 2026-07-21
impl(bigquery): apply templates for pub(crate) builder fields

haphungw:
Regenerate `google-cloud-bigquery-v2` to surface the internal builder fields as `pub(crate)`, paving the way for handwritten job polling logic.


haphungw:
outdated, closing

PR #6123, opened 2026-07-21
impl(bigquery): add support for Range type

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery/src/query/from_sql.rs
+++ b/src/bigquery/src/query/from_sql.rs
@@ -341,14 +341,15 @@ impl<T: FromSql> FromSql for Range<T> {
         match value {
             wkt::Value::String(s) => {
                 let trimmed = s.trim();
-                if trimmed.len() < 2 {
-                    return Err(ConvertError::Convert(
-                        "invalid range format: too short".into(),
-                    ));
-                }
-
                 // Strip leading [ or ( and trailing ] or )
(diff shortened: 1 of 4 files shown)


src/bigquery/src/query/from_sql.rs
```diff
@@ -330,6 +330,68 @@ impl FromSql for rust_decimal::Decimal {
...
+    /// The exclusive end of the range (or None if unbounded).
+    pub end: Option<T>,
+}
+
+impl<T: FromSql> FromSql for Range<T> {
+    fn from_sql(value: wkt::Value) -> Result<Self, ConvertError> {
+        match value {
+            wkt::Value::String(s) => {
+                let trimmed = s.trim();
+                // Strip leading [ or ( and trailing ] or )
```
dbolduc: Q: is a leading `(` valid? and same q for a trailing `]`

dbolduc (approved):

src/bigquery/src/query/from_sql.rs
alvarowolfx: good catch, I wrote good assertions (with only leading `[` and trailing `)`), but the AI overlords wrote impl for non existing cases and I ended up missing that. I added some extra test cases of invalid ranges with leading `(` and trailing `]`

Issue #5741, opened 2026-05-26
BigQuery samples


dbolduc:
We need to write the samples for BigQuery. We are signing up for:

dbolduc:
These are all done.

PR #6124, opened 2026-07-21
docs(bigquery): add DML and DDL query operation samples

alvarowolfx:
Adds sample code demonstrating Standard SQL DML record updates, logical DDL view creation, stored SQL DDL routine definitions, and schema-expanding destination table row appends for the following region tags:

* `bigquery_update_with_dml`: modifying existing table records using Standard SQL DML `UPDATE`statements
* `bigquery_ddl_create_view`: creating logical SQL views programmatically using DDL statements
* `bigquery_create_routine_ddl`: creating persistent custom SQL routines and user-defined functions using DDL statements
* `bigquery_add_column_query_append`: appending rows to a destination table while introducing new columns using the `ALLOW_FIELD_ADDITION` schema update option

Towards #5741


src/bigquery/examples/src/query/ddl_create_routine.rs
```diff
@@ -0,0 +1,40 @@
...
+
+    // Create a SQL UDF routine using standard SQL DDL
+    let ddl_sql = format!(
+        "CREATE OR REPLACE FUNCTION `{project_id}.{dataset_id}.{routine_id}`(x INT64, y INT64) \
+         RETURNS INT64 \
+         AS (x * y);"
+    );
+
+    client
+        .query(&ddl_sql)
```
dbolduc: nit: 

```suggestion
        .query(ddl_sql)
```

here and below

dbolduc (approved):

src/bigquery/examples/src/query/ddl_create_routine.rs
```diff
@@ -0,0 +1,40 @@
...
+
+// [START bigquery_create_routine_ddl]
+use google_cloud_bigquery::client::BigQuery;
+
+pub async fn sample(project_id: &str, dataset_id: &str, routine_id: &str) -> anyhow::Result<()> {
+    let client = BigQuery::builder().build().await?;
+
+    // Create a SQL UDF routine using standard SQL DDL
+    let ddl_sql = format!(
+        "CREATE OR REPLACE FUNCTION `{project_id}.{dataset_id}.{routine_id}`(x INT64, y INT64) \
```
coryan, line 29: another drive-by... do we want to have examples that can be subject to SQL injection?

alvarowolfx: our samples are only using `format!` for the table identifier (project + dataset + table id) and it's aligned with other language sample (this one for example [here](https://docs.cloud.google.com/bigquery/docs/samples/bigquery-create-routine-ddl#bigquery_create_routine_ddl-go)). For parameters, there are other samples that are using named and positional parameters, which are safer. On top of that, BigQuery doesn't allow for such parameters on resource identifiers iirc, will double check.

src/bigquery/examples/src/query/ddl_create_routine.rs
alvarowolfx: fixed here and other places. Thanks

src/bigquery/examples/src/query/ddl_create_routine.rs
coryan: >  it's aligned with other language sample

The fact that other people are doing it wrong is never a good argument to use with me...  It just summons "particularly grumpy Carlos" as opposed to "normally grumpy Carlos".

> are only using format! for the table identifier (project + dataset + table id) 

So? The risk is something like: ```project_id="your-project.foo.bar` do something innocuous ; do something evil; -- ignore the existing stuff"``` 

note the semicolons, and the backtick ... People can do SQL injection with just that format, and I am not even good at SQL injection.

Not that we expect the examples to do these things, or that this would work in these queries ... but we don't want to steer customers towards text interpolation. If there is a better way to represent "a table name received as an argument" we should be using it (a quick search says that there isn't ... sad). 

If there isn't... maybe the example should say something. Or maybe we should call a `validateInputs` that is not part of the example, something like:

```rust
validateIdentifiers(project_id, dataset_id, routine_id)?; // always check inputs to prevent SQL injections
let ddl_sql = format!(
        "CREATE OR REPLACE FUNCTION `{project_id}.{dataset_id}.{routine_id}`
```

The `validateIdentifiers` can verify the string values are only alphanumeric.

alvarowolfx: Life is too short to be that grumpy, the last paragraph with the idea was enough. I'm well aware of what SQL Injection is about. Thanks for the tip.

src/bigquery/examples/src/query/ddl_create_routine.rs
```diff
@@ -0,0 +1,46 @@
...
+    let client = BigQuery::builder().build().await?;
+
+    // Query parameters cannot be used as substitutes for identifiers, column names or table
+    // names. Ensure resource identifiers are validated before formatting into SQL statements
+    // to prevent SQL injection.
+    validate_resource_names(project_id, dataset_id, routine_id)?;
+
+    // Create a SQL UDF routine using standard SQL DDL
+    let ddl_sql = format!(
+        "CREATE OR REPLACE FUNCTION `{project_id}.{dataset_id}.{routine_id}`(x INT64, y INT64) \
```
coryan, line 29: nit: consider raw strings for these too

coryan (approved):

src/bigquery/examples/src/query/ddl_create_routine.rs
alvarowolfx: will move to raw strings on other samples too, all at once in a separated PR.

PR #6125, opened 2026-07-21
impl(bigquery): add support for byte type

alvarowolfx:
Towards #5844

### DIFF
--- a/src/bigquery/src/query/from_sql.rs
+++ b/src/bigquery/src/query/from_sql.rs
@@ -526,23 +526,29 @@ mod tests {
     }
 
     #[test_case(wkt::Value::String("AQIDBA==".to_string()) => Ok(vec![1, 2, 3, 4]) ; "vec u8 from base64")]
+    #[test_case(wkt::Value::String("".to_string()) => Ok(vec![]) ; "vec u8 from empty base64")]
     #[test_case(wkt::Value::Null => Err(TestConvertError::NotNull) ; "null vec u8")]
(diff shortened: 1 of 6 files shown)


src/bigquery/src/query/from_sql.rs
```diff
@@ -501,6 +525,26 @@ mod tests {
         FromSql::from_sql(value).map_err(TestConvertError::from)
     }
 
+    #[test_case(wkt::Value::String("AQIDBA==".to_string()) => Ok(vec![1, 2, 3, 4]) ; "vec u8 from base64")]
```
coryan, line 528: Can you throw in a test for the empty string too?

coryan: And a test with a string that is not quite a base 64 like `"AQIDBA"` and `"Not a base64 string"`...

coryan (approved):
Please improve the testing a bit and then merge.

src/bigquery/src/query/from_sql.rs
alvarowolfx: > Can you throw in a test for the empty string too?
ack

> And a test with a string that is not quite a base 64 like "AQIDBA" and "Not a base64 string"...

This test is doing that. It's separated to avoid having to match `ConvertError::Convert`  with the full Base64 decode error. 
https://github.com/googleapis/google-cloud-rust/pull/6125/changes#diff-e5149826e7de7c08ad94bfd71fbb320741cc7908d12e62503079f2888a7c12bbR535-R539

But will add more test cases to it and be more explicit that is not a base64 string.

PR #6126, opened 2026-07-21
impl(bigquery): introduce `AppendError`

dbolduc:
Define a custom error to return when we append rows. The error is crate private while we do not have a public client.

We need a custom error. There will at least be another variant for `UnexpectedEndOfStream` for when the stream ends in an error without having responded to some of the requests.

I am sure there will be more variants we will need as we write the implementation.

Part of the work for #5833

### DIFF
--- a/src/bigquery-write/src/error.rs
+++ b/src/bigquery-write/src/error.rs
@@ -1,0 +1,49 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::Error;
+
+/// Represents an error that can occur when appending rows.
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+pub enum AppendError {
+    /// The underlying RPC failed.
+    #[non_exhaustive]
+    #[error("the operation failed. RPC error: {source}")]
+    Rpc {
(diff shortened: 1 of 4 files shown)


src/bigquery-write/src/error.rs
```diff
@@ -0,0 +1,49 @@
...
+#[derive(thiserror::Error, Debug)]
+#[non_exhaustive]
+pub enum AppendError {
+    /// The underlying RPC failed.
+    #[non_exhaustive]
+    #[error("the operation failed. RPC error: {source}")]
+    Rpc {
+        /// The error returned by the service for the request.
+        #[source]
+        source: Error,
```
alvarowolfx, line 27: What is the plan to surface [StorageError](https://docs.cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#google.cloud.bigquery.storage.v1.StorageError)s ? It is contained on the service status details: https://github.com/googleapis/google-cloud-rust/blob/d16877cb48309ae3fdb59fb0d334f016fa785cdb/src/gax/src/error/rpc.rs#L352

The service can attach extra metadata to the RPC errors, that needs to be extracted. Some options that I see: 
* provide a helper that extract from gax::Error 
* a branch on AppendError which contains more information 
* add a []StorageError to the AppendError::Rpc branch.

Non blocking, but wondering how the error story is gonna take shape here.

alvarowolfx (approved):
no blockers, but I have a question on how the AppendError is going to allow customers to extract the service specific information (aka StorageError)

src/bigquery-write/src/error.rs
dbolduc: > What is the plan to surface [StorageError](https://docs.cloud.google.com/bigquery/docs/reference/storage/rpc/google.cloud.bigquery.storage.v1#google.cloud.bigquery.storage.v1.StorageError)s ? It is contained on the service status details:

TIL, thanks for the call out!

At the moment, we are dropping an encoded `StorageError` on the floor when converting from gRPC-generated types to our types. I opened #6127 to fix that. (looks annoying, tbh).

> add a []StorageError to the AppendError::Rpc branch.

I like this. It is convenient to extract this for the application. The branch is `non_exhaustive`, so we could add to it later.

Issue #5831, opened 2026-06-03
Implement basic connection pool (no multiplexing, simple routing)


dbolduc:

dbolduc:
Note to self: Gemini had a simplification to the program flow for stream failures. I should implement it:

https://github.com/googleapis/google-cloud-rust/pull/6663#discussion_r3936215012

PR #6128, opened 2026-07-21
impl(bigquery): task to read from stream

dbolduc:
Add the beginning of a helper to run the stream event loop. At the moment, it handles the first write and opening a stream (but not reading from it at all).

A follow up PR will actually add the event loop that reads from the stream, and accepts more writes. It will also expand on the `success()` test case.

Towards #5831

### DIFF
--- a/src/bigquery-write/src/runner.rs
+++ b/src/bigquery-write/src/runner.rs
@@ -13,7 +13,7 @@
 // limitations under the License.
 
 use crate::Result;
-use crate::error::{AppendError, AppendResult};
+use crate::error::AppendResult;
 use crate::google::cloud::bigquery::storage::v1::{AppendRowsRequest, AppendRowsResponse};
 use crate::stream::Stream;
 use crate::transport::Transport;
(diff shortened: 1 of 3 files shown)


src/bigquery-write/src/runner.rs
```diff
@@ -0,0 +1,191 @@
...
+}
+
+fn process_gax_response(
+    resp_txs: &mut VecDeque<oneshot::Sender<AppendResult<AppendRowsResponse>>>,
+    result: Result<AppendRowsResponse>,
+) {
+    // Pop the response channel associated with this response.
+    let resp_tx = resp_txs
+        .pop_front()
+        .expect("the service sends one response per request");
```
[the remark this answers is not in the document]
dbolduc: This is a valid use of expect. We don't need to be defensive.

src/bigquery-write/src/runner.rs
```diff
@@ -0,0 +1,191 @@
...
+
+    // Forward the result.
+    let resp = result.map_err(|e| AppendError::Rpc { source: e });
```
alvarowolfx: if we add `#[from]` here: https://github.com/googleapis/google-cloud-rust/blob/6334a13575e11f7b8e065168c192930a92824b97/src/bigquery-write/src/error.rs#L27 we can make this code like this:

```suggestion
    let resp = result?;
```

coryan (approved):

alvarowolfx (approved):
All good, just a minor code improvement on the RPC error instantiation, which we used on the BigQuery query client.

src/bigquery-write/src/runner.rs
dbolduc: I don't want to use `?` because I have to send the error on the channel (not return it from the `fn`).

Still, `#[from]` + `e.into()` is slightly cleaner, so thanks!

PR #6131, opened 2026-07-22
feat(storage): object-level checksum in gRPC bidi read

### LINKED ISSUE #6056: feat(storage): object-level checksum validation for gRPC bidi read
**Is your feature request related to a problem?**
The Rust SDK currently performs chunk-level validation for gRPC `BidiReadObject` streams. While this detects wire-level corruption, it doesn't protect against dropped, duplicated, or reordered chunks. We need full object-level validation.

*(Note: This is the gRPC equivalent to sister issue #2049 for JSON reads).*

**Describe the solution you'd like**
Implement cumulative object-level checksum validation (CRC32C and MD5) for the gRPC Bidi stream.

xlai20:
### Description

This PR introduces object-level checksum validation (CRC32C and MD5) for gRPC bidirectional streaming reads (`open_object`).

go/rust-sdk-grpc-bidi-object-checksum, Issue https://github.com/googleapis/google-cloud-rust/issues/6056

#### Key Changes:
* **Checksum Validation Options:** Added `compute_crc32c(bool)` and `compute_md5()` configuration toggles to `open_object`. CRC32C is enabled by default.
* **Resilient Error Handling:** Validating checksums on the foreground reader prevents a mismatch from tearing down the underlying gRPC connection for other multiplexed streams. Additionally, streams now gracefully exhaust themselves on error to prevent infinite polling loops.
* **Testing:** Added comprehensive unit and integration tests, including verification that `content-encoding: gzip` objects successfully validate their CRC32C hashes over gRPC without false mismatches.


src/storage/src/storage/open_object.rs
```diff
@@ -124,6 +124,74 @@ impl<S> OpenObject<S> {
...
+        if enable {
+            self.options
+                .checksum
+                .crc32c
+                .get_or_insert_with(crate::storage::checksum::details::Crc32c::default);
+        } else {
+            self.options.checksum.crc32c = None;
+        }
+        self
+    }
```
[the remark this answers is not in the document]
xlai20: Done.

src/storage/src/storage/bidi/range_reader.rs
```diff
@@ -68,8 +75,30 @@ impl ReadObjectResponse for RangeReader {
...
+            return None;
+        }
+        let msg = self.inner.recv().await;
+        match msg {
+            Some(Ok(b)) => {
+                self.checksum.update(self.offset, &b);
+                self.offset += b.len() as u64;
+                Some(Ok(b))
+            }
+            Some(Err(e)) => Some(Err(Error::io(e))),
```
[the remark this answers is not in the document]
xlai20: Done.

src/storage/src/storage/open_object.rs
```diff
@@ -124,6 +124,74 @@ impl<S> OpenObject<S> {
...
+    ///     .await?;
+    /// let mut contents = Vec::new();
+    /// while let Some(chunk) = reader.next().await.transpose()? {
+    ///     contents.extend_from_slice(&chunk);
+    /// }
+    /// println!("object contents={:?}", contents);
+    /// # Ok(()) }
+    /// ```
+    pub fn compute_md5(self) -> Self {
+        let mut this = self;
```
joshuatants: why not just `pbu fn compute_md5(mut self)`?

src/storage/src/storage/open_object.rs
```diff
@@ -124,6 +124,74 @@ impl<S> OpenObject<S> {
...
+    pub fn compute_crc32c(mut self, enable: bool) -> Self {
+        if !enable {
+            self.options.checksum.crc32c = None;
+            return self;
+        }
+        self.options
+            .checksum
+            .crc32c
+            .get_or_insert_with(crate::storage::checksum::details::Crc32c::default);
+        self
```
joshuatants, line 186: Let's avoid negation unless necessary.

```
if enable {
 self.options
            .checksum
            .crc32c
            .get_or_insert_with(crate::storage::checksum::details::Crc32c::default);
} else {
...
}
```

src/storage/src/storage/bidi/transport.rs
```diff
@@ -94,10 +98,21 @@ impl ObjectDescriptorTransport {
...
         let active = ActiveRead::new(tx, range);
-        let reader = RangeReader::new(rx, object.clone(), requests.clone());
+
+        // Skip checksums for ranged reads.
+        // Note: We do not skip checksums for gzip-encoded objects (decompressive transcoding)
+        // because gRPC automatic object decompression is not supported by the server.
+        // The client receives the raw compressed bytes which match the stored checksum.
+        let is_ranged = !matches!(range, RequestedRange::Offset(0));
+        if is_ranged {
+            checksum = crate::storage::checksum::details::Checksum::default();
```
joshuatants: for my learning: so setting Checksum::default() means that we are skipping the checksum?

src/storage/src/storage/bidi/transport.rs
```diff
@@ -94,10 +98,21 @@ impl ObjectDescriptorTransport {
...
     ) -> (ActiveRead, ReadObjectResponse) {
         let (tx, rx) = tokio::sync::mpsc::channel(100);
         let active = ActiveRead::new(tx, range);
-        let reader = RangeReader::new(rx, object.clone(), requests.clone());
+
+        // Skip checksums for ranged reads.
+        // Note: We do not skip checksums for gzip-encoded objects (decompressive transcoding)
+        // because gRPC automatic object decompression is not supported by the server.
+        // The client receives the raw compressed bytes which match the stored checksum.
+        let is_ranged = !matches!(range, RequestedRange::Offset(0));
```
joshuatants: Instead of passing in `mut checksum` just so that checksum can be reassigned to Checksum::default(), let's use shadowing instead:

```
let checksum = if !matches!(range, RequestedRange::Offset(0)) {
        crate::storage::checksum::details::Checksum::default()
    } else {
        checksum // Drops into the expression if not ranged
    };

    let reader = RangeReader::new(rx, object.clone(), requests.clone(), checksum);
```

Passing in `mut` suggests that you are going to modify checksum itself.

joshuatants (changes requested):
Can this PR be split s.t. that are at most 200 new LOC per PR?

xlai20:
> Can this PR be split s.t. that are at most 200 new LOC per PR?

It is not a good idea to to split this PR because the PR is doing 1 single thing (i.e., adding object-level checksum for gRPC read), and also, most of the changes are tests and comments.

Below is the stats:

1. Production Code (~130 lines total)
    - Executable Code: 70 lines
    - Comments & RustDocs: 54 lines 
    - Blank Lines: 6 lines 
2. Test Code (~361 lines total)

As you can see, the actual logic added to open_object.rs, range_reader.rs, etc. is merely 70 lines of code.

src/storage/src/storage/bidi/transport.rs
xlai20: Yes. Setting `Checksum::default()` entirely disables (skips) checksum computation.

1. **`Checksum::default()` (The top-level configuration):**
   The `Checksum` struct holds `Option` wrappers for the specific algorithms:
   ```rust
   pub struct Checksum {
       pub crc32c: Option<Crc32c>,
       pub md5_hash: Option<Md5>,
   }
   ```
   Because it derives `Default`, calling `Checksum::default()` initializes both fields to `None`. When the `RangeReader` receives chunks of bytes and calls `checksum.update(data)`, it acts as a no-op because the inner algorithms are `None`. At the end of the stream, it returns an empty validation result, effectively bypassing the validation check entirely.

2. **`Crc32c::default()` and `Md5::default()` (The active hashers):**
   Conversely, if we wanted to *enable* a checksum algorithm, we would initialize the inner hasher using its default state, which starts the math at zero.
   ```rust
   pub struct Crc32c {
       checksum: u32, // Starts at 0 via Default
       offset: u64,   // Starts at 0 via Default
   }
   ```
   For example, when a user explicitly calls `.compute_crc32c(true)`, the SDK constructs the object like this:
   `Checksum { crc32c: Some(Crc32c::default()), md5_hash: None }`

src/storage/src/storage/open_object.rs
xlai20: Done. Changed this to a 1-liner that when enable is true, crc32 checksum struct is created and when enable is false, crc32 checksum becomes None.

src/storage/src/storage/bidi/transport.rs
xlai20: Done in updating "mut checksum" to "checksum" in function signature and the code accordingly.

src/storage/src/storage/open_object.rs
xlai20: Done. 

As discussed offline, the original function signature is trying to mimic the same pattern as the read_object API. But `compute_md5(mut self)` is indeed more an idiomatic Rust style, hence updating this function signature based on feedback.

src/storage/src/storage/bidi/transport.rs
```diff
@@ -94,10 +98,22 @@ impl ObjectDescriptorTransport {
...
     ) -> (ActiveRead, ReadObjectResponse) {
         let (tx, rx) = tokio::sync::mpsc::channel(100);
         let active = ActiveRead::new(tx, range);
-        let reader = RangeReader::new(rx, object.clone(), requests.clone());
+
+        // Skip checksums for partial ranged reads.
+        // Note: We do not skip checksums for gzip-encoded objects (decompressive transcoding)
+        // because gRPC automatic object decompression is not supported by the server.
+        // The client receives the raw compressed bytes which match the stored checksum.
+        let checksum = if matches!(range, RequestedRange::Offset(0)) {
```
joshuatants: If the user makes a range call that results in RequestedRange::Segment { 0, 0 }, this will still return the whole object but the checksum won't be calculated.

A more robust detection method is to either convert the RequestadRange into the actual offset and length used in the RPC call (I think this function already exists somewhere in the code), or handle all the possible RequestedRange cases.

src/storage/src/storage/open_object.rs
```diff
@@ -124,6 +124,67 @@ impl<S> OpenObject<S> {
         }
     }
 
+    /// Enables computation of MD5 checksums.
+    ///
+    /// By default, MD5 checksums are disabled, as they are not supported by the GCS gRPC API.
```
joshuatants: If they're not supported by gRPC, and open-object is gRPC, then why offer this setter?

src/storage/src/storage/bidi/range_reader.rs
```diff
@@ -68,8 +75,33 @@ impl ReadObjectResponse for RangeReader {
     }
 
     async fn next(&mut self) -> Option<crate::Result<bytes::Bytes>> {
-        let msg = self.inner.recv().await?;
-        Some(msg.map_err(Error::io))
+        if self.exhausted {
```
joshuatants, line 92: Let's take this opportunity to improve the readability of the code. Let's add:

1. a docstring to the function
2. comments at suitable locations, e.g. logical blocks, to explain what the code does. For example, what would be helpful here is a comment explaining what each match case handles

joshuatants:
> > Can this PR be split s.t. that are at most 200 new LOC per PR?
> 
> It is not a good idea to to split this PR because the PR is doing 1 single thing (i.e., adding object-level checksum for gRPC read), and also, most of the changes are tests and comments.
> 
> Below is the stats:
> 
> 1. Production Code (~130 lines total)
>    
>    * Executable Code: 70 lines
>    * Comments & RustDocs: 54 lines
>    * Blank Lines: 6 lines
> 2. Test Code (~361 lines total)
> 
> As you can see, the actual logic added to open_object.rs, range_reader.rs, etc. is merely 70 lines of code.

I review tests too. Perhaps for PRs like these we can have the logic mainly in 1 PR and tests in another.

Let's do that for future PRs. I reviewed the ones in this one already.

src/storage/src/storage/bidi/range_reader.rs
```diff
@@ -143,4 +180,152 @@ mod tests {
...
+
+        // Subsequent calls should yield None
+        let got_none = reader.next().await;
+        assert!(got_none.is_none(), "{got_none:?}");
+
+        Ok(())
+    }
+
+    #[tokio::test]
+    async fn object_with_checksum_bypassed_due_to_ranged_read() -> anyhow::Result<()> {
```
joshuatants: Is it possible to have the test enable a checksum and then perform a ranged read to verify that no checksum is calculated?

Right now, what the test verifies is that if Checksum::default is passed in, then no validation is performed. Which is a valuable test, but not what the name suggests.

src/storage/src/storage/open_object.rs
```diff
@@ -800,6 +926,30 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
```
joshuatants: Doesn't need to be async

src/storage/src/storage/bidi/range_reader.rs
```diff
@@ -68,8 +75,33 @@ impl ReadObjectResponse for RangeReader {
...
+                self.offset += b.len() as u64;
+                Some(Ok(b))
+            }
+            Some(Err(e)) => {
+                self.exhausted = true;
+                Some(Err(Error::io(e)))
+            }
+            None => {
+                self.exhausted = true;
+                if let Some(expected) = &self.object.checksums {
```
joshuatants, line 115: To double check: if it's possible for object.checksums to be None, then I think that in that case we should skip calling self.checksum.update in order to save computation

src/storage/src/storage/open_object.rs
```diff
@@ -726,6 +787,71 @@ mod tests {
...
+
+        let mut got_payload = Vec::new();
+        let mut got_err = false;
+        while let Some(res) = reader.next().await {
+            match res {
+                Ok(chunk) => got_payload.extend_from_slice(&chunk),
+                Err(e) => {
+                    got_err = true;
+                    let fmt = format!("{e:?}");
+                    assert!(fmt.contains("ChecksumMismatch"), "{fmt}");
```
joshuatants: Instead of checking the error string, can the error itself be checked like what is done in object_with_checksum_mismatch?

joshuatants (changes requested):

src/storage/src/storage/bidi/range_reader.rs
xlai20: Done.
Have Added: 
1. A detailed docstring to range_reader::next() function
2. Comment on each logic branch of the msg match case to explain what the code block is handling.

src/storage/src/storage/bidi/transport.rs
xlai20: Done.
Using a more robust method to directly checking the offset/length used in the rpc call to evaluate it is a full object read. (According to gRPC proto documentation, when both are 0, it is a full object read)
Re-used the existing function RequestedRange::as_proto() to do this conversion.

src/storage/src/storage/open_object.rs
xlai20: Md5 is supported in gRPC; the comment is wrong. Updated the comment.

src/storage/src/storage/bidi/range_reader.rs
xlai20: Done.

1. Renamed this test to `object_with_default_checksum_skips_all_validation()` to purposely test that when a default empty Checksum is passed in to RangeReader, the checksum validation will be bypassed.
2. Added another test `send_and_read_checksum_bypassed_due_to_ranged_read()` to `open_object.rs` to purposely test on a client level that when crc32c checksum is enabled by default,  the checksum validation is still bypassed in a ranged read.

src/storage/src/storage/open_object.rs
xlai20: Done.

Changed to `#[test]`

src/storage/src/storage/bidi/range_reader.rs
xlai20: Done.

Yes, sometimes the object.checksums sent by server can be None (e.g., composite objects), so in this case, it would be waste to do cumulative checksum update. I've added a if block in `RangeReader::new` function to check if the object.checksums is present, if not, CheckSum struct will be set to default empty even though user enables it, effectively avoiding unnecessary checksum calculation.

src/storage/src/storage/open_object.rs
xlai20: Done.

I've updated the test assert statement in `send_and_read_checksum_mismatch()` test to become `assert!(
                        matches!(source, Some(crate::error::ReadError::ChecksumMismatch(_))),
                        "Expected ChecksumMismatch, got: {source:?}"
                    );`. In this way, it is asserting against the struct type of mismatch instead of an error string.

joshuatants (approved):

PR #6137, opened 2026-07-22
impl(bigquery): add support for Interval type

alvarowolfx:
Towards #5844


src/bigquery/src/query/from_sql.rs
```diff
@@ -330,6 +335,106 @@ impl FromSql for rust_decimal::Decimal {
...
+                let time_str = parts.next();
+                let extra = parts.next();
+
+                let (ym_str, days_str, time_str) = match (ym_str, days_str, time_str, extra) {
+                    (Some(ym), Some(d), Some(t), None) => (ym, d, t),
+                    _ => {
+                        let count = s.split_whitespace().count();
+                        return Err(ConvertError::Convert(
+                            format!("invalid interval format: expected 3 parts, got {count}")
+                                .into(),
```
dbolduc: I think displaying `{s}` would be more helpful than `{count}`

src/bigquery/src/query/from_sql.rs
```diff
@@ -330,6 +335,106 @@ impl FromSql for rust_decimal::Decimal {
     }
 }
 
+/// Represents a BigQuery TIME INTERVAL value.
```
dbolduc: consider linking to https://docs.cloud.google.com/bigquery/docs/reference/standard-sql/data-types#interval_type

and (not in this PR) maybe do a pass on the other non-trivial types?

src/bigquery/src/query/from_sql.rs
```diff
@@ -330,6 +335,106 @@ impl FromSql for rust_decimal::Decimal {
...
+#[derive(Clone, Debug, Default, PartialEq)]
+pub struct Interval {
```
dbolduc, line 342: thought for later: might be nice to have conversion helpers for `wkt::Duration`.

another thought for later: are we going to impl mathy traits like `Add`: https://doc.rust-lang.org/std/ops/trait.Add.html

I will answer my own questions and say: let's not do this any time soon. It is a trivial task to do later if we are asked to.

dbolduc (approved):

PR #6143, opened 2026-07-22
impl(bigquery): add write stream event loop

dbolduc:
Towards #5831

### DIFF
--- a/src/bigquery-write/src/runner.rs
+++ b/src/bigquery-write/src/runner.rs
@@ -17,10 +17,15 @@ use crate::error::{AppendError, AppendResult};
 use crate::google::cloud::bigquery::storage::v1::{AppendRowsRequest, AppendRowsResponse};
 use crate::stream::Stream;
 use crate::transport::Transport;
+use gaxi::grpc::from_status::to_gax_error;
+use gaxi::grpc::tonic::Status as TonicStatus;
+use gaxi::grpc::tonic::Streaming;
 use std::collections::VecDeque;
 use std::sync::Arc;
 use tokio::sync::{mpsc, oneshot};
 use tokio::task::JoinHandle;
+
+type TonicResult<T> = std::result::Result<T, TonicStatus>;
 
 #[derive(Debug)]
 pub(crate) struct WriteRequest {
@@ -69,15 +74,55 @@ async fn run_stream_task(inner: Arc<Transport>, mut req_rx: mpsc::Receiver<Write
     resp_txs.push_back(initial_req.resp_tx);
 
     // Open the stream.
-    let Stream { stream, request_tx } = match Stream::new(inner, initial_req.req).await {
+    let Stream {
+        mut stream,
+        request_tx,
+    } = match Stream::new(inner, initial_req.req).await {
         Ok(s) => s,
         Err(e) => {
             process_gax_response(&mut resp_txs, Err(e));
             return;
         }
     };
-    // TODO(#5831) - implement stream event loop
-    let _ = (stream, request_tx);
+
+    loop {
+        tokio::select! {
+            req = req_rx.recv() => {
+                match req {
+                    Some(r) => {
+                        // Keep track of the response channel.
+                        resp_txs.push_back(r.resp_tx);
+
+                        // Forward the request to the stream.
+                        let _ = request_tx.send(r.req).await;
+                    }
+                    None => break drain_stream(stream, resp_txs).await,
+                }
+            }
+            resp = stream.message() => {
+                match resp.transpose() {
+                    Some(r) => process_response(&mut resp_txs, r),
+                    None => break,
+                }
+            }
+        }
+    }
+}
+
+async fn drain_stream(
+    mut stream: Streaming<AppendRowsResponse>,
+    mut resp_txs: VecDeque<oneshot::Sender<AppendResult<AppendRowsResponse>>>,
+) {
+    while let Some(r) = stream.message().await.transpose() {
+        process_response(&mut resp_txs, r);
+    }
+}
+
+fn process_response(
+    resp_txs: &mut VecDeque<oneshot::Sender<AppendResult<AppendRowsResponse>>>,
+    resp: TonicResult<AppendRowsResponse>,
+) {
+    process_gax_response(resp_txs, resp.map_err(to_gax_error))
 }
 
 fn process_gax_response(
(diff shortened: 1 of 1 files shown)


src/bigquery-write/src/runner.rs
```diff
@@ -69,15 +74,55 @@ async fn run_stream_task(inner: Arc<Transport>, mut req_rx: mpsc::Receiver<Write
...
+                    Some(r) => {
+                        // Keep track of the response channel.
+                        resp_txs.push_back(r.resp_tx);
+
+                        // Forward the request to the stream.
+                        let _ = request_tx.send(r.req).await;
+                    }
+                    None => break drain_stream(stream, resp_txs).await,
+                }
+            }
```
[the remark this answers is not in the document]
dbolduc: Hm, the effect of the code is almost exactly the same, but this is uglier.

In both cases we are ultimately dropping the `WriteRequest`, and the `resp_tx` is satisfied with a `RecvError`.

src/bigquery-write/src/runner.rs
```diff
@@ -69,15 +74,55 @@ async fn run_stream_task(inner: Arc<Transport>, mut req_rx: mpsc::Receiver<Write
...
+            resp = stream.message() => {
+                match resp.transpose() {
+                    Some(r) => process_response(&mut resp_txs, r),
+                    None => break,
+                }
+            }
```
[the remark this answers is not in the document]
dbolduc: > race conditions where new requests are accepted on a dead stream

The effect is the same.

src/bigquery-write/src/runner.rs
```diff
@@ -69,15 +74,55 @@ async fn run_stream_task(inner: Arc<Transport>, mut req_rx: mpsc::Receiver<Write
...
+}
+
+async fn drain_stream(
+    mut stream: Streaming<AppendRowsResponse>,
+    mut resp_txs: VecDeque<oneshot::Sender<AppendResult<AppendRowsResponse>>>,
+) {
+    while let Some(r) = stream.message().await.transpose() {
+        process_response(&mut resp_txs, r);
+    }
+}
```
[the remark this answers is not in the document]
dbolduc: no need.

src/bigquery-write/src/runner.rs
```diff
@@ -96,6 +141,10 @@ fn process_gax_response(
 #[cfg(test)]
 mod tests {
     use super::*;
+    use crate::google::cloud::bigquery::storage::v1::append_rows_response::{
+        AppendResult, Response,
+    };
+    use crate::transport::tests::*;
     use crate::transport::tests::*;
```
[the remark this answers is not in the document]
dbolduc: Fixed.

dbolduc:
Gemini does not like that I rely on tonic returning a `None` after a `Some(Err(e))`, but I am fine with this.

src/bigquery-write/src/runner.rs
```diff
@@ -69,15 +73,56 @@ async fn run_stream_task(inner: Arc<Transport>, mut req_rx: mpsc::Receiver<Write
...
+    loop {
+        tokio::select! {
+            req = req_rx.recv() => {
+                match req {
+                    Some(r) => {
+                        // Keep track of the response channel.
+                        resp_txs.push_back(r.resp_tx);
+
+                        // Forward the request to the stream.
+                        let _ = request_tx.send(r.req).await;
```
alvarowolfx, line 96: one thing that I'm still not following (and maybe not seeing the big picture), is how retries are going to work here (specially if the stream closes by some service error). Ideally we need to put that WriteRequest back on the queue, reconnect, and resend all queue requests. We don't have a queue for WriteRequest yet, maybe is going to hold by a layer that is going to hold this Runner ?

src/bigquery-write/src/runner.rs
```diff
@@ -180,11 +257,199 @@ mod tests {
...
+        };
+        req_tx.send(write3).await?;
+
+        // resp 1
+        response_tx.send(Ok(convert(&test_response(1)))).await?;
+        let resp1 = resp_rx1.await??;
+        assert_eq!(resp1, test_response(1));
+
+        // Drop the request sender before all the writes are finished.
+        drop(req_tx);
```
alvarowolfx, line 365: is this essentially closing the stream from the Client side ? How the other writes still work and don't return an Error (like stream closed) ? In the real world scenario, we can't guarantee that writes 2 and 3 are going to arrive on the server right ?

Maybe I need to learn more about tokio, tokio::select and other things 😬

alvarowolfx: oh, I see now that this actually happens on test `unexpected_end_of_stream`

alvarowolfx (approved):
overall looks good to me, but I still need to learn more tokio, mpsc and etc. I see a lot of things are closer to Go channels, but the mapping still not written that well in my mind.

But I do have questions on how later down the line, we are going to be able to resend requests that were inflight and the stream was closed unexpectedly

src/bigquery-write/src/runner.rs
dbolduc: Fair question, and I can only give you the plan. I don't have an implementation to share.

My plan is for the retry logic to live outside of this layer. I am pretty sure it will live in the writers.

Here: I want one stream per task. (we will only retry when opening the stream, not after a stream errors)

The caller should be holding onto the `resp_rx` oneshot. If the stream errors, this future will contain an error (whether an `Err(RecvError)` or an `Ok(Err(AppendError))`). And the caller would decide to retry (or bubble up the error, if the caller is not the writer).

So the writer is holding onto a copy of the `AppendRowsRequest`, and in the case of ordered writes, would maintain that queue.

---

It makes some sense in my head, but we will see if I can actually turn it into code. :p

alvarowolfx: that makes sense, that helps build the mental model to help on other PR reviews too. Thanks

src/bigquery-write/src/runner.rs
dbolduc: > is this essentially closing the stream from the Client side ?

Sort of. It is saying no more write requests are coming.

For simplex streams, the request sender is dropped when the writer is dropped, which could maybe happen while there are outstanding writes.

For a multiplex stream, the implementation is less obvious to me, but I assume the thing responsible for multiplexing holds a sender. And so that channel will stay open until there is some stream error (i.e. it is only ever closed from the receiver side).

If not that, then the request sender is closed when all N writers are dropped. This would have the side effect of forcing us to recreate streams more often, but having fewer idle streams? (I think that's a minor point. I think either impl is fine with me).

> How the other writes still work and don't return an Error (like stream closed) ?

Because we remember to drain the stream.

> In the real world scenario, we can't guarantee that writes 2 and 3 are going to arrive on the server right ?

Correct. These just get errors, which we test in `unexpected_end_of_stream`, as you pointed out

PR #6146, opened 2026-07-22
chore: prepare to publish BQ v2

dbolduc:
Let's start with an unstable version, just to be conservative.


librarian.yaml
```diff
@@ -511,11 +511,10 @@ libraries:
     version: 1.11.0
     copyright_year: "2025"
   - name: google-cloud-bigquery-v2
-    version: 1.0.0
+    version: 0.0.1
```
coryan: Maybe too extreme? 0.1 would be fine, or matching the version for the veneer?

🤷

coryan (changes requested):
I think you should match the `0.x` from the veneer.

librarian.yaml
dbolduc: @alvarowolfx informed me that `0.0.1` is a bad version. I went with `0.1.0`.

dbolduc: I don't think we should tie the version of this crate to the veneer. We can and probably will GA this crate first.

coryan (approved):

PR #6154, opened 2026-07-23
impl(bigquery): implement job-level retries

haphungw:
Implement job-level retries for BigQuery job execution within `JobPoller::until_done()`.

When a BigQuery job fails with a transient error reason (`jobBackendError`, `jobInternalError`, `jobRateLimitExceeded`, etc), the poller automatically resubmits the job with exponential backoff up to the configured retry limit.

### DIFF
--- a/src/generated/cloud/bigquery/v2/src/operation.rs
+++ b/src/generated/cloud/bigquery/v2/src/operation.rs
@@ -38,6 +38,33 @@ impl google_cloud_lro::internal::DiscoveryOperation for Job {
     }
 }
 
+/// Determines if a BigQuery job failure reason is transient and eligible for job-level retry.
+///
+/// Returns `true` for retryable reasons (`jobBackendError`, `jobInternalError`, `jobRateLimitExceeded`,
+/// `tableUnavailable`) per BigQuery error handling specification.
+pub(crate) fn is_retryable_job_error(reason: &str) -> bool {
+    matches!(
+        reason,
+        "jobBackendError" | "jobInternalError" | "jobRateLimitExceeded" | "tableUnavailable"
+    )
+}
+
+/// Prepares a `Job` instance for retry by assigning a new synthetic job ID
+/// and clearing existing execution status.
+///
+/// To preserve idempotency and avoid job execution collisions, each job-level retry must
+/// use a unique job ID while retaining original reference details (project ID, location)
+/// and configuration settings.
+pub(crate) fn prepare_job_for_retry(mut job: Job) -> Job {
+    let existing_ref = job.job_reference.unwrap_or_default();
+    job.job_reference = Some(crate::model::JobReference {
+        job_id: uuid::Uuid::new_v4().to_string(),
+        ..existing_ref
+    });
+    job.status = None;
+    job
+}
+
 #[cfg(test)]
 mod tests {
     use super::*;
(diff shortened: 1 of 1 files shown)


src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -183,8 +351,34 @@ impl JobPoller {
...
+                let retry_state = google_cloud_gax::retry_state::RetryState::new(true)
+                    .set_start(start_time)
+                    .set_attempt_count(attempts as u32);
+                let delay = backoff.on_failure(&retry_state);
+                tokio::time::sleep(delay).await;
+                continue;
+            }
+            return Ok(job_result);
+        }
     }
```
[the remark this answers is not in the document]
haphungw: done

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -125,6 +153,143 @@ mod tests {
...
+                    reason: "jobBackendError".to_string(),
+                    message: "backend failed".to_string(),
+                    ..Default::default()
+                }),
+                ..Default::default()
+            }),
+            ..Default::default()
+        };
+
+        let retried_job = prepare_job_for_retry(&original_job);
```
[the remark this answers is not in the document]

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -38,6 +38,33 @@ impl google_cloud_lro::internal::DiscoveryOperation for Job {
...
+///
+/// To preserve idempotency and avoid job execution collisions, each job-level retry must
+/// use a unique job ID while retaining original reference details (project ID, location)
+/// and configuration settings.
+pub(crate) fn prepare_job_for_retry(mut job: Job) -> Job {
+    let existing_ref = job.job_reference.unwrap_or_default();
+    job.job_reference = Some(crate::model::JobReference {
+        job_id: uuid::Uuid::new_v4().to_string(),
+        ..existing_ref
+    });
```
dbolduc: nit:

```suggestion
    job.job_reference.get_or_insert_default().job_id = uuid::Uuid::new_v4().to_string();
```

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -183,8 +350,34 @@ impl JobPoller {
...
     pub async fn until_done(self) -> google_cloud_gax::Result<Job> {
-        // Scaffolding: just pass through to standard poller for now
-        self.builder.poller().until_done().await
+        let mut attempts = 0;
+        let mut builder = self.builder;
+        let backoff = self.policy.backoff;
+        let start_time = std::time::Instant::now();
+        use google_cloud_gax::backoff_policy::BackoffPolicy;
+
+        loop {
```
dbolduc: this is complicated code. we need unit tests for it.

Did you want to include it in this PR? we could start with the helpers, and add the main implementation and all of its tests in a follow up

dbolduc: Advice on testing:

- Make a client with a mock stub
- Return a job that is `DONE` on the initial `InsertJob` RPC. (i.e. don't do any polling. We can assume the poller works).

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -125,6 +152,143 @@ mod tests {
...
+                state: "DONE".to_string(),
+                error_result: Some(ErrorProto {
+                    reason: "jobBackendError".to_string(),
+                    message: "backend failed".to_string(),
+                    ..Default::default()
+                }),
+                ..Default::default()
+            }),
+            ..Default::default()
+        };
```
dbolduc: use the fluent setters for message types. All of the `to_string()`s go away.

dbolduc: here and below.

dbolduc (changes requested):

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -38,6 +38,33 @@ impl google_cloud_lro::internal::DiscoveryOperation for Job {
     }
 }
 
+/// Determines if a BigQuery job failure reason is transient and eligible for job-level retry.
+///
+/// Returns `true` for retryable reasons (`jobBackendError`, `jobInternalError`, `jobRateLimitExceeded`,
+/// `tableUnavailable`) per BigQuery error handling specification.
```
dbolduc: nit: try to flow the lines to 80 characters. We don't have any linters enforcing this, so not a big deal.

dbolduc (commented):
Also Q: have we caught one of these job-level errors in the wild? It would be nice to catch one and see that our code handles it.

"Catching one" looks like running a bajillion jobs until something fails.

src/generated/cloud/bigquery/v2/src/operation.rs
coryan: AFAIK, `cargo fmt` wraps lines over 100 characters (not 80). So I would argue that we have a linter. I saw no reason to override that default.

haphungw: TIL

src/generated/cloud/bigquery/v2/src/operation.rs

src/generated/cloud/bigquery/v2/src/operation.rs
haphungw: I did plan for a mock stub in a later PR. 
I decided to group the detail implementation of `until_done()` with that in the next PR.

src/generated/cloud/bigquery/v2/src/operation.rs

dbolduc (approved):

haphungw:
> Also Q: have we caught one of these job-level errors in the wild? It would be nice to catch one and see that our code handles it.
> 
> "Catching one" looks like running a bajillion jobs until something fails.

I tried running concurrent jobs against live GCP to catch a `jobBackendError`. As expected, high request volume mostly just ran into API Gateway `HTTP 403 RATE_LIMIT_EXCEEDED`. 

Edge proxy rate limits kick in before transient server-side worker drops mid-query, load testing wasn't able to trigger job-level status errors on demand.

Any suggestion?

dbolduc:
> Any suggestions

Just to run an application for days on end.

Don't worry about it though. This will happen naturally when we start benchmarking

PR #6157, opened 2026-07-24
feat(gax): implement grpc-rust outbound request streaming

joshuatants:
For #5991 

Integration tests covering the happy path cases will be added in an upcoming PR.

### DIFF
--- a/src/gax-internal/src/grpc/grpc_rust/send.rs
+++ b/src/gax-internal/src/grpc/grpc_rust/send.rs
@@ -1,0 +1,285 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use bytes::{Buf, Bytes};
+use grpc::client::{SendOptions, SendStream};
+use grpc::core::SendMessage;
+use prost::Message;
+
+/// A [`SendMessage`] adapter that encodes Prost protobuf messages for `grpc-rust`.
+#[derive(Debug)]
+pub struct GrpcRustSend<T>(pub T);
+
+/// Implementation of [`SendMessage`] so that [`GrpcRustSend`] knows how to encode a [`prost::Message`] into raw protobuf bytes.
+impl<T> SendMessage for GrpcRustSend<T>
+where
+    T: prost::Message,
+{
+    fn encode(&self) -> std::result::Result<Box<dyn Buf + Send + Sync>, String> {
+        Ok(Box::new(Bytes::from(self.0.encode_to_vec())))
+    }
+}
+
+/// A handle for the background task pumping outbound requests to a `grpc-rust` [`SendStream`].
+///
+/// This background pump is required because [`SendStream::send`](grpc::client::SendStream::send) is not cancellation-safe.
+/// If an outer caller cancels an async operation (e.g., via a timeout, `tokio::select!`, or dropping a stream early) that directly
+/// uses `SendStream::send`, the stream could become corrupted. This might corrupt the entire HTTP/2 connection,
+/// causing all other concurrent RPCs multiplexed over that same connection to fail or disconnect.
+///
+/// Isolating `SendStream` in a dedicated background task ensures `send` calls run to completion, while aborting the task
+/// on drop cleanly releases the owned [`SendStream`]. Because `SendStream` is owned by the background loop,
+/// `send()` is never interrupted mid-call by outer task drops.
+pub(super) struct SendTask {
+    // The handle is wrapped in an `Option` to ensure it is not aborted twice if the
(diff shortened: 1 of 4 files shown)


src/gax-internal/src/grpc/grpc_rust/send.rs
```diff
@@ -0,0 +1,285 @@
...
+    pub(super) async fn join(&mut self) -> tonic::Result<()> {
+        // Await the background task to finish before taking it so that if `join()` is cancelled
+        // mid-await, `self.handle` remains `Some(...)` and `SendTask::drop()` can still abort the
+        // task.
+        //
+        // If we were to `take()` the handle before awaiting it, a cancellation during `join()`
+        // would leave `self.handle` as `None`, preventing `drop()` from aborting the handle and
+        // causing the background task to become an orphaned "ghost" task.
+        let Some(handle) = self.handle.as_mut() else {
+            return Ok(());
```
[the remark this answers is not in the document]
joshuatants: SendTask::join is intended to be used together with SendTask::is_joinable in a select statement like:

```
tokio::select! {
 // Inbound branch here
...
// Outbound branch
result = send_task.join(), if send_task.is_joinable() => {...}
}
```

Don't foresee join() being called twice.

src/gax-internal/src/grpc/grpc_rust/send.rs
```diff
@@ -0,0 +1,285 @@
...
+        // Assert
+        assert!(result.is_ok(), "finished cancelled task should return Ok");
+        assert!(
+            !task.is_joinable(),
+            "task should no longer be joinable after join"
+        );
+        assert!(task.handle.is_none(), "handle should be cleared after join");
+        drop(task); // Shouldn't panic
+        Ok(())
+    }
```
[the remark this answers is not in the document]
joshuatants: Addressed above.

src/gax-internal/src/grpc/grpc_rust/send.rs
```diff
@@ -0,0 +1,285 @@
...
+    Request: Message + 'static,
+{
+    use tokio_stream::StreamExt;
+
+    tokio::pin!(requests);
+    while let Some(request) = requests.next().await {
+        send.send(&GrpcRustSend(request), SendOptions::default())
+            .await
+            .map_err(|_| tonic::Status::internal(SendTask::ERROR_MESSAGE_STREAM_CLOSED))?;
+    }
```
vsharonlynn, line 129: Here we always use `SendOptions::default()`.
IIUC [SendOptions](https://github.com/grpc/grpc-rust/blob/ccc36e6e416b054434847a8cfc8060e5e7396998/grpc/src/client/mod.rs#L274-L279) has a `final_message` bool.
Do we need to send a `SendOptions` with final_message false to end the stream?

src/gax-internal/src/grpc/grpc_rust/send.rs
```diff
@@ -0,0 +1,285 @@
...
+        };
+        let result = handle.await;
+        self.handle.take();
+        match result {
+            // Task ran to completion without encountering send errors.
+            Ok(Ok(())) => Ok(()),
+            // Task completed, but send loop returned a status error (e.g., stream closed).
+            Ok(Err(status)) => Err(status),
+            // Task was cleanly cancelled (e.g., overall RPC finished or dropped).
+            Err(error) if error.is_cancelled() => Ok(()),
```
vsharonlynn: I'm trying to understand the cancellation error here, and still wrapping my head around many concepts including Rust borrow checker...

Is the only way for this task to be aborted is if the `SendTask` is dropped? I see a custom Drop implementation for SendTask [below](https://github.com/googleapis/google-cloud-rust/pull/6157/changes#diff-e0b00744579684d3a2019bdc1b6419b8857d059d8ae8b7a035a6512cbe17b361R108).

This function `join(&mut self)` uses a mutabl reference, so when `join()` is using this, the compiler will not allow anything else drop or move it.
So, we can't drop `SendTask` when we are waiting for `join()` to finish.
Because we can't drop it, `handle.abort()` will not be called.
Because `handle.abort()` is not called, there can't be a cancellation error.

So is this arm possible to reach?

src/gax-internal/src/grpc/grpc_rust/send.rs
joshuatants: No, based on https://github.com/grpc/grpc-rust/blob/ccc36e6e416b054434847a8cfc8060e5e7396998/grpc/src/client/mod.rs#L228-L235 , simply dropping the SendStream is enough to close the stream properly.

what `final_msg` allows is closing the stream when the SendStream is not owned by the caller; in this case, we own the SendStream, so we can drop it

furthermore, since we are pulling msgs off requests.next(), unless we peek ahead, we can't actually set final_msg correctly

src/gax-internal/src/grpc/grpc_rust/send.rs
joshuatants: join() can be cancelled. For example, if it is used in a select statement like:

```
 tokio::select! {
                response = responses.recv() => match response {
                ...
                },
                status = self.send_task.join(), if self.send_task.is_joinable() => {
                 ...
                    }
                }
            }
````

When join() is cancelled it releases the mutable ref. It is then possible for `drop()` to be called.

vsharonlynn:
LGTM.

src/gax-internal/src/grpc/grpc_rust/send.rs
joshuatants: However,  given how the code is written, it's impossible to reach this arm via a drop, because you're right that after the SendTask is dropped, we can't call join() again.

Removing the arm and associated tests.

PR #6169, opened 2026-07-27
test(storage): test open_object single request/response

joshuatants:


src/storage/tests/open_object_grpc.rs
```diff
@@ -34,6 +35,95 @@ const OBJECT_CONTENT: &[u8] = b"0123456789abcdefghijklmnopqrstuvwxyzABCDEFGHIJKL
 const ERR_STREAM_CLOSED_PREMATURELY: &str = "gRPC stream closed before the request was received";
 const ERR_RECV_ERROR: &str = "error while reading the request";
 
+#[tokio::test]
+async fn send_and_read_single_response_success() -> anyhow::Result<()> {
```
xlai20, line 39: Just to double confirm, will this test be run in the CI/CD set-up for both situations, before and after the gRPC-Rust migration?

xlai20 (approved):

vsharonlynn:
LGTM.

src/storage/tests/open_object_grpc.rs
joshuatants: Yes it will.

Issue #6104, opened 2026-07-20
Timeouts caused by parallel resource cleanup


coryan:
All / most of the integration tests and samples for storage cleanup stale resources (objects, buckets, etc). If the timing is just right, all of them run simultaneously and try to delete the same resources.

The cleanup "succeeds" because any errors caused by duplicate deletes are ignored. However, the cleanup may be very slow, leading to timeouts such as the one in #6099.

Maybe we need to stop the cleanup after X seconds, to prevent slowing down things. Or maybe we should randomize the order in which we delete things, so simultaneous cleanups have a smaller chance of clashes.

Or maybe a combination of both things.

PR #6189, opened 2026-07-28
fix(storage): bound stale bucket cleanup

joshuatants:
For #6104.

Switch from AlwaysRetry to RetryableErrors to avoid retrying permanent cleanup failures. Also cap stale bucket cleanup at 5 minutes.

### DIFF
--- a/src/storage/examples/src/lib.rs
+++ b/src/storage/examples/src/lib.rs
@@ -666,7 +666,7 @@ async fn create_bucket_kms_key(
 pub async fn create_test_bucket() -> anyhow::Result<(StorageControl, Bucket)> {
     let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
     let client = client_for_create_bucket().await?;
-    cleanup_stale_buckets(&client, &project_id).await?;
+    cleanup_stale_buckets(&client, &project_id).await;
 
     let bucket_id = crate::random_bucket_id();
 
@@ -691,7 +691,7 @@ pub async fn create_test_bucket() -> anyhow::Result<(StorageControl, Bucket)> {
 pub async fn create_test_hns_bucket() -> anyhow::Result<(StorageControl, Bucket)> {
     let project_id = std::env::var("GOOGLE_CLOUD_PROJECT")?;
     let client = client_for_create_bucket().await?;
-    cleanup_stale_buckets(&client, &project_id).await?;
+    cleanup_stale_buckets(&client, &project_id).await;
 
     let bucket_id = crate::random_bucket_id();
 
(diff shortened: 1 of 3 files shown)


src/storage/examples/src/lib.rs
```diff
@@ -770,7 +772,38 @@ pub async fn cleanup_stale_buckets(
...
+        }
+        Err(_) => {
+            tracing::warn!(
+                "stale bucket cleanup timed out after {:.1}s; continuing",
+                timeout.as_secs_f64()
+            );
+            Ok(())
+        }
+    }
+}
```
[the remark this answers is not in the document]
joshuatants: Fixed

src/storage/examples/src/lib.rs
```diff
@@ -1056,4 +1089,25 @@ mod tests {
...
+    #[tokio::test]
+    async fn stale_bucket_cleanup_stops_at_timeout() {
+        let result = run_stale_bucket_cleanup(
+            Duration::from_millis(1),
+            std::future::pending::<anyhow::Result<()>>(),
+        )
+        .await;
+
+        assert!(result.is_ok(), "{result:?}");
+    }
```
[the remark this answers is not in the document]
joshuatants: Fixed

src/storage/examples/src/lib.rs
```diff
@@ -38,6 +38,8 @@ use google_cloud_test_utils::resource_names::random_bucket_id;
 use google_cloud_wkt::FieldMask;
 use std::time::Duration;
 
+const STALE_BUCKET_CLEANUP_TIMEOUT: Duration = Duration::from_secs(30);
```
coryan: That seems fairly short considering (a) any retry needs to backoff for 2 seconds, (b) you may need to delete some objects too.

🤷  you would know better.

src/storage/examples/src/lib.rs
```diff
@@ -757,7 +759,7 @@ async fn client_for_create_bucket() -> anyhow::Result<StorageControl> {
                 .unwrap(),
         )
         .with_retry_policy(
-            google_cloud_gax::retry_policy::AlwaysRetry
+            RetryableErrors
                 .with_attempt_limit(16)
                 .with_time_limit(Duration::from_secs(32)),
```
coryan: Note that a single RPC is given 32 seconds, and yet the overall cleanup is supposed to take at most 30. Something is not consistent.

coryan (approved):
I would think the overall timeout needs to be longer, maybe as much as 5 minutes.  Your call.

src/storage/examples/src/lib.rs
joshuatants: Switched to 5 mins. Let's see how it performs.

src/storage/examples/src/lib.rs
joshuatants: Set a max of 5 retries and removed the timeout here.

PR #6190, opened 2026-07-28
feat(gax): implement grpc-rust bidi stream lifecycle

joshuatants:
For #5991

### DIFF
--- a/src/gax-internal/src/grpc/grpc_rust/bidi.rs
+++ b/src/gax-internal/src/grpc/grpc_rust/bidi.rs
@@ -88,6 +88,7 @@ where
         if let Some(message) = self.pending_response.take() {
             return Ok(Some(message));
         }
+        // If the response stream is absent, signal end-of-stream.
         let Some(responses) = self.responses.as_mut() else {
             return Ok(None);
         };
@@ -95,6 +96,11 @@ where
         // Check for inbound responses or the status of the outbound send.
         loop {
             tokio::select! {
+                // Prioritize the response arm: a send failure may result
+                // from the server closing after sending its final response,
+                // and we want to inspect that response to know what the
+                // error is.
+                biased;
                 // Get the next response from the background receive task.
                 response = responses.recv() => match response {
                     // Successfully received a response.
@@ -121,6 +127,9 @@ where
                 // Monitor the background send task. If sending failed, fail early and terminate the stream.
                 status = self.send_task.join(), if self.send_task.is_joinable() => {
                     if let Err(status) = status {
+                        // TODO(#5991): Consider waiting until the receive
+                        // side terminates, so server responses that haven't
+                        // arrived are not lost.
                         self.terminate();
                         return Err(status);
                     }
--- a/src/gax-internal/src/grpc/grpc_rust/bidi.rs
+++ b/src/gax-internal/src/grpc/grpc_rust/bidi.rs
@@ -14,3 +14,136 @@
 
 pub use super::receive::ReceiveTask;
 pub use super::send::GrpcRustSend;
+use super::send::SendTask;
+use grpc::client::RecvStream;
+use prost::Message;
+
+/// A `tonic` adapter for `grpc-rust` RPCs.
+///
+/// Incoming server response messages are pulled via [`GrpcRustStreaming::message`].
+///
+/// Internally, this stream orchestrates:
+/// - A background [`ReceiveTask`](super::receive::ReceiveTask) that decodes inbound protobuf messages into `Response`s.
+/// - A background [`SendTask`](super::send::SendTask) that pumps outbound request messages.
+///
+/// `grpc-rust` stream operations ([`RecvStream::recv`](grpc::client::RecvStream::recv)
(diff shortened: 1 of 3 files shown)


src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -14,3 +14,140 @@
...
+
+        // Check for inbound responses or the status of the outbound send.
+        loop {
+            let responses = self
+                .responses
+                .as_mut()
+                .expect("active streams must have a response channel");
+            tokio::select! {
+                // Get the next response from the background receive task.
+                response = responses.recv() => match response {
```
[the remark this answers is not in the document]
joshuatants: Fixed.

src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -14,3 +14,136 @@
...
+pub struct GrpcRustStreaming<Response> {
+    responses: Option<tokio::sync::mpsc::Receiver<tonic::Result<Option<Response>>>>,
+    receive_task: Option<ReceiveTask>,
+    send_task: SendTask,
+    /// Holds a pre-decoded initial response message (e.g. from stream setup or handshake)
+    /// to yield on the first call to [`GrpcRustStreaming::message`] before polling `responses`.
+    pending_response: Option<Response>,
+}
+
+impl<Response> GrpcRustStreaming<Response>
```
vsharonlynn, line 42: Do we need to implement a `Drop` trait to clean up its resources, especially the background tasks?

src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -14,3 +14,136 @@
...
+                    }
+                }
+            }
+        }
+    }
+
+    /// Terminates the stream, cleaning up background tasks and response channels.
+    fn terminate(&mut self) {
+        self.responses.take();
+        self.receive_task.take();
```
vsharonlynn, line 144: For my learning... send task has an abort function while receive task doesn't. Is this intentional?

src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -14,3 +14,136 @@
...
+                            .receive_task
+                            .as_mut()
+                            .expect("active streams must have a receive task")
+                            .join()
+                            .await;
+                        return Err(status);
+                    }
+                },
+                // Monitor the background send task. If sending failed, fail early and terminate the stream.
+                status = self.send_task.join(), if self.send_task.is_joinable() => {
```
vsharonlynn, line 128: I'm thinking about this scenario:
When the server closes the connection, the `send_task` will fail, but there is still an actual server response in the responses.recv() channel that can be read.
Is it possible that this arm is triggered first instead of the `response =` branch? If it is called first, `self.terminate()` will be called and this destroys the responses channel and aborts the receive task. Then the server's actual response will be lost.

vsharonlynn (changes requested):

src/gax-internal/src/grpc/grpc_rust/bidi.rs
joshuatants: Not for GrpcRustStreaming.

Send/ReceiveTask need a custom impl of `Drop` because we terminate a tokio task by abort()ing it (dropping a tokio task merely detaches it). GrpcRustStreaming itself has no fields that require special handling on drop, and when GrpcRustStreaming is dropped, Rust automatically invokes drop on its fields.

src/gax-internal/src/grpc/grpc_rust/bidi.rs
joshuatants: receive_task is as Option because a terminal stream does not have an active receive task. Since it's an Option, it can be take()n and effectively drop()ped because it's unused, triggering abort() on its task handle.

send_task is currently stored unconditionally because it's created when making the initial request to the server. Rust prevents calling drop() on send_task in a &mut self function, because drop() would move the field out from the struct,  so I have an explicit abort() .

These constructors will be used in the next few PRs when processing the initial server response. Happy to revisit this again over time as the architecture becomes clearer.

src/gax-internal/src/grpc/grpc_rust/bidi.rs
joshuatants: I originally had `biased` but was wondering whether it'd over-prioritise the recv stream. You're right though that in that scenario we'd want to inspect the recv() channel to examine the server's messages.

I added `biased` back, and also a TODO to possibly drain the rest of the recv() channel if an error occurs on send.

src/gax-internal/src/grpc/grpc_rust/bidi.rs
vsharonlynn: Ok! Thank you for your explanation.

src/gax-internal/src/grpc/grpc_rust/bidi.rs
vsharonlynn: Ok. Thanks!

vsharonlynn (approved):

PR #6202, opened 2026-07-28
impl(bigquery): bubble up `JobPoller` errors instead of returning `Ok`

### LINKED ISSUE #6199: bigquery: `JobPoller` should return an error when job completes with `error_result`
Currently, if an inserted BigQuery job completes successfully but contains an `error_result` in its status payload (indicating the job itself failed, e.g., due to an invalid query), `JobPoller::until_done()` still returns `Ok(Job)`. 

We need to update the `JobPoller` in the BigQuery client to correctly surface job failures as errors.

haphungw:
Update `JobPoller::until_done` to surface errors when a job completes with a failure in its `status.error_result`.

Previously, `until_done` would return `Ok(job)` even if the job ultimately completed with an error state (as long as the underlying RPC succeeded). This updates the behavior to return a dedicated error type.

Fixes #6199

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -309,10 +309,9 @@ libraries:
           match: |2
                 - `folders/<folder-number>`
                 - `organizations/<organization-number>`
-          replace: |+
+          replace: |
             - `folders/<folder-number>`
             - `organizations/<organization-number>`
-
   - name: google-cloud-api-servicemanagement-v1
     version: 1.13.0
     apis:
@@ -524,6 +523,9 @@ libraries:
     skip_release: true
     rust:
       package_dependencies:
+        - name: thiserror
+          package: thiserror
+          force_used: true
         - name: tokio
           package: tokio
           force_used: true
--- a/src/bigquery/tests/job_poller_tests.rs
+++ b/src/bigquery/tests/job_poller_tests.rs
@@ -132,12 +132,11 @@ async fn non_retryable_error() -> anyhow::Result<()> {
     let result = poller.until_done().await;
 
     // Should return immediately after 1st attempt
-    let JobPollerError::Job(failed_job) = result.unwrap_err() else {
-        panic!("expected JobPollerError::Job");
+    let JobPollerError::ErrorProto(err) = result.unwrap_err() else {
+        panic!("expected JobPollerError::ErrorProto");
     };
 
-    let status = failed_job.status.unwrap();
-    assert_eq!(status.error_result.unwrap().reason.as_str(), "invalidQuery");
+    assert_eq!(err.reason.as_str(), "invalidQuery");
     Ok(())
 }
 
@@ -155,15 +154,11 @@ async fn retry_exhausted() -> anyhow::Result<()> {
     let result = poller.until_done().await;
 
     // Should stop retrying after limit of 3
-    let JobPollerError::Job(failed_job) = result.unwrap_err() else {
-        panic!("expected JobPollerError::Job");
+    let JobPollerError::ErrorProto(err) = result.unwrap_err() else {
+        panic!("expected JobPollerError::ErrorProto");
     };
 
-    let status = failed_job.status.unwrap();
-    assert_eq!(
-        status.error_result.unwrap().reason.as_str(),
-        "jobBackendError"
-    );
+    assert_eq!(err.reason.as_str(), "jobBackendError");
     Ok(())
 }
 
--- a/src/bigquery/tests/job_poller_tests.rs
+++ b/src/bigquery/tests/job_poller_tests.rs
@@ -14,6 +14,7 @@
 
 use google_cloud_bigquery_v2::client::JobService;
 use google_cloud_bigquery_v2::model::{ErrorProto, InsertJobRequest, Job, JobReference, JobStatus};
+use google_cloud_bigquery_v2::operation::JobPollerError;
 use google_cloud_bigquery_v2::stub::JobService as JobServiceStub;
 use google_cloud_gax::Result as GaxResult;
 use google_cloud_gax::options::RequestOptions;
@@ -28,6 +29,12 @@ mock! {
         async fn insert_job(
             &self,
             req: InsertJobRequest,
+            options: RequestOptions,
+        ) -> GaxResult<Response<Job>>;
+
+        async fn get_job(
+            &self,
+            req: google_cloud_bigquery_v2::model::GetJobRequest,
             options: RequestOptions,
         ) -> GaxResult<Response<Job>>;
     }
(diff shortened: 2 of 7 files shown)


src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -89,6 +89,45 @@ impl Default for JobRetryPolicy {
     }
 }
 
+/// Errors returned by the JobPoller.
+#[derive(Debug)]
+pub enum JobPollerError {
+    /// An error occurred during the RPC or LRO polling.
+    Gax(GaxError),
```
dbolduc: nit: I think we have been calling this type of branch `Rpc`

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -89,6 +89,45 @@ impl Default for JobRetryPolicy {
...
+    /// The job completed, but the BigQuery service reported an error in `status.error_result`.
+    Job(Box<Job>),
```
dbolduc: Q: Why use the full `Job` and not the `ErrorResult`? Is there relevant information in the `Job`?

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -89,6 +89,45 @@ impl Default for JobRetryPolicy {
     }
 }
 
+/// Errors returned by the JobPoller.
+#[derive(Debug)]
```
dbolduc: can we derive `thiserror::Error` ?

Look at other custom errors like the pubsub `PublishError`. I think it would save us all of the std::error::Error trait impls.

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -105,7 +144,7 @@ impl JobPoller {
     }
 
     /// Sets the maximum number of job-level attempts.
-    pub fn set_job_level_attempt_limit(mut self, limit: usize) -> Self {
+    pub fn with_attempt_limit(mut self, limit: u32) -> Self {
```
dbolduc, line 119: comment: I agree with the new name. Just be careful about refactoring in a feature PR like this. Life is easiest when PRs do one thing.

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -155,6 +198,9 @@ impl InsertJob {
     /// `InsertJob` operation. Note that the client library will supply a
     /// synthetic job ID for any retries.
     ///
+    /// WARNING: Unlike `poller()`, the `JobPoller` will return an error if the
+    /// final job completes with an `error_result` in its status.
+    ///
```
dbolduc: I think the warning belongs on the other method. This is the behavior I would expect. 🙂

src/bigquery/tests/job_poller_tests.rs
```diff
@@ -141,17 +150,65 @@ async fn retry_exhausted() -> anyhow::Result<()> {
...
     assert_eq!(
-        result.status.unwrap().error_result.unwrap().reason.as_str(),
+        status.error_result.unwrap().reason.as_str(),
         "jobBackendError"
     );
     Ok(())
 }
+
+#[tokio::test]
+async fn polling_success() -> anyhow::Result<()> {
```
dbolduc, line 166: same nit about doing too much in one PR. This seems like an independent change to the rest of the PR.

No need to change anything, just an FYI for the future.

dbolduc (commented):
Nice!

src/generated/cloud/bigquery/v2/src/operation.rs
haphungw: I was thinking that the full `Job` would give us the `job_reference` metadata. In case we have retries, we can use the final `job_id` for debugging.

src/generated/cloud/bigquery/v2/src/operation.rs
haphungw: Ack, done.

src/generated/cloud/bigquery/v2/src/operation.rs
haphungw: removed

src/generated/cloud/bigquery/v2/src/operation.rs

src/generated/cloud/bigquery/v2/src/operation.rs
dbolduc: Ack, that is reasonable.

dbolduc (approved):

src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -89,6 +89,17 @@ impl Default for JobRetryPolicy {
...
 
+/// Errors returned by the JobPoller.
+#[derive(Debug, thiserror::Error)]
+pub enum JobPollerError {
+    /// An error occurred during the RPC or LRO polling.
+    #[error(transparent)]
+    Rpc(#[from] GaxError),
+    /// The job completed, but the BigQuery service reported an error in `status.error_result`.
+    #[error("BigQuery job failed ({}): {}", .0.reason, .0.message)]
+    Job(crate::model::ErrorProto),
```
dbolduc: I think I prefer `ErrorProto`, but I am happy to go with either `ErrorProto` or `Job`.

My argument for `ErrorProto` is that the application could turn on tracing if they want to recover a failed job ID.

src/bigquery/tests/job_poller_tests.rs
```diff
@@ -122,13 +129,15 @@ async fn non_retryable_error() -> anyhow::Result<()> {
...
-    assert_eq!(
-        result.status.unwrap().error_result.unwrap().reason.as_str(),
-        "invalidQuery"
-    );
+    let JobPollerError::Job(failed_job) = result.unwrap_err() else {
+        panic!("expected JobPollerError::Job");
+    };
+
+    let status = failed_job.status.unwrap();
+    assert_eq!(status.error_result.unwrap().reason.as_str(), "invalidQuery");
```
dbolduc: just noticed that this need to match the new type of the error.

src/bigquery/tests/job_poller_tests.rs
```diff
@@ -141,17 +150,65 @@ async fn retry_exhausted() -> anyhow::Result<()> {
...
-        .insert_job()
-        .into_job_poller()
-        .set_job_level_attempt_limit(3);
+    let poller = client.insert_job().into_job_poller().with_attempt_limit(3);
 
-    let result = poller.until_done().await?;
+    let result = poller.until_done().await;
 
     // Should stop retrying after limit of 3
+    let JobPollerError::Job(failed_job) = result.unwrap_err() else {
```
dbolduc: same

src/generated/cloud/bigquery/v2/src/operation.rs
dbolduc: Also in the docs, maybe just say:

"The job completed, but the BigQuery service reported an internal error."

librarian.yaml
```diff
@@ -309,10 +309,9 @@ libraries:
           match: |2
                 - `folders/<folder-number>`
                 - `organizations/<organization-number>`
-          replace: |+
```
coryan, line 312: This changed the output for an unrelated library ... was that the intention?

haphungw: I ran `librarian tidy` and got this change

coryan: Fair, but that is a red flag.... The committed `librarian.yaml` file should be tidy all the time, we have a CI build to prevent drift.  If you run tidy and it changes the file then something is odd: you are using the wrong version of librarian, or librarian has an incompatible change, or something.

Issue #6218, opened 2026-07-29
Implement Query Retries


alvarowolfx:
To make query execution more reliable, so customers do not need to write complex retry loops. The SDK should automatically handles temporary network drops, rate limits, and job backend errors for both fast query path (jobs.query) and full jobs (jobs.insert). This should be done while allowing users to customize retry limits and policies when needed.

Reference for job backend errors: https://cloud.google.com/bigquery/docs/error-messages

See internal go/cloud-rust:bigquery-query-retries-design

PR #6219, opened 2026-07-29
impl(bigquery): generate job/request IDs and mark query RPCs idempotent

alvarowolfx:
Towards #5844 #6218


src/bigquery/src/query/run_query.rs
```diff
@@ -167,8 +223,12 @@ mod tests {
     #[tokio::test]
     async fn test_run_jobs_query() -> TestResult {
         let mut mock = MockJobService::new();
-        mock.expect_query()
-            .returning(move |_, _| Ok(Response::from(QueryResponse::new())));
+        mock.expect_query().returning(move |req, _| {
+            let req_id = &req.query_request.as_ref().unwrap().request_id;
+            assert!(req_id.starts_with(QUERY_REQUEST_ID_PREFIX), "{req_id:?}");
+            assert!(req_id.len() <= 36, "{req_id:?}"); // bigquery limits request id to 36 characters
```
suzmue: I would prefer this as a const, but since it is unused maybe thats fine. Maybe const _BIGQUERY_REQ_ID_LIMIT = 36 or something in the code?

src/bigquery/src/query/run_query.rs
```diff
@@ -105,6 +132,7 @@ mod tests {
         Job, JobConfiguration, JobReference, JobStatus, QueryRequest, QueryResponse,
     };
     use google_cloud_gax::response::Response;
+    use uuid::Uuid;
```
suzmue: I think this might be unnecessary with use super::*;

src/bigquery/src/query/run_query.rs
```diff
@@ -94,6 +104,23 @@ impl RunQuery {
...
+fn generate_job_reference(project_id: &str, location: &str) -> JobReference {
+    let job_id = generate_prefixed_id(JOB_ID_PREFIX);
+    let mut job_ref = JobReference::new()
+        .set_project_id(project_id.to_string())
+        .set_job_id(job_id);
+
+    if !location.is_empty() {
+        job_ref = job_ref.set_location(location.to_string());
+    }
+
```
suzmue: nit(style): consider removing the new line before the implicit return.

src/bigquery/src/query/run_query.rs
```diff
@@ -94,6 +104,23 @@ impl RunQuery {
...
+    job_ref
+}
+
+fn generate_prefixed_id(prefix: &str) -> String {
```
suzmue, line 131: Since the size of the ID matters for each of the places we use this (to varying degrees), it would be nice to have a comment explaining why this is in range for both of them / maybe defining what the returned size of the id will be as a guarantee of this function.

src/bigquery/src/query/run_query.rs
```diff
@@ -94,6 +104,23 @@ impl RunQuery {
     }
 }
 
+fn generate_job_reference(project_id: &str, location: &str) -> JobReference {
+    let job_id = generate_prefixed_id(JOB_ID_PREFIX);
```
suzmue, line 112: nit: could be nice to have a link to any specs for what this ID should look like.

src/bigquery/src/query/run_query.rs
```diff
@@ -77,12 +84,15 @@ impl RunQuery {
                 .execute()
                 .await
         } else {
+            let query_request_id = generate_prefixed_id(QUERY_REQUEST_ID_PREFIX);
```
suzmue, line 87: nit: could be nice to have a link to any specs for what this ID should look like (including that 36 limit mentioned below)

supernit: for symmetry across if-else block, consider moving this initialization to `let query_request = generate_query_request(...);`

src/bigquery/src/query/execution.rs
```diff
@@ -97,6 +101,9 @@ impl InsertJobExecutor {
             .job_service
             .insert_job()
             .with_request(self.request)
+            // jobs.insert is idempotent because every request
+            // carries a generated UUID job_id.
```
suzmue: supernit: remove UUID (or add it above), since you also add the prefix it is not a pure UUID

src/bigquery/src/query/run_query.rs
```diff
@@ -141,10 +169,38 @@ mod tests {
...
+        assert_eq!(job_ref.location.as_deref(), Some("us-central1"));
+    }
+
     #[tokio::test]
     async fn test_run_jobs_insert() -> TestResult {
         let mut mock = MockJobService::new();
-        mock.expect_insert_job().returning(|_, _| {
+        mock.expect_insert_job().returning(|req, _| {
+            let job_ref = req.job.as_ref().unwrap().job_reference.as_ref().unwrap();
+            assert!(job_ref.job_id.starts_with(JOB_ID_PREFIX), "{job_ref:?}");
```
suzmue, line 216: nit: This technically has a length limit too (but its very large). But could still be worth checking explicitly? Your call.

suzmue (approved):

src/bigquery/src/query/run_query.rs
alvarowolfx: fair, but we are far far away from the 1024 limit. For request id is more tricky and easy to do things wrong since it's just 36

src/bigquery/src/query/run_query.rs
alvarowolfx: added links on the `generate_prefixed_id` method. The downside of movings more things to a `generate_query_request` method is that is going to be less visible in this functions to see the default values for generating those requests. I think I still prefer to keep just the request id generation being called here.

PR #6222, opened 2026-07-30
feat(gax): add grpc-rust bidi call driver

joshuatants:
For #5991

### DIFF
--- a/src/gax-internal/src/grpc/grpc_rust/bidi.rs
+++ b/src/gax-internal/src/grpc/grpc_rust/bidi.rs
@@ -246,6 +246,7 @@ mod tests {
         value: String,
     }
 
+    // TODO(#5991): Add tests for failure paths.
     #[tokio::test]
     async fn bidi_call_yields_response_messages() -> anyhow::Result<()> {
         // Arrange
--- a/src/gax-internal/src/grpc/grpc_rust/bidi.rs
+++ b/src/gax-internal/src/grpc/grpc_rust/bidi.rs
@@ -258,6 +258,7 @@ mod tests {
         struct TestInvoker {
             observed_headers: Arc<Mutex<Option<RequestHeaders>>>,
             observed_messages: Arc<Mutex<Vec<TestMessage>>>,
+            notify: Arc<tokio::sync::Notify>,
         }
 
         impl Invoke for TestInvoker {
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -165,4 +167,241 @@ impl<Response> std::fmt::Debug for GrpcRustStreaming<Response> {
...
+                        self.state = StreamState::HeadersSent;
+                        // Wait for the client request message to be sent before yielding initial headers.
+                        while self
+                            .observed_messages
+                            .lock()
+                            .expect("lock messages")
+                            .is_empty()
+                        {
+                            tokio::task::yield_now().await;
+                        }
```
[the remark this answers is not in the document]
joshuatants: Addressed.

src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -165,4 +167,256 @@ impl<Response> std::fmt::Debug for GrpcRustStreaming<Response> {
...
+                    send_task.abort();
+                    let status = receive_task.join().await;
+                    return Err(status);
+                }
+            },
+            // TODO(#5991): Consider waiting until the receive
+            // side terminates, so server responses that haven't
+            // arrived are not lost. Perhaps we can removed biased;
+            // once that's in place.
+            status = send_task.join(), if send_task.is_joinable() => {
```
vsharonlynn, line 225: IIUC `send_task.join()` is not cancel-safe, so it must not be polled directly inside a `tokio::select!` loop with other network operations?

What happens when:
1. tokio::select! calls `send_task.join()`
2. `join()` executes `self.handle.take()`. The `JoinHandle` is moved out of the `SendTask` struct and held inside the `join()` future. The `SendTask` struct now holds `None`.
3. While `join()` is waiting for background task to finish, the server responds with its HTTP/2 headers. `response.recv()` wakes up, processes the header.
4. `response.recv()` is selected in `tokio::select!`, so `tokio::select!` cancels and drops the pending `send_task.join()` future.
5. `JoinHandle` is being held inside the dropped `join()` future, so the handle is deleted from memory.

As a result, the `tokio::seelct!` block successfully receives the headers and returns `GrpcRustStreaming::new(..., send_task)` to the user.
But `send_task`'s internal task handle was destroyed.
Later when the application finishes the stream and tries to check final `SendTask` status, it will fail with "already joined" error because the handle is missing.

Is this possible?

src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -165,4 +167,256 @@ impl<Response> std::fmt::Debug for GrpcRustStreaming<Response> {
...
+    use std::sync::{Arc, Mutex};
+
+    #[derive(Clone, PartialEq, Message)]
+    struct TestMessage {
+        #[prost(string, tag = "1")]
+        value: String,
+    }
+
+    #[tokio::test]
+    async fn bidi_call_yields_response_messages() -> anyhow::Result<()> {
```
vsharonlynn, line 251: This only tests the happy path. ARe you planning to add tests for the other arms? I'm ok with adding a TODO.

vsharonlynn (approved):

src/gax-internal/src/grpc/grpc_rust/bidi.rs
joshuatants: This can't happen: in send_task.join(), take() only executes after handle.await, so either the task completes and the handle is set to `None`, or the `join()` future is dropped pre-take() and the handle remains `Some`.

Also, note that task cancellation can only happen at `await`s, so after handle.await, join() cannot be cancelled and so we can't have a situation where the handle has been `take()`n but the `match` statement hasn't executed.

Issue #5716, opened 2026-05-21
Implement appendable upload


vsharonlynn:
Tracking issue for PRs.

PR #6225, opened 2026-07-30
feat(storage): add code samples

vsharonlynn:
Issue #5716.

### DIFF
--- a/src/storage/examples/Cargo.toml
+++ b/src/storage/examples/Cargo.toml
@@ -47,5 +47,4 @@ tracing.workspace             = true
 
 [features]
 run-integration-tests     = []
-unstable-stream           = ["google-cloud-storage/unstable-stream"]
 skipped-integration-tests = []
--- a/src/storage/examples/Cargo.toml
+++ b/src/storage/examples/Cargo.toml
@@ -23,6 +23,9 @@ repository.workspace = true
 keywords.workspace   = true
 categories.workspace = true
 
+[lints]
+workspace = true
+
 [dependencies]
 anyhow.workspace              = true
 bytes.workspace               = true
@@ -44,4 +47,5 @@ tracing.workspace             = true
 
 [features]
 run-integration-tests     = []
+unstable-stream           = ["google-cloud-storage/unstable-stream"]
 skipped-integration-tests = []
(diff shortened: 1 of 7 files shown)


src/storage/examples/src/lib.rs
```diff
@@ -521,6 +521,77 @@ pub async fn run_object_examples(buckets: &mut Vec<String>) -> anyhow::Result<()
...
+            .create_bucket()
+            .set_parent("projects/_")
+            .set_bucket_id(rapid_bucket_id.clone())
+            .set_bucket(
+                Bucket::new()
+                    .set_project(format!("projects/{project_id}"))
+                    .set_location("us-central1")
+                    .set_custom_placement_config(
+                        CustomPlacementConfig::new()
+                            .set_data_locations(["us-central1-a".to_string()]),
```
coryan: Please use these to pick a zone and region:

https://github.com/googleapis/google-cloud-rust/blob/4ee1b7b01dd5adc39ffce1307570df66e015633f/src/test-utils/src/runtime_config.rs#L34-L46

If we ever want to move the tests to a different location, it would be nice to keep the configuration in a single place.

src/storage/examples/Cargo.toml
```diff
@@ -44,4 +47,5 @@ tracing.workspace             = true
 
 [features]
 run-integration-tests     = []
+unstable-stream           = ["google-cloud-storage/unstable-stream"]
```
coryan: I am not sure you need to use a feature the the samples...  First, because in the examples you should just enable the feature we don't need to worry about the stability of the asynchronous streams in that case, and second, because I don't see any use of `.into_stream()` or otherwise using the streaming traits.

src/storage/examples/src/lib.rs
vsharonlynn: Thanks for this, wasn't aware of these test utils!
Updated bucket location and data placement configuration to use `region_id()` and `zone_id()` from runtime_config.

src/storage/examples/Cargo.toml
vsharonlynn: Yes, you're right. My bad. I removed it. Thanks for catching that.

joshuatants (approved):

PR #6233, opened 2026-07-30
impl(bigquery): add default job retry policies

alvarowolfx:
Towards #6218 #5844


src/bigquery/src/retry_policy.rs
```diff
@@ -0,0 +1,336 @@
...
+                | Code::DeadlineExceeded
+                | Code::Internal
+                | Code::ResourceExhausted
+                | Code::Unavailable
+                | Code::Unknown => RetryResult::Continue(error),
+                _ => RetryResult::Permanent(error),
+            };
+        }
+        RetryResult::Permanent(error)
+    }
```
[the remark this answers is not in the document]
alvarowolfx: we are gonna send JobID or RequestID on the RPCs that idempotency could be a problem, so we don't need to check here

src/bigquery/src/retry_policy.rs
```diff
@@ -0,0 +1,336 @@
...
+        let permanent_http = [400, 404, 408, 409, 501];
+        for code in permanent_http {
+            let err = GaxError::http(code, HeaderMap::new(), bytes::Bytes::new());
+            assert!(
+                p.on_error(&state, err).is_permanent(),
+                "expected permanent for HTTP {}",
+                code
+            );
+        }
+    }
```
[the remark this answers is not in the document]
alvarowolfx: we are gonna send JobID or RequestID on the RPCs that idempotency could be a problem, so we don't need to check here

src/bigquery/src/retry_policy.rs
```diff
@@ -0,0 +1,336 @@
...
+/// The result of evaluating a BigQuery job error against a [`JobRetryPolicy`].
+#[derive(Debug)]
+#[allow(dead_code)]
+pub(crate) enum JobRetryResult {
+    Continue(Duration, QueryError),
+    Exhausted(QueryError),
+    Permanent(QueryError),
+}
+
+#[cfg(test)]
```
suzmue: nit: Consider moving this into the test module

src/bigquery/src/retry_policy.rs
```diff
@@ -0,0 +1,336 @@
...
+pub(crate) fn is_query_error_retryable(err: &crate::error::QueryError) -> bool {
+    match err {
+        crate::error::QueryError::JobFailed { errors } => is_retryable_errors(errors),
+        _ => false,
+    }
+}
+
+#[allow(dead_code)]
+pub(crate) fn is_retryable_errors(errors: &[ErrorProto]) -> bool {
+    !errors.is_empty() && errors.iter().all(|e| is_retryable_error_reason(&e.reason))
```
suzmue, line 173: question: empty errors is not retryable?

src/bigquery/src/lib.rs
```diff
@@ -28,6 +28,7 @@
 pub use google_cloud_gax::Result;
 pub use google_cloud_gax::error::Error;
 pub mod error;
+pub mod retry_policy;
```
suzmue: Everything in here is pub(crate) I believe. Maybe keep the whole module pub(crate) for now? Your choice.

src/bigquery/src/retry_policy.rs
```diff
@@ -0,0 +1,336 @@
...
+
+/// A policy trait for handling BigQuery job execution retries.
+///
+/// Note: This trait is kept crate-internal for now. It is a copy/adaptation of GAX's
+/// `RetryPolicy` trait specifically tailored for BigQuery job-level re-issuance.
+/// In the future, we plan to refine this trait and discuss how to make it public
+/// so customers can provide their own custom job retry policies.
+#[allow(dead_code)]
+pub(crate) trait JobRetryPolicy<S = RetryState>: Send + Sync + std::fmt::Debug {
+    fn on_error(&self, state: &S, error: crate::error::QueryError) -> JobRetryResult;
```
suzmue: You already have a use for crate::error::QueryError so you should be able to use it directly (as QueryError) throughout this file.

src/bigquery/src/retry_policy.rs
```diff
@@ -0,0 +1,336 @@
...
+pub(crate) fn default_job_retry_policy() -> Arc<dyn JobRetryPolicy> {
+    Arc::new(RetryableJobErrors::default())
+}
+
+#[allow(dead_code)]
+pub(crate) fn is_query_error_retryable(err: &crate::error::QueryError) -> bool {
+    match err {
+        crate::error::QueryError::JobFailed { errors } => is_retryable_errors(errors),
+        _ => false,
+    }
```
suzmue, line 168: style nit, take it or leave it:

```suggestion
     let QueryError::JobFailed { errors } = err else {
        return false;
    };
    is_retryable_errors(errors)

```

suzmue (approved):

src/bigquery/src/lib.rs
alvarowolfx: good catch, this should be pub(crate) for now.

src/bigquery/src/retry_policy.rs
alvarowolfx: yeah, in this case we need to be strict on trying to re-insert a job (which is what retrying means here). We should only do it if we find one of the backend error checked on `is_retryable_error_reason`. If the error list is empty, is not safe to retry (it could be a 400 error with an invalid query for example)

PR #6232, opened 2026-07-30
test(bigquery): add integration tests for job polling

haphungw:
Add live integration tests for `JobPoller` in `google-cloud-bigquery-v2`, covering successful query polling and error handling paths against live BigQuery APIs.

Fixes #5754

PR #6237, opened 2026-07-30
feat(bigquery): heap-box futures and builder in JobPoller to prevent stack overflow

haphungw:
Heap-box `JobPoller::until_done` future, `JobPoller` `builder`, and `DiscoveryPoller::until_done`.

Job structs are large (~7.1 KB) and cause deep unboxed async future state machines to overflow the default 2 MB thread stack during integration tests.

### DIFF
--- a/src/generated/cloud/bigquery/v2/src/operation.rs
+++ b/src/generated/cloud/bigquery/v2/src/operation.rs
@@ -104,14 +104,14 @@ pub enum JobPollerError {
 #[derive(Debug)]
 pub struct JobPoller {
     policy: JobRetryPolicy,
-    builder: InsertJob,
+    builder: Box<InsertJob>,
 }
 
 impl JobPoller {
     pub(crate) fn new(builder: InsertJob) -> Self {
         Self {
             policy: JobRetryPolicy::default(),
-            builder,
+            builder: Box::new(builder),
         }
     }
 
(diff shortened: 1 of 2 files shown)


src/generated/cloud/bigquery/v2/src/operation.rs
```diff
@@ -128,38 +128,43 @@ impl JobPoller {
...
-            attempts += 1;
-
-            if let Some(status) = &job_result.status
-                && let Some(err) = &status.error_result
-            {
-                if is_retryable_job_error(&err.reason)
-                    && attempts < self.policy.job_level_attempt_limit
+    pub fn until_done(
+        self,
+    ) -> std::pin::Pin<Box<dyn std::future::Future<Output = Result<Job, JobPollerError>> + Send>>
```
alvarowolfx, line 133: why does the signature has to change that much ? isn't this gonna diverge from other LROs that are async and return just a Result ?

Maybe I don't understand Pin/Box enough to understand what is going on here.

haphungw: I was struggling to get this test pass due to stack overflow
https://github.com/googleapis/google-cloud-rust/pull/6232/changes#diff-824e62dc84d6a981c4c62f1f254e61712feacc64d410a3a1b0744f981cb657edR267
https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/6916363e-91b1-4147-81c4-2bf87a03abd7?project=rust-sdk-testing&e=SqlStudioMysqlLaunch::SqlStudioMysqlEnabled

The builder and polling loop state machine get pretty huge on the stack, so boxing it offloads that to the heap.
The `the .until_done().await` caller syntax is still the same, but it seems like I haven't been able to fix the issue.

dbolduc:
I think this was done in #6390, #6392. Can we close the PR?

haphungw:
yes, closing

Issue #5664, opened 2026-05-13
Define basic client surface


dbolduc:

dbolduc:
Some prototyping from your favorite robot:

https://github.com/googleapis/google-cloud-rust/compare/main...dbolduc:bq-write-sketch?expand=1

See: https://github.com/dbolduc/google-cloud-rust/blob/bb851ad7176c19c78a6c046e14cbb7e1fe0ae0db/TASK.md

PR #6241, opened 2026-07-30
impl(bigquery): add default writer

dbolduc:
Add in the default writer implementation.

We take plenty of todo's on send optimization, multiplexing. For now we just want something, anything that works.

Towards #5664


alvarowolfx (approved):
nice to see things coming together

PR #6244, opened 2026-07-31
chore(spanner): implement RequestIdCreator for Spanner request IDs

olavloite:
Implement `RequestIdCreator` and process-wide random ID caching as Step 2 of client-side request ID (`x-goog-spanner-request-id`) support.

### DIFF
--- a/src/spanner/src/request_id.rs
+++ b/src/spanner/src/request_id.rs
@@ -1,0 +1,176 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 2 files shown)


src/spanner/src/request_id.rs
```diff
@@ -0,0 +1,176 @@
...
+        result
+    }
+
+    /// Returns a full Request ID string for an RPC attempt:
+    /// `"1.<RAND_PROCESS_ID>.<client_id>.<channel_id>.<request_id>.<attempt>"`
+    ///
+    /// # Arguments
+    /// * `channel_id` - The 1-based channel identifier (`1, 2, ...`), or `0` for an unknown channel.
+    /// * `attempt` - The 1-based attempt number (`1` for initial attempt, `2` for first retry).
+    pub(crate) fn next_id(&self, channel_id: usize, attempt: u32) -> String {
```
sakthivelmanii: next_id vs next_base_id -> can we better fn names for this? When I look at the name, it's confusing. Looking at deeper with parameters, I can see the difference.

sakthivelmanii (approved):

src/spanner/src/request_id.rs
olavloite: Good point. Renamed to:
- next_id_prefix
- next_id_with_attempt

PR #6247, opened 2026-07-31
impl(bigquery): add arrow writer builder

dbolduc:
Adds a writer builder for arrow format. This thing is used to make a stream.

At the moment it can only attach to the default stream. Note that the default stream always exists, so we do not need to send an RPC to Get/Create it first. We can just plumb the resource name.

In the future, the API will look more like:

```rs
impl WriterBuilder {
  // stream-level options. We list one as an example.
  pub fn set_multiplexing(self, bool) -> Self;

  // Use the default stream for a table.
  pub fn default(self) -> DefaultWriter;

  // Create a stream of a given type for a table.
  //
  // Note that these send a `CreateWriteStream` RPC, so they are async and can fail.
  pub async fn pending(self, table: String) -> Result<PendingWriter>;
  pub async fn committed(self, table: String) -> Result<CommittedWriter>;
  pub async fn buffered(self, table: String) -> Result<BufferedWriter>;

  // Attach to an existing write stream, by name
  pub async fn attach<W: Writer>(self, write_stream: String) -> Result<W>;
}
```

Towards #5664

### DIFF
--- a/src/bigquery-write/src/arrow/default.rs
+++ b/src/bigquery-write/src/arrow/default.rs
@@ -19,16 +19,17 @@ use crate::runner::Runner;
 use crate::transport::Transport;
 use std::sync::Arc;
 
-/// A writer for the default stream
+/// A writer for the [default stream]
+///
+/// [default stream]: https://docs.cloud.google.com/bigquery/docs/write-api#default_stream
 pub struct DefaultWriter {
     // TODO(#5744) - support multiplexed connections
     runner: Runner,
-    write_stream: String,
-    schema: ArrowSchema,
+    pub(crate) write_stream: String,
+    pub(crate) schema: ArrowSchema,
 }
 
 impl DefaultWriter {
-    #[cfg_attr(not(test), expect(dead_code))]
     pub(crate) fn new(inner: Arc<Transport>, write_stream: String, schema: ArrowSchema) -> Self {
         let runner = Runner::new(inner);
         Self {
--- a/src/bigquery-write/src/arrow/default.rs
+++ b/src/bigquery-write/src/arrow/default.rs
@@ -22,6 +22,7 @@ use std::sync::Arc;
 /// A writer for the [default stream]
 ///
 /// [default stream]: https://docs.cloud.google.com/bigquery/docs/write-api#default_stream
+#[derive(Debug)]
 pub struct DefaultWriter {
     // TODO(#5744) - support multiplexed connections
     runner: Runner,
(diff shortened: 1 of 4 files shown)


src/bigquery-write/src/arrow/default.rs
```diff
@@ -19,16 +19,17 @@ use crate::runner::Runner;
 use crate::transport::Transport;
 use std::sync::Arc;
 
-/// A writer for the default stream
+/// A writer for the [default stream]
+///
+/// [default stream]: https://docs.cloud.google.com/bigquery/docs/write-api#default_stream
 pub struct DefaultWriter {
```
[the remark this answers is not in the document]
dbolduc: We don't want to derive `Clone`. I added `Debug`.

src/bigquery-write/src/arrow/writer_builder.rs
```diff
@@ -0,0 +1,59 @@
...
+    pub(crate) fn new(inner: Arc<Transport>, schema: ArrowSchema) -> Self {
+        Self { inner, schema }
+    }
+
+    // TODO(#6224) - add an example showing the format of `table`
+    /// Create a writer for the [default stream] for the given table.
+    ///
+    /// [default stream]: https://docs.cloud.google.com/bigquery/docs/write-api#default_stream
+    pub fn default<T: Into<String>>(self, table: T) -> DefaultWriter {
+        let mut write_stream = table.into();
```
alvarowolfx, line 40: q: should we do any kind of validation for the resource name ?

alvarowolfx (approved):

src/bigquery-write/src/arrow/writer_builder.rs
dbolduc: That is a good q.

I considered making this return a `Result<>` and yielding a `BindingError` on a badly formatted resource name. That's similar to how the GAPICs work. And the error tends to be better than what the server reports.

The service error looks like:

```
the service reports an error with code INVALID_ARGUMENT described as: Invalid BigQuery dataset name . Non empty project_id and dataset_id are required
```

or

```
the service reports an error with code INVALID_ARGUMENT described as: Invalid stream name. Entity: projects/dbolduc-test/<omitted>/streams/_default
```

---

Ok, I think I will do this, partially because I realized I messed up the format of the table in all of my unit tests.

dbolduc: I changed the signature in preparation for validating the resource. I am going to merge this then send a follow up with the implementation.

PR #6255, opened 2026-08-03
feat(gax): add grpc-rust bidi streaming

joshuatants:
For #5991

### DIFF
--- a/src/gax-internal/src/grpc/grpc_rust.rs
+++ b/src/gax-internal/src/grpc/grpc_rust.rs
@@ -14,6 +14,7 @@
 
 use super::grpc_helpers;
 use crate::grpc::tonic::{Extensions, Response as TonicResponse, Result as TonicResult};
+use crate::observability::RequestRecorder;
 use crate::options::{ClientConfig, InstrumentationClientInfo};
 use crate::universe_domain::DEFAULT_UNIVERSE_DOMAIN;
 use google_cloud_auth::credentials::Credentials;
@@ -148,9 +149,7 @@ impl GrpcRustClient {
             .with_method_name(path.as_str())
             .with_metadata(metadata);
 
-        if let Some(recorder) = crate::observability::RequestRecorder::current() {
-            recorder.on_grpc_request(&path);
-        }
+        Self::record_grpc_request(&path);
         match bidi::invoke_bidi(&self.inner.invoker, request_headers, request).await {
             Ok(response) => {
                 Self::record_grpc_response();
@@ -240,14 +239,20 @@ impl GrpcRustClient {
         );
     }
 
+    fn record_grpc_request(path: &http::uri::PathAndQuery) {
+        if let Some(recorder) = RequestRecorder::current() {
+            recorder.on_grpc_request(path);
+        }
+    }
+
     fn record_grpc_response() {
-        if let Some(recorder) = crate::observability::RequestRecorder::current() {
+        if let Some(recorder) = RequestRecorder::current() {
             recorder.on_grpc_response();
         }
     }
 
     fn record_grpc_error(status: &tonic::Status) {
-        if let Some(recorder) = crate::observability::RequestRecorder::current() {
+        if let Some(recorder) = RequestRecorder::current() {
             recorder.on_grpc_error(&super::from_status::to_gax_error(status.clone()));
         }
     }
--- a/src/gax-internal/src/grpc/grpc_rust.rs
+++ b/src/gax-internal/src/grpc/grpc_rust.rs
@@ -19,10 +19,12 @@ use crate::universe_domain::DEFAULT_UNIVERSE_DOMAIN;
 use google_cloud_auth::credentials::Credentials;
 use google_cloud_gax::Result as GaxResult;
 use google_cloud_gax::client_builder::{Error as BuilderError, Result as ClientBuilderResult};
+use google_cloud_gax::error::Error;
 use google_cloud_gax::options::RequestOptions;
 use google_cloud_gax::polling_backoff_policy::PollingBackoffPolicy;
 use google_cloud_gax::polling_error_policy::PollingErrorPolicy;
 use grpc::client::{Channel, ChannelOptions};
+use grpc::core::RequestHeaders;
 use grpc::credentials::LocalChannelCredentials;
 // TODO(#5991): remove once grpc-rust corrects the typo.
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -209,6 +246,37 @@ impl GrpcRustClient {
...
+    fn record_grpc_error(status: &tonic::Status) {
+        if let Some(recorder) = crate::observability::RequestRecorder::current() {
+            recorder.on_grpc_error(&super::from_status::to_gax_error(status.clone()));
+        }
+    }
+
+    fn note_ignored_extensions(extensions: &Extensions) {
+        if !extensions.is_empty() {
+            tracing::debug!(
+                extension_count = extensions.len(),
```
xlai20, line 263: What's the purpose of keeping a variable of `extension_count`?

src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -209,6 +246,37 @@ impl GrpcRustClient {
...
+                "grpc-rust request extensions are not propagated"
```
xlai20, line 264: Why do we not propagate request extensions? What are these `extensions` btw?

src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -103,34 +105,69 @@ impl GrpcRustClient {
...
+        let metadata = MetadataMap::from_headers(headers)
+            .try_into()
+            .map_err(Error::ser)?;
+        let request_headers = RequestHeaders::new()
+            .with_method_name(path.as_str())
+            .with_metadata(metadata);
+
+        if let Some(recorder) = crate::observability::RequestRecorder::current() {
+            recorder.on_grpc_request(&path);
+        }
```
xlai20: Since you already created `record_grpc_response` function, for symmetric purpose, why not extract this as `record_grpc_request` function as well?

src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -360,6 +428,7 @@ fn make_tls_credentials() -> ClientBuilderResult<Arc<RustlsChannelCredentials>>
         })
 }
 
+// TODO(#5991): Add integration tests for `GrpcRustClient::bidi_stream` and `GrpcRustClient::bidi_stream_with_status` (covering happy paths and request stream failures).
```
xlai20, line 419: I was long educated with the convention that "production code changes must be accompanied by automated tests in the same Changelist", even having some unit tests coverage are better than none. 

Is there any strong reason that you add the production code change first, and then submit the tests in a second PR?

src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -209,6 +246,37 @@ impl GrpcRustClient {
...
         );
     }
+
+    fn channel_state(&self) -> ConnectivityState {
+        let mut channel = self.inner.invoker.clone();
+        channel.get_state(false)
+    }
+
+    fn is_connection_failure(status: &tonic::Status, connection_state: ConnectivityState) -> bool {
+        status.code() == tonic::Code::Unavailable
```
xlai20: Do we need to take care of any backoff and retry strategy?  The "Unavailable" error code looks like something that needs to be re-attempted at some exponential backoff time.
https://docs.cloud.google.com/storage/docs/retry-strategy#rest-apis_1

xlai20 (changes requested):

src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -209,15 +245,39 @@ impl GrpcRustClient {
...
+        Some(channel.get_state(false))
+    }
+
+    fn is_connection_failure(
+        status: &tonic::Status,
+        connection_state: Option<ConnectivityState>,
+    ) -> bool {
+        status.code() == tonic::Code::Unavailable
+            && matches!(connection_state, Some(ConnectivityState::TransientFailure))
+    }
```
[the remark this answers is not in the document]
joshuatants: Done

src/gax-internal/src/grpc/grpc_rust.rs
joshuatants: This isn't a variable but a named field for structured logging: https://docs.rs/tracing/latest/tracing/index.html#recording-fields

src/gax-internal/src/grpc/grpc_rust.rs
joshuatants: `grpc-rust` doesn't support it: https://docs.google.com/document/d/15DPV_-Pfjd6Myf_QTwhdjwFwy8f0nsUrjV7Vl9QueSU/edit?pli=1&tab=t.0#bookmark=id.j19dssbyc825

src/gax-internal/src/grpc/grpc_rust.rs
joshuatants: Good catch, done.

src/gax-internal/src/grpc/grpc_rust.rs
joshuatants: The tests are long and I want to keep the PRs small so that they're easier to review. 

Also, this PR isn't going to be merged into `main` directly.  I plan to add the tests before merging to `main`.

src/gax-internal/src/grpc/grpc_rust.rs
joshuatants: No: bidi uses streams; if a stream fails mid-way, the transport layer has no knowledge of what the correct retry semantics are. This is something that only the layers above know and so any backoff/retry logic should be implemented there.

For comparison, the `tonic`-backed Client (in grpc.rs) also does not implement backoff/retry for bidi.

src/gax-internal/src/grpc/grpc_rust.rs
xlai20: SGTM

src/gax-internal/src/grpc/grpc_rust.rs
xlai20: Thanks for the explanation. Retry strategy is actually sitting at a higher-level service layer, not at this transport layer.

xlai20 (approved):

PR #6260, opened 2026-08-03
feat(gax): request-level custom HTTP/gRPC headers

xlai20:
## Overview
Adds support for request-level custom HTTP/gRPC headers via `RequestOptionsBuilder::with_custom_header`, allowing developers to inject tracking or tracing headers (e.g., `x-client-tracking-id`, `X-B3-TraceId`) on individual API calls without signature changes.

DD: go/rust-sdk-custom-header
Part of [#5997](https://github.com/googleapis/google-cloud-rust/issues/5997)

## Key Changes
* **`google-cloud-gax`**:
  * Added default method `with_custom_header<K, V>(mut self, name: K, value: V) -> Self` to `RequestOptionsBuilder`.
* **`google-cloud-gax-internal`**:
  * Updated `ReqwestClient::request()` to extract custom headers and enforce a **"System-Wins" precedence policy**.
* **Tests**:
  * Added unit tests for `RequestOptionsBuilder::with_custom_header`.
  * Added HTTP transport wire emission and precedence tests in `http_custom_header.rs` (via `echo_server`).
  * Integration tests will be added later in google-cloud-storage upon version bump in next PR.


src/gax-internal/src/http.rs
```diff
@@ -416,12 +416,33 @@ impl ReqwestClient {
...
+        ] {
+            headers.remove(key);
+        }
+
+        // System headers overwrite custom headers.
+        for (k, v) in cred_headers.into_iter() {
+            if let Some(k) = k {
+                headers.insert(k, v);
+            }
+        }
```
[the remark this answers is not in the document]
xlai20: Done.

src/gax/src/options.rs
```diff
@@ -221,6 +221,28 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
...
+                .get_extension::<http::HeaderMap>()
+                .cloned()
+                .unwrap_or_default();
+            headers.insert(name, value);
+            let mut options = std::mem::take(self.request_options());
+            options = options.insert_extension(headers);
+            *self.request_options() = options;
+        }
+        self
+    }
```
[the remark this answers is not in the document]
xlai20: Done.

src/gax-internal/tests/http_custom_header.rs
```diff
@@ -0,0 +1,167 @@
...
+        let mut options = RequestOptions::default();
+        options = with_custom_header(options, "user-agent", "fake-agent");
+        options = with_custom_header(options, X_GOOG_USER_PROJECT, "fake-project");
+
+        let response: serde_json::Value = client
+            .execute(builder, Some(json!({})), options)
+            .await?
+            .into_body();
+
+        assert!(
```
joshuatants, line 155: Is 'authorization' not a system header too?

src/gax/src/options.rs
```diff
@@ -221,6 +221,31 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
...
+    {
+        match (name.try_into(), value.try_into()) {
+            (Ok(name), Ok(value)) => {
+                use internal::RequestOptionsExt;
+                let mut headers = self
+                    .request_options()
+                    .get_extension::<http::HeaderMap>()
+                    .cloned()
+                    .unwrap_or_default();
+                headers.insert(name, value);
```
joshuatants: How about repeated headers, e.g. `Cache-Control` or custom repeated headers?

src/gax-internal/tests/http_custom_header.rs
```diff
@@ -0,0 +1,167 @@
...
+    use super::mock_credentials::{MockCredentials, mock_credentials};
+    use google_cloud_auth::credentials::{CacheableResource, Credentials, EntityTag};
+    use google_cloud_gax::options::RequestOptions;
+    use google_cloud_gax::options::internal::RequestOptionsExt;
+    use http::{HeaderMap, HeaderValue, header::HeaderName};
+    use serde_json::json;
+
+    const X_GOOG_USER_PROJECT: &str = "x-goog-user-project";
+
+    fn with_custom_header(
```
joshuatants, line 28: Why is the logic reimplemented here instead of using the implementation in RequestOptionsBuilder?

src/gax/src/options.rs
```diff
@@ -221,6 +221,31 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
     {
         unimplemented!();
     }
+
+    /// Injects a custom HTTP header into this specific request.
```
joshuatants: Let's include more information about how custom headers behave:
- precedence vs system headers 
- behaviour when used with with_user_project/with_user_agent

src/gax-internal/src/http.rs
```diff
@@ -416,12 +417,27 @@ impl ReqwestClient {
             builder = builder.timeout(timeout);
         }
 
-        let mut headers = match self.cred.headers(Extensions::new()).await {
+        let cred_headers = match self.cred.headers(Extensions::new()).await {
```
joshuatants: How about x-goog-user-project?

Are there other reserved headers that should be removed as well? What is the behaviour of the other SDKs?

src/gax/src/options.rs
```diff
@@ -221,6 +221,31 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
...
+                    .request_options()
+                    .get_extension::<http::HeaderMap>()
+                    .cloned()
+                    .unwrap_or_default();
+                headers.insert(name, value);
+                let mut options = std::mem::take(self.request_options());
+                options = options.insert_extension(headers);
+                *self.request_options() = options;
+            }
+            _ => panic!("invalid header name or value"),
```
joshuatants: Instead of panicking, I think either:

1. Have name and value be HeaderName/Value and make callers do the validation, or
2. Return a Result instead, renaming the function to `try_with_custom_header`

Otherwise, right now there's no way for the application to fail gracefully on malformed input.

src/gax-internal/src/http.rs
```diff
@@ -416,12 +417,27 @@ impl ReqwestClient {
...
         };
 
+        use google_cloud_gax::options::internal::RequestOptionsExt;
+        let mut headers = options
+            .get_extension::<http::HeaderMap>()
+            .cloned()
+            .unwrap_or_default();
+
+        // Strip any custom headers that collide with system headers or credential headers.
+        for key in cred_headers
```
joshuatants: Can you double check this logic in the gRPC path (https://github.com/googleapis/google-cloud-rust/blob/35b95901b68b407fc39d4cb8b51df320ab5d9898/src/gax-internal/src/grpc/grpc_helpers.rs#L78)? If `Authorization` is suppled as a custom header, in HTTP it'll be removed, but it doesn't look like it'll be removed there.

joshuatants (changes requested):

src/gax-internal/src/http.rs
xlai20: `x-goog-user-project` is removed at line 435.

src/gax/src/options.rs
xlai20: No code changes are needed for this, as standard repeated headers work out-of-the-box using RFC-compliant comma-separated strings:
 
 * Under HTTP RFC 9110 (Section 5.3), standard repeated headers like `Cache-Control`, `Accept`, or `Allow` are defined as comma-separated lists (e.g., `Cache-Control: no-cache, no-store`).
 * Callers can supply repeated values natively as a single comma-separated `HeaderValue` without needing multiple calls:
   ```rust
   builder.with_custom_header(
       HeaderName::from_static("cache-control"),
       HeaderValue::from_static("no-cache, no-store"),
   )
   ```
 * I checked other language SDK implementations (such as Python), and they follow this same pattern.

src/gax-internal/tests/http_custom_header.rs
xlai20: No, this entire test file `src/gax-internal/tests/http_custom_header.rs` is *not* testing the `RequestOptionsBuilder` function in `gax/src/options.rs`. Instead, it is an integration test for the HTTP wire-level transport logic in `src/gax-internal/src/http.rs`.

The function `with_custom_header` here is a test helper function that inserts custom headers into the extensions. The test verifies that when `client.execute(..., options)` runs, `ReqwestClient::request`:
 - Retrieves the custom `HeaderMap` from `RequestOptions` extensions
 - Sanitizes and strips conflicting system/auth headers
 - Correctly merges custom headers with credential headers and system headers on the wire.

src/gax/src/options.rs
xlai20: Done, added detailed explanation to the `with_custom_header` function.

src/gax/src/options.rs
xlai20: Done. 

I chose to implement your suggested Option 1 to shift HeaderName and HeaderValue to the function signature, letting callers to do the validation. It's better than Option 2 as it preserves Fluent Builder ergonomics.

src/gax-internal/src/http.rs
xlai20: Done.

I align the gRPC path and HTTP/JSON path as follows:
1. Ensure the order of "adding authorization headers" and "adding custom headers" in both paths are the same: add custom headers, strip away system headers, then add authorization headers. Previously the order is different in gRPC and JSON, resulting in different implementations in both sites; now they are the same.
2. In the sanitization code, now I add "authorization" and other authorization-related headers into the list of system headers. They would all be stripped away from user-input custom headers, before any kinds of authorization/system headers are added.
3. The current implementation is aligned with other language SDKs.

src/gax-internal/tests/http_custom_header.rs
xlai20: Yes, added it and explained in another comment.

src/gax/src/options.rs
```diff
@@ -221,6 +221,37 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
     {
         unimplemented!();
     }
+
+    /// Injects a custom HTTP header (or gRPC metadata) into this specific request.
+    ///
+    /// This method is dedicated to adding custom headers. Callers cannot use this
+    /// function to inject system or authentication headers (such as `user-agent`,
+    /// `authorization`, `x-goog-api-key`, `x-goog-api-client`, `x-goog-user-project`,
+    /// or `x-goog-request-params`); any attempt to do so will be silently ignored.
```
coryan: Consider:

```suggestion
    /// This method will ignore attempts to set headers used by the client
    /// library. These include: such as `user-agent`,
    /// `authorization`, `x-goog-api-key`, `x-goog-api-client`, `x-goog-user-project`,
    /// and `x-goog-request-params`).
```

Saying what the method does upfront is easier to reason about.

src/gax/src/options.rs
```diff
@@ -221,6 +221,37 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
...
+    ///
+    /// Callers who want to configure respective system headers should use the
+    /// dedicated builder methods instead, such as [`with_quota_project`](Self::with_quota_project)
+    /// and [`with_user_agent`](Self::with_user_agent).
```
coryan: If I am a customer, how do I find out the full list of headers controlled by "dedicated builder methods" ? What happens in the future if we start using `x-goog-server-timeout` (a real possibility)?

It may be easier if we enforced something like:

> This method ignores any `x-goog-*` headers, as well as `authorization` and `user-agent`.
> Use `with_user-agent` to set the user agent header.
> Use the credentials to set the authorization header.

That is something we can easily enforce, and maybe not silently.

coryan (changes requested):
In general, I like the changes.

The documentation needs improving, and trying to write the docs suggests improvements in the behavior and future guarantees for this.

src/gax/src/options.rs
xlai20: Done. 

Edited the function comment as suggested.

src/gax/src/options.rs
xlai20: 1. Regarding the `x-goog-*` headers, I opted not to strictly block them in the code to maintain parity with other language SDKs. Specifically:
   - Python, Go, and C++ do not actively block or drop `x-goog-*` custom headers.
   - In Python, injecting headers like `x-goog-custom-audit-job` is actually a common pattern.
   - In C++, the SDK explicitly uses `x-goog-emulator-instructions` as a test case for custom headers.

2. I've updated the function documentation to explicitly map each system header to its "dedicated builder method." This exhaustively guides users on how to properly configure these headers and clarifies which ones are auto-generated and cannot be modified.

src/gax/src/options.rs
joshuatants: Can the docstring for this function be updated to specify that that is how repeated headers should be passed in?

src/gax/src/options.rs
joshuatants: Nice, thank you

joshuatants (approved):

src/gax/src/options.rs
xlai20: Done. 

In my latest commit "[document repeated custom headers in with_custom_header](https://github.com/googleapis/google-cloud-rust/pull/6260/commits/6ec12a1a9e527eed30e300363026647a9159e03e)" I've added an extra paragraph explaining to user on how to pass in repeated headers.

coryan (approved):

PR #6277, opened 2026-08-04
impl(bigquery): regenerate with skipped fields on librarian.yaml file

alvarowolfx:
With https://github.com/googleapis/librarian/pull/7189 we can move skipped field configuration to librarian.yaml

Towards #5844


librarian.yaml
```diff
@@ -490,6 +490,21 @@ libraries:
         - output: src/bigquery/src/generated
           api_path: google/cloud/bigquery/v2
           template: bigquery
+          skipped_ids:
```
suzmue, line 491: This doesn't result in a code change in bigquery, is that expected?

librarian.yaml
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 language: rust
-version: v0.31.2-0.20260803194200-96709e948366
+version: v0.31.2-0.20260804190831-b07a8a79b63a
```
suzmue, line 15: Since this results in changes in shopping, maybe this bump as a separate PR?

librarian.yaml
suzmue: Answer: yes it is, support for skipping was moved to librarian.yaml

suzmue (approved):

PR #6279, opened 2026-08-05
chore(spanner): implement location-aware routing key recipe evaluation and caching

olavloite:
Add `key_recipe.rs` and `key_recipe_cache.rs` to support client-side evaluation of Spanner storage specification routing keys for location-aware routing.


olavloite:
Closing this PR, as it is getting too big. I will use a different strategy to try to reduce the size of it.

PR #6291, opened 2026-08-05
chore: bump MSRV to 1.90

alvarowolfx:
It is more than a year since 1.88 and 1.89 turns 1 year old next Friday. So jumping to 1.90 as MSRV.

There is still a `extract_if()` cleanup to do in PubSub, but will leave for a separated PR.


.gcb/builds/triggers/main.tf
```diff
@@ -138,10 +138,8 @@ locals {
       script = "test"
     }
     test-msrv = {
-      config       = "complex.yaml"
-      flags        = local.unstable_flags
-      script       = "test"
-      rust_version = "1.88"
+      config = "msrv.yaml"
+      flags  = local.unstable_flags
     }
```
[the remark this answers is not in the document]
alvarowolfx: most configs are not passing `rust_version` and it's not being a problem, to GCB is handling `null` properly.

Cargo.toml
```diff
@@ -398,7 +398,7 @@ license      = "Apache-2.0"
 repository   = "https://github.com/googleapis/google-cloud-rust/tree/main"
 keywords     = ["gcp", "google-cloud", "google-cloud-rust", "sdk"]
 categories   = ["network-programming"]
-rust-version = "1.88.0"
+rust-version = "1.90.0"
```
coryan, line 407: The trigger is still using the 1.88 docker image.  You need to manually update the trigger first (using the old school method) and then send this PR.

coryan: And then update the trigger again to use the new `msrv.yaml` file.

alvarowolfx: yeah, make sense. I was afraid of pushing the terraform updates and breaking other PR CI checks in the meantime while this PR is not merged. Will sync with the team to do so.

coryan (approved):

PR #6292, opened 2026-08-05
impl(showcase): update bidi to use request builder

suzmue:
Update the librarian version to get these changes. The integration was updated to use the new request builder.

For #2318


suzmue:
FYI @alvarowolfx the query_creation_metadata.rs file is added in this PR but is not included in the build.

alvarowolfx:
> FYI @alvarowolfx the query_creation_metadata.rs file is added in this PR but is not included in the build.

cool, and that's expected. After this PR I'll send one to wire it up. Thanks

alvarowolfx (approved):

Issue #6198, opened 2026-07-28
Implement `attach_job` method on BigQuery Query Client


alvarowolfx:
Allow to get a `Query` handle from an existing query job.

```
impl BigQuery {
    /// Re-attach hook to bind an existing out-of-process job reference.
    pub async fn attach_job(
       &self,
       job_ref: google_cloud_bigquery_v2::model::JobReference,
    ) -> Result<Query>;
}
```

alvarowolfx:
done via https://github.com/googleapis/google-cloud-rust/pull/6338

PR #6293, opened 2026-08-05
impl(bigquery): support default `project_id` on `ClientBuilder`

haphungw:
Add support for configuring a default GCP `project_id` on `ClientBuilder`, which is stored on the `BigQuery` client and automatically used as the billing project when running queries or attaching to jobs.

For https://github.com/googleapis/google-cloud-rust/issues/6198

### DIFF
--- a/src/bigquery/src/client.rs
+++ b/src/bigquery/src/client.rs
@@ -28,6 +28,7 @@ use std::sync::Arc;
 ///
 /// Common configuration customizations include:
 ///
+/// - [`with_project_id()`][crate::builder::bigquery::ClientBuilder::with_project_id]: Sets the default Google Cloud project ID for the client.
 /// - [`with_endpoint()`][crate::builder::bigquery::ClientBuilder::with_endpoint]: Overrides the default API endpoint (`https://bigquery.googleapis.com`). Useful when testing against mock servers or running in restricted network environments (for example, with VPC Service Controls).
 /// - [`with_credentials()`][crate::builder::bigquery::ClientBuilder::with_credentials]: Overrides the default Application Default Credentials with explicit or custom authentication credentials.
 ///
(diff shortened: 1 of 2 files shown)


src/bigquery/src/client.rs
```diff
@@ -55,14 +56,21 @@ impl BigQuery {
...
 
     /// Execute a SQL query.
     ///
     /// This builder internally routes to either `jobs.query` (fast path) or `jobs.insert` (job path)
     pub fn query<S: Into<String>>(&self, sql: S) -> RunQuery {
-        RunQuery::new(self.job_service.clone(), sql.into())
+        let mut run_query = RunQuery::new(self.job_service.clone(), sql.into());
+        if let Some(project_id) = &self.project_id {
+            run_query = run_query.with_project_id(project_id);
+        }
```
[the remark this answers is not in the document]
haphungw: done

suzmue (approved):

src/bigquery/src/client.rs
```diff
@@ -150,7 +154,11 @@ impl BigQuery {
     /// # }
     /// ```
     pub fn query<S: Into<String>>(&self, sql: S) -> RunQuery {
-        RunQuery::new(self.job_service.clone(), sql.into())
+        let mut run_query = RunQuery::new(self.job_service.clone(), sql.into());
+        if let Some(project_id) = self.project_id.as_deref() {
+            run_query = run_query.with_project_id(project_id);
+        }
+        run_query
```
alvarowolfx: nit: we can avoid `mut` here:

```suggestion
        let builder = RunQuery::new(self.job_service.clone(), sql.into());
        self.project_id
            .as_deref()
            .into_iter()
            .fold(builder, |builder, project_id| builder.with_project_id(project_id))
```

alvarowolfx (approved):

src/bigquery/src/client_builder.rs
```diff
@@ -46,9 +47,16 @@ impl ClientBuilder {
     pub fn new() -> Self {
         Self {
             config: ClientConfig::default(),
+            project_id: None,
         }
     }
 
+    /// Sets the default Google Cloud project ID for the client.
+    pub fn with_project_id<V: Into<String>>(mut self, project_id: V) -> Self {
```
alvarowolfx, line 66: this is missing example like in the other methods

alvarowolfx (changes requested):
I think we can add a example section on the new `with_project_id` method.

src/bigquery/src/client_builder.rs
haphungw: done

src/bigquery/src/client.rs
haphungw: done

alvarowolfx (approved):

PR #6307, opened 2026-08-07
ci: move MSRV build to a dedicated definition

alvarowolfx:
Move MSRV check to have it's own GCB file, so we don't need to push Terraform updates on every MSRV bump, just update version on that file.


.gcb/msrv.yaml
```diff
@@ -0,0 +1,58 @@
...
+    name: 'rust:${_RUST_VERSION}-bookworm'
+    args: ['.gcb/scripts/cargo-fetch.sh']
+    volumes:
+      - name: 'fetched'
+        path: /builder/home/.cargo
+  - name: 'rust:${_RUST_VERSION}-bookworm'
+    volumes:
+      - name: 'fetched'
+        path: /builder/home/.cargo
+    args: ['.gcb/scripts/test.sh']
```
coryan: Note to self: I was going to suggest keeping the variable here, but YAGNI.

coryan (approved):

.gcb/msrv.yaml
alvarowolfx: yeah, I was going back and forth on that. I think I'm gonna keep it to keep all GCB files more consistent

coryan (approved):

PR #6309, opened 2026-08-07
impl(bigquery): wire up QueryCreationMetadata

alvarowolfx:
Wire up generated QueryCreationMetadata and use it internally. `Query::metadata()` to be added in a future PR, since we are probably gonna remove `query_reference`.

Towards #5844

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -497,8 +497,7 @@ libraries:
             - .google.cloud.bigquery.v2.QueryRequest.kind
             - .google.cloud.bigquery.v2.JobConfiguration.kind
             - .google.cloud.bigquery.v2.JobConfiguration.job_type
-            - .google.cloud.bigquery.v2.GetQueryResultsResponse.rows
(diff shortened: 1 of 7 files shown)


librarian.yaml
```diff
@@ -497,8 +497,7 @@ libraries:
             - .google.cloud.bigquery.v2.QueryRequest.kind
             - .google.cloud.bigquery.v2.JobConfiguration.kind
             - .google.cloud.bigquery.v2.JobConfiguration.job_type
-            - .google.cloud.bigquery.v2.GetQueryResultsResponse.rows
-            - .google.cloud.bigquery.v2.QueryResponse.rows
+            - rows
```
alvarowolfx, line 500: had to change this because I made a mistake on the QueryCreationMetadata generation on sidekick and it's not checking for full field ID.

There is a PR to fix (https://github.com/googleapis/librarian/pull/7232), but it's harmless to move to check by name and and move to ID again back later.

suzmue (approved):

PR #6313, opened 2026-08-10
test(gax): add tests for bidi_stream and bidi_stream_with_status

joshuatants:
For #5991.

Also remove previous test-only helper function `invoker`.

### DIFF
--- a/src/gax-internal/src/grpc/grpc_rust.rs
+++ b/src/gax-internal/src/grpc/grpc_rust.rs
@@ -82,11 +82,6 @@ impl GrpcRustClient {
         Self::build(config, default_endpoint, Some(instrumentation)).await
     }
 
-    // TODO(#5991): Temporary helper for testing. Remove once `bidi_stream` is implemented.
-    pub fn invoker(&self) -> &Channel {
-        &self.inner.invoker
-    }
-
     pub async fn execute<Request, Response>(
         &self,
         _extensions: Extensions,
@@ -423,7 +418,6 @@ fn make_tls_credentials() -> ClientBuilderResult<Arc<RustlsChannelCredentials>> 
         })
 }
 
-// TODO(#5991): Add integration tests for `GrpcRustClient::bidi_stream` and `GrpcRustClient::bidi_stream_with_status` (covering happy paths and request stream failures).
 #[cfg(test)]
 mod tests {
     use super::*;
--- a/src/gax-internal/src/grpc/grpc_rust/bidi.rs
+++ b/src/gax-internal/src/grpc/grpc_rust/bidi.rs
@@ -246,12 +246,29 @@ mod tests {
         value: String,
     }
 
-    struct TestSendStream {
+    #[derive(Clone, Default)]
+    struct MockSendStream {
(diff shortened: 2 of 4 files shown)


src/gax-internal/tests/grpc_rust_client_streaming.rs
```diff
@@ -34,177 +33,331 @@ mod tests {
...
-        // Consume and assert initial headers emitted upon RPC start
-        let first_item = rx
-            .recv()
-            .await
-            .expect("response pump should yield at least one item")?;
+        let TestBidiSession {
+            tx,
+            mut stream,
+            metadata,
+            ..
```
vsharonlynn: In Rust, when we destructure a struct like this, the unmentioned fields (e.g. _server_task, _client) will be dropped, right? So these tasks might continue running in the background, and any panic or failure in the background will be silently ignored.

src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -420,4 +419,74 @@ mod tests {
         );
         Ok(())
     }
+
+    #[tokio::test]
+    async fn bidi_call_yields_error_on_server_error_status() -> anyhow::Result<()> {
```
vsharonlynn, line 493: Quick check of my understanding:
IIUC in this test case, the error is only observed when `stream.message().await` is called.

So this doesn't test the immediate error check in `invoke_bidi` on [line 214](https://github.com/googleapis/google-cloud-rust/pull/6313/changes#diff-2f17ccdd28f5b24b883fdc2648d69041804b8a103bc26e2cc32d924fc2cda0c7R214).

Do you want to test the case where invoke_bidi returns an error immediately?

src/gax-internal/src/grpc/grpc_rust.rs
```diff
@@ -423,7 +418,6 @@ fn make_tls_credentials() -> ClientBuilderResult<Arc<RustlsChannelCredentials>>
         })
 }
 
-// TODO(#5991): Add integration tests for `GrpcRustClient::bidi_stream` and `GrpcRustClient::bidi_stream_with_status` (covering happy paths and request stream failures).
```
vsharonlynn, line 426: I think the unit test cases do not cover the case of request stream failure? Do we need to test this case?

vsharonlynn (approved):

src/gax-internal/tests/grpc_rust_client_streaming.rs
joshuatants: You're right. 

Changed to store the returned struct in a variable instead.

src/gax-internal/src/grpc/grpc_rust/bidi.rs
joshuatants: Good catch. Added a new test `bidi_call_returns_error_on_immediate_trailers_only_status`.

src/gax-internal/src/grpc/grpc_rust.rs
joshuatants: Done. Added a test `bidi_call_immediately_fails_if_initial_request_fails`

PR #6315, opened 2026-08-10
chore(gax-internal): add attempt start and complete lifecycle hooks to AttemptInterceptor

olavloite:
Extend the internal AttemptInterceptor trait with on_attempt_start and on_attempt_complete callbacks to allow tracking per-attempt latency, metadata, and response headers across unary gRPC attempts.

Bump google-cloud-gax-internal version to 0.7.17.

Needed for https://github.com/googleapis/google-cloud-rust/pull/6311

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3185,7 +3185,7 @@ dependencies = [
 
 [[package]]
 name = "google-cloud-gax-internal"
-version = "0.7.16"
+version = "0.7.17"
 dependencies = [
  "anyhow",
  "bytes",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -523,7 +523,7 @@ tokio-stream      = { default-features = false, version = "0.1.16" }
 # Local packages used as dependencies.
 google-cloud-auth        = { default-features = false, version = "1.15.0", path = "src/auth" }
 google-cloud-gax         = { default-features = false, version = "1.13.0", path = "src/gax" }
-gaxi                     = { default-features = false, version = "0.7.16", path = "src/gax-internal", package = "google-cloud-gax-internal" }
+gaxi                     = { default-features = false, version = "0.7.17", path = "src/gax-internal", package = "google-cloud-gax-internal" }
 wkt                      = { default-features = false, version = "1.7.0", path = "src/wkt", package = "google-cloud-wkt" }
 google-cloud-wkt         = { default-features = false, version = "1.7.0", path = "src/wkt", package = "google-cloud-wkt" }
 google-cloud-api         = { default-features = false, version = "1.7.0", path = "src/generated/api/types" }
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -904,7 +904,7 @@ libraries:
     copyright_year: "2025"
     output: src/gax
   - name: google-cloud-gax-internal
-    version: 0.7.16
+    version: 0.7.17
     copyright_year: "2025"
     output: src/gax-internal
     rust:
--- a/src/gax-internal/Cargo.toml
+++ b/src/gax-internal/Cargo.toml
@@ -14,7 +14,7 @@
 
 [package]
 name        = "google-cloud-gax-internal"
-version     = "0.7.16"
+version     = "0.7.17"
 description = "Google Cloud Client Libraries for Rust - Implementation Details"
 build       = "build.rs"
 # Inherit other attributes from the workspace.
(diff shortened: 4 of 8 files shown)


coryan (commented):
> to allow tracking per-attempt latency,

We already have a way to do that.... Look at the implementation of transport-level metrics. This may be a good place to start:

https://github.com/googleapis/google-cloud-rust/blob/main/src/gax-internal/src/observability/client_signals/with_transport_metric.rs

Can you examine the other approach (maybe you already did) and let us know why it does not work in your case?

olavloite:
> > to allow tracking per-attempt latency,
> 
> We already have a way to do that.... Look at the implementation of transport-level metrics. This may be a good place to start:
> 
> https://github.com/googleapis/google-cloud-rust/blob/main/src/gax-internal/src/observability/client_signals/with_transport_metric.rs
> 
> Can you examine the other approach (maybe you already did) and let us know why it does not work in your case?

Yeah, looked into that possibility (sorry, I should have added that directly to the PR description). The short answer is 'While not utterly __impossible__, it would require major changes to `WithTransportMetric`'. The main reasons are:
1. For Spanner, we need to capture some relatively specific metrics and also include header values that are returned by Spanner, so we can compare end-to-end latency with the latency that was measured by the GFE. This allows us to make a reasonable judgement call on whether end-to-end latency is coming from Spanner (i.e. the GFE is also reporting high latency), or whether the end-to-end latency is caused by something 'after' the GFE (network, client, ...).
2. We need to capture metrics per attempt, and not only per overall RPC. Meaning: We need a hook into the gRPC retry loop.


----

Or to quote a verbose friend of mine:

### 1. Why `WithTransportMetric` Cannot Be Used As-Is

`WithTransportMetric` and `TransportMetric` are designed specifically for standard Google Cloud Client Signals (T4 transport signals):

1. **Hardcoded Metric Definition**:
   - `TransportMetric` is hardcoded to emit `gcp.client.attempt.duration` in seconds (`s`) with 16 fixed bucket boundaries to the global `opentelemetry::global::meter_provider()`.
   - Spanner Built-In Metrics requires emitting 8 distinct metrics (`attempt_latencies`, `operation_latencies`, `gfe_latencies`, `afe_latencies`, `attempt_count`, `operation_count`, `gfe_connectivity_error_count`, `afe_connectivity_error_count`) in **milliseconds (`ms`)**, with Spanner's explicit 50-bucket distribution (`0.0ms` to `3,200,000.0ms`), under a custom monitored resource (`spanner_instance_client`).
2. **Missing Response Headers in `RequestRecorder`**:
   - Spanner Built-In Metrics requires measuring GFE and AFE latencies and connectivity error counts by parsing the `server-timing` headers returned in gRPC response metadata.
   - `WithTransportMetric` reads from `RequestRecorder` (`tokio::task_local!` storing `TransportSnapshot`), which captures URL, status code, and duration, but does not capture gRPC response metadata/headers on successful attempts.

---

### 2. What Would It Take to Adapt `WithTransportMetric`?

If we tried to route Spanner Built-In Metrics through `WithTransportMetric` / `RequestRecorder`, it would require significant architectural changes across GAX:

1. **Refactoring `RequestRecorder`**:
   - Extend `RequestRecorder` and `TransportSnapshot` in `gax-internal` to capture and store gRPC response `HeaderMap` metadata on every attempt.
   - Update `grpc.rs` to extract response headers from `tonic::Response` on every attempt and write them into the task-local `RequestRecorder`.
2. **Pluggable Metric Emitters in GAX**:
   - `WithTransportMetric` currently only calls `TransportMetric.with_recorder_ok()` and `with_recorder_error()`.
   - It would need to be generalized to support custom metric listeners or service-specific metric pipelines.
3. **Retry Loop Visibility**:
   - Intermediate attempt results are encapsulated within `grpc::Client::execute`. To notify Spanner of intermediate attempt outcomes (for attempt-level metric recording), GAX still needs to provide an attempt-level hook/callback during the retry loop.

---

### Summary Comparison

| Requirement | `WithTransportMetric` / `RequestRecorder` | `AttemptInterceptor` Lifecycle Hooks |
|---|---|---|
| **Response Headers (`server-timing`)** | ❌ Not captured in `TransportSnapshot` | ✅ Directly passed as `Option<&HeaderMap>` |
| **Spanner Metric Suite (8 instruments)** | ❌ Hardcoded to `gcp.client.attempt.duration` | ✅ Delegated directly to Spanner's `Observability` |
| **Custom Monitored Resource (`spanner_instance_client`)** | ❌ Hardcoded to generic OTel resource conventions | ✅ Managed by Spanner's `SdkMeterProvider` |
| **Intermediate Attempt Notification** | ❌ Encapsulated inside GAX retry loop | ✅ Direct callback per attempt |

coryan (approved):

PR #6317, opened 2026-08-10
impl(bigquery): remove QueryReference in favor of QueryCreationMetadata

alvarowolfx:
Previously was assumed that QueryID and JobID were mutually exclusive (they're once, but now it isn't anymore). QueryReference was created with that in mind, so it doesn't make sense anymore and customers can access QueryID and JobID via QueryCreationMetadata now, giving access to the full RPC responses, without additional abstractions. 

Towards #5844


src/bigquery/examples/src/query/job_optional.rs
```diff
@@ -35,20 +34,21 @@ pub async fn sample(project_id: &str) -> anyhow::Result<()> {
...
-        }
-        _ => {}
+    let metadata = query.metadata();
+    if !metadata.query_id.is_empty() {
+        println!(
+            "Query was run in optional job mode. Query ID: \"{}\"",
+            metadata.query_id
+        );
+    }
+    if let Some(job) = &metadata.job_reference {
```
dbolduc, line 44: comment: it is a tiny bit sad that the mutual exclusivity of these things is not encoded in the type. I think it's fine though.

tests/bigquery/src/lib.rs
```diff
@@ -235,12 +234,9 @@ pub async fn query_client() -> Result<()> {
...
     // BigQuery client sets JobCreationMode::JobCreationOptional by default
-    let query_ref = query.query_reference();
-    let QueryReference::Stateless { ref query_id } = query_ref else {
-        anyhow::bail!("expected a stateless query reference, got {query_ref:?}");
-    };
-
-    assert!(!query_id.is_empty(), "{query_ref:?}");
+    let metadata = query.metadata();
+    let query_id = &metadata.query_id;
+    assert!(!query_id.is_empty(), "expected non-empty query_id");
```
dbolduc, line 239: optional nit:

```suggestion
    assert!(!metadata.query_id.is_empty(), "expected non-empty query_id");
```

dbolduc (approved):

src/bigquery/examples/src/query/job_optional.rs
alvarowolfx: yeah, unfortunately was an error from the service side to not declared it nullable. It's a hard call to us to add an override to change the type, even that I would like to.

PR #6318, opened 2026-08-10
impl(bigquery): job level retry for queries

alvarowolfx:
When a query job fails with a retryable BigQuery error (for example `backendError` or `rateLimitExceeded`), the client now automatically creates a new job attempt with backoff instead of failing immediately. Move all logic for query execution away from builder layer (Query) to the execution layer. 

Towards #5844 #6218


src/bigquery/src/query/builder.rs
```diff
@@ -35,6 +33,7 @@ pub struct RunQuery {
     pub(crate) job_service: Arc<JobService>,
     pub(crate) request: RunQueryRequest,
     pub(crate) project_id: Option<String>,
+    pub(crate) job_retry_policy: Arc<dyn JobRetryPolicy>,
```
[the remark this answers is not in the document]
alvarowolfx: will do that in a separated PR. Need to figure out how to add `#[derive(Debug)]` on the generated types

src/bigquery/src/retry_policy.rs
```diff
@@ -107,13 +105,11 @@ impl JobRetryResult {
 /// `RetryPolicy` trait specifically tailored for BigQuery job-level re-issuance.
 /// In the future, we plan to refine this trait and discuss how to make it public
 /// so customers can provide their own custom job retry policies.
-#[allow(dead_code)]
 pub(crate) trait JobRetryPolicy<S = RetryState>: Send + Sync + std::fmt::Debug {
     fn on_error(&self, state: &S, error: QueryError) -> JobRetryResult;
 }
 
 #[derive(Clone, Debug)]
-#[allow(dead_code)]
```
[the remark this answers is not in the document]
alvarowolfx: will fix in a separated PR

src/bigquery/src/query/execution.rs
```diff
@@ -107,11 +100,104 @@ impl InsertJobExecutor {
...
+            let job_config: JobConfiguration = self.template.request.clone().into();
+            let job_ref = generate_job_reference(project_id, &self.template.request.location);
+            let job = Job::new()
+                .set_configuration(job_config)
+                .set_job_reference(job_ref);
+            let req = InsertJobRequest::new()
+                .set_job(job)
+                .set_project_id(project_id);
+
+            let job = Box::pin(InsertJobExecutor::new(job_service.clone(), req).execute()).await?;
```
suzmue: Could we add a brief comment explaining why we're using Box::pin here? (e.g. boxing the RPC execution future to keep the future / stack frame size small across await points in the retry loop)

src/bigquery/src/query/execution.rs
```diff
@@ -107,11 +100,104 @@ impl InsertJobExecutor {
...
+
+            let res = Box::pin(PostQueryExecutor::new(job_service.clone(), req).execute()).await?;
+
+            Ok(Query::from_query_response(
+                job_service,
+                res,
+                Some(self.clone()),
+                max_results,
+            ))
+        }
```
suzmue, line 158: nit: consider refactoring to make it extra self documenting
```suggestion
        if self.template.request.force_job_path() {
           self.execute_insert_job(project_id)
        } else {
            self.execute_post_query(project_id)
        }
```

src/bigquery/src/query/run_query.rs
```diff
@@ -66,49 +66,15 @@ impl RunQuery {
...
-                .await
-        }
+        RetryContext::new(self).execute().await
     }
 }
 
 // Create a job reference with a generated job ID.
 //
 // BigQuery does not strictly define a format for job IDs, just a limit in size.
 // See https://docs.cloud.google.com/bigquery/docs/reference/rest/v2/JobReference
```
suzmue, line 178: nit: convert this to a doc comment
```suggestion
/// Create a job reference with a generated job ID.
///
/// BigQuery does not strictly define a format for job IDs, just a limit in size.
/// See https://docs.cloud.google.com/bigquery/docs/reference/rest/v2/JobReference
```

src/bigquery/src/query/run_query.rs
```diff
@@ -128,7 +94,7 @@ fn generate_job_reference(project_id: &str, location: &str) -> JobReference {
 //
 // See https://docs.cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#QueryRequest for request ID
 // and https://docs.cloud.google.com/bigquery/docs/reference/rest/v2/JobReference for job ID.
```
suzmue, line 198: nit: convert this to a doc comment
```suggestion
/// Create a random ID with the given prefix and a UUID.
///
/// BigQuery does not strictly define a format for request and job IDs, just a limit in size.
/// However, request IDs are more restrictive and have a limit of 36 characters.
/// UUID v4 simple format is used to reduce length.
///
/// See https://docs.cloud.google.com/bigquery/docs/reference/rest/v2/jobs/query#QueryRequest for request ID
/// and https://docs.cloud.google.com/bigquery/docs/reference/rest/v2/JobReference for job ID.
```

src/bigquery/src/query/query_handle.rs
```diff
@@ -106,41 +113,62 @@ impl Query {
...
+                    metadata,
+                    cached_rows,
+                    max_results,
+                ));
+            }
+
+            let job_ref = metadata
+                .job_reference
+                .as_ref()
+                .expect("query job should have job reference at this point");
```
suzmue, line 220: I am a little concerned about this expect. If this job_reference is coming from the server (even if we set it in every request) I think we should error here instead. If self was purely constructed by us and guaranteed to have the job_reference I think its fine to panic (because it would really be a problem in our code), but would maybe add an extra comment saying why you believe that this really cannot panic.

src/bigquery/src/query/query_handle.rs
```diff
@@ -106,41 +113,62 @@ impl Query {
...
+
+            let backoff_policy = Arc::new(
+                ExponentialBackoffBuilder::default()
+                    .with_initial_delay(std::time::Duration::from_secs(10))
```
suzmue, line 224: nit; magic numbers, should we make this a const?

src/bigquery/src/query/query_handle.rs
```diff
@@ -106,41 +113,62 @@ impl Query {
...
+                        res,
+                        max_results,
+                    ));
+                }
+                Err(err) => {
+                    let Some(retry_ctx) = retry_context else {
+                        return Err(err);
+                    };
+                    match retry_ctx.on_error(err) {
+                        JobRetryResult::Continue(delay, _) => {
```
suzmue, line 243: nit: will this eventually be used? If so maybe a TODO here. If not maybe a comment at the Continue() site for why it will not be used ever.

src/bigquery/src/query/execution.rs
suzmue, line 98: Is it possible for s.error_result to include an error but errors to not? 

It looks like this code drops any error in error_result. When retrying, we are then only looking at "errors" (and then not retrying because it is empty.

suzmue (approved):

src/bigquery/src/query/execution.rs
alvarowolfx: this section is on `jobs.insert` call, which only has `error_result`. The `errors` array is for `jobs.query` path and for when jobs need to be polled via `jobs.getQueryResults`, where both have a `errors` array.

src/bigquery/src/query/execution.rs
alvarowolfx: ack

src/bigquery/src/query/execution.rs
alvarowolfx: good call

src/bigquery/src/query/query_handle.rs
alvarowolfx: this is really an impossible state. We are basically in a point where the query is not complete (still running) and we don't have a way to reference it to poll it.

src/bigquery/src/query/query_handle.rs
alvarowolfx: the `Error` on `Continue` is just to mirror the same types in gax. We are not gonna make that public yet, so we can change it later

PR #6322, opened 2026-08-11
test(bigquery): test basic arrow writes against prod

### LINKED ISSUE #5311: Bootstrap BigQuery Write API
Add the generated gRPC code for the BigQuery storage write API.

We are not married to these decisions. But my initial plan is to:

- Develop this under `src/bigquery`
  - Read + write + query/jobs veneers all live in this crate
- Continue using `grpc_mock` strategy for mocking the service.
  - Aside: it might be time to teach sidekick how to make this thing

---

cc: @yirutang, @yifatgortler

### LINKED ISSUE #5665: Write an integration test for writing data to bigquery
akin to this [java sample](https://github.com/googleapis/java-bigquerystorage/blob/44534dd37bcee0f89c8ecf158b140d1d7a1f2e1a/samples/snippets/src/main/java/com/example/bigquerystorage/WriteToDefaultStreamWithArrow.java)

dbolduc:
Add an integration test using the write client to stream data to BQ in Arrow format over the default stream.

### Organization

After this PR, I will do a pass to tidy up the `lib.rs`. It is a bit too unruly for my taste.

### Arrow types

It is definitely verbose to construct the schema and record batches. I know that `serde_arrow` is a thing, but (1) it is unstable and (2) I am not sure it is efficient. I am having the robot look into whether it can simplify the example code with it.

Fixes #5665, fixes #5311

### DIFF
--- a/tests/bigquery/src/writes.rs
+++ b/tests/bigquery/src/writes.rs
@@ -1,0 +1,88 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+mod arrow;
+
+use super::*;
+use google_cloud_bigquery_v2::client::TableService;
+use google_cloud_bigquery_v2::model::{Table, TableFieldSchema, TableReference, TableSchema};
+
+pub async fn run_writes() -> Result<()> {
+    let project_id = project_id()?;
+    let dataset_service = DatasetService::builder().with_tracing().build().await?;
+    cleanup_stale_datasets(&dataset_service, &project_id).await?;
+
+    let dataset_id = random_dataset_id();
+    let _ = create_dataset(&dataset_service, &project_id, &dataset_id).await?;
+
+    let table_service = TableService::builder().with_tracing().build().await?;
+    let table_id = random_table_id();
+    let _ = create_table(&table_service, &project_id, &dataset_id, &table_id).await?;
+
+    let result = arrow::basic(&project_id, &dataset_id, &table_id).await;
+
+    let _ = delete_dataset(&dataset_service, &project_id, &dataset_id).await;
+    result
+}
+
+async fn create_table(
+    table_service: &TableService,
+    project_id: &str,
+    dataset_id: &str,
+    table_id: &str,
+) -> Result<()> {
+    println!("CREATING TABLE WITH ID: {table_id}");
+    let schema = TableSchema::new().set_fields([
+        TableFieldSchema::new().set_name("name").set_type("STRING"),
+        TableFieldSchema::new().set_name("age").set_type("INTEGER"),
+    ]);
+    table_service
+        .insert_table()
+        .set_project_id(project_id)
+        .set_dataset_id(dataset_id)
+        .set_table(
+            Table::new()
+                .set_table_reference(
+                    TableReference::new()
+                        .set_project_id(project_id)
+                        .set_dataset_id(dataset_id)
+                        .set_table_id(table_id),
+                )
+                .set_schema(schema),
+        )
+        .send()
+        .await?;
+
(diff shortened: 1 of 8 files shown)


tests/bigquery/src/writes.rs
```diff
@@ -0,0 +1,88 @@
...
+    let _ = create_dataset(&dataset_service, &project_id, &dataset_id).await?;
+
+    let table_service = TableService::builder().with_tracing().build().await?;
+    let table_id = random_table_id();
+    let _ = create_table(&table_service, &project_id, &dataset_id, &table_id).await?;
+
+    let result = arrow::basic(&project_id, &dataset_id, &table_id).await;
+
+    let _ = delete_dataset(&dataset_service, &project_id, &dataset_id).await;
+    result
```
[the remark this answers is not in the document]
dbolduc: you're right. fixed.

tests/bigquery/src/writes/arrow.rs
```diff
@@ -0,0 +1,94 @@
...
+    let _ = StreamWriter::try_new(&mut buf, schema)?;
+    Ok(buf)
+}
+
+fn serialize_batch(batch: &RecordBatch, schema_len: usize) -> Result<Vec<u8>> {
+    let mut buf = Vec::new();
+    let mut writer = StreamWriter::try_new(&mut buf, &batch.schema())?;
+    writer.write(batch)?;
+    Ok(buf[schema_len..].to_vec())
+}
```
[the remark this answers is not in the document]
dbolduc: Added a comment explaining why this is safe.

tests/bigquery/src/writes/arrow.rs
```diff
@@ -0,0 +1,96 @@
...
+    Ok(buf)
+}
+
+fn serialize_batch(batch: &RecordBatch, schema_len: usize) -> Result<Vec<u8>> {
+    let mut buf = Vec::new();
+    let mut writer = StreamWriter::try_new(&mut buf, &batch.schema())?;
+    writer.write(batch)?;
+    // Note that the schema is encoded in the front of the record batch. We need
+    // to strip it.
+    Ok(buf[schema_len..].to_vec())
```
alvarowolfx, line 95: fun fact: this is defined on the IPC spec and I tried to convince BQ folks to support streams with the schema encoded in the beginning of the stream, but they rejected that. Is a bit sad because a lot of other tools use IPC format by default and to stream to BQ we have to remove the schema and send just the row data.

tests/bigquery/src/writes/arrow.rs
```diff
@@ -0,0 +1,96 @@
...
+use std::sync::Arc;
+
+pub async fn basic(project_id: &str, dataset_id: &str, table_id: &str) -> Result<()> {
+    let table = format!("projects/{project_id}/datasets/{dataset_id}/tables/{table_id}");
+
+    // Create a Schema
+    let arrow_schema = Arc::new(Schema::new(vec![
+        Field::new("name", DataType::Utf8, false),
+        Field::new("age", DataType::Int64, false),
+    ]));
```
alvarowolfx, line 31: will be cool later to use our own internal helpers to convert from TableSchema -> ArrowSchema

tests/bigquery/src/writes.rs
```diff
@@ -0,0 +1,91 @@
...
+                )
+                .set_schema(schema),
+        )
+        .send()
+        .await?;
+
+    Ok(())
+}
+
+async fn read_table(project_id: &str, dataset_id: &str, table_id: &str) -> Result<Vec<UserRecord>> {
```
alvarowolfx, line 73: later sounds like this can be even more generic to read tables as a `T` that derives FromRow

alvarowolfx (approved):

tests/bigquery/src/writes.rs
dbolduc: Yeah, maybe with abstracting the query out too. (e.g. I order on the name field so the results are deterministic).

I might need the generic read when we get to the JSON writer.

tests/bigquery/src/writes/arrow.rs
dbolduc: can't wait. :smile:

tests/bigquery/src/writes/arrow.rs
dbolduc: Dumb Q: Is the IPC spec the only way to encode this stuff?

Should I be using something else that doesn't include the schema at the beginning? (which has to be more efficient, but maybe takes some extra code).

I think this PR is fine as-is for an integration test, but at some point I am going to need to write a public sample, where we will want to guide customers towards the right thing

PR #6338, opened 2026-08-11
impl(bigquery): add `attach_job` method to query client

haphungw:
Add `BigQuery::attach_job` to bind an existing out-of-process query job reference (`JobReference`) to a high-level `Query` handle without re-submitting SQL.

When #6295 was originally merged, it conflicted with #6317 (which was merged before and refactored `Query` struct fields to use `QueryCreationMetadata`). This caused post-merge build failures because `attach_job` referenced the old struct fields.

This PR re-adds the changes with fixes to be compatible with `QueryCreationMetadata`.

### DIFF
--- a/src/bigquery/src/client.rs
+++ b/src/bigquery/src/client.rs
@@ -14,8 +14,10 @@
 
 use crate::ClientBuilderResult as BuilderResult;
 use crate::client_builder::ClientBuilder;
-use crate::query::RunQuery;
+use crate::error::QueryError;
+use crate::query::{Query, Result as QueryResult, RunQuery};
 use google_cloud_bigquery_v2::client::JobService;
+use google_cloud_bigquery_v2::model::JobReference;
 use std::sync::Arc;
 
 /// A high-level BigQuery client for executing queries and managing jobs.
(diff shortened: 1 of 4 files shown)


src/bigquery/src/client.rs
```diff
@@ -163,12 +165,60 @@ impl BigQuery {
...
+        }
+        if job_ref.project_id.is_empty() {
+            let Some(proj) = &self.project_id else {
+                return Err(QueryError::InvalidArgument(
+                    "no project ID was provided".to_string(),
+                ));
+            };
+            job_ref.project_id = proj.clone();
+        }
+
```
alvarowolfx, line 212: I missed something really important on the previous PR. Here we need to fetch the `Job` using `job_service.get_job`. And them call existing `Query::from_job`. This way the Query::metadata has the inital job metadata

alvarowolfx (changes requested):

src/bigquery/src/error.rs
```diff
@@ -21,9 +21,9 @@ use google_cloud_gax::error::Error;
 #[derive(thiserror::Error, Debug)]
 #[non_exhaustive]
 pub enum QueryError {
-    /// The project ID was not provided or could not be determined.
-    #[error("no project ID was provided")]
-    MissingProjectId,
+    /// An invalid argument was provided (e.g. missing job ID or project ID).
+    #[error("invalid argument: {0}")]
+    InvalidArgument(String),
```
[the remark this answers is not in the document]
alvarowolfx: I'm still on a fence on this subject and was going to bring it on the API review, but the bot brought this argument and similar to @suzmue said, having a generic InvalidArgument is not really useful and I'm leaning more towards having `MissingProjectId` and `MissingJobId`, because they are actionable in specific ways. The problem is that later if we have another required argument, are we going to keep adding `MissingX` ?

src/bigquery/src/client.rs
```diff
@@ -163,12 +165,60 @@ impl BigQuery {
...
+        Ok(Query::from_job_reference(
+            self.job_service.clone(),
+            job_ref,
+            None,
+        ))
+    }
```
alvarowolfx: we could do something like this and not even have a explicit check for Job IDs.

```suggestion
    pub async fn attach_job(
        &self,
        job_ref: google_cloud_bigquery_v2::model::JobReference,
    ) -> Result<Query> {
        let project_id = job_ref.project_id.clone();
        let project_id = if project_id.is_empty() {
            self.project_id.clone()
        } else {
            Some(project_id)
        };
        let project_id = project_id.ok_or(QueryError::MissingProjectId)?;
        
        let req = self
            .job_service
            .get_job()
            // maybe we don't even have to check for Job ID and just let it throws a 404 if Job Id is empty ?
            // GCS does that and let it fail on the service level
            .set_job_id(job_ref.job_id.clone())
            .set_project_id(project_id.clone());
        

        let req = job_ref
            .location
            .into_iter()
            .fold(req, |req, location| req.set_location(location));

        let job = req.send().await?;

        Ok(Query::from_job(self.job_service.clone(), job, None, None))
    }
```

haphungw: done

src/bigquery/src/error.rs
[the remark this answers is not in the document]
haphungw: That's fair.

Now that we no longer handle `job_id`, it's not a problem for this PR anymore, but I think we should still consider this for future cases?

alvarowolfx (approved):

PR #6356, opened 2026-08-12
impl(bigquery): hide set_request_id on query builder

alvarowolfx:
Initially we are not gonna allow customers to set a request id or job_id on the query level. The client will control idempotency and generate those ids.

Towards #6218

### DIFF
--- a/src/bigquery/src/generated/run_query_request.rs
+++ b/src/bigquery/src/generated/run_query_request.rs
@@ -34,7 +34,7 @@ pub struct RunQueryRequest {
     /// Clustering specification for the destination table.
     pub clustering: std::option::Option<crate::model::Clustering>,
 
-    /// Connection properties which can modify the query behavior.
+    /// Optional. Connection properties which can modify the query behavior.
     pub connection_properties: std::vec::Vec<crate::model::ConnectionProperty>,
 
     /// Optional. Whether to run the query as continuous or a regular query.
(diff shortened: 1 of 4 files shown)


src/bigquery/src/generated/run_query_request.rs
```diff
@@ -107,13 +107,12 @@ pub struct RunQueryRequest {
...
-    /// than 60 seconds to complete has a better chance of being stopped than a job
-    /// that takes 10 seconds to complete. This timeout applies to the query even
-    /// if a job does not need to be created.
-    pub job_timeout_ms: std::option::Option<i64>,
+    /// Optional. Job timeout in milliseconds relative to the job creation time. If
+    /// this time limit is exceeded, BigQuery attempts to stop the job, but might
+    /// not always succeed in canceling it before the job completes. For example, a
+    /// job that takes more than 60 seconds to complete has a better chance of
+    /// being stopped than a job that takes 10 seconds to complete.
+    pub job_timeout_ms: std::option::Option<wkt::Int64Value>,
```
dbolduc: Q: why is this type changing? This would be a break if we were GA.

dbolduc (approved):

src/bigquery/src/generated/run_query_request.rs
alvarowolfx: I thought that was due to some changes in librarian. Now I need to double check

alvarowolfx: oh, it's bouncing between JobConfiguration:: job_timeout_ms and QueryRequest::job_timeout_ms, which has different types. Totally forgot that case, I ended handling just for bools. Need to solve that

alvarowolfx: Actually the way that I decided to solve that is to prioritize fields on the order of the messages that are being merged (https://github.com/googleapis/librarian/blob/5b51a25eeaa2be4ff5a6b6045281d36e077665dd/internal/sidekick/rust/bigquery.go#L105), in this case, QueryRequest field should take priority. But I think as I went refactoring some stuff, that behavior got lost, specially here when fields are being dedup:

https://github.com/googleapis/librarian/blob/5b51a25eeaa2be4ff5a6b6045281d36e077665dd/internal/sidekick/rust/bigquery.go#L70

alvarowolfx: using `slices.SortStableFunc` do the trick here:
https://github.com/googleapis/librarian/blob/5b51a25eeaa2be4ff5a6b6045281d36e077665dd/internal/sidekick/rust/bigquery.go#L67

will send a PR

alvarowolfx: Fix for this behavior https://github.com/googleapis/librarian/pull/7255

PR #6361, opened 2026-08-12
refactor(bigquery): reorg query integration tests

dbolduc:
Move tests of the query client into its own module

Follow up to #6344, #6345


tests/bigquery/src/writes.rs
```diff
@@ -28,7 +32,7 @@ pub async fn run_writes() -> Result<()> {
     let _ = create_dataset(&dataset_service, &project_id, &dataset_id).await?;
 
     let table_service = TableService::builder().with_tracing().build().await?;
-    let table_id = random_table_id();
+    let table_id = "writes";
```
dbolduc, line 35: Note that the table belongs to a dataset with a random name. There is one table per dataset. So the name will not clash.

suzmue (approved):

PR #6340, opened 2026-08-11
docs(bigquery): add core job lifecycle samples

haphungw:
Demonstrate how to manage the lifecycle of BigQuery Jobs. These samples poll `JobService` to track status and metadata.

Added:
- `bigquery_create_job`: Creates and executes a job.
- `bigquery_get_job`: Retrieves the execution status and metadata of a job by its ID.
- `bigquery_list_jobs`: Lists recently executed jobs in the project. 
- `bigquery_cancel_job`: Cancels an active analytical query job.

Fixes #5762, fixes #5761

PR #2832, opened 2025-08-07
fix: empty body is object for PUT, POST

dbolduc:
Fixes #2823 

No integration tests because typically when you create a resource, it will error on the default body. It basically always expects some sort of ID.

But the difference in the error statuses are:

```
                status: Status {
                    code: InvalidArgument,
                    message: "Invalid JSON payload received. Unknown name \"\": Root element must be a message.",
                    details: [
                        BadRequest(
                            BadRequest {
                                field_violations: [
                                    FieldViolation {
                                        field: "",
                                        description: "Invalid JSON payload received. Unknown name \"\": Root element must be a message.",
                                        reason: "",
                                        localized_message: None,
                                    },
                                ],
                            },
                        ),
                    ],
                },
```

vs.

```
                status: Status {
                    code: InvalidArgument,
                    message: "Following fields are missing or invalid: Display Name, External Company ID. (Possible reasons: non-numeric for double field, invalid language code, etc.). Request ID for tracking: d51e19c2-edde-401a-8d45-9540b597ac31:APAb7IQowEf1xfF5gvMw7aI1IWLFxa8mlQ==",
                    details: [
                        DebugInfo(
                            DebugInfo {
                                stack_entries: [],
                                detail: <omitted...>,
                            },
                        ),
                    ],
                },
```

Issue #2823, opened 2025-08-06
Improve error message when the POST body is empty


coryan:
The service error message can be confusing when a request is missing the body for a `POST` request.  For example, calling [PublicCertificateAuthorityService::create_external_account_key](https://docs.rs/google-cloud-security-publicca-v1/0.4.3/google_cloud_security_publicca_v1/client/struct.PublicCertificateAuthorityService.html#method.create_external_account_key) returns:

```rust
Err(
    Error {
        kind: Service(
            ServiceDetails {
                status_code: Some(
                    400,
                ),
                headers: Some(
                    {
                        "x-google-esf-cloud-client-params": "backend_service_name: \"publicca.googleapis.com\" backend_fully_qualified_method: \"google.cloud.security.publicca.v1.PublicCertificateAuthorityService.CreateExternalAccountKey\"",
                        "x-google-session-info": "GgQYECgLIAE6GRIXcHVibGljY2EuZ29vZ2xlYXBpcy5jb20",
                        "vary": "X-Origin",
                        "vary": "Referer",
                        "vary": "Origin,Accept-Encoding",
                        "content-type": "application/json; charset=UTF-8",
                        "date": "Wed, 06 Aug 2025 19:08:44 GMT",
                        "server": "ESF",
                        "x-google-netmon-label": "/bns/ug/borg/ug/bns/tarsier-publicca-service/tarsier-publicca-prod.frontend/2",
                        "x-google-security-signals": "FRAMEWORK=HTTPSERVER2,BUILD=GOOGLE3,BUILD_DEBUG=cl:767553382,ENV=borg,ENV_DEBUG=borg_user:tarsier-publicca-service;borg_job:tarsier-publicca-prod.frontend",
                        "x-xss-protection": "0",
                        "x-frame-options": "SAMEORIGIN",
                        "x-content-type-options": "nosniff",
                        "x-google-gfe-service-trace": "tarsier-publicca-service-publicca-service-prod",
                        "x-google-backends": "/bns/ug/borg/ug/bns/tarsier-publicca-service/tarsier-publicca-prod.frontend/2,/bns/lclgaa/borg/lclgaa/bns/blue-layer1-gfe-prod-edge/prod.blue-layer1-gfe.lga34s37/95",
                        "x-google-gfe-request-trace": "aclgaee2:443,/bns/ug/borg/ug/bns/tarsier-publicca-service/tarsier-publicca-prod.frontend/2,aclgaee2:443",
[...]

PR #6364, opened 2026-08-12
fix(gax-internal): serialize `NoBody` as empty JSON object

haphungw:
Fix an edge case from #2823 where empty `POST`/`PUT` requests were failing with `Root element must be a message` on BigQuery's `cancel_job` https://github.com/googleapis/google-cloud-rust/pull/6340#discussion_r3764327581.

#2832 updated `gaxi::http::handle_empty` to return `Some(NoBody::default())` to make sure a payload was sent. However, since `NoBody` was defined as `pub struct NoBody;`, `serde_json` serializes it as "null". Our APIs expect an empty JSON object `{}` instead.

To fix this without breaking semver, we added a custom serialization to explicitly map `NoBody` to an empty size `0` struct, forcing the JSON payload out as `{}`. This resolves the empty JSON problem while maintaining backward compatibility.

### DIFF
--- a/src/gax-internal/src/http.rs
+++ b/src/gax-internal/src/http.rs
@@ -576,8 +576,19 @@ pub fn map_send_error(err: ::reqwest::Error) -> Error {
     }
 }
 
-#[derive(Default, serde::Serialize)]
+#[derive(Default)]
 pub struct NoBody;
+
+impl serde::Serialize for NoBody {
+    fn serialize<S>(&self, serializer: S) -> std::result::Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        use serde::ser::SerializeMap;
+        let map = serializer.serialize_map(Some(0))?;
+        map.end()
+    }
+}
 
 pub fn handle_empty<T: Default>(body: Option<T>, method: &Method) -> Option<T> {
     body.or_else(|| {
@@ -638,6 +649,13 @@ async fn to_http_response<O: serde::de::DeserializeOwned + Default>(
 
 #[cfg(test)]
 mod tests {
+    #[test]
+    fn nobody_serializes_to_empty_object() -> anyhow::Result<()> {
+        let serialized = serde_json::to_string(&super::NoBody)?;
+        assert_eq!(serialized, "{}");
+        Ok(())
+    }
+
     use super::*;
     use crate::options::ClientConfig;
     use crate::options::InstrumentationClientInfo;


src/gax-internal/src/http.rs
```diff
@@ -1026,4 +1039,10 @@ mod tests {
         assert_eq!(result.err().unwrap().http_status_code(), Some(308));
         Ok(())
     }
+
+    #[test]
+    fn nobody_serializes_to_empty_object() {
+        let serialized = serde_json::to_string(&super::NoBody).unwrap();
+        assert_eq!(serialized, "{}");
+    }
```
dbolduc: nit:
```suggestion
    fn nobody_serializes_to_empty_object() -> anyhow::Result<()> {
        let serialized = serde_json::to_string(&super::NoBody)?;
        assert_eq!(serialized, "{}");
        Ok(())
    }
```

dbolduc (approved):
Thanks!

coryan (changes requested):
> Fix an edge case from #2823 where empty `POST`/`PUT` requests were failing with `Root element must be a message` on BigQuery's `cancel_job` [#6340 (comment)](https://github.com/googleapis/google-cloud-rust/pull/6340#discussion_r3764327581).

Ack.

> #2832 updated `gaxi::http::handle_empty` to return `Some(NoBody::default())` to make sure a payload was sent. However, since `NoBody` was defined as `pub struct NoBody;`, `serde_json` serializes it as "null". Our APIs expect an empty JSON object `{}` instead.

Can the pass `Some(wkt::Empty)` to these POST/PUT requests (maybe through changes in the generator)?  The definition of `NoBody` was intentional, it was supposed to represent "this is a request that has no body".  The BigQuery requests seem to have a body. They need a different type.

dbolduc:
> > #2832 updated `gaxi::http::handle_empty` to return `Some(NoBody::default())` to make sure a payload was sent. However, since `NoBody` was defined as `pub struct NoBody;`, `serde_json` serializes it as "null". Our APIs expect an empty JSON object `{}` instead.

My bad. I think I was only testing for a body that was a message field that didn't have presence.

https://github.com/googleapis/googleapis/blob/29da218f022fcd6af34142e9128699ed04820a11/google/cloud/security/publicca/v1/service.proto#L48

I did not test for a PUT/POST RPC without a `body` annotation.

https://github.com/googleapis/googleapis/blob/29da218f022fcd6af34142e9128699ed04820a11/google/cloud/bigquery/v2/job.proto#L57-L59

> Can the pass `Some(wkt::Empty)` to these POST/PUT requests (maybe through changes in the generator)? The definition of `NoBody` was intentional, it was supposed to represent "this is a request that has no body". The BigQuery requests seem to have a body. They need a different type.

My understanding is that we pass in `None` to represent "this is a request that has no body" and `Some(NoBody)` to mean a body with an empty object `{}`

https://github.com/googleapis/google-cloud-rust/blob/2a279e6a600f3655c69b0f79bc1c543ab02e5a9d/src/gax-internal/src/http.rs#L271-L273

I agree that we could/should use `wkt::Empty` and avoid `NoBody` altogether, but I also want to accept this PR as it is a legitimate bug fix.

src/gax-internal/src/http.rs
haphungw: done

dbolduc (approved):

coryan (approved):

PR #6374, opened 2026-08-13
impl(bigquery): rename run to send and add until_done shortcut to run queries

alvarowolfx:
Based on API review discussion, we thought that was better to rename the run method to send, to be more in line with other methods in the SDK and still fits in the BigQuery scenario. Also added a until_done shortcut to the query builder, which run and polls the query in sequence.

Towards #5844


src/bigquery/src/query/run_query.rs
```diff
@@ -188,6 +188,38 @@ impl RunQuery {
...
+    /// Sends the query execution request and polls until execution completes.
+    ///
+    /// This is a convenience method equivalent to calling
+    /// [`.send().await?.until_done().await`](RunQuery::send).
+    ///
+    /// # Errors
+    ///
+    /// Returns [`QueryError::MissingProjectId`](crate::error::QueryError::MissingProjectId)
+    /// if no project ID was configured, or an error if query execution or polling fails.
+    ///
```
dbolduc: This is changing. Remove the section?

src/bigquery/src/query/run_query.rs
```diff
@@ -188,6 +188,38 @@ impl RunQuery {
...
+    /// let mut rows = completed_query.read();
+    /// if let Some(row) = rows.next().await.transpose()? {
+    ///     let now: String = row.get("now");
+    ///     println!("Current time: {now}");
+    /// }
+    /// # Ok(())
+    /// # }
+    /// ```
+    pub async fn until_done(self) -> Result<CompleteQuery> {
```
dbolduc, line 213: cool

dbolduc (approved):

PR #6376, opened 2026-08-13
test: deflake `go run ..librarian`

coryan:
Minimize flakes by making multiple attempts to download `librarian` and its dependencies.

Towards #6373

### DIFF
--- a/.gcb/format.yaml
+++ b/.gcb/format.yaml
@@ -121,8 +121,13 @@ steps:
         go install github.com/google/yamlfmt/cmd/yamlfmt@v0.21.0
       git ls-files -z -- '*.yaml' '*.yml' ':!:testdata/**' ':!:**/generated/**' ':!:librarian.yaml' | \
         xargs -0 yamlfmt
+      go run github.com/googleapis/librarian/cmd/librarian@latest config get version || \
+        go run github.com/googleapis/librarian/cmd/librarian@latest config get version || \
+        go run github.com/googleapis/librarian/cmd/librarian@latest config get version
       V=$(go run github.com/googleapis/librarian/cmd/librarian@latest config get version)
-      go run github.com/googleapis/librarian/cmd/librarian@${V} tidy
+      go run github.com/googleapis/librarian/cmd/librarian@${V} tidy || \
+      go run github.com/googleapis/librarian/cmd/librarian@${V} tidy || \
+        go run github.com/googleapis/librarian/cmd/librarian@${V} tidy
     waitFor: ['-']
     env:
       - GOTOOLCHAIN=auto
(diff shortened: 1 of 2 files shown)


.gcb/scripts/regenerate.sh
```diff
@@ -36,6 +36,10 @@ rustup show active-toolchain -v
 
 echo "Regenerate all the code"
 version=$(sed -n 's/^version: *//p' /workspace/librarian.yaml)
+# Make multiple download attempts to avoid flakes
+go run github.com/googleapis/librarian/cmd/librarian@${version} >/dev/null ||
+  go run github.com/googleapis/librarian/cmd/librarian@${version} >/dev/null ||
+  go run github.com/googleapis/librarian/cmd/librarian@${version} >/dev/null
```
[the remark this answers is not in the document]
coryan: Fixed

.gcb/scripts/regenerate.sh
```diff
@@ -36,6 +36,10 @@ rustup show active-toolchain -v
 
 echo "Regenerate all the code"
 version=$(sed -n 's/^version: *//p' /workspace/librarian.yaml)
+# Make multiple download attempts to avoid flakes
+go run github.com/googleapis/librarian/cmd/librarian@${version} help >/dev/null ||
```
dbolduc: why not `go install` ?

.gcb/format.yaml
```diff
@@ -121,8 +121,13 @@ steps:
         go install github.com/google/yamlfmt/cmd/yamlfmt@v0.21.0
       git ls-files -z -- '*.yaml' '*.yml' ':!:testdata/**' ':!:**/generated/**' ':!:librarian.yaml' | \
         xargs -0 yamlfmt
+      go run github.com/googleapis/librarian/cmd/librarian@latest config get version || \
```
dbolduc: why fetch `latest` AND the version in `librarian.yaml` ? Shouldn't we just do the version in `librarian.yaml` ?

dbolduc: and same Q about `go install` instead of `go run ...`

.gcb/scripts/regenerate.sh
coryan: Fixed.

.gcb/format.yaml
coryan: > why fetch `latest` AND the version in `librarian.yaml` ? Shouldn't we just do the version in `librarian.yaml` ?

Done. And left some comments explaining why we use `sed` here and document `librarian config get version` everywhere else.

dbolduc (approved):
Thanks

.gcb/format.yaml
dbolduc: Ah. We had to download some version of `librarian` in order to determine which version of `librarian` we need to download. :upside_down_face:

PR #6377, opened 2026-08-13
docs(gax): fix `with_custom_header()` sample

coryan:
The sample was marked `ignore`. The generated uses `ignore` to speed up the PRs. One of the CI builds compiles post-merge.

Towards #6373

### DIFF
--- a/src/gax/src/options.rs
+++ b/src/gax/src/options.rs
@@ -331,10 +331,14 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
     /// comma-separated [`HeaderValue`]:
     ///
     /// ```rust,ignore
+    /// # use google_cloud_gax::options::RequestOptionsBuilder;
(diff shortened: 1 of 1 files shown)


src/gax/src/options.rs
```diff
@@ -331,10 +331,14 @@ pub trait RequestOptionsBuilder: internal::RequestBuilder {
     /// comma-separated [`HeaderValue`]:
     ///
     /// ```rust,ignore
```
dbolduc: We can just run the sample, right?

```suggestion
    /// ```
```

If it doesn't compile, the PM build would keep failing anyway.

dbolduc (approved):
Thanks.

Also, I'll just leave this here: https://github.com/googleapis/google-cloud-rust/issues/5500

src/gax/src/options.rs
coryan: Done.

PR #6390, opened 2026-08-14
fix(lro): reduce stack usage

### LINKED ISSUE #6389: Reduce stack size of LRO futures
## Problem

We have massive futures for LROs. Taking up space on the stack can lead to stack overflows in practice.

For example, `cargo clippy -p integration-tests-lro --all-features -- -W clippy::large_futures` flags:

```
warning: large future with a size of 31392 bytes
  --> tests/lro/src/production.rs:47:20
   |
47 |       let response = client
   |  ____________________^
48 | |         .create_workflow()
49 | |         .set_parent(format!("projects/{project_id}/locations/{location_id}"))
50 | |         .set_workflow_id(&workflow_id)
...  |
58 | |         .poller()
59 | |         .until_done()
```

## Analysis

There are several opportunities for reducing the size:

1. `start` captures the initial request builder. We can box it and save the size of messages.

The same is true for `query`, but `GetOperation` is a fixed size that is not sooo big. The real gains are in boxing `start`.

https://github.com/googleapis/google-cloud-rust/blob/5dbecc17c5302b125e95cf7a32752279ddb6d24b/src/lro/src/internal/aip151.rs#L56

2. The `poll()` future is 1-2 kB, we can pin that.

https://github.com/googleapis/google-cloud-rust/blob/5dbecc17c5302b125e95cf7a32752279ddb6d24b/src/lro/src/lib.rs#L177

3. The result of the poll, `p` is held across await points. We can make sure it is out of scope.

https://github.com/googleapis/google-cloud-rust/blob/5dbecc17c5302b125e95cf7a32752279ddb6d24b/src/lro/src/lib.rs#L178

## Results

With all these changes, I can get the size of the future from `31392` down to `2712` bytes. 😏

dbolduc:
(see the linked issue for the problem write up)

Nobody will ever notice that we made some extra heap allocations on a longrunning operation. They already take a while.

Fixes #6389

### DIFF
--- a/src/lro/src/internal/aip151.rs
+++ b/src/lro/src/internal/aip151.rs
@@ -31,6 +31,11 @@ use super::LroRecorder;
 
 pub type Operation<R, M> = crate::details::Operation<R, M>;
 
+type BoxedStartFuture<R, M> =
+    std::pin::Pin<Box<dyn std::future::Future<Output = Result<Operation<R, M>>> + Send>>;
+
+type BoxedQuery<R, M> = Box<dyn Fn(String) -> BoxedStartFuture<R, M> + Send + Sync>;
+
 /// Creates a new `impl Poller<R, M>` from the closures created by the generator.
 ///
 /// This is intended as an implementation detail of the generated clients.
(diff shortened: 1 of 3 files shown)


src/lro/src/lib.rs
```diff
@@ -174,25 +174,18 @@ where
...
+            Some(PollingResult::InProgress(_)) => (),
+            Some(PollingResult::PollingError(_)) => (),
+            // `poll()` only returns `None` after it returned `Polling::Completed`
+            // therefore this is never reached.
+            None => unreachable!("loop should exit via the `Completed` branch vs. this line"),
+        };
         state.attempt_count += 1;
-        poller.backoff(&state).await
+        Box::pin(poller.backoff(&state)).await;
     }
```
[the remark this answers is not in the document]
dbolduc: I considered this and it saved ~1.5 kB to keep in the boxing.

Given that this is a longrunning operation I am not going to sweat some heap allocations.

src/lro/src/internal/aip151.rs
```diff
@@ -42,18 +47,27 @@ pub fn new_poller<ResponseType, MetadataType, S, SF, Q, QF>(
...
         + Send
         + 'static,
-    Q: Fn(String) -> QF + Send + Sync + Clone,
+    Q: Fn(String) -> QF + Send + Sync + 'static,
     QF: std::future::Future<Output = Result<Operation<ResponseType, MetadataType>>>
         + Send
         + 'static,
 {
-    PollerImpl::new(polling_error_policy, polling_backoff_policy, start, query)
+    let start_boxed =
```
coryan: nit: just call it `start`

coryan (approved):

src/lro/src/internal/aip151.rs
dbolduc: Done.

PR #6393, opened 2026-08-14
fix(pubsub): make stream opening cancellation safe

Zarathustra2:
Create the keep alive cancellation guard before awaiting StreamingPull so
canceling an in progress open does not orphan its keep alive task.

------------------

Given the following setup where you either stream messages or on a timeout do some other task like flushing.

```rust
loop {
    select! {
        message = stream.next() => process(message),
        _ = timer.tick() => flush_messages(),
    }
}
```

When the timer wins it can happen that we have already spawned the keep alive task before we have created the cancellation task since we are still awaiting the StreamingPull. 

```rust
// Currently on main
let shutdown = CancellationToken::new();

// spawn keep alive task
keepalive::spawn(request_tx, shutdown.clone());

// Await opening the stream. Now if the task gets canceled
// we will never cancel the keepalive task
// because we have not yet created the cancellation task which happens
// when we return.
//
// The fix is to create the keepalive_guard before we spawn the keep alive task.
let stream = inner
        .streaming_pull(&request_params, request_rx, RequestOptions::default())
        .await?
        .into_inner();

Ok(Stream {
    _keepalive_guard: shutdown.drop_guard(),
    _keepalive_guard: keepalive_guard,
    stream,
})

```

As a result we now have an orphaned keep alive task that has not been canceled.

### DIFF
--- a/src/pubsub/src/subscriber/stream.rs
+++ b/src/pubsub/src/subscriber/stream.rs
@@ -115,7 +115,8 @@ where
     //
     // [^1]: https://github.com/hyperium/tonic/issues/515
     let shutdown = CancellationToken::new();
-    keepalive::spawn(request_tx, shutdown.clone());
+    let keepalive_guard = shutdown.clone().drop_guard();
+    keepalive::spawn(request_tx, shutdown);
 
     let stream = inner
         .streaming_pull(&request_params, request_rx, RequestOptions::default())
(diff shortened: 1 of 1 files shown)


coryan:
@suzmue can you take a look at this? I am not Pub/Sub-savvy enough to be of any use.

dbolduc (approved):
Makes perfect sense. Thank you for the fix and the test!

coryan (approved):

PR #6395, opened 2026-08-14
refactor(bigquery): rename query metadata and builder

alvarowolfx:
Based on API reviews, we decided to simplify some names:

* QueryCreationMetadata -> QueryMetadata
* QueryMetadata -> CompleteQueryMetadata
* RunQuery -> Query
* RunQueryRequest -> QueryRequest

Towards #5844

### DIFF
--- a/src/bigquery/src/generated/builder.rs
+++ b/src/bigquery/src/generated/builder.rs
@@ -18,9 +18,9 @@
 #[allow(rustdoc::broken_intra_doc_links)]
 #[allow(rustdoc::invalid_html_tags)]
 #[allow(rustdoc::redundant_explicit_links)]
-impl QueryBuilder {
+impl Query {
 
-    /// Sets the value of [request.allow_large_results][crate::generated::QueryBuilderRequest::allow_large_results].
+    /// Sets the value of [request.allow_large_results][crate::builder::bigquery::QueryRequest::allow_large_results].
     pub fn set_allow_large_results<T>(mut self, v: T) -> Self
     where T: std::convert::Into<wkt::BoolValue>
     {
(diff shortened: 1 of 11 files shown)


alvarowolfx:
depends on https://github.com/googleapis/librarian/pull/7285

src/bigquery/src/generated/builder.rs
```diff
@@ -18,41 +18,41 @@
 #[allow(rustdoc::broken_intra_doc_links)]
 #[allow(rustdoc::invalid_html_tags)]
 #[allow(rustdoc::redundant_explicit_links)]
-impl RunQuery {
+impl Query {
 
-    /// Sets the value of [request.allow_large_results][crate::model::RunQueryRequest::allow_large_results].
+    /// Sets the value of [request.allow_large_results][crate::builder::bigquery::QueryRequest::allow_large_results].
```
suzmue, line 23: request is a private field of Query, so I would consider dropping `request.` or making it `QueryRequest.allow_large_results`
```suggestion
    /// Sets the value of [allow_large_results][crate::builder::bigquery::QueryRequest::allow_large_results].
```

suzmue (approved):

src/bigquery/src/generated/builder.rs
alvarowolfx: it's not that easy to change since both places in template uses `Codec.FieldName`. So I would have to change all templates to have new fields for the comment vs field accessor.

* https://github.com/googleapis/librarian/blob/main/internal/sidekick/rust/templates/common/setter_preamble/singular_value.mustache#L16
* https://github.com/googleapis/librarian/blob/f51a5f9543f50ba59aae726783a7f9dd9d9f1e3b/internal/sidekick/rust/templates/common/setters.mustache#L32

PR #6409, opened 2026-08-14
impl(bigquery): remove continuous and incremental query options

alvarowolfx:
Remove ways for customers to run continuous and incremental queries for now. They will need special handling on the client SDK which is not worth to tackle right now.

Towards #5844


dbolduc (approved):
comment: In the future, we are never going to notice when BQ adds a veneer-only field like this 🤷

alvarowolfx:
> comment: In the future, we are never going to notice when BQ adds a veneer-only field like this 🤷

I was thinking about this and wondering if we should make this list a include list instead of a skip list, so we can avoid that. But them, we have to keep that list up to date with the latest and it can drift with time.

PR #6411, opened 2026-08-14
test(showcase): add bidi streaming integration tests

suzmue:
Add integration tests for bidirectional streaming using the Showcase service. The tests verify full-duplex streaming, client half-close via sender drop, clean server EOF handling, server-side error propagation, early receiver drop cancellation, and custom headers/channel options.


tests/showcase/src/streaming.rs
```diff
@@ -0,0 +1,139 @@
...
+use google_cloud_gax::error::rpc::Code;
+use google_cloud_gax::options::RequestOptionsBuilder as _;
+use google_cloud_showcase_v1beta1::client::Echo;
+use google_cloud_showcase_v1beta1::model::EchoRequest;
+
+pub async fn run() -> Result<()> {
+    let client = Echo::builder()
+        .with_endpoint("http://localhost:7469")
+        .with_credentials(Anonymous::new().build())
+        .with_retry_policy(NeverRetry)
```
coryan, line 25: That may fail, even in CI with a local server... Consider something simple like retrying I/O errors

coryan: Or maybe it is interesting to see the failure.

coryan (approved):

tests/showcase/src/streaming.rs
suzmue: This is what we use for other showcase tests, and AFAIK havent had problems. It seems like it will be interesting to see the failure.

PR #6415, opened 2026-08-15
cleanup(storage): skip troublesome field

### LINKED ISSUE #6353: Generate StorageControl's `ViewObjectFullContext` RPC
This RPC uses a message which has a `google.protobuf.Any` field, which has all sorts of problems over gRPC:

https://github.com/googleapis/googleapis/blame/668dd277e2e0915aa7bcf6164993fced2ba9f5a8/google/storage/control/v2/storage_control.proto#L2556

A reasonable solution would be to skip the field in the model. I am not sure if librarian supports this yet.

Until then, we will omit the RPC and the offending message.

coryan:
Skipping the field is cleaner than skipping the Sidekick-codegen <-> Prost-codegen conversion. If nothing else, the application developer get a clear signal that they cannot set the field, as there is no field to set.

Fixes #6353

### DIFF
--- a/src/generated/showcase/src/transport.rs
+++ b/src/generated/showcase/src/transport.rs
@@ -1829,7 +1829,7 @@ impl super::stub::Echo for Echo {
                 req_stream,
                 options.into(),
                 &crate::info::X_GOOG_API_CLIENT_HEADER,
-                &x_goog_request_params,
+                x_goog_request_params,
             )
             .await?;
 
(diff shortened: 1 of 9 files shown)


dbolduc (approved):
Nice

src/generated/showcase/src/transport.rs
```diff
@@ -1798,6 +1798,8 @@ impl super::stub::Echo for Echo {
         let req =
             req.ok_or_else(|| google_cloud_gax::error::Error::binding("a request is required"))?;
 
+        let x_goog_request_params = "";
```
coryan: This is explained by https://github.com/googleapis/librarian/pull/7273 rolled into the librarian update.

PR #6421, opened 2026-08-17
test(gax): add simple unary integration tests for GrpcRustClient

joshuatants:
For #5991.

Mirrors tests from `src/gax-internal/tests/grpc_simple_request.rs`:
- `default_endpoint`
- `no_request_params`
- `override_endpoint`

Omitted:
- `multiple_endpoints`: subchannel connection pooling (`grpc_subchannel_count`) is not supported by `GrpcRustClient`.
- `non_default_endpoint_ipv6`: currently `#[ignore]`'d in `grpc_simple_request.rs`.
- default `user-agent` header checks: grpc-rust does not inject a default "tonic/..." user-agent.

### DIFF
--- a/src/gax-internal/tests/grpc_rust_simple_request.rs
+++ b/src/gax-internal/tests/grpc_rust_simple_request.rs
@@ -26,7 +26,7 @@ mod tests {
     use grpc_server::{google, start_echo_server};
     use pretty_assertions::assert_eq;
 
-    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    #[tokio::test]
     async fn default_endpoint() -> anyhow::Result<()> {
         // Arrange
         let (endpoint, _server) = start_echo_server().await?;
--- a/src/gax-internal/tests/grpc_rust_simple_request.rs
+++ b/src/gax-internal/tests/grpc_rust_simple_request.rs
@@ -30,8 +30,7 @@ mod tests {
     async fn default_endpoint() -> anyhow::Result<()> {
         // Arrange
         let (endpoint, _server) = start_echo_server().await?;
-        let client =
-            GrpcRustClient::new(test_config(endpoint), "https://storage.googleapis.com").await?;
+        let client = GrpcRustClient::new(test_config(), &endpoint).await?;
 
         // Act & Assert
         check_simple_request(client).await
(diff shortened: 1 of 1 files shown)


src/gax-internal/tests/grpc_rust_simple_request.rs
```diff
@@ -0,0 +1,132 @@
...
+        // Arrange
+        let (endpoint, _server) = start_echo_server().await?;
+        let client =
+            GrpcRustClient::new(test_config(endpoint), "https://test-only.googleapis.com").await?;
+
+        // Act
+        let response = send_request(client, "test message", "").await?;
+
+        // Assert
+        assert_eq!(
```
vsharonlynn, line 50: Nit: For this case, no need to assert the x-goog-api-client and message body?

src/gax-internal/tests/grpc_rust_simple_request.rs
```diff
@@ -0,0 +1,132 @@
...
+    use google_cloud_auth::credentials::{Credentials, anonymous::Builder as Anonymous};
+    use google_cloud_gax::options::RequestOptions;
+    use google_cloud_gax::retry_policy::NeverRetry;
+    use google_cloud_gax_internal::grpc::GrpcRustClient;
+    use google_cloud_gax_internal::options::ClientConfig;
+    use grpc_server::{google, start_echo_server};
+    use pretty_assertions::assert_eq;
+
+    #[tokio::test]
+    async fn default_endpoint() -> anyhow::Result<()> {
```
vsharonlynn, line 30: For my learning...
What should this test case test?

Thinking out loud...
In `test_config(endpoint)` below, `config.endpoint` is unconditionally populated with `Some(endpoint.into())`.

So for `GrpcRustClient::new(test_config(endpoint), "https://storage.googleapis.com").await?;`
Because `config.endpoint` is `Some(...)`, grpc_rust.rs `resolve_target_endpoint` evaluates the override branch (`endpoint.map(...)`) and completely ignores "https://storage.googleapis.com". As a result, both `default_endpoint` and `override_endpoint` test the endpoint override code path. We never test the default endpoint branch which I assume is `config.endpoint == None`?

Github is down and I can't access the grpc_rust.rs code now. I'll double check the code tomorrow morning.

vsharonlynn (approved):

vsharonlynn:
Approved with some questions.

src/gax-internal/tests/grpc_rust_simple_request.rs
joshuatants: It's already checked in `check_simple_request` which is triggered by the other tests.

src/gax-internal/tests/grpc_rust_simple_request.rs
joshuatants: Thanks for catching this. Updated the tests.

PR #6427, opened 2026-08-17
chore(bigquery): publish crates

dbolduc:


dbolduc:
Argh, the builds passed, but GH dropped the notification. The only way I know how to fix this is to retrigger all of the builds and cross my fingers. :shrug:

src/bigquery/Cargo.toml
```diff
@@ -14,7 +14,6 @@
 
 [package]
 name        = "google-cloud-bigquery"
-publish     = false
 version     = "0.16.0-preview"
```
coryan, line 17: So this one can / should stay `-preview` ?

coryan (approved):

src/bigquery/Cargo.toml
dbolduc: Yes, it should stay `-preview` because of: https://crates.io/crates/google-cloud-bigquery.

The others are new crates, so I don't really see the point.

Issue #6430, opened 2026-08-17
Flaky BigQuery `run_writes` integration test fails with `UNAVAILABLE (OnChannelError)`


haphungw:
```
test bigquery::run_writes ... FAILED
failures:
---- bigquery::run_writes stdout ----
found 0 stale datasets
CREATING DATASET WITH ID: rust_bq_test_dataset_FT1sRSez
CREATING TABLE WITH ID: writes
full error: the operation failed. RPC error: the service reports an error with code UNAVAILABLE described as: Connection to server broken (OnChannelError)
Caused by:
    0: the service reports an error with code UNAVAILABLE described as: Connection to server broken (OnChannelError)
    1: code: 'The service is currently unavailable', message: "Connection to server broken (OnChannelError)"
  because: Rpc { source: Error { kind: Service(ServiceDetails { status_code: None, headers: Some({}), status: Status { code: Unavailable, message: "Connection to server broken (OnChannelError)", details: [] } }), source: Some(Status { code: Unavailable, message: "Connection to server broken (OnChannelError)", source: None }) } }
```

https://pantheon.corp.google.com/cloud-build/builds;region=us-central1/d43eb715-e4d1-438c-9168-678530826cb2?project=rust-sdk-testing&e=13803378&mods=allow_workbench_image_override

PR #6431, opened 2026-08-17
test(bigquery): disable flaky `run_writes` test

haphungw:
For #6430


dbolduc:
Can we not disable the test? It is our only test for this client. We know we need to implement retries.

PR #6432, opened 2026-08-17
feat(gax): support request stream channel capacity on RequestOptions

suzmue:
Add with_request_stream_channel_capacity directly to RequestOptionsBuilder, allowing streaming RPC builders to configure the internal request channel buffer capacity.

Retain BidiStreamOptions for compatibility during migration, updating its conversions to store and extract the channel capacity from RequestOptions.

All additions are gated behind google_cloud_unstable_gapic_streaming.

For #2318

### DIFF
--- a/src/gax/src/options.rs
+++ b/src/gax/src/options.rs
@@ -196,9 +196,6 @@ impl RequestOptions {
 #[cfg(google_cloud_unstable_gapic_streaming)]
 const MAX_REQUEST_CHANNEL_CAPACITY: usize = usize::MAX >> 3;
 
-#[cfg(google_cloud_unstable_gapic_streaming)]
-const DEFAULT_REQUEST_CHANNEL_CAPACITY: usize = 16;
-
 /// Options for configuring bidirectional streaming RPCs.
 #[cfg(google_cloud_unstable_gapic_streaming)]
 #[derive(Clone, Debug, Default)]
@@ -239,7 +236,7 @@ impl BidiStreamOptions {
     pub fn request_channel_capacity(&self) -> usize {
         self.request_options
             .request_stream_channel_capacity()
-            .unwrap_or(DEFAULT_REQUEST_CHANNEL_CAPACITY)
+            .unwrap_or(internal::DEFAULT_REQUEST_CHANNEL_CAPACITY)
     }
 
     /// Sets all request options, replacing any prior values.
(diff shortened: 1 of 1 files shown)


src/gax/src/options.rs
```diff
@@ -170,51 +172,60 @@ impl RequestOptions {
...
+    }
 }
 
 /// The maximum allowed capacity for a bidirectional stream request channel.
 ///
 /// This upper limit matches Tokio's internal `mpsc::channel` capacity limit
 /// (`usize::MAX >> 3`).
 #[cfg(google_cloud_unstable_gapic_streaming)]
-pub const MAX_REQUEST_CHANNEL_CAPACITY: usize = usize::MAX >> 3;
+const MAX_REQUEST_CHANNEL_CAPACITY: usize = usize::MAX >> 3;
```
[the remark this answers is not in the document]
suzmue: no I am removing it.

src/gax/src/options.rs
```diff
@@ -170,51 +172,60 @@ impl RequestOptions {
...
 const DEFAULT_REQUEST_CHANNEL_CAPACITY: usize = 16;
 
-/// Configuration options specific to bidirectional streaming RPCs.
-///
-/// Wraps standard [`RequestOptions`] while adding settings unique to bidirectional
-/// streaming RPCs, such as internal channel capacity for buffering requests.
+/// Options for configuring bidirectional streaming RPCs.
 #[cfg(google_cloud_unstable_gapic_streaming)]
-#[derive(Clone, Debug)]
+#[derive(Clone, Debug, PartialEq)]
```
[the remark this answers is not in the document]
suzmue: removing it, I'm getting rid of this type.

src/gax/src/options.rs
```diff
@@ -50,6 +50,8 @@ pub struct RequestOptions {
     retry_throttler: Option<SharedRetryThrottler>,
     polling_error_policy: Option<Arc<dyn PollingErrorPolicy>>,
     polling_backoff_policy: Option<Arc<dyn PollingBackoffPolicy>>,
+    #[cfg(google_cloud_unstable_gapic_streaming)]
+    request_stream_channel_capacity: Option<usize>,
```
dbolduc, line 54: Why not a `usize` initialized to the default value?

src/gax/src/options.rs
```diff
@@ -170,51 +172,51 @@ impl RequestOptions {
     pub fn set_polling_backoff_policy<V: Into<PollingBackoffPolicyArg>>(&mut self, v: V) {
         self.polling_backoff_policy = Some(v.into().0);
     }
+
+    /// Gets the current request stream channel capacity, or the default `16`.
+    #[cfg(google_cloud_unstable_gapic_streaming)]
+    pub fn request_stream_channel_capacity(&self) -> usize {
+        self.request_stream_channel_capacity
+            .unwrap_or(DEFAULT_REQUEST_CHANNEL_CAPACITY)
```
dbolduc: and then no unwraps here

dbolduc (approved):

src/gax/src/options.rs
suzmue: Oh this is so we can distinguish between unset and default value. I think the return type of request_stream_channel_capacity is actually wrong though, and should be option and we do the unwrap_or in the transport code.

PR #6433, opened 2026-08-17
feat(bigquery): introduce `AppendFuture` for nonblocking writes

### LINKED ISSUE #6400: BQ Write: Add `AppendFuture`
`append().send()`s for custom stream will need to return a custom future.

We need to synchronize some work (to preserve ordering of incoming writes) while also allowing for async execution (to queue up multiple writes before waiting for the previous responses)

Googlers can see: https://docs.google.com/document/d/1ALXaIw5KJKzI-cN5TB4Jn36c3fPAAmrwVJ6AKlg2llY/edit?tab=t.0#bookmark=id.lrhva3bhyn78 for more of the motivation and a code sketch

I prototyped the struct:

https://github.com/dbolduc/google-cloud-rust/blob/bq-write-arrow-writer-prototype/src/bigquery-write/src/append_future.rs#L64

haphungw:
`Append::send` is currently an `async fn`, which means network requests aren't sent to the underlying gRPC client until the caller `.awaits` them. To support custom exactly-once streams, we change `Append::send` to synchronous so that we can dispatch requests into the background immediately.

Fixes #6400

### DIFF
--- a/src/bigquery-write/src/append_future.rs
+++ b/src/bigquery-write/src/append_future.rs
@@ -31,6 +31,7 @@ pub struct AppendFuture {
 }
 
 impl AppendFuture {
+    #[allow(dead_code)]
     pub(crate) fn new(handle: JoinHandle<AppendResult<AppendResponse>>) -> Self {
         Self { handle }
     }
@@ -46,9 +47,8 @@ impl Future for AppendFuture {
             Err(e) => {
                 if e.is_panic() {
                     std::panic::resume_unwind(e.into_panic());
-                } else {
-                    Poll::Ready(Err(AppendError::UnexpectedEndOfStream))
                 }
+                Poll::Ready(Err(AppendError::UnexpectedEndOfStream))
             }
         }
     }
--- a/src/bigquery-write/src/append_future.rs
+++ b/src/bigquery-write/src/append_future.rs
@@ -1,0 +1,56 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use tokio::task::JoinHandle;
+
+use crate::error::{AppendError, AppendResult};
+use crate::model::AppendResponse;
+
+/// A future that resolves to the result of an async append operation.
+///
+/// This future represents a write request that has already been dispatched to the
+/// background network task. Awaiting this future yields the server's acknowledgment
+/// or an error if the stream fails.
+#[derive(Debug)]
+pub struct AppendFuture {
+    handle: JoinHandle<AppendResult<AppendResponse>>,
+}
+
+impl AppendFuture {
+    pub(crate) fn new(handle: JoinHandle<AppendResult<AppendResponse>>) -> Self {
+        Self { handle }
+    }
+}
+
+impl Future for AppendFuture {
+    type Output = AppendResult<AppendResponse>;
+
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let result = std::task::ready!(Pin::new(&mut self.handle).poll(cx));
+        match result {
+            Ok(res) => Poll::Ready(res),
+            Err(e) => {
+                if e.is_panic() {
+                    std::panic::resume_unwind(e.into_panic());
+                } else {
+                    // Task was cancelled or dropped.
+                    Poll::Ready(Err(AppendError::UnexpectedEndOfStream))
+                }
+            }
+        }
+    }
+}
--- a/src/bigquery-write/src/append_future.rs
+++ b/src/bigquery-write/src/append_future.rs
@@ -47,7 +47,6 @@ impl Future for AppendFuture {
                 if e.is_panic() {
                     std::panic::resume_unwind(e.into_panic());
                 } else {
-                    // Task was cancelled or dropped.
                     Poll::Ready(Err(AppendError::UnexpectedEndOfStream))
                 }
             }
--- a/src/bigquery-write/src/append_future.rs
+++ b/src/bigquery-write/src/append_future.rs
@@ -54,3 +54,61 @@ impl Future for AppendFuture {
         }
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use crate::model::TableSchema;
+
+    fn fake_response() -> AppendResponse {
+        AppendResponse {
+            offset: None,
+            updated_schema: Some(TableSchema::default()),
+        }
+    }
+
+    #[tokio::test]
+    async fn happy_path() {
+        let handle = tokio::spawn(async { Ok(fake_response()) });
+        let future = AppendFuture::new(handle);
+        let resp = future.await.expect("should succeed");
+        assert_eq!(resp.offset, None);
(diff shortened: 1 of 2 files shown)


src/bigquery-write/src/append_builder.rs
```diff
@@ -33,16 +34,22 @@ impl Append {
     }
 
     /// Append rows to the stream.
-    pub async fn send(self) -> AppendResult<AppendResponse> {
```
dbolduc, line 36: Can you leave this alone? We don't need to use `AppendFuture` in the `DefaultWriter`/`AppendBuilder` AFAICT.

I think we will only need it for the other stream types.

src/bigquery-write/src/append_builder.rs
```diff
@@ -33,16 +34,22 @@ impl Append {
...
-        let resp = resp_rx
-            .await
-            .map_err(|_| AppendError::UnexpectedEndOfStream)??;
-        let resp = resp.cnv().map_err(Error::ser)?;
-        to_result(resp)
+    ///
+    /// This eagerly dispatches the request to the network channel. It returns an
+    /// `AppendFuture` which you can `.await` to receive the server's response.
+    pub fn send(self) -> AppendFuture {
+        let handle = tokio::spawn(async move {
```
dbolduc: Gemini is right, but its suggestion is nonsense.

Sorry, I realize I did not get to solving this problem.

https://github.com/dbolduc/google-cloud-rust/blob/92dd3e5f7b0e2287fe00f79edfd26fc53138ac6b/src/bigquery-write/src/append_builder.rs#L84-L86

I think the answer looks like using `tokio::sync::mpsc::UnboundedSender` over `tokio::sync::mpsc::Sender`. I think i will be introducing other mechanisms for flow control. We won't rely on the type to deal with back pressure.

dbolduc: But this is a problem for the next PR (whenever you add the `AppendWithOffset` builder)

src/bigquery-write/src/append_future.rs
```diff
@@ -0,0 +1,113 @@
...
+    fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let result = std::task::ready!(Pin::new(&mut self.handle).poll(cx));
+        match result {
+            Ok(res) => Poll::Ready(res),
+            Err(e) => {
+                if e.is_panic() {
+                    std::panic::resume_unwind(e.into_panic());
+                } else {
+                    Poll::Ready(Err(AppendError::UnexpectedEndOfStream))
+                }
```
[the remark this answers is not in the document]
haphungw: done

src/bigquery-write/src/append_builder.rs
haphungw: I see. I reverted the changes to `append_builder`.

src/bigquery-write/src/append_builder.rs
haphungw: done

src/bigquery-write/src/append_builder.rs
```diff
@@ -33,16 +34,22 @@ impl Append {
...
+            let write = WriteRequest { req, resp_tx };
+            let _ = self.req_tx.send(write).await;
+            let resp = resp_rx
+                .await
+                .map_err(|_| AppendError::UnexpectedEndOfStream)??;
+            let resp = resp.cnv().map_err(Error::ser)?;
+            to_result(resp)
+        });
+        AppendFuture::new(handle)
     }
```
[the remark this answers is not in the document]
haphungw: There is indeed an OOM risk with `tokio::spawn` and a bounded channel. However, reverting `send()` back to `async` is against our design. We need it to be `sync` so that users can queue multiple writes without immediately `.await`ing them.

The default stream actually doesn't need pipelining. The specific fix of unbounded channel is reserved for the `AppendWithOffset` of custom streams.

src/bigquery-write/src/append_builder.rs
dbolduc: Yeah, I remember we thought about having a double `await` in pubsub. (i.e. one `await` for when the library has accepted the message, and another `await` for when the operation completes).

And I think ultimately we said that's too much. And instead just landed on a single `await` (we only return the appendfuture once we accept the message) and an `UnboundedSender` to implement the thing.

https://github.com/googleapis/google-cloud-rust/blob/a00654c73398379dc37e72571750c3f4fd8bfc35/src/pubsub/src/publisher/client.rs#L72

src/bigquery-write/src/append_builder.rs
dbolduc, line 1: revert these changes?

src/bigquery-write/src/lib.rs
```diff
@@ -41,6 +41,7 @@ pub mod builder {
     /// Request and client builders for the [Write][crate::client::Write] client
     pub mod write {
         pub use crate::append_builder::Append;
+        pub use crate::append_future::AppendFuture;
```
dbolduc: These are the builders. Can we just move the `use` to the top-level of the crate?

I think I need to reorg this stuff anyway

src/bigquery-write/src/append_future.rs
```diff
@@ -0,0 +1,113 @@
...
+use tokio::task::JoinHandle;
+
+use crate::error::{AppendError, AppendResult};
+use crate::model::AppendResponse;
+
+/// A future that resolves to the result of an async append operation.
+///
+/// This future represents a write request that has already been dispatched to the
+/// background network task. Awaiting this future yields the server's acknowledgment
+/// or an error if the stream fails.
```
dbolduc: Maybe:
```suggestion
/// This future represents a write request that has already been queued by the
/// client library to send over the network. Awaiting this future yields the server's acknowledgment
/// or an error if the write fails.
```

src/bigquery-write/src/append_future.rs
```diff
@@ -0,0 +1,113 @@
...
+        assert_eq!(resp.offset, None);
+        assert_eq!(resp.updated_schema, Some(TableSchema::default()));
+    }
+
+    #[tokio::test]
+    async fn dropped_task() {
+        let handle = tokio::spawn(async {
+            // Sleep forever to allow tests to abort it.
+            let _ = tokio::time::sleep(tokio::time::Duration::from_secs(10)).await;
+            Ok(fake_response())
```
dbolduc: This is a timing-based test. 10s != forever

You want to return a pending future instead: https://doc.rust-lang.org/std/future/fn.pending.html

src/bigquery-write/src/append_future.rs
```diff
@@ -0,0 +1,113 @@
...
+    use super::*;
+    use crate::model::TableSchema;
+
+    fn fake_response() -> AppendResponse {
+        AppendResponse {
+            offset: None,
+            updated_schema: Some(TableSchema::default()),
+        }
+    }
+
```
dbolduc: style nit: probably not worth having a function. Just inline it.

src/bigquery-write/src/append_future.rs
```diff
@@ -0,0 +1,113 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use std::future::Future;
+use std::pin::Pin;
+use std::task::{Context, Poll};
+use tokio::task::JoinHandle;
+
```
dbolduc, line 21: nit: remove newline

src/bigquery-write/src/append_builder.rs
haphungw: this was done already, i just pushed the updated changes.

src/bigquery-write/src/append_future.rs
haphungw: done

src/bigquery-write/src/append_future.rs
haphungw: done

src/bigquery-write/src/append_future.rs
haphungw: ack. done.

src/bigquery-write/src/lib.rs
haphungw: done

dbolduc (approved):

Issue #6401, opened 2026-08-14
BQ Write: Add `AppendWithOffset` builder


dbolduc:
Custom streams get their own `append()` builder that has the ability to supply an offset.

https://github.com/dbolduc/google-cloud-rust/blob/92dd3e5f7b0e2287fe00f79edfd26fc53138ac6b/src/bigquery-write/src/append_builder.rs#L64-L93

the implementation code should be reusable from the existing builder. They should both pass along the `Request` object to a lower layer.

PR #6434, opened 2026-08-18
impl(bigquery): migrate stream to unbounded channels

haphungw:
Migrate the background stream runner from an `mpsc` bounded channel to an unbounded channel, so that we no longer need to wrap every client append in a `tokio::spawn` task.

For #6401


dbolduc (approved):
Thanks!

PR #6435, opened 2026-08-18
fix(gax): grpc-rust wait for trailing server errors

joshuatants:
For #5991.

Previously, when `self.send_task.join()` returned an error (meaning sending failed or the pipe broke), the client immediately called `self.terminate()` and returned the send error. This meant we might lose any error trailers that the server had sent over.

This PR fixes that by introducing a `SendState` state machine that manages the `SendTask`. In the event of a send error, the `SendState` records the error *without* aborting receiving responses. This allows the client to continue polling inbound responses so that the receive stream is drained until it receives the server trailers. If the send error was due to the server terminating the stream, then the trailers should contain the server error message and we surface that; if the send error was due to an error on the client-side, then the `SendState` preserves the error so that we can surface it even if the server reports `Ok()`.

### DIFF
--- a/src/gax-internal/src/grpc/grpc_rust/bidi.rs
+++ b/src/gax-internal/src/grpc/grpc_rust/bidi.rs
@@ -15,7 +15,7 @@
 pub use super::receive::ReceiveTask;
 pub use super::receive::RecvItem;
 pub use super::send::GrpcRustSend;
-use super::send::SendTask;
+use super::send::{SendState, SendTask};
 use grpc::client::{CallOptions, Invoke};
 use grpc::core::RequestHeaders;
 use prost::Message;
@@ -36,7 +36,7 @@ use tokio::sync::mpsc::Receiver;
 pub struct GrpcRustStreaming<Response> {
     responses: Option<Receiver<tonic::Result<Option<RecvItem<Response>>>>>,
     receive_task: Option<ReceiveTask>,
-    send_task: SendTask,
+    send_state: SendState,
     /// Holds a pre-decoded initial response message (e.g. from stream setup or handshake)
     /// to yield on the first call to [`GrpcRustStreaming::message`] before polling `responses`.
     pending_response: Option<Response>,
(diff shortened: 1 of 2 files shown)


src/gax-internal/src/grpc/grpc_rust/send.rs
```diff
@@ -118,6 +120,62 @@ impl Drop for SendTask {
...
+            Self::Complete | Self::Failed(_) => return,
+        };
+        *self = match result {
+            Ok(()) => Self::Complete,
+            Err(status) => Self::Failed(status),
+        };
+    }
+
+    pub(super) async fn join_if_finished(&mut self) {
+        let is_finished = match self {
```
vsharonlynn: Lines 153-159 can be:
```
    if let Self::Active(send_task) = self {
        if send_task.is_finished() {
            self.join().await;
        }
    }
```

src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -126,7 +129,7 @@ where
                     // Join the background receive task to extract the final exit status/error.
                     None => {
                         self.responses.take();
-                        self.send_task.abort();
+                        self.send_state.abort();
```
vsharonlynn, line 132: For my learning: Do we need to join and check failure before abort is called?

vsharonlynn (approved):
Approve with some comments. Thanks!

src/gax-internal/src/grpc/grpc_rust/send.rs
```diff
@@ -118,6 +120,62 @@ impl Drop for SendTask {
...
+        }
+    }
+
+    /// Aborts the background request task and marks the state as complete.
+    pub(super) fn abort(&mut self) {
+        if let Self::Active(send_task) = self {
+            send_task.abort();
+        }
+        *self = Self::Complete;
+    }
```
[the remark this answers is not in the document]
joshuatants: Changed `abort` to preserve failure state.

src/gax-internal/src/grpc/grpc_rust/send.rs
joshuatants: Nice, updated.

src/gax-internal/src/grpc/grpc_rust/bidi.rs
joshuatants: `abort` only works if the state is `Active`, so if the state is already `Failed` it'll be a no-op.

PR #6438, opened 2026-08-18
fix(gax): update grpc-rust to 7482441 and add metadata conversion helpers

joshuatants:
For #5991.

`grpc-rust` is removing direct `From/Into` type conversions between `grpc::metadata::MetadataMap` and `tonic::metadata::MetadataMap`. This PR introduces explicit conversion helpers to perform the conversion until built-in conversions are provided in a future `grpc-rust` release.

This will becoming a breaking change when switching to `0.10`, so fixing now.

### DIFF
--- a/src/gax-internal/src/grpc/grpc_rust/metadata.rs
+++ b/src/gax-internal/src/grpc/grpc_rust/metadata.rs
@@ -1,0 +1,135 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO(#5991): Replace with `grpc-rust`'s built-in header conversion methods
+// when available in a future release.
+
+use google_cloud_gax::Result as GaxResult;
+use grpc::metadata::{
+    Ascii as GrpcAscii, AsciiMetadataValue as GrpcAsciiMetadataValue, Binary as GrpcBinary,
+    KeyAndValueRef as GrpcKeyAndValueRef, MetadataKey as GrpcMetadataKey,
+    MetadataMap as GrpcMetadataMap, MetadataValue as GrpcMetadataValue,
+};
+use http::HeaderMap;
+use tonic::metadata::{
+    MetadataKey as TonicMetadataKey, MetadataMap as TonicMetadataMap,
+    MetadataValue as TonicMetadataValue,
+};
+
+/// Converts a [`HeaderMap`] into a [`GrpcMetadataMap`].
+pub(super) fn from_header_map(headers: &HeaderMap) -> GaxResult<GrpcMetadataMap> {
+    let mut grpc_map = GrpcMetadataMap::new();
+    for (key, value) in headers {
+        if key.as_str().ends_with("-bin") {
+            if let Ok(k) = GrpcMetadataKey::<GrpcBinary>::from_bytes(key.as_str().as_bytes()) {
+                grpc_map.append_bin(k, GrpcMetadataValue::from_bytes(value.as_bytes()));
+            }
+        } else if let (Ok(k), Ok(val)) = (
+            GrpcMetadataKey::<GrpcAscii>::from_bytes(key.as_str().as_bytes()),
+            GrpcAsciiMetadataValue::try_from(value.as_bytes()),
+        ) {
+            grpc_map.append(k, val);
+        }
+    }
+    Ok(grpc_map)
+}
+
+/// Converts a [`GrpcMetadataMap`] into a [`TonicMetadataMap`].
+pub(super) fn to_tonic_map(metadata: &GrpcMetadataMap) -> TonicMetadataMap {
+    let mut tonic_map = TonicMetadataMap::new();
+    for item in metadata.iter() {
+        match item {
+            GrpcKeyAndValueRef::Ascii(key, val) => {
+                if let (Ok(name), Ok(val)) = (
+                    TonicMetadataKey::from_bytes(key.as_str().as_bytes()),
+                    TonicMetadataValue::try_from(val.to_str().as_bytes()),
+                ) {
+                    tonic_map.append(name, val);
+                }
+            }
+            GrpcKeyAndValueRef::Binary(key, val) => {
+                if let Ok(name) = TonicMetadataKey::from_bytes(key.as_str().as_bytes()) {
+                    let val = TonicMetadataValue::from_bytes(val.as_bytes());
+                    tonic_map.append_bin(name, val);
+                }
+            }
+        }
+    }
+    tonic_map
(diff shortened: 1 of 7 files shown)


src/gax-internal/src/grpc/grpc_rust/metadata.rs
```diff
@@ -0,0 +1,135 @@
...
+            }
+        } else if let (Ok(k), Ok(val)) = (
+            GrpcMetadataKey::<GrpcAscii>::from_bytes(key.as_str().as_bytes()),
+            GrpcAsciiMetadataValue::try_from(value.as_bytes()),
+        ) {
+            grpc_map.append(k, val);
+        }
+    }
+    Ok(grpc_map)
+}
```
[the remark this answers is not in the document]
joshuatants: We do not expect malformed headers from the server.

src/gax-internal/src/grpc/grpc_rust/metadata.rs
```diff
@@ -0,0 +1,135 @@
...
+            GrpcKeyAndValueRef::Binary(key, val) => {
+                if let Ok(name) = TonicMetadataKey::from_bytes(key.as_str().as_bytes()) {
+                    let val = TonicMetadataValue::from_bytes(val.as_bytes());
+                    tonic_map.append_bin(name, val);
+                }
+            }
+        }
+    }
+    tonic_map
+}
```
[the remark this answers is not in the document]
joshuatants: We do not expect malformed headers from the server.

src/gax-internal/src/grpc/grpc_rust/metadata.rs
```diff
@@ -0,0 +1,135 @@
...
+    for (key, value) in headers {
+        if key.as_str().ends_with("-bin") {
+            if let Ok(k) = GrpcMetadataKey::<GrpcBinary>::from_bytes(key.as_str().as_bytes()) {
+                grpc_map.append_bin(k, GrpcMetadataValue::from_bytes(value.as_bytes()));
+            }
+        } else if let (Ok(k), Ok(val)) = (
+            GrpcMetadataKey::<GrpcAscii>::from_bytes(key.as_str().as_bytes()),
+            GrpcAsciiMetadataValue::try_from(value.as_bytes()),
+        ) {
+            grpc_map.append(k, val);
```
vsharonlynn, line 43: Is the key value mapping guaranteed to always be valid? If it's not, it doesn't return an error.

And the function return is a GaxResult, but the funtion never returns an Err.

src/gax-internal/src/grpc/grpc_rust/metadata.rs
```diff
@@ -0,0 +1,135 @@
...
+pub(super) fn to_tonic_map(metadata: &GrpcMetadataMap) -> TonicMetadataMap {
+    let mut tonic_map = TonicMetadataMap::new();
+    for item in metadata.iter() {
+        match item {
+            GrpcKeyAndValueRef::Ascii(key, val) => {
+                if let (Ok(name), Ok(val)) = (
+                    TonicMetadataKey::from_bytes(key.as_str().as_bytes()),
+                    TonicMetadataValue::try_from(val.to_str().as_bytes()),
+                ) {
+                    tonic_map.append(name, val);
```
vsharonlynn: Ditto.

src/gax-internal/src/grpc/grpc_rust/metadata.rs
```diff
@@ -0,0 +1,135 @@
...
+use tonic::metadata::{
+    MetadataKey as TonicMetadataKey, MetadataMap as TonicMetadataMap,
+    MetadataValue as TonicMetadataValue,
+};
+
+/// Converts a [`HeaderMap`] into a [`GrpcMetadataMap`].
+pub(super) fn from_header_map(headers: &HeaderMap) -> GaxResult<GrpcMetadataMap> {
+    let mut grpc_map = GrpcMetadataMap::new();
+    for (key, value) in headers {
+        if key.as_str().ends_with("-bin") {
```
vsharonlynn, line 35: -bin means this is binary encoded64, right? Do we need to decode it before copying it to grpc metadata?

src/gax-internal/src/grpc/grpc_rust/metadata.rs
```diff
@@ -0,0 +1,135 @@
...
+                if let (Ok(name), Ok(val)) = (
+                    TonicMetadataKey::from_bytes(key.as_str().as_bytes()),
+                    TonicMetadataValue::try_from(val.to_str().as_bytes()),
+                ) {
+                    tonic_map.append(name, val);
+                }
+            }
+            GrpcKeyAndValueRef::Binary(key, val) => {
+                if let Ok(name) = TonicMetadataKey::from_bytes(key.as_str().as_bytes()) {
+                    let val = TonicMetadataValue::from_bytes(val.as_bytes());
```
vsharonlynn: In the other direction, do we need to encode the grpc metadata into the tonic bin params?

vsharonlynn (changes requested):

src/gax-internal/src/grpc/grpc_rust/metadata.rs
joshuatants: Nope. I use `from_bytes` and `append_bin` here which are designed to handle raw binary bytes; when this data is sent/received over the wire, `grpc-rust` will take care of converting them into the necessary format.

src/gax-internal/src/grpc/grpc_rust/metadata.rs
joshuatants: Not sure what you mean - this function handles GrpcMetadata -> TonicMetadata. If you're asking whether the other direction TonicMetadat -> GrpcMetadata is necessary, there's presently no use for it.

If you're asking whether we need to do more than use `as_bytes` and `from_bytes` - no, as long as we're passing raw bytes around between Grpc -> Tonic, the libraries will take care of the underlying representation on their own.

src/gax-internal/src/grpc/grpc_rust/metadata.rs
joshuatants: `grpc-rust` already does validation under the hood, so by the time this conversion is done, we know we have valid keys/values. Furthermore, the set of vaild keys/values that `grpc-rust` allows is at most as large as the set of valid keys/values for `tonic`.

Added `expect` statements to make this clearer.

src/gax-internal/src/grpc/grpc_rust/metadata.rs
joshuatants: Good catch. No it's not guaranteed to succeed.

Updated to propagate parse errors via `Error::ser`.

src/gax-internal/src/grpc/grpc_rust/metadata.rs
vsharonlynn: Cool. Thanks.

src/gax-internal/src/grpc/grpc_rust/metadata.rs
vsharonlynn: Oh... I didn't know the fact that the libraries already takes care of the underlying representation on their own. Thanks.

src/gax-internal/src/grpc/grpc_rust/metadata.rs
vsharonlynn: Ditto with the comment above. I didn't know the fact that the libraries already takes care of the underlying representation on their own. Thanks for explaining and sorry for the confusion!

vsharonlynn (approved):

PR #6444, opened 2026-08-18
docs(bigquery): remove RPC routing explanation

alvarowolfx:
Don't explain how the sausage factory works. 

Towards #5844

### DIFF
--- a/src/bigquery/src/query/builder.rs
+++ b/src/bigquery/src/query/builder.rs
@@ -27,19 +27,13 @@ use uuid::Uuid;
 pub(crate) const JOB_ID_PREFIX: &str = "job_";
 pub(crate) const QUERY_REQUEST_ID_PREFIX: &str = "req_";
 
-/// A builder for executing a SQL query.
+/// A builder for configuring and executing a SQL query.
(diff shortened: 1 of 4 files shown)


src/bigquery/src/query/builder.rs
```diff
@@ -188,7 +176,7 @@ impl Query {
         }
     }
 
-    /// Sends the query execution request and polls until execution completes.
+    /// Sends the query execution request and waits until execution completes.
```
dbolduc, line 179: super nit: why not "polls"? It seems more active, like we are being helpful.

dbolduc (approved):

src/bigquery/src/query/builder.rs
alvarowolfx: I was leaning on consistency with BigQuery docs and other SDKs, that just call out that users should "wait" instead of "poll". Also as a non native speaker, waits sounds better to me personally, but 🤷

dbolduc: Ok, SGTM.

PR #6446, opened 2026-08-18
feat(gax-internal): re-export Client and ClientBuilder in http::reqwest

olavloite:
Re-export `reqwest::Client` and `reqwest::ClientBuilder` from `gaxi::http::reqwest`. It does not bump the gax-internal version number, as that version has already been updated for the next release by other pull requests.

This allows client libraries to construct HTTP clients with custom connection timeouts (such as for querying the Compute Engine Metadata Service) without declaring direct dependencies on `reqwest`.

Needed for #6445


src/gax-internal/Cargo.toml
```diff
@@ -14,7 +14,7 @@
 
 [package]
 name        = "google-cloud-gax-internal"
-version     = "0.7.18"
+version     = "0.7.19"
```
coryan: We don't need to bump this again, it was already bumped (twice) since the last release.

coryan (changes requested):
Please revert the bumps:

```
git describe --tags --abbrev=0 upstream/main
git diff release-20260730 -- src/gax-internal/Cargo.toml
```

src/gax-internal/Cargo.toml
olavloite: Done

coryan (approved):

PR #6451, opened 2026-08-19
feat(bigquery): introduce `AppendWithOffset` builder

### LINKED ISSUE #6401: BQ Write: Add `AppendWithOffset` builder

haphungw:
Add an `AppendWithOffset` struct for application-created streams. This exposes an offset argument and delegates network sending into a background `AppendFuture` task.

Fixes #6401

### DIFF
--- a/src/bigquery-write/src/append_builder.rs
+++ b/src/bigquery-write/src/append_builder.rs
@@ -46,7 +46,7 @@ impl AppendWithOffset {
         let AppendWithOffset { req_tx, req } = self;
         
         tokio::spawn(async move {
-            let res = send_append_request(&req_tx, req).await;
+            let res = send_append_request(req_tx, req).await;
             let _ = tx.send(res);
         });
         
@@ -63,7 +63,7 @@ pub struct Append {
 
 /// Executes the network transmission and underlying proto translation for an `AppendRowsRequest`.
 pub(crate) async fn send_append_request(
-    req_tx: &mpsc::UnboundedSender<WriteRequest>,
+    req_tx: mpsc::UnboundedSender<WriteRequest>,
     req: AppendRowsRequest,
 ) -> AppendResult<AppendResponse> {
     let (resp_tx, resp_rx) = oneshot::channel();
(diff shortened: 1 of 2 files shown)


dbolduc (approved):

src/bigquery-write/src/append_builder.rs
```diff
@@ -20,6 +20,40 @@ use crate::runner::WriteRequest;
...
+        self
+    }
+
+    /// Append rows to the stream.
+    pub fn send(self) -> crate::append_future::AppendFuture {
+        let (tx, rx) = oneshot::channel();
+        let AppendWithOffset { req_tx, req } = self;
+        
+        tokio::spawn(async move {
+            let res = send_append_request(&req_tx, req).await;
```
dbolduc: why take a reference to the Sender vs. accepting it by value? We don't use it again.

haphungw: thank you, fixed

src/bigquery-write/src/append_builder.rs
```diff
@@ -20,6 +20,40 @@ use crate::runner::WriteRequest;
...
+    /// Sets the target stream offset to guarantee exactly-once execution.
+    pub fn set_offset(mut self, offset: i64) -> Self {
+        self.req.offset = Some(offset);
+        self
+    }
+
+    /// Append rows to the stream.
+    pub fn send(self) -> crate::append_future::AppendFuture {
+        let (tx, rx) = oneshot::channel();
+        let AppendWithOffset { req_tx, req } = self;
```
dbolduc: nit: skip this and just move out of `self` below 🤷

dbolduc (approved):

src/bigquery-write/src/append_builder.rs
haphungw: done

PR #6454, opened 2026-08-19
fix(storage): split large append payloads to avoid gRPC message limits

vsharonlynn:
Issue #5716

### DIFF
--- a/src/storage/src/storage/bidi_write/transport.rs
+++ b/src/storage/src/storage/bidi_write/transport.rs
@@ -165,10 +165,8 @@ impl AppendableObjectWriterTransport {
 
         Ok(())
     }
-}
 
-impl AppendableObjectWriter for AppendableObjectWriterTransport {
-    async fn append(&mut self, chunk: Bytes) -> Result<()> {
+    async fn append_sub_chunk(&mut self, chunk: Bytes) -> Result<()> {
         let length = chunk.len() as i64;
         let crc32c = crc32c::crc32c(&chunk);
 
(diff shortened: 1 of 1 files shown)


src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -419,6 +443,72 @@ mod tests {
...
+            transport.append(large_payload).await.unwrap();
+            transport
+        });
+
+        // First chunk (2 MiB)
+        let intent1 = rx.recv().await.unwrap();
+        if let UploadIntent::Append(req) = intent1 {
+            assert_eq!(req.write_offset, 0);
+            if let Some(Data::ChecksummedData(data)) = req.data {
+                assert_eq!(data.content.len(), 2 * 1024 * 1024);
```
joshuatants: How about reusing the constant MAX_WRITE_CHUNK_SIZE where possible?

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -194,6 +192,32 @@ impl AppendableObjectWriter for AppendableObjectWriterTransport {
...
+///
+/// The GCS v2 API specification (`google/storage/v2/storage.proto`) restricts the
+/// `ChecksummedData.content` payload length to at most 2 MiB per chunk. Splitting
+/// larger writes into <= 2 MiB chunks also ensures individual gRPC messages remain
+/// safely under gRPC's default 4 MiB message size limit.
+const MAX_WRITE_CHUNK_SIZE: usize = 2 * 1024 * 1024;
+
+impl AppendableObjectWriter for AppendableObjectWriterTransport {
+    async fn append(&mut self, chunk: Bytes) -> Result<()> {
+        if chunk.is_empty() {
```
joshuatants, line 207: If there's nothing to write, why call append_sub_chunk?

joshuatants (approved):

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Yes of course. Done.

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Fixed. Thanks!

PR #6455, opened 2026-08-19
test(showcase): verify PQC transport with TLS in http and grpc

suzmue:
Add an integration test target for gapic-showcase running with Auto-TLS and restricted key exchange groups (--tls-groups 0x11ec). This verifies that the HTTP (reqwest) and gRPC (tonic) transports in google-cloud-rust successfully perform post-quantum hybrid key exchange using X25519MLKEM768.


tests/showcase/src/pqc.rs
```diff
@@ -0,0 +1,207 @@
...
+    let mut ready = false;
+    for _ in 0..50 {
+        if let Ok(pem) = std::fs::read(&ca_cert_path)
+            && !pem.is_empty()
+        {
+            // SAFETY: Setting environment variable before issuing network requests in the test process.
+            // This configures `rustls-native-certs` to trust the autogenerated Showcase CA.
+            unsafe {
+                std::env::set_var("SSL_CERT_FILE", &ca_cert_path);
+            }
```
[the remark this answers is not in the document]
suzmue: its running in its own process

tests/showcase/src/pqc.rs
```diff
@@ -0,0 +1,212 @@
...
+    tracing::info!("started showcase PQC server: {child:?}");
+
+    // Wait for the Showcase server to write its autogenerated CA certificate and become ready.
+    let mut ready = false;
+    for _ in 0..50 {
+        if let Ok(pem) = std::fs::read(&ca_cert_path)
+            && !pem.is_empty()
+        {
+            // SAFETY: Setting environment variable before issuing network requests in the test process.
+            // This configures `rustls-native-certs` to trust the autogenerated Showcase CA.
```
coryan: `set_var()` is unsafe because if more than two threads call it at the same time then the program gets undefined behavior.  The right comment might be "This is safe because this is the only test in the repo that changes SSL_CERT_FILE" ... The "Right Thing" :tm: is to set the certificate file via some client option (that we do not have).

coryan (approved):
Please verify my suggestion for the SAFETY comment is correct and if so, change the comment and merge.

tests/showcase/src/pqc.rs
suzmue: The safety is from no other tests will run set_env at the same time on the same thread, which we do by putting these in its own tests/pqc.rs that only runs this test in serial. Updated the comment to reflect this.

PR #6458, opened 2026-08-19
feat(storage): global custom header configuration

xlai20:
DD: go/rust-sdk-custom-header
Part of https://github.com/googleapis/google-cloud-rust/issues/5997

This PR introduces client-level (global) custom header configuration for Google Cloud Storage, while ensuring strict behavioral and architectural parity between HTTP/gRPC transports and global/request-level configuration.

### 1. Storage Client API
* **Public API Added:** Introduced [`ClientBuilder::with_custom_header(mut self, key: K, value: V) -> Self`](file:///usr/local/google/home/xlai/feature-parity-rust/write/google-cloud-rust/src/storage/src/storage/client.rs) to configure client-wide custom headers applied to every outgoing request.
* **Precedence Rules:**
  * **Request-level overrides Global-level:** When the same header key is configured at both levels, the request-specific header explicitly overwrites the global header.
  * **System-level overrides Custom:** Reserved system headers (e.g., `authorization`, `user-agent`, `x-goog-user-project`, `x-goog-request-params`, `x-goog-api-client`, `x-goog-api-key`) always take precedence and strip conflicting custom header keys at both global and request scopes.

### 2. Parity between HTTP and gRPC
* **Aligned Sanitization:** Harmonized HTTP and gRPC transport sanitization to scrub the reserved system keys.
* **Uniform Behavior:** Ensured identical merging order and sanitization rules are enforced whether a request executes via HTTP (`ReqwestClient`) or gRPC (`grpc_helpers::make_headers`).

### 3. Parity between Global-level and Request-level Custom Headers
* **Public API Parity:** `ClientBuilder::with_custom_header` shares the identical function signature, generic bounds, and error ergonomics as the existing [`RequestOptionsBuilder::with_custom_header`](file:///usr/local/google/home/xlai/feature-parity-rust/write/google-cloud-rust/src/storage/src/storage/client.rs).
* **Internal `Extensions` Architecture:** Both layers use the shared `Extensions` container holding an `http::HeaderMap` (`ClientConfig.extensions` for global, `RequestOptions.extensions` for request-level). This enables transport layers to extract and unpack configuration uniformly without requiring dedicated struct fields.

### 4. Testing
* **Isolated & Integration Tests:** Added unit tests verifying isolated global headers, isolated request-level headers, and global vs. request precedence (`make_headers_precedence_request_over_global`).
* **System Precedence Tests:** Added unit tests verifying that system keys strip conflicting custom keys across both populated and empty parameter states.
* **Documentation Tests:** Included full doctests for `ClientBuilder::with_custom_header` validating public documentation, usage patterns, and compiler checks.


src/gax-internal/src/http.rs
```diff
@@ -60,6 +60,23 @@ const X_GOOG_USER_PROJECT: http::header::HeaderName =
...
+        http::header::USER_AGENT,
+        http::header::AUTHORIZATION,
+        X_GOOG_API_KEY,
+        X_GOOG_USER_PROJECT,
+        X_GOOG_REQUEST_PARAMS,
+        X_GOOG_API_CLIENT,
+    ] {
+        headers.remove(key);
+    }
+}
```
joshuatants: Can this be refactored with grpc_helpers.rs:91-100?

src/storage/src/storage/client.rs
```diff
@@ -805,6 +805,44 @@ impl ClientBuilder {
         self
     }
 
+    /// Injects a global custom header for all HTTP and gRPC requests made by this client.
```
joshuatants, line 808: Let's also mention how to handle repeated values (similar to the last PR)

src/gax-internal/src/grpc/grpc_helpers.rs
```diff
@@ -69,13 +69,21 @@ pub(crate) fn make_credentials(config: &ClientConfig) -> ClientBuilderResult<Cre
 /// Custom headers can be provided through `RequestOptions`.
```
joshuatants: Sholud update this to also mention it can come through the Extensions

src/gax-internal/src/options.rs
```diff
@@ -15,6 +15,7 @@
 pub use google_cloud_auth::credentials::Credentials;
 
 // The client configuration for [crate::http::ReqwestClient] and [crate::grpc::Client].
+pub use google_cloud_gax::client_builder::internal::Extensions;
 pub type ClientConfig = google_cloud_gax::client_builder::internal::ClientConfig<Credentials>;
```
joshuatants, line 19: Does the comment on L17 apply to this as well? Or only to L19?

joshuatants (approved):

src/gax-internal/src/grpc/grpc_helpers.rs
xlai20: Done.

src/gax-internal/src/http.rs
xlai20: Done.

This function is now shared between http and grpc. It is in a new module called "src/gax-internal/src/headers.rs".

src/gax-internal/src/options.rs
xlai20: This is an accident. This line should not insert in between the comment and the existing code.

Moved it to the right position.

src/storage/src/storage/client.rs
xlai20: Done.

PR #6472, opened 2026-08-19
impl(bigquery): use arrow format with jobs.query

alvarowolfx:
Trying out arrow support on jobs.query. This is gonna break on CI because support for it is behind an allowlist.


alvarowolfx:
closing for now to break into smaller PRs

PR #6477, opened 2026-08-19
feat(speech): generate bidi streaming methods and add integration test

suzmue:
Enable bidirectional streaming support for Google Cloud Speech v2 by setting `include_bidi_streaming_methods: true` in `librarian.yaml` and regenerating the crate.

Add a live integration test for `Speech::streaming_recognize()` that streams 4KB chunks from the public `hello.wav` sample, verifies bidirectional streaming lifecycle and half-closing, and validates that the transcribed text contains "hello".


src/generated/cloud/speech/v2/README.md
```diff
@@ -2,13 +2,6 @@
 
 <!-- Code generated by sidekick. DO NOT EDIT. -->
 
-**WARNING:** some RPCs have no corresponding Rust function to call them.
```
coryan, line 5: :shipit:

tests/integration/src/lib.rs
```diff
@@ -13,4 +13,6 @@
 // limitations under the License.
 
 pub mod error_details;
+#[cfg(google_cloud_unstable_gapic_streaming)]
```
coryan: The changes in the README and using a `#[cfg()]` to get all the features is slightly odd.

tests/integration/tests/driver.rs
```diff
@@ -48,4 +48,13 @@ mod driver {
             .await
             .inspect_err(anydump)
     }
+
```
coryan: Please create a separate `tests/*` directory for the bidi streaming tests? I would rather not pile up the dependencies even more.

coryan (approved):
drive-by

tests/integration/src/lib.rs
suzmue: I agree 100%.

I plan to hold onto this change until after the release, so we don't make the README updates and will then have the config flags removed before the release that makes those updates.

suzmue: Release went out, preparing to submit this and then have a follow up with removing `#[cfg()]`

coryan (approved):

PR #6489, opened 2026-08-21
feat(storage): support initial payload in bidi_write transport

vsharonlynn:
Issue #5716 .

### DIFF
--- a/src/storage/src/storage/bidi_write/connector.rs
+++ b/src/storage/src/storage/bidi_write/connector.rs
@@ -78,6 +78,7 @@ where
             spec: Arc::new(Mutex::new(AppendObjectSpecState::Write(
                 Box::default(),
                 None,
+                None,
             ))),
             options,
             client,
@@ -88,6 +89,14 @@ where
     pub async fn connect_open(
         &mut self,
         req: crate::model_ext::OpenAppendableObjectRequest,
+    ) -> Result<(BidiWriteObjectResponse, Connection<T::Stream>)> {
+        self.connect_open_and_append(req, None).await
+    }
+
+    pub async fn connect_open_and_append(
+        &mut self,
+        req: crate::model_ext::OpenAppendableObjectRequest,
+        initial_chunk: Option<bytes::Bytes>,
     ) -> Result<(BidiWriteObjectResponse, Connection<T::Stream>)> {
         let resource = match req.spec.resource {
             Some(r) => {
@@ -111,7 +120,7 @@ where
             .map(|p| p.to_proto().map_err(Error::deser))
             .transpose()?;
         *self.spec.lock().expect("never poisoned") =
-            AppendObjectSpecState::Write(Box::new(spec), None);
+            AppendObjectSpecState::Write(Box::new(spec), None, initial_chunk);
         self.connect_attempt_loop().await
     }
 
@@ -134,7 +143,7 @@ where
             .params
             .map(|p| p.to_proto().map_err(Error::deser))
             .transpose()?;
-        *self.spec.lock().expect("never poisoned") = AppendObjectSpecState::Append(spec);
+        *self.spec.lock().expect("never poisoned") = AppendObjectSpecState::Append(spec, None);
         self.connect_attempt_loop().await
     }
 
@@ -231,20 +240,40 @@ fn prepare_request(
     state: &AppendObjectSpecState,
     params: Option<CommonObjectRequestParams>,
 ) -> Result<(BidiWriteObjectRequest, String)> {
-    let (first_message, routing_token) = match state {
-        AppendObjectSpecState::Write(spec, rt) => {
-            (FirstMessage::WriteObjectSpec((**spec).clone()), rt.clone())
-        }
-        AppendObjectSpecState::Append(spec) => (
+    let (first_message, routing_token, initial_chunk) = match state {
+        AppendObjectSpecState::Write(spec, rt, data) => (
+            FirstMessage::WriteObjectSpec((**spec).clone()),
+            rt.clone(),
+            data.clone(),
+        ),
+        AppendObjectSpecState::Append(spec, data) => (
             FirstMessage::AppendObjectSpec(spec.clone()),
             spec.routing_token.clone(),
+            data.clone(),
         ),
     };
 
     let state_lookup = matches!(first_message, FirstMessage::AppendObjectSpec(_));
 
+    let data = match initial_chunk {
+        Some(chunk) if !chunk.is_empty() => {
+            let crc = crc32c::crc32c(&chunk);
+            Some(
+                crate::google::storage::v2::bidi_write_object_request::Data::ChecksummedData(
+                    crate::google::storage::v2::ChecksummedData {
+                        content: chunk,
+                        crc32c: Some(crc),
+                    },
+                ),
+            )
+        }
+        _ => None,
+    };
+
     let request = BidiWriteObjectRequest {
         first_message: Some(first_message),
+        write_offset: 0,
+        data,
         common_object_request_params: params,
         state_lookup,
         ..BidiWriteObjectRequest::default()
@@ -583,8 +612,8 @@ mod tests {
 
         let got = connector.spec.lock().expect("never poisoned").clone();
         match got {
-            AppendObjectSpecState::Write(_, _) => panic!("Should be Append"),
-            AppendObjectSpecState::Append(got) => {
+            AppendObjectSpecState::Write(..) => panic!("Should be Append"),
+            AppendObjectSpecState::Append(got, _) => {
                 assert_eq!(got.routing_token.as_deref(), Some("r1"));
             }
         }
@@ -698,7 +727,7 @@ mod tests {
         assert_eq!(response, initial);
 
         let guard = connector.spec.lock().expect("never poisoned");
-        if let AppendObjectSpecState::Append(s) = &*guard {
+        if let AppendObjectSpecState::Append(s, _) = &*guard {
             assert!(s.routing_token.is_none(), "{s:?}");
             assert_eq!(s.generation, 123456, "{s:?}");
             assert_eq!(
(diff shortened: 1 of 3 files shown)


src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -88,6 +89,14 @@ where
...
         req: crate::model_ext::OpenAppendableObjectRequest,
+    ) -> Result<(BidiWriteObjectResponse, Connection<T::Stream>)> {
+        self.connect_open_and_append(req, None).await
+    }
+
+    pub async fn connect_open_and_append(
+        &mut self,
+        req: crate::model_ext::OpenAppendableObjectRequest,
+        initial_chunk: Option<bytes::Bytes>,
     ) -> Result<(BidiWriteObjectResponse, Connection<T::Stream>)> {
```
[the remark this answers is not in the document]
vsharonlynn: This function is self-explanatory.

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -51,7 +51,45 @@ impl AppendableObjectWriterTransport {
...
         let (initial, connection) = connector.connect_open(req).await?;
-        Self::start_worker(connector, initial, connection, 0)
+        Self::start_worker(connector, initial, connection, 0, 0, None)
+    }
+
+    pub async fn new_open_and_append<T>(
+        mut connector: Connector<T>,
+        req: OpenAppendableObjectRequest,
+        chunk: Bytes,
+    ) -> Result<Self>
```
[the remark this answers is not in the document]
vsharonlynn: This function is self-explanatory.

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -713,6 +742,83 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
+    async fn connect_open_and_append_sends_payload_in_first_request() -> Result<()> {
+        // Arrange
```
joshuatants: Style: going forward, for long tests like these, let's also add comments at suitable locations to make the test more readable.

src/storage/src/storage/bidi_write/state.rs
```diff
@@ -36,23 +36,26 @@ impl AppendObjectSpecState {
...
+                    AppendObjectSpec {
+                        bucket,
+                        object,
+                        generation: new_generation.unwrap_or(0),
+                        write_handle: m.write_handle.clone(),
+                        if_metageneration_match: spec.if_metageneration_match,
+                        if_metageneration_not_match: spec.if_metageneration_not_match,
+                        routing_token: token.take(),
+                    },
+                    data.take(),
```
joshuatants: I think `data` should be released by setting it to `None`; otherwise, even after the initial 2MiB chunk succeeds, we still keep it around in memory.

Pl. also check the `Append` branch below.

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -713,6 +742,83 @@ mod tests {
         Ok(())
     }
 
+    #[tokio::test]
```
joshuatants: Let's also add tests for these cases:

- `open_and_append` with empty chunk
- `open_and_append` with a chunk exactly equal to the max chunk size
-  `open_and_append` with a chunk that's bigger than the max chunk size

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -51,7 +51,45 @@ impl AppendableObjectWriterTransport {
         <T as Client>::Stream: TonicStreaming,
     {
         let (initial, connection) = connector.connect_open(req).await?;
-        Self::start_worker(connector, initial, connection, 0)
+        Self::start_worker(connector, initial, connection, 0, 0, None)
```
joshuatants: This parameter list is getting pretty long. Maybe refactor to pass in a struct? 

Can be done as a follow-up.

src/storage/src/storage/bidi_write/state.rs
```diff
@@ -24,8 +24,8 @@ use prost::Message;
 /// Represents the state of the initial request in the stream.
 #[derive(Clone, Debug)]
 pub enum AppendObjectSpecState {
-    Write(Box<WriteObjectSpec>, Option<String>),
-    Append(AppendObjectSpec),
+    Write(Box<WriteObjectSpec>, Option<String>, Option<bytes::Bytes>),
+    Append(AppendObjectSpec, Option<bytes::Bytes>),
```
joshuatants: Consider using named enums for readability

joshuatants (changes requested):

src/storage/src/storage/bidi_write/state.rs
xlai20: +1 to Joshua's suggestion.

Consider refactoring `AppendObjectSpecState` variants from positional tuple fields to named struct fields. Having multiple `Option<...>` fields in a tuple makes pattern matching less readable (e.g., `Write(_, ref token, _)`) and increases refactoring friction:

```rust
pub enum AppendObjectSpecState {
    Write {
        spec: Box<WriteObjectSpec>,
        routing_token: Option<String>,
        initial_chunk: Option<bytes::Bytes>,
    },
    Append {
        spec: AppendObjectSpec,
        initial_chunk: Option<bytes::Bytes>,
    },
}

File: src/storage/src/storage/bidi_write/connector.rs (hunk 1 above)
Comment (xlai20) [inline src/storage/src/storage/bidi_write/connector.rs:102]: Need to add doc to public function.

Please add doc comments (`///`) describing `connect_open_and_append`, explaining the role of `initial_chunk` and the return value to adhere to the crate's documentation guidelines.

File: src/storage/src/storage/bidi_write/transport.rs (hunk 2 above)
Comment (xlai20) [inline src/storage/src/storage/bidi_write/transport.rs:72]: Please add a doc comment (`///`) for `new_open_and_append` explaining how the initial payload is partitioned into the opening request (up to `MAX_WRITE_CHUNK_SIZE`) and subsequent append queue.

File: src/storage/src/storage/bidi_write/transport.rs [hunk 8]
```diff
@@ -64,14 +102,16 @@ impl AppendableObjectWriterTransport {
...
     }
 
     fn start_worker<T>(
         connector: Connector<T>,
         initial: BidiWriteObjectResponse,
         connection: Connection<<T as Client>::Stream>,
         generation: i64,
+        initial_chunk_len: usize,
+        initial_chunk_crc: Option<u32>,
     ) -> Result<Self>
```
xlai20, line 128: The parameter list for `start_worker` now has 6 arguments. Since `initial_chunk_len` and `initial_chunk_crc` are coupled (always either `(0, None)` or `(len, Some(crc))`), consider grouping them into a struct or `Option<InitialPayload>`:

```rust
struct InitialPayload {
    len: usize,
    crc32c: u32,
}

File: src/storage/src/storage/bidi_write/transport.rs [hunk 9]
```diff
@@ -1028,4 +1074,51 @@ mod tests {
...
+        let transport =
+            AppendableObjectWriterTransport::new_open_and_append(connector, req, chunk.clone())
+                .await?;
+
+        assert_eq!(transport.generation(), 987654321);
+        assert_eq!(transport.persisted_size(), 0);
+        assert_eq!(transport.write_offset, chunk.len() as i64);
+        assert_eq!(transport.running_crc32c, Some(expected_crc));
+        Ok(())
+    }
```
xlai20, line 1154: Suggest to add more test cases:

1. `new_open_and_append` with an empty payload (`Bytes::new()`), asserting that it cleanly delegates to `new_open` without attaching data.
2. `new_open_and_append` with a chunk exactly equal to `MAX_WRITE_CHUNK_SIZE` (2 MiB).
3. `new_open_and_append` with a payload exceeding `MAX_WRITE_CHUNK_SIZE` (e.g., 3 MiB), verifying that the first 2 MiB is sent in the initial connect request and the remaining 1 MiB is dispatched via `transport.append()` with properly incremented offsets and combined CRC32C.

You can choose not to add if there is already a test coverage somewhere else that I didn't know, or there is a reason that some test cases are better to be in just 1 to avoid test bloating.

xlai20 (changes requested):

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Yes, thanks. Sorry I forgot to clean this up.

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Done. Thanks!

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Good idea. Added TODO in the function definition of `start_worker`.

src/storage/src/storage/bidi_write/state.rs
vsharonlynn: Good idea. Done. Thank you.

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Thanks for the input! Done.

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Done. Thanks!

joshuatants (approved):

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -75,10 +75,11 @@ where
 {
     pub fn new(options: RequestOptions, client: T) -> Self {
```
xlai20, line 76: Nits: for public function (even it's public within the crate), it's better to come with some function /// doc comments. It can be helpful for internal contributors and maintainability.

The same comment applies to all the undocumented pub function in these files, e.g, new(), connect_open().

xlai20 (approved):
lgtm with an unresolved comment.

src/storage/src/storage/bidi_write/connector.rs
vsharonlynn: Yes, I will send a follow-up PR for these other functions. For this PR, I only added public function docs for the functions being modified. Sorry, I forgot to mention when resolving your previous comments about this. Thanks for the reminder.

PR #6490, opened 2026-08-21
feat(storage): add send_and_append on OpenAppendableObject

vsharonlynn:
Issue #5716 .

This PR follows PR #6489 .

### DIFF
--- a/src/storage/src/storage/stub.rs
+++ b/src/storage/src/storage/stub.rs
@@ -97,7 +97,8 @@ pub trait Storage: std::fmt::Debug + Send + Sync {
     }
 
     #[cfg(google_cloud_unstable_storage_bidi)]
-    /// Implements [crate::client::Storage::open_appendable_object] with an initial chunk.
+    /// Implements [crate::client::Storage::open_appendable_object]. Also sends an
+    /// initial chunk of data with the opening request.
     fn open_appendable_object_and_append(
         &self,
         _request: crate::model_ext::OpenAppendableObjectRequest,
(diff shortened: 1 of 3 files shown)


src/storage/src/storage/stub.rs
```diff
@@ -96,6 +96,19 @@ pub trait Storage: std::fmt::Debug + Send + Sync {
         unimplemented_stub::<crate::appendable_object_writer::AppendableObjectWriter>()
     }
 
+    #[cfg(google_cloud_unstable_storage_bidi)]
+    /// Implements [crate::client::Storage::open_appendable_object] with an initial chunk.
```
joshuatants: Nit: maybe something like "Implements XYZ. Also sends an appendable chunk with the initial request".

Right now it reads funny.

src/storage/src/storage/transport.rs
```diff
@@ -990,6 +1059,132 @@ mod tests {
         Ok(())
     }
 
+    #[cfg(google_cloud_unstable_storage_bidi)]
+    #[tokio::test]
+    async fn open_appendable_object_and_append_not_found() -> anyhow::Result<()> {
```
joshuatants: what is the thing that is "not_found"? can the test name include that so that it's clearer?

joshuatants (approved):

src/storage/src/storage/stub.rs
vsharonlynn: Done. THanks!

src/storage/src/storage/transport.rs
vsharonlynn: Done. Thanks!

PR #6503, opened 2026-08-21
perf(bigquery): borrow schemas during REST row conversion

fornwall:
Row conversion cloned table field schemas and allocated temporary schema wrappers for every cell, even though conversion only reads schema metadata.

Pass borrowed field slices through the recursive conversion instead, which removes a deep `TableFieldSchema` clone, an `Arc`, and a `TableSchema` allocation per cell for scalar, nested and repeated values.

Also reserve the cell vector up front, now that the column count is known before the loop runs.

### DIFF
--- a/src/bigquery/src/query/row.rs
+++ b/src/bigquery/src/query/row.rs
@@ -14,6 +14,7 @@
 
 use crate::error::{ConvertError, RowError};
 use crate::query::{FromSql, Schema};
+use google_cloud_bigquery_v2::model::TableFieldSchema;
 use std::sync::Arc;
 use wkt::{ListValue, Struct, Value};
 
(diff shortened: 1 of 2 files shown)


src/bigquery/src/query/row.rs
```diff
@@ -264,49 +260,35 @@ fn get_field_value(value: Value) -> Result<Value> {
...
 }
 
-fn convert_repeated(
-    value: ListValue,
-    field_name: &str,
-    field_type: &str,
-    schema: &Arc<Schema>,
-) -> Result<Value> {
+fn convert_repeated(value: ListValue, field: &TableFieldSchema) -> Result<Value> {
     let mut values = ListValue::new();
```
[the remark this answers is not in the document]
fornwall: Done in 4c7dbc9b2f3155668b4ee029dfd1af6074fd60e6.

src/bigquery/src/query/row.rs
```diff
@@ -264,49 +259,33 @@ fn get_field_value(value: Value) -> Result<Value> {
...
-        for (i, val) in list.into_iter().enumerate() {
-            if let Some(field) = schema.get_field_by_index(i) {
-                obj.insert(field.name.clone(), val);
-            }
-        }
+    for (field, value) in fields.iter().zip(values) {
+        obj.insert(field.name.clone(), value);
     }
     Ok(Value::Object(obj))
 }
```
alvarowolfx, line 292: I think we don't need to `mut` here anymore.

```suggestion
fn convert_nested(value: Struct, fields: &[TableFieldSchema]) -> Result<Value> {
    let values = convert_row(value, fields)?;
    let obj: Struct = fields
        .iter()
        .zip(values)
        .map(|(field, value)| (field.name.clone(), value))
        .collect();
    Ok(Value::Object(obj))
}
```

src/bigquery/src/query/row.rs
```diff
@@ -246,6 +224,23 @@ impl Row {
...
+            fields.len(),
+            field_list.len()
+        )));
+    }
+
+    for (cell, field) in field_list.iter_mut().zip(fields) {
+        *cell = convert_value(get_field_value(cell.take())?, field)?;
+    }
+    Ok(field_list)
+}
```
alvarowolfx, line 242: We usually try to avoid `mut`, but in this case it seems like is necessary to take advantage of `cell.take()`

alvarowolfx (approved):
Thanks for the improvements @fornwall 

We usually  try to avoid `mut`, so I added some minor style suggestions. But I can do it later since I'm doing some work semi-related to this.

src/bigquery/src/query/row.rs
fornwall: Thanks - pushed this in e949a43f6972f4fef71762449df6750ed18c1e49.

fornwall:
> 1 line in your changes missing coverage

Just pushed a test hitting that.

dbolduc (approved):

PR #6514, opened 2026-08-23
fix(bigquery): make get_job return None for job references without job IDs

fornwall:
Dry-run queries return a job reference that has no job ID (verified against a real project for both `jobs.query` and `jobs.insert`):

    "jobReference": {"projectId": "p", "location": "US"}

This is distinct from the no-job `JOB_CREATION_OPTIONAL` case, which omits `jobReference` altogether and returns a queryId instead (that case was already handled).

`Query::get_job` and `CompleteQuery::get_job` only checked that the job reference was present, so on dry runs they handed back `Some(GetJob)` - a builder for a `jobs.get` call carrying an empty job ID.

Fix this by returning `None` when the job ID is empty, matching the documented contract that `get_job` returns `None` when no job was created.

### DIFF
--- a/src/bigquery/src/query/query_handle.rs
+++ b/src/bigquery/src/query/query_handle.rs
@@ -119,7 +119,8 @@ impl Query {
     /// Build a request to fetch full [Job] execution metadata from the service for this query.
     ///
     /// > Returns `None` if the query was executed without creating a job
-    /// > (for example, when using [`JobCreationMode::JobCreationOptional`]).
+    /// > (for example, when using [`JobCreationMode::JobCreationOptional`],
+    /// > or when the query was a dry run).
     ///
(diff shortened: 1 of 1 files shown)


alvarowolfx (approved):
LGTM, thanks for the fix. It seems like dry run were a blind spot in my work

coryan (approved):

fornwall:
> It seems like dry run were a blind spot in my work

The server behaviour here is surprising / undocumented / not really logical, right? Or do these job references from dry runs make any sense?

alvarowolfx:
> The server behaviour here is surprising / undocumented / not really logical, right? Or does these job references from dry run make any sense?

It's confusing indeed, but maybe they thought about adding routing information on where the job was going to run, by returning a partially filled JobReference without job id. Will check with the service team

PR #6518, opened 2026-08-24
fix: skip redundant body fields

### LINKED ISSUE #6515: aiplatform-v1: PredictionService::embed_content is rejected by Vertex with "oneof field '_model' is already set"
## What happened

Every call through the typed `PredictionService::embed_content` builder fails with:

```
INVALID_ARGUMENT: Invalid value (oneof), oneof field '_model' is already set. Cannot set 'model'
```

## Repro

```rust
use google_cloud_aiplatform_v1::client::PredictionService;
use google_cloud_aiplatform_v1::model::{Content, Part};
use google_cloud_aiplatform_v1::model::embed_content_request::EmbedContentConfig;

let client = PredictionService::builder()
    .with_endpoint("https://aiplatform.googleapis.com")
    .build()
    .await?;

let content = Content::new().set_parts([Part::new().set_text("hello world")]);
let cfg = EmbedContentConfig::new()
    .set_output_dimensionality(3072)
    .set_auto_truncate(true);

let resp = client
    .embed_content()
    .set_model("projects/PROJECT/locations/global/publishers/google/models/gemini-embedding-2")
    .set_content(content)
    .set_embed_content_config(cfg)
    .send()
    .await?; // <-- INVALID_ARGUMENT: oneof field '_model' is already set. Cannot set 'model'
```

Environment: `google-cloud-aiplatform-v1` 1.15.0 (features `prediction-service`, `default-rustls-provider`), Rust 1.96, ADC on Vertex AI, `global` endpoint.

## Root cause (as analyzed by my coding agent)

For the `EmbedContent` REST method, the transcoding binds `model` into the URL path **and also** serializes it into the JSON request body:

- `src/transport.rs` builds the path from `req.model` — `format!("/v1/{}:embedContent", var_model)` — then sends the **entire** request as the body via `handle_empty(Some(req), &method)` (no path-field projection).
- `src/model/serialize.rs`, `impl serde::ser::Serialize for EmbedContentRequest`, writes the field whenever it is set:

  ```rust
  if self.model.is_some() {
      state.serialize_entry("model", &self.model)?;
  }
  ```

So `model` ends up in **both** the URL path and the body. Because `EmbedContentRequest.model` is a proto3-`optional` (synthetic `_model` oneof), the Vertex server binds it from the path and then rejects the duplicate value in the body.

`PredictionService::generate_content` is unaffected only because `GenerateContentRequest.model` is a plain `string` (not an `optional`/oneof), so the duplicated value is tolerated by the server. Same transcoding shape, different field kind, different outcome.

dbolduc:
Omit path variables from the body of requests.

On testing: I opted for a fake HTTP server. While I did verify the RPC works against prod, I think the fake server is better for the CI than a live test hitting prod.

Fixes #6515 

Corresponds to librarian changes in https://github.com/googleapis/librarian/pull/7378

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 language: rust
-version: v0.38.1-0.20260821183352-4eaeba5d8e1a
+version: v0.38.1-0.20260824162858-1baf7d92dfda
 repo: googleapis/google-cloud-rust
 sources:
   conformance:
--- a/src/generated/ads/datamanager/v1/src/transport.rs
+++ b/src/generated/ads/datamanager/v1/src/transport.rs
@@ -358,6 +358,7 @@ impl super::stub::MarketingDataInsightsService for MarketingDataInsightsService 
         use gaxi::path_parameter::try_match;
         use gaxi::routing_parameter::Segment;
         use google_cloud_gax::error::binding::BindingError;
+        let mut req = req;
         let (builder, method, _path_template) = None
             .or_else(|| {
                 let var_parent = try_match(
@@ -372,6 +373,7 @@ impl super::stub::MarketingDataInsightsService for MarketingDataInsightsService 
                 let path = format!("/v1/{}/insights:retrieve", var_parent,);
                 let path_template = "/v1/{parent}/insights:retrieve";
 
+                let _ = Some(&mut req).map(|m| std::mem::take(&mut m.parent));
                 let builder = self.inner.builder(Method::POST, path);
                 let builder = Ok(builder);
                 Some(builder.map(|b| (b, Method::POST, path_template)))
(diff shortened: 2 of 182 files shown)


src/generated/cloud/aiplatform/v1/src/transport.rs
```diff
@@ -383431,6 +383949,7 @@ impl super::stub::PredictionService for PredictionService {
                 let path_template = "/v1/{model}:embedContent";
 
                 let resource_name = format!("//aiplatform.googleapis.com/{}", var_model,);
+                let _ = Some(&mut req).map(|m| std::mem::take(&mut m.model));
```
coryan, line 383952: Note to self: this seems to match the original bug report.

dbolduc: Yes, I repro'd the issue and confirmed this fixed it.

I should add it as an integration test in this PR...

dbolduc: I added a test. PTAL.

coryan (approved):

PR #6519, opened 2026-08-24
refactor(gax): consolidate grpc-rust test mocks

joshuatants:
For #5991.

### DIFF
--- a/src/gax-internal/src/grpc/grpc_rust/bidi.rs
+++ b/src/gax-internal/src/grpc/grpc_rust/bidi.rs
@@ -261,16 +261,12 @@ mod tests {
             MockRecvStream::new([
                 MockRecvAction::Wait(send_stream.notify_handle()),
                 MockRecvAction::Headers(ResponseHeaders::new().with_metadata(metadata)),
-                MockRecvAction::Message(TestMessage {
-                    value: RESPONSE_VALUE.to_string(),
-                }),
+                MockRecvAction::Message(TestMessage::new(RESPONSE_VALUE)),
                 MockRecvAction::Trailers(Trailers::new(Ok(()))),
             ]),
         );
         let headers = RequestHeaders::new().with_method_name(METHOD_NAME);
-        let request = TestMessage {
-            value: REQUEST_VALUE.to_string(),
-        };
+        let request = TestMessage::new(REQUEST_VALUE);
 
         // Act
         let response = invoke_bidi::<TestMessage, TestMessage, _>(
@@ -291,9 +287,7 @@ mod tests {
         let mut stream = response.into_inner();
         assert_eq!(
             stream.message().await?,
-            Some(TestMessage {
-                value: RESPONSE_VALUE.to_string()
-            })
+            Some(TestMessage::new(RESPONSE_VALUE))
         );
         assert_eq!(stream.message().await?, None);
         assert_eq!(
@@ -376,9 +370,7 @@ mod tests {
 
         let invoker = MockInvoker::new(FailingSendStream, PendingRecvStream);
         let headers = RequestHeaders::new().with_method_name(METHOD_NAME);
-        let request = TestMessage {
-            value: "msg".to_string(),
-        };
+        let request = TestMessage::new("msg");
 
         // Act
         let result = tokio::time::timeout(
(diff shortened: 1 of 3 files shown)


src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -480,27 +252,20 @@ mod tests {
...
             MockRecvStream::new([
-                MockRecvAction::WaitForMessage {
-                    observed_messages: observed_messages.clone(),
-                    notify,
-                },
+                MockRecvAction::Wait(send_stream.notify_handle()),
                 MockRecvAction::Headers(ResponseHeaders::new().with_metadata(metadata)),
                 MockRecvAction::Message(TestMessage {
                     value: RESPONSE_VALUE.to_string(),
                 }),
```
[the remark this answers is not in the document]
joshuatants: Done

src/gax-internal/src/grpc/grpc_rust/testing.rs
```diff
@@ -164,3 +274,58 @@ impl RecvStream for PanicRecvStream {
...
+
+impl<S, R> MockInvoker<S, R> {
+    /// Creates a new [`MockInvoker`] with the provided send and receive streams.
+    pub fn new(send_stream: S, recv_stream: R) -> Self {
+        Self {
+            send_stream: Mutex::new(Some(send_stream)),
+            recv_stream: Mutex::new(Some(recv_stream)),
+            observed_headers: Arc::new(Mutex::new(None)),
+        }
+    }
```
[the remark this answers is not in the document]
joshuatants: Done

vsharonlynn (approved):

src/gax-internal/src/grpc/grpc_rust/testing.rs
```diff
@@ -78,10 +133,48 @@ impl SendStream for PanicSendStream {
...
+        )
+    }
+}
+
+impl SendStream for FailAfterFirstSendStream {
+    async fn send(&mut self, _message: &dyn SendMessage, _options: SendOptions) -> Result<(), ()> {
+        self.sent_count += 1;
+        if self.sent_count == 1 {
+            return Ok(());
+        }
```
vsharonlynn, line 174: For my learning:
When sent_count >= 3, the below block will be executed again and the task will keep on running / hangs until the test timeouts? Is this the intended behaviour?

src/gax-internal/src/grpc/grpc_rust/testing.rs
```diff
@@ -38,6 +40,59 @@ impl TestMessage {
...
+    /// Returns the notification handle triggered when a message is sent.
+    pub fn notify_handle(&self) -> Arc<tokio::sync::Notify> {
+        self.notify.clone()
+    }
+}
+
+impl SendStream for MockSendStream {
+    async fn send(&mut self, message: &dyn SendMessage, options: SendOptions) -> Result<(), ()> {
+        let mut encoded = message.encode().map_err(|_| ())?;
+        let decoded = TestMessage::decode(&mut encoded).map_err(|_| ())?;
```
vsharonlynn: For my learning:
Do we need to expect anything during the encode and decode here?
Right now if the test passes a different message type then these two lines will return Err(()) ?
At [line 226](https://github.com/googleapis/google-cloud-rust/pull/6519/changes#diff-9d9dd83e75daf17f32179fdf7e2fe4cc1559a15365a66f5e134a100f97f6e65cR226) MockRecvStream::recv uses:
```
message
    .decode(&mut encoded)
    .expect("decode response message");
```

src/gax-internal/src/grpc/grpc_rust/testing.rs
joshuatants: The tests are set up to always expect the first send() to succeed and the second to fail: there will never be a third call to send(). To make this assumption more explicit, I added an `unreachable!` in case it's called a third time.

src/gax-internal/src/grpc/grpc_rust/testing.rs
joshuatants: Good catch. `MockSendStream` is the happy path mock so we don't expect `Err`s.

Changed to `expect` so that we fail fast.

PR #6520, opened 2026-08-24
test(storage): serialize transport tracing tests to fix flaky test

### LINKED ISSUE #6324: Flake: storage appendable object tracing missing finalize
**Failure in Google Cloud Build on `main`:**
- **Build:** `gcb-pm-test-msrv (rust-sdk-testing)`
- **Build Log:** https://console.cloud.google.com/cloud-build/builds;region=us-central1/ac3b31e8-33f8-4e95-ae64-9e89bb2ea18a?project=1044885715437
- **Commit:** 053bdfff4e4d345fafcc27457240d66d8040b601

**Associated Pull Requests:**
- #6323

vsharonlynn:
Fixes #6324 .

### DIFF
--- a/src/storage/src/storage/transport.rs
+++ b/src/storage/src/storage/transport.rs
@@ -459,6 +459,7 @@ mod tests {
     use google_cloud_test_utils::test_layer::{CapturedSpan, TestLayer};
     use httptest::{Expectation, Server, matchers::*, responders::status_code};
     use pretty_assertions::assert_eq;
+    use serial_test::serial;
     use std::collections::BTreeMap;
     use std::sync::Arc;
 
@@ -469,6 +470,7 @@ mod tests {
     }
 
     #[tokio::test]
+    #[serial]
     async fn read_object() -> anyhow::Result<()> {
         let guard = TestLayer::initialize();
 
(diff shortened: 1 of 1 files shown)


joshuatants:
Out of curiosity, why did allowing the tests to run in parallel cause flakiness?

joshuatants (approved):

coryan (changes requested):
> Out of curiosity, why did allowing the tests to run in parallel cause flakiness?

Because only one test can set the global tracing dispatcher at a time.  However, this is unlikely to work. This is serializing all the tests in this module, but cargo runs the tests in the **crate** in parallel. Any other test setting the global will break you.

There is a way to set the dispatcher for the current task, but that has failed us elsewhere.

We are finding that all these tracing tests are flaky across the board. We may need to move to `cargo nextest` for CI. In the interim, I think it is safer to leave them disabled.

vsharonlynn:
> > Out of curiosity, why did allowing the tests to run in parallel cause flakiness?
> 
> Because only one test can set the global tracing dispatcher at a time. However, this is unlikely to work. This is serializing all the tests in this module, but cargo runs the tests in the **crate** in parallel. Any other test setting the global will break you.
> 
> There is a way to set the dispatcher for the current task, but that has failed us elsewhere.
> 
> We are finding that all these tracing tests are flaky across the board. We may need to move to `cargo nextest` for CI. In the interim, I think it is safer to leave them disabled.

Ok, got it. Thanks, Carlos. I'll close this PR.

PR #6522, opened 2026-08-24
chore(spanner): core implementation for a dynamic channel pool

olavloite:
Introduce modular channel pooling supporting static fixed-size pools and dynamic load-based scaling with Power of Two Least Busy (P2C) selection, parallel channel priming, and transaction affinity routing.


src/spanner/src/channel_pool/scaler.rs
```diff
@@ -0,0 +1,2296 @@
...
+            .consecutive_low_load_checks
+            .store(0, Ordering::Relaxed);
+        return;
+    }
+
+    // Note: Scale-down deliberately evaluates real in-flight RPCs only (entry.in_flight()),
+    // excluding synthetic error penalties so failing channels are not retained as busy.
+    let total_in_flight: u32 = active_write.iter().map(|entry| entry.in_flight()).sum();
+    let avg_load = (total_in_flight as f64) / (active_write.len() as f64);
+
```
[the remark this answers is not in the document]
olavloite: This is intentional.

src/spanner/src/channel_pool/config.rs
```diff
@@ -0,0 +1,694 @@
...
+    pub(crate) prime_max_attempts: usize,
+    /// Channel selection strategy (default: PowerOfTwoLeastBusy).
+    pub(crate) selection_strategy: ChannelSelectionStrategy,
+}
+
+impl Default for DynamicChannelPoolConfig {
+    fn default() -> Self {
+        Self {
+            initial_channels: 4,
+            min_channels: 4,
```
rahul2393, line 151: In other langages and approved design we set this as min_channel: 2

src/spanner/src/channel_pool/pool.rs
```diff
@@ -0,0 +1,839 @@
...
+    }
+}
+
+impl ChannelPoolInner {
+    pub(crate) fn make_guard(&self, entry: Arc<ChannelEntry>) -> ActiveRpcGuard {
+        let (step, duration, max_penalty) = match &self.config {
+            ChannelPoolConfig::Dynamic(dynamic_config) => (
+                dynamic_config.error_penalty_step,
+                dynamic_config.error_penalty_duration,
+                dynamic_config.error_penalty_max,
```
rahul2393: In other languages, the max error penalty also depends on the maxRPCPerChannel config, which can differ from default if it's user-configurable

 Math.min(maxRpcPerChannel, current + errorPenaltyStep).

src/spanner/src/channel_pool/config.rs
```diff
@@ -0,0 +1,694 @@
...
+    /// Cooldown period between consecutive scale-up bursts (default: 10 seconds).
+    pub(crate) scale_up_cooldown: Duration,
+    /// Number of consecutive low-load checks required before scale-down (default: 3).
+    pub(crate) consecutive_low_load_checks: usize,
+    /// Maximum percentage of current pool size added per scale-up event (default: 30%, min 2).
+    pub(crate) max_scale_up_percent: u32,
+    /// Maximum number of channels marked draining per scale-down cycle (default: 2).
+    pub(crate) max_remove_channels: usize,
+    /// Idle grace duration a draining channel is kept alive after load drops to 0 (default: 5 seconds).
+    pub(crate) drain_idle_grace: Duration,
```
rahul2393, line 138: Similar to my other question why is it not a minute as agreed in desing?

src/spanner/src/channel_pool/pool.rs
```diff
@@ -0,0 +1,839 @@
...
+        let guard = self.inner.make_guard(entry);
+
+        // Signal scale-up if high load threshold is exceeded on dynamic pools.
+        // Works consistently across both standalone RPCs and pinned affinity transactions.
+        if self
+            .inner
+            .config
+            .dynamic_config()
+            .is_some_and(|dynamic_config| {
+                guard.entry.effective_pick_load() as f64 > dynamic_config.max_rpc_per_channel
```
rahul2393, line 137: We don't consider averages?

src/spanner/src/channel_pool/config.rs
olavloite: Yeah, I think that we need to revisit / rediscuss that. I think that `min_channel: 2` is too low, because it effectively means that very many systems will scale down to 2 channels relatively quickly, and then stay there. And having only 2 channels makes it more vulnerable to issues with one of the channels. The reason is the folllowing:

1. The threshold for scaling down is min_rpc_per_channel: 15
2. Assuming for simplicity an application that only does reads.
3. Standard read latency: ~5ms
4. Calculation for `desired_channels = ceil(total_in_flight / 20.0).max(min_channels)`
5. Assume that `total_in_flight <= 40` => `desired_channels = 2`
6. So at what QPS does `total_in_flight` fall below 40? That depends on the average latency:
    1. 5ms: `40 / 0.005s = 8,000 QPS`
    2. 10ms: `40 / 0.010s = 4,000 QPS`

Meaning: Very many systems would quickly scale down to 2 channels (and also stay there). I therefore strongly prefer sticking to `min_channels: 4`, which corresponds to the current number of channels that is used by default for the fixed channel pool.

rahul2393: mmm, ok thanks for clarification.

src/spanner/src/channel_pool/pool.rs
olavloite: Fixed (once new changes have been pushed)

src/spanner/src/channel_pool/config.rs
olavloite: The Rust implementation does not really need a grace period at all. It tracks all:
1. Active requests
2. Channel affiliations (for read/write transactions)

Once all active requests and channel affiliations have dropped to zero, it is guaranteed that no more traffic will ever be sent to the channel.

In addition, the implementation tracks the time since last request for a channel affiliation (so for a read/write transaction). If a read/write transaction does not send a request in 10s or more, then that affiliation is also dropped, as the transaction has been aborted by Spanner anyways. So waiting for 60s (or even 5s) before closing the channel is not really necessary.

I have changed it to 60s in this PR to be consistent with the other languages, but I will probably remove the entire option and corresponding logic in a follow-up PR.

src/spanner/src/channel_pool/pool.rs
olavloite: We do consider averages, but in the background scale-up job instead of on the hot path:

1. In `make_lease`, we only check if the load on the selected channel is > 25. That gives us a lock-free signal that the pool might need to scale.
2. If it is, we wake up the background worker (note: the background job is not scheduled, but woken by a notify signal).
3. The worker then calculates the average, and only scales up if it really needed.
4. Calculating the average directly in `make_lease` would require locking or a shared atomic counter across all channels on every single RPC, which we want to avoid on the hot path.

The check for average is [here](scaler.rs#L177-L185):

```rust
let total_load: u32 = active_guard
    .iter()
    .map(|entry| entry.effective_pick_load())
    .sum();
let desired_channels = config.desired_channel_count(total_load);
if desired_channels <= current_len {
    return 0;
}
```

rahul2393 (approved):

PR #6523, opened 2026-08-24
impl(bigquery): dry run queries should be routed to jobs.insert

alvarowolfx:
`jobs.query` don't provide full job statistics, so routing to `jobs.insert` is more useful. Also, since dry run queries don't have queries results or are marked as complete, we change examples to show just using a `Query` handle, not `until_done` or a `CompleteQuery`. Right now calling `until_done` it's going to throw a RPC Binding Error (empty job_id), but will eval if worth it introducing a new `QueryError::DryRunQuery`

Supersedes https://github.com/googleapis/google-cloud-rust/pull/6508

### DIFF
--- a/src/bigquery/src/query/execution.rs
+++ b/src/bigquery/src/query/execution.rs
@@ -150,8 +150,15 @@ impl RetryContext {
         }
     }
 
+    /// Returns true if the query should use the jobs.insert API.
+    fn force_job_path(&self) -> bool {
+        // jobs.query doesn't return full statistics information, so we route to jobs.insert
+        let dry_run = self.template.request.dry_run;
+        self.template.request.force_job_path() || dry_run
+    }
+
     async fn execute_once(&self, project_id: &str) -> Result<QueryHandle> {
-        if self.template.request.force_job_path() {
+        if self.force_job_path() {
             self.execute_jobs_insert(project_id).await
         } else {
             self.execute_jobs_query(project_id).await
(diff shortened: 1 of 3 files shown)


src/bigquery/src/query/execution.rs
```diff
@@ -150,8 +150,15 @@ impl RetryContext {
         }
     }
 
+    /// Returns true if the query should use the jobs.insert API.
+    fn force_job_path(&self) -> bool {
+        // jobs.query doesn't return full statistics information, so we route to jobs.insert
+        let dry_run = self.template.request.dry_run;
+        self.template.request.force_job_path() || dry_run
+    }
```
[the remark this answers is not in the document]
alvarowolfx: I did this exact change locally and it felt odd to return a `CompleteQuery` here, because `CompleteQuery` will not have the full job stats, as it derives from `jobs.getQueryResults` and `jobs.query`, so it misses having the jobs.insert data. We either have to make `CompleteQueryMetadata` to merge all 3 RPCs or have a new `QueryError` branch to mention that a dry run query can't become a `CompleteQuery`. So forcing users to just use the `Query` handle seems more logical, since the `metadata()` for it contains the full `Job` info

src/bigquery/src/query/execution.rs
```diff
@@ -406,4 +413,45 @@ mod tests {
         assert_eq!(query.metadata.job_reference, Some(job_ref));
         Ok(())
     }
+
+    #[test_case(true, "insert-job"; "dry_run routes to jobs.insert")]
+    #[test_case(false, "query-job"; "non dry_run routes to jobs.query")]
+    #[tokio::test]
+    async fn test_dry_run_routing(dry_run: bool, expected_job_id: &'static str) -> TestResult {
+        let mut mock = MockJobService::new();
+        if dry_run {
```
dbolduc: this suggests we want two separate tests

src/bigquery/src/query/execution.rs
```diff
@@ -150,8 +150,15 @@ impl RetryContext {
         }
     }
 
+    /// Returns true if the query should use the jobs.insert API.
+    fn force_job_path(&self) -> bool {
```
dbolduc, line 154: optional: arguably too clever, but we could exclude `QueryRequest.dry_run` in the librarian config and then we don't need this function.

src/bigquery/src/query/execution.rs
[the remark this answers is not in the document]
dbolduc: > or have a new `QueryError` branch to mention that a dry run query can't become a `CompleteQuery`

Probably this.

dbolduc (approved):

src/bigquery/src/query/execution.rs
alvarowolfx: that's a good idea, but dry_run would change to a `wkt::BoolValue` instead of a simple `bool` due to jobs.insert weirdness. Now I'm on a fence what is better here. If they fix things in `jobs.query` and we remove `QueryRequest.dry_run` the type is going to change.

dbolduc: Ah, good point. I forgot that `dry_run` was the one field that didn't line up. :unamused:

Issue #6405, opened 2026-08-14
BQ Write: support attaching to existing streams


dbolduc:
Applications that need to resume writing across processes can attach to an existing write stream by supplying its name. Reminder: the write stream is just a GCP resource.

Because our writers are strongly typed,
```rs
impl builder::ArrowWriter {
  pub async fn attach<U: TryFromStream>(self, write_stream: String) -> Result<U>;
}
```

Application code:
```rs
let writer: CommittedWriter = client
    .proto_writer(proto_schema())
    .attach("projects/p/datasets/d/tables/t/streams/s")
    .await?;
```

---

Implementation sketch:

https://play.rust-lang.org/?version=stable&mode=debug&edition=2024&gist=040a0c3472de935a0d503c1cd3f63fc7

We will need some trait, which we want to seal. The trait basically just tells us what `StreamType` each `Writer` is.

For naming... uh.... maybe just: `trait WriteStream` ? open to other suggestions. Customers shouldn't be naming this trait really.

PR #6527, opened 2026-08-24
refactor(bigquery): refactor duplicated target streams behind `BaseWriter`

haphungw:
For https://github.com/googleapis/google-cloud-rust/issues/6405

### DIFF
--- a/src/bigquery/src/write/arrow/buffered.rs
+++ b/src/bigquery/src/write/arrow/buffered.rs
@@ -12,7 +12,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-use super::base::BaseWriter;
+use super::inner::InnerWriter;
 use crate::Result;
 use crate::model::{ArrowRecordBatch, ArrowSchema, FinalizeWriteStreamResponse, FlushRowsResponse};
 use crate::write::append_builder::AppendWithOffset;
@@ -24,30 +24,30 @@ use std::sync::Arc;
 /// [buffered stream]: https://docs.cloud.google.com/bigquery/docs/write-api-grpc#buffered_type
 #[derive(Debug)]
 pub struct BufferedWriter {
-    pub(crate) base: BaseWriter,
+    pub(crate) inner: InnerWriter,
 }
 
 impl BufferedWriter {
     pub(crate) fn new(inner: Arc<Transport>, write_stream: String, schema: ArrowSchema) -> Self {
         Self {
-            base: BaseWriter::new(inner, write_stream, schema),
+            inner: InnerWriter::new(inner, write_stream, schema),
         }
     }
 
     /// Append rows to the buffered stream.
     pub fn append(&self, rows: ArrowRecordBatch) -> AppendWithOffset {
         AppendWithOffset::new(
-            self.base.runner.req_tx.clone(),
-            self.base.append_request(rows),
+            self.inner.runner.req_tx.clone(),
+            self.inner.append_request(rows),
         )
     }
 
     /// Flush the buffered stream, making rows up to the specified offset available for reading.
     pub async fn flush(&self, offset: i64) -> Result<FlushRowsResponse> {
-        self.base
+        self.inner
             .client
             .flush_rows()
-            .set_write_stream(&self.base.write_stream)
+            .set_write_stream(&self.inner.write_stream)
             .set_offset(offset)
             .send()
             .await
@@ -55,7 +55,7 @@ impl BufferedWriter {
 
     /// Finalize the buffered stream, preventing further writes.
     pub async fn finalize(&self) -> Result<FinalizeWriteStreamResponse> {
-        self.base.finalize().await
+        self.inner.finalize().await
     }
 }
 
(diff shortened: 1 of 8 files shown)


src/bigquery/src/write/arrow/buffered.rs
```diff
@@ -12,70 +12,50 @@
...
 /// A writer for a [buffered stream].
 ///
 /// [buffered stream]: https://docs.cloud.google.com/bigquery/docs/write-api-grpc#buffered_type
 #[derive(Debug)]
 pub struct BufferedWriter {
-    runner: Runner,
-    pub(crate) write_stream: String,
-    pub(crate) schema: ArrowSchema,
-    client: BigQueryWrite,
+    pub(crate) base: BaseWriter,
```
dbolduc: nit: we more commonly call this `inner`.

src/bigquery/src/write/arrow/default.rs
dbolduc, line 1: can you revert these changes?

The default stream implementation will be way different. It will involve multiplexing channel pools. And it has a different retry mechanism. I am not sure we will benefit from trying to consolidate it.

haphungw: make sense, done

src/bigquery/src/write/arrow/buffered.rs
```diff
@@ -12,70 +12,50 @@
...
     pub async fn flush(&self, offset: i64) -> Result<FlushRowsResponse> {
-        self.client
+        self.inner
+            .client
             .flush_rows()
-            .set_write_stream(&self.write_stream)
+            .set_write_stream(&self.inner.write_stream)
             .set_offset(offset)
             .send()
             .await
```
suzmue: nit: I would have expected self.inner.flush(offset).await instead of this request building, to match the way the others were refactored.

src/bigquery/src/write/arrow/inner.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::write::runner::Runner;
+use crate::write::transport::Transport;
+use std::sync::Arc;
+
+#[derive(Debug)]
+pub(crate) struct InnerWriter {
+    pub(crate) runner: Runner,
+    pub(crate) write_stream: String,
+    pub(crate) schema: ArrowSchema,
+    pub(crate) client: BigQueryWrite,
```
suzmue: nit: Consider which of these you actually want to be pub(crate). By keeping these internal to this module, it forces the rest of our crate to use the API that you expose in `impl InnerWriter`, which allows you to encapsulate the implementation. 

(If they all need to be pub(crate) because you need to access each of the fields outside of this module that is also totally fine.)

src/bigquery/src/write/arrow/inner.rs
```diff
@@ -0,0 +1,61 @@
...
+use crate::model::append_rows_request::ArrowData;
+use crate::model::{AppendRowsRequest, ArrowRecordBatch, ArrowSchema, FinalizeWriteStreamResponse};
+
+use crate::write::generated::gapic_storage::client::BigQueryWrite;
+use crate::write::runner::Runner;
+use crate::write::transport::Transport;
+use std::sync::Arc;
+
+#[derive(Debug)]
+pub(crate) struct InnerWriter {
```
dbolduc: nit: maybe `ExclusiveWriter`? Or `BaseWriter` or `WriterImpl`

src/bigquery/src/write/arrow/inner.rs
```diff
@@ -0,0 +1,61 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::Result;
+use crate::model::append_rows_request::ArrowData;
+use crate::model::{AppendRowsRequest, ArrowRecordBatch, ArrowSchema, FinalizeWriteStreamResponse};
+
```
dbolduc: nit: remove newline

src/bigquery/src/write/arrow/inner.rs
suzmue: Adding a doc comment would help clarify the API Boundary of this InnerWriter (why some are moved in and others are not even though they require no additional configuration from the parent writer)

My interpretation: It seems like this is meant to provide a convenient place to hold these subfields, and provide implementations for logic that needs to be shared for BufferedWriter, CommittedWriter, and PendingWriter. (Leaving anything that is not common to both of these to the implementations of those writers).

dbolduc (approved):

suzmue (approved):

src/bigquery/src/write/arrow/inner.rs
haphungw: done

src/bigquery/src/write/arrow/inner.rs
haphungw: done

src/bigquery/src/write/arrow/inner.rs
haphungw: thank you, done

PR #6528, opened 2026-08-24
chore: update Rust toolchain to 1.98

suzmue:
Updates the stable Rust compiler version from `1.97` to `1.98` across GitHub Actions and Google Cloud Build configurations.

### DIFF
--- a/.gcb/scripts/semver-checks.sh
+++ b/.gcb/scripts/semver-checks.sh
@@ -15,7 +15,7 @@
 
 set -ev
 
-cargo install --locked cargo-semver-checks@0.46.0
+cargo install --locked cargo-semver-checks@0.50.0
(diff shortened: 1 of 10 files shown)


.gcb/scripts/semver-checks.sh
```diff
@@ -15,7 +15,7 @@
 
 set -ev
 
-cargo install --locked cargo-semver-checks@0.46.0
```
coryan, line 18: We should bump the definition in `librarian.yaml` too?

coryan (approved):

.gcb/scripts/semver-checks.sh
suzmue: TIL, updated.

PR #6532, opened 2026-08-24
refactor: adopt let_chains in handwritten crates

suzmue:
Let chains were stabilized in Rust 1.88 and we now declare 1.90 as our MSRV.

Simplify nested pattern matches and tuple match constructs using let_chains across storage, auth, gax-internal, spanner, bigquery, and lro crates.

This flattens conditional logic when inspecting response headers, resolving external credentials, handling LRO polling states, and parsing telemetry metadata.

### DIFF
--- a/src/spanner/src/partitioned_dml_transaction.rs
+++ b/src/spanner/src/partitioned_dml_transaction.rs
@@ -239,9 +239,7 @@ async fn extract_lower_bound_update_count_from_stream(
             .take()
             .and_then(|cache_update| cache_update.cnv().ok());
         client.observe_cache_update(cache_update);
-        if let Some(stats) = partial_result_set.stats
-            && let Some(RowCountLowerBound(val)) = stats.row_count
-        {
+        if let Some(RowCountLowerBound(val)) = partial_result_set.stats.and_then(|s| s.row_count) {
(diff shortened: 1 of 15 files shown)


src/spanner/src/partitioned_dml_transaction.rs
```diff
@@ -239,7 +239,9 @@ async fn extract_lower_bound_update_count_from_stream(
             .take()
             .and_then(|cache_update| cache_update.cnv().ok());
         client.observe_cache_update(cache_update);
-        if let Some(RowCountLowerBound(val)) = partial_result_set.stats.and_then(|s| s.row_count) {
```
dbolduc, line 242: comment: I probably prefer an `and_then` chain to an `if let` chain. 🤷

dbolduc (approved):
How did you find these?

suzmue:
> How did you find these?

Gemini of course. I did not manually read through the whole repository, so entirely possible this does not identify all locations this improvement could be applied.

src/spanner/src/partitioned_dml_transaction.rs
suzmue: Fair enough. will revert this one.

PR #6533, opened 2026-08-24
refactor(pubsub): use extract_if in lease state management

suzmue:
Use HashMap::extract_if to evict expired leased messages in a single pass, avoiding intermediate Vec allocations and secondary key lookups.

This simplifies the retain loop by directly taking ownership of expired lease state info to send expiration errors on their respective channels, and consolidates the remaining lease extension logic across acking and leased statuses.

### DIFF
--- a/src/pubsub/src/subscriber/lease_state/exactly_once.rs
+++ b/src/pubsub/src/subscriber/lease_state/exactly_once.rs
@@ -105,36 +105,41 @@ impl Leases {
         max_lease_extension: Duration,
     ) -> Vec<Vec<String>> {
         let now = Instant::now();
+        let mut batches = Vec::new();
+        let mut batch = Vec::new();
 
-        // Extract and notify expired leases.
         self.under_lease
-            .extract_if(|_id, info| {
-                matches!(info.status, MessageStatus::Leased) && info.receive_time + max_lease < now
+            .extract_if(|id, info| match info.status {
+                MessageStatus::Leased if info.receive_time + max_lease < now => {
+                    // Extract expired leases.
+                    true
+                }
+                MessageStatus::Nacking => false,
+                MessageStatus::Acking | MessageStatus::Leased => {
+                    if info.last_extension.is_some_and(|i| {
+                        i + max_lease_extension > now + EXTEND_PERIOD + EXTEND_BUFFER
+                    }) {
+                        // The current lease is valid for a while. Retain the message,
+                        // but do not extend its lease.
+                    } else {
+                        // Extend leases for all other messages.
+                        batch.push(id.clone());
+                        if batch.len() == MAX_IDS_PER_RPC {
+                            // Flush the batch when it is full.
+                            batches.push(std::mem::take(&mut batch));
+                        }
+                    }
+                    false
+                }
             })
             .for_each(|(_id, info)| {
                 let _ = info.result_tx.send(Err(AckError::LeaseExpired));
             });
 
-        self.under_lease
-            .iter()
-            .filter_map(|(id, info)| match info.status {
-                MessageStatus::Nacking => None,
-                MessageStatus::Acking | MessageStatus::Leased => {
-                    if info.last_extension.is_some_and(|i| {
-                        i + max_lease_extension > now + EXTEND_PERIOD + EXTEND_BUFFER
-                    }) {
-                        // The lease is still valid for a while. No need to extend.
-                        None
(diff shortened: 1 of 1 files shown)


src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -106,61 +106,35 @@ impl Leases {
...
-            .collect::<Vec<_>>();
-
-        expired
-            .into_iter()
-            .filter_map(|id| self.under_lease.remove_entry(&id))
-            .for_each(|(_id, info)| {
-                let _ = info.result_tx.send(Err(AckError::LeaseExpired));
-            });
-        remaining
+            .collect::<Vec<_>>()
```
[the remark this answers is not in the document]
suzmue: This is existing code not introduced by this refactor. There are other places within pub/sub where we do this chunking so would proabbly want to update those as well so not doing in this PR.

src/pubsub/src/subscriber/lease_state/exactly_once.rs
```diff
@@ -106,61 +106,35 @@ impl Leases {
...
+        // Extract and notify expired leases.
+        self.under_lease
+            .extract_if(|_id, info| {
+                matches!(info.status, MessageStatus::Leased) && info.receive_time + max_lease < now
+            })
+            .for_each(|(_id, info)| {
+                let _ = info.result_tx.send(Err(AckError::LeaseExpired));
+            });
+
+        self.under_lease
```
dbolduc: I think the idea was that if we use `extract_if()` we can both collect the IDs we need to extend and remove expired leases in one pass over the data structure.

This code is doing two iterations.

dbolduc: Gemini came up with this impl which passes my eye test and our unit tests and fixes the clones from `chunks()`:

```rs
    pub fn retain(
        &mut self,
        max_lease: Duration,
        max_lease_extension: Duration,
    ) -> Vec<Vec<String>> {
        let now = Instant::now();
        let mut batches = Vec::new();
        let mut batch = Vec::new();

        self.under_lease
            .extract_if(|id, info| {
                if matches!(info.status, MessageStatus::Leased) && info.receive_time + max_lease < now {
                    // Extract expired leases.
                    true
                } else {
                    match info.status {
                        MessageStatus::Nacking => {}
                        MessageStatus::Acking | MessageStatus::Leased => {
                            if info.last_extension.is_some_and(|i| {
                                i + max_lease_extension > now + EXTEND_PERIOD + EXTEND_BUFFER
                            }) {
                                // The current lease is valid for a while. Retain the message,
                                // but do not extend its lease.
                            } else {
                                // Extend leases for all other messages.
                                batch.push(id.clone());
                                if batch.len() == MAX_IDS_PER_RPC {
                                    // Flush the batch when it is full.
                                    batches.push(std::mem::take(&mut batch));
                                }
                            }
                        }
                    }
                    false
                }
            })
            .for_each(|(_id, info)| {
                let _ = info.result_tx.send(Err(AckError::LeaseExpired));
            });

        if !batch.is_empty() {
            batches.push(batch);
        }
        batches
    }
```

dbolduc (changes requested):

src/pubsub/src/subscriber/lease_state/exactly_once.rs
suzmue: Cool. Will update (though from a readability I prefer the current one but performance wins).

This implementation was still an improvement than previous because it didn't have to do each self.under_lease.remove_entry via a lookup (its iterating straight over them instead).

suzmue: FWIW the performance diff I could measure was 1% for the combining the loops, vs the original PR. Personally, I'm not convinced that the performance boost of lumping them into a single loop is worth the reduced readability but I made the change.

(fixing the allocation does yield a more significant performance diff of 15-20%)

dbolduc (approved):

src/pubsub/src/subscriber/lease_state/exactly_once.rs
dbolduc: Thanks for measuring something.

I generally think it's a good habit to write optimal code (even in this case when any performance gain is going to get dominated by the latency of a network call).

Sometimes it does matter, and 1% here, 1% there can start adding up.

PR #6534, opened 2026-08-24
feat(bigquery): abstract attachable writers

haphungw:
Isolate the foundation to support stream attachment.

For https://github.com/googleapis/google-cloud-rust/issues/6405

### DIFF
--- a/src/bigquery/src/write/arrow/buffered.rs
+++ b/src/bigquery/src/write/arrow/buffered.rs
@@ -40,6 +40,7 @@ impl BufferedWriter {
         &self.inner.write_stream
     }
 
+    /// Append rows to the stream.
     pub fn append(&self, rows: ArrowRecordBatch) -> AppendWithOffset {
         AppendWithOffset::new(
             self.inner.runner.req_tx.clone(),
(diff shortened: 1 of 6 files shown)


src/bigquery/src/write/arrow/committed.rs
```diff
@@ -35,6 +35,11 @@ impl CommittedWriter {
     }
 
     /// Append rows to the stream.
```
dbolduc: and this

src/bigquery/src/write/arrow/pending.rs
```diff
@@ -37,6 +37,11 @@ impl PendingWriter {
     }
 
     /// Appends rows to the pending stream.
```
dbolduc: and this

src/bigquery/src/write/arrow/buffered.rs
```diff
@@ -35,6 +35,12 @@ impl BufferedWriter {
     }
 
     /// Append rows to the buffered stream.
```
dbolduc: Need to delete this

PR #6536, opened 2026-08-25
fix(gax): distinguish connection from RPC errors using ConnectionInfo

joshuatants:
For #5991.

### DIFF
--- a/src/gax-internal/src/grpc/grpc_rust/bidi.rs
+++ b/src/gax-internal/src/grpc/grpc_rust/bidi.rs
@@ -435,11 +435,24 @@ mod tests {
         // Arrange
         use std::error::Error;
         const METHOD_NAME: &str = "/google.test.v1.Test/Bidi";
-        const ERROR_MESSAGE: &str = "Service was not ready: connection refused to 127.0.0.1:1";
-        let connect_error = StatusError::new(StatusCodeError::Unavailable, ERROR_MESSAGE);
+        const TEST_HEADER: &str = "x-test-header";
+        const TEST_VALUE: &str = "test-value";
+        const GRPC_STATUS_DETAILS_BIN: &str = "grpc-status-details-bin";
+        const ERROR_MESSAGE_UNAVAILABLE: &str = "Service was not ready: transport connect failed";
+
+        let details = b"status-details";
+        let mut metadata = grpc::metadata::MetadataMap::new();
+        metadata.insert_bin(GRPC_STATUS_DETAILS_BIN, MetadataValue::from_bytes(details));
+        metadata.insert(TEST_HEADER, MetadataValue::from_static(TEST_VALUE));
+        let trailers = Trailers::new(Err(StatusError::new(
+            StatusCodeError::Unavailable,
+            ERROR_MESSAGE_UNAVAILABLE,
+        )))
+        .with_metadata(metadata);
+
         let invoker = MockInvoker::new(
             FailingSendStream,
-            MockRecvStream::with_immediate_trailers(Trailers::new(Err(connect_error))),
+            MockRecvStream::with_immediate_trailers(trailers),
         );
         let headers = RequestHeaders::new().with_method_name(METHOD_NAME);
         let request = TestMessage::new("request");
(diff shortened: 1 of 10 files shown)


src/gax-internal/src/grpc/grpc_rust/receive.rs
```diff
@@ -197,19 +202,40 @@ async fn receive_responses<Response, R>(
...
         details,
         metadata,
-    ))
+    );
+
+    if !seen_headers
+        && trailers.connection_info().is_none()
+        && status.code() == StatusCodeError::Unavailable
+    {
+        return Some(tonic::Status::from_error(Box::new(ConnectError(
```
xlai20, line 233: A lot of wrapping of this error status here.

IIUC the output of this function `trailers_to_tonic_status` will be further passed on to `Error::connect(status)` in from_status.rs, is it? there seems to be a nested error chain here. Please double check.

src/storage/src/storage/bidi_write/connector.rs
```diff
@@ -354,14 +354,10 @@ mod tests {
         need_send(&start);
 
         let err = start.await.unwrap_err();
-        assert!(err.is_connect(), "{err:?}");
-        // TODO(#5991): Re-enable when `grpc-rust` preserves `.source()` and endpoint URIs.
-        // See `to_gax_error` in `gax-internal/src/grpc/from_status.rs` for more background.
-        #[cfg(not(google_cloud_unstable_grpc_rust))]
-        {
-            let source = err.source().unwrap().to_string();
```
xlai20, line 362: why is the verification of source being deleted here?

xlai20 (changes requested):

src/gax-internal/src/grpc/grpc_rust/bidi.rs
```diff
@@ -424,6 +430,71 @@ mod tests {
...
+        let status = invoke_bidi::<TestMessage, TestMessage, _>(
+            &invoker,
+            headers,
+            tokio_stream::iter([request]),
+        )
+        .await
+        .expect_err("should fail with connect error");
+
+        // Assert
+        let gax_err = crate::grpc::from_status::to_gax_error(status);
```
vsharonlynn: Should we import at the top of test mod definition for better readability?

src/gax-internal/src/grpc/grpc_rust/receive.rs
```diff
@@ -197,19 +202,40 @@ async fn receive_responses<Response, R>(
...
+            tonic_status,
```
vsharonlynn, line 234: tonic::Status::from_error wraps ConnectError (which envelopes tonic::Status). Is this intended?

vsharonlynn (changes requested):

src/storage/src/storage/bidi_write/connector.rs
joshuatants: 1. There's no guarantee that the `source` string contains the target IP+port - this is platform specific behaviour.
2. I checked with @vsharonlynn who added this check - we couldn't find a reason why source verification was necessary.
3. Similar tests, e.g. in `bidi/connector.rs`and elsewhere in the repo don't check for `source`, or for `is_connect`.

src/gax-internal/src/grpc/grpc_rust/receive.rs
joshuatants: Yes, there's a nested error chain, but this is deliberate.

Inner `tonic::Status`: Holds the actual gRPC error data (status code Unavailable, message, headers, and `grpc-status-details-bin`). This is packaged into a `ConnectError` so callers can access the error details if needed.

Outer `tonic::Status`: The internal API passes results as `tonic::Result<Response<...>>`. In order to pass the `ConnectError` around,  we need to use `Status::from_error(ConnectError)`. `Status::from_error` produces another `tonic::Status`.

src/gax-internal/src/grpc/grpc_rust/receive.rs
joshuatants: Yep, see my reply in https://github.com/googleapis/google-cloud-rust/pull/6536#discussion_r3850286534

src/gax-internal/src/grpc/grpc_rust/bidi.rs
joshuatants: Done.

vsharonlynn (approved):

xlai20 (approved):

Issue #6461, opened 2026-08-19
feat(storage): Support trailing checksum metadata for single-shot REST uploads


xlai20:
**Is your feature request related to a problem? Please describe.**
Currently, when performing unbuffered, single-shot object uploads via the JSON (REST) API, the payload is streamed continuously in a single HTTP `PUT` request for maximum performance. However, because HTTP/1.1 requires all headers to be sent *before* the request body, we cannot append an on-the-fly computed CRC32C or MD5 checksum via the traditional `x-goog-hash` header. Achieving this currently requires passing over the payload twice (once to pre-compute the checksum and once to stream the data), creating a significant disk I/O performance penalty.

**Describe the solution you'd like**
To provide server-side validation against network data corruption without the double-read overhead, we should leverage the GCS JSON API's multipart trailing metadata feature. 

**Implementation Constraints**
This trailing checksum strategy should be mutually exclusive with other kinds of checksum validation opt-in'ed by users.

b/521188176

xlai20:
Updated the title of this issue to increase its scope of coverage for all `Seek` data upload in GCS. `Seek` data are rewindable data which is usually the data in the disk; this contrasts with non-`Seek` data which is stream/network/pipe data that is non-rewindable.

**Single-shot:** 
- Using trailing checksum metadata to support the server-side checksumming. This is already merged in #6537 .

**Resumable Upload for large object:**
- Currently served as a opt-in feature via the `precompute_checksums()` function, but a recent benchmarking effort (#6611 ) has proved that the performance penalty is in an acceptable range. Hence the next step is to turn on precomputation of checksums.

PR #6537, opened 2026-08-25
feat(storage): support trailing checksum metadata for single-shot unbuffered uploads

xlai20:
Part of https://github.com/googleapis/google-cloud-rust/issues/6461

## Description
This PR implements trailing checksum metadata for single-shot unbuffered (`Seek` data source) REST uploads in Google Cloud Storage.

Previously, streaming single-shot uploads from a `Seek` source either required a separate precomputation disk pass (`precompute_checksums()`) or omitted server-side checksum validation during on-the-fly streaming. 

This change leverages the GCS JSON multipart upload protocol by appending a third, trailing JSON metadata part containing the on-the-fly computed CRC32C (and optionally MD5) checksums. This allows GCS to validate data integrity upon receiving the final chunk with zero RAM buffering and in a single disk pass.

---

### 1. Multipart Payload Construction (`src/storage/src/storage/perform_upload/unbuffered.rs`)
- Modified `single_shot_builder()` to assemble a 3-part multipart request:
  - **Part 1 (`metadata`):** JSON object metadata (e.g. name, generation match, metadata) without upfront checksums.
  - **Part 2 (`media`):** File payload streamed directly from `Seek(0)`.
  - **Part 3 (`checksums`):** Trailing JSON metadata containing base64-encoded `crc32c` and/or `md5Hash`.
- Constructed Part 3 symmetrically to Part 2 via a lazy stream (`trailing_checksums_to_body()`). This guarantees the checksum metadata is evaluated only after the media stream completes and the hash accumulator is finalized.
- Enforced mutual exclusivity: Part 3 is only appended when on-the-fly calculation is active (`self.options.checksum`). When upfront checksums are provided or precomputed (`self.resource().checksums`), Part 3 is omitted and checksums are placed into Part 1 as before.

### 2. Unit Tests (`src/storage/src/storage/perform_upload/unbuffered/single_shot_tests.rs`)
- Updated `parse_multipart_body()` test helper to extract and validate the optional 3rd multipart field.
- Added `mod trailing_checksums` covering:
  - `default_expecting_trailing_crc32c`
  - `compute_md5_expecting_trailing_both`
  - `known_crc32c_expecting_no_trailing`
  - `known_both_expecting_no_trailing`
  - `known_md5_expecting_trailing_crc32c`
  - `precompute_expecting_no_trailing`

### 3. Live Integration Tests (`tests/storage/src/write_object.rs`)
- Added `"unbuffered/verify/default"` to `write_object::checksums()` to verify the 3-part multipart trailing checksum wire format against live Google Cloud Storage infrastructure.

---

## Verification
- `cargo fmt --check`
- `cargo check --all-targets`
- `cargo test --package google-cloud-storage --lib single_shot_tests` (All 25 tests pass)
- `cargo test --package google-cloud-storage --lib checksums` (All 40 tests pass)


src/storage/src/storage/perform_upload/unbuffered.rs
```diff
@@ -204,14 +205,33 @@ where
             .seek(0)
             .await
             .map_err(Error::ser)?;
+
+        // Multipart part 2: Media payload stream
         let payload = self.payload_to_body().await?;
-        let form = multipart::Form::new().part("metadata", metadata);
-        let form = if let Some(exact) = hint.exact() {
+        let mut form = multipart::Form::new().part("metadata", metadata);
+        form = if let Some(exact) = hint.exact() {
```
joshuatants: This doesn't look like it needs `mut` - shadowing was (and still should be) enough?

src/storage/src/storage/perform_upload/unbuffered.rs
```diff
@@ -204,14 +205,33 @@ where
...
             form.part("media", multipart::Part::stream_with_length(payload, exact))
         } else {
             form.part("media", multipart::Part::stream(payload))
         };
 
+        // Multipart part 3: Trailing checksums metadata
+        // `self.options.checksum` tracks checksum state for on-the-fly streaming calculation.
+        // When known hashes are provided or `precompute_checksums()` is called, `self.options.checksum`
+        // is deactivated while the checksum is stored separately in `self.resource().checksums`.
```
joshuatants: WDYT of adding an `assert!` to catch an violations of this, e.g. somehow we have self.options.checksum AND self.resource().checksums both set.

src/storage/src/storage/perform_upload/unbuffered.rs
```diff
@@ -204,14 +205,33 @@ where
...
+        // Hence, an active `self.options.checksum` indicates no upfront checksum exists, so we
```
joshuatants: Nit i think we can drop "Part 3", i.e. "append the on-the-fly hash via trailing metadata".

src/storage/src/storage/perform_upload/unbuffered.rs
```diff
@@ -204,14 +205,33 @@ where
...
-        let form = multipart::Form::new().part("metadata", metadata);
-        let form = if let Some(exact) = hint.exact() {
+        let mut form = multipart::Form::new().part("metadata", metadata);
+        form = if let Some(exact) = hint.exact() {
             form.part("media", multipart::Part::stream_with_length(payload, exact))
         } else {
             form.part("media", multipart::Part::stream(payload))
         };
 
+        // Multipart part 3: Trailing checksums metadata
```
joshuatants, line 218: Double checking: I believe this code path is shared with buffered single-shot too? In buffered.rs `send_buffered_single_shot`, the checksum is precalculated and put into the `spec`, the checksums are then disabled, and `options` is copied over. This means that L225 here could still pass for the buffered if `options` sets checksums to true. Will this cause any errors?

src/storage/src/storage/perform_upload/unbuffered.rs
```diff
@@ -204,14 +205,33 @@ where
...
+        // `self.options.checksum` tracks checksum state for on-the-fly streaming calculation.
+        // When known hashes are provided or `precompute_checksums()` is called, `self.options.checksum`
```
joshuatants: We should update the doc here as it's no longer accurate after this PR: https://github.com/googleapis/google-cloud-rust/blob/b21daf27743ac2da0205bdbcb9570f254b4e6a06/src/storage/src/storage/write_object.rs#L1055-L1056

joshuatants: "`self.options.checksum`   is deactivated " make it sound like all checksumming is switched off. Can the docstring be updated so that it's clear that it's on a per algorithm basis?

joshuatants (changes requested):

src/storage/src/storage/perform_upload/unbuffered.rs
xlai20: Thank you for spotting this. Wow, I didn't expect deep inside the "buffered.rs" it is silently calling a function "unbuffered.rs", this existing code is quite an anti-pattern. I think it violates the principle of least astonishment.

So the current behavior (before my fix) is that the send_buffered single-shot will upload both the checksum in the Part 1 header and and Part 3 trailing metadata.

So here is my fix:
-  I have updated `single_shot_builder()` to explicitly check whether hash already exists upfront in `self.resource().checksums`. In this way, if upfront checksums are already present in Part 1 (as is the case when delegated from `buffered.rs`), Part 3 trailing metadata is now suppressed, guaranteeing clean 2-part multipart uploads without redundant trailing parts.
- The fix will make sure the send_buffered behave exactly same as what it does before.

However, I think the current buffered.rs needs some more fix and I will do it in follow-up PRs, because for this PR I want to just keep the scope to changing only unbuffered single-shot.
1. De-activate `self.options.checksum = Checksum::default()` in `send_buffered_single_shot()` to synchronize `options` with `upload.payload`.
2. Consider refactoring both modules to share a common `multipart.rs` builder rather than having `buffered.rs` delegate into `unbuffered.rs`.

src/storage/src/storage/perform_upload/unbuffered.rs
xlai20: I think there's no need to add assert! here given the latest change in the if statement of Part 3.

src/storage/src/storage/perform_upload/unbuffered.rs
xlai20: I think having the "Part 3" comment makes things clearer--this is a multipart http request, and each part is added in a clear manner.

src/storage/src/storage/perform_upload/unbuffered.rs
xlai20: Done.

src/storage/src/storage/perform_upload/unbuffered.rs
xlai20: With the latest update in the comment, I think the word "self.options.checksum is deactivated" is no longer used, we just say on-the-fly streaming calculation is active.

src/storage/src/storage/perform_upload/unbuffered.rs
joshuatants: I meant as in instead of saying "Part 3 trailing metadat", we just have "trailing metadata", since the comment is already in a block that starts with "Multipart part 3".

Not a big deal.

src/storage/src/storage/perform_upload/unbuffered.rs
joshuatants: Was this addressed? Or do you plan to update this in a future PR?

src/storage/src/storage/perform_upload/unbuffered.rs
joshuatants: Would it be better to first fix bufferred.rs (or at least decouple it from the code here) before updating 'single_shot_builder'? Otherwise, there's a risk we break buffered.rs.

src/storage/src/storage/perform_upload/unbuffered.rs
xlai20: Done. I do a minor update on the function doc here. 

But I think we need to discuss on how we should position the `precompute_checksums()` function to the user. My feeling is that the send_buffered() cases this function is not relevant. That can be a separate PR.

src/storage/src/storage/perform_upload/unbuffered.rs
xlai20: I'm pretty sure that the buffered single-shot path will not be affected by this change, as newly added code is now protected by `!has_upfront_crc32c` condition. In buffered.rs, when the payload is buffered in memory, its CRC32C is computed, and spec.resource.checksums is populated upfront. 

To address your concern of regression, I just added a unit test `buffered_single_shot_expecting_two_parts`
 in buffered/single_shot_tests.rs. It executes send_buffered() through a mock server, parses the resulting multipart payload with multer, and asserts that no 3rd trailing metadata part is generated.

Why am I not doing the de-coupling right now:
- send_buffered and send_unbuffered single shots are still sharing a huge block of common code on constructing the first 2 parts of http multipart request, which comes with a lot of other things like retry-policy construction.
- Decoupling them either means a lot of copy-and-paste code, or extracting a shared builder abstraction across modues, which would significantly expand the scope of this PR. 
- Hence, I prefer to keep the current status quo, and consider moving the single_shot_builder out into a shared module in a separate refactoring PR.

src/storage/src/storage/perform_upload/unbuffered.rs
joshuatants: SGTM we can discuss offline.

joshuatants (approved):

PR #6546, opened 2026-08-25
chore(ci): add rust-toolchain-bump skill and instructions

suzmue:
Add the rust-toolchain-bump skill and a howto guide for humans to assist with detecting new stable Rust minor compiler releases, executing workspace clippy checks, and updating CI configurations across Google Cloud Build and GitHub Actions.


coryan (commented):
This seems deterministic ... could we ask the LLM to write us a script instead of asking the LLM to behave like a script?

suzmue:
Extracted initial deterministic steps into a script.

Left the following non-deterministic parts separate:
- The files that reference the current version are also subject to change, so I'd worry about hardcoding them. 
- using a script to replace the version number 1.YY is brittle, other dependencies could have the same version. 
- Updating the clippy lints is also not deterministic.

.agents/skills/rust-toolchain-bump/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+______________________________________________________________________
+
+## Step 2: Handle Diagnostics & Tooling Issues
+
+- **If `check-rust-toolchain.sh` exits successfully:** Proceed directly to Step
+  3\.
+- **If clippy fails in handwritten crates (`src/auth`, `src/gax`, `src/storage`,
+  etc.):** Fix code diagnostics directly in the working branch, then re-run
+  `.agents/skills/rust-toolchain-bump/scripts/check-rust-toolchain.sh` until
+  clean.
```
coryan: Most of the time this should work, no?

https://doc.rust-lang.org/clippy/usage.html#automatically-applying-clippy-suggestions

.agents/skills/rust-toolchain-bump/SKILL.md
```diff
@@ -0,0 +1,102 @@
...
+- **If clippy fails in generated code (`src/generated/`):** Do NOT edit
```
coryan: there is generated code in `src/*/generated/` too ...

coryan (commented):
You need somebody more enthusiastic about LLMs than I am to approve.  I would rather run this manually myself, but then again, I am not the one running it most of the time.

Just a couple of nits.

suzmue:
> I would rather run this manually myself.

It is not required to use the skill, just there as an option 😄  I'll consider moving the script to a more human-user friendly place so we can share it with our robot helpers.

> I am not the one running it most of the time.

I have some hope this can be run by someone on the broader client libraries team as well, that may not have (and may not want to have) all of the context to create the PR themselves.

> Just a couple of nits.

Will fix.

.agents/skills/rust-toolchain-bump/SKILL.md
suzmue: Yes. I added it to the script so that we in the common case everything is fixed via script.

.agents/skills/rust-toolchain-bump/SKILL.md
suzmue: Updated.

suzmue:
Added instructions for people as well in howto guide

.agents/skills/rust-toolchain-bump/SKILL.md
```diff
@@ -0,0 +1,143 @@
...
+Common files to update include:
+
+1. **GitHub Actions Workflows**:
+
+   - `.github/workflows/sdk.yaml`
+     (`GHA_RUST_VERSIONS: '{ "rust:current": "1.XX" }'`)
+   - `.github/workflows/rust-toolchain-check.yaml`
+     (`CURRENT_RUST_VERSION: '1.XX'`)
+
+1. **Google Cloud Build Configurations**: Update all `_RUST_VERSION: '1.YY'`
```
haphungw: nit: `1.XX` everywhere for consistency?

haphungw (approved):

.agents/skills/rust-toolchain-bump/SKILL.md
suzmue: Done!

scripts/check-rust-toolchain.sh
```diff
@@ -0,0 +1,85 @@
...
+set -euo pipefail
+
+WORKFLOW_FILE=".github/workflows/rust-toolchain-check.yaml"
+
+if [[ ! -f "${WORKFLOW_FILE}" ]]; then
+  echo "Error: ${WORKFLOW_FILE} not found." >&2
+  exit 1
+fi
+
+CURRENT_VERSION=$(grep -oP "(?<=CURRENT_RUST_VERSION: ')[0-9.]+" "${WORKFLOW_FILE}")
```
haphungw: I ran the script on my local mac and encountered this error.
```
grep: invalid option -- P
usage: grep [-abcdDEFGHhIiJLlMmnOopqRSsUVvwXxZz] [-A num] [-B num] [-C[num]]
	[-e pattern] [-f file] [--binary-files=value] [--color=when]
	[--context[=num]] [--directories=action] [--label] [--line-buffered]
	[--null] [pattern] [file ...]
```

Gemini suggests we change this from `grep -oP` to `sed -nE` to support both Linux and macOS.

suzmue: done.

PR #6549, opened 2026-08-25
chore(ci): add MSRV 1-year check to scheduled workflow

suzmue:
Add an automated step to .github/workflows/rust-toolchain-check.yaml to calculate MSRV eligibility under the 1-year policy and file a tracking issue when we can drop a version.


dbolduc (approved):
Nice

PR #6555, opened 2026-08-26
impl(bigquery-read): support server-side streaming for ReadRows

suzmue:
Enable include_server_streaming_methods for google-cloud-bigquery-read in librarian.yaml and regenerate the client with server-side streaming support for the ReadRows RPC.

Also add an integration test.

### DIFF
--- a/src/bigquery-read/src/generated/gapic_storage/builder.rs
+++ b/src/bigquery-read/src/generated/gapic_storage/builder.rs
@@ -166,6 +166,105 @@ pub mod read {
         }
     }
 
+    /// The request builder for [Read::read_rows][crate::client::Read::read_rows] calls.
+    ///
+    /// # Example
+    /// ```
+    /// # use google_cloud_bigquery_read::builder::read::ReadRows;
+    /// # async fn sample() -> google_cloud_bigquery_read::Result<()> {
+    /// let builder = prepare_request_builder();
+    /// let mut receiver = builder.send().await?;
+    /// # Ok(()) }
+    ///
+    /// fn prepare_request_builder() -> ReadRows {
+    ///   # panic!();
+    ///   // ... details omitted ...
+    /// }
+    /// ```
+    #[cfg(google_cloud_unstable_gapic_streaming)]
+    #[derive(Clone, Debug)]
+    pub struct ReadRows(RequestBuilder<crate::model::ReadRowsRequest>);
+
+    #[cfg(google_cloud_unstable_gapic_streaming)]
+    impl ReadRows {
+        pub(crate) fn new(stub: std::sync::Arc<dyn super::super::stub::dynamic::Read>) -> Self {
+            Self(RequestBuilder::new(stub))
+        }
+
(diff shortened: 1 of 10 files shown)


tests/bigquery/src/reads.rs
```diff
@@ -39,15 +40,31 @@ pub async fn run_reads() -> Result<()> {
...
     }
     .await;
 
     let _ = delete_dataset(&dataset_service, &project_id, &dataset_id).await;
     result
 }
 
-// Calls the one unary RPC the service has to offer.
+// Calls the unary CreateReadSession RPC.
 pub async fn basic(project_id: &str, dataset_id: &str, table_id: &str) -> Result<()> {
```
dbolduc: comment: Feel free to delete this. My intention was to `read_rows` in here when it became available.

tests/bigquery/src/reads.rs
```diff
@@ -66,3 +83,43 @@ pub async fn basic(project_id: &str, dataset_id: &str, table_id: &str) -> Result
...
+        .set_read_stream(stream_name)
+        .send()
+        .await?;
+
+    let mut total_rows = 0;
+    while let Some(response) = stream.recv().await {
+        let response = response?;
+        total_rows += response.row_count;
+    }
+    assert_eq!(total_rows, 3, "expected 3 rows to be read from stream");
```
dbolduc: maybe gate the read rows piece only?

```suggestion
#[cfg(google_cloud_unstable_gapic_streaming)]
{
    let stream_name = &session.streams[0].name;
    let mut stream = client
        .read_rows()
        .set_read_stream(stream_name)
        .send()
        .await?;

    let mut total_rows = 0;
    while let Some(response) = stream.recv().await {
        let response = response?;
        total_rows += response.row_count;
    }
    assert_eq!(total_rows, 3, "expected 3 rows to be read from stream");
}
```

dbolduc (approved):

tests/bigquery/src/reads.rs
suzmue: I originally had this, but I thought it was less clear to have the test only half run. But if we want to ditch the basic test then it would make sense to have something testing the client set up.

suzmue: Done.

tests/bigquery/src/reads.rs
suzmue: Done.

src/bigquery-read/src/generated/gapic_storage/builder.rs
```diff
@@ -166,6 +166,105 @@ pub mod read {
...
+        }
+
+        /// Sets all the options, replacing any prior values.
+        pub fn with_options<V: Into<crate::RequestOptions>>(mut self, v: V) -> Self {
+            self.0.options = v.into();
+            self
+        }
+
+        /// Initiates the server stream.
+        pub async fn send(
```
dbolduc, line 207: 😎

dbolduc (approved):

PR #6559, opened 2026-08-26
fix(secretmanager): apply retry policy to openapi integration tests

haphungw:
This proactively prevents `503 Unavailable` flakes in the REST clients similar to the protobuf integration tests.

Related to #6550


dbolduc (changes requested):
I think we should mark the POSTs as idempotent and keep using `Aip194Strict` instead of `AlwaysRetry`.

dbolduc (approved):

PR #6571, opened 2026-08-28
feat(storage): implement 2 MiB coalescing buffer for appendable upload

vsharonlynn:


src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -31,23 +32,36 @@ use gaxi::prost::FromProto;
 use tokio::sync::mpsc::Sender;
 use tokio::sync::oneshot;
 
+/// Defines the maximum number of queued intents in the foreground-to-worker channel.
+///
+/// Each write intent carries a coalesced chunk of up to [`MAX_WRITE_CHUNK_SIZE`] (2 MiB).
+/// Sizing this queue to 4 slots (8 MiB total in-flight payload) balances two goals:
+/// 1. Pipelining: Keeps the background worker stream continuously saturated without network starvation.
+/// 2. Backpressure and Memory Footprint: Bounds queued channel memory to 8 MiB so that combined
+///    with the 32 MiB worker replay buffer and 2 MiB coalescing buffer, total in-flight data
```
joshuatants: this doesn't exist yet?

src/storage/src/storage/bidi_write/transport.rs
```diff
@@ -31,23 +32,36 @@ use gaxi::prost::FromProto;
...
+///    remains predictably capped around 42 MiB before foreground `.append()` calls suspend.
```
joshuatants: 32 + 2 * 8 = 40, how is thsi 42?

src/storage/src/storage/bidi_write/coalescing_buffer.rs
```diff
@@ -0,0 +1,212 @@
...
+            ready_chunks.push(chunk.slice(..COALESCING_CHUNK_SIZE));
+            chunk = chunk.slice(COALESCING_CHUNK_SIZE..);
+        }
+
+        // 3. Store remaining trailing bytes (< 2 MiB) in accumulator.
+        if !chunk.is_empty() {
+            self.buffer.extend_from_slice(&chunk);
+        }
+
+        ready_chunks
```
joshuatants, line 92: Since the bytes are moved to the caller, what happens if there's an error and the `append` operation fails? Seems like the bytes would be lost, and there's no way to recover them.

src/storage/src/storage/bidi_write/coalescing_buffer.rs
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    /// Drains any residual unsealed buffered bytes (< 2 MiB).
+    pub fn flush(&mut self) -> Option<Bytes> {
+        if self.buffer.is_empty() {
+            None
+        } else {
+            let residual = self.buffer.split().freeze();
+            self.buffer.reserve(COALESCING_CHUNK_SIZE);
+            Some(residual)
```
joshuatants: Ditto above - if this fails midway, the bytes are lost.

src/storage/src/storage/bidi_write/coalescing_buffer.rs
```diff
@@ -0,0 +1,212 @@
...
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl CoalescingBuffer {
+    /// Creates a new, empty coalescing buffer with pre-allocated 2 MiB capacity.
+    pub fn new() -> Self {
+        Self {
+            buffer: BytesMut::with_capacity(COALESCING_CHUNK_SIZE),
```
joshuatants: Would it make more sense to do lazy initialisation instead?

src/storage/src/storage/bidi_write/coalescing_buffer.rs
```diff
@@ -0,0 +1,212 @@
...
+        ready_chunks
+    }
+
+    /// Drains any residual unsealed buffered bytes (< 2 MiB).
+    pub fn flush(&mut self) -> Option<Bytes> {
+        if self.buffer.is_empty() {
+            None
+        } else {
+            let residual = self.buffer.split().freeze();
+            self.buffer.reserve(COALESCING_CHUNK_SIZE);
```
joshuatants: Similar to the above qn on lazy initialisation, would it mkae more sense to only reserve, if we receive more bytes that need to be written? For example, calling flush() during finalize() is pointless.

src/storage/src/storage/bidi_write/coalescing_buffer.rs
```diff
@@ -0,0 +1,212 @@
...
+        assert_eq!(ready.len(), 1);
+        assert_eq!(ready[0].len(), COALESCING_CHUNK_SIZE);
+        assert_eq!(&ready[0][..1024 * 1024], payload_1mib.as_ref());
+        assert_eq!(&ready[0][1024 * 1024..], payload_1mib.as_ref());
+        assert!(buf.is_empty());
+        assert_eq!(buf.len(), 0);
+    }
+
+    #[test]
+    fn large_append_fast_path_zero_copy() {
```
joshuatants, line 172: How is the zero-copy property tested here?

joshuatants (changes requested):

src/storage/src/storage/bidi_write/coalescing_buffer.rs
vsharonlynn: Good question. I added a code doc for the `CoalescingBuffer` struct to explain the error recovery semantics. PTAL. Thanks.

src/storage/src/storage/bidi_write/coalescing_buffer.rs
vsharonlynn: Ditto. I added a code doc for the CoalescingBuffer struct to explain the error recovery semantics. PTAL. Thanks.

src/storage/src/storage/bidi_write/coalescing_buffer.rs
vsharonlynn: Yes, that makes sense. Thank you.

src/storage/src/storage/bidi_write/coalescing_buffer.rs
vsharonlynn: Yes, removed. Thanks!

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: Removed. Sorry about that. I was working on the same branch in the beginning.

src/storage/src/storage/bidi_write/transport.rs
vsharonlynn: The math is
4 channels * 2 MiB each = 8
2 MiB coalescing buffer = 2
32 MiB replay buffer = 32
Total 32+2+8 = 42.

I removed the replay buffer since it's not in this PR.
For this PR the total is 10 MiB.

src/storage/src/storage/bidi_write/coalescing_buffer.rs
vsharonlynn: My bad. Added the assertion to compare chunk memory pointers to the slices of the source payload.

joshuatants (approved):

PR #6605, opened 2026-09-01
feat(storage): implement replay buffer for appendable upload

vsharonlynn:
Issue #5716 .

### DIFF
--- a/src/storage/src/storage/bidi_write/replay_buffer.rs
+++ b/src/storage/src/storage/bidi_write/replay_buffer.rs
@@ -1,0 +1,300 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+//! Retains unacknowledged chunks, trims acknowledged data, and provides chunks
+//! for resending upon stream reconnect.
+
+use crate::google::storage::v2::{
+    BidiWriteObjectRequest, ChecksummedData, bidi_write_object_request::Data,
+};
+use bytes::Bytes;
+use std::collections::VecDeque;
+
+/// Defines the maximum capacity of the replay buffer in bytes (32 MiB).
+pub const MAX_REPLAY_BUFFER_SIZE: usize = 32 * 1024 * 1024;
+
+/// Represents an unacknowledged data chunk retained in the [`ReplayBuffer`].
+#[derive(Clone, Debug, PartialEq, Eq)]
+pub struct ReplayChunk {
+    /// Holds the logical starting write offset of this chunk.
+    pub write_offset: i64,
+    /// Contains the raw payload bytes.
+    pub data: Bytes,
+    /// Stores the precomputed CRC32C checksum of [`data`][Self::data].
+    pub crc32c: u32,
+}
+
+impl ReplayChunk {
+    /// Creates a new replay chunk.
+    pub fn new(write_offset: i64, data: Bytes, crc32c: u32) -> Self {
+        Self {
+            write_offset,
+            data,
+            crc32c,
+        }
+    }
+
+    /// Returns the ending byte offset (exclusive) of this chunk.
+    pub fn end_offset(&self) -> i64 {
+        self.write_offset + self.data.len() as i64
+    }
+
+    /// Converts this replay chunk into a [`BidiWriteObjectRequest`] for transmission.
+    pub fn to_request(&self) -> BidiWriteObjectRequest {
+        BidiWriteObjectRequest {
+            write_offset: self.write_offset,
+            data: Some(Data::ChecksummedData(ChecksummedData {
+                content: self.data.clone(),
+                crc32c: Some(self.crc32c),
+            })),
+            ..BidiWriteObjectRequest::default()
+        }
+    }
+}
+
+/// Manages an in-memory FIFO queue of unacknowledged chunks up to [`MAX_REPLAY_BUFFER_SIZE`].
+#[derive(Debug, Default)]
+pub struct ReplayBuffer {
+    queue: VecDeque<ReplayChunk>,
+    current_size: usize,
+}
+
+impl ReplayBuffer {
+    /// Creates a new, empty replay buffer.
+    pub fn new() -> Self {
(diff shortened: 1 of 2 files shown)


src/storage/src/storage/bidi_write/replay_buffer.rs
```diff
@@ -0,0 +1,300 @@
...
+//! for resending upon stream reconnect.
+
+use crate::google::storage::v2::{
+    BidiWriteObjectRequest, ChecksummedData, bidi_write_object_request::Data,
+};
+use bytes::Bytes;
+use std::collections::VecDeque;
+
+/// Defines the maximum capacity of the replay buffer in bytes (32 MiB).
+pub const MAX_REPLAY_BUFFER_SIZE: usize = 32 * 1024 * 1024;
```
coryan, line 25: Is  32MiB the "One True Value" for some reason? Shouldn't this be configurable? If not, could you add a comment explaining why this is the only correct value?

src/storage/src/storage/bidi_write/replay_buffer.rs
```diff
@@ -0,0 +1,300 @@
...
+    pub fn push(&mut self, chunk: ReplayChunk) {
+        self.current_size += chunk.data.len();
+        self.queue.push_back(chunk);
+    }
+
+    /// Trims acknowledged chunks up to `persisted_size`.
+    ///
+    /// If `persisted_size` lands inside a chunk, that chunk is sliced in-place
+    /// and its CRC32C is recomputed for the unpersisted sub-slice only.
+    pub fn acknowledge(&mut self, persisted_size: i64) {
```
joshuatants, line 92: Nit: maybe use `ack`? Hardly ever see the full word in code

src/storage/src/storage/bidi_write/replay_buffer.rs
```diff
@@ -0,0 +1,300 @@
...
+        self.current_size >= MAX_REPLAY_BUFFER_SIZE
+    }
+
+    /// Returns `true` if the replay buffer contains no chunks.
+    pub fn is_empty(&self) -> bool {
+        self.queue.is_empty()
+    }
+
+    /// Returns the number of chunks currently held in the buffer.
+    pub fn len(&self) -> usize {
```
joshuatants, line 125: I think `num_chunks` is  less ambiguous

src/storage/src/storage/bidi_write/replay_buffer.rs
```diff
@@ -0,0 +1,300 @@
...
+        self.queue.len()
+    }
+
+    /// Returns the total unpersisted byte count currently retained in the buffer.
+    pub fn current_size(&self) -> usize {
```
joshuatants, line 130: `bytes_remaining`, `bytes_unpersisted`, `unpersisted_bytes`?

Since there's a distinction between the no. of chunks vs bytes, thinking that "size"/"length" can be ambiguous.

joshuatants (dismissed):

src/storage/src/storage/bidi_write/replay_buffer.rs
vsharonlynn: Done. Thanks.

src/storage/src/storage/bidi_write/replay_buffer.rs
vsharonlynn: Done. Thanks.

src/storage/src/storage/bidi_write/replay_buffer.rs
vsharonlynn: Agreen. Went with `unpersisted_bytes`. Thanks.

src/storage/src/storage/bidi_write/replay_buffer.rs
vsharonlynn: No, it's not the one true value. It is the intended default.

Added `with_capacity` method. Currently default value is 32MiB.
I'll write follow-up PRs on setting this at the request level (and maybe client-level) to keep this PR focused on replay_buffer.rs.

Thanks Carlos.

src/storage/src/storage/bidi_write/replay_buffer.rs
```diff
@@ -0,0 +1,363 @@
...
+            } else {
+                break;
+            }
+        }
+
+        if let Some(front) = self.queue.front_mut()
+            && front.write_offset < persisted_size
+        {
+            let trimmed_bytes = (persisted_size - front.write_offset) as usize;
+            // SAFETY: persisted_size is guaranteed to be within the bounds of the chunk because front.write_offset < persisted_size and front.end_offset() > persisted_size.
```
xlai20: Nit: In Rust convention, `// SAFETY:` comments are typically reserved for documenting contracts justifying `unsafe` blocks. Because `Bytes::slice` is completely safe Rust, this comment is documenting an algorithmic invariant rather than an unsafe memory-safety contract.

Consider updating the prefix to `// Invariant:` or `// Note:` to align with standard conventions:
```rust
// Invariant: persisted_size is guaranteed to be within the bounds of the chunk because front.write_offset < persisted_size and front.end_offset() > persisted_size.

File: src/storage/src/storage/bidi_write/replay_buffer.rs [hunk 6]
```diff
@@ -0,0 +1,363 @@
...
+    /// and its CRC32C is recomputed for the unpersisted sub-slice only.
+    pub fn ack(&mut self, persisted_size: i64) {
+        while let Some(front) = self.queue.front() {
+            if front.end_offset() <= persisted_size {
+                let chunk = self.queue.pop_front().expect("front chunk must exist");
+                self.unpersisted_bytes -= chunk.data.len();
+            } else {
+                break;
+            }
+        }
```
xlai20, line 119: Nit: Since the only reason we unwrap `chunk` here is to read its byte length (`chunk.data.len()`), we can read `front.data.len()` directly before popping:

```rust
while let Some(front) = self.queue.front() {
    if front.end_offset() <= persisted_size {
        self.unpersisted_bytes -= front.data.len();
        self.queue.pop_front();
    } else {
        break;
    }
}
```

Why I suggest this change:
1. "expect()" is unreachable because to enter the while loop, the queue must contain at least one item. This makes compiler to generate an unnecessary panic branch
3. Cleaner Rust idioms: the immutable borrow of front finishes the moment its length is evaluated. you save an unnecessary intermediate variable

xlai20 (changes requested):
I would say it's in general a ready to merge state except 2 nits.

src/storage/src/storage/bidi_write/replay_buffer.rs
vsharonlynn: Done. Thanks!

src/storage/src/storage/bidi_write/replay_buffer.rs
vsharonlynn: Ah, ok. Done. Thanks for explaining!

xlai20 (approved):

Issue #5315, opened 2026-04-07
Conversion for `wkt::DescriptorProto`


dbolduc:
Needs to be defined in `gaxi`, if that can even work

dbolduc:
I am starting to suspect this is impossible to fully generate via sidekick. I think it requires the protobuf source used by librarian to match the one used by `prost_types`.

For example, our `wkt::DescriptorProto` depends on the `wkt::Edition` enum. But there is no `prost_types::Edition`.

dbolduc:
We can always start with the `ArrowSchema` side of things.

And worst case, we write a conversion function by hand for the `ProtoSchema`.

coryan:
> We can always start with the `ArrowSchema` side of things.

Yes.

> 
> And worst case, we write a conversion function by hand for the `ProtoSchema`.

You could limit support to Proto3 protos (or whatever Prost supports), and add full support when we migrate to Google's Protobuf. Or ... you could accept the Proto descriptor as a binary blob, representing the serialized schema.

dbolduc:
We are not going to offer a protobuf based surface any time soon. Of the input formats, it is the least prevalent and hardest to use.

PR #6608, opened 2026-09-01
feat(wkt): trim `wkt::DescriptorProto` fields to allow AST translation

haphungw:
BigQuery `TableSchema` tracks structure (names, column formats `field`, nested structs `nested_type`, and enums `enum_type`), and ignores protobuf extensions, one-ofs, and reserved namespaces.

This PR trims recursive and interconnected definitions out of `google-cloud-wkt`, allowing us to map `wkt::DescriptorProto` directly to `prost_types::DescriptorProto` without hitting recursive dependency issues in `librarian`.

For #5315

### DIFF
--- a/librarian.yaml
+++ b/librarian.yaml
@@ -1816,6 +1816,13 @@ libraries:
             - descriptor.proto
           module_path: crate
           output: src/wkt/src/generated
+          skipped_ids:
+            - .google.protobuf.DescriptorProto.extension
+            - .google.protobuf.DescriptorProto.oneof_decl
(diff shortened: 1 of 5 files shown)


dbolduc (changes requested):
We can't do this. It would be a breaking change.

I was thinking of using `skipped_ids` in the convert prost template.

haphungw:
This approach is incorrect. Replaced by https://github.com/googleapis/google-cloud-rust/pull/6609 and https://github.com/googleapis/google-cloud-rust/pull/6613.

PR #6609, opened 2026-09-01
feat(gax-internal): add `FromProto` and `ToProto` for `wkt::DescriptorProto`

haphungw:
Provide `gaxi::prost::FromProto` and `ToProto`.

For #5315

### DIFF
--- a/src/gax-internal/Cargo.toml
+++ b/src/gax-internal/Cargo.toml
@@ -36,10 +36,6 @@ workspace = true
 
 [features]
 default = ["_default-rustls-provider"]
-# Enables the narrow, intentionally incomplete `wkt::DescriptorProto` conversion.
-#
-# Used exclusively by BigQuery to bypass librarian struct generation constraints.
-_internal-descriptor-proto-conversion = []
 # Enables the HTTP client.
 #
 # Currently this is based on reqwest, and this crate implicitly uses the same
--- a/src/gax-internal/Cargo.toml
+++ b/src/gax-internal/Cargo.toml
@@ -36,6 +36,10 @@ workspace = true
 
 [features]
 default = ["_default-rustls-provider"]
+# Enables the narrow, intentionally incomplete `wkt::DescriptorProto` conversion.
+#
+# Used exclusively by BigQuery to bypass librarian struct generation constraints.
+_internal-descriptor-proto-conversion = []
 # Enables the HTTP client.
 #
 # Currently this is based on reqwest, and this crate implicitly uses the same
--- a/src/gax-internal/src/prost.rs
+++ b/src/gax-internal/src/prost.rs
@@ -231,6 +231,89 @@ impl FromProto<wkt::NullValue> for prost_types::NullValue {
     }
 }
 
+impl FromProto<wkt::DescriptorProto> for prost_types::DescriptorProto {
+    fn cnv(self) -> Result<wkt::DescriptorProto> {
+        Ok(wkt::DescriptorProto::default()
+            .set_name(self.name.unwrap_or_default())
+            .set_field(
+                self.field
+                    .into_iter()
+                    .map(|v| {
+                        wkt::FieldDescriptorProto::default()
+                            .set_name(v.name.unwrap_or_default())
+                            .set_number(v.number.unwrap_or_default())
+                    })
+                    .collect::<Vec<_>>(),
+            )
+            .set_nested_type(
+                self.nested_type
+                    .into_iter()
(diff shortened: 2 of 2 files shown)


src/gax-internal/src/prost.rs
```diff
@@ -231,6 +231,89 @@ impl FromProto<wkt::NullValue> for prost_types::NullValue {
     }
 }
 
+impl FromProto<wkt::DescriptorProto> for prost_types::DescriptorProto {
```
suzmue, line 240: I believe there are several fields that are being skipped during conversion.

This seems acceptable as a narrow conversion to be used in Bigquery. Please add a clear comment noting that it is intentionally incomplete so future readers do not treat it as a general-purpose DescriptorProto.

If there’s any risk that other libraries could start depending on this, we could also consider gating it behind an internal feature like _internal-descriptor-proto-conversion to prevent accidental use.

suzmue (approved):
Would be nice if we could generate this.

src/gax-internal/src/prost.rs
haphungw: That's a good idea. I added a `_internal-descriptor-proto-conversion` flag to gate this conversion.

dbolduc: FWIW, BigQuery is currently the only service that uses a DescriptorProto, and I expect it to stay that way.

https://github.com/search?q=repo%3Agoogleapis%2Fgoogleapis+google.protobuf.descriptorproto&type=code

... and we are skipping fields because the `prost_types::DescriptorProto` we need to convert to does not have all of the fields of our `wkt::DescriptorProto` (it was generated off a different version of the protos). So even if any other library used this in their API and they required gRPC transport, they would have the same problem.

So I think the feature flag is unnecessary, and I think we should remove it.

---

+1 to "we should generate this". If you are closing #5315, can you open a new issue + TODO for generating the code?

src/gax-internal/Cargo.toml
```diff
@@ -36,6 +36,10 @@ workspace = true
 
 [features]
 default = ["_default-rustls-provider"]
+# Enables the narrow, intentionally incomplete `wkt::DescriptorProto` conversion.
+#
+# Used exclusively by BigQuery to bypass librarian struct generation constraints.
+_internal-descriptor-proto-conversion = []
```
dbolduc: We should drop the feature. Otherwise, LGTM.

src/gax-internal/src/prost.rs
```diff
@@ -231,6 +231,101 @@ impl FromProto<wkt::NullValue> for prost_types::NullValue {
     }
 }
 
+/// A narrow, intentionally incomplete conversion for `DescriptorProto`.
+///  
+/// This conversion drops several AST fields and must not be used as
+/// a general-purpose converter. It exists solely to bypass librarian
+/// limitations when generating `ProtoSchema` in BigQuery.
+#[cfg(feature = "_internal-descriptor-proto-conversion")]
```
dbolduc: Add a TODO to track generating this code.

dbolduc (changes requested):

src/gax-internal/src/prost.rs
haphungw: done!

src/gax-internal/src/prost.rs
haphungw: makes sense, done

dbolduc (approved):

src/gax-internal/src/prost.rs
```diff
@@ -308,6 +403,7 @@ mod tests {
     #[test_case(Err(err()), Ok(2))]
     #[test_case(Ok(1), Err(err()))]
     #[test_case(Err(err()), Err(err()))]
+
```
dbolduc: weird newline

src/gax-internal/src/prost.rs
```diff
@@ -523,4 +619,21 @@ mod tests {
         assert_eq!(url, want);
         Ok(())
     }
+    #[test]
```
dbolduc, line 636: weird missing newline

src/gax-internal/src/prost.rs
```diff
@@ -523,4 +619,21 @@ mod tests {
...
+    fn test_descriptor_proto_conversion() -> anyhow::Result<()> {
```
dbolduc, line 637: test more fields?

haphungw: done

PR #6610, opened 2026-09-01
test(storage): add error metadata preservation test for gRPC

joshuatants:
For #5991

### DIFF
--- a/src/storage/tests/grpc_error_metadata.rs
+++ b/src/storage/tests/grpc_error_metadata.rs
@@ -49,8 +49,7 @@ async fn open_object_error_exposes_grpc_metadata_in_headers() -> Result<()> {
     metadata.insert(
         CUSTOM_METADATA_KEY,
         CUSTOM_METADATA_VALUE
-            .parse()
-            .expect("valid custom metadata value"),
+            .parse()?,
     );
 
     let status = Status::with_details_and_metadata(
--- a/src/storage/tests/grpc_error_metadata.rs
+++ b/src/storage/tests/grpc_error_metadata.rs
@@ -46,11 +46,7 @@ async fn open_object_error_exposes_grpc_metadata_in_headers() -> Result<()> {
     };
 
     let mut metadata = MetadataMap::new();
-    metadata.insert(
-        CUSTOM_METADATA_KEY,
-        CUSTOM_METADATA_VALUE
-            .parse()?,
-    );
+    metadata.insert(CUSTOM_METADATA_KEY, CUSTOM_METADATA_VALUE.parse()?);
 
     let status = Status::with_details_and_metadata(
         Code::PermissionDenied,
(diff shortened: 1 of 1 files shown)


src/storage/tests/grpc_error_metadata.rs
```diff
@@ -0,0 +1,96 @@
...
+        details: vec![prost_types::Any::from_msg(&detail)?],
+    };
+
+    let mut metadata = MetadataMap::new();
+    metadata.insert(
+        CUSTOM_METADATA_KEY,
+        CUSTOM_METADATA_VALUE
+            .parse()
+            .expect("valid custom metadata value"),
+    );
```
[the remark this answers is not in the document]
joshuatants: Done.

src/storage/tests/grpc_error_metadata.rs
```diff
@@ -0,0 +1,91 @@
...
+#[tokio::test]
+async fn open_object_error_exposes_grpc_metadata_in_headers() -> Result<()> {
+    // Arrange
+    const BIND_ADDRESS: &str = "127.0.0.1:0";
+    const BUCKET: &str = "projects/_/buckets/error-metadata-bucket";
+    const OBJECT: &str = "error-metadata-object";
+    const CUSTOM_METADATA_KEY: &str = "x-parity-probe";
+    const CUSTOM_METADATA_VALUE: &str = "present";
+
+    // Construct gRPC error details
```
vsharonlynn: Since it's only for BidiReadObjectRedirectError, the comment should be "gRPC bidi read".

src/storage/tests/grpc_error_metadata.rs
```diff
@@ -0,0 +1,91 @@
...
+
+    // Act
+    let error = client
+        .open_object(BUCKET, OBJECT)
+        .send()
+        .await
+        .expect_err("mock must return a permission error");
+
+    // Assert
+    assert!(error.status().is_some());
```
vsharonlynn, line 78: Should we also assert the contents of the error?
```
let returned_status = error.status().expect(...);
assert_eq!(returned_status.code(), Code::PermissionDenied);
assert_eq!(returned_status.details(), wire_status.encode_to_vec().as_slice());
```

I understand that this test is for metadata, so if you want to not include this here that's ok too.

vsharonlynn (approved):
Approved with comments. THanks.

src/storage/tests/grpc_error_metadata.rs
joshuatants: Done.

src/storage/tests/grpc_error_metadata.rs
joshuatants: Will leave it out - the error code and details are already tested in other tests.

PR #6611, opened 2026-09-01
perf(storage): add write_object upload strategies benchmarking suite

xlai20:
## Summary

Adds an independent benchmarking suite crate `storage-benchmark-write-object` under `src/storage/benchmarks/write_object` to evaluate and compare different upload strategies and checksum validation mechanisms for `write_object`.

This benchmarking suite provides empirical performance and throughput data across various payload sizes to inform the decision on the default `write_object` behavior for seekable sources (e.g., local files).

## Scenarios Evaluated

1. **Option A (`Option_A_Unbuffered_Baseline`)**:
   - `.send_unbuffered()`
   - Single continuous HTTP stream; 0 application heap buffer.
   - Checksum computed on the fly and validated client-side upon upload completion.
2. **Option B (`Option_B_Unbuffered_2Pass`)**:
   - `.precompute_checksums().await?.send_unbuffered()`
   - **Pass 1:** Fast local disk scan to compute SIMD CRC32C checksum (populates page cache).
   - **Pass 2:** Single continuous HTTP PUT stream with `x-goog-hash` server-side validation header.
   - 0 application heap buffer.
3. **Option C (`Option_C_Buffered_Chunked`)**:
   - `.send_buffered()`
   - 1-pass chunked upload using sequential 8 MiB HTTP PUT requests.
   - Server-side validation enforced by attaching checksum to the final chunk.
   - Requires 8 MiB in-memory buffer per upload.

## Benchmark Features

- **5 Object Size Tiers**: 12 MiB (single-shot multipart), 64 MiB (small resumable / 8 chunks), 512 MiB (medium resumable / 64 chunks), 2 GiB (large resumable / 256 chunks), and 8 GiB (stress resumable / 1,024 chunks).
- **Pre-flight Warmup (512 KiB)**: Executes a one-time global warmup per process to verify authentication, check bucket permissions, and prime the TLS 1.3 connection pool before running measured iterations.
- **Physical Disk & Page Cache Control**
- **Reporting & Orchestration**:
  - Real-time terminal summary tables (Mean, P50, P90, P99 latency, and throughput in MiB/s).
  - Optional raw CSV latency and JSON summary persistence via `GOOGLE_CLOUD_RUST_BENCHMARKS_STATS_OUTPUT_PATH`.
  - `run_all.sh` driver script to run all 5 tiers sequentially.

## Testing

- Unit tests added for CLI parsing and statistical metric calculations.
- Verified workspace compilation, `cargo clippy --all-targets -- -D warnings`, and `cargo fmt`.
- Executed full 5-tier benchmark runs on physical SSD storage against live GCS multi-region buckets.

### DIFF
--- a/src/storage/benchmarks/write_object/src/args.rs
+++ b/src/storage/benchmarks/write_object/src/args.rs
@@ -1,0 +1,60 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use clap::{Parser, ValueEnum};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq, ValueEnum)]
+pub enum UploadScenario {
+    /// Option A: Baseline 1-pass unbuffered stream (no precomputed hash, client-side validation only)
+    OptionA,
+    /// Option B: 2-pass unbuffered stream (Pass 1: precomputed hash, Pass 2: continuous stream, server-side validation)
+    OptionB,
+    /// Option C: 1-pass buffered chunked upload (8 MiB chunks in RAM, server-side validation on final chunk)
+    OptionC,
+    /// Run all three scenarios (Option A, Option B, and Option C) for side-by-side comparison
+    All,
+}
+
+#[derive(Parser, Debug)]
+#[command(author, version, about = "Benchmark write_object upload strategies in GCS", long_about = None)]
+pub struct Args {
+    /// The name of the bucket to use for the benchmark.
+    #[arg(long, env = "GOOGLE_CLOUD_RUST_BENCHMARKS_BUCKET")]
+    pub bucket_name: String,
+
+    /// Number of warmup iterations.
+    #[arg(long, default_value_t = 1)]
+    pub warmup_iterations: usize,
+
+    /// Number of measured iterations.
+    #[arg(long, default_value_t = 5)]
+    pub measured_iterations: usize,
+
+    /// The size of the object to upload in bytes.
+    #[arg(long, default_value_t = 67_108_864)] // 64 MiB default
+    pub object_size: usize,
+
+    /// Upload scenario / strategy to benchmark.
+    #[arg(long, value_enum, default_value_t = UploadScenario::All)]
+    pub scenario: UploadScenario,
+
+    /// Directory for output artifacts (raw CSV latencies and summary JSON). If not provided, file
+    /// reporting is skipped.
+    #[arg(long)]
+    pub output_dir: Option<String>,
+
+    /// Whether to delete test objects from GCS after each iteration.
+    #[arg(long, default_value_t = true)]
+    pub cleanup: bool,
+}
--- a/src/storage/benchmarks/write_object/src/args.rs
+++ b/src/storage/benchmarks/write_object/src/args.rs
@@ -31,7 +31,7 @@ pub enum UploadScenario {
 pub struct Args {
     /// The name of the bucket to use for the benchmark.
     #[arg(long, env = "GOOGLE_CLOUD_RUST_BENCHMARKS_BUCKET")]
-    pub bucket_name: String,
+    pub bucket_name: Option<String>,
 
     /// Number of measured iterations per scenario.
     #[arg(long, default_value_t = 5)]
@@ -51,18 +51,13 @@ pub struct Args {
     pub cold_cache: bool,
 
     /// Directory on physical SSD storage for creating the temporary test file.
-    #[arg(
-        long,
-        default_value = "/usr/local/google/tmp/rust-write-object-benchmarking-data"
-    )]
-    pub temp_dir: String,
+    #[arg(long, env = "GOOGLE_CLOUD_RUST_BENCHMARKS_DATA_PATH")]
+    pub temp_dir: Option<String>,
 
     /// Directory for saving output artifacts (raw CSV latencies and summary JSON).
-    #[arg(
-        long,
-        default_value = "/usr/local/google/tmp/rust-write-object-benchmarking-data/results"
-    )]
-    pub output_dir: String,
+    /// If omitted or empty, file reporting is skipped and results are printed to the terminal.
+    #[arg(long, env = "GOOGLE_CLOUD_RUST_BENCHMARKS_STATS_OUTPUT_PATH")]
+    pub output_dir: Option<String>,
 
     /// Whether to delete test objects from GCS after each iteration.
     #[arg(long, action = clap::ArgAction::Set, default_value_t = true)]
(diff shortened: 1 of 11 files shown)


xlai20:
This is part of the Issue #6461 , a benchmarking effort that is helping the final API definition decision.

For an 8GB file sit on disk, 
- Option A (Baseline): 1 continuous PUT stream without upfront hash (client-side validation only).
- Option B (2-Pass Stream): Precompute local hash + 1 continuous PUT stream.
  - Result: +30% P50 latency (-16% throughput) vs Option A.
- Option C (1-Pass Chunked): Sequential 8 MiB in-memory chunked PUTs.
  - Result: +75% P50 latency (-41% throughput) vs Option A.
 
The executive decision is to turn on Option B to prioritize data integrity over speed; Option B incurs less performance penalty than Option C.

src/storage/benchmarks/write_object/src/scenarios.rs
```diff
@@ -0,0 +1,147 @@
...
+    bucket_name: &str,
+    object_name: &str,
+    file_path: &Path,
+    object_size: usize,
+) -> anyhow::Result<IterationResult> {
+    let file = File::open(file_path).await?;
+    let start_time = Instant::now();
+
+    let object = client
+        .write_object(bucket_name, object_name, file)
```
vsharonlynn, line 43: After your other PR #6769 , the default will be to automatically do precomputation, right?
If yes, then we'll need to set `.with_checksum_precomputation(false)` explicitly here.

src/storage/benchmarks/write_object/src/scenarios.rs
```diff
@@ -0,0 +1,147 @@
...
+    file_path: &Path,
+    object_size: usize,
+) -> anyhow::Result<IterationResult> {
+    let file = File::open(file_path).await?;
+    let total_start = Instant::now();
+
+    let precompute_start = Instant::now();
+    let write_builder = client
+        .write_object(bucket_name, object_name, file)
+        .precompute_checksums()
```
vsharonlynn, line 79: Nit: Maybe add a note that this is deprecated in #6769. I think it's ok to keep using this here because you did the benchmark before this was deprecated.

src/storage/benchmarks/write_object/src/args.rs
```diff
@@ -0,0 +1,100 @@
...
+    #[arg(long, env = "GOOGLE_CLOUD_RUST_BENCHMARKS_BUCKET")]
+    pub bucket_name: Option<String>,
+
+    /// Number of measured iterations per scenario.
+    #[arg(long, default_value_t = 5)]
+    pub measured_iterations: usize,
+
+    /// The size of the object to upload in bytes.
+    #[arg(long, default_value_t = 67_108_864)] // 64 MiB default
+    pub object_size: usize,
```
vsharonlynn, line 42: I think we need u64 here.
In run_all.sh, the Tier 5 sets `8589934592` (8 GiB) https://github.com/googleapis/google-cloud-rust/pull/6611/changes#diff-975f37e0a9f9b2e60be2aabcf50bd7f88e34925f118a0430dbd29c02e6fdab21R56.
This will overflow `usize` on 32-bit platforms.

src/storage/benchmarks/write_object/src/source.rs
```diff
@@ -0,0 +1,123 @@
...
+        remaining -= to_write;
+    }
+    async_file.flush().await?;
+
+    Ok((temp_file, path))
+}
+
+/// Evicts the given file's data from the OS page cache (RAM) to simulate a cold physical disk read.
+pub fn drop_file_from_page_cache(path: &Path) -> std::io::Result<()> {
+    #[cfg(unix)]
```
vsharonlynn, line 108: We might need to change the guard to target linux only, `#[cfg(target_os = "linux")]`.

`libc::posix_fadvise` is gated with `#[cfg(unix)]`. macOS is Unix, but Apple Darwin libc does not implement `posix_fadvise`, so this will cause compilation on macOS to fail.

src/storage/benchmarks/write_object/src/source.rs
```diff
@@ -0,0 +1,123 @@
...
+    {
+        let std_file = File::open(path)?;
+        let fd = std_file.as_raw_fd();
+        // Sync dirty pages to disk first.
+        // SAFETY: `fdatasync` is called with a valid open file descriptor owned by `std_file`.
+        let _ = unsafe { libc::fdatasync(fd) };
```
vsharonlynn, line 114: Nit: This is more of a defensive programming... If fdatasync fails, posix_fadvise(..., POSIX_FADV_DONTNEED) cannot discard dirty pages (PageDirty in the Linux kernel).
So maybe better to not ignore the error.
```
let ret = unsafe { libc::fdatasync(fd) };
if ret != 0 {
    return Err(std::io::Error::last_os_error());
}
```

src/storage/benchmarks/write_object/src/source.rs
```diff
@@ -0,0 +1,123 @@
...
+    size_bytes: usize,
+    temp_dir: &str,
+) -> anyhow::Result<(NamedTempFile, PathBuf)> {
+    // Ensure parent directory exists
+    tokio::fs::create_dir_all(temp_dir).await?;
+
+    let temp_file = NamedTempFile::new_in(temp_dir)?;
+    let path = temp_file.path().to_path_buf();
+
+    // Use a 1 MiB chunk of pseudo-random data written repeatedly to disk
```
vsharonlynn, line 88: Nit: THis is generating a 1MiB block and writing it repeatedly. This might produce data that is highly compressible on filesystems which might distort I/O measurements. Pointing this out because in the PR description it's explicitly mentioned "Generates **non-compressible** pseudo-random files on physical storage".
Maybe just mention that 1MiB pattern is used to minimize setup complexity?

vsharonlynn (changes requested):

src/storage/benchmarks/write_object/src/scenarios.rs
xlai20: I think it's better to preserve the logical order of PRs here. It is because of the benchmark outcome in this PR, then we have the API change in the other #6769 . Hence, I would like to commit the benchmark first, have the evidence of the need to change API, then proceed to commit the other PR.

As a result, we must use the older API here in the benchmark.

src/storage/benchmarks/write_object/src/scenarios.rs
xlai20: Ditto. 

Deprecation of this API will only follow when #6769 is merged.

src/storage/benchmarks/write_object/src/source.rs
xlai20: Done. remove the outdated statement in PR description.

src/storage/benchmarks/write_object/src/source.rs
xlai20: Done.

src/storage/benchmarks/write_object/src/source.rs
xlai20: Done.

src/storage/benchmarks/write_object/src/args.rs
xlai20: Done. Thanks for pointing it out!

src/storage/benchmarks/write_object/src/scenarios.rs
vsharonlynn: SGTM. Thanks!
Nit: We can have a TODO here to track.

src/storage/benchmarks/write_object/src/scenarios.rs
vsharonlynn: Ok. Thanks.

vsharonlynn (approved):
Approved. With a nit. Thank you!

PR #6634, opened 2026-09-02
impl(pubsub): enable streaming method generation

suzmue:
The generated clients are internal so these changes are all implementation details. The only change is to generate streaming_pull. It was included so we could use the types needed to do it. This does *not* update the veneer to use the generated method.


dbolduc (approved):
I assume the point is to stop using the `include_*_streaming_methods` configuration options.

We should probably do some benchmarking before swapping out the implementations

suzmue:
> I assume the point is to stop using the `include_*_streaming_methods` configuration options.
> 
> We should probably do some benchmarking before swapping out the implementations

Correct. Switching to using the generated code is not a requirement.

PR #6638, opened 2026-09-02
fix(gax): parse status detail errors from non AIP-193 complaint APIs

alvarowolfx:
Some older Google Cloud APIs are not AIP-193 complain and return errors with an `error.errors` array containing the details. Issue was found when writing logic to retry BigQuery errors.

Some examples:
* BigQuery REST API error messages - https://cloud.google.com/bigquery/docs/error-messages
* Cloud Storage JSON API status codes - https://cloud.google.com/storage/docs/json_api/v1/status-codes

Since only `domain` and` reason` fields are common to `rpc::ErrorInfo`, the rest is going to be set as `metadata` on it.

Towards #6218


dbolduc (approved):
Nice

PR #6641, opened 2026-09-02
impl(bigquery): retry on RPC errors with structured reason

alvarowolfx:
Towards #6218


alvarowolfx:
Found this issue when first call to `jobs.query` threw a 400 error that is retryable:

```
{
  "error": {
    "code": 400,
    "message": "The job encountered an error during execution. Retrying the job may solve the problem.",
    "errors": [
      {
        "message": "The job encountered an error during execution. Retrying the job may solve the problem.",
        "domain": "global",
        "reason": "backendError"
      }
    ],
    "status": "INVALID_ARGUMENT"
  }
}
```

This depends on a fix on gax at #6638

dbolduc (approved):

PR #6643, opened 2026-09-02
ci: use e2-highmem-16 for large worker pool

suzmue:
Update the private worker pool `rust-sdk-pool-large` to use `e2-highmem-16` instances instead of `e2-standard-32`. This maintains the 128 GB memory capacity while naturally scaling down Cargo parallelism, providing 8 GB RAM per core to avoid Out of Memory (OOM) failures on large crates like `google-cloud-compute-v1`.


coryan (approved):

suzmue:
Considering this PR is failing, I don't think this changes solved our issues...

PR #6647, opened 2026-09-02
fix(bigquery): parse INTERVAL values with 24 or more hours

fornwall:
BigQuery renders an `INTERVAL` as `[-]Y-M [-]D [-]H:M:S[.F]` and never carries hours over into days.

The clock part was parsed as a `time::Time`, which rejects hours above 23, so conversion failed for values such as:

- `TIMESTAMP '2026-01-03' - TIMESTAMP '2026-01-01'`, rendered as `0-0 0 48:0:0`
- `INTERVAL 25 HOUR`, rendered as `0-0 0 25:0:0`

Fix that by parsing the clock part directly, allowing hours larger than 23.

### DIFF
--- a/src/bigquery/src/datatypes.rs
+++ b/src/bigquery/src/datatypes.rs
@@ -19,7 +19,6 @@
 
 use crate::error::ConvertError;
 use crate::query::FromSql;
-use crate::query::from_sql::parse_time;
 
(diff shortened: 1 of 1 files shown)


dbolduc (approved):
Thanks!

alvarowolfx (approved):

PR #6654, opened 2026-09-03
impl(bigquery-read): migrate to default crate template with default_transport

suzmue:
Migrate google-cloud-bigquery-read from multi-module handwritten veneer to the default crate template with default_transport: grpc. Update integration tests to use BigQueryRead client.

Notable difference is public from_ methods for enum conversions are missing.


librarian.yaml
```diff
@@ -579,31 +579,16 @@ libraries:
     version: 1.12.0
     copyright_year: "2025"
   - name: google-cloud-bigquery-read
+    version: 0.1.0
+    apis:
+      - path: google/cloud/bigquery/storage/v1
     copyright_year: "2026"
     output: src/bigquery-read
-    skip_release: true
```
dbolduc, line 584: drive by so I don't forget: can we keep `publish: false`, at least in this PR?

suzmue: yes absolutely!

dbolduc:
I deleted this crate. close the PR?

PR #6659, opened 2026-09-04
test(storage): verify appendable object chunking and checksums

joshuatants:
For #5991


src/storage/tests/appendable_object_grpc.rs
```diff
@@ -0,0 +1,199 @@
...
+            WriteEvent::Chunk {
+                write_offset: (2 * MIB) as i64,
+                size: 2 * MIB,
+                crc32c: Some(crc32c::crc32c(&vec![0x42; 2 * MIB])),
+            },
+            WriteEvent::Chunk {
+                write_offset: (4 * MIB) as i64,
+                size: MIB,
+                crc32c: Some(crc32c::crc32c(&vec![0x42; MIB])),
+            },
```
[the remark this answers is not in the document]
joshuatants: Done.

src/storage/tests/appendable_object_grpc.rs
```diff
@@ -0,0 +1,199 @@
...
+        [
+            WriteEvent::Chunk {
+                write_offset: 0,
+                size: 2 * MIB,
+                crc32c: Some(crc32c::crc32c(&payload[0..2 * MIB])),
+            },
+            WriteEvent::Chunk {
+                write_offset: (2 * MIB) as i64,
+                size: 2 * MIB,
+                crc32c: Some(crc32c::crc32c(&payload[0..2 * MIB])),
```
vsharonlynn: Shouldn't this be the second chunk, therefore `2 * MiB..4 * MiB` ?

src/storage/tests/appendable_object_grpc.rs
```diff
@@ -0,0 +1,199 @@
...
+            },
+            WriteEvent::Chunk {
+                write_offset: (4 * MIB) as i64,
+                size: MIB,
+                crc32c: Some(crc32c::crc32c(&payload[0..MIB])),
```
vsharonlynn: Third chunk is 4 to 5 ?

src/storage/tests/appendable_object_grpc.rs
vsharonlynn: I think it only passes because the payload is 0x42...

vsharonlynn (changes requested):

PR #6664, opened 2026-09-04
impl(bigquery): evict streams from stream pool

dbolduc:
Add the ability for writers to evict a failed stream from the pool, and get a new stream. This is how they react to failures on their cached stream.

Multiple writers can be serviced by the same gRPC stream. If the stream dies, they will race to evict and replace. One will win. The rest will be moved to other streams in the pool (likely the newly created stream, which is at 0 load)

Towards #5831

### DIFF
--- a/src/bigquery/src/write/pool.rs
+++ b/src/bigquery/src/write/pool.rs
@@ -72,7 +72,20 @@ impl StreamPool {
         self.get_impl(&mut streams)
     }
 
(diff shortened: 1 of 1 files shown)


dbolduc:
> I have no feedback to provide as the implementation is clean and well-tested.

Highlight of my life.

haphungw (approved):

Issue #2138, opened 2025-05-14
Fix veneer API client header


dbolduc:
Veneers should use "veneer" instead of "gapic" in their client header.

Services are individually marked as veneers (`HasVeneer`) in the template. We will probably need to iterate over them, and scope the `info` mod in `transport.rs` (for gRPC).

dbolduc:
Given that the surface for Storage Control is fully generated, I think it is appropriate to call it a `gapic`.

We still want to do this work for other clients, e.g. `firestore`.

PR #6665, opened 2026-09-04
feat(gax): allow Google API client header to be overrided

alvarowolfx:
This will allow veneers to change header from GAPIC to veneer (gccl).

Towards https://github.com/googleapis/google-cloud-rust/issues/2138


alvarowolfx:
~need to be done in sync with https://github.com/googleapis/librarian/pull/7514~ removed template changes

src/gax-internal/src/http.rs
```diff
@@ -270,10 +272,15 @@ impl ReqwestClient {
         mut builder: reqwest::RequestBuilder,
         body: Option<I>,
         options: RequestOptions,
+        api_client_header: &'static str,
```
coryan: This is a breaking change, no?

alvarowolfx: yeah, I'm not sure to be honest and decided to send to CI if this would be automatically detected as a breaking change. My theory is that we don't expose internal details (reqwest), so I thought that changing ReqwestClient methods would not cause a break, but there is a high change of me being wrong.

I had an impl that was not changing the `pub execute` method, but this part here https://github.com/googleapis/google-cloud-rust/pull/6665/changes/a652b593c28fa99ea32cdea68050a03a17abec6c#diff-75de0fe3a13c8c814cedaa31ced499d3a24c233149f8386afec07af02a596f89L457 and mutating the request if an override was found for the x-goog-api-client header. But this was uglier to me and not in line on how things are being handled in gRPC. So I decided to try this out 🤷

dbolduc: Can we build it into something like `RequestOptions`, which can grow? It could be an extension.

coryan: > My theory is that we don't expose internal details (reqwest), so I thought that changing ReqwestClient methods would not cause a break, 

Consider a released version of `google-cloud-secretmanager-v1`, or any of our crates that use this method.... If you release a (minor) version of `gaxi` with a new version of `execute()` that requires more parameters then that crate would break.

@dbolduc's idea may work.

alvarowolfx: > Can we build it into something like `RequestOptions`, which can grow? It could be an extension.

yeah, I had some impl supporting overriding on a per request basis, but decided to go down the route of doing just on the client config level. But you're right, maybe the method can set x-goog-api-client on RequestOptions and we check against client level one. With the correct precedence, things will work without breaking changes.

> Consider a released version of google-cloud-secretmanager-v1, or any of our crates that use this method.... If you release a (minor) version of gaxi with a new version of execute() that requires more parameters then that crate would break.

true, at least a major in gaxi would be needed anyway. Will work on avoiding that. I already had a uglier solution for that, but will try now this idea on RequestOptions.

dbolduc (approved):

Issue #6122, opened 2026-07-21
Add flow control to connection pool


dbolduc:

dbolduc:
We are going to follow suit with the Pub/Sub Publisher and have the application own flow control. In Rust it is simple enough to hold a semaphore to track your outstanding writes.

Note that the client supports dynamic scale up of the connection pool to handle load.

PR #6741, opened 2026-09-08
impl(bigquery): track load on stream entry

dbolduc:
Towards #6122


src/bigquery/src/write/entry.rs
```diff
@@ -53,12 +55,40 @@ impl StreamEntry {
     }
 }
 
+/// RAII guard that increments load metrics on entry and decrements on drop.
```
alvarowolfx, line 58: TIL RAII

alvarowolfx (approved):

PR #6743, opened 2026-09-08
impl(bigquery): introduce stream dispatcher

dbolduc:
Add a type that handles interactions with the stream pool. Eventually it will handle retries for individual writes.

It caches the `StreamEntry` (basically a handle on a gRPC stream) so that pool resizing is independent of writing on valid streams.

Eventually, we can even rebalance writers to other streams by swapping out the Dispatcher's `ArcSwap<StreamEntry>` also without slowing down the hot path.

Towards #5831

### DIFF
--- a/src/bigquery/src/write/dispatcher.rs
+++ b/src/bigquery/src/write/dispatcher.rs
@@ -1,0 +1,262 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use super::append_response::{AppendResponse, to_result};
+use super::entry::StreamEntry;
+use super::error::{AppendError, AppendResult};
+use super::pool::StreamPool;
+use crate::Error;
+use crate::model::AppendRowsRequest;
+use arc_swap::ArcSwap;
+use gaxi::prost::{FromProto, ToProto};
+use std::sync::Arc;
+
+/// Efficiently dispatches writes to a stream in a stream pool.
+///
+/// This struct caches a `StreamEntry` and loads it atomically on each write.
+///
+/// On a transient error, the `Dispatcher` notifies the `StreamPool` of the
+/// failed `StreamEntry` and receives a new `StreamEntry` to use for future
+/// writes.
+///
+/// This struct is also responsible for retrying individual writes.
+#[derive(Debug)]
+pub(crate) struct Dispatcher {
+    pub(crate) pool: Arc<StreamPool>,
+    pub(crate) entry: ArcSwap<StreamEntry>,
+}
+
+impl Dispatcher {
+    /// Creates a new `Dispatcher` for a given `StreamPool`.
+    pub(crate) fn new(pool: Arc<StreamPool>) -> Self {
+        let stream = pool.get();
+        Self {
+            pool,
+            entry: ArcSwap::from_pointee(stream),
+        }
+    }
+
+    /// Send the write and process the response.
+    ///
+    /// Evicts and updates its cached stream on transient errors.
+    pub(crate) async fn send(&self, req: AppendRowsRequest) -> AppendResult<AppendResponse> {
+        let req = req.to_proto().map_err(Error::deser)?;
+
+        let stream = self.entry.load_full();
+        let stream_id = stream.id;
+
+        let resp = match stream.send(req).await {
+            Ok(resp) => Ok(resp),
+            Err(err) => {
+                if is_transient_error(&err) {
+                    // Atomically evicts failed_id and returns a new stream for use.
+                    let new_stream = self.pool.evict_and_replace(stream_id);
+
+                    // The application can `send()` multiple writes
+                    // concurrently. Only one `send()` should update the cached
+                    // stream on a transient error.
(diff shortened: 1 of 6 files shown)


src/bigquery/src/write/dispatcher.rs
```diff
@@ -0,0 +1,262 @@
...
+            Ok(resp) => Ok(resp),
+            Err(err) => {
+                if is_transient_error(&err) {
+                    // Atomically evicts failed_id and returns a new stream for use.
+                    let new_stream = self.pool.evict_and_replace(stream_id);
+
+                    // The application can `send()` multiple writes
+                    // concurrently. Only one `send()` should update the cached
+                    // stream on a transient error.
+                    let _ = self.entry.compare_and_swap(&stream, Arc::new(new_stream));
```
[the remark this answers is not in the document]
dbolduc: Hallucination. https://docs.rs/arc-swap/latest/arc_swap/type.ArcSwap.html#method.compare_and_swap

`compare_and_swap` was deprecated for the std atomic types.

src/bigquery/src/write/dispatcher.rs
```diff
@@ -0,0 +1,262 @@
...
+    }
+}
+
+pub(crate) fn is_transient_error(err: &AppendError) -> bool {
+    match err {
+        AppendError::UnexpectedEndOfStream => true,
+        // TODO(#6355): classify transient RPC errors
+        _ => false,
+    }
+}
```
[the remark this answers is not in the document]
dbolduc: I am aware and tracking this in https://github.com/googleapis/google-cloud-rust/issues/5831#issuecomment-5544851764

> by avoiding unnecessary failed requests.

these would be retried immediately by the client.

src/bigquery/src/write/dispatcher.rs
```diff
@@ -0,0 +1,269 @@
...
+                    let _ = self.entry.compare_and_swap(&stream, Arc::new(new_stream));
+
+                    // TODO(#6355): implement retries
+                }
+                Err(err)
+            }
+        }?;
+
+        let resp = resp.cnv().map_err(Error::ser)?;
+        to_result(resp)
```
suzmue: style nit, chain these?:

```suggestion
        resp.cnv()
            .map_err(Error::ser)
            .and_then(to_result)
```

src/bigquery/src/write/dispatcher.rs
```diff
@@ -0,0 +1,269 @@
...
+        let resp = match stream.send(req).await {
+            Ok(resp) => Ok(resp),
+            Err(err) => {
+                if is_transient_error(&err) {
+                    // Atomically evicts failed_id and returns a new stream for use.
+                    let new_stream = self.pool.evict_and_replace(stream_id);
+
+                    // The application can `send()` multiple writes
+                    // concurrently. Only one `send()` should update the cached
+                    // stream on a transient error.
```
suzmue, line 68: nit: wording, "should" makes it more confusing to me.

```suggestion
// The application can `send()` multiple writes
// concurrently. Only one `send()` will update the cached
// stream on a transient error.
```

or 

```
// The application can `send()` multiple writes concurrently.
// On transient errors, only one `send()` will swap the cached stream:
// `compare_and_swap` succeeds only if `self.entry` still points to `stream`.
```

src/bigquery/src/write/dispatcher.rs
```diff
@@ -0,0 +1,269 @@
...
+        // Simulate the stream closing before responding to the requests.
+        drop(response_tx);
+
+        while let Some(write) = writes.join_next().await {
+            let err = write?.expect_err("should return an error");
+            assert!(matches!(err, AppendError::UnexpectedEndOfStream));
+        }
+
+        // We ran into a transient error. We should now have a new stream. Only
+        // one of the dispatchers should have evicted the failed stream.
```
suzmue: "only one" is not tested here, consider making that more clear.

One idea would be to check the result of the compare_and_swap to see whether a swap actually occurred, behind #[cfg(test)] perhaps? we could keep a counter on the dispatcher, also gated, for how many swaps actually occur.

I think its ok to not do this as well because the compare_and_swap is well documented and used correctly.

src/bigquery/src/write/dispatcher.rs
```diff
@@ -0,0 +1,269 @@
...
+            pool,
+            entry: ArcSwap::from_pointee(stream),
+        }
+    }
+
+    /// Send the write and process the response.
+    ///
+    /// Evicts and updates its cached stream on transient errors.
+    pub(crate) async fn send(&self, req: AppendRowsRequest) -> AppendResult<AppendResponse> {
+        let req = req.to_proto().map_err(Error::deser)?;
```
suzmue:
```suggestion
        let req = req.to_proto().map_err(Error::ser)?;
```

suzmue (approved):

src/bigquery/src/write/dispatcher.rs
dbolduc: Doesn't work, the error types are different.

The error type on L77 is a `gax::Error` and the `?` is converting it into an `AppendError`.

src/bigquery/src/write/dispatcher.rs
dbolduc: Fixed

src/bigquery/src/write/dispatcher.rs
dbolduc: Yeah true. It's hard to reproduce.

We could have N writes outstanding on a stream that fails with a transient error. They all `evict_and_replace`. The first one creates a new stream in the pool. The pool *could* return a different stream for the rest of the callers. But as implemented... the new stream has 0 load, so it gets picked. And there is no good way to modify the load in between calls.

I think I will not test it because it is not an observable behavior that is too important.

---

Also, I `s/dispatchers/callers/` because there is only one dispatcher in the test.

src/bigquery/src/write/dispatcher.rs
dbolduc: Fixed. I think I had these backwards but only like 70% sure.

Issue #6598, opened 2026-09-01
Internal `ProtoWriter` API


haphungw:
Implement `pub(crate)` types (`ProtoWriter`, `ProtoRows`) based on the `Arrow` interface semantics.

PR #6744, opened 2026-09-08
impl(bigquery): add internal proto default writer

haphungw:
Introduce the internal `write::proto` module with `DefaultWriter` and `WriterBuilder`, mirroring the `arrow` default stream implementation https://github.com/googleapis/google-cloud-rust/pull/6241.

For #6598

### DIFF
--- a/src/bigquery/src/write/proto/writer_builder.rs
+++ b/src/bigquery/src/write/proto/writer_builder.rs
@@ -1,0 +1,107 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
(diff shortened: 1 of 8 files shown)


src/bigquery/src/write/proto/writer_builder.rs
```diff
@@ -0,0 +1,107 @@
...
+    pub fn default<T: Into<String>>(self, table: T) -> Result<DefaultWriter> {
+        let table = table.into();
+        validate_table(table.as_str())?;
+        let mut write_stream = table;
+        write_stream.push_str("/streams/_default");
+        Ok(DefaultWriter::new(self.inner, write_stream, self.schema))
+    }
+}
+
+fn validate_table(table: &str) -> Result<()> {
```
suzmue: Can we share these utils instead for the different writers?

suzmue (approved):

src/bigquery/src/write/proto/writer_builder.rs
haphungw: good idea! done

PR #6745, opened 2026-09-08
feat(storage): make GrpcRustClient available to Storage GAPIC methods

joshuatants:
Configure `grpc_client: crate::storage::bidi::GrpcClient` specifically for `google-cloud-storage` in librarian.yaml and regenerate the GAPIC transports.

This makes `GrpcRustClient` available exclusively to Storage's GAPIC transports under `--cfg google_cloud_unstable_grpc_rust`, while defaulting to `gaxi::grpc::Client` (Tonic) when the flag is absent.

For #5991.

### DIFF
--- a/src/storage/src/generated/gapic_control/transport.rs
+++ b/src/storage/src/generated/gapic_control/transport.rs
@@ -37,7 +37,7 @@ mod info {
 /// Implements [StorageControl](super::stub::StorageControl) using a gRPC client.
 #[derive(Clone)]
 pub struct StorageControl {
-    inner: gaxi::grpc::Client,
+    inner: crate::storage::bidi::GrpcClient,
 }
 
 impl std::fmt::Debug for StorageControl {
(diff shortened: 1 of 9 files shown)


vsharonlynn (approved):

src/storage/src/generated/gapic_control/transport.rs
```diff
@@ -51,14 +51,14 @@ impl std::fmt::Debug for StorageControl {
 impl StorageControl {
     pub async fn new(config: gaxi::options::ClientConfig) -> crate::ClientBuilderResult<Self> {
         let inner = if gaxi::options::tracing_enabled(&config) {
-            gaxi::grpc::Client::new_with_instrumentation(
+            crate::storage::bidi::GrpcClient::new_with_instrumentation(
```
xlai20: I have bit of concern that worth thinking before proceeding to with PR, Please let me know if such a concern makes sense:

**Current situation:**
In `librarian.yaml`, `grpc_client` is configured as `crate::storage::bidi::GrpcClient`. This causes the generated unary GAPIC transports (`StorageControl` in `gapic_control/transport.rs`) to import their transport client from `bidi.rs`.

**Concern:**
`bidi.rs` is specifically dedicated to bidirectional streaming operations (`BidiReadObject`, etc.). Having unary control-plane clients reference a type alias inside `bidi` creates a misleading conceptual coupling (unary code depending on a streaming submodule) and could be brittle if `bidi.rs` is ever refactored or moved.

**Actionable proposal:**
Consider defining or re-exporting `GrpcClient` in a neutral location, such as `src/storage/src/storage.rs`:
```rust
// in src/storage/src/storage.rs
pub(crate) use self::bidi::GrpcClient;
```
Then update `librarian.yaml` to:
```yaml
grpc_client: crate::storage::GrpcClient
```
This allows both unary GAPIC transports and `bidi.rs` to share a common, logically neutral alias.

xlai20 (changes requested):

src/storage/src/generated/gapic_control/transport.rs
joshuatants: Good catch. Placed the alias in `src/storage/lib.rs` to avoid coupling it with the data-plane crate.

xlai20 (approved):

PR #6747, opened 2026-09-09
refactor(generated): regenerate crates with first-class convert module

suzmue:
Convert.rs is now discoverable and is formatted.


dbolduc (approved):

suzmue:
Obsolete.

PR #6759, opened 2026-09-09
Revert "chore: update sources and regenerate"

dbolduc:
Reverts googleapis/google-cloud-rust#6757


alvarowolfx (approved):

dbolduc:
FTR, my plan is to revert this, then send another update sources, regeneration PR but with compute set to `skip_generate: true:

PR #6769, opened 2026-09-10
feat(storage): precompute checksums by default for unbuffered resumable uploads

xlai20:
This PR enables automatic checksum precomputation for unbuffered resumable uploads and updates the public API to give callers explicit control over this behavior.

Part of #6461

### Key Public API Changes

* **Deprecated API:**
  * [`WriteObject::precompute_checksums()`](file:///usr/local/google/home/xlai/feature-parity-rust/write/google-cloud-rust/src/storage/src/storage/write_object.rs): Deprecated because checksums are now automatically computed and validated by default across all upload modes.
* **New Replacement API:**
  * [`WriteObject::with_checksum_precomputation(bool)`](file:///usr/local/google/home/xlai/feature-parity-rust/write/google-cloud-rust/src/storage/src/storage/write_object.rs): Added with a default value of `true`. Callers performing unbuffered resumable uploads can pass `false` to opt out and prioritize upload latency over upfront checksum validation.

### Implementation Details

* **Unbuffered Resumable Precomputation:** When precomputation is enabled (default) and an upfront CRC32C is not already known, the client computes the checksum over the seekable payload before initiating the resumable session, passes the precomputed hash in session creation metadata, and streams the upload without redundant double-hashing.
* **Version Bump:** Bumped `google-cloud-storage` version to `1.19.0` across `Cargo.toml`, `librarian.yaml`, and `README.md`.
* **Tests:** Added comprehensive unit tests covering builder defaults, wire hash verification, payload rewinding, error handling on seek/read failures, retry memoization (ensuring no re-computation across transient retries), and non-redundant integration test cases.

### DIFF
--- a/src/storage/src/storage/checksum/details.rs
+++ b/src/storage/src/storage/checksum/details.rs
@@ -173,10 +173,6 @@ impl<S> ChecksummedSource<S> {
         self.checksum.finalize()
     }
 
-    /// Resets the internal checksumming state back to default (`None`).
-    ///
-    /// This disables on-the-fly checksumming for subsequent stream reads, which is
-    /// used after precomputing checksums to avoid redundant re-hashing during upload.
     pub fn reset_checksum(&mut self) {
         self.checksum = Checksum::default();
     }
--- a/src/storage/src/storage/checksum/details.rs
+++ b/src/storage/src/storage/checksum/details.rs
@@ -172,6 +172,18 @@ impl<S> ChecksummedSource<S> {
     pub fn final_checksum(&self) -> ObjectChecksums {
         self.checksum.finalize()
     }
+
+    /// Finalizes the computed checksum and resets internal checksumming state to
+    /// default (i.e. disabled / `None`).
+    ///
+    /// This is used when precomputing checksums before an upload pass: it extracts
+    /// the calculated checksum for metadata headers and leaves the source with no
+    /// active hashers, avoiding redundant recomputation during subsequent streaming.
+    pub fn take_checksum(&mut self) -> ObjectChecksums {
+        let computed = self.checksum.finalize();
+        self.checksum = Checksum::default();
+        computed
+    }
 }
 
 impl<S> StreamingSource for ChecksummedSource<S>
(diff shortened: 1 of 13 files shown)


src/storage/src/storage/checksum/details.rs
```diff
@@ -172,6 +172,10 @@ impl<S> ChecksummedSource<S> {
     pub fn final_checksum(&self) -> ObjectChecksums {
         self.checksum.finalize()
     }
+
+    pub fn reset_checksum(&mut self) {
+        self.checksum = Checksum::default();
```
vsharonlynn, line 177: Do we also need to reset the offset to 0 here?

src/storage/src/storage/checksum/details.rs
```diff
@@ -172,6 +172,10 @@ impl<S> ChecksummedSource<S> {
     pub fn final_checksum(&self) -> ObjectChecksums {
         self.checksum.finalize()
     }
+
+    pub fn reset_checksum(&mut self) {
```
vsharonlynn, line 176: `Checksum::default()` will initialise `Checksum {crc32c: None, md5_hash: None}`, so this function disables all further checksum computation on ChecksummedSource.
Please add docs to state this explicitly.

src/storage/src/storage/write_object.rs
```diff
@@ -1024,6 +1025,35 @@ where
...
+    /// let payload = tokio::fs::File::open("my-data").await?;
+    /// let response = client
+    ///     .write_object("projects/_/buckets/my-bucket", "my-object", payload)
+    ///     .with_checksum_precomputation(false)
+    ///     .send_unbuffered()
+    ///     .await?;
+    /// println!("response details={response:?}");
+    /// # Ok(()) }
+    /// ```
+    pub fn with_checksum_precomputation<V: Into<bool>>(mut self, enable: V) -> Self {
```
vsharonlynn, line 1052: Do we need to use `Into<bool>`? Can we just have this signature: `pub fn with_checksum_precomputation(mut self, enable: bool) -> Self`

vsharonlynn (changes requested):

src/storage/src/storage/checksum/details.rs
xlai20: No need. In ChecksummedSource, the "offset" field refers to the stream's physical cursor, not the status of checksum.

This function `reset_checksum()` only handles the checksum. It shouldn't be doing anything more than what its name entails.

src/storage/src/storage/checksum/details.rs
xlai20: Done. Added a doc for this function.

In general, when Checksum is set as None in both crc32c it is disabling computation; this was also mentioned in the validate() function in the same file.

src/storage/src/storage/write_object.rs
xlai20: Oops, I think the AI skill is over-learning the GAPIC machine-generated code that follows this pattern of function signature. Thanks for the comment!

Done in fixing this.

src/storage/src/storage/checksum/details.rs
vsharonlynn: Ok. Thanks for clarifying!

src/storage/src/storage/checksum/details.rs
vsharonlynn: Ok. Thanks!

src/storage/src/storage/write_object.rs
vsharonlynn: Cool. Thanks!

vsharonlynn (approved):

PR #6778, opened 2026-09-11
fix(gax-internal): omit empty descriptor strings

haphungw:
Only populate `type_name`, `json_name` and `default_value` when non-empty.

`wkt` represents unset optional strings as empty strings, while `prost_types` uses None. The conversion mapped `""` to `Some("")`, which is not the same as an absent field: it yields e.g. an empty `type_name` on a scalar field, which reflection engines reject when building a descriptor pool.

For #5315

### DIFF
--- a/src/gax-internal/src/prost.rs
+++ b/src/gax-internal/src/prost.rs
@@ -307,9 +307,9 @@ impl ToProto<prost_types::DescriptorProto> for wkt::DescriptorProto {
                     number: Some(v.number),
                     label: v.label.value(),
                     r#type: v.r#type.value(),
-                    type_name: (!v.type_name.is_empty()).then_some(v.type_name),
-                    json_name: (!v.json_name.is_empty()).then_some(v.json_name),
-                    default_value: (!v.default_value.is_empty()).then_some(v.default_value),
+                    type_name: Some(v.type_name).filter(|s| !s.is_empty()),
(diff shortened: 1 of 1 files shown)


src/gax-internal/src/prost.rs
```diff
@@ -307,9 +307,9 @@ impl ToProto<prost_types::DescriptorProto> for wkt::DescriptorProto {
                     number: Some(v.number),
                     label: v.label.value(),
                     r#type: v.r#type.value(),
-                    type_name: Some(v.type_name),
-                    json_name: Some(v.json_name),
-                    default_value: Some(v.default_value),
+                    type_name: (!v.type_name.is_empty()).then_some(v.type_name),
+                    json_name: (!v.json_name.is_empty()).then_some(v.json_name),
+                    default_value: (!v.default_value.is_empty()).then_some(v.default_value),
```
suzmue: alt:
```suggestion
                    type_name: Some(v.type_name).filter(|s| !s.is_empty()),
                    json_name: Some(v.json_name).filter(|s| !s.is_empty()),
                    default_value: Some(v.default_value).filter(|s| !s.is_empty()),
```

suzmue (approved):

src/gax-internal/src/prost.rs
haphungw: done

Issue #6776, opened 2026-09-11
Support publish hedging in  pubsub::client::Publisher


suzmue:
Support publish request hedging in `google-cloud-pubsub` to reduce p99/p99.9 tail latency caused by transient network stalls or slow backend tasks.
### Design
- **Hedging Mechanism:** When an in-flight batch publish exceeds a configured delay threshold, send a concurrent duplicate publish RPC and resolve the batch with whichever attempt finishes first.
- **Rate Limiting:** Rate-limit hedged RPCs using a token bucket that refills on successful publish RPCs.
- **Telemetry:** Attach the `x-goog-pubsub-client-telemetry` header to hedged attempts for backend tracking.
- **Ordering Safety:** Hedging does not apply for messages with ordering keys. 
- 
See java reference implementation: https://github.com/googleapis/google-cloud-java/pull/13735/changes

suzmue:
During implementation, I evaluated an alternative design that integrated hedging directly into the existing batch worker rather than using a dedicated background queue (which is used in the Java and Go implementations). 

While that alternative fits more cleanly into the existing publisher architecture, benchmark testing revealed noticeable differences in hedging performance and tail-latency mitigation. Not necessarily worse overall but worse for certain percentiles (though also better for others).  Because of this difference, I plan to proceed with implementing the queue design.

coryan:
> While that alternative fits more cleanly into the existing publisher architecture, benchmark testing revealed noticeable differences in hedging performance and tail-latency mitigation. Not necessarily worse overall but worse for certain percentiles (though also better for others).

nit (optional): there math-y are ways to compare distributions: as I recall, "stochastic dominance" is the magic thing to search for. Basically you are trying to ask "By blind luck, how often would I get a better outcome on distribution A vs. distribution B".

PR #6779, opened 2026-09-11
impl(pubsub): implement token bucket for request hedging

suzmue:
The token bucket will be used to limit the number of requests that are hedged. Use a scaled implementation similar to Java's to avoid unnecessary floating point math.

For #6776

### DIFF
--- a/src/pubsub/src/publisher/token_bucket.rs
+++ b/src/pubsub/src/publisher/token_bucket.rs
@@ -68,7 +68,7 @@ impl TokenBucket {
             .tokens
             .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
                 if current >= self.max_scaled_tokens {
-                    None
+                    None // optimization to save an atomic update.
                 } else {
                     Some(
                         current
@@ -84,7 +84,6 @@ impl TokenBucket {
 mod tests {
     use super::*;
 
-    use google_cloud_test_macros::tokio_test_no_panics;
     use tokio::task::{JoinSet, yield_now};
 
     impl TokenBucket {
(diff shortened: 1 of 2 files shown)


src/pubsub/src/publisher/token_bucket.rs
```diff
@@ -0,0 +1,199 @@
...
+    /// Refills the bucket by the configured refill amount.
+    ///
+    /// Tokens are capped at `max_tokens * SCALE`.
+    fn refill(&self) {
+        let _ = self
+            .tokens
+            .fetch_update(Ordering::Relaxed, Ordering::Relaxed, |current| {
+                if current >= self.max_scaled_tokens {
+                    None
+                } else {
```
dbolduc, line 72: optional: could remove this branch. Although it saves an atomic update.

src/pubsub/src/publisher/token_bucket.rs
```diff
@@ -0,0 +1,199 @@
...
+        }
+
+        // Spawn acquirer tasks
+        let mut acquirers = JoinSet::new();
+        for _ in 0..num_acquirers {
+            let bucket = Arc::clone(&bucket);
+            acquirers.spawn(async move {
+                let mut total_acquired = 0;
+                loop {
+                    if total_acquired == refills_per_task {
```
dbolduc: comment: I was originally confused, but I just missed that `scale == 1.0`

dbolduc: But this code assumes that `num_refillers == num_acquirers`... so I think we want to consolidate those variables.

dbolduc (approved):

src/pubsub/src/publisher/token_bucket.rs
suzmue: Yeah I did consider this, and decided to save the atomic update.

src/pubsub/src/publisher/token_bucket.rs
suzmue: True, cleaning up...

src/pubsub/src/publisher/token_bucket.rs
suzmue: adding a comment.

Issue #6717, opened 2026-09-07
jobs.insert retried with same job_id after timeout surfaces 409 Conflict as permanent error


coryan:
Unconfirmed.  This is the result of a search with Claude, may be a false positive.

In `src/bigquery/src/query/execution.rs`, `execute_jobs_insert` generates a `job_id` and marks the insert request idempotent:

https://github.com/googleapis/google-cloud-rust/blob/f02b42c2ac5720334cd2b5335824bd4166264527/src/bigquery/src/query/execution.rs#L87-L95

https://github.com/googleapis/google-cloud-rust/blob/f02b42c2ac5720334cd2b5335824bd4166264527/src/bigquery/src/query/execution.rs#L176-L185

If an initial `jobs.insert` request reaches BigQuery but times out or experiences a network reset during response delivery, GAX retries the identical request with the same `job_id`.

BigQuery responds to duplicate job creation requests with HTTP 409 Conflict. In `retry_policy.rs:42`, HTTP 409 is not in the retryable status list, so it is treated as a permanent error and returned to the caller, even though the query job is actively executing and being billed in BigQuery.

dbolduc:
I read this as: the client library could potentially recover from a 409 in the initial request of the polling loop.

Note that worst case, the application will see the job ID in the error message and can use `attach_job` to poll the job, and read the results.

alvarowolfx:
Endurance tests detected this error:

```
2026-09-11T17:31:19.209043Z ERROR ThreadId(15) bigquery_benchmark_queries::runner: Query::send failed: Rpc { source: Error { kind: Service(ServiceDetails { status_code: Some(409), headers: Some({"vary": "X-Origin", "vary": "Referer", "vary": "Origin,Accept-Encoding", "content-type": "application/json; charset=UTF-8", "date": "Fri, 11 Sep 2026 17:31:19 GMT", "server": "ESF", "x-xss-protection": "0", "x-frame-options": "SAMEORIGIN", "x-content-type-options": "nosniff", "accept-ranges": "none", "transfer-encoding": "chunked"}), status: Status { code: AlreadyExists, message: "Already Exists: Job aviebrantz-testing:US.job_bg6chsrbcA_YyHf2M84LRI7To7i4", details: [ErrorInfo(ErrorInfo { reason: "duplicate", domain: "global", metadata: {"message": "Already Exists: Job REDACTED"} })] } }), source: None } } self.task_id=15 iteration=71083
```

[...]

PR #6780, opened 2026-09-11
fix(bigquery): handle 409 errors and attach to existing job

alvarowolfx:
Towards #6717


alvarowolfx:
might need to split since handling this kind of error is tricky. 

`request_id` only merges duplicates that are still in flight. Once the request has been answered, later duplicates collide with the job it created. 409 on jobs.query only happens if the query executes correctly, but we get a transport failure to read results, so RPC level retries try to re-execute and we get an 409.

alvarowolfx:
superseded by #6801 #6799

PR #6787, opened 2026-09-12
impl(wkt): add helper for duplicate field detection

fornwall:
The generated deserializers track duplicate fields with a HashSet, which hashes each insertion and allocates storage as fields are recorded. Add wkt::internal::SeenFields<N>, a bitset backed by [u64; N], to support duplicate detection without those allocations or hashing.

The generator will assign each field a consecutive index and choose N as field_count.div_ceil(64). insert() returns false for an already recorded field, matching HashSet::insert(). Both new() and insert() can be used in const contexts. Multiple words support messages with more than 64 fields.

This change adds the helper and its tests. Generator adoption and regenerated code will follow in separate PRs. The generator must assign the same index to a field's JSON name and original proto name to preserve existing duplicate handling.

Towards #6788

### DIFF
--- a/src/wkt/src/internal.rs
+++ b/src/wkt/src/internal.rs
@@ -41,6 +41,9 @@ pub use float32::F32;
 mod float64;
 pub use float64::F64;
 
+mod seen_fields;
+pub use seen_fields::SeenFields;
+
 // For skipping serialization of default values of bool/numeric types.
 pub fn is_default<T>(t: &T) -> bool
 where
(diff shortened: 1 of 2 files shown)


coryan:
I created an issue and referenced said issue in your description. I have specific requests for the PR (here) and some questions about the improvements, which we should continue on that issue.

It would be nice to see the changes to the generator and the resulting output before merging this.  Can you send those as draft PRs and reference them here?

PR #6791, opened 2026-09-14
test(bigquery): add write throughput benchmark

dbolduc:
A throughput/endurance benchmark for the write client.

Apologies for the large PR, but I think we can squint at the code and be like "yeah, looks fine". (That is what I did).

Disclaimer: this is a tool. We are not going to run it in the CI. It is not production code. We really just want to make sure it measures what we think it is measuring.

For #6792


src/bigquery/benchmarks/write-throughput/src/args.rs
```diff
@@ -0,0 +1,105 @@
...
+    pub project: String,
+
+    #[arg(long, value_parser = parse_duration, default_value = "5s")]
+    pub report_interval: Duration,
+
+    #[arg(long, value_parser = parse_duration, default_value = "1m")]
+    pub duration: Duration,
+
+    #[arg(long, default_value_t = 1024)]
+    pub row_size: usize,
```
coryan: It would be nice to express these as KiB, MiB, etc. there is code in the w1r3 benchmark you could borrow.

alvarowolfx (approved):
that reminded myself to also submit the benchmark that I have been running for queries

PR #6794, opened 2026-09-14
test(bigquery): add query benchmark

alvarowolfx:
Benchmark for the query aspect of the bigquery crate to check resilience to retryable errors. Has different scenario with synthetic and hitting real tables, publishes metrics via Otel, writes errors to a dedicated error file (so we can better act on them) and also writes sample data to a .csv file to be queried later.

Disclaimer: this is a tool. We are not going to run it in the CI. It is not production code. We really just want to make sure it measures what we think it is measuring.

### DIFF
--- a/src/bigquery/benchmarks/queries/src/main.rs
+++ b/src/bigquery/benchmarks/queries/src/main.rs
@@ -1,0 +1,142 @@
+// Copyright 2026 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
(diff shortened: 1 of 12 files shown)


src/bigquery/benchmarks/queries/src/main.rs
```diff
@@ -0,0 +1,157 @@
...
+            let result = runner.run().await;
+            (task_id, result)
+        });
+    }
+
+    // Drop main sender so receiver terminates after all tasks complete
+    drop(tx);
+
+    tokio::select! {
+        _ = tokio::signal::ctrl_c() => {
```
dbolduc, line 125: TIL

src/bigquery/benchmarks/queries/src/telemetry.rs
```diff
@@ -0,0 +1,179 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+use crate::args::Args;
+use google_cloud_auth::credentials::Credentials;
+use integration_tests_o11y::detector::GoogleCloudResourceDetector;
```
dbolduc, line 17: cool

dbolduc (approved):

PR #6798, opened 2026-09-14
fix(bigquery): enforce idempotency requirements in retry policy

alvarowolfx:
Add check for duplicated job error and increase test coverage for idempotency settings.

Towards #6717 #6218


alvarowolfx:
Split from #6780

dbolduc (approved):

PR #6799, opened 2026-09-14
feat(bigquery): handle duplicate job error with jobs.insert

alvarowolfx:
A transport/io error can happen after query was inserted and RPC level retries might try to re-insert it with same job ID. Instead of re running the query, we can connect to the existing one to avoid running and billing a new query. 

Towards #6218 #6717


src/bigquery/src/query/execution.rs
```diff
@@ -176,13 +169,32 @@ impl RetryContext {
...
+                // an earlier attempt of this request reached the service. Adopt
+                // the job it created rather than fail a running, billing query.
+                let existing_job = match build_get_job(&job_service, &job_ref) {
+                    Some(get) => Box::pin(get.send()).await.ok(),
+                    None => None,
+                };
+                let Some(existing_job) = existing_job else {
+                    // The original error names the running job, and unlike a
+                    // `jobs.get` failure it never makes the job retry loop
+                    // reissue the query.
```
suzmue, line 191: super-nit: I think this comment is slightly more helpful if it is at the site where we throw away the jobs.get error with `.ok()`.

suzmue (approved):

PR #6801, opened 2026-09-14
feat(bigquery): handle duplicate job error with jobs.query

alvarowolfx:
A transport/io error can happen after query was submitted via jobs.query and finished execution, and RPC level retries might try to re-run it with same request ID. But request ID only dedups in-flight queries, this causes a 409 error. Instead of re running the query, we can connect to the existing one to avoid extra billing.

Towards https://github.com/googleapis/google-cloud-rust/issues/6218 https://github.com/googleapis/google-cloud-rust/issues/6717


alvarowolfx:
The debatable aspect here is that 409 errors on jobs.query don't have a structured way to provide job ID, so we have to do some error message parsing. Thoughts @dbolduc ?

dbolduc:
> The debatable aspect here is that 409 errors on jobs.query don't have a structured way to provide job ID, so we have to do some error message parsing. Thoughts @dbolduc ?

Error parsing is definitely a smell and brittle, but sometimes it must be done. We can open an issue / feature request with the service team to make it a structured error and reference that in the code.

src/bigquery/src/query/execution.rs
```diff
@@ -210,7 +204,33 @@ impl RetryContext {
...
+                // The request ID only dedups in-flight queries, so a resend
+                // after the query finished collides with the job the first
+                // attempt created. That job ran and was billed, so we try to
+                // attach to it.
+                let get = parse_duplicate_job_reference(&err)
+                    .and_then(|job_ref| build_get_job(&job_service, &job_ref));
+                let existing_job = match get {
+                    Some(get) => Box::pin(get.send()).await.ok(),
+                    None => None,
+                };
```
dbolduc, line 249: nit: this looks like an `.and_then()`

src/bigquery/src/query/execution.rs
```diff
@@ -210,7 +204,33 @@ impl RetryContext {
...
+                let Some(existing_job) = existing_job else {
+                    // This error names the job and, unlike a `jobs.get`
+                    // failure, never makes the retry loop reissue the query.
```
dbolduc: nit: there are a few ways to end up here. Should it be like?

```
// We were unable to parse the job in the error message, or successfully send a
// `GetJob` RPC. Return the original error message. 
```

Or: `// Oh well, we tried.` 😛

dbolduc (approved):
Looks good. I like that we are making the effort to retry these.

src/bigquery/src/query/execution.rs
alvarowolfx: the `.await` mess up trying to use `and_then` here, unless am I missing something. I tried to do that to chain on the previous `and_then`

PR #6809, opened 2026-09-15
impl(pubsub): implement publish request hedging runtime

suzmue:
Implement request hedging for batch publishing in the Pub/Sub publisher. When request hedging is enabled, if an initial publish RPC does not complete within the configured delay threshold and tokens are available in the token bucket, a hedged publish RPC is dispatched with `NeverRetry`. The first attempt to succeed resolves the pending publish futures, refills the token bucket, and cancels any remaining in-flight attempts via a shared CancellationToken. Only errors from the initial request are returned to the user, all other errors are discarded.

Rate-limited hedges (where no token is available at deadline expiry) are discarded to shed load and avoid sending excessive requests during sustained backend latency spikes.

Hedged request timeouts are capped at 10 seconds pending retry policy timeout integration (it should be the minimum of 10 seconds and total_timeout of the original request).

For #6776


dbolduc:
I think this PR is too big for me to provide good feedback. Can we break it down, e.g. start with some of the types from `hedging.rs`?

suzmue:
> I think this PR is too big for me to provide good feedback. Can we break it down, e.g. start with some of the types from `hedging.rs`?

Closing this, I am breaking it up: 

So far see:
#6818 
#6820

Issue #6355, opened 2026-08-12
BQ Write: Add stream retries


dbolduc:
Default stream is relatively simple. We just grab a new stream and retry the requests independently.

Exclusive streams will be more complicated. We need a retry manager that holds state (the ordered request queue) and makes sure that if any fail, all subsequent requests are retried in the correct order.

dbolduc:
I caught a stream disconnect in the wild. The error message looks like:

```
the operation failed. RPC error: the service reports an error with code ABORTED described as: Closing the stream because server is restarted. This is expected and client is advised to reconnect. Entity: projects/dbolduc-test/datasets/rust_bq_bench_dataset_v6tclwzk/tables/table_0/streams/_default
```

dbolduc:
We added retries for the default stream. We are tracking retries for exclusive streams in another issue, which is not part of this milestone. So closing this issue.

PR #6813, opened 2026-09-15
impl(bigquery): default write retry policy

dbolduc:
Define the default retry policy for the Write client. We will eventually expose this.

Note that all appends are idempotent as the service offers at-least-once semantics. The policy does not need to check it, it can assume.

Part of the work for #6355


src/bigquery/src/write/retry_policy.rs
```diff
@@ -0,0 +1,189 @@
...
+
+/// Follows the RPC retry strategy recommended for BigQuery Storage Write API.
+///
+/// This policy must be decorated to limit the duration of the retry loop or
+/// the number of attempts.
+#[derive(Clone, Debug)]
+#[allow(dead_code)]
+pub struct RetryableErrors;
+
+impl RetryPolicy for RetryableErrors {
```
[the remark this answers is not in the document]
dbolduc: This name is consistent with other APIs of ours. I am keeping it.

src/bigquery/src/write/retry_policy.rs
```diff
@@ -0,0 +1,189 @@
...
+        if let Some(429 | 500 | 502 | 503 | 504) = error.http_status_code() {
+            return RetryResult::Continue(error);
+        }
+        if let Some(status) = error.status() {
+            return match status.code {
+                Code::Aborted
+                | Code::DeadlineExceeded
+                | Code::Internal
+                | Code::ResourceExhausted
+                | Code::Unavailable => RetryResult::Continue(error),
```
alvarowolfx, line 53: Code::Cancelled should also be added here. Based on some history on reconnection/retry errors

b/326242484, cl/934719642

https://github.com/googleapis/google-cloud-go/pull/9642/
https://github.com/googleapis/nodejs-bigquery-storage/blob/057959ade8c920bb5dfcdc801eaf5693442ae7d9/src/managedwriter/stream_connection.ts#L157

alvarowolfx (approved):
Just make sure to add Code::Cancelled to the retry list

src/bigquery/src/write/retry_policy.rs
dbolduc: Fixed, thanks.

PR #6820, opened 2026-09-15
impl(pubsub): implement send functions for hedging

suzmue:
Implement the state to send and handle responses when hedging. All other requests are cancelled once one completes. Hedged responses only resolve on success.

For #6776


src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,360 @@
...
+            .publish()
+            .set_topic(self.topic.clone())
+            .set_messages((*self.msgs).clone());
+
+        tokio::select! {
+            _ = self.cancel_token.cancelled() => {}
+            res = request.send() => {
+                self.complete(res);
+            }
+        }
```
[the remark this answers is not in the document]
suzmue: This is the design. The initial attempt always wins if it comes back first.

src/pubsub/src/publisher/actor.rs
```diff
@@ -554,6 +554,33 @@ impl SequentialBatchActor {
...
+        Ok(result) => {
+            txs.into_iter()
+                .zip(result.message_ids)
+                .for_each(|(tx, result)| {
+                    // The user may have dropped the handle, so it is ok if this fails.
+                    let _ = tx.send(Ok(result));
+                });
+            Ok(())
+        }
+    }
```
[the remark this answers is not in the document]
suzmue: This isnt really a part of this PR, this is just needed for this code to pass. This was part of the refactor.

src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,360 @@
...
+
+/// Shared state for a single publish batch across initial and hedged RPC attempts.
+///
+/// Only the first attempt to finish (either initial or hedged) resolves the batch's
+/// pending `PublishFuture` channels and triggers cancellation of any remaining in-flight
+/// attempts for this batch.
+pub(crate) struct BatchState {
+    pub msgs: Arc<Vec<crate::model::Message>>,
+    /// Protected senders for the batch: user response channels and an internal `done_tx`
+    /// signal for the actor's inflight `JoinSet`. Taken atomically by the winning attempt.
```
dbolduc: "Taken atomically" seems wrong for a `Mutex` ?

src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,360 @@
...
+        let msg_id = rx.await??;
+        assert_eq!(msg_id, "msg-initial");
+        assert!(state.cancel_token.is_cancelled());
+        assert!(done_rx.await.is_ok_and(|r| r.is_ok()));
+
+        Ok(())
+    }
+
+    #[tokio_test_no_panics(start_paused = true)]
+    async fn test_initial_fails_option_1() -> anyhow::Result<()> {
```
dbolduc: option 1 ?

src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,360 @@
...
+        assert!(publish_res.is_err());
+        assert!(state.cancel_token.is_cancelled());
+        let done_res = done_rx.await?;
+        assert!(done_res.is_err());
+
+        Ok(())
+    }
+
+    #[tokio_test_no_panics(start_paused = true)]
+    async fn test_cancellation_when_initial_finishes_first() -> anyhow::Result<()> {
```
dbolduc, line 231: What is this testing that `test_initial_succeeds_fast` isn't?

src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,360 @@
...
+        let client = GapicPublisher::from_stub(mock);
+        let token_bucket = Arc::new(TokenBucket::new(10, 0.1));
+        let (state, rx, done_rx) = test_batch_state(client, token_bucket);
+
+        let initial_handle = tokio::spawn({
+            let state = state.clone();
+            async move {
+                state.send_initial().await;
+            }
+        });
```
dbolduc: optional nit:


```suggestion
        let initial_handle = {
          state = state.clone();
          tokio::spawn(async move { state.send_initial().await; })
        };
```

src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,360 @@
...
+            })
+        });
+
+        // Hedged succeeds after 20ms
+        mock.expect_publish().times(1).returning(|_, _| {
+            Box::pin(async {
+                tokio::time::sleep(Duration::from_millis(20)).await;
+                mock_publish_response("msg-hedged")
+            })
+        });
```
dbolduc: Q: do we actually know which is which?

Seems like the tokio::tasks could race. But the requests and options are the same as far as the mock is concerned so we can't tell them apart.

not a big deal, just wondering.

---

nit: use `mockall::Sequence`? I don't think the ordering of the responses matter, but I would prefer if its deterministic (which I think only happens with `mockall:Sequence`?)

src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,360 @@
...
+            return;
+        }
+
+        // TODO(#6776): clamp the timeout to the remaining time of the initial request.
+        let timeout = Duration::from_secs(10);
+
+        let request = self
+            .client
+            .publish()
+            .set_topic(self.topic.clone())
```
dbolduc, line 107: nit: if we remove this line, all of the tests still pass. Same for the `initial_rpc`. 🤷 

Eventually an integration test would catch it

src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,360 @@
...
+        #[derive(Debug)]
+        GapicPublisherWithFuture {}
+        impl crate::generated::gapic_dataplane::stub::Publisher for GapicPublisherWithFuture {
+            fn publish(&self, req: crate::model::PublishRequest, _options: google_cloud_gax::options::RequestOptions) -> impl Future<Output=google_cloud_gax::Result<google_cloud_gax::response::Response<crate::model::PublishResponse>>> + Send;
+        }
+    }
+
+    fn mock_publish_response(msg_id: &str) -> crate::Result<crate::Response<PublishResponse>> {
+        Ok(crate::Response::from(
+            PublishResponse::new().set_message_ids([msg_id.to_string()]),
```
dbolduc: nit:

```suggestion
            PublishResponse::new().set_message_ids([msg_id]),
```

src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,360 @@
...
+
+        Ok(())
+    }
+
+    #[tokio_test_no_panics(start_paused = true)]
+    async fn test_cancellation_when_hedged_succeeds_first() -> anyhow::Result<()> {
+        let mut mock = MockGapicPublisherWithFuture::new();
+
+        // Initial hangs for 10s
+        mock.expect_publish().times(1).returning(|_, _| {
```
dbolduc: optional nit: this seems to be deterministic, but use `mockall::Sequence`?

src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,360 @@
...
+
+        let _ = hedged_handle.await;
+        // Upon hedged completion, cancel_token was cancelled so initial finishes quickly
+        assert!(state.cancel_token.is_cancelled());
+        initial_handle.await?;
+
+        let msg_id = rx.await??;
+        assert_eq!(msg_id, "msg-hedged");
+        assert!(state.cancel_token.is_cancelled());
+        assert!(done_rx.await.is_ok_and(|r| r.is_ok()));
```
dbolduc:
```suggestion
        done_rx.await??;
```

here and elsewhere

dbolduc (approved):

src/pubsub/src/publisher/hedging.rs
suzmue: Done.

src/pubsub/src/publisher/hedging.rs
suzmue: fixed, this was leftover.

src/pubsub/src/publisher/hedging.rs
suzmue: Updated to make sure the hedged attempt actually is created.

src/pubsub/src/publisher/hedging.rs
suzmue: done.

src/pubsub/src/publisher/hedging.rs
suzmue: done.

src/pubsub/src/publisher/hedging.rs
suzmue: done.

PR #6822, opened 2026-09-15
impl(pubsub): add runner for hedging

suzmue:
The HedgingScheduler is used to process new batches that are inflight and schedule hedges for them if possible. If a batch is unable to acquire tokens, it is dropped from attempting to be hedged. All batches have the same delay, so now + delay is guaranteed to put it at the end of the queue.

For #6776


src/pubsub/src/publisher/hedging.rs
```diff
@@ -0,0 +1,661 @@
...
+            self.cancel_token.cancel();
+            if resp.is_ok() {
+                self.token_bucket.refill();
+            }
+            let _ = done_tx.send(batch_resolve_publish_futures(resp, txs));
+        }
+    }
+}
+
+/// An entry in the queue of scheduled hedges.
```
suzmue, line 143: NOTE: REVIEW BEGINS HERE

Code before this line is in PR #6820

src/pubsub/src/publisher/hedging.rs
```diff
@@ -135,6 +140,152 @@ impl BatchState {
...
+    pub token_bucket: Arc<TokenBucket>,
+}
+
+impl HedgingSchedulerHandle {
+    /// Dispatches a batch to the network and registers it with the hedging scheduler.
+    ///
+    /// 1. Immediately spawns `send_initial()` on a separate task.
+    /// 2. Spawns `done_rx` into the actor's `inflight` set so `flush()`/shutdown waits for the batch to resolve.
+    /// 3. Registers `state` with the `HedgingScheduler` to schedule hedged attempts if the initial request is slow.
+    pub(crate) fn dispatch(
```
dbolduc, line 162: codecov says this isn't covered. save it for the next PR?

src/pubsub/src/publisher/hedging.rs
```diff
@@ -135,6 +140,152 @@ impl BatchState {
...
+        // Send the initial RPC immediately instead of sending over the channel.
+        let state_clone = state.clone();
+        tokio::spawn(async move {
+            state_clone.send_initial().await;
+        });
+        let _ = self.tx.send(state);
+    }
+}
+
+/// Dedicated background task that manages scheduled hedges using a FIFO queue.
```
dbolduc, line 192: Why do we need a scheduler? Why can't we spawn background tasks for hedging attempts and let tokio do the scheduling?

Can we chain the RPCs with something like:

```rs
/// Dispatches an attempt and schedules the next hedge recursively via Weak reference.
    pub fn send_with_hedge(self: &Arc<Self>, attempt: usize) {
        // 1. Prepare weak reference to state
        let weak_state = Arc::downgrade(self);
        let delay = self.hedge_delay;

        // 2. Spawn next hedge timer
        tokio::spawn(async move {
            tokio::time::sleep(delay).await;

            // If the main batch hasn't finished, downgrade can upgrade back to Arc
            if let Some(state) = weak_state.upgrade() {
                // Check if the RPC is still waiting for a winner
                let is_pending = state.tx.lock().unwrap().is_some();
                if is_pending {
                    // Recursively trigger the next attempt and its subsequent hedge timer
                    state.send_with_hedge(attempt + 1);
                }
            }
        });

        // 3. Send the RPC for this attempt
        let state = Arc::clone(self);
        tokio::spawn(async move {
            let result = Self::rpc_call(attempt).await;

            // Deliver the response if we are the first to complete successfully
            state.complete(result);
        });
    }
```

^ pseudo code but you get the point

dbolduc: Basically, what is the scheduler doing that thhe runtime can't?

suzmue: I mention this in my comment on the actual bug, but I did implement a dispatching that does it from an individual attempt, instead of the queue mechanism that java is using, but I got different results. I would need to do a comparison of the implementations to see which is "better" as Carlos suggested. This is the implementation that Java has been finetuning and testing so I am mirroring it here.

The difference as I see it is the creation of timers for all of the batches that then need to manage themselves. We recently had a bug based on timers firing too often, so I am hesitant to use an implementation that will rely on a timer per batch without sufficient testing that it is not worse than having the background task.

src/pubsub/src/publisher/hedging.rs
suzmue: I would strongly prefer not to but if you insist I will do it. I have to rebase all of these PRs on top of each other a gazillion times and I'd rather just get the main hedging.rs PR done so the PR that uses it doesnt have to change the same file. If you want to see it tested used look at https://github.com/googleapis/google-cloud-rust/pull/6825.

dbolduc: > I would strongly prefer not to but if you insist I will do it. 

No problem. I will wait for the next PR

src/pubsub/src/publisher/hedging.rs
dbolduc: Ok. It seems like we either (1) spawn a task to do the hedge, or (2) send the thing on a channel and have an event loop that drains the channel and spawns a task to do the hedge.

That's why I assume (1) is simpler, but I think the extra cost of (2) is like < 1 microsecond per message. So it's not showing up next to a network call. And ownership of the hedged RPCs is much clearer with (2).

So I'm happy to go with (2).

src/pubsub/src/publisher/hedging.rs
```diff
@@ -415,4 +566,327 @@ mod tests {
...
+        scheduler_handle.tx.send(state.clone())?;
+
+        // Advance time by 150ms to trigger hedge
+        tokio::time::advance(Duration::from_millis(150)).await;
+        tokio::task::yield_now().await;
+
+        let msg_id = rx.await??;
+        assert_eq!(msg_id, "msg-hedged");
+        assert!(state.cancel_token.is_cancelled());
+        assert!(done_rx.await.is_ok_and(|r| r.is_ok()));
```
dbolduc:
```suggestion
        done_rx.await??;
```

src/pubsub/src/publisher/hedging.rs
```diff
@@ -135,6 +140,152 @@ impl BatchState {
...
+        }
+    }
+
+    fn handle_expired_hedges(&mut self) {
+        let now = tokio::time::Instant::now();
+        while let Some(item) = self.queue.front() {
+            if item.deadline <= now {
+                let item = self.queue.pop_front().unwrap();
+                // If the batch has already resolved or cannot acquire a token from the bucket,
+                // the hedge is discarded and not rescheduled to avoid sending excessive requests
```
dbolduc, line 242: Q: I think this is a fine choice, and pretty sure it won't matter in practice... but I am curious if you considered not sending a request, but still queuing another hedge if we can't acquire a token from the bucket.

I don't think anything should change. It just seems like that would also be valid behavior.

src/pubsub/src/publisher/hedging.rs
```diff
@@ -415,4 +566,327 @@ mod tests {
...
+        // Advance 60ms to t = 110ms -> Batch A's deadline (100ms) has expired, hedge A fires
+        tokio::time::advance(Duration::from_millis(60)).await;
+        tokio::task::yield_now().await;
+
+        let msg_a = rx_a.await??;
+        assert_eq!(msg_a, "msg-hedged-a");
+        assert!(state_a.cancel_token.is_cancelled());
+        assert!(done_rx_a.await.is_ok_and(|r| r.is_ok()));
+
+        // Batch B should not have completed yet at t = 110ms (deadline is 150ms)
```
dbolduc, line 878: nit: `.await`ing can advance the clock. If you want to assert what time it is, you can say like:

```
let start = tokio::time::Instant::now();
// do stuff
assert_eq!(start.elapsed, Duration::from_millis(110));
```

but warning: when background tasks are in the mix, I have seen non-deterministic timing and scheduling 🤷

dbolduc: Basically everything looks right... but this sort of timing has been a struggle for me in the past. It am not always doing what I think I am doing.

dbolduc (approved):

src/pubsub/src/publisher/hedging.rs
suzmue: Yeah I did consider this, and it seems valid to me too.

This is a part of the original design, but I think it would be worthwhile to test these differing implementations against each other to see which produces the best distribution.

src/pubsub/src/publisher/hedging.rs
suzmue: oops sorry I missed this.

PR #6823, opened 2026-09-15
feat(bigquery): add Read client

### LINKED ISSUE #5745: BigQuery read GAPIC
We need to generate a GAPIC for the BigQuery storage read API. The plan (go/cloud-rust:bigquery-high-level-design) is to put it in a `google-cloud-bigquery-storage-v1` crate.

- [x] Prototype streaming reads in GAPICs: #2318
      Some work was done in: #3377
- [x] Implement streaming read librarian templates, regenerate
- [ ] Design for resuming streams. Maybe stream resume policies?

dbolduc:
Add the `BigQueryRead` GAPIC to the `google-cloud-bigquery` crate. This is simpler, and makes the `Read` client compatible with the `Write` client as they share message types, like an `ArrowRecordBatch`.

Add all the models it needs to `mod model` and introduce a  `mod stub` for mocking the `Read` client (which has a `from_stub<T: Read>` method).

Of note, this PR has a drawback in that we are making public messages that go unused because they are covered by the Write veneer. (Like an `AppendRowsRequest`). It is not perfect but I accept the trade off of simplicity and 0-maintenance over trying to make the generated API surface minimal.

Fixes #5745, related to https://github.com/googleapis/librarian/pull/7593

cc: @tswast and ty: @suzmue for the full GAPIC :)


haphungw (approved):

src/bigquery/src/write/generated/gapic_storage/builder.rs
```diff
@@ -14,6 +14,350 @@
...
+            self,
+        ) -> Result<
+            google_cloud_gax::streaming::ResponseStream<
+                crate::write::generated::gapic_storage::model::ReadRowsResponse,
+            >,
+        > {
+            (*self.0.stub)
+                .read_rows(self.0.request, self.0.options)
+                .await
+        }
```
tswast, line 232: Would you all be open to a contribution that adds some reconnection logic here? The state machine adds a good bit of complexity though

https://github.com/pola-rs/polars-bigquery-client/blob/216de0ebc6e32bf31b1c3ef87ed0763cef0f62be/crates/arrow-bigquery/src/bigquery_read_stream.rs#L11-L29

so "no" is a perfectly valid answer.

tswast (commented):
Thanks so much @dbolduc and @suzmue ! Agreed on keeping the handwritten portion to a minimum, but I do have one "feature" that might make sense at this layer as opposed to every customer having to implement something similar.

What do you think?

src/bigquery/src/write/generated/gapic_storage/builder.rs
dbolduc: We are definitely open to it. We won't even need to touch thiis generated code. Rust is cool and lets us define functions on types in other places of the crate.

For example, @haphungw extended the generated builder for the `InsertJob` RPC with a handwritten `JobPoller` that holds state and can retry on job-level errors.

https://github.com/googleapis/google-cloud-rust/blob/506463276afab6accf14a206d890c476f9529243/src/generated/cloud/bigquery/v2/src/job_poller.rs#L192-L210

And that `JobPoller` supports its own custom options to configure its retry loop: https://docs.rs/google-cloud-bigquery-v2/latest/google_cloud_bigquery_v2/operation/struct.JobPoller.html#method.with_attempt_limit

So see if you can do something like that. Introduce some new type with state management, like maybe a `RowReader`. Then extend the builder to add `into_row_reader() ->  RowReader`.

tswast: Love that! I'll look into this next week.

PR #6828, opened 2026-09-15
fix(storage): preserve parent span context in appendable writer

alvarowolfx:
#### Root Cause
In `TestLayer::initialize()`, a `"test_layer"` span is created and entered into the calling thread's thread-local storage (TLS) via `tracing::span::EnteredSpan`. Because `EnteredSpan` is purely a thread-local guard, it is not preserved across `.await` yield points when Tokio resumes execution on a different worker thread.

When calling `writer.append(...)`, `writer.flush(...)`, `writer.finalize()`, or `writer.close()`:
- `TracingAppendableObjectWriter` previously relied on `#[tracing::instrument]` with contextual parent (`tracing::Span::current()`).
- After multiple `.await` yields (network handshake, streaming chunks, worker channel sends/joins), execution frequently resumed on worker threads whose TLS had `Span::none()` or an unrelated span.
- Consequently, the `finalize` (or `close`/`append`/`flush`) span was created without a link to `"test_layer"`.
- `find_test_id` in `TestLayer` failed to locate the `test_id` when traversing the span ancestry, so the span was recorded with `test_id: None` and excluded from `TestLayer::capture(&guard)`.
- This triggered the assertion failure: `missing finalize span in capture: [...]`.

#### Fix
1. **Preserve Parent Span in [`TracingAppendableObjectWriter`](file:///usr/local/google/home/aviebrantz/rust/google-cloud-rust.worktrees/fix-6324-tracing-flake/src/storage/src/storage/tracing.rs#L115-L127)**:
   - Added `parent: tracing::Span` to `TracingAppendableObjectWriter<T>`.
   - In `TracingAppendableObjectWriter::new`, captured `parent: tracing::Span::current()`. Since `new` is called within the instrumented `open_appendable_object` / `reopen_appendable_object` methods, `parent` reliably captures the client request span linked to `"test_layer"`.
   - Configured `parent = &self.parent` on `#[tracing::instrument]` for `append`, `flush`, `finalize`, and `close`.
2. **Re-enabled Tests in [`src/storage/src/storage/transport.rs`](file:///usr/local/google/home/aviebrantz/rust/google-cloud-rust.worktrees/fix-6324-tracing-flake/src/storage/src/storage/transport.rs)**:
   - Removed `#[ignore = "TODO(#6324) - disabled because it was flaky"]` from `open_appendable_object_success` and `open_appendable_object_finalize_success`.

#### Verification
- **Unit & Integration Tests**: All 7 bidi appendable object tests passed.
- **Concurrency Stress Test**: Ran 30 iterations in a loop with `--test-threads 16`; all 30 iterations passed with 0 flakes.
- **Code Formatter**: `cargo fmt -p google-cloud-storage -- --check` passed.
- **Strict Clippy**: `cargo clippy --all-features --no-deps -p google-cloud-storage -- -D missing_docs -D clippy::exhaustive_enums` passed cleanly.

Towards #6324


dbolduc (approved):
Thanks

PR #6830, opened 2026-09-15
impl(pubsub): pipe publish timeout to concurrent batch actor

suzmue:
Extract the total timeout duration from the configured retry policy in BasePublisher and forward it through PublisherPartialBuilder and Dispatcher to ConcurrentBatchActor in preparation for publish request hedging. Calculating it this way means we may be off from the actual total_timeout by a small amount (the time between creating the state and querying total_timeout). 

This is used by hedging to make sure that the timeout for a hedged request does not exceed the total timeout of the original request.

For #6776


suzmue:
I think we will want to include the whole retry policy, it turns out we want to resolve responses with non-transient errors, so we want the policy to find out which ones are considered non-transient.

suzmue:
> I think we will want to include the whole retry policy, it turns out we want to resolve responses with non-transient errors, so we want the policy to find out which ones are considered non-transient.

NVM we don't need it. just saving total_timeout for now.

src/pubsub/src/publisher/base_publisher.rs
```diff
@@ -102,4 +114,50 @@ mod tests {
...
+    #[tokio::test]
+    async fn default_total_timeout() -> anyhow::Result<()> {
+        let client = BasePublisher::builder()
+            .with_credentials(Anonymous::new().build())
+            .build()
+            .await?;
+        let timeout = client
+            .total_timeout
+            .expect("default total_timeout should be present");
+        assert!(timeout <= Duration::from_secs(600) && timeout >= Duration::from_secs(590));
```
dbolduc: This is an odd acceptable range for the default value

src/pubsub/src/publisher/base_publisher.rs
```diff
@@ -102,4 +114,50 @@ mod tests {
...
+    async fn custom_total_timeout() -> anyhow::Result<()> {
+        let client = BasePublisher::builder()
+            .with_credentials(Anonymous::new().build())
+            .with_retry_policy(AlwaysRetry.with_time_limit(Duration::from_secs(45)))
+            .build()
+            .await?;
+        let timeout = client
+            .total_timeout
+            .expect("custom total_timeout should be present");
+        assert!(timeout <= Duration::from_secs(45) && timeout >= Duration::from_secs(40));
```
dbolduc: Oh, it is that the time is moving. Can we `start_paused = true` and just `assert_eq!(...)`

dbolduc (approved):

src/pubsub/src/publisher/base_publisher.rs
suzmue: To make this work we also have to set_start on the RetryState. In prod, the clocks will be the same so it seems fine to do, especially to calculate the timeout.

Made this change.

PR #6832, opened 2026-09-15
impl(pubsub): add hedging options to the publisher builder

suzmue:
Skip the pubsub release temporarily. This is to allow all of the publish hedging changes to land without risking them getting accidentally released.

For #6776


src/pubsub/src/publisher/builder.rs
```diff
@@ -387,6 +440,56 @@ impl PublisherPartialBuilder {
...
+    /// # use google_cloud_pubsub::publisher::HedgingOptions;
+    /// # async fn sample() -> anyhow::Result<()> {
+    /// # let client: BasePublisher = BasePublisher::builder().build().await?;
+    /// let publisher = client
+    ///     .publisher("projects/my-project/topics/my-topic")
+    ///     .set_or_clear_hedging_options(None::<HedgingOptions>)
+    ///     .build();
+    /// # Ok(()) }
+    /// ```
+    pub fn set_or_clear_hedging_options<V>(mut self, v: Option<V>) -> Self
```
dbolduc, line 485: nit: I would vote against this unless hedging was enabled by default. I don't think clearing is necessary.

dbolduc (approved):

src/pubsub/src/publisher/builder.rs
suzmue: It will never be enabled by default. I guess the scenario where this could be useful is to be able to take a BasePublisher where it is set and then create one Publisher where it clears it. I'm going to leave it in and will schedule a design review where we can decide (I think deleting is easier than adding especially since I skipped publish).

PR #6833, opened 2026-09-16
feat(bigquery): retry loop for default writer

dbolduc:
Add a retry loop around requests on the default stream, which is enabled by default.

We need a custom loop to handle the nuances of the service.

The next PR will add support for handling the attempt and operation limits on the retry loop.

Towards #6355


alvarowolfx (approved):
LGTM

PR #6834, opened 2026-09-16
feat(spanner): add support for static and dynamic channel pooling

olavloite:
Adds support for configuring gRPC channel pooling when creating a Spanner client.

Customers can now configure the channel pool on `Spanner::builder()` using `.with_channel_pool(...)`:
- Static channel pool (`StaticChannelPoolConfig`): Configures a fixed number of gRPC channels (defaults to 4 channels, or 1 channel when targeting the Spanner emulator).
- Dynamic channel pool (`DynamicChannelPoolConfig`): Automatically scales the number of channels up or down based on RPC concurrency and load.


src/spanner/src/channel_pool/affinity.rs
```diff
@@ -55,16 +54,6 @@ impl TransactionAffinity {
         }
     }
 
-    /// Returns the provided affinity handle, or creates a new default `ReadOnly` affinity if `None`.
```
olavloite, line 58: Removed, as they were not in use

src/spanner/src/channel_pool/affinity.rs
```diff
@@ -255,37 +245,6 @@ mod tests {
         );
     }
 
-    #[test]
```
olavloite, line 258: Removed, as the methods that it tested were removed.

src/spanner/src/channel_pool/entry.rs
```diff
@@ -327,31 +327,12 @@ impl ChannelLease {
         Self { guard }
     }
 
-    /// Consumes the lease, returning the underlying active RPC guard.
```
olavloite, line 330: The methods that are being removed here are also removed because they were not used. (Note that error penalties are being used by this implementation, it was just the method here that is not needed)

src/spanner/src/session_maintainer.rs
```diff
@@ -66,6 +66,10 @@ impl ManagedSessionMaintainer {
         let session =
             Self::create_session(&spanner, &database_name, &database_role, &options, &o11y).await?;
 
+        spanner
```
olavloite, line 69: This is strictly speaking a bug fix that came up while working on this, and not directly related to making it publicly available. I did not create a separate PR for it, as the pool was not yet public before this, meaning that no one has been able to use the pool without this priming.

src/spanner/README.md
```diff
@@ -69,6 +69,31 @@ export SPANNER_EMULATOR_HOST=localhost:9010
 The client builder automatically detects this variable, connects to the emulator
 endpoint, and configures anonymous credentials.
 
+### Configuring the Channel Pool
+
+By default, the client uses a static pool of 4 gRPC channels. You can configure
```
olavloite, line 74: The default is likely to be changed to dynamic in the (near) future, once we have benchmarked it.

src/spanner/src/channel_pool/config.rs
```diff
@@ -239,120 +244,200 @@ impl Default for DynamicChannelPoolConfig {
...
+    /// Returns the maximum retry attempts for `SELECT 1` priming.
+    pub fn prime_max_attempts(&self) -> usize {
+        self.prime_max_attempts
+    }
+
+    /// Returns the channel selection strategy.
+    pub fn selection_strategy(&self) -> ChannelSelectionStrategy {
+        self.selection_strategy
+    }
+
```
alvarowolfx, line 440: nit: do we really need to make all those available to read ? At this point, is almost just easier to make the fields in `ChannelSelectionStrategy` `pub`. I would remove those methods and let customers request it (unless there is a known need). 

Just to keep in mind that making something public later is easier, but the opposite is a breaking change.

olavloite:
(Reverted to Draft for now, as we are still looking into some shortcomings of this implementation)

PR #6846, opened 2026-09-17
impl(bigquery): identify write client to server

### LINKED ISSUE #5910: Identify BQ Write client to server
We need to set the `trace_id` field to identify our library. Probably something like: "rust-streamwriter" plus some other identifying information.

https://github.com/googleapis/googleapis/blob/2622f170ca5441590b472785b230eac029c67ab5/google/cloud/bigquery/storage/v1/storage.proto#L527-L529

Googlers can see the spec at: https://docs.google.com/document/d/1BlnBv_ijvpDcq9kaH-wQF-INu_QFfdfJpPxQfwk6hxg

dbolduc:
The BQ Storage Write server folks have their own telemetry to identify client languages. We will tell them ours is the rust client, and what version it is.

This only needs to be set on the first request of a stream.

Fixes #5910


src/bigquery/src/write/runner.rs
```diff
@@ -70,6 +70,10 @@ async fn run_stream_task(inner: Arc<Transport>, mut req_rx: mpsc::UnboundedRecei
         return;
     };
 
+    // Identify the client library to the server on the initial request.
+    let mut req = initial_req.req;
+    req.trace_id = format!("rust-writer:{VERSION}");
```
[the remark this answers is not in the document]
dbolduc: We don't expose this field. If we do expose it as an option, we would have to address this.

src/bigquery/src/write/transport.rs
```diff
@@ -19,11 +19,11 @@ use gaxi::grpc::tonic::{Response as TonicResponse, Streaming};
 use tokio::sync::mpsc::Receiver;
 use tokio_stream::wrappers::ReceiverStream;
 
-mod info {
+pub(super) mod info {
     use std::sync::LazyLock;
 
     const NAME: &str = env!("CARGO_PKG_NAME");
-    const VERSION: &str = env!("CARGO_PKG_VERSION");
+    pub(in super::super) const VERSION: &str = env!("CARGO_PKG_VERSION");
```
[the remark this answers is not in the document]
dbolduc: Fixed

alvarowolfx (approved):

src/bigquery/src/write/runner.rs
```diff
@@ -506,4 +510,65 @@ mod tests {
...
     }
+
+    #[tokio::test]
+    async fn trace_id() -> anyhow::Result<()> {
+        // We use this channel to surface writes (requests) from outside our
+        // mock expectation.
+        let (recover_writes_tx, mut recover_writes_rx) = mpsc::channel(10);
+        let (_, response_rx) = mpsc::channel(10);
+        let mut mock = MockBigQueryWrite::new();
+        mock.expect_append_rows().return_once(move |request| {
```
alvarowolfx, line 521: isn't `append_rows` gonna be called more than once here since we have two write requests ? I think that's why the integration tests are failing on the second request.

alvarowolfx: oh, nvm. It is going to be called just once to open the stream.

dbolduc: d'oh. thanks for looking, I did not notice the build failures from my flaky test. Fixed by closing the stream gracefully and awaiting the stream task completing (instead of having that race with exiting the test).

```sh
$ cargo nextest run -p google-cloud-bigquery write::runner::tests::trace_id --stress-count 1000

Summary [  21.417s] 1000/1000 stress run iterations: 1000 passed
```

PR #6849, opened 2026-09-17
test(bigquery): add Storage Write API tests for flaky tables

alvarowolfx:
Add integration tests for BigQuery Storage Write API exercising reconnection and retry behaviors on simulated failure tables in `us-east7`:

- `_reconnect_on_close`: sequential and parallel appends to pending streams
  verifying connection recovery on periodic drops (every 10 requests).
- `_initial_connect_failure`: sequential and parallel appends verifying
  retry logic during frequent initial stream connect failures.
- Default stream appends under simulated connection drops.
- Stream finalization, commit validation, and row count verification.

Towards ???


tests/bigquery/src/writes/flaky.rs
```diff
@@ -0,0 +1,326 @@
...
+
+    // Pre-generate all batches
+    let mut batches = Vec::with_capacity(ITERATIONS);
+    for i in 0..ITERATIONS {
+        let offset = (i * ROWS_PER_BATCH) as i64;
+        let batch = serializer.generate_batch(ROWS_PER_BATCH, i * ROWS_PER_BATCH)?;
+        batches.push((offset, batch));
+    }
+
+    let writer = Arc::new(client.arrow(serializer.schema()).pending(table).await?);
```
dbolduc, line 110: I was seeing errors from the server because the serialized schema length was 0.

I think you are assuming the `ArrowSerializer` test helper was written well, when it wasn't. Oops, sorry.

At HEAD, you have to grab the schema before serializing any batches. I just sent #6854 to make `ArrowSerializer` safer, so either order will do. You will probably want to rebase on it.

tests/bigquery/src/writes/flaky.rs
```diff
@@ -0,0 +1,331 @@
...
+    let table_id = format!(
+        "{}{RECONNECT_ON_CLOSE_SUFFIX}",
+        bigquery_samples::random_id_suffix()
+    );
+    create_flaky_table(table_service, project_id, dataset_id, &table_id, schema).await?;
+
+    let table = format!("projects/{project_id}/datasets/{dataset_id}/tables/{table_id}");
+    let mut serializer = ArrowSerializer::new("reconnect_seq")?;
+
+    let writer = client.arrow(serializer.schema()).pending(table).await?;
```
dbolduc, line 50: We don't have retries for pending streams yet...

I suspect that we are getting well behaved tables bc our test project is not allowlisted to activate the flaky behavior.

Does this stuff pass on your test project? Can you see if we are actually getting disconnects from the server?

alvarowolfx: yeah, running on my test project they pass too. I still need to do some work to add our CI test to the allowlist. 

I'll add some more debug logs here to check if the behavior is being triggered.

PR #6857, opened 2026-09-18
chore(storage): send veneer header instead of gapic

alvarowolfx:
Towards #2138


src/storage/src/control/client.rs
```diff
@@ -120,6 +120,11 @@ pub(crate) mod client_builder {
             if config.backoff_policy.is_none() {
                 config.backoff_policy = Some(Arc::new(crate::backoff_policy::default()));
             }
+            config.extensions.insert(gaxi::api_header::XGoogApiClient {
+                name: crate::storage::info::NAME,
+                version: crate::storage::info::VERSION,
+                library_type: gaxi::api_header::GCCL,
+            });
```
dbolduc, line 127: While I think either works, I would rather keep calling `StorageControl` a GAPIC because it has a fully generated surface. (even tho that fully generated surface is a veneer over two other GAPICs 🤷)

src/storage/src/storage/client.rs
```diff
@@ -389,6 +389,11 @@ impl StorageInner {
         let (mut config, options) = builder.into_parts()?;
         config.disable_automatic_decompression = true;
         config.disable_follow_redirects = true;
+        config.extensions.insert(gaxi::api_header::XGoogApiClient {
```
dbolduc, line 392: hm, it looks like we define the veneer header in `src/storage/src/storage.rs` for the gRPC based requests. But maybe we are missing some.

When I look at our telemetry I see some GAPICs and some veneers for `storage.object.get`, e.g.:

https://screenshot-v2.corp.google.com/5o7dsl4mpvajo

I would approve this change, but maybe worth seeing if we use the GAPIC header anywhere outside of `StorageControl`.

PR #6861, opened 2026-09-18
feat(pubsub): support publish hedging

suzmue:
Finalize API and enable release of pubsub.

For #6776


dbolduc (approved):

src/pubsub/src/publisher/builder.rs
```diff
@@ -478,14 +475,11 @@ impl PublisherPartialBuilder {
...
-    ///     .set_or_clear_hedging_options(None::<HedgingOptions>)
+    ///     .set_or_clear_hedging_options(None)
     ///     .build();
     /// # Ok(()) }
     /// ```
-    pub fn set_or_clear_hedging_options<V>(mut self, v: Option<V>) -> Self
-    where
-        V: Into<HedgingOptions>,
-    {
+    pub fn set_or_clear_hedging_options(mut self, v: Option<HedgingOptions>) -> Self {
```
dbolduc, line 482: Oh, we do have `set_or_clear_*` for fields with presence like messages:

https://docs.rs/google-cloud-secretmanager-v1/latest/google_cloud_secretmanager_v1/builder/secret_manager_service/struct.AddSecretVersion.html#method.set_or_clear_payload

LGTM

src/pubsub/src/publisher/builder.rs
```diff
@@ -460,7 +457,7 @@ impl PublisherPartialBuilder {
     ///     .build();
     /// # Ok(()) }
     /// ```
-    pub fn set_hedging_options<V: Into<HedgingOptions>>(mut self, v: V) -> Self {
+    pub fn set_hedging_options(mut self, v: HedgingOptions) -> Self {
```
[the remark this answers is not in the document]
suzmue: fixed.

src/pubsub/src/publisher/builder.rs
dbolduc: Another argument for keeping the API is that an application might want to explicitly disable hedging in case the client library ever changes its default behavior.

dbolduc (approved):

PR #6870, opened 2026-09-19
fix(bigquery-derive): prevent field names from shadowing macro locals

### LINKED ISSUE #6869: BQ Query: Variable shadowing in derive macros

dbolduc:
Use struct initializations instead of local variables in a macro. When we do use a local variable, put it in a scope.

Without this application structs with a field like `row` could not be annotated with `derive(FromSql)`. They would have to use the rename feature to workaround this.

Fixes #6869


alvarowolfx (approved):
nice, thanks

PR #6907, opened 2026-09-21
fix(bigquery): preserve SQL field order and support anonymous structs for derive macros

### LINKED ISSUE #6892: BQ Query: `RECORD`, `STRUCT` stored in a Map, losing anonymous and duplicated fields.

dbolduc:
Change the internal representation of Rows to (mostly) avoid `serde_json`.

For structs:

A `wkt::Value` is just a `serde_json::Map`, which is implemented by a `BTreeMap`. Using an ordered map means if we have duplicated column names (which can happen when joining tables) the last value will overwrite all previous ones. It also means that statements like `SELECT (1 as z, 2 as a)` will sort the inputs and flip the order when indexing by number.

The difference is that looking up a column by name is now a find on a Vec<> instead of aHashMap lookup. O(N) instead of O(1). For small N (columns) this is just as fast. BQ allows up to 10,000 columns in a query. It might be noticeable at that scale, but most queries with 10,000 columns are probably indexing by number, not name.

I think most use cases are improved with the Vec and the Vec is more correct. I will happily take this trade off.

Avoiding serde_json also helps avoid serialization costs.

I will change Number to have `i64` and `f64` variants instead of storing it as a `serde_json::number` in a follow up PR.

Fixes #6892


dbolduc:
This is a good change, but it's doing two things at once. (1) changing the internal representation of a value and (2) adding tuple support in the macro.

That's a dead giveaway that this should be two PRs.

I will leave this PR open until its sub-PRs are out.

Issue #6874, opened 2026-09-20
BQ Query: JSON not supported by derive macros


dbolduc:

coryan:
Could you say more words ? This is a bit scary ...

coryan:
I see the parent bug has the details, sorry.

dbolduc:
> I see the parent bug has the details, sorry.

No problem. Converting tasks to subissues makes the context less apparent.

PR #6930, opened 2026-09-21
fix(bigquery): preserve SQL field order and anonymous fields in SqlValue

dbolduc:
Fixes three correctness issues in `STRUCT` deserialization (`convert_nested` and `SqlValue`):
1. **Anonymous `STRUCT` data loss (`SELECT STRUCT(10, 'hello', true)`):**
   BigQuery sets `field.name = ""` for unnamed subfields inside a `STRUCT`. Previously, `convert_nested` collected subfields into a `wkt::Struct` (`BTreeMap<String, Value>`), causing every unnamed subfield to overwrite the `""` key and drop all but the last value.
2. **Alphabetical field-order scrambling (`SELECT STRUCT(1 AS z, 2 AS a)`):**
   Because `wkt::Struct` is backed by a `BTreeMap`, keys were reordered alphabetically (`"a"` before `"z"`) instead of preserving the SQL `SELECT` declaration order.
3. **Duplicate field-name inconsistency (`SELECT STRUCT(100 AS id, 200 AS id)`):**
   On a top-level `Row`, `row.get("id")` returns the *first* column named `"id"` (`100`). Inside a `STRUCT`, `BTreeMap` insertion overwrote earlier keys and returned the *last* `"id"` (`200`), destroying the first value.
### Changes
- Introduces `SqlValueInner` (`SqlValueInner::Struct(Vec<(String, SqlValueInner)>)`) inside `SqlValue`, avoiding per-row `BTreeMap` allocations unless the caller explicitly requests `wkt::Struct` / `wkt::Value`.
- Updates `SqlValue::take` (`usize` and `&str`) to swap taken slots in-place with `Null` (`std::mem::replace`), matching `Row::take` semantics (no index shifting, and subsequent reads of a taken field see `Null`).

---

^ That was the bot's write up and here are some of my thoughts:

The difference is that looking up a column by name is now a find on a `Vec<>` instead of a `HashMap` lookup. `O(N)` instead of `O(1)`. For small `N` (columns) this is just as fast. BQ allows up to 10,000 columns in a query. It might be noticeable at that scale, but most queries with `10,000` columns are probably indexing by number, not name.

I think most use cases are improved with the Vec and the Vec is more correct. I will happily take this trade off.

---

I considered testing the `SqlValueInner` interface directly, but decided against it. I think the end-to-end test coverage we have is what we want. We care about the observable behavior of the public interfaces, which is already covered.

---

Avoiding serde_json also helps avoid serialization costs.

I will change Number to have i64 and f64 variants instead of storing it as a `serde_json::number` in a follow up PR.

Towards #6892


alvarowolfx (changes requested):
Can we add to the integration tests to exercise anonymous structs ?

src/bigquery/src/query/from_sql.rs
```diff
@@ -33,19 +33,57 @@ pub(crate) const BIGQUERY_DATETIME_SUBSEC_FORMAT: &[time::format_description::Fo
...
+            wkt::Value::Null => Self::Null,
+            wkt::Value::Bool(b) => Self::Bool(b),
+            wkt::Value::Number(n) => Self::Number(n),
+            wkt::Value::String(s) => Self::String(s),
+            wkt::Value::Array(arr) => Self::Array(arr.into_iter().map(Self::from_wkt).collect()),
+            wkt::Value::Object(obj) => Self::Struct(
+                obj.into_iter()
+                    .map(|(k, v)| (k, Self::from_wkt(v)))
+                    .collect(),
+            ),
```
alvarowolfx, line 69: is this allocating a new wkt::Value::Objct for every row that has a struct type ?

src/bigquery/src/query/from_sql.rs
```diff
@@ -123,15 +161,31 @@ pub trait FromSql: Sized {
...
+                arr.into_iter()
+                    .map(|v| wkt::Value::from_value(SqlValue::from_inner(v)))
+                    .collect::<Result<Vec<_>, _>>()?,
+            ),
+            SqlValueInner::Struct(entries) => wkt::Value::Object(
+                entries
+                    .into_iter()
+                    .map(|(k, v)| Ok((k, wkt::Value::from_value(SqlValue::from_inner(v))?)))
+                    .collect::<Result<wkt::Struct, ConvertError>>()?,
+            ),
```
alvarowolfx, line 179: same comment here around allocations, I wonder if we should check this against some of the bench's that I wrote on allocations

tests/bigquery/src/query.rs
```diff
@@ -417,12 +417,50 @@ struct UserProfile {
...
+struct PositionalPair(i64, i64);
+
+// TODO(#6892) - use the derive macro when it supports tuples
+impl FromSql for PositionalPair {
+    fn from_value(
+        mut value: google_cloud_bigquery::query::SqlValue,
+    ) -> std::result::Result<Self, google_cloud_bigquery::error::ConvertError> {
+        Ok(Self(value.take(0)?, value.take(1)?))
+    }
+}
```
alvarowolfx, line 448: this is cool, even without the derive macro

alvarowolfx (approved):

src/bigquery/src/query/from_sql.rs
dbolduc: No this `from_wkt` business is mostly used in tests (so they didn't change much when we flipped the representation).

But then the only prod use case was an artifact of: https://github.com/googleapis/google-cloud-rust/issues/6874

It only comes into play when the application tries to deserialize a field of their struct into a custom type where the column is type `JSON`.

Maybe we should be conservative and just revert that, so it is only in tests for now. We can add it later, and the workaround is to represent something as a `String` and then applications deserialize it themselves.

src/bigquery/src/query/from_sql.rs
dbolduc: I would very much like to see the allocations benchmark with this impl. I think it has to do better to not put all of the data into a `serde_json::Map`.

PR #6932, opened 2026-09-21
fix(auth): avoid busy-loop in token cache on short/expired tokens

### LINKED ISSUE #6678: Tight infinite busy-loop in token cache refresh on short expiration tokens
Unconfirmed.  This is the result of a search with Gemini, may be a false positive.

In `TokenCache::refresh_task`, the loop calculates sleep duration until expiry using `expiry.checked_duration_since(Instant::now())`:

https://github.com/googleapis/google-cloud-rust/blob/f02b42c2ac5720334cd2b5335824bd4166264527/src/auth/src/token_cache.rs#L180-L196

Specifically lines 186–194:

https://github.com/googleapis/google-cloud-rust/blob/f02b42c2ac5720334cd2b5335824bd4166264527/src/auth/src/token_cache.rs#L186-L194

```rust
if time_until_expiry > NORMAL_REFRESH_SLACK {
    sleep(time_until_expiry - NORMAL_REFRESH_SLACK).await;
} else if time_until_expiry > SHORT_REFRESH_SLACK {
    // If expiry is less than 4 mins, try to refresh every 10 seconds
    // This is to handle cases where MDS **repeatedly** returns about to expire tokens.
    sleep(SHORT_REFRESH_SLACK).await;
}
```

If the token provider returns a token expiring in `<= SHORT_REFRESH_SLACK` (10 seconds), or an already-expired token (where `checked_duration_since` returns `None`), neither `if` branch executes.

Because there is no fallback `else` branch or backoff sleep, the loop immediately ends its iteration and restarts, querying `token_provider.token().await` without any delay. If the metadata server (MDS) or auth backend repeatedly returns an about-to-expire token, this leads to an infinite 100% CPU busy loop that hammers the auth server and starves other async tasks.

colinmoy:
Fixes #6678


src/auth/src/token_cache.rs
```diff
@@ -435,16 +433,21 @@ mod tests {
             refresh_task(Arc::new(mock), tx).await;
         });
 
-        // Give the refresh task a chance to run
-        sleep(Duration::from_millis(100)).await;
```
dbolduc, line 439: comment: lol, we have since learned about `tokio::task::yield_now()` to pause execution in the current task and run other tasks

src/auth/src/token_cache.rs
```diff
@@ -435,16 +433,21 @@ mod tests {
...
-
         rx.changed().await.unwrap();
+        let (actual, ..) = rx.borrow().clone().unwrap().unwrap();
```
dbolduc, line 437: nit: we have since learned about `anyhow::Result<()>` which can let us use the `?` for early returns on error (vs a panic from unwrap.

and nit: i tend to use `.expect("reason")` over unwrap


```suggestion
         rx.changed().await?;
        let (actual, ..) = rx.borrow().clone().expect("should have a token")?;
```

src/auth/src/token_cache.rs
```diff
@@ -435,16 +433,21 @@ mod tests {
...
+        assert_eq!(actual, token1);
 
+        // Advance past SHORT_REFRESH_SLACK; now the refresh task should fetch token2.
+        tokio::time::advance(SHORT_REFRESH_SLACK).await;
+        rx.changed().await.unwrap();
         let (actual, ..) = rx.borrow().clone().unwrap().unwrap();
-        assert_eq!(actual, token2.clone());
+        assert_eq!(actual, token2);
     }
```
dbolduc, line 451: and then here we would return Ok
```suggestion
        Ok(())
     }
```

src/auth/src/token_cache.rs
```diff
@@ -400,14 +395,17 @@ mod tests {
         Ok(())
     }
 
-    #[tokio::test]
-    async fn refresh_task_expired_token_loop() {
+    #[test_case::test_case(Duration::ZERO; "already_expired")]
+    #[test_case::test_case(SHORT_REFRESH_SLACK / 2; "less_than_short_slack")]
+    #[test_case::test_case(SHORT_REFRESH_SLACK; "equal_to_short_slack")]
+    #[tokio::test(start_paused = true)]
+    async fn refresh_task_expired_token_loop(initial_ttl: Duration) {
```
dbolduc, line 402: And the test function would return an `anyhow::Result<()>`


```suggestion
    async fn refresh_task_expired_token_loop(initial_ttl: Duration) -> anyhow::Result<()> {
```

dbolduc (approved):
Thanks, this is a good fix!

I have non-blocking nits in the test code. Feel free to merge the PR as is because addressing the nits will make the other tests in the file look bad.

PR #6933, opened 2026-09-21
refactor(bigquery): invert the writer builder order

### LINKED ISSUE #6855: BQ Write: improve writer builder patterns
## Requirements

Based on feedback from API reviews, we want a few things:

- Multiplexing to only be an option on the default stream (it doesn't apply to others)
- Client APIs that are more explicit and descriptive (and read like verbs)
- Consistency between `attach` and `create` APIs

## Overview

After workshopping the great ideas from the team, I am proposing we move from:

```rs
let w = client.arrow(as).default(t).await?;
let w = client.arrow(as).pending(t).await?;
let w: PendingWriter = client.arrow(as).attach(ws).await?;
```

to

```rs
let w = client.open_default_stream(t).build_arrow(as).await?;
let w = client.create_stream<Pending>(t).build_arrow(as).await?;
let w = client.attach_to_stream<Pending>(ws).build_arrow(as).await?;
```

## Details

We need a new trait representing the stream type. I am proposing:

```rs
trait StreamType : sealed StreamType {
  type Writer;
}

// Marker implementations
struct Default;
impl StreamType for Default {
  type Writer = DefaultWriter;
}
struct Pending;
impl StreamType for Pending {
  type Writer = PendingWriter;
}
// etc.
```

And then we still need our existing trait (now called `Writer`) to represent the non-default streams. We restrict `create` and `attach` to only accept marker structs that also implement this trait.

```rs
trait CreatedStreamType : sealed CreatedStreamType { }

// Marker implementations - note: no default
struct Pending;
impl CreatedStreamType for Pending {
  type Writer = PendingWriter;
}
// etc.
```

The clients three APIs -- `open_default_stream`, `create_stream`, `attach_to_stream` -- all yield a writer builder which is generic on the StreamType: `WriterBuilder<StreamType>`.

Then we have generic impls such that any `WriterBuilder<T>` can `async fn build_arrow(schema: ArrowSchema) -> StreamType::Writer<ArrowFormat>`. And later we add analogous methods for `json` and `proto`.

This also lets us specialize and define multiplexing only on the default writer:

```rs
impl WriterBuilder<DefaultStream> {
  fn with_multiplexing(self, bool: enable) -> Self;
}
```

(this was the main motivation for inverting the builder from data format first to stream type first.)

And then minor point: we delay making the RPC until the `build_*`. That is because in cases where the application supplies a schema, we do not need to ask for the full view on a `GetWriteStream` RPC. But when we add support for JSON, we will want to ask for the full view so we can use the `TableSchema`. Therefore, we must wait until the application tells us the format to do async work. (Also, it just feels more natural to say `.await?` at the end.)

dbolduc:
This PR flips the BQ Write flow to be action first (`open_default_stream`, `create_stream`, `attach_to_stream`) and format second (for now just `build_arrow()`).

These names are more explicit, and it allows us to only define `with_multiplexing()` on a `WriterBuilder<DefaultStream>`. (Which is great, because the option doesn't apply to other streams.

We also remove the old way of doing things.

I apologize for the ugly diff. I wanted to break this down, but I ran out of time and need to shortcut to the ending. The reviewer can look at individual commits to see an implementation path (which involves scaffolding). Or they can just look at the final state.

Fixes #6855


alvarowolfx (approved):

src/bigquery/src/write/client.rs
```diff
@@ -64,28 +63,94 @@ impl Write {
...
+    /// # Ok(()) }
+    ///
+    /// use google_cloud_bigquery::model::ArrowSchema;
+    /// fn schema() -> ArrowSchema {
+    ///   todo!("Define your table's schema...")
+    /// }
+    /// ```
+    ///
+    /// [application-created stream]: https://docs.cloud.google.com/bigquery/docs/write-api-grpc#application-created_streams
+    pub fn create_stream<S: ApplicationCreatedStream, T: Into<String>>(
```
suzmue, line 104: nit: I don't think we lose too much by making table just accept a String instead of Into<String>, though I know it is not consistent with the rest of our APIs. Since we expect users to need to specify S, it could be a nice cleanup.

src/bigquery/src/write/client.rs
```diff
@@ -64,28 +63,94 @@ impl Write {
...
-    // For now we share the stream pool. This is safe because the
-    // protobuf-based surface is not exposed. Ideally they would have separate
-    // pools.
-    #[allow(dead_code)]
-    pub(crate) fn proto(&self, schema: ProtoSchema) -> WriterBuilder<Proto> {
-        let format = Proto { schema };
-        WriterBuilder::new(
+    /// Creates a new [application-created stream] for the given table.
+    ///
+    /// The stream type `S` can be inferred from the variable's writer type
```
suzmue, line 78: Wait is this true? I thought it couldn't be. Really cool!

dbolduc: Yeah! The writers impl `HasStream`, which points back to the `Stream` type.

The examples speak for themselves :sunglasses:

src/bigquery/src/write/client.rs
dbolduc: hmm that is also a reasonable trade off.

I would sooner accept just `&str`, but I think I like the tradeoff of accepting `"foo"` or `foo` over having the neater turbofish notation with one generic.

In my head, applications are not naming any of the types from `mod stream_types`, so I have the turbofish notation as kind of a second-class experience. And adding `_` is... fine. It's sorta comparable to having to say like `let v: Vec<_> = blah.collect();`

suzmue (approved):

PR #6937, opened 2026-09-22
perf(bigquery): remove `tokio::spawn` from `AppendWithOffset`

### LINKED ISSUE #6898: BQ Write: `AppendWithOffset` does not need a `tokio::spawn`

haphungw:
In `AppendWithOffset::send()`, eliminate the intermediate oneshot channel and background `tokio::spawn` task by returning an inline async future via `AppendFuture::from_future`.

Fixes #6898


src/bigquery/src/write/builder/append_with_offset.rs
```diff
@@ -236,4 +231,66 @@ mod tests {
...
+            .expect_err("should return unexpected end of stream");
+        assert!(matches!(err, AppendError::UnexpectedEndOfStream));
+    }
+
+    #[tokio::test]
+    async fn send_serialization_error() {
+        use crate::model::append_rows_request::MissingValueInterpretation;
+
+        let (req_tx, _req_rx) = mpsc::unbounded_channel();
+        let invalid: MissingValueInterpretation = serde_json::from_str("\"INVALID\"").unwrap();
```
dbolduc, line 255: comment: Interesting. I never thought about how to force this error.

src/bigquery/src/write/builder/append_with_offset.rs
```diff
@@ -236,4 +231,66 @@ mod tests {
         write_handle.await?;
         Ok(())
     }
+
+    #[test]
+    fn send_and_poll_outside_tokio_runtime() {
```
dbolduc: nit: I don't think we need to guarantee our code works outside of the tokio runtime. We are pretty locked in already

src/bigquery/src/write/builder/append_with_offset.rs
```diff
@@ -84,29 +84,24 @@ impl AppendWithOffset {
...
-            .await;
-            let _ = tx.send(res);
-        });
-        AppendFuture::new(rx)
+        if self.req_tx.send(write).is_err() {
+            return AppendFuture::from_future(
+                async move { Err(AppendError::UnexpectedEndOfStream) },
+            );
+        }
+        AppendFuture::from_future(async move {
```
dbolduc, line 98: Ah, I see

dbolduc (approved):
I had to think about whether behavior has changed and I think the answer is no. The request is immediately sent on the channel, serviced by a `Runner` (an already spawned task). So we still get the eager execution without needing to await the future.

So I think this is just a win. Thanks!

src/bigquery/src/write/builder/append_with_offset.rs
haphungw: done

